PR #71, opened 2021-02-09
Add container autoscaling based on CPU utilization

ndmckinley:
### Description
Targets 75% CPU utilization, scales up to 5 containers, or down to 2.  Fewer than 2 would be bad because it leaves the entire application down with a single failure, and higher than 5 is just unlikely.  Might increase that number for launch / public announcement?

### Checklist

### Issue(s)
[Fixes] #<issue_number>

### DIFF
--- a/infra/containers.yaml
+++ b/infra/containers.yaml
@@ -113,3 +113,25 @@ Resources:
               awslogs-region: us-west-2
               awslogs-stream-prefix: uatsrv
           Essential: true
+  autoscalingtarget:
+    Type: AWS::ApplicationAutoScaling::ScalableTarget
+    Properties:
+      MinCapacity: 2
+      MaxCapacity: 5
+      ResourceId: !Join ['/', [service, !Ref ecscluster, !GetAtt ecsservice.Name]]
+      ScalableDimension: ecs:service:DesiredCount
+      ServiceNamespace: ecs
+      RoleARN: # This is where the ARN of the SLR goes.
+  autoscalingpolict:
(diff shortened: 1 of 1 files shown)


infra/containers.yaml
```diff
@@ -113,3 +113,25 @@ Resources:
...
+      MinCapacity: 2
+      MaxCapacity: 5
+      ResourceId: !Join ['/', [service, !Ref ecscluster, !GetAtt ecsservice.Name]]
+      ScalableDimension: ecs:service:DesiredCount
+      ServiceNamespace: ecs
+      RoleARN: arn:aws:iam::405662711265:role/aws-service-role/ecs.application-autoscaling.amazonaws.com/AWSServiceRoleForApplicationAutoScaling_ECSService
+  autoscalingpolicy:
+    Type: AWS::ApplicationAutoScaling::ScalingPolicy
+    Properties:
+      PolicyName: UATAutoScalingPolicy
```
hmhsu, line 128: **Question** (non-blocking): What's the naming convention for these names? I've seen lower case (`uat`, `uatsrv`) as well as camelcase like this one.

hmhsu (approved):

infra/containers.yaml
ndmckinley: Haha, sorry!  Basically I try to use upper case if I think it's going to be in a UI, and lowercase if I think it's going to go in a log.  I have guessed wrong a few times - a cleanup is probably in order at the end of the week.

PR #73, opened 2021-02-09
Turn off database logSql

hmhsu:
### Description
Turn off database logSql to avoid confusion in test output. This should be fine since we use Ebean to manage our database interaction.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/conf/application.conf
+++ b/universal-application-tool-0.0.1/conf/application.conf
@@ -346,7 +346,7 @@ db {
 
   # You can turn on SQL logging for any datasource
   # https://www.playframework.com/documentation/latest/Highlights25#Logging-SQL-statements
-  default.logSql=true
+  default.logSql=false
 }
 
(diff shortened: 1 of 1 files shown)


ndmckinley:
Oh, is that what that does?  I thought that logged the actual sql statements being sent - that's not so useful, then, good to turn it off.

ndmckinley (approved):

hmhsu:
> Oh, is that what that does? I thought that logged the actual sql statements being sent - that's not so useful, then, good to turn it off.

Yeah, it actually does multiple things according the [documentation](https://www.playframework.com/documentation/2.8.x/AccessingAnSQLDatabase#How-to-configure-SQL-log-statement). It can log connection status, SQL statements, and ResultSet status. We do not configure the logging level in `logback.xml` so I think the default behavior is simply log errors if any.
I think in case that there is any real errors, the error should be surfaced to our application code so this logging is not necessary.

natsid:
Thanks, Hung-Ming! Test output is so much better now.

PR #79, opened 2021-02-10
Add monitoring to ELB and DB, add alert channels, do not populate them.

ndmckinley:
### Description
This adds a handful of alerts, described in comments in the file.

Page on:
* no db connections
* no healthy containers
* 5xx > 5%

Send ticket on:
* lots of 4xxs
* db free space low
* db cpu usage high

This doesn't actually hook up any sms or email or pager hooks, but you can do that in the UI - also we don't want those until we're live anyway.

### Checklist

### Issue(s)
[Fixes] #<issue_number>

### DIFF
--- a/infra/monitoring.yaml
+++ b/infra/monitoring.yaml
@@ -50,6 +50,9 @@ Resources:
             Metric:
               MetricName: HTTPCode_Target_2XX_Count
               Namespace: AWS/ApplicationELB
+              Dimensions:
+                - Name: LoadBalancer
+                  Value: !Ref LBName
             Period: 60
             Stat: Sum
           ReturnData: false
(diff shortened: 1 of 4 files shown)


infra/monitoring.yaml
```diff
@@ -0,0 +1,225 @@
...
+          Expression: (total4xx / (total2xx + total3xx + total4xx))
+          ReturnData: true
+        - Id: total2xx
+          MetricStat:
+            Metric:
+              MetricName: HTTPCode_Target_2XX_Count
+              Namespace: AWS/ApplicationELB
+              Dimensions:
+                - Name: LoadBalancer
+                  Value: !Ref LBName
```
hmhsu, line 98: **Question**: We are not referencing this in above alert. Do we need it? What is the difference?

hmhsu (approved):

infra/monitoring.yaml
ndmckinley: Oh, we *do* need it there.  I was *wondering* why that was showing up as "not enough data".  Thanks, good eye!

PR #82, opened 2021-02-10
Add QuestionDefinition and test

carolinedanzi:
### Description
Added a basic QuestionDefinition with only getters, and a test.

Paired with: @AWBaumann 

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -1,20 +1,31 @@
 package services.question;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import java.util.Locale;
 
 public class AddressQuestionDefinition extends QuestionDefinition {
-    @Override 
-    public QuestionType questionType() {
-      return QuestionType.ADDRESS;
-    }
-     
-    @Override
-    public ImmutableMap<String, Class> getScalars() {
-        return ImmutableMap.of(
-            "street", String.class,
-            "unit", String.class,
-            "city", String.class,
-            "state", String.class,
-            "zip", String.class);
-    }
+
+  public AddressQuestionDefinition(
+      String id,
+      String version,
+      String name,
+      String path,
+      String description,
+      ImmutableMap<Locale, String> questionText,
+      ImmutableMap<Locale, String> questionHelpText,
+      ImmutableSet<String> tags) {
+    super(id, version, name, path, description, questionText, questionHelpText, tags);
+  }
+
+  @Override
+  public QuestionType getQuestionType() {
+    return QuestionType.ADDRESS;
+  }
+
+  @Override
+  public ImmutableMap<String, Class> getScalars() {
+    return ImmutableMap.of(
+        "street", String.class, "city", String.class, "state", String.class, "zip", String.class);
+  }
 }
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -1,0 +1,20 @@
+package services.question;
+
+import com.google.common.collect.ImmutableMap;
+
+public class AddressQuestionDefinition extends QuestionDefinition {
+    @Override 
+    public QuestionType questionType() {
+      return QuestionType.ADDRESS;
+    }
+     
+    @Override
+    public ImmutableMap[] getScalars() {
+        return ImmutableMap.of(
+            "street", String,
+            "unit", String,
+            "city", String,
+            "state", String,
+            "zip", String);
+    }
+}
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -9,7 +9,7 @@ public class AddressQuestionDefinition extends QuestionDefinition {
     }
      
     @Override
-    public ImmutableMap[] getScalars() {
+    public ImmutableMap<String, Class> getScalars() {
         return ImmutableMap.of(
             "street", String.class,
             "unit", String.class,
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -11,10 +11,10 @@ public class AddressQuestionDefinition extends QuestionDefinition {
     @Override
     public ImmutableMap[] getScalars() {
         return ImmutableMap.of(
-            "street", String,
-            "unit", String,
-            "city", String,
-            "state", String,
-            "zip", String);
+            "street", String.class,
+            "unit", String.class,
+            "city", String.class,
+            "state", String.class,
+            "zip", String.class);
     }
 }
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -15,7 +15,7 @@ public class AddressQuestionDefinition extends QuestionDefinition {
       ImmutableMap<Locale, String> questionText,
       ImmutableMap<Locale, String> questionHelpText,
       ImmutableSet<String> tags) {
-    super(id, version, name, path, description, questionText, questionHelpText, tags);
+    super(id, version, name, path, description, questionText, questionHelpText);
   }
 
   @Override
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -2,6 +2,7 @@ package services.question;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.Locale;
+import java.util.Optional;
 
 public class AddressQuestionDefinition extends QuestionDefinition {
 
@@ -12,7 +13,7 @@ public class AddressQuestionDefinition extends QuestionDefinition {
       String path,
       String description,
       ImmutableMap<Locale, String> questionText,
-      ImmutableMap<Locale, String> questionHelpText) {
+      Optional<ImmutableMap<Locale, String>> questionHelpText) {
     super(id, version, name, path, description, questionText, questionHelpText);
   }
 
--- a/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
@@ -1,7 +1,6 @@
 package services.question;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 import java.util.Locale;
 
 public class AddressQuestionDefinition extends QuestionDefinition {
@@ -13,8 +12,7 @@ public class AddressQuestionDefinition extends QuestionDefinition {
       String path,
       String description,
       ImmutableMap<Locale, String> questionText,
-      ImmutableMap<Locale, String> questionHelpText,
-      ImmutableSet<String> tags) {
+      ImmutableMap<Locale, String> questionHelpText) {
     super(id, version, name, path, description, questionText, questionHelpText);
   }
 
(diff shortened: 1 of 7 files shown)


universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+    return this.tags;
+  }
+
+  /** Get the type of this question. */
+  public QuestionType getQuestionType() {
+    return QuestionType.TEXT;
+  }
+
+  /** Get a set of scalars stored by this question definition. */
+  public ImmutableSet<String> getScalars() {
```
AWBaumann, line 87: We'll probably change this to an ImmutableMap<String, Object> or something like that so we can enforce types.
Then this would be be...

```
public ImmutableMap<String, Object> getScalars() {
    return ImmutableMap.of("text",  String);
}
```

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /** Get a human-readable description for the data this question collects. */
+  public String getDescription() {
+    return this.description;
+  }
+
+  /** Localized question text. Keys are ISO language codes, and values are the translated text. */
+  public ImmutableMap<Locale, String> getQuestionText() {
+    return this.questionText;
+  }
```
bion, line 65: Suggestion (blocking): for this and `getQuestionHelpText()` I suggest favoring encapsulation, absent a reason not to.

A possible approach:
```
public String getQuestionText(Locale locale) {
  if (!this.questionText.containsKey(locale)) {
    throw new RuntimeException("locale not found: " + locale);
  }

  return this.questionText.get(localCode);
}
```

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+      ImmutableMap<Locale, String> questionHelpText,
+      ImmutableSet<String> tags) {
+    this.id = id;
+    this.version = version;
+    this.name = name;
+    this.path = path;
+    this.description = description;
+    this.questionText = questionText;
+    this.questionHelpText = questionHelpText;
+    this.tags = tags;
```
bion, line 34: Suggestion (blocking): for anything that isn't explicitly nullable please include a null check.

```

this.id = checkNotNull(id);
```

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+  public ImmutableMap<Locale, String> getQuestionHelpText() {
+    return this.questionHelpText;
+  }
+
+  /**
+   * Get a set of admin-visible tags for sorting and searching (ex: "General", "Financial", etc.)
+   */
+  public ImmutableSet<String> getTags() {
+    return this.tags;
+  }
```
bion, line 79: Suggestion (non-blocking): I don't think we have an immediate use-case for this for the first several milestones. Since it's unclear how it'll be consumed I recommend not including it for the time being. We may well end up deciding we want a `Tag` class or something instead of passing strings around for them.

universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public QuestionType getQuestionType() {
+    return QuestionType.ADDRESS;
+  }
+
+  @Override
+  public ImmutableMap<String, Class> getScalars() {
+    return ImmutableMap.of(
+        "street", String.class, "city", String.class, "state", String.class, "zip", String.class);
```
bion, line 29: Issue (blocking): Java switch statements don't allow instances of `Class`, i.e. the following code doesn't compile:

```java
switch(questionDef.getScalars().get("street")) {
  case String.class: {
    ...
    break;
  }
}
```

Suggestion: another option could be to use an enum for the scalar types.

bion: It has been 0 days since I've been disappointed in Java.

AWBaumann: I considered the enum approach as well. I like it since we only want to support Java primitives and not any Class (at least for now)

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
carolinedanzi: Done! Though I flipped the condition, let me know if you have a preference there

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
carolinedanzi: Done

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
carolinedanzi: Great point! I removed tags

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,85 @@
...
+      String description,
+      ImmutableMap<Locale, String> questionText,
+      ImmutableMap<Locale, String> questionHelpText) {
+    this.id = checkNotNull(id);
+    this.version = checkNotNull(version);
+    this.name = checkNotNull(name);
+    this.path = checkNotNull(path);
+    this.description = checkNotNull(description);
+    this.questionText = checkNotNull(questionText);
+    this.questionHelpText = checkNotNull(questionHelpText);
```
natsid, line 32: issue: I think questionHelpText should be nullable. Some questions may not require additional text beyond the questionText.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -59,23 +57,20 @@ public String getDescription() {
...
-   * Localized question help text. Keys are ISO language codes, and values are the translated text.
-   */
-  public ImmutableMap<Locale, String> getQuestionHelpText() {
-    return this.questionHelpText;
+  /** Get the question text for the given locale. */
+  public String getQuestionText(Locale locale) {
+    if (this.questionText.containsKey(locale)) {
+      return this.questionText.get(locale);
+    }
+    throw new RuntimeException("Locale not found: " + locale);
```
bion, line 65: Suggestion (non-blocking): please add a space between the `if` and the `throw` 🙏🏻

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
bion: LGTM 👍🏻

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
carolinedanzi: That's a good point - would it be null, or just an empty map?

bion: Suggestion: an `Optional<ImmutableMap<yada, yada>>` would make it explicit whether the help text was intentionally left blank or simply hadn't been added yet.

universal-application-tool-0.0.1/app/services/question/AddressQuestionDefinition.java
```diff
@@ -0,0 +1,36 @@
+package services.question;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Locale;
+
+public class AddressQuestionDefinition extends QuestionDefinition {
```
bion, line 6: Issue (blocking): this class as well as its superclass represent a question type, rather than a specific question definition.

Suggestion: rename `*QuestionDefinition` to `*QuestionTypeDefinition`

bion: Errr wait no they don't, sorry.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+  public String getVersion() {
+    return this.version;
+  }
+
+  /** Get the name of this question. */
+  public String getName() {
+    return this.name;
+  }
+
+  /** Get the path of this question's parent, in JSON notation. */
```
bion, line 50: Suggestion (non-blocking): recommend including in comment that the question definition paths are what define the structure of an applicant's ApplicationData

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+  public String getPath() {
+    return this.path;
+  }
+
+  /** Get a human-readable description for the data this question collects. */
```
bion, line 55: Suggestion (non-blocking): recommend mentioning that the description is for the admin only, and that's why it's not localized.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+
+    throw new RuntimeException("Locale not found: " + locale);
+  }
+
+  /** Get the type of this question. */
+  public QuestionType getQuestionType() {
+    return QuestionType.TEXT;
+  }
+
+  /** Get a set of scalars stored by this question definition. */
```
bion, line 83: Suggestion (non-blocking): update comment to match that this returns a map.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
bion: Question (non-blocking): is the path to the question's parent or to the question itself? I'd think the question itself e.g. for the applicant's name I'd expect the question definition path to be `applicant.name`.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+  public String getId() {
+    return this.id;
+  }
+
+  /** Get the system version this question is pinned to. */
+  public String getVersion() {
+    return this.version;
+  }
+
+  /** Get the name of this question. */
```
bion, line 45: Suggestion (non-blocking): recommend mentioning that the name is for the admin only, and that's why it's not localized.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,91 @@
...
+    return QuestionType.TEXT;
+  }
+
+  /** Get a set of scalars stored by this question definition. */
+  public ImmutableMap<String, ScalarType> getScalars() {
+    return ImmutableMap.of("text", ScalarType.STRING);
+  }
+
+  public ScalarType getScalarType(String key) {
+    return this.getScalars().get(key);
```
bion, line 89: Suggestion (blocking): since this can return `null` recommend returning it wrapped in an `Optional.of` to force consumers to check for presence and avoid null pointer exception.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
bion: Maybe we discussed using the question name as the last segment of the path? If that's the case I'd advise against it because the name is meant to be human readable whereas the path should be optimized for machine-readability e.g. should not contain spaces and punctuation.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
bion: Alternatively could throw a runtime exception if the key doesn't exist, that'd reduce code complexity for the consumer while avoiding annoying to track down null pointer bugs

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
AWBaumann: Yeah, I think we were assuming that the name would be a valid camelCase string, but that may be a bit to code-y for admins so I agree that the name should be human readable in which case the path should be the full path instead of the path to the parent.

Plus, we can always get the parent given the full path - or we can just add a simple utility method.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -0,0 +1,102 @@
...
+  /** Get the question help text for the given locale. */
+  public String getQuestionHelpText(Locale locale) {
+    if (!this.questionHelpText.isPresent()) {
+      return "";
+    }
+    if (this.questionHelpText.get().containsKey(locale)) {
+      return this.questionHelpText.get().get(locale);
+    }
+    throw new RuntimeException("Locale not found: " + locale);
+  }
```
bion, line 87: nit (non-blocking): add a newline between 82,83 and 85,86

bion (changes requested):
Looks good! Ready to merge after resolving https://github.com/seattle-uat/universal-application-tool/pull/82/files#r573988626

natsid (approved):

bion (approved):

Issue #99, opened 2021-02-10
Create Question service facade


bion:
Create a Java interface for the internal question service.

Something along the lines of

```java
interface QuestionService {
  // create, update, list, get question definitions
  // get question definition for path
}
```

Additional context
This is how the question service controllers, program service, and applicant service will interact with the question service.

Done when
Done when the interface is available in the UAT server.

PR #115, opened 2021-02-10
Create QuestionService interface

AWBaumann:
### Description
Added QuestionService interface with basic get, set, and lookup abstract methods.

### Checklist

### Issue(s)
Eventually [Fixes] #99

### DIFF
--- a/universal-application-tool-0.0.1/app/services/question/InvalidPathException.java
+++ b/universal-application-tool-0.0.1/app/services/question/InvalidPathException.java
@@ -1,0 +1,7 @@
+package services.question;
+
+public class InvalidPathException extends Exception {
+  public InvalidPathException(String pathName) {
+    super("Path not found: " + pathName);
+  }
+}
--- a/universal-application-tool-0.0.1/app/services/question/PathType.java
+++ b/universal-application-tool-0.0.1/app/services/question/PathType.java
@@ -1,0 +1,7 @@
+package services.question;
+
+enum PathType {
+  // INTERMEDIATE,
+  QUESTION,
+  SCALAR
+}
--- a/universal-application-tool-0.0.1/app/services/question/PathType.java
+++ b/universal-application-tool-0.0.1/app/services/question/PathType.java
@@ -1,7 +1,6 @@
 package services.question;
 
 enum PathType {
-  // INTERMEDIATE,
   QUESTION,
   SCALAR
 }
--- a/universal-application-tool-0.0.1/app/services/question/QuestionService.java
+++ b/universal-application-tool-0.0.1/app/services/question/QuestionService.java
@@ -1,5 +1,6 @@
 package services.question;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.util.Locale;
 import java.util.Optional;
@@ -10,22 +11,25 @@ public interface QuestionService {
    * Creates a new Question Definition. Returns a QuestionDefinition object on success and null on
    * failure.
    *
-   * <p>This will fail if: - The path provided already resolves to a QuestionDefinition or Scalar.
+   * <p>This will fail if he path provided already resolves to a QuestionDefinition or Scalar.
    *
    * <p>NOTE: This does not update the version.
    */
-  public abstract Optional<QuestionDefinition> create(QuestionDefinition definition);
+  Optional<QuestionDefinition> create(QuestionDefinition definition);
 
   /**
    * Adds a new translation to an existing question definition. Returns true if the write is
    * successful.
    *
-   * <p>The write will fail if: - The path does not resolve to a QuestionDefinition. - A translation
-   * with that Locale already exists for a given question path.
+   * <p>The write will fail if:
+   *
+   * <p>- The path does not resolve to a QuestionDefinition.
+   *
+   * <p>- A translation with that Locale already exists for a given question path.
    *
    * <p>NOTE: This does not update the version.
    */
-  public abstract boolean addTranslation(
+  boolean addTranslation(
       String path, Locale Locale, String questionText, Optional<String> questionHelpText);
 
   /**
@@ -33,28 +37,45 @@ public interface QuestionService {
    *
    * <p>NOTE: This updates the service and question versions.
    */
-  public abstract QuestionDefinition update(QuestionDefinition definition);
+  QuestionDefinition update(QuestionDefinition definition);
 
   /** Checks whether a specific path is valid. */
-  public abstract boolean isValid(String pathString);
-
-  /** Gets the question definition for a given path. */
-  public abstract Optional<QuestionDefinition> getQuestionDefinition(String pathString);
+  boolean isValid(String pathString);
 
   /**
-   * Returns all of the scalar properties for a given path. If the path is invalid this returns
-   * Optional.empty().
+   * Gets the question definition for a given path.
+   *
+   * <p>If the path is to a QUESTION, it will return that.
+   *
+   * <p>If the path is to a SCALAR, it will return the parent QuestionDefinition for that Scalar.
+   *
+   * <p>If the path is invalid it will throw an InvalidPathException.
    */
-  public abstract Optional<ImmutableMap<String, ScalarType>> getPathScalars(String pathString);
+  QuestionDefinition getQuestionDefinition(String pathString);
 
-  /** Gets the type of the node if it exist. Otherwise returns Optional.empty(). */
-  public abstract Optional<PathType> getPathType(String pathString);
+  /**
+   * Returns all of the scalar properties for a given path.
+   *
+   * <p>If the path is to a QUESTION, it will return the question's scalar objects.
+   *
+   * <p>If the path is to a SCALAR, it will return a single scalar.
+   *
+   * <p>If the path is invalid it will throw an InvalidPathException.
+   */
+  ImmutableMap<String, ScalarType> getPathScalars(String pathString);
+
+  /**
+   * Gets the type of the node if it exist.
+   *
+   * <p>If the path is invalid it will throw an InvalidPathException.
+   */
+  PathType getPathType(String pathString);
 
   /** Returns all question definitions for this version. */
-  public abstract QuestionDefinition[] getAllQuestions();
+  ImmutableList<QuestionDefinition> getAllQuestions();
 
   /**
    * Returns a map of full path to ScalarType for all scalars referenced in the question service.
    */
-  public abstract ImmutableMap<String, ScalarType> getFullyQualifiedScalars();
+  ImmutableMap<String, ScalarType> getFullyQualifiedScalars();
 }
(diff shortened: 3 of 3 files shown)


universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Creates a new Question Definition. Returns a QuestionDefinition object on success and null on
+   * failure.
+   *
+   * <p>This will fail if: - The path provided already resolves to a QuestionDefinition or Scalar.
+   *
+   * <p>NOTE: This does not update the version.
+   */
+  public abstract Optional<QuestionDefinition> create(QuestionDefinition definition);
```
bion, line 17: `public` and `abstract` are redundant for an interface and can be removed.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+   * Adds a new translation to an existing question definition. Returns true if the write is
+   * successful.
+   *
+   * <p>The write will fail if: - The path does not resolve to a QuestionDefinition. - A translation
+   * with that Locale already exists for a given question path.
+   *
+   * <p>NOTE: This does not update the version.
+   */
+  public abstract boolean addTranslation(
+      String path, Locale Locale, String questionText, Optional<String> questionHelpText);
```
bion, line 29: Question: why `path` rather than the qd id?

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * <p>NOTE: This updates the service and question versions.
+   */
+  public abstract QuestionDefinition update(QuestionDefinition definition);
+
+  /** Checks whether a specific path is valid. */
+  public abstract boolean isValid(String pathString);
+
+  /** Gets the question definition for a given path. */
+  public abstract Optional<QuestionDefinition> getQuestionDefinition(String pathString);
```
bion, line 42: Suggestion (): The docstring could use a little more info on this one. If the path is to a scalar, does it return the scalar's question? I like that behavior but could understand the objection that the method should return the question at the path or throw an exception if the path doesn't point to a question.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Returns all of the scalar properties for a given path. If the path is invalid this returns
+   * Optional.empty().
```
bion, line 46: Suggestion: rather than return `Optional.empty()` throw a `InvalidPathException` whenever the path is invalid

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public abstract Optional<ImmutableMap<String, ScalarType>> getPathScalars(String pathString);
+
+  /** Gets the type of the node if it exist. Otherwise returns Optional.empty(). */
+  public abstract Optional<PathType> getPathType(String pathString);
+
+  /** Returns all question definitions for this version. */
+  public abstract QuestionDefinition[] getAllQuestions();
```
bion, line 54: Suggestion: return a Guava `ImmutableList` rather than a raw Java array

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Returns a map of full path to ScalarType for all scalars referenced in the question service.
+   */
+  public abstract ImmutableMap<String, ScalarType> getFullyQualifiedScalars();
```
bion, line 59: Question: what use case do we have for this? If there isn't one for it yet I recommend not investing in implementing it (same with any others that we can't see an immediate or near-immediate use for).

universal-application-tool-0.0.1/app/services/question/QuestionService.java
AWBaumann: I guess I'm viewing it from an API perspective. Path will always be more human readable, and if I'm hand writing these (without a user interface) I'd much rather write `path: 'applicant.name'` than `id: 'fhwdgads'`

It does raise a question though... once we implement versioning we'll have

| id                | path  | version |
| ----------- | ----------- | ----------- 
| fhwgads      | applicant.name | 1
| hgnsdhjsd   | applicant.name | 2

If an admin wants to add another translation to `applicant.name`will it only be available in version 2 or should it be available to both?

Remember, this is strictly for `addTranslation`. If we already have an existing translation, we would need to use `update` and increment the version of the question.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
AWBaumann: This would essentially give us the empty applicant data model for a given version of the service.

| key | value |
| -------- | -------- |
| applicant.name.first | string |
| applicant.name.last | string |

universal-application-tool-0.0.1/app/services/question/QuestionService.java
AWBaumann: Yeah, I left it intentionally vague because I had the same thought, but I'm fine saying that if the path is to a scalar then this method should return the scalar's QuestionDefinition parent.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
AWBaumann: Of course!

universal-application-tool-0.0.1/app/services/question/QuestionService.java
bion: I'm hesitant to think too far into versioning yet. That being said, my guess is that when we create a new version we'll copy all of the questions from the previous version so they can be edited/deleted totally independent of the previous one. I don't know what the right behavior is with locale updates specifically 🤔 . I'm inclined to hold off on making decisions around it since it's not in the current milestone if we can.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
bion: I'm not sure we have anything in scope for M1 or M2 that needs that. Also (and unfortunately), repeated questions will mean we can't generate a complete empty applicant data model.

AWBaumann: Achievement unlocked: Scope reduced!

universal-application-tool-0.0.1/app/services/question/QuestionService.java
bion: I tried to think through what the update applicant data codepath will look like, and I think I actually like the stricter way of doing things a bit more where `getQuestionDefinition` throws if the path isn't to a question definition.

Seems like it'd be fairly easy behavior to change down the line if we'd like?

universal-application-tool-0.0.1/app/services/question/QuestionService.java
AWBaumann: Additional reasoning, I also wanted to enforce the notion that an admin can add translations without having to re-version of the question. (I believe that this is the only question update that we can safely do without re-versioning.)

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Creates a new Question Definition. Returns a QuestionDefinition object on success and null on
+   * failure.
+   *
+   * <p>This will fail if he path provided already resolves to a QuestionDefinition or Scalar.
+   *
+   * <p>NOTE: This does not update the version.
+   */
+  Optional<QuestionDefinition> create(QuestionDefinition definition);
```
bion, line 18: Thought (non-blocking): This is making me wonder where the right place to detect and surface validation errors is. This *could* return an `ErrorsOr<QuestionDefinition>` where the outer class is something like

```java
class ErrorsOr<T> {
  private final Optional<T> wrapped;
  private final ImmutableMap<String, String> errors;

  ErrorsOr(T wrapped) {
    this.wrapped = Optional.of(wrapped);
    this.errors = ImmutableMap.of();
  }

  ErrorsOr(ImmutableMap<String, String> errors) {
    this.wrapped = Optional.empty();
    this.errors = errors;
  }
  
  // getters and predicate methods
}
```

Or we could do validation at the form-object level.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,76 @@
...
+   * <p>If the path is invalid it will throw an InvalidPathException.
+   */
+  ImmutableMap<String, ScalarType> getPathScalars(String pathString);
+
+  /**
+   * Gets the type of the node if it exist.
+   *
+   * <p>If the path is invalid it will throw an InvalidPathException.
+   */
+  PathType getPathType(String pathString);
```
bion, line 72: Suggestion: add the `InvalidPathException` class and add `throws InvalidPathException` to the methods that do in this interface.

universal-application-tool-0.0.1/app/services/question/QuestionService.java
bion: That's a good call.

bion: I'm fine with leaving it path for now but think whoever implements locale updates should have the right to change it to whatever is most convenient

universal-application-tool-0.0.1/app/services/question/InvalidPathException.java
```diff
@@ -0,0 +1,7 @@
+package services.question;
+
+public class InvalidPathException extends Exception {
+  public InvalidPathException(String pathName) {
+    super("Path not found: " + pathName);
```
bion, line 5: Thought (non-blocking): when we do versioning we'll likely want to include the version here as well

bion (approved):

universal-application-tool-0.0.1/app/services/question/PathType.java
```diff
@@ -0,0 +1,7 @@
+package services.question;
+
+enum PathType {
+  // INTERMEDIATE,
```
carolinedanzi, line 4: suggestion: remove commented-out code, we can always add back later :)

universal-application-tool-0.0.1/app/services/question/QuestionService.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.Locale;
+import java.util.Optional;
+
+public interface QuestionService {
+
+  /**
+   * Creates a new Question Definition. Returns a QuestionDefinition object on success and null on
```
carolinedanzi, line 11: nit: `Returns...{@link Optional#empty} on failure`

carolinedanzi (commented):
This is looking really nice!

bion (approved):

PR #117, opened 2021-02-10
Create a basic Program Interface

### LINKED ISSUE #98: Create program service facade

Create a Java `interface` for the internal program service.

Something along the lines of

```java
interface ProgramService {
  ImmutableList<ProgramDefinition> listProgramDefinitions();

  ProgramDefinition getProgramDefinition(String programId);

  // also create, update, destroy ProgramDefinition

 // may also include CRUD for block actions
}
```

**Additional context**
This is how the program service controllers and applicant service will interact with the program service.

**Done when**
Done when the interface is available in the UAT server.

carolinedanzi:
### Description
Adds a very basic interface for programs that only allows getting and create (basic).

Paired with @kendrickt 

### Checklist

### Issue(s)
Fixes #98


natsid (approved):

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -0,0 +1,17 @@
+package services.program;
+
+import com.google.common.collect.ImmutableList;
+
+/** Operations you can perform on Programs. */
```
natsid, line 5: suggestion (non-blocking): "Programs" --> "ProgramDefinitions"

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -0,0 +1,17 @@
...
+public interface ProgramService {
+
+    /** List all programs. */
+    public ImmutableList<ProgramDefinition> listProgramDefinitions();
+
+    /** Get the definition for a given program. */
+    public ProgramDefinition getProgramDefinition(String id);
+
+    /** Create a new program. */
+    public ProgramDefinition createProgram(String name, String description);
```
bion, line 15: See comment [here](https://github.com/seattle-uat/universal-application-tool/pull/115/files#r574175484) on the possibility of surfacing validation errors here.

natsid (approved):

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -0,0 +1,17 @@
...
+import com.google.common.collect.ImmutableList;
+
+/** Operations you can perform on Programs. */
+public interface ProgramService {
+
+    /** List all programs. */
+    public ImmutableList<ProgramDefinition> listProgramDefinitions();
+
+    /** Get the definition for a given program. */
+    public ProgramDefinition getProgramDefinition(String id);
```
bion, line 12: Question: should the ID be a `long` since that's what the DB generates?

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -0,0 +1,17 @@
+package services.program;
+
+import com.google.common.collect.ImmutableList;
+
+/** Operations you can perform on Programs. */
+public interface ProgramService {
+
+    /** List all programs. */
+    public ImmutableList<ProgramDefinition> listProgramDefinitions();
```
bion, line 9: Suggestion: 💁🏻 All abstract, default, and static methods in an interface are implicitly public, so you can omit the public modifier.

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Makes sense to me!

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Done!

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Good catch, thank you!

bion:
Thoughts on including actions on blocks in this PR?

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Interesting! Are we preferring to return validation errors rather than throw? Or a mix of both? Just trying to wrap my head around these methods

Kendrick and I were wondering what kind of validation errors initial create would give, if it's just setting up a name and description

bion: We don't have to decide this now (forgot to specify this is a non-blocking discussion), but I'm trying to think about where the right place to surface validation errors is. For create program we'll probably want to validate that at the very least they include a name for the program, likely more stuff down the line (department, etc).

carolinedanzi: That makes sense to me, thanks for clarifying! I suppose this will either need to throw if validation fails and something upstream can format the errors nicely, or it should return a validation error class as you suggested

bion: If it throws I'm not sure how it would be able to return multiple errors if there are multiple validation problems.

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -1,17 +1,88 @@
...
+   * Insert a new {@link BlockDefinition} at a specific index for a given program.
+   *
+   * @param programId the ID of the program to update
+   * @param blockIndex the index in the block list for the new block
+   * @param blockName a name for the block to add
+   * @param blockDescription a description of what the questions in the block address
+   * @return the updated {@link ProgramDefinition}
+   */
+  ProgramDefinition addBlockToProgram(
+      long programId, int blockIndex, String blockName, String blockDescription);
```
bion, line 54: Suggestion (non-blocking): recommend removing this. Specifying block order isn't in-scope for M1 and it's unclear what the semantics/UI we'll be supporting is. Even with a simple action like this there are questions that are tricky to answer like if the index collides with an existing block what happens?

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -1,17 +1,88 @@
...
+  /**
+   * Update a {@link BlockDefinition} with a set of questions.
+   *
+   * @param programId the ID of the program to update
+   * @param blockIndex the index of the block to update
+   * @param questionIds an {@link ImmutableList} of questions for the block
+   * @return the updated {@link ProgramDefinition}
+   */
+  ProgramDefinition setBlockQuestions(
+      long programId, int blockIndex, ImmutableList<String> questionIds);
```
bion, line 65: Suggestion (blocking): rename `questionIds` to `questionDefinitionIds`
Suggestion (non-blocking): check if qd IDs are strings or longs

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -1,17 +1,88 @@
...
+  /**
+   * Set the hide {@link Predicate} for a block. This predicate describes under what conditions the
+   * block should be hidden from an applicant filling out the program form.
+   *
+   * @param programId the ID of the program to update
+   * @param blockIndex the index of the block to update
+   * @param predicate the {@link Predicate} for hiding the block
+   * @return the updated {@link ProgramDefinition}
+   */
+  ProgramDefinition setBlockHidePredicate(long programId, int blockIndex, Predicate predicate);
```
bion, line 76: Thought (blocking): possible to reference the block by ID rather than index?

carolinedanzi: I don't think the BlockDefinition has an ID yet - is that something we want?

carolinedanzi: I guess a larger discussion is "are BlockDefinitions their own model / db object, or are they included in ProgramDefinition"?

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Nice catch! Looks like qd IDs are Strings

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Good point - removed

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Good point - I'm used to having validators that just throw on the first issue they see. But it would be a much nicer experience to see all your issues at once instead

universal-application-tool-0.0.1/app/services/program/ProgramService.java
bion: I think they're child resources of the program. Giving them their own ID is convenient for updates and generating URLs imo. It *could* be done using the index but I don't think that's a good practice.

bion (approved):

universal-application-tool-0.0.1/app/services/program/ProgramService.java
carolinedanzi: Makes sense to me - I'll make sure we add the ID field on BlockDefinition

universal-application-tool-0.0.1/app/services/program/ProgramService.java
natsid: I wonder if qd IDs should possibly be `long`s instead, based on our repo code. See [this test](https://github.com/seattle-uat/universal-application-tool/blob/main/universal-application-tool-0.0.1/test/repository/QuestionRepositoryTest.java#L29) that references a question ID.

carolinedanzi:
Messed up git, so re-created with bug fixes in https://github.com/seattle-uat/universal-application-tool/pull/125

PR #118, opened 2021-02-10
Github action to validate correct configuration of cloudformation

ndmckinley:
If you make a change to a file under `infra/`, we will create a CFN changeset, which we will check to see which resources will be replaced.  This will crash if your cfn configuration is invalid or would require permissions we don't have, but will succeed regardless of the impact of your change.

It does print the likely impact of the change, which might help to keep updates safe, but I'll also add some stack rules later to keep us from accidentally deleting the database, for instance.

Hopefully, you'll see the message on this PR, due to the dummy file - which I will remove after testing to show that it works.  There's a chance there will be some back-and-forth since we don't control the creds used here - they may be invalid for this sort of operation.  But it works on my fork: https://github.com/ndmckinley/universal-application-tool/pull/2#issuecomment-777107518

### Description

### Checklist

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -51,6 +51,7 @@ jobs:
 
       - uses: mshick/add-pr-comment@v1
         if: failure() && contains(toJSON(steps.file_changes.outputs.files), 'infra/')
+        name: Send failure comment.
         with:
           message: |
             I noticed that you have created a PR which seems to contain a file that updates CloudFormation.
@@ -60,6 +61,7 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: mshick/add-pr-comment@v1
+        name: Send success comment.
         if: success() && contains(toJSON(steps.file_changes.outputs.files), 'infra/')
         with:
           message: |
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -18,4 +18,49 @@ jobs:
         # Confirm that we get a response on port 9000.
         run: while ! docker run --network host appropriate/curl -v -s --retry-max-time 180 --retry-connrefused http://localhost:9000/ ; do sleep 5; done
         timeout-minutes: 3
+  cloudformation:
+    runs-on: ubuntu-18.04 # for older version of awscli.
+    steps:
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.3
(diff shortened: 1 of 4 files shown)


ndmckinley:
Nice.  Now to remove that file and show that it does not print a message if nothing under infra/ is changed...

ndmckinley:
Nice.  Ready for review by @hmhsu and optional review by @bion.  Especially interested in what instructions might be useful and in neat ideas for exploring whether a change is likely to be safe.

hmhsu:
> The following objects will be recreated (or conditionally recreated) - please take a look and make sure
> this is correct before submitting your PR.
> 
> ```
> ["autoscalingpolicy","autoscalingtarget","ecscluster","ecsservice","ecstaskdef","loggroup","postgresdb","droptraffic","lbrule","lbtarget","listener","publiclb","containersecgroup","dbsecgroup","publiclbsecgroup"]
> ```

Want to confirm if I understand this correctly: this outputs all resources that could potentially be recreated by *any* change, not by the *proposed* change. Is that correct? Since I see `postgresdb` listed while the test file is empty.

Another interesting thing I found is that this does not catch all the errors, e.g. there is a typo in `monitoring.yaml` causing the change set status to be `failed`, but the test still passes. See below command for repo.

```shell
$ aws cloudformation describe-change-set --stack-name uat --change-set-name auto-changeset-pr-118 | tail -n 14 | head -n 4
    "CreationTime": "2021-02-10T23:45:07.227000+00:00",
    "ExecutionStatus": "UNAVAILABLE",
    "Status": "FAILED",
    "StatusReason": "Template format error: Unresolved resource dependencies [ticketalert] in the Resources block of the template",
```

bion (approved):

ndmckinley:
@hmhsu - thanks, you're right, I need to check the change set statuses as well - will do!  And those are the list of resources that the system believes will be recreated as part of this change.  It's wildly overestimating - I'm not sure why!  I'll want to look into that.

ndmckinley:
Okay cool!  @hmhsu would it be all right with you if I merge this as-is to prevent issues like the typo from recurring, then investigate why it's overestimating the impact of changes?

hmhsu (approved):
Yeah, great work! Thanks for looking into this.

ndmckinley:
Okay - that removed a *few* of them - I'll merge this - but, odd, there are still a few shown as possible updates that are not really going to be updated.

PR #122, opened 2021-02-11
Rely on generated IDs only

natsid:
### Description
Stop setting an explicit ID for applicant in repository test. Instead, we should only ever rely on generated IDs. This avoids conflicts in the database.

### Checklist

### Needs review from
ONEOF: [hmhsu, ndmckinley]


ndmckinley:
It looks like there's a compile error:

> Error:  /usr/src/universal-application-tool-0.0.1/app/services/program/BlockDefinition.java:53:1: Parameter type com.google.common.collect.ImmutableList<java.lang.String> of setter method should be com.google.common.collect.ImmutableList<services.question.QuestionDefinition> to match getter services.program.BlockDefinition.questionDefinitions, or it should be a type that can be passed to ImmutableList.copyOf to produce com.google.common.collect.ImmutableList<services.question.QuestionDefinition>

ndmckinley:
Never mind, that's in another PR!  I'll wait for that to fix and you to rebase.

ndmckinley (approved):

PR #143, opened 2021-02-12
Deploy to prod when we merge to main.

ndmckinley:
### Description
Just run `bin/deploy-prod` - the ubuntu-18 environment runs it just fine, which is great and lucky for us.

### Checklist

### DIFF
--- a/.github/workflows/docker_build.yaml
+++ b/.github/workflows/docker_build.yaml
@@ -19,3 +19,21 @@ jobs:
           tags: uat:latest
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
+
+  deploy_to_cfn:
+    runs-on: ubuntu-18.04 # for older version of awscli.
+    steps:
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.3
+
+      - name: Checkout
+        if: contains(toJSON(steps.file_changes.outputs.files), 'infra/')
+        uses: actions/checkout@v2
+
+      - name: Run deploy-prod.
+        if: contains(toJSON(steps.file_changes.outputs.files), 'infra/')
+        id: deploy_prod
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.ACCESS_KEY_ID }}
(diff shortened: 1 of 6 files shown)


.github/workflows/docker_build.yaml
```diff
@@ -19,3 +19,21 @@ jobs:
...
+    steps:
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.3
+
+      - name: Checkout
+        if: contains(toJSON(steps.file_changes.outputs.files), 'infra/')
+        uses: actions/checkout@v2
+
+      - name: Run deploy-prod.
+        if: contains(toJSON(steps.file_changes.outputs.files), 'infra/')
```
hmhsu, line 34: **Suggestion** (non-blocking): Might worth deploying even if we don't change the stack, i.e. push a new image?

hmhsu (approved):

.github/workflows/docker_build.yaml
ndmckinley: That's a good idea, but I think we want a much less invasive change there - just want to update the container.  Let me see what I can do there...

ndmckinley: Great, got it.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build_container:
+    runs-on: ubuntu-18.04 # for older version of awscli.
+    steps:
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.3
+
+      - name: Checkout
+        if: contains(toJSON(steps.file_changes.outputs.files), 'infra/')
```
hmhsu, line 15: **Question**: Do we now still need this condition? We checkout no matter the change is in `infra/` or not.

hmhsu (approved):
Looks good. Thanks!

.github/workflows/push.yaml
ndmckinley: Hm, thanks!  You're right.

ndmckinley:
Okay, here we goooo

PR #144, opened 2021-02-12
The ability to bring up a full dev stack which will automatically tear itself down.

ndmckinley:
### Description
There is one TODO - the actual ARN for the role that needs to bring up the stack needs to change, or else the perms for that role need to be changed.  Waiting on John for that.  So I figured I'd send this for review and then edit that before submitting.

Let's see if that testing github action works!  (Will add reviewers once I see whether that needs changes)

### Checklist

### DIFF
--- a/infra/iam_for_stack_ttl.yaml
+++ b/infra/iam_for_stack_ttl.yaml
@@ -1,0 +1,54 @@
+# provided by AWS quickstart - but not available hosted on s3, so reproduced here from
+# https://github.com/aws-quickstart/quickstart-examples/blob/main/samples/cloudformation-stack-ttl/templates/cloudformation-stack-ttl.yaml
+# Applied by the account admins since we lack permissions for creating IAM roles.
+AWSTemplateFormatVersion: '2010-09-09'
+Description: ''
+Resources:
+  DeleteCFNLambdaExecutionRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+        - Effect: "Allow"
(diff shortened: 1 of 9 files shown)


ndmckinley:
Cool - I think that's right.

infra/ttl.yaml
```diff
@@ -0,0 +1,125 @@
...
+                    cfnresponse.send(event, context, status, {})
+                else:
+                    ttl = event['ResourceProperties']['ttl']
+                    responseData = {}
+                    responseData['cron_exp'] = deletion_time(ttl)
+                    cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
+            except Exception as e:
+                logging.error('Exception: %s' % e, exc_info=True)
+                status = cfnresponse.FAILED
+                cfnresponse.send(event, context, status, {}, None)
```
hmhsu, line 115: **Question**: Based on reading this [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html), the fifth argument is default to `None`. Curious why we are setting it here, but not in other `send` invocation.

infra/iam_for_stack_ttl.yaml
```diff
@@ -0,0 +1,54 @@
+# provided by AWS quickstart - but not available hosted on s3, so reproduced here from
```
hmhsu, line 1: **Question**: This file is not referenced in the main `stack.yaml` or anywhere else. Is that expected? Would like to know how it works.

hmhsu (approved):

ndmckinley:
closing because staging does what this does, better!

PR #150, opened 2021-02-13
Filestorage stack

hmhsu:
### Description
 - Add S3 as file storage to our AWS stack.
 - Add a `files` table in DB to keep track of file name, i.e. key.
 - Add a S3Client class to talk to S3
 - Add a repository to interact with DB and also upload file to S3.

### Checklist

### DIFF
--- a/infra/filestorage.yaml
+++ b/infra/filestorage.yaml
@@ -1,0 +1,47 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: ''
+Resources:
+  key:
+    Type: AWS::KMS::Key
+    Properties:
+      EnableKeyRotation: true
+      Description: The key used to encrypt/decrypt secure uploads to the UAT.
+      KeySpec: SYMMETRIC_DEFAULT
+      KeyPolicy:
+        Version: '2012-10-17'
+        Id: secureuploads-key
+        Statement:
+        - Sid: Enable IAM User Permissions
+          Effect: Allow
+          Principal:
+            AWS:
+              Fn::Join:
+              - ''
+              - - 'arn:aws:iam::'
+                - Ref: AWS::AccountId
+                - :root
+          Action: kms:*
+          Resource: '*'
+  bucket:
+    Type: 'AWS::S3::Bucket'
+    Properties:
+      BucketName: !Sub 'secureuploads-${AWS::Region}-${AWS::AccountId}'
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: 'aws:kms'
+              KMSMasterKeyID: !GetAtt key.Arn
+            BucketKeyEnabled: true
+      PublicAccessBlockConfiguration:
+        # This totally prevents anyone outside the AWS account using this bucket.
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+    DeletionPolicy: Delete
+Outputs:
+  bucket:
+    Description: The location of the bucket for secure uploads.
+    Value: !Ref bucket
+    Export:
+      Name: !Sub '${AWS::StackName}-SecureBucket'
--- a/infra/filestorage.yaml
+++ b/infra/filestorage.yaml
@@ -1,5 +1,5 @@
 AWSTemplateFormatVersion: '2010-09-09'
-Description: ''
+Description: 'We have a S3 bucket encrypted with "Customer master keys" (CMKs). The bucket is limited access to from within our VPC. This means the console does not have access to it, either.'
 Parameters:
   S3VPCEndpoint:
     Description: The vpc endpoint from which UAT connects to S3.
@@ -36,7 +36,8 @@ Resources:
               KMSMasterKeyID: !GetAtt key.Arn
             BucketKeyEnabled: true
       PublicAccessBlockConfiguration:
-        # This totally prevents anyone outside the AWS account using this bucket.
+        # This prevents anyone outside the AWS account using this bucket, but does not limit roles within the AWS account from accessing.
+        # This is only effective when the bucket is public without this setting. Our bucket policy below makes our bucket non-public so this setting has no effect.
         BlockPublicAcls: true
         BlockPublicPolicy: true
         IgnorePublicAcls: true
--- a/infra/filestorage.yaml
+++ b/infra/filestorage.yaml
@@ -26,6 +26,19 @@ Resources:
                 - :root
           Action: kms:*
           Resource: '*'
+        - Sid: Enable UAT Server Permissions
+          Effect: Allow
+          Principal:
+            AWS: arn:aws:iam::405662711265:role/AmazonECSTaskS3BucketRole
+          Action:
+            - kms:Encrypt
+            - kms:Decrypt
+            - kms:ReEncryptFrom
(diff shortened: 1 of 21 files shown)


infra/filestorage.yaml
```diff
@@ -0,0 +1,69 @@
...
+      KeySpec: SYMMETRIC_DEFAULT
+      KeyPolicy:
+        Version: '2012-10-17'
+        Id: secureuploads-key
+        Statement:
+        - Sid: Enable IAM User Permissions
+          Effect: Allow
+          Principal:
+            AWS:
+              Fn::Join:
```
ndmckinley, line 22: *suggestion*: Can you use `!Join` for consistency?

infra/filestorage.yaml
```diff
@@ -0,0 +1,69 @@
...
+            Effect: Deny
+            Resource: !Join
+              - ''
+              - - 'arn:aws:s3:::'
+                - !Ref bucket
+                - /*
+            Principal: '*'
+            Condition:
+              StringNotEquals:
+                'aws:sourceVpce': !Ref 'S3VPCEndpoint'
```
ndmckinley, line 63: *praise*: heh, nice, that's even better than the public access block.

infra/network.yaml
```diff
@@ -119,6 +119,21 @@ Resources:
...
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal: '*'
+            Action: '*'
+            Resource: '*'
+      RouteTableIds:
+        - !Ref 'privroutetable1'
+        - !Ref 'privroutetable2'
+      ServiceName: !Sub 'com.amazonaws.${AWS::Region}.s3'
```
ndmckinley, line 135: *question*: Do we use AWS::Region elsewhere, or did I hardcode us-west-2 by accident?

universal-application-tool-0.0.1/app/models/StoredFile.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  @Constraints.Required String name;
+
+  public byte[] getContent() {
+    return data;
```
ndmckinley, line 25: *question*: I'm a little concerned about this being in-memory - is that avoidable?  Can we use streams?  Should we use streams?

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -0,0 +1,101 @@
...
+          return CompletableFuture.completedFuture(null);
+        });
+  }
+
+  public boolean enabled() {
+    return config.hasPath(AWS_S3_BUCKET);
+  }
+
+  public void putObject(String key, byte[] data) {
+    throwIfUninitialized();
```
ndmckinley, line 55: *suggestion*: I mildly prefer a conditional reinitialize, which is a little more forgiving.

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private void getTestObject() {
+    byte[] data = getObject("file1");
+    log.info("got data from s3: " + new String(data));
+  }
+
+  private void connect() {
+    bucket = config.getString(AWS_S3_BUCKET);
+
+    Region region = Region.US_WEST_2;
```
ndmckinley, line 98: *question*: is it better if this is configurable?

universal-application-tool-0.0.1/test/repository/StoredFileRepositoryTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.mockito.Mockito.when;
+
+import models.StoredFile;
+import org.junit.Test;
+import play.db.ebean.EbeanConfig;
+
+public class StoredFileRepositoryTest extends WithPostgresContainer {
+
+  @Test
+  public void createFile() {
```
ndmckinley, line 15: *praise*: nice!

ndmckinley (commented):
Looks awesome - let's get that role from John and we'll get this in.

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
hmhsu: Sure thing!

ndmckinley (approved):
Great, and I think you got the role you needed - go for it!

universal-application-tool-0.0.1/app/models/StoredFile.java
bion: Applicant-uploaded files should never have a reason for their contents to be on the UAT server itself.

Admin-uploaded PDF templates, generated PDFs, and CSVs will have to be on the UAT when they're consumed or generated. I agree that it'd be preferable to use streams for those when feasible.

bion: Question for either of you: if we need to shell out for PDF generation, what kind of disk or /tmp directory behavior can we expect with our current infra setup?

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -0,0 +1,108 @@
...
+  private final ApplicationLifecycle appLifecycle;
+  private final Config config;
+  private Region region;
+  private String bucket;
+  private S3Client s3;
+
+  @Inject
+  public AmazonS3Client(ApplicationLifecycle appLifecycle, Config config) {
+    this.appLifecycle = appLifecycle;
+    this.config = config;
```
bion, line 34: nit: use `checkNotNull` on constructor args

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -0,0 +1,108 @@
...
+        () -> {
+          if (s3 != null) {
+            s3.close();
+          }
+          return CompletableFuture.completedFuture(null);
+        });
+  }
+
+  public boolean enabled() {
+    return (config.hasPath(AWS_S3_REGION) && config.hasPath(AWS_S3_BUCKET));
```
bion, line 53: Question: is there a non-error case where this would be called and return false?

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
hmhsu: oops, I have merged the PR. Please leave your comments and I will follow up with another PR.

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
hmhsu: These config are set by passing in environment variables and we do not provide default values. So it is possible to run locally without connecting to S3. This allows us to run dev server without providing AWS access keys/token every time. Do we want to have a test bucket that we connect to in dev server? Should we hardcode that bucket name in our default values?

universal-application-tool-0.0.1/app/models/StoredFile.java
hmhsu: Looks like we get [fixed 20GB temp disk storage](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-task-storage.html)

bion: Ah cool, that's way more than enough.

PR #152, opened 2021-02-13
Configure ObjectMapper for ebean

carolinedanzi:
### Description
Found the correct place for configuring a Jackson ObjectMapper for ebean (it is using a different one than Play), but it doesn't serialize/deserialize correctly (just has an empty JSON object).

### Checklist


carolinedanzi:
Closing since not needed

PR #163, opened 2021-02-17
Add application.test.conf that sets db.default settings for tests. 

kendrickt:
Add application.test.conf that sets db.default settings for tests and add config to build.sbt to use the test conf for java tests.

### Description
The `application.test.conf` file now provides database bindings to an effectively ineffective in-memory postgres database. This database isn't actually compatible with things that actually need a database, but the binding is helpful for tests that `extends WithApplication` that don't need a real database.

The `WithPostgresContainer` is what test classes should extend when they need to use a real database. It has a `@Before` method to truncate tables.

### Checklist


universal-application-tool-0.0.1/conf/application.test.conf
```diff
@@ -0,0 +1,12 @@
+include "application.conf"
```
ndmckinley, line 1: *Question(non-blocking)*: This is going to start the database container for every test, including if you don't need it.  Is that worth it, do you think?  I'm not sure personally.  If you think so, LGTM!

ndmckinley (approved):

universal-application-tool-0.0.1/conf/application.test.conf
kendrickt: oo that's true. I didn't think about that.

I'm gonna script and run some timing tests to see how much time this actually costs.

ndmckinley (approved):

PR #165, opened 2021-02-17
Testing (unit and browser) for basic authentication scheme.

ndmckinley:
### Description
Adds a unit test for an authenticated page (which overrides the security configuration so that prod and test are different).
Adds a browser test for the login page to show that submitting the form logs you in.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/views/BaseHtmlView.java
+++ b/universal-application-tool-0.0.1/app/views/BaseHtmlView.java
@@ -21,17 +21,18 @@ abstract class BaseHtmlView {
   }
 
   protected Tag textField(String fieldName, String labelText) {
-    return label(text(labelText), input().withType("text").withName(fieldName))
+    return label(text(labelText), input().withType("text").withName(fieldName).withId(fieldName))
         .attr("for", fieldName);
   }
 
   protected Tag passwordField(String fieldName, String labelText) {
-    return label(text(labelText), input().withType("password").withName(fieldName))
+    return label(
+            text(labelText), input().withType("password").withName(fieldName).withId(fieldName))
         .attr("for", fieldName);
   }
 
   protected Tag submitButton(String textContents) {
-    return input().withType("submit").withValue(textContents);
+    return input().withType("submit").withId("submit").withValue(textContents);
   }
 
   /**
--- a/universal-application-tool-0.0.1/app/views/BaseHtmlView.java
+++ b/universal-application-tool-0.0.1/app/views/BaseHtmlView.java
@@ -21,18 +21,31 @@ abstract class BaseHtmlView {
   }
 
   protected Tag textField(String fieldName, String labelText) {
-    return label(text(labelText), input().withType("text").withName(fieldName).withId(fieldName))
+    return label(text(labelText), input().withType("text").withName(fieldName))
+        .attr("for", fieldName);
+  }
+
+  protected Tag textField(String id, String fieldName, String labelText) {
+    return label(text(labelText), input().withType("text").withName(fieldName).withId(id))
+        .attr("for", fieldName);
+  }
+
+  protected Tag passwordField(String id, String fieldName, String labelText) {
+    return label(text(labelText), input().withType("password").withName(fieldName).withId(id))
         .attr("for", fieldName);
   }
 
   protected Tag passwordField(String fieldName, String labelText) {
-    return label(
-            text(labelText), input().withType("password").withName(fieldName).withId(fieldName))
+    return label(text(labelText), input().withType("password").withName(fieldName))
         .attr("for", fieldName);
   }
 
   protected Tag submitButton(String textContents) {
-    return input().withType("submit").withId("submit").withValue(textContents);
+    return input().withType("submit").withValue(textContents);
+  }
+
+  protected Tag submitButton(String id, String textContents) {
+    return input().withType("submit").withId(id).withValue(textContents);
   }
 
   /**
--- a/universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
+++ b/universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
@@ -4,13 +4,19 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static play.api.test.Helpers.testServerPort;
 import static play.test.Helpers.*;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.util.List;
 import org.junit.Before;
 import org.junit.Test;
+import org.pac4j.core.client.Client;
 import org.pac4j.core.client.Clients;
 import org.pac4j.core.client.direct.AnonymousClient;
+import org.pac4j.core.client.finder.ClientFinder;
 import org.pac4j.core.config.Config;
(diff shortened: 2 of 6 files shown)


universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -21,17 +21,18 @@ protected Content htmlContent(DomContent... domContents) {
   }
 
   protected Tag textField(String fieldName, String labelText) {
-    return label(text(labelText), input().withType("text").withName(fieldName))
+    return label(text(labelText), input().withType("text").withName(fieldName).withId(fieldName))
```
carolinedanzi, line 24: Out of curiosity, why are these IDs necessary now?

carolinedanzi: Never mind, answered my own question - they are needed in the test

universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
```diff
@@ -40,10 +65,7 @@ public void testAuthenticatedSecurePage() {
         fakeRequest(routes.HomeController.secureIndex())
             .header(Http.HeaderNames.HOST, "localhost:" + testServerPort());
     Result result = route(app, request);
-    // This should work, but it doesn't!
-    // assertThat(result.status()).isEqualTo(HttpConstants.OK);
-
-    // This is the opposite of what we want, but it's what happens now anyway.
-    assertThat(result.status()).isEqualTo(HttpConstants.UNAUTHORIZED);
+    Logger.of("application").info(result.headers().toString());
```
carolinedanzi, line 68: suggestion (blocking): instead of logging here, can we assert something about the header?

universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
```diff
@@ -30,8 +39,24 @@ protected Application provideApplication() {
...
     Config config = app.injector().instanceOf(Config.class);
-    config.setClients(new Clients(AnonymousClient.INSTANCE));
+    AnonymousClient client = AnonymousClient.INSTANCE;
+    config.setClients(new Clients(client));
+
+    // The SecurityLogic wants to use some smarts to figure out which client to use, but
+    // those smarts are never going to pick this client (since none of the endpoints are
+    // configured to use it), so we implement an anonymous client finder which always returns
+    // our client.
+    DefaultSecurityLogic securityLogic = new DefaultSecurityLogic();
```
carolinedanzi, line 51: question: Is this something we will need to do for every controller test? If so, should we (in a future PR) refactor this out so all the controller tests can inherit the same logic?

carolinedanzi (approved):
This looks nice, thank you for figuring out the unit tests!

universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
ndmckinley: Definitely, that was my intent for this class in the medium term.  For now I figure we'll just put all those tests in here.

universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
ndmckinley: I actually left that in there by accident, removed!

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
bion: Issue (blocking): `<input>` names and element IDs have different behavior and we should not encourage them defaulting them to the same value. It's common for multiple elements on the page or even within the same `<form>` (as is the case for checkboxes) to have the same name, but no two elements on the page should have the same ID.

Suggestion: have a second version of this method that allows specifying the ID.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -21,17 +21,18 @@ protected Content htmlContent(DomContent... domContents) {
...
         .attr("for", fieldName);
   }
 
   protected Tag passwordField(String fieldName, String labelText) {
-    return label(text(labelText), input().withType("password").withName(fieldName))
+    return label(
+            text(labelText), input().withType("password").withName(fieldName).withId(fieldName))
```
bion, line 30: (same as above)

Issue (blocking): <input> names and element IDs have different behavior and we should not encourage them defaulting them to the same value. It's common for multiple elements on the page or even within the same <form> (as is the case for checkboxes) to have the same name, but no two elements on the page should have the same ID.

Suggestion: have a second version of this method that allows specifying the ID.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -21,17 +21,18 @@ protected Content htmlContent(DomContent... domContents) {
...
         .attr("for", fieldName);
   }
 
   protected Tag submitButton(String textContents) {
-    return input().withType("submit").withValue(textContents);
+    return input().withType("submit").withId("submit").withValue(textContents);
```
bion, line 35: (same issue as above)

Issue (blocking): names and element IDs have different behavior and we should not encourage them defaulting them to the same value. It's common for multiple elements on the page or even within the same

(as is the case for checkboxes) to have the same name, but no two elements on the page should have the same ID.
Suggestion: have a second version of this method that allows specifying the ID.

universal-application-tool-0.0.1/test/controllers/HomeControllerAuthenticatedTest.java
```diff
@@ -30,8 +36,24 @@ protected Application provideApplication() {
 
   @Before
   public void setUp() {
+    // Get the config, and hack it so that all requests appear authorized.
     Config config = app.injector().instanceOf(Config.class);
-    config.setClients(new Clients(AnonymousClient.INSTANCE));
+    AnonymousClient client = AnonymousClient.INSTANCE;
```
bion, line 41: question: what's the relationship between a client an account? does an account always have one and only one client?

ndmckinley: Think "client" like "IDCSClient" - it's an interface to whatever your authentication / authorization system is.  So it's more like clients retrieve or create "profiles", which is pac4j's name for what you're calling an account.  Multiple clients could interact with the same profile - that's probably what we'll do for secure links or other no-creds login stuff.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
ndmckinley: Got it - agree, that seemed like a temporary hack when I wrote it, better not to let it go in at all.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
ndmckinley: Done as above (textfield also)

ndmckinley:
boy that formatter is real particular!  Let me clean up that history a little by removing those unrelated changes.

Issue #106, opened 2021-02-10
Admin can view the list of programs


bion:
blocked on #98, #97 

**Done when**
Admin is able to see a list of program definitions

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354355841621&cot=14

lianne-oy:
M1 Review - Done!

PR #167, opened 2021-02-17
Views and controllers for listing and creating new programs

kendrickt:
### Description
Add `AdminProgramController` with methods for listing and creating new programs, with views for `ProgramList` and `ProgramNewOne`. 

Creating new programs does not mean finalizing/publishing them, or adding blocks or questions. 

### Checklist

### Issue(s)
[Fixes] #106


universal-application-tool-0.0.1/conf/routes
```diff
@@ -15,7 +15,9 @@ POST    /postgres                   controllers.PostgresController.create(reques
 POST    /postgresSync               controllers.PostgresController.createSync(request: Request)
 
 # A controller for a page for an admin to create a new application
-GET     /admin/create               controllers.admin.CreateAppController.index
+GET     /admin/programs             controllers.admin.AdminProgramController.list
```
natsid, line 18: issue (blocking): Our [Html routing conventions](https://github.com/seattle-uat/universal-application-tool#html-routing-convention) uses `index` for the controller method name for listing a resource. We should either use that here, or if there's a good reason to use `list` instead, update the README.

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java
```diff
@@ -0,0 +1,104 @@
...
+    assertThat(contentAsString(result)).contains("This is a new program");
+  }
+
+  private static void insertProgram(String name) {
+    Program program = new Program(name, "description");
+    program.save();
+  }
+
+  /**
+   * For our HTTP responses, we set the charset to "UTF-8; charset=utf-8", which is not
```
natsid, line 97: question: Is this something we can / need to fix in the [`contentType` method](https://github.com/seattle-uat/universal-application-tool/blob/main/universal-application-tool-0.0.1/app/views/BaseHtmlView.java#L62)? I know @kendrickt mentioned this at standup, but I didn't fully follow the issue and what we might do to resolve it.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -0,0 +1,51 @@
...
+  public Result newOne(Request request) {
+    return ok(newOneView.render(request));
+  }
+
+  public Result create(Request request) {
+    DynamicForm requestData = this.formFactory.form().bindFromRequest(request);
+    String name = requestData.get("name");
+    String description = requestData.get("description");
+    service.createProgramDefinition(name, description);
+    return list();
```
natsid, line 49: question (blocking): I'm wondering if instead of returning `list()` we may want to return a different `Result`, such as a redirect like:

```java
return redirect("/admin/programs");
```

[Play docs](https://www.playframework.com/documentation/2.8.x/JavaActions#Redirects-are-simple-results-too) on redirect Results.

I also just found all of these [`created` methods](https://github.com/playframework/playframework/blob/master/core/play/src/main/java/play/mvc/Results.java#L776) on `Results`. The 201 "created" response type seems like the right thing semantically, but it doesn't looks like you can pass it a URL string to redirect to, so it might not be right in this case where we just want to send the user back to the list page.

Not sure if it matters, but thought I would flag this in case there is some "proper" response type for create/POSTs.

natsid (commented):
Going to take a closer look at the unit tests now, but sending my first batch of comments. Looks great!!

natsid (commented):
Update: tests look good 👍

universal-application-tool-0.0.1/app/views/admin/ProgramList.java
```diff
@@ -0,0 +1,27 @@
...
+import static j2html.TagCreator.h2;
+import static j2html.TagCreator.p;
+import static j2html.attributes.Attr.FORMACTION;
+
+import com.google.common.collect.ImmutableList;
+import play.twirl.api.Content;
+import services.program.ProgramDefinition;
+import views.BaseHtmlView;
+
+public final class ProgramList extends BaseHtmlView {
```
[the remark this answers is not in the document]
carolinedanzi: I like your idea of a `View` suffix - makes it clearer at first glance in the controller what their purpose is

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java
carolinedanzi: Yes, thank you! That was the issue - the `contentType` method is returning both content type and charset, which means the charset is in there twice. Removing it allows us to use Play's `Helper#contentAsString` method successfully

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
bion: Definitely redirect rather than directly calling the list method. Recommend using a 302 redirect: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -0,0 +1,51 @@
...
+  @Inject
+  public AdminProgramController(
+      ProgramService service,
+      ProgramList listView,
+      ProgramNewOne newOneView,
+      FormFactory formFactory) {
+    this.service = service;
+    this.listView = listView;
+    this.newOneView = newOneView;
+    this.formFactory = formFactory;
```
bion, line 33: nit: please use `checkNotNull` on injected dependencies

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public Result newOne(Request request) {
+    return ok(newOneView.render(request));
+  }
+
+  public Result create(Request request) {
+    DynamicForm requestData = this.formFactory.form().bindFromRequest(request);
+    String name = requestData.get("name");
+    String description = requestData.get("description");
```
bion, line 47: Suggestion (blocking): recommend defining and using a [form class](https://www.playframework.com/documentation/2.8.x/JavaForms#Handling-form-submission) rather than `DynamicForm` when we know the structure of the form data this controller is meant to handle ahead of time.

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java
natsid: Glad to hear that resolves it!! 🎉

universal-application-tool-0.0.1/app/views/admin/ProgramList.java
```diff
@@ -0,0 +1,27 @@
...
+
+  public Content render(ImmutableList<ProgramDefinition> programs) {
+    return htmlContent(
+        body(
+            h1("Programs"),
+            div(form(button("Add a Program").attr(FORMACTION, "/admin/programs/new"))),
```
bion, line 24: Suggestion (blocking): use the reverse router to get the URL path

```suggestion
            div(form(button("Add a Program").attr(FORMACTION, controllers.routes.AdminProgramController.newOne()))),
```

universal-application-tool-0.0.1/app/views/admin/ProgramNewOne.java
```diff
@@ -0,0 +1,24 @@
...
+        body(
+            h1("Create a new Program"),
+            div(
+                form(
+                        makeCsrfTokenInputTag(request),
+                        textField("name", "Program Name"),
+                        textField("description", "Program Description"),
+                        submitButton("Create"))
+                    .withMethod("post")
+                    .withAction("/admin/programs"))));
```
bion, line 22: Suggestion (blocking): use the reverse router to get the URL path

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -0,0 +1,51 @@
...
+import play.data.FormFactory;
+import play.mvc.Controller;
+import play.mvc.Http.Request;
+import play.mvc.Result;
+import services.program.ProgramService;
+import views.admin.ProgramList;
+import views.admin.ProgramNewOne;
+
+/**
+ * This controller contains an action to handle HTTP requests to the application's Admin "Create
```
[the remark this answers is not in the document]
kendrickt: done!

universal-application-tool-0.0.1/conf/routes
kendrickt: done!

universal-application-tool-0.0.1/app/views/admin/ProgramNewOne.java
kendrickt: done!

universal-application-tool-0.0.1/app/views/admin/ProgramList.java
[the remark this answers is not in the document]
kendrickt: done!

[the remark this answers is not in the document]
kendrickt: done

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
kendrickt: using `Results.found` for 302.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
kendrickt: done

universal-application-tool-0.0.1/app/views/admin/ProgramList.java
kendrickt: done!

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
kendrickt: done!

universal-application-tool-0.0.1/app/views/admin/ProgramNewOne.java
kendrickt: there isn't actually a `withAction` that takes a `Call`, so I think I have to use `controllers.routes.AdminProgramController.newOne().url()`

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
kendrickt: The tests had to change for the redirect. They now test that the response is 302 FOUND, and then makes another request and checks the list of programs to see if the new programs show up.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public Result create(Request request) {
+    Form<NewProgram> newProgramForm = formFactory.form(NewProgram.class);
+    NewProgram newProgram = newProgramForm.bindFromRequest(request).get();
+    service.createProgramDefinition(newProgram.getName(), newProgram.getDescription());
+    return found(controllers.admin.routes.AdminProgramController.index());
+  }
+
+  public static class NewProgram {
```
bion, line 53: Suggestion (blocking): this will be the same class for the edit/update form, recommend a more generic name e.g. `ProgramForm`.

Suggestion (non-blocking): form classes will likely get more complex and can have additional concerns such as validation. It'd probably be good to have a separate package for them e.g. `forms.ProgramForm` rather than expand the size of the controller.

bion (approved):

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java
```diff
@@ -0,0 +1,100 @@
+package controllers.admin;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static play.api.test.CSRFTokenHelper.addCSRFToken;
+import static play.mvc.Http.Status.OK;
+import static play.test.Helpers.*;
```
bion, line 6: suggestion (blocking): [remove wildcard import](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)

kendrickt: done!

kendrickt: either intellij or bin/fmt is using wildcards when the list gets real long. i'll see if its just my intellij settings.

natsid (approved):

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java
bion: Screen Shot 2021-02-19 at 9 46 22 AM

natsid: good tip!

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
kendrickt: done

universal-application-tool-0.0.1/app/views/admin/ProgramListView.java
```diff
@@ -0,0 +1,31 @@
...
+import views.BaseHtmlView;
+
+public final class ProgramListView extends BaseHtmlView {
+
+  public Content render(ImmutableList<ProgramDefinition> programs) {
+    return htmlContent(
+        body(
+            h1("Programs"),
+            div(
+                form(
```
AWBaumann, line 25: This doesn't need to be a form. 

Other options:
1. Just use a simple `<a>` and we can style it to look like a button later
2. `<button onclick="window.location.href='...'">Add a Program</button>`

universal-application-tool-0.0.1/app/views/admin/ProgramListView.java
```diff
@@ -0,0 +1,31 @@
...
+                    button("Add a Program")
+                        .attr(
+                            FORMACTION, controllers.admin.routes.AdminProgramController.newOne()))),
+            div(each(programs, program -> div(h2(program.name()), p(program.description()))))));
+  }
+}
```
AWBaumann, line 31: I think we need to split these out into sub-components as frequently as possible. When we start adding styles it should make things a lot simpler if the methods are small/distinct.

```
/** This should be in utils. */
Tag headerTag(String title) {
    return h1(title);
}

Tag renderLinkButton(String buttonText, Uri href) {
    return button(text).attr(onclick, String.format("window.location.href='%1s'", href))
}

Tag renderedProgramInfo(ProgramDefinition pd) {
    return div(
        h2(program.name()),
        p(program.description())
    );
}
```

PR #169, opened 2021-02-17
Setup correct java version in GitHub formatter check

hmhsu:
### Description
Make sure the Java formatter in GitHub uses the same version of JDK as our app (currently 11.0.10), so the formatter version is correctly selected (v1.9).

### Checklist


carolinedanzi (approved):
This looks promising - thanks for looking into this!

PR #185, opened 2021-02-19
Question model and tests

carolinedanzi:
### Description
Updated the Question model to match the question definition.

Paired with: @AWBaumann 

### Checklist

### Approvals
Want approval from any one reviewer


universal-application-tool-0.0.1/app/models/Question.java
```diff
@@ -1,34 +1,77 @@
...
+  private @Constraints.Required String name;
+
+  private @Constraints.Required String description;
+
+  private @Constraints.Required @DbJsonB ImmutableMap<Locale, String> questionText;
+
+  private ImmutableMap<Locale, String> questionHelpText;
+
+  public Question(QuestionDefinition questionDefinition) {
+    this.questionDefinition = questionDefinition;
```
bion, line 37: nit: 
```suggestion
    this.questionDefinition = checkNotNull(questionDefinition);
```

universal-application-tool-0.0.1/conf/evolutions/default/3.sql
```diff
@@ -9,7 +9,12 @@ create table if not exists applicants (
 
 create table if not exists questions (
   id bigserial primary key,
-  object jsonb not null
+  version bigserial,
+  path varchar,
+  name varchar,
+  description varchar,
+  question_text jsonb,
+  question_help_text jsonb
```
bion, line 17: question (non-blocking): what's the plan for persisting the question's question type?

bion (approved):

universal-application-tool-0.0.1/app/models/Question.java
carolinedanzi: Done!

universal-application-tool-0.0.1/app/models/Question.java
```diff
@@ -1,34 +1,77 @@
...
+  /** Populates column values from {@link QuestionDefinition}. */
+  @PreUpdate
+  public void persistChangesToQuestionDefinition() {
+    this.id = questionDefinition.getId();
+    this.version = questionDefinition.getVersion();
+    this.path = questionDefinition.getPath();
+    this.name = questionDefinition.getName();
+    this.description = questionDefinition.getDescription();
+    this.questionText = questionDefinition.getQuestionText();
+    this.questionHelpText = questionDefinition.getQuestionHelpText().orElse(null);
```
kendrickt, line 55: If this was `orElse(ImmutableMap.of())`, would `loadQuestionDefinition` be able to use `Optional.of(this.questionHelpText)` and we could avoid nulls altogether?

universal-application-tool-0.0.1/test/models/QuestionTest.java
```diff
@@ -0,0 +1,61 @@
...
+    QuestionRepository repo = app.injector().instanceOf(QuestionRepository.class);
+    QuestionDefinition definition =
+        new QuestionDefinition(1L, 1L, "test", "my.path", "", ImmutableMap.of(), Optional.empty());
+    Question question = new Question(definition);
+
+    question.save();
+
+    Question found = repo.lookupQuestion(definition.getId()).toCompletableFuture().join().get();
+
+    assertThat(found.getQuestionDefinition().getId()).isEqualTo(definition.getId());
```
kendrickt, line 26: nit(non-blocking): I like to see the `.isEqualTo` values as raw values when possible.

kendrickt: ditto for the `canSerializeLocalizationMaps` test case.

kendrickt:
Is it weird that the `Program` model stores serialized `QuestionDefinition`s and not `Question`s?

universal-application-tool-0.0.1/app/models/Question.java
carolinedanzi: Agree that anytime we have null values is cause for concern, but I think we want nulls here - if a question doesn't have help text, we want `Optional.empty()`, not an `Optional` with an empty map

universal-application-tool-0.0.1/test/models/QuestionTest.java
carolinedanzi: Sounds good to me! Will update tests

universal-application-tool-0.0.1/conf/evolutions/default/3.sql
carolinedanzi: We're working on it - we need to persist the QuestionType enum safely, then reform the question into the proper class on load. We're working on this today (Friday)

Issue #170, opened 2021-02-18
Create user account and Applicant for unauth'd visitors


natsid:
A user account should automatically and silently be created when an unauthorized visitor visits a page that requires auth. In the context of the "applicant" user role, an associated `Applicant` (with foreign key to the new user/profile) should also be created.

**Done when**
There exists a way to silently create a user account and associated applicant, likely by using the `@Secure` annotation.

lianne-oy:
M1 Review - In progress

PR #186, opened 2021-02-19
Allow credentialless login.

ndmckinley:
### Description
This allows credentialless logins through a new pac4j client which creates a new account and stores it in the session token.  It persists until the user logs out or deletes their cookies.

### Checklist
- [X] Extended the README / documentation, if necessary

### Issue(s)
Related to, but does not complete, #170.


universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -12,13 +12,18 @@
...
                     makeCsrfTokenInputTag(request),
                     textField("uname", "username", "Username"),
                     passwordField("pwd", "password", "Password"),
                     submitButton("login", "Submit"))
                 .withMethod("POST")
-                .withAction("/callback?client_name=FormClient")));
+                .withAction("/callback?client_name=FormClient"),
+            h1("Or, continue as guest."),
+            button("guest", "continue")
+                .attr("onclick", "window.location = '/callback?client_name=GuestClient';")));
```
bion, line 27: suggestion (non-blocking): this can be done without using JS

```java
import static j2html.attributes.Attr.FORMACTION;
...
form(
    button("guest", "continue")
    .attr(
        FORMACTION, route))
```

ndmckinley: Tried it, but it doesn't work with URL parameters.  We can get around that by making it a `POST` form, or by providing the url parameter as a hidden input, both of which seem like slightly too much workaround to avoid a JS fragment.  Plus, if we make it a form, I think we have to interface with Play's XSRF tokens, which is easy, but also seems like a little too much workaround.

This is some super standard javascript, the entire front page of a google search for "html button redirect" is all a variation on this, so I'm sure it's just as widely supported in browsers as html forms are, and I think it's marginally more readable than the formaction + hidden input + xsrf token approach.

Guessing by non-blocking you don't feel _too_ strongly and the added complexity there will be persuasive, but post again if you disagree - willing to go the formaction route if you're thinking we should go a bit further to avoid JS.

bion: Yep, didn't feel strongly about it. Thanks for the info though it's helpful to know!

universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -70,17 +98,22 @@ protected FormClient provideFormClient() {
...
     Clients clients = new Clients(baseUrl + "/callback");
-    clients.setClients(directBasicAuthClient, formClient);
+    clients.setClients(guestClient, formClient);
     PlayHttpActionAdapter.INSTANCE
         .getResults()
         .putAll(
             ImmutableMap.of(
                 HttpConstants.UNAUTHORIZED,
-                unauthorized("401 not authorized").as(HttpConstants.HTML_CONTENT_TYPE),
+                redirect("/loginForm?message=login"),
```
bion, line 116: Suggestion (blocking): use the reverse router when feasible: `controllers.routes.HomeController.loginForm().url() + "?message=login"`

universal-application-tool-0.0.1/app/views/LoginForm.java
natsid: 👍  Good to know.

bion: Though as Aaron points out in another PR, factoring this out would be helpful: https://github.com/seattle-uat/universal-application-tool/pull/167/files#r579390736

universal-application-tool-0.0.1/conf/evolutions/default/5.sql
```diff
@@ -0,0 +1,16 @@
+# --- Accounts
+
+# --- !Ups
+
+create table accounts (
+  id bigserial primary key
+  -- table does not contain a username or password hash yet.
+);
+
+alter table applicants add account bigint;
```
bion, line 10: suggestion (blocking): renaming `account` to `account_id` would make working with this column easier to read

universal-application-tool-0.0.1/conf/logback.xml
```diff
@@ -32,7 +32,7 @@
   <!-- Off these ones as they are annoying, and anyway we manage configuration ourselves -->
   <logger name="com.gargoylesoftware.htmlunit.javascript" level="OFF" />
 
-  <root level="WARN">
+  <root level="DEBUG">
```
bion, line 35: question: is this intentionally checked in?

universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -12,13 +12,18 @@
...
+                .withCondHidden(!request.queryString("message").orElse("").equals("login")),
             h1("Log In"),
             form(
                     makeCsrfTokenInputTag(request),
                     textField("uname", "username", "Username"),
                     passwordField("pwd", "password", "Password"),
                     submitButton("login", "Submit"))
                 .withMethod("POST")
-                .withAction("/callback?client_name=FormClient")));
+                .withAction("/callback?client_name=FormClient"),
```
bion, line 24: suggestion (blocking): please use the reverse router `org.pac4j.play.routes.CallbackController.callback().url()`

universal-application-tool-0.0.1/app/controllers/ProfileController.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public Result myProfile(Http.Request request) {
+    // Fetch the current profile from the session cookie, which the ProfileManager
+    // will fetch from the request's cookies, using the session store to decrypt it.
+    PlayWebContext webContext = new PlayWebContext(request);
+    ProfileManager profileManager = new ProfileManager(webContext, sessionStore);
+    return ok(profileView.render(profileManager.getProfile()));
+  }
+
+  public Result profilePage(Http.Request request, Long id) {
```
natsid, line 30: question (non-blocking): Is the idea behind this one for trusted intermediaries to be able to view different profiles?

universal-application-tool-0.0.1/app/views/LoginForm.java
natsid, line 24: suggestion (non-blocking): I found [these docs](https://www.playframework.com/documentation/2.8.x/JavaRouting#Call-to-action-generator-method) on query strings. It could be convenient to use a `client_name` param in the `callback` controller method. There's also a way to specify a default value.

Likewise for potentially adding a `message` param on the `loginForm` controller method. (Unless you need the entire `request` in these cases.)

universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -12,13 +12,18 @@
...
+            h1("Or, continue as guest."),
+            button("guest", "continue")
```
natsid, line 26: nit: Capitalize "Continue" to match "Submit" above.

natsid (approved):
Other than updating the URLs to use reverse routing, LGTM!

bion (approved):

Issue #113, opened 2021-02-10
Applicant can view a program form (blocks rendered as forms)


bion:
**Describe the solution you'd like**
Renders the block's questions as form inputs. Has a deep-link URL to the program block (must be able to copy the URL and open a new tab that goes to the same block in the same program).

**Done when**
Applicant can view a program form

gcapiel:
wireframes: https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354365631793&cot=14

lianne-oy:
M1 Review - In progress

PR #202, opened 2021-02-21
Applicant can view program form

bion:
### Description
Renders a block for an applicant

Some thoughts:

* Question types should have named methods for whatever scalars they've got, otherwise we'll end up hard-coding string values to cobble together the full path (see nested classes in `services.applicant.ApplicantQuestion`)

### Issue(s)
[Fixes] #113


natsid (approved):
LGTM after reviewing with you over VC!

PR #218, opened 2021-02-23
Minor cleanup to use instanceof in tests

carolinedanzi:
### Description
Use `instanceOf(class)` instead of `app.injector().instanceOf()` in tests that extend WithPostgresContainer or WithApplication


bion (approved):
This is a great improvement, thanks!

Issue #103, opened 2021-02-10
Admin can update an existing question


bion:
Blocked by #101 

**Done when**
Admin can view the form for an existing question definition and submit it to update the question definition

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354353430975&cot=14

lianne-oy:
M1 Review - In progress

Issue #102, opened 2021-02-10
Admin can view list of questions


bion:
**Done when**
Admin can view list of questions i.e. the question bank

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354353430975&cot=14

lianne-oy:
M1 Review - In progress

Issue #101, opened 2021-02-10
Admin can submit question form


bion:
Blocked by #95

**Done when**
Admin submit the form from #95 and persist a new question definition.

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354352546632&cot=14

lianne-oy:
M1 Review - In progress

Issue #95, opened 2021-02-10
Admin can view question form


bion:
This form allows an admin to create a new question (i.e. a `QuestionDefinition`) using the `NAME` question type.

- Simple text input field for question path for now
- question and help text are only for english locale
- question type selected from dropdown
- version hard-coded to "1" for now
- no tags

Done when admin can view the question form.

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354352415135&cot=14

lianne-oy:
M1 Review:  In progress should be done by EOD

PR #219, opened 2021-02-23
Implement question controller with routes for viewing, creating, and editing QuestionDefinitions.

AWBaumann:
### Description
Created QuestionController, which is responsible for routing to the appropriate views and calling the necessary services.
Updated ReadOnlyQuestionService to include a nextId field. (This should be fixed in a follow-up PR)

### Checklist

### Issue(s)
[Fixes] #95 
[Fixes] #101 
[Fixes] #102 

### Follow ups (will attach issues to these)
**Implement `QuestionService.update`** #103
Right now it just throws an exception.

**Unwrap `questionHelpText` because an `Optional<ImmutableList>` doesn't make sense** #237
Why use `Optional.isPresent()` when we can just use `ImmutableList.isEmpty()`?

**Add helper methods for getting the question text and question help text** #237 
This was my thought process behind `ReadOnlyQuestionService.setPreferredLocale`. That would result in us not having to pass around the locale all of the time.

### Review
Will submit after I get at least 2 approvals.


universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Inject
+  public QuestionEditView(BaseHtmlLayout layout) {
+    this.layout = layout;
+  }
+
+  public Content render(Request request, Optional<QuestionDefinition> question) {
+    String headerText = question.isPresent() ? "Edit Question" : "New Question";
+    return layout.htmlContent(
+        body(
```
hmhsu, line 34: question: should this be `body().with()` to be consistent with elsewhere? I am just curious if they are the same or there is some difference there.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,100 @@
...
+            renderHeader(headerText),
+            buildQuestionForm(question.orElse(null)).with(makeCsrfTokenInputTag(request))));
```
hmhsu, line 36: Suggestion (non-blocking): I think it would be more readable to pass down optional and check `isPresent` below instead of `null`.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,100 @@
...
+          label("id: " + definition.getId()),
+          br(),
+          label("version: " + definition.getVersion()),
+          br(),
+          br());
+      try {
+        questionText = Optional.of(definition.getQuestionText(Locale.ENGLISH));
+        questionHelpText = Optional.of(definition.getQuestionHelpText(Locale.ENGLISH));
+      } catch (TranslationNotFoundException e) {
+        questionText = Optional.of("Error");
```
hmhsu, line 60: Suggestion (non-blocking): consider using `missing text` rather than `error`?

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
```diff
@@ -0,0 +1,97 @@
...
+          QuestionRow.QUESTION_HELP_TEXT,
+          QuestionRow.QUESTION_TYPE,
+          QuestionRow.ACTIONS);
+
+  @Inject
+  public QuestionsListView(BaseHtmlLayout layout) {
+    this.layout = layout;
+  }
+
+  /** Renders a page with either a table or a list view of all questions. */
```
carolinedanzi, line 40: question: this is really cool, but why the two views? The wireframes for M1/M2 just have a list of the question text (in English) for the question bank, but I may be missing a lot of context

universal-application-tool-0.0.1/conf/routes
```diff
@@ -21,6 +21,13 @@ GET     /admin/programs/:id/edit    controllers.admin.AdminProgramController.edi
 POST    /admin/programs/:id         controllers.admin.AdminProgramController.update(request: Request, id: Long)
 POST    /admin/programs             controllers.admin.AdminProgramController.create(request: Request)
 
+# A controller for a page for an admin to view, edit, and create questions
+GET     /admin/questions             controllers.admin.QuestionController.list(renderAs: String ?= "table")
```
kendrickt, line 25: controller method should be `index`.

https://github.com/seattle-uat/universal-application-tool#html-routing-convention

universal-application-tool-0.0.1/conf/routes
```diff
@@ -21,6 +21,13 @@ GET     /admin/programs/:id/edit    controllers.admin.AdminProgramController.edi
...
+GET     /admin/questions/new         controllers.admin.QuestionController.create(request: Request)
```
kendrickt, line 26: controller method should be `newOne`

universal-application-tool-0.0.1/conf/routes
```diff
@@ -21,6 +21,13 @@ GET     /admin/programs/:id/edit    controllers.admin.AdminProgramController.edi
...
+GET     /admin/questions/edit/:path  controllers.admin.QuestionController.edit(request: Request, path: String)
```
kendrickt, line 27: Use `/admin/questions/:path/edit`

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+              try {
+                definition = Optional.of(readOnlyService.getQuestionDefinition(path));
+              } catch (InvalidPathException e) { // If the path doesn't exist, redirect to create.
+                System.out.println(e); // TODO: What are we using for logging?
+              }
+              if (definition.isPresent()) {
+                return ok(editView.render(request, definition));
+              } else {
+                return redirect("/admin/questions/new");
+              }
```
carolinedanzi, line 68: suggestion (non-blocking): This might be easier to read if the redirects are included in the try/catch (i.e. redirect for invalid path after logging)

universal-application-tool-0.0.1/conf/routes
```diff
@@ -21,6 +21,13 @@ GET     /admin/programs/:id/edit    controllers.admin.AdminProgramController.edi
...
+POST    /admin/questions/update      controllers.admin.QuestionController.update(request: Request)
```
kendrickt, line 28: These two methods don't need the `update` because of the `POST`. So just `/admin/questions/:id` for the edit, and `/admin/questions` for the new one.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      QuestionDefinition definition = questionForm.getBuilder().setId(0L).setVersion(1L).build();
+      service.update(definition);
+    } catch (UnsupportedQuestionTypeException e) {
+      // I'm not sure why this would happen here, so we'll just log and redirect.
+      System.out.println(e);
+    } catch (UnsupportedOperationException e) {
+      // This is expected for now until we implement update on QuestionService.
+    }
+    return list("table");
```
carolinedanzi, line 84: suggestion (blocking): Instead of calling the method, we'll want to use reverse routing to redirect: `found(routes.QuestionController.list("table"))`

universal-application-tool-0.0.1/test/controllers/admin/QuestionControllerTest.java
```diff
@@ -0,0 +1,146 @@
+package controllers.admin;
```
hmhsu, line 1: praise: nice!

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+                return redirect("/admin/questions/new");
+              }
+            });
+  }
+
+  public CompletionStage<Result> update(Request request) {
+    Form<QuestionForm> form = formFactory.form(QuestionForm.class);
+    QuestionForm questionForm = form.bindFromRequest(request).get();
+    try {
+      QuestionDefinition definition = questionForm.getBuilder().setId(0L).setVersion(1L).build();
```
carolinedanzi, line 76: question: why are ID and version set to these values?

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+      service.update(definition);
+    } catch (UnsupportedQuestionTypeException e) {
+      // I'm not sure why this would happen here, so we'll just log and redirect.
+      System.out.println(e);
```
carolinedanzi, line 80: suggestion (blocking): log (see Nathan's comment)

universal-application-tool-0.0.1/app/forms/QuestionForm.java
```diff
@@ -0,0 +1,84 @@
...
+  public void setQuestionHelpText(String questionHelpText) {
+    this.questionHelpText = questionHelpText;
+  }
+
+  public QuestionDefinitionBuilder getBuilder() {
+    ImmutableMap<Locale, String> questionTextMap =
+        (questionText.length() > 0)
+            ? ImmutableMap.of(Locale.ENGLISH, questionText)
+            : ImmutableMap.of();
+    Optional<ImmutableMap<Locale, String>> questionHelpTextMap =
```
kendrickt, line 70: Not quite in scope for this PR, but what's the point of Optional<ImmutableMap> when the map can be empty?

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+    } catch (UnsupportedOperationException e) {
+      // This is expected for now until we implement update on QuestionService.
```
carolinedanzi, line 82: Thanks for the note! Mind adding a TODO?

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
```diff
@@ -0,0 +1,50 @@
...
+            Optional.empty());
+    QuestionDefinition actual = builder.build();
+
+    assertThat(actual.getId()).isEqualTo(expected.getId());
+    assertThat(actual.getVersion()).isEqualTo(expected.getVersion());
+    assertThat(actual.getName()).isEqualTo(expected.getName());
+    assertThat(actual.getQuestionType()).isEqualTo(expected.getQuestionType());
+    assertThat(actual.getPath()).isEqualTo(expected.getPath());
+    assertThat(actual.getDescription()).isEqualTo(expected.getDescription());
+  }
```
hmhsu, line 49: suggestion (non-blocking): add checks for questionType and help text.

hmhsu (approved):
Looks great!

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+      QuestionsListView listView,
+      QuestionEditView editView,
+      FormFactory formFactory) {
+    this.service = checkNotNull(service);
+    this.listView = checkNotNull(listView);
+    this.editView = checkNotNull(editView);
+    this.formFactory = checkNotNull(formFactory);
+  }
+
+  public CompletionStage<Result> list(String renderAs) {
```
carolinedanzi, line 40: suggestion (blocking): This should be `index`: https://github.com/seattle-uat/universal-application-tool#routing-and-controller-methods

universal-application-tool-0.0.1/app/forms/QuestionForm.java
AWBaumann: Yeah, I think I've discussed that with a few others offline. `ImmutableMap.isEmpty()` is equivalent.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+    return service
+        .getReadOnlyQuestionService()
+        .thenApplyAsync(
+            readOnlyService -> {
+              return ok(listView.render(readOnlyService.getAllQuestions(), renderAs));
+            });
+  }
+
+  public Result create(Request request) {
```
carolinedanzi, line 49: suggestion (blocking): This should be `edit`: https://github.com/seattle-uat/universal-application-tool#routing-and-controller-methods

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+    } catch (UnsupportedQuestionTypeException e) {
+      // I'm not sure why this would happen here, so we'll just log and redirect.
+      System.out.println(e);
+    } catch (UnsupportedOperationException e) {
+      // This is expected for now until we implement update on QuestionService.
+    }
+    return list("table");
+  }
+
+  public CompletionStage<Result> write(Request request) {
```
carolinedanzi, line 87: suggestion (blocking): this should be `create`: https://github.com/seattle-uat/universal-application-tool#routing-and-controller-methods

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+              Optional<QuestionDefinition> definition = Optional.empty();
+              try {
+                definition = Optional.of(readOnlyService.getQuestionDefinition(path));
+              } catch (InvalidPathException e) { // If the path doesn't exist, redirect to create.
+                System.out.println(e); // TODO: What are we using for logging?
+              }
+              if (definition.isPresent()) {
+                return ok(editView.render(request, definition));
+              } else {
+                return redirect("/admin/questions/new");
```
carolinedanzi, line 67: suggestion (blocking): You can use reverse routing here too: `routes.QuestionController.new()`

universal-application-tool-0.0.1/app/forms/QuestionForm.java
```diff
@@ -0,0 +1,84 @@
...
+    return questionHelpText;
+  }
+
+  public void setQuestionHelpText(String questionHelpText) {
+    this.questionHelpText = questionHelpText;
+  }
+
+  public QuestionDefinitionBuilder getBuilder() {
+    ImmutableMap<Locale, String> questionTextMap =
+        (questionText.length() > 0)
```
carolinedanzi, line 67: suggestion (blocking): this can be `!questionText.isEmpty()`

universal-application-tool-0.0.1/app/forms/QuestionForm.java
```diff
@@ -0,0 +1,84 @@
...
+    Optional<ImmutableMap<Locale, String>> questionHelpTextMap =
+        (questionHelpText.length() > 0)
+            ? Optional.of(ImmutableMap.of(Locale.ENGLISH, questionHelpText))
+            : Optional.empty();
+    QuestionDefinitionBuilder builder =
+        new QuestionDefinitionBuilder()
+            .setQuestionType(QuestionType.valueOf(questionType))
+            .setName(questionName == null ? "" : questionName)
+            .setPath(questionPath == null ? "" : questionPath)
+            .setDescription(questionDescription == null ? "" : questionDescription)
```
carolinedanzi, line 79: question: will we be adding validation for these later?

universal-application-tool-0.0.1/app/services/question/ReadOnlyQuestionService.java
```diff
@@ -8,7 +8,13 @@
  * The ReadOnlyQuestionService contains all synchronous, in-memory operations for
  * QuestionDefinitions.
  */
-interface ReadOnlyQuestionService {
+public interface ReadOnlyQuestionService {
+
+  /**
+   * Returns a valid id for new QuestionDefinitions. TODO: Remove. The id should be set by the
```
carolinedanzi, line 14: nit: can we link an issue?

universal-application-tool-0.0.1/app/services/question/ReadOnlyQuestionServiceImpl.java
```diff
@@ -22,6 +23,12 @@ public ReadOnlyQuestionServiceImpl(ImmutableList<QuestionDefinition> questions)
     ImmutableMap.Builder<String, QuestionDefinition> scalarParentsMap = ImmutableMap.builder();
     for (QuestionDefinition qd : questions) {
       String questionPath = qd.getPath();
+      // Remove this code.
```
carolinedanzi, line 26: nit: add TODO with link to issue

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: They're _mostly_ the same. 

If the tag that you're using is a `ContainerTag` (which most are) then you can use `body().with(tag).with(tag2)` or `body(tag1, tag2)`

The base tag class doesn't implement `with` but we'll likely never be using that unless we run into an unsupported type.

I try to use `.with` because I think it's cleaner, but that's not always true (and can largely depend on what the auto-formatter does to your code.)

universal-application-tool-0.0.1/conf/routes
```diff
@@ -21,6 +21,13 @@ GET     /admin/programs/:id/edit    controllers.admin.AdminProgramController.edi
...
+POST    /admin/questions/write       controllers.admin.QuestionController.write(request: Request)
```
carolinedanzi, line 29: suggestion (blocking): make sure these align with https://github.com/seattle-uat/universal-application-tool#routing-and-controller-methods

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+
+public class QuestionController extends Controller {
+
+  private final QuestionService service;
+  private final QuestionsListView listView;
+  private final QuestionEditView editView;
+  private final FormFactory formFactory;
+
+  @Inject
+  public QuestionController(
```
carolinedanzi, line 29: suggestion (blocking): We'll want to use the HttpExecutionContext for async controllers (I learned this today!): https://www.playframework.com/documentation/2.8.x/JavaAsync#Using-HttpExecutionContext

carolinedanzi (commented):
Exciting to see this coming together! Leaving a first round of comments (mostly nits about naming conventions in the controller)

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
AWBaumann: It was mainly an exercise/example. When we're actually implementing UX we'll want to do branching calls like this to do A/B testing or implement different DOM structures or styles so it's just an example of how we can quickly prototype rendering the same object in different ways.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+    this.service = checkNotNull(service);
+    this.listView = checkNotNull(listView);
+    this.editView = checkNotNull(editView);
+    this.formFactory = checkNotNull(formFactory);
+  }
+
+  public CompletionStage<Result> list(String renderAs) {
+    return service
+        .getReadOnlyQuestionService()
+        .thenApplyAsync(
```
kendrickt, line 43: inject an `HttpExecutionContext` and explicitly use `httpExecutionContext.current()`.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,109 @@
...
+  public CompletionStage<Result> edit(Request request, String path) {
+    return service
+        .getReadOnlyQuestionService()
+        .thenApplyAsync(
+            readOnlyService -> {
+              Optional<QuestionDefinition> definition = Optional.empty();
+              try {
+                definition = Optional.of(readOnlyService.getQuestionDefinition(path));
+              } catch (InvalidPathException e) { // If the path doesn't exist, redirect to create.
+                System.out.println(e); // TODO: What are we using for logging?
```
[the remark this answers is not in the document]
AWBaumann: Done.

universal-application-tool-0.0.1/conf/routes
AWBaumann: Done.

universal-application-tool-0.0.1/conf/routes
AWBaumann: Done.

universal-application-tool-0.0.1/conf/routes
AWBaumann: Done.

universal-application-tool-0.0.1/conf/routes
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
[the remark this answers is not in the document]
AWBaumann: Used `org.slf4j.LoggerFactory` per Nathan's suggestion.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
kendrickt, line 67: (nit) is it more readable to put the `return redirect` in `catch`? I don't like the comment on line 61 and the `return redirect` because if we change what is happening here, we have to remember to update that comment.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done.

universal-application-tool-0.0.1/conf/routes
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Edit/Update isn't implemented yet, but for the questionForm to build the questionDefinition it needs and id and version. These aren't currently set in the form (we may want hidden inputs for them)  but I wanted to avoid setting the id in the `QuestionForm`

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
kendrickt, line 84: i think this should be a redirect

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
  */
 public abstract class BaseHtmlView {
 
+  public Tag renderHeader(String headerText) {
+    return h1(headerText);
+  }
+
+  protected ImmutableList<DomContent> inputWithLabel(
```
bion, line 33: Suggestion (blocking): rename to `textInputWithLabel` since this specifically for an input type of "text"

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
...
+      String labelValue, String inputId, Optional<String> value) {
+    Tag labelTag = label(labelValue).attr("for", inputId);
+    Tag inputTag = input().withType("text").withId(inputId).withName(inputId);
+    if (value.isPresent()) {
+      inputTag.withValue(value.get());
+    }
+    return ImmutableList.of(labelTag, br(), inputTag, br(), br());
```
bion, line 40: nit: please add whitespace before the return statement and to separate the multi-line `if`

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
...
+  }
+
+  public ImmutableList<DomContent> textAreaWithLabel(
+      String labelValue, String inputId, Optional<String> value) {
+    Tag labelTag = label(labelValue).attr("for", inputId);
+    Tag textAreaTag = textarea(value.orElse("")).withType("text").withId(inputId).withName(inputId);
+    return ImmutableList.of(labelTag, br(), textAreaTag, br(), br());
```
bion, line 47: nit: please add whitespace before the return statement

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
...
+  }
+
+  public ImmutableList<DomContent> formSelect(
+      String labelValue,
+      String selectId,
+      String[] optionLabels,
+      String[] optionValues,
```
bion, line 54: Suggestion (blocking): use `ImmutableList` rather than a raw Java array. We should generally avoid using raw arrays unless a library forces us to.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
...
+      String selectedValue) {
+    Tag labelTag = label(labelValue).attr("for", selectId);
+    ContainerTag selectTag = select().withId(selectId).withName(selectId);
+    for (int i = 0; i < optionLabels.length && i < optionValues.length; i++) {
```
bion, line 58: Suggestion (blocking): assert that the lists are of the same length rather than check both, this could hide a bug where one is longer than other and we get unexpected output

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +26,45 @@
...
+      Tag optionTag = option(optionLabels[i]).withValue(optionValues[i]);
+      if (optionValues[i].equals(selectedValue)) {
+        optionTag.attr("selected");
+      }
+      selectTag.with(optionTag);
+    }
+    return ImmutableList.of(labelTag, br(), selectTag, br(), br());
```
bion, line 65: nit: please add empty lines before return and around the mulit-line `for` and `if`

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
bion: +1 please do not intentionally pass `null` to methods or constructors if ever avoidable

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: DONE

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,103 @@
...
+    String buttonText = "";
+
+    ContainerTag formTag = form().withMethod("POST");
+
+    Optional<String> questionText = Optional.empty();
+    Optional<String> questionHelpText = Optional.empty();
+
+    if (definition != null) { // Editing a question.
+      buttonText = "Update";
+      formTag.withAction("/admin/questions/" + definition.getId());
```
bion, line 49: Suggestion (blocking): use the reverse router. We should never need to create URL paths with string concatenation.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,103 @@
...
+      try {
+        questionText = Optional.of(definition.getQuestionText(Locale.ENGLISH));
+      } catch (TranslationNotFoundException e) {
+        questionText = Optional.of("Missing Text");
+      }
+      try {
+        questionHelpText = Optional.of(definition.getQuestionHelpText(Locale.ENGLISH));
+      } catch (TranslationNotFoundException e) {
+        questionHelpText = Optional.of("Missing Text");
+      }
```
bion, line 65: Suggestion (blocking): I think we can extract a method for this to reduce duplication and also drop the `Optional`.

```java
protected String getLocaleString(Function<Locale, String> retrievalMethod, Locale locale) {
  try {
    return retrievalMethod.apply(locale);
  } catch (TranslationNotFoundException e) {
     logger.warn("Missing locale string for method: " + retrievalMethod.toString());
     return "Missing Text";
   }
   
... 
  questionHelpText = getLocaleString(definition::getQuestionHelpText, Locale.ENGLISH);
```

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+      buttonText = "Create";
+      formTag.withAction("/admin/questions");
```
bion, line 68: Suggestion (blocking): use reverse router to get path

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,103 @@
...
+        .with(submitButton(buttonText));
+    return formTag;
+  }
+
+  private ImmutableList<DomContent> formQuestionTypeSelect(QuestionType selectedType) {
+    QuestionType[] questionTypes = QuestionType.values();
+    String[] labels =
+        Arrays.stream(questionTypes).map(item -> item.toString()).toArray(String[]::new);
+    String[] values = Arrays.stream(questionTypes).map(item -> item.name()).toArray(String[]::new);
+    return this.formSelect("Question type: ", "questionType", labels, values, selectedType.name());
```
bion, line 101: nit: `this` can be omitted here

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionRow.java
```diff
@@ -0,0 +1,67 @@
...
+import static j2html.TagCreator.a;
+import static j2html.TagCreator.span;
+import static j2html.TagCreator.td;
+
+import j2html.tags.Tag;
+import java.util.Locale;
+import services.question.QuestionDefinition;
+import services.question.TranslationNotFoundException;
+
+enum QuestionRow {
```
bion, line 12: praise: I like how this encapsulates the logic for rendering a table cell. One nit on naming: should the name of the enum be `QuestionTableCell`? A specific instance of the enum is for a cell rather than a row...

universal-application-tool-0.0.1/app/views/admin/questions/QuestionRow.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+          return td(definition.getQuestionText(Locale.ENGLISH));
+        } catch (TranslationNotFoundException e) {
+          return td("");
+        }
+      case QUESTION_HELP_TEXT:
+        try {
+          return td(definition.getQuestionHelpText(Locale.ENGLISH));
+        } catch (TranslationNotFoundException e) {
+          return td("");
```
bion, line 55: Thought: this and line 49 look like they could also use the `getLocaleString` method suggested above.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
```diff
@@ -0,0 +1,97 @@
...
+    return layout.htmlContent(
+        body()
+            .with(renderHeader("All Questions"))
+            .with(renderAllQuestions(questions, renderAs))
+            .with(renderSummary(questions))
+            .with(renderAddQuestionLink()));
+  }
+
+  private Tag renderAddQuestionLink() {
+    return a("Create a new question").withHref("/admin/questions/new");
```
bion, line 51: Suggestion (blocking): use reverse router for path

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,113 @@
...
+                return redirect("/admin/questions/new");
+              }
+            });
+  }
+
+  public Result newOne(Request request) {
+    return ok(editView.render(request, Optional.empty()));
+  }
+
+  public CompletionStage<Result> index(String renderAs) {
```
bion, line 90: Suggestion (blocking): `renderAs` when it is passed in here to `index` is user input. We should validate that it is one of an allowed set of values and then convert to an enum to pass to the view for rendering on line 95. Alternatively,  we could `switch (renderAs)` and call a different render method on `listView` for each valid value of `renderAs` and throw bad request by default.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,113 @@
...
+    try {
+      QuestionDefinition definition = questionForm.getBuilder().setId(0L).setVersion(1L).build();
+      service.update(definition);
+    } catch (UnsupportedQuestionTypeException e) {
+      // I'm not sure why this would happen here, so we'll just log and redirect.
+      LOG.info(e.toString());
+    } catch (UnsupportedOperationException e) {
+      // This is expected for now until we implement update on QuestionService.
+    }
+    return CompletableFuture.completedFuture(redirect("/admin/questions"));
```
bion, line 111: Suggestion (blocking): use reverse router for path

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,113 @@
...
+                        .getBuilder()
+                        .setId(readOnlyService.getNextId())
+                        .setVersion(1L)
+                        .build();
+                service.create(definition);
+              } catch (UnsupportedQuestionTypeException e) {
+                // I'm not sure why this would happen here, so we'll just log and redirect.
+                LOG.info(e.toString());
+              }
+              return redirect("/admin/questions");
```
bion, line 63: Suggestion (blocking): use reverse router for path

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionRow.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
bion: Recommend returning NOT FOUND if the question definition isn't found rather than redirecting to `new`

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Done

universal-application-tool-0.0.1/app/forms/QuestionForm.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Thanks! I opted for the switch in the controller.

universal-application-tool-0.0.1/app/forms/QuestionForm.java
AWBaumann: I'm not entirely sure where the validation will be, but short answer is "yes".

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -0,0 +1,121 @@
...
+    return service
+        .getReadOnlyQuestionService()
+        .thenApplyAsync(
+            readOnlyService -> {
+              Optional<QuestionDefinition> definition = Optional.empty();
+              try {
+                definition = Optional.of(readOnlyService.getQuestionDefinition(path));
+              } catch (InvalidPathException e) { // If the path doesn't exist, redirect to create.
+                LOG.info(e.toString());
+                return redirect(routes.QuestionController.create());
```
hmhsu, line 77: Question: did you mean `newOne`?

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,117 @@
...
+        questionText = Optional.of("Missing Text");
+      }
+      try {
+        questionHelpText = Optional.of(definition.getQuestionHelpText(Locale.ENGLISH));
+      } catch (TranslationNotFoundException e) {
+        questionHelpText = Optional.of("Missing Text");
+      }
+    } else {
+      buttonText = "Create";
+      formTag.withAction(controllers.admin.routes.QuestionController.newOne().toString());
```
hmhsu, line 71: question: did you mean `create()`?

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
AWBaumann: Yup.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Yes, yes I did!

I was just debugging this

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,117 @@
...
+    ContainerTag formTag = form().withMethod("POST");
+
+    Optional<String> questionText = Optional.empty();
+    Optional<String> questionHelpText = Optional.empty();
+
+    if (optionalDefinition.isPresent()) { // Editing a question.
+      QuestionDefinition definition = optionalDefinition.get();
+      buttonText = "Update";
+      formTag.withAction(
+          controllers.admin.routes.QuestionController.edit(definition.getPath()).toString());
```
hmhsu, line 52: question: did you mean `update`? Also update takes an `id`, maybe should be `path`.

AWBaumann: Yup. It's getting late.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
```diff
@@ -0,0 +1,52 @@
...
+    form.setQuestionType("TEXT");
+    QuestionDefinitionBuilder builder = form.getBuilder();
+
+    // The QuestionForm does not set version or id, which are needed in order to build the
+    // QuestionDefinition.
+    builder.setVersion(1L);
+
+    TextQuestionDefinition expected =
+        new TextQuestionDefinition(
+            OptionalLong.empty(),
```
hmhsu, line 34: nit (blocking): we can omit specifying this now. Please use the constructor that doesn't take an `id` as argument.

universal-application-tool-0.0.1/test/controllers/admin/QuestionControllerTest.java
```diff
@@ -0,0 +1,159 @@
...
+            result -> {
+              assertThat(contentAsString(result)).contains("Total Questions: 1");
+              assertThat(contentAsString(result)).contains("All Questions");
+            });
+  }
+
+  private void buildQuestionsList() throws UnsupportedQuestionTypeException {
+    QuestionDefinitionBuilder builder =
+        new QuestionDefinitionBuilder()
+            .setId(1L)
```
hmhsu, line 147: nit (blocking): we can skip setting id in builder now. The default value is empty.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
```diff
@@ -0,0 +1,52 @@
...
+    QuestionForm form = new QuestionForm();
+    form.setQuestionName("name");
+    form.setQuestionDescription("description");
+    form.setQuestionPath("my.question.path");
+    form.setQuestionText("What is the question text?");
+    form.setQuestionHelpText("");
+    form.setQuestionType("TEXT");
+    QuestionDefinitionBuilder builder = form.getBuilder();
+
+    // The QuestionForm does not set version or id, which are needed in order to build the
```
hmhsu, line 28: nit (blocking): please update the comment, id is no longer required for new questionDefinition.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: Added Issue #237 as a follow up.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionRow.java
AWBaumann: Added Issue #237 as a follow up.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
hmhsu: If we are to update the exact row in db, we need the id from the existing question (hidden inputs) so eventually `QuestionForm` needs to have a `questionId` field (likely `questionVersion` as well).

For now, we no longer need to explicitly set `id` for builder to successfully build a (new) definition. Please remove it.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public Content render(Request request, Optional<QuestionDefinition> question) {
+    String headerText = question.isPresent() ? "Edit Question" : "New Question";
+    return layout.htmlContent(
+        body(
+            renderHeader(headerText),
+            buildQuestionForm(question).with(makeCsrfTokenInputTag(request))));
+  }
+
+  private ContainerTag buildQuestionForm(Optional<QuestionDefinition> optionalDefinition) {
```
kendrickt, line 40: suggestion (non-blocking): Would this be clearer as two methods, instead of trying to combine the `create new` and `edit` form into one method?

universal-application-tool-0.0.1/test/controllers/admin/QuestionControllerTest.java
AWBaumann: Done.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionTableCell.java
```diff
@@ -0,0 +1,71 @@
...
+      case QUESTION_TYPE:
+        return td(definition.getQuestionType().toString());
+      case ACTIONS:
+        return td().with(span("view"))
+            .with(span(" | "))
+            .with(
+                a("edit")
+                    .withHref(
+                        controllers.admin.routes.QuestionController.edit(definition.getPath())
+                            .toString()));
```
kendrickt, line 66: (blocking): use `routes.QuestionController.edit(...).url()` here instead of `toString`.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
```diff
@@ -0,0 +1,95 @@
...
+          QuestionTableCell.QUESTION_TYPE,
+          QuestionTableCell.ACTIONS);
+
+  @Inject
+  public QuestionsListView(BaseHtmlLayout layout) {
+    this.layout = layout;
+  }
+
+  /** Renders a page with either a table or a list view of all questions. */
+  public Content render(ImmutableList<Tag> questionContent) {
```
kendrickt, line 41: question (blocking): should this be a private method? Should any controller be allowed to call this render method and pass in whatever tags it wants?

kendrickt (changes requested):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
AWBaumann: Done

universal-application-tool-0.0.1/app/views/admin/questions/QuestionTableCell.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
AWBaumann: I feel like it'd be a lot of repeated code. The biggest difference is just whether we have data to populate the inputs. I think that we could probably reuse `QuestionForm` to clean it up a bit, though.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -42,6 +42,14 @@ public Tag renderHeader(String headerText) {
     return ImmutableList.of(labelTag, br(), inputTag, br(), br());
   }
 
+  public ImmutableList<DomContent> textInputWithLabel(
```
kendrickt, line 45: (blocking): what's the difference between these new methods and `textField` methods? Should we get rid of the `textField` methods?

universal-application-tool-0.0.1/app/forms/QuestionForm.java
```diff
@@ -3,16 +3,42 @@
...
+  private String questionName = "";
+  private String questionDescription = "";
+  private String questionPath = "";
+  private String questionType = "";
+  private String questionText = "";
+  private String questionHelpText = "";
+
+  public QuestionForm() {}
+
+  public QuestionForm(Optional<QuestionDefinition> definition) {
```
kendrickt, line 21: suggestion (non-blocking) this constructor should take a `QuestionDefinition` to keep the constructor more simple. and whatever is calling this can do something like

`QuestionForm form = qd.isPresent() ? new QuestionForm(qd) : new QuestionForm();`

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: suggestion (blocking): you could factor out the part that adds `QuestionForm` fields to the `formTag` and split this into an `edit` and `create` method. 

I find the optional input confusing. It would be much clearer to me if there were a `render()` and a `render(QuestionDefinition)`, and `buildquestionForm()` and `buildQuestionForm(QuestionDefinition)`.

kendrickt (changes requested):

kendrickt:
The `Optional<QuestionDefinition>` is confusing. It suggests the presence of `QuestionDefinition` may not be deterministic, but in each place it is used, we should know exactly whether it is present or not. I'd rather see a clear code path for when we are editing a  `QuestionDefinition`, and when we are creating a new one.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: They're similar, and in the end we'll keep one or the other, but I don't think we'll have a clear picture of what methods go in here until we start getting rolling with Tailwind.

kendrickt: can you group these together with the other ones?

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +27,64 @@
  */
 public abstract class BaseHtmlView {
 
+  public Tag renderHeader(String headerText) {
+    return h1(headerText);
+  }
+
+  protected ImmutableList<DomContent> textInputWithLabel(
+      String labelValue, String inputId, Optional<String> value) {
+    Tag labelTag = label(labelValue).attr("for", inputId);
```
carolinedanzi, line 36: suggestion: prefer to use `Attr` constants: https://javadoc.io/static/com.j2html/j2html/1.2.1/j2html/attributes/Attr.html

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
```diff
@@ -17,6 +27,64 @@
...
+      inputTag.withValue(value.get());
+    }
+
+    return ImmutableList.of(labelTag, br(), inputTag, br(), br());
+  }
+
+  public ImmutableList<DomContent> textInputWithLabel(
+      String labelValue, String inputId, String value) {
+    Optional<String> optionalValue =
+        (value == null || value.isEmpty()) ? Optional.empty() : Optional.of(value);
```
carolinedanzi, line 48: question: I think you can shorten this to `Optional.ofNullable(value)`, but I'm not sure value can/should be null, can it? (same for below)

universal-application-tool-0.0.1/app/views/admin/questions/QuestionsListView.java
carolinedanzi: Thanks for the explanation!

kendrickt (approved):

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Done.

AWBaumann: Done.

universal-application-tool-0.0.1/app/views/BaseHtmlView.java
AWBaumann: Replaced with `Optional<String> optionalValue = Optional.ofNullable(value).filter(s -> !s.trim().isEmpty());`

PR #225, opened 2021-02-23
Don't format unnecessary files.

ndmckinley:
### Description
Also move bin/fmt to use an ECR image.

### Checklist


hmhsu (approved):
Thanks for fixing it!

bion (approved):

PR #227, opened 2021-02-24
do not manually set question id

hmhsu:
### Description
We want ebean to auto-generate question id for us.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/models/Question.java
+++ b/universal-application-tool-0.0.1/app/models/Question.java
@@ -41,8 +41,8 @@ public class Question extends BaseModel {
 
   public Question(QuestionDefinition questionDefinition) {
     this.questionDefinition = checkNotNull(questionDefinition);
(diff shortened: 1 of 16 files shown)


kendrickt:
I think if we get rid of `QuestionService.create(QuestionDefinition)` and copy the constructor arguments for `QuestionDefinition` into `create` then the `QuestionDefinition` can just use `long` and not never need an optional.

hmhsu:
> I think if we get rid of `QuestionService.create(QuestionDefinition)` and copy the constructor arguments for `QuestionDefinition` into `create` then the `QuestionDefinition` can just use `long` and we would never need an optionallong.

We still decided to go with `OptionalLong`. This should be mostly hidden from users with overloaded constructors and builder except few cases in tests.

kendrickt (approved):

AWBaumann (approved):

PR #229, opened 2021-02-24
Add a ProgramIndexView for applicants

### LINKED ISSUE #178: Applicant can browse list of all programs offered in interface

carolinedanzi:
### Description
Adds an applicant controller that has two methods: `index` for displaying a list of available programs, and `edit`, which for now just returns a simple string (will be implemented later). `index` uses a new view to render the list of programs, with buttons for starting the application process

### Checklist

### Issue(s)
Fixes #178 

### Approvals
Want approval from all please

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
+++ b/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
@@ -1,0 +1,38 @@
+package controllers.applicant;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import javax.inject.Inject;
+import play.mvc.Controller;
+import play.mvc.Result;
+import services.program.ProgramService;
+import views.applicant.ProgramIndexView;
+
+/** Controller for handling methods for an applicant applying to programs. */
+public class ApplicantProgramsController extends Controller {
+
+  private final ProgramService programService;
+  private final ProgramIndexView programIndexView;
+
+  @Inject
+  public ApplicantProgramsController(
+      ProgramService programService, ProgramIndexView programIndexView) {
+    this.programService = checkNotNull(programService);
+    this.programIndexView = checkNotNull(programIndexView);
+  }
+
+  public CompletionStage<Result> index(long applicantId) {
+    return programService
+        .listProgramDefinitionsAsync()
+        .thenApply(programs -> ok(programIndexView.render(applicantId, programs)));
+  }
+
+  // TODO(https://github.com/seattle-uat/universal-application-tool/issues/224): Go to the next
+  // block for the chosen program
+  public CompletionStage<Result> edit(long applicantId, long programId) {
+    return CompletableFuture.completedFuture(
+        ok("Applicant " + applicantId + " chose program " + programId));
+  }
+}
--- a/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
+++ b/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
@@ -5,6 +5,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
+import play.libs.concurrent.HttpExecutionContext;
 import play.mvc.Controller;
 import play.mvc.Result;
 import services.program.ProgramService;
@@ -13,12 +14,16 @@ import views.applicant.ProgramIndexView;
 /** Controller for handling methods for an applicant applying to programs. */
 public class ApplicantProgramsController extends Controller {
 
+  private final HttpExecutionContext httpContext;
   private final ProgramService programService;
   private final ProgramIndexView programIndexView;
 
   @Inject
   public ApplicantProgramsController(
-      ProgramService programService, ProgramIndexView programIndexView) {
+      HttpExecutionContext httpContext,
+      ProgramService programService,
+      ProgramIndexView programIndexView) {
+    this.httpContext = httpContext;
     this.programService = checkNotNull(programService);
     this.programIndexView = checkNotNull(programIndexView);
   }
@@ -26,7 +31,8 @@ public class ApplicantProgramsController extends Controller {
   public CompletionStage<Result> index(long applicantId) {
     return programService
         .listProgramDefinitionsAsync()
-        .thenApply(programs -> ok(programIndexView.render(applicantId, programs)));
+        .thenApplyAsync(
+            programs -> ok(programIndexView.render(applicantId, programs)), httpContext.current());
   }
 
   // TODO(https://github.com/seattle-uat/universal-application-tool/issues/224): Go to the next
--- a/universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
+++ b/universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
@@ -1,4 +1,4 @@
-package views;
+package views.applicant;
 
 import static j2html.TagCreator.body;
 import static j2html.TagCreator.head;
@@ -7,6 +7,8 @@ import static j2html.TagCreator.main;
 import j2html.tags.DomContent;
 import javax.inject.Inject;
 import play.twirl.api.Content;
+import views.BaseHtmlLayout;
+import views.ViewUtils;
 
 public class ApplicantLayout extends BaseHtmlLayout {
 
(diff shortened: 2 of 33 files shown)


universal-application-tool-0.0.1/conf/routes
```diff
@@ -24,6 +24,10 @@ POST    /admin/programs             controllers.admin.AdminProgramController.cre
 GET     /demo/new                   controllers.J2HtmlDemoController.newOne(request: Request)
 POST    /demo                       controllers.J2HtmlDemoController.create(request: Request)
 
+# Methods for applicants
+GET /applicants/:id/programs        controllers.applicant.ApplicantProgramsController.index(id: Long)
+GET /applicants/:aid/programs/:pid  controllers.applicant.ApplicantProgramsController.edit(aid: Long, pid: Long)
```
bion, line 29: nit: route for edit action should be `/applicants/:id/programs/:id/edit`

suggestion (non-blocking): recommend spelling the IDs out for readability e.g. `applicantId: Long, programId: Long`

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -0,0 +1,38 @@
...
+  public ApplicantProgramsController(
+      ProgramService programService, ProgramIndexView programIndexView) {
+    this.programService = checkNotNull(programService);
+    this.programIndexView = checkNotNull(programIndexView);
+  }
+
+  public CompletionStage<Result> index(long applicantId) {
+    return programService
+        .listProgramDefinitionsAsync()
+        .thenApply(programs -> ok(programIndexView.render(applicantId, programs)));
```
bion, line 29: Suggestion (blocking): when continuing from async calls like this the play docs recommend explicitly specifying the `HttpExecutionContext` to ensure that the class loader is available in its correct state.

```java
// inject it in the c'tor
@Inject
ApplicantProgramsController(HttpExecutionContext httpExecutionContext)
```
then
```java
         .thenApply(programs -> ok(programIndexView.render(applicantId, programs)),
```

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
```diff
@@ -1,4 +1,4 @@
-package views;
+package views.applicant;
```
bion, line 1: praise: 👍🏻

universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
```diff
@@ -0,0 +1,54 @@
...
+            .with(h1("Programs"))
+            .with(each(programs, program -> shortProgram(applicantId, program))));
+  }
+
+  private Tag shortProgram(long applicantId, ProgramDefinition program) {
+    return div()
+        .with(h2(program.name()))
+        .with(p(program.description()))
+        .with(
+            redirectButton(
```
bion, line 47: Suggestion (blocking): use an `a` tag instead of this method. If we need it to look like a button we can style it as such later. `redirectButton` button isn't accessible so I'd like to remove it eventually.

universal-application-tool-0.0.1/test/BrowserTest.java
```diff
@@ -31,24 +35,49 @@ protected TestBrowser provideBrowser(int port) {
...
+
+    goTo(routes.ProfileController.myProfile());
+    assertThat(browser.pageSource()).contains("GuestClient");
+    assertThat(browser.pageSource()).contains("{\"created_time\":");
+    assertThat(browser.pageSource()).contains(Roles.ROLE_APPLICANT.toString());
+  }
+
+  private void goTo(Call method) {
+    browser.goTo(BASE_URL + method.url());
+  }
```
bion, line 76: Suggestion (non-blocking): this looks super helpful to have for browser tests! What do you think of adding it to the superclass?

universal-application-tool-0.0.1/test/BrowserTest.java
```diff
@@ -31,24 +35,49 @@ protected TestBrowser provideBrowser(int port) {
   }
 
   @Test
-  public void test() {
+  public void homePage() {
     browser.goTo("http://localhost:" + play.api.test.Helpers.testServerPort());
```
bion, line 39: nit: the hard-coded string can be replaced with `LOCALHOST` now

bion: Actually the whole arg can be replaced with `BASE_URL`

universal-application-tool-0.0.1/test/BrowserTest.java
```diff
@@ -31,24 +35,49 @@ protected TestBrowser provideBrowser(int port) {
...
+  @Test
+  public void applicantProgramList_selectApply_redirectsToEdit() {
+    Program program = insertProgram("My Program");
+    goTo(controllers.applicant.routes.ApplicantProgramsController.index(1L));
+    assertThat(browser.pageSource()).contains("My Program");
+
+    // Redirect when "apply" button is clicked.
+    String applyButtonId = "#apply" + program.id;
+    browser.$(applyButtonId).click();
+    assertThat("/" + browser.url())
```
bion, line 52: suggestion (non-blocking): this string cat is a little cryptic. Recommend moving into a named method, maybe in the superclass since it looks like something we'll want to do in a few tests

universal-application-tool-0.0.1/test/BrowserTest.java
bion: Just realized we don't have a superclass for these yet.

I think we'll want to separate our browser tests into multiple files as they get bigger and more involved. If you'd like to split them out now and make a superclass for them go ahead, otherwise please file an issue to do it.

bion (approved):

universal-application-tool-0.0.1/test/BrowserTest.java
carolinedanzi: Good call - I filed https://github.com/seattle-uat/universal-application-tool/issues/232 and will work on that today

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Test
+  public void edit_returnsTemporaryString() {
+    Result result = controller.edit(123L, 456L).toCompletableFuture().join();
+
+    assertThat(result.status()).isEqualTo(OK);
+    assertThat(contentAsString(result)).isEqualTo("Applicant 123 chose program 456");
+  }
+
+  private static Program insertProgram(String name) {
```
natsid, line 63: suggestion (non-blocking): Would it make sense to share this helper method with the [same one in AdminProgramsController](https://github.com/seattle-uat/universal-application-tool/blob/main/universal-application-tool-0.0.1/test/controllers/admin/AdminProgramControllerTest.java#L148)?

I'm not sure if it's stylistically desirable in Java, but in TypeScript, we'll often have separate test utility files that have helpers like this. This method is pretty simple, but it might make sense to break it out in a future PR, as our tests add more common database interactions.

natsid (approved):

universal-application-tool-0.0.1/conf/routes
carolinedanzi: Done for both - thanks for the reminder on `edit`

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramsController.java
carolinedanzi: Thanks for linking the docs, I completely missed that section

universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
carolinedanzi: Thanks! Learned something new :) This is now an a tag

universal-application-tool-0.0.1/test/BrowserTest.java
carolinedanzi: Thanks for catching that, it was my reason for making the string in the first place

universal-application-tool-0.0.1/test/BrowserTest.java
carolinedanzi: I created a new `assertUrlEquals` method and will add it to the base class in #232

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: That's a really good idea, we can have test utilities. I filed https://github.com/seattle-uat/universal-application-tool/issues/234

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
carolinedanzi: :)

PR #230, opened 2021-02-24
Add ErrorProne to our compiler, and also fix all existing ErrorProne warnings and errors.

### LINKED ISSUE #21: Look into adding ErrorProne to the project
[Error Prone](https://github.com/google/error-prone) is a static analysis tool for Java that catches common programming mistakes at compile-time. It's designed for plug-and-play usage with several build systems that are not the one we're using. This task is to take a look at it and get an idea of how much work it'd be to set up for our project. If it ends up being fairly straightforward, go ahead and do it.

ndmckinley:
### Checklist

### Issue(s)
Fixes #21.

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/UatProfileData.java
+++ b/universal-application-tool-0.0.1/app/auth/UatProfileData.java
@@ -4,6 +4,7 @@ import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import com.google.common.base.Preconditions;
 import java.time.Clock;
+import java.time.ZoneId;
 import models.Account;
 import models.Applicant;
 import org.pac4j.core.profile.CommonProfile;
@@ -20,7 +21,7 @@ public class UatProfileData extends CommonProfile {
   private Clock clock;
 
   public UatProfileData() {
-    this(Clock.systemDefaultZone());
+    this(Clock.system(ZoneId.of("America/Los_Angeles")));
   }
 
   public UatProfileData(Clock clock) {
--- a/universal-application-tool-0.0.1/app/modules/MainModule.java
+++ b/universal-application-tool-0.0.1/app/modules/MainModule.java
@@ -1,5 +1,8 @@
+package modules;
+
 import com.google.inject.AbstractModule;
 import java.time.Clock;
+import java.time.ZoneId;
 import repository.AmazonS3Client;
 import services.program.ProgramService;
 import services.program.ProgramServiceImpl;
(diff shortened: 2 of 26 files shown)


universal-application-tool-0.0.1/app/modules/MainModule.java
```diff
@@ -14,12 +17,12 @@
  * create modules in other locations by adding `play.modules.enabled` settings to the
  * `application.conf` configuration file.
  */
-public class Module extends AbstractModule {
+public class MainModule extends AbstractModule {
```
carolinedanzi, line 20: question: is this ok with Play? For some reason in my head this needed to be called "Module", but looks like that's not true since the tests pass

edit: https://www.playframework.com/documentation/2.8.x/JavaPlayModules#Module-registration - I think I got confused by "will load any class called `Module`" and took that literally - I think this is fine

universal-application-tool-0.0.1/app/auth/UatProfileData.java
```diff
@@ -20,7 +21,7 @@
   private Clock clock;
 
   public UatProfileData() {
```
carolinedanzi, line 23: suggestion (non-blocking): At some point we should probably change this to an `@Inject` constructor that takes the `Clock` as a parameter, since it's already bound in the main module. That way we won't need to keep track of two clocks

universal-application-tool-0.0.1/build.sbt
```diff
@@ -49,12 +49,18 @@ lazy val root = (project in file("."))
...
+      // Errorprone
+      "com.google.errorprone" % "error_prone_core" % "2.5.1",
     ),
     javacOptions ++= Seq(
       "-encoding", "UTF-8",
       "-parameters",
       "-Xlint:unchecked",
       "-Xlint:deprecation",
+      "-XDcompilePolicy=simple",
+      "-Xplugin:ErrorProne -Xep:AutoValueSubclassLeaked:OFF",
```
carolinedanzi, line 62: question: what does this mean - is it ok that it's off?

carolinedanzi (approved):
Beautiful! Love it! Super happy to have ErrorProne enabled

universal-application-tool-0.0.1/app/auth/UatProfileData.java
ndmckinley: Sure - I don't understand how to call an `@Inject` constructor, though.  Can you make the change in a tiny PR and send it to me so I can see how it works?

universal-application-tool-0.0.1/app/modules/MainModule.java
ndmckinley: Oh, no, you are right, but ErrorProne doesn't allow default-package classes.  So instead we just specify it directly in `application.conf`.

universal-application-tool-0.0.1/build.sbt
ndmckinley: Added comment - it's an error present in the generated code, so nothing we can do about it.

PR #231, opened 2021-02-24
allow question service to create a question entry without a questionD…

hmhsu:
…efinition class

### Description

### Checklist

### Issue(s)
[Fixes] #<issue_number>

### DIFF
--- a/universal-application-tool-0.0.1/app/models/Question.java
+++ b/universal-application-tool-0.0.1/app/models/Question.java
@@ -50,6 +50,16 @@ public class Question extends BaseModel {
     questionType = questionDefinition.getQuestionType().toString();
   }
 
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/models/Question.java
```diff
@@ -50,6 +50,16 @@ public Question(QuestionDefinition questionDefinition) {
     questionType = questionDefinition.getQuestionType().toString();
   }
 
+  public Question(long version, String path, String name, String description, ImmutableMap<Locale, String> questionText, ImmutableMap<Locale, String> questionHelpText, String questionType) {
```
bion, line 53: Suggestion: It's a lot of arguments to pass around individually. Another alternative is to create a play form object specifically for question creation and editing. ID would be optional on it since it won't exist when the form is submitted for create.

https://www.playframework.com/documentation/2.8.x/JavaForms#Defining-a-form

Issue #194, opened 2021-02-19
Implement path validation when inserting new questions


carolinedanzi:
**Describe the solution you'd like**
The QuestionService states that create "will fail if the path provided already resolves to a QuestionDefinition or Scalar." We need to implement this logic in the service implementation

**Done when**
A new question insertion fails when the path already exists

lianne-oy:
M1 Review - Not worked on

PR #238, opened 2021-02-25
Add path validation for creating new questions

hmhsu:
### Checklist

### Issue(s)
[Fixes] #194

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -55,6 +55,8 @@ public class QuestionRepository {
     }
 
     public static boolean pathConflicts(String path, String otherPath) {
+      path = path.toLowerCase();
+      otherPath = otherPath.toLowerCase();
       return path.startsWith(otherPath) || otherPath.startsWith(path);
     }
   }
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -34,21 +34,21 @@ public class QuestionRepository {
   }
 
   static class PathConflictDetector {
-    private boolean conflict = false;
+    private Optional<Question> conflictedQuestion = Optional.empty();
     private final String newPath;
 
     PathConflictDetector(String newPath) {
       this.newPath = checkNotNull(newPath);
     }
 
-    public Boolean hasConflict() {
-      return Boolean.valueOf(conflict);
+    public Optional<Question> getConflictedQuestion() {
+      return conflictedQuestion;
     }
 
     public boolean checkConflict(Question question) {
       boolean proceed = true;
       if (pathConflicts(question.getPath(), newPath)) {
-        conflict = true;
+        conflictedQuestion = Optional.of(question);
         proceed = false;
       }
       return proceed;
@@ -61,12 +61,12 @@ public class QuestionRepository {
     }
   }
 
-  public CompletionStage<Boolean> lookupPathConflict(String newPath) {
+  public CompletionStage<Optional<Question>> findConflictingQuestion(String newPath) {
     return supplyAsync(
         () -> {
           PathConflictDetector detector = new PathConflictDetector(newPath);
           ebeanServer.find(Question.class).findEachWhile(detector::checkConflict);
-          return detector.hasConflict();
+          return detector.getConflictedQuestion();
         },
         executionContext);
   }
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -33,6 +33,38 @@ public class QuestionRepository {
         executionContext);
   }
 
+  static class PathConflictDetector {
+    Boolean conflict = Boolean.FALSE;
+    String newPath;
+
+    public PathConflictDetector(String newPath) {
+      this.newPath = checkNotNull(newPath);
+    }
+
+    public Boolean hasConflict() {
+      return conflict;
+    }
+
+    public boolean checkConflict(Question question) {
+      boolean proceed = true;
+      if (question.pathConflictsWith(newPath)) {
+        conflict = Boolean.TRUE;
+        proceed = false;
+      }
+      return proceed;
+    }
+  }
+
+  public CompletionStage<Boolean> lookupPathConflict(String newPath) {
+    return supplyAsync(
+        () -> {
+          PathConflictDetector detector = new PathConflictDetector(newPath);
+          ebeanServer.find(Question.class).findEachWhile(detector::checkConflict);
+          return detector.hasConflict();
+        },
+        executionContext);
+  }
+
   public CompletionStage<Optional<Question>> lookupQuestionByPath(String path) {
     return supplyAsync(
         () ->
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -33,6 +33,21 @@ public class QuestionRepository {
         executionContext);
   }
 
+  public CompletionStage<Optional<Question>> lookupQuestionByPath(String path) {
+    return supplyAsync(
+        () ->
+            Optional.ofNullable(
+                ebeanServer
+                    .find(Question.class)
+                    .where()
+                    .eq("path", path)
+                    .orderBy()
+                    .desc("version")
+                    .setMaxRows(1)
+                    .findOne()),
+        executionContext);
+  }
+
   public CompletionStage<Question> insertQuestion(Question question) {
     return supplyAsync(
         () -> {
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -34,21 +34,21 @@ public class QuestionRepository {
   }
 
   static class PathConflictDetector {
-    Boolean conflict = Boolean.FALSE;
-    String newPath;
+    private boolean conflict = false;
+    private final String newPath;
 
-    public PathConflictDetector(String newPath) {
+    PathConflictDetector(String newPath) {
       this.newPath = checkNotNull(newPath);
     }
 
     public Boolean hasConflict() {
-      return conflict;
+      return Boolean.valueOf(conflict);
     }
 
     public boolean checkConflict(Question question) {
       boolean proceed = true;
       if (pathConflicts(question.getPath(), newPath)) {
-        conflict = Boolean.TRUE;
+        conflict = true;
         proceed = false;
       }
       return proceed;
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -47,11 +47,15 @@ public class QuestionRepository {
 
     public boolean checkConflict(Question question) {
       boolean proceed = true;
-      if (question.pathConflictsWith(newPath)) {
+      if (pathConflicts(question.getPath(), newPath)) {
         conflict = Boolean.TRUE;
         proceed = false;
       }
       return proceed;
+    }
+
+    public static boolean pathConflicts(String path, String otherPath) {
+      return path.startsWith(otherPath) || otherPath.startsWith(path);
     }
   }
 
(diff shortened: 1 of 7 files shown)


universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,57 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+    }
+
+    public Boolean hasConflict() {
+      return conflict;
+    }
+
+    public boolean checkConflict(Question question) {
+      boolean proceed = true;
+      if (pathConflicts(question.getPath(), newPath)) {
+        conflict = Boolean.TRUE;
```
bion, line 51: question: why `Boolean.TRUE` rather than `true`?

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,57 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
         executionContext);
   }
 
+  static class PathConflictDetector {
+    Boolean conflict = Boolean.FALSE;
+    String newPath;
```
bion, line 38: Suggestion (blocking): `private final` for `newPath`, `private` for `conflict`

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,57 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+
+    public PathConflictDetector(String newPath) {
```
bion, line 40: Question (blocking): does this need to be public? the class itself is package-private and I don't see it accessed outside of this package.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: It ultimately needs to be passed to `CompletionStage<Boolean>`, we can convert the type later though. (I am not sure if `CompletionStage<boolean>` is allowed. If so we should just use that.)

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Sure thing. Will change it.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,57 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+  public CompletionStage<Optional<Question>> lookupQuestionByPath(String path) {
+    return supplyAsync(
+        () ->
+            Optional.ofNullable(
+                ebeanServer
+                    .find(Question.class)
+                    .where()
+                    .eq("path", path)
+                    .orderBy()
+                    .desc("version")
```
bion, line 81: Suggestion (blocking): we have implemented versioning yet, I recommend removing this code for now since it may cause confusion

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -53,4 +55,14 @@ public QuestionDefinition update(QuestionDefinition definition) {
                     .map(question -> question.getQuestionDefinition())
                     .collect(ImmutableList.toImmutableList()));
   }
+
+  private boolean isValid(QuestionDefinition definition) {
```
AWBaumann, line 59: We probably also want to run this through a url encoder. I think any string where the encoded value isn't equal to the raw value should be denied. We can also validate that client side, but as a general rule client side validation should only be done as a service to the user, not to ensure things are legitimate on the server side.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
bion: Ah I see. `CompletionStage<boolean>` isn't allowed because Java doesn't support primitives in generics.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
hmhsu: That definitely sounds something we want to do! We probably want to do that for all string fields. For now, let me add a super basic path pattern matching check. PTAL.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -53,4 +55,18 @@ public QuestionDefinition update(QuestionDefinition definition) {
...
+      return false;
+    }
+    boolean hasConflict =
+        questionRepository.lookupPathConflict(newPath).toCompletableFuture().join().booleanValue();
+    return !hasConflict;
+  }
+
+  private boolean validPathPattern(String path) {
+    return path.matches("^([a-z]+[.])*[a-z]+$");
+  }
```
AWBaumann, line 71: How about using ``java.net.URLEncoder` here?

```
import java.net.URLEncoder;
import java.nio.charset.StandardCharsets;

public boolean validPathPattern(String path) { 
  try {
    return URLEncoder.encode(path, StandardCharsets.UTF_8.name()).equals(path);
  } catch (Exception e) {
    return false;
  }
}
```

hmhsu: Good idea. Done.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
hmhsu: Done.

AWBaumann (approved):

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,52 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+      otherPath = otherPath.toLowerCase();
+      return path.startsWith(otherPath) || otherPath.startsWith(path);
+    }
+  }
+
+  public CompletionStage<Boolean> lookupPathConflict(String newPath) {
+    return supplyAsync(
+        () -> {
+          PathConflictDetector detector = new PathConflictDetector(newPath);
+          ebeanServer.find(Question.class).findEachWhile(detector::checkConflict);
```
bion, line 68: Question (blocking): I would be curious to know how this behaves when there are 100s of questions. Does it generate a separate query for each question? If so it could end up being very slow.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,52 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+    }
+
+    public static boolean pathConflicts(String path, String otherPath) {
+      path = path.toLowerCase();
+      otherPath = otherPath.toLowerCase();
+      return path.startsWith(otherPath) || otherPath.startsWith(path);
+    }
+  }
+
+  public CompletionStage<Boolean> lookupPathConflict(String newPath) {
```
bion, line 64: Suggestion (blocking): this method is called `lookupPathConflict` but returns a `boolean`. It seems likely that the question it's in conflict with could be useful for displaying a helpful error message so I recommend returning the question it's in conflict with and changing the name to `findConflictingQuestion`

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -53,4 +57,22 @@ public QuestionDefinition update(QuestionDefinition definition) {
...
+    }
+    boolean hasConflict =
+        questionRepository.lookupPathConflict(newPath).toCompletableFuture().join().booleanValue();
+    return !hasConflict;
+  }
+
+  private boolean isValidPathPattern(String path) {
+    try {
+      return URLEncoder.encode(path, StandardCharsets.UTF_8.name()).equals(path);
+    } catch (Exception e) {
```
bion, line 74: Suggestion (blocking): never catch the generic `Exception` since it could hide bugs that have nothing to do with the exception you're expecting to catch here. Instead, catch the exception you expect `URLEncoder.encode` to throw on bad input specifically (you can use a test to discover what it is).

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
bion: You can examine the behavior by turning SQL logging on and running a test that has many questions in the DB.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
hmhsu: It turns out that this is not necessary. The only exception is thrown when the `Charset` name is unrecognized. I changed it to the static field.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: From the test I ran on my local machine with 200+ questions, Ebean only issues one select statement. From the [document](https://ebean.io/docs/query/findEachWhile) this is actually recommended for large queries. For smaller queries `findList` is more performant.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
bion: Glad to hear it, thanks!

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,52 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+    public boolean checkConflict(Question question) {
+      boolean proceed = true;
+      if (pathConflicts(question.getPath(), newPath)) {
+        conflictedQuestion = Optional.of(question);
+        proceed = false;
+      }
+      return proceed;
+    }
+
+    public static boolean pathConflicts(String path, String otherPath) {
```
bion, line 57: nit: this can be a package-private since the test is in the same package

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,52 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
+
+    PathConflictDetector(String newPath) {
+      this.newPath = checkNotNull(newPath);
+    }
+
+    public Optional<Question> getConflictedQuestion() {
+      return conflictedQuestion;
+    }
+
+    public boolean checkConflict(Question question) {
```
bion, line 48: nit: this can be a package-private

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
```diff
@@ -33,6 +33,52 @@ public QuestionRepository(EbeanConfig ebeanConfig, DatabaseExecutionContext exec
...
 
+  static class PathConflictDetector {
+    private Optional<Question> conflictedQuestion = Optional.empty();
+    private final String newPath;
+
+    PathConflictDetector(String newPath) {
+      this.newPath = checkNotNull(newPath);
+    }
+
+    public Optional<Question> getConflictedQuestion() {
```
bion, line 44: nit: this can be a package-private

bion (approved):

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

universal-application-tool-0.0.1/app/repository/QuestionRepository.java
hmhsu: Done.

PR #253, opened 2021-02-25
Convert QuestionHelpText to ImmutableMap, unwrapping the unnecessary Optional.

### LINKED ISSUE #237: Improve/Clean up question text and help text methods.
**Done when**
`QuestionDefinition.questionHelpText` is an `ImmutableMap<Locale, String>` instead of `Optional<ImmutableMap<Locale, String>` and we have better methods for getting these strings.

AWBaumann:
### Description
It's not necessary to have an `Optional<ImmutableList>` so we're unwrapping it.

### Checklist

### Issue(s)
#237

### Follow ups
In order to close #237 we should implement `QuestionDefinition.getPreferredQuestionText()` and `QuestionDefinition.getPreferredQuestionHelpText()` that uses a global `preferredLocale` variable so we're not passing around `Locale.ENGLISH` everywhere.

### DIFF
--- a/universal-application-tool-0.0.1/app/forms/QuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/QuestionForm.java
@@ -2,7 +2,6 @@ package forms;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.Locale;
-import java.util.Optional;
 import services.question.QuestionDefinition;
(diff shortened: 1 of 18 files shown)


carolinedanzi (approved):
This looks nice, thank you for doing this!

PR #259, opened 2021-02-26
Update package-lock.json with webpack deps

bion:
This diff is created when I run `./bin/fmt` on main right now. It includes a bunch of stuff about webpack. Was this missed in the webpack PR?


natsid:
Iiiinteresting...

I wouldn't think anything was missed since I definitely ran `bin/run-test` (which should have installed everything / updated the package-lock.json accordingly) before merging. Not exactly sure why this would have such a diff.

PR #261, opened 2021-02-26
Swap to alpine.  Ran and confirmed no changes on main.

ndmckinley:
### Description
The formatter used to use a debian derived OS, which provided a different default version of NPM, which wrote incompatible lockfiles, since `tsfmt` actually reads and rewrites the lockfile.  Now we use the same version in `uat` and `fmt`.  This will fix the issue raised in https://github.com/seattle-uat/universal-application-tool/pull/259.

### Checklist

### DIFF
--- a/formatter.Dockerfile
+++ b/formatter.Dockerfile
@@ -1,10 +1,8 @@
-FROM openjdk:slim
-RUN apt-get update
-RUN apt-get install -y wget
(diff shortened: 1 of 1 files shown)


hmhsu (approved):
Thanks for fixing this!

bion (approved):

Issue #196, opened 2021-02-19
Implement a .equals() method for QuestionDefinition


carolinedanzi:
**Is your feature request related to a problem? Please describe.**
Testing QuestionDefinition equality in unit tests is annoying, since there is no `.equals()` method override

**Describe the solution you'd like**
Override `.equals()` in QuestionDefinition and update current unit tests to use this for equality checks

**Done when**
Unit tests use `assertThat().isEqualTo()` for QuestionDefinitions

carolinedanzi:
Putting this on pause for now - we're rethinking the QuestionDefinition classes, and having subtypes complicates things. .equals is a nice-to-have

lianne-oy:
M1 Review - Not worked on currently

PR #272, opened 2021-02-27
Update QuestionDefinition with `equals` and `equalsIgnoreId` methods.

AWBaumann:
### Description
Overrides equals on QuestionDefinition.

### Checklist

### Issue(s)
[Fixes] #196


universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -134,4 +135,43 @@ public String getQuestionHelpText(Locale locale) throws TranslationNotFoundExcep
...
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other instanceof QuestionDefinition) {
+      QuestionDefinition o = (QuestionDefinition) other;
+
+      return this.equalsIgnoreId(other)
+          && this.isPersisted() == o.isPersisted()
+          && (!this.isPersisted() || this.getId() == o.getId());
```
bion, line 151: Thought (blocking): imo two models should be considered the same model if they have the same ID. If we want to assert that their state is the same we can use `equalsIgnoreId`. I'm curious how others feel about this though. If folks are up for it, I recommend changing the boolean logic here to:

```java
return (this.isPersisted() && this.getId() == o.getId())
    || this.equalsIgnoreId(other);
```

AWBaumann: Wouldn't this break the `equals` contract because there would be instances where `a.equals(b)` is true but `b.equals(a)` is false?

bion: Ah that's a good point 🤔

hmhsu: imo we want to keep the convention of `equals`, i.e. checking all fields. Sounds like comparing IDs is valuable, then we can have a `IdEquals` method?

AWBaumann: So maybe we want 

```
public boolean equals(Object other) {
  return this.idEquals(other) && this.equalsIgnoreId(other);
}

public boolean idEquals(Object other) {
    if (other instanceof QuestionDefinition) {
      QuestionDefinition o = (QuestionDefinition) other;

      return this.isPersisted() == o.isPersisted()
          && (!this.isPersisted() || this.getId() == o.getId());
    }

    return false; // or do we want to return true even when the class is different?
}

public boolean equalsIgnoreId(Object other) {
    if (other instanceof QuestionDefinition) {
      QuestionDefinition o = (QuestionDefinition) other;

      return this.getQuestionType().equals(o.getQuestionType())
          && this.version == o.getVersion()
          && this.name.equals(o.getName())
          && this.path.equals(o.getPath())
          && this.description.equals(o.getDescription())
          && this.questionText.equals(o.getQuestionText())
          && this.questionHelpText.equals(o.getQuestionHelpText());
    }
    return false;
}
```

bion: In that case they can't be `equal` unless they're persisted. I'd be ok with only checking the IDs in `equals` only if both are persisted.

bion: And also if one is persisted and the other isn't they're not equal

hmhsu: The change looks good to me. I couldn't think of an example where we want to compare a non-persisted question to a persisted one and expect them to be equal. (If we are to edit an existing question, we would have loaded it from DB and so it has an id.) I think we persist a new question as soon as it is created, or at least we prefer to do so. Are there any cases that this is not true?

Issue #277, opened 2021-03-01
Implement "Address" question type


neel-a-mehta:
See go/uat-question-requirements

PR #292, opened 2021-03-02
Implement the address question renderer

carolinedanzi:
### Description
Implement the rendering logic for address questions

### Checklist

### Issue(s)
#277
Screen Shot 2021-03-01 at 6 08 11 PM

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
@@ -35,6 +35,15 @@ public class ApplicantQuestion {
     }
   }
 
+  public AddressQuestion getAddressQuestion() {
+    if (!getType().equals(QuestionType.ADDRESS)) {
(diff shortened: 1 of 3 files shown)


carolinedanzi:
@bion - there are no current tests for the rendering logic, shall I add some? Or do you feel they are not needed since it's mostly boilerplate?

bion (approved):
I think it's ok to add without unit test. However I think it'd be good for us to introduce a browser test that includes making questions with all of the different question types and  filling them out as the applicant. That's not in the scope of this PR though.

carolinedanzi:
> I think it's ok to add without unit test. However I think it'd be good for us to introduce a browser test that includes making questions with all of the different question types and filling them out as the applicant. That's not in the scope of this PR though.

Thanks for clarifying - I filed https://github.com/seattle-uat/universal-application-tool/issues/318 for the browser tests (we have no current way in the browser to navigate to the applicant program list, so this is tricky for now)

PR #293, opened 2021-03-02
compile and save dependencies in docker image

hmhsu:
We don't want to refetch dependencies every time we run the tests.

### Checklist

### Issue(s)
[Fixes]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ENV PROJECT_HOME /usr/src
 ENV PROJECT_NAME universal-application-tool-0.0.1
 
 COPY ${PROJECT_NAME} ${PROJECT_HOME}/${PROJECT_NAME}
-RUN cd $PROJECT_HOME/$PROJECT_NAME && sbt reload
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -29,7 +29,7 @@ ENV PROJECT_HOME /usr/src
 ENV PROJECT_NAME universal-application-tool-0.0.1
 
 COPY ${PROJECT_NAME} ${PROJECT_HOME}/${PROJECT_NAME}
-RUN cd $PROJECT_HOME/$PROJECT_NAME && sbt reload
+RUN cd $PROJECT_HOME/$PROJECT_NAME && sbt reload compile
```
hmhsu, line 32: I don't know why we removed `compile` when we set up tailwind. Let me know if compiling here breaks anything.

bion (approved):

Dockerfile
natsid: I don't actually know why either... But if you tested this locally, then I guess it's probably fine?

natsid (approved):

PR #328, opened 2021-03-02
Pull Authorizers out of Roles into its own class, and add authorizers…

kendrickt:
… to admin controller methods.

### Description
Authorizers cannot _just_ be an enum because annotations require values to be constants and a `toString()` method on the enum class doesn't work. 

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/Authorizers.java
+++ b/universal-application-tool-0.0.1/app/auth/Authorizers.java
@@ -1,0 +1,36 @@
+package auth;
+
+/**
+ * This enum represents Authorizers to be used with pac4j.
+ *
+ * <p>The enum values should always look like &#60;Authorizer&#62;(Labels.&#60;Authorizer&#62;).
+ */
+public enum Authorizers {
+  APPLICANT(Labels.APPLICANT),
+  UAT_ADMIN(Labels.UAT_ADMIN),
+  TI(Labels.TI),
+  PROGRAM_ADMIN(Labels.PROGRAM_ADMIN);
+
+  /**
+   * This Labels class is used to provide constant variables for annotations. This nested static
+   * class should only contain constant string values that are used to initialize {@link
+   * Authorizers}.
+   */
+  public static final class Labels {
+    public static final String APPLICANT = "applicant";
+    public static final String UAT_ADMIN = "uatadmin";
+    public static final String TI = "trustedintermediary";
+    public static final String PROGRAM_ADMIN = "programadmin";
+  }
+
+  private final String label;
+
+  private Authorizers(String label) {
+    this.label = label;
+  }
+
+  @Override
+  public String toString() {
+    return this.label;
+  }
+}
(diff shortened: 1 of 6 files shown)


kendrickt:
Is this too much? Could we just use a list of static strings in an `Authorizers` class (`public static final String UAT_ADMIN = "uatadmin";`)?

universal-application-tool-0.0.1/app/auth/Authorizers.java
```diff
@@ -0,0 +1,36 @@
...
+   * This Labels class is used to provide constant variables for annotations. This nested static
+   * class should only contain constant string values that are used to initialize {@link
+   * Authorizers}.
+   */
+  public static final class Labels {
+    public static final String APPLICANT = "applicant";
+    public static final String UAT_ADMIN = "uatadmin";
+    public static final String TI = "trustedintermediary";
+    public static final String PROGRAM_ADMIN = "programadmin";
+  }
```
bion, line 24: Suggestion (blocking): I don't think the inner `Labels` class is necessary. You can initialize the enums on lines 9-12 with string literals and then access them external to this class using `toString()`

universal-application-tool-0.0.1/app/auth/Authorizers.java
```diff
@@ -0,0 +1,36 @@
...
+  APPLICANT(Labels.APPLICANT),
+  UAT_ADMIN(Labels.UAT_ADMIN),
+  TI(Labels.TI),
+  PROGRAM_ADMIN(Labels.PROGRAM_ADMIN);
+
+  /**
+   * This Labels class is used to provide constant variables for annotations. This nested static
+   * class should only contain constant string values that are used to initialize {@link
+   * Authorizers}.
+   */
```
bion, line 18: suggestion (blocking): please mention in the comment that the labels are necessary for making constant references in annotations

universal-application-tool-0.0.1/app/auth/Authorizers.java
```diff
@@ -0,0 +1,36 @@
+package auth;
+
+/**
+ * This enum represents Authorizers to be used with pac4j.
+ *
+ * <p>The enum values should always look like &#60;Authorizer&#62;(Labels.&#60;Authorizer&#62;).
```
hmhsu, line 6: Question: Is this (`&#60;` etc) necessary? Can we remove it? Also I am curious how you generate these.

kendrickt: I just looked up how to get < and > characters in javadoc. So it'll render like `<Authorizer>(Labels.<Authorizer>)`. I should have used `&lt;` and `&gt;`!

universal-application-tool-0.0.1/app/auth/Authorizers.java
kendrickt: done!

bion (approved):

PR #329, opened 2021-03-02
Inject localized messages for applicant program view

carolinedanzi:
### Description
1. Adds a column on the Applicant table for preferredLocale
2. On login for applicants sets the PLAY_LANG cookie to the applicant's preferred locale
2. Injects `MessagesApi` into applicant controllers so messages are rendered in the preferred locale based on the PLAY_LANG cookie set on the request
2. Adds `messages.en` for our default messages and `messages.es-US` as another locale - primarily for testing this worked (note: I am not a native Spanish speaker so translations are rough)
3. Localizes the `ProgramIndexView` and `ApplicantProgramBlockEditView` pages for now as an examples (having two helped test that the cookie was set correctly across pages)

Follows https://www.playframework.com/documentation/latest/JavaI18N

### Checklist

### Issue(s)
#354 
Screen Shot 2021-03-10 at 4 33 39 PM
Screen Shot 2021-03-10 at 4 34 04 PM


universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -39,7 +43,13 @@ public ApplicantProgramBlocksController(
...
 
               if (block.isPresent()) {
-                return ok(editView.render(request, applicantId, programId, block.get()));
+                return ok(
+                    editView.render(
+                        request,
+                        messagesApi.preferred(request),
+                        applicantId,
+                        programId,
+                        block.get()));
```
bion, line 52: Suggestion (non-blocking): at five params I wonder if the view class should have an inner `public static Params` class that's an autovalue holding all of these. Lengthy param lists are unpleasant and having two `long` param types next to each other makes me nervous since it'd be easy to pass them in swapped.

universal-application-tool-0.0.1/conf/messages.en
```diff
@@ -0,0 +1,3 @@
+apply=Apply
+programs=Programs
+nextBlock=Save and continue
```
bion, line 3: Question (blocking): how might we organize these files so they don't become a mess? Is it possible to put comments in these files? Could we add sections for different parts of the UI e.g. a COMPONENTS section with the different button texts, a PAGES section with sub-sections for different pages?

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramBlocksControllerTest.java
```diff
@@ -2,17 +2,29 @@
...
+import static play.test.Helpers.contentAsString;
 import static play.test.Helpers.fakeRequest;
+import static play.test.Helpers.stubMessagesApi;
 
+import java.util.Collections;
+import java.util.Map;
 import models.Applicant;
 import org.junit.Before;
 import org.junit.Test;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
bion, line 17:
```suggestion
import com.google.common.collect.ImmutableMap;
```

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
```diff
@@ -1,31 +1,44 @@
...
+import static play.test.Helpers.stubMessagesApi;
 
+import java.util.Collections;
+import java.util.Map;
 import models.Applicant;
 import models.Program;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
bion, line 18:
```suggestion
import com.google.common.collect.ImmutableMap;
```

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
```diff
@@ -49,13 +62,40 @@ public void index_withPrograms_returnsAllPrograms() {
...
   }
 
+  @Test
+  public void index_usesMessagesForUserPreferredLocale() {
+    // Stub the MessagesApi with fake translations.
+    Langs langs = new Langs(new DefaultLangs());
+    Map<String, String> translations = Collections.singletonMap("programs", "Different");
+    Map<String, Map<String, String>> messageMap =
+        ImmutableMap.of(Lang.defaultLang().code(), translations);
+    MessagesApi messagesApi = stubMessagesApi(messageMap, langs);
```
bion, line 79: Issue (blocking): this is a decent amount of setup and probably more than we want to do for each controller test once we've localized the whole app. How might we reduce the level of effort for testing controllers that render localized views?

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: Yikes, completely missed that - it keeps importing the wrong one. Thanks for catching!

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramBlocksControllerTest.java
carolinedanzi: Thanks!

universal-application-tool-0.0.1/conf/messages.en
carolinedanzi: Great point! I added comments to `messages.en` - this is the default and the one we should keep the most organized. Thoughts?

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
carolinedanzi: Interesting! I had never thought to have a parameters class, I'll look into it

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: I added a `LocalizationHelper` to the controller test directory for now, but we may want to update `WithPostgresContainer` or some other test class with the override.

Alternatively, it occurs to me we can simply set a different language on the request (using a simple `stubMessageApi()`) and then use the actual translations - which would you prefer?

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
carolinedanzi: I added a Params class that has only a builder (to avoid the ambiguous `create` variable ordering). Errorprone complained that `Block` is final and doesn't override `toString` so I added a very dumb toString method - suggestions for a better one?

universal-application-tool-0.0.1/conf/messages.en
bion: Looks great!

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
bion: The latter sounds really good -- then we'd be testing more of the real codepath.

universal-application-tool-0.0.1/app/services/applicant/Block.java
```diff
@@ -62,4 +62,9 @@ public boolean hasErrors() {
     }
     return errorsMemo.get();
   }
+
+  @Override
+  public String toString() {
+    return "Id: " + this.id;
```
bion, line 68: suggestion (blocking): include the class name here e.g. `'Block [id=" + this.id + "]"`

bion (approved):

universal-application-tool-0.0.1/app/services/applicant/Block.java
carolinedanzi: Much better, thank you!

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: Done! Much simpler

Issue #111, opened 2021-02-10
Admin can remove a question from a block using block dashboard


bion:
blocked on #110 

**Describe the solution you'd like**
In block dashboard center panel is current block with list of questions in the block. Each question has a checkbox next to it. Single "Remove from Block" button removes the checked questions from the block.

See https://tiny.cc/uat-flow for lightweight UI mock.

**Done when**
Admin can remove a question from a block using block dashboard

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354358891906&cot=14

lianne-oy:
M1 Review - In progress

Issue #110, opened 2021-02-10
Admin can add questions to a block using block dashboard


bion:
blocked by #109 and #107 

**Describe the solution you'd like**
Right-side panel in block dashboard has question bank. Each question not already in the program has a checkbox next to it. Single "Add to Block" button submits a form that persists the questions in the current block.

See https://tiny.cc/uat-flow for lightweight UI mock

**Done when**
Admin can add questions to a block using block dashboard

gcapiel:
wireframes: https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354358571331&cot=14

lianne-oy:
M1 Review - In progress

Issue #88, opened 2021-02-10
Admin can group questions into blocks in the Form Builder


lianne-oy:

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354358236952&cot=14

neel-a-mehta:
#108, #109, #110, #111 are all connected to this.

lianne-oy:
M1 Review - In progress

Issue #83, opened 2021-02-10
Admin can define form with single question (using Text widget)


lianne-oy:
- Does not need to be with UI; can be hardcoded for now

neel-a-mehta:
#95 includes this.

gcapiel:
wireframes : https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354355841620&cot=14

lianne-oy:
M1 Review - In progress

PR #330, opened 2021-03-02
Controllers and views for adding/removing questions to blocks

bion:
### Description
Adds `ProgramService` methods `addQuestionsToBlock` and `removeQuestionsFromBlock.

Adds `DuplicateProgramQuestionException` to be thrown when `addQuestionsToBlock` attempts to add a question that is already in the program.

Adds `AdminProgramBlockQuestionsController` for POST methods to add and remove questions to a block.

Adds forms to `ProgramBlockEditView` to add and remove questions to a block. 

### Checklist

### Issue(s)
[Fixes] #110 
[Fixes] #111 
[Fixes] #83 
[Fixes] #88 
[Fixes] #139

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockQuestionsController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockQuestionsController.java
@@ -11,7 +11,10 @@ import play.data.FormFactory;
 import play.mvc.Controller;
 import play.mvc.Http.Request;
 import play.mvc.Result;
+import services.program.ProgramBlockNotFoundException;
+import services.program.ProgramNotFoundException;
 import services.program.ProgramService;
+import services.question.QuestionNotFoundException;
 
 public class AdminProgramBlockQuestionsController extends Controller {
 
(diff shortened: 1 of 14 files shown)


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockQuestionsController.java
```diff
@@ -0,0 +1,80 @@
...
+      return notFound(String.format("Program ID %d not found.", programId));
+    } catch (ProgramBlockNotFoundException e) {
+      return notFound(String.format("Block ID %d not found for Program %d", blockId, programId));
+    } catch (QuestionNotFoundException e) {
+      return notFound(String.format("Question ID %s not found", questionIds));
+    }
+
+    return redirect(controllers.admin.routes.AdminProgramBlocksController.edit(programId, blockId));
+  }
+}
```
bion, line 80: Thought (non-blocking): There's a lot of structural duplication between these two methods, and we could probably figure out a way to extract some of it. That being said, I don't think it's worth it since we don't know that there will be a third piece of duplication and the generic code would itself probably be fairly complicated and verbose.

universal-application-tool-0.0.1/app/services/program/ProgramBlockNotFoundException.java
```diff
@@ -1,7 +1,7 @@
 package services.program;
 
 public class ProgramBlockNotFoundException extends Exception {
-  public ProgramBlockNotFoundException(long programid, long blockid) {
-    super("Block not found in Program (ID " + programid + ") for block ID " + blockid);
+  public ProgramBlockNotFoundException(long programId, long blockId) {
```
AWBaumann, line 4: Nit (non-blocking): For consistency let's use `String.format`

```
super(String.format("Block not found in Program (ID %1$d) for block ID %2$d", programId, blockId));
```

AWBaumann (approved):

PR #334, opened 2021-03-02
Update program list view

AWBaumann:
### Description
- Added some tailwind styles.
- Added `StyleUtils` class
- Added some utility methods to `BlockDefinition` and `ProgramDefinition`
- TODO: Add Tailwind line-clamp plugin.
- TODO: Bind [Draft|Live] and [Last Update] to real data.
 
### Checklist

image


universal-application-tool-0.0.1/app/views/StyleUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+  public static String RESPONSIVE_SM = "sm";
+  public static String RESPONSIVE_MD = "md";
+  public static String RESPONSIVE_LG = "lg";
+  public static String RESPONSIVE_XL = "xl";
+  public static String RESPONSIVE_2XL = "2xl";
+
+  public static String applyUtilityClass(String utility, String style) {
+    return utility + ":" + style;
+  }
```
bion, line 18: question (non-blocking): thoughts on methods for the utilities? e.g. 

```java
public static String focus(String style) {
  return applyUtilityClass(FOCUS, style);
}
```

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -44,6 +44,19 @@ public QuestionDefinition getQuestionDefinition(int blockIndex, int questionInde
...
+    return blockDefinitions().size();
+  }
+
+  public int getQuestionCount() {
+    int total = 0;
+    for (BlockDefinition block : blockDefinitions()) {
+      total += block.getQuestionCount();
+    }
+    return total;
+  }
```
bion, line 58: suggestion (non-blocking): totally optional, but could instead do

```java
public int getQuestionCount() {
  return blockDefinitions().stream().mapToInt(BockDefinition::getQuestionCount).sum();
}
```

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -25,30 +27,96 @@ public ProgramIndexView(AdminLayout layout) {
...
+            Styles.PX_3,
+            Styles.MY_2,
+            Styles.ROUNDED_MD,
+            Styles.RING_BLUE_200,
+            Styles.RING_OFFSET_2,
+            Styles.BG_BLUE_400,
+            Styles.TEXT_WHITE,
+            StyleUtils.applyUtilityClass(StyleUtils.HOVER, Styles.BG_BLUE_500),
+            StyleUtils.applyUtilityClass(
+                StyleUtils.FOCUS, ImmutableList.of(Styles.OUTLINE_NONE, Styles.RING_2)));
```
bion, line 58: Question (non-blocking): would you be up for extracting these styles to a method for creating buttons with these styles so we could start re-using?

bion (approved):

universal-application-tool-0.0.1/app/views/StyleUtils.java
AWBaumann: I'm all about making code shorter where possible and

 `StyleUtils.focus(style)`
  is both shorter and more readable than `StyleUtils.applyUtilityClass(StyleUtils.FOCUS, style)`
  
For now I'll add `focus` and `hover`. I'm not sure that it makes sense for the responsive utils so I'll keep the current methods around for those and any shortcuts we may want later.

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
natsid: In addition to (rather than instead of) what Bion mentioned, another thought I had was to create other constants that are lists of classes that are useful, e.g., `CREATE_BUTTON_CLASSES`, `SUBMIT_BUTTON_CLASSES`, `EDIT_LINK_CLASSES`, etc.

I sort of attempted to do this earlier, but found it didn't play that nicely with the variable args that `withClasses` takes. Definitely not necessary to do in this PR, but just musing about potential ways to keep our style code somewhat separate from our template code so that it's easier to scan the template code...

natsid (approved):
This looks amazing!!!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
AWBaumann: Yeah, I think we're going to want to reuse components as much as possible, but this is probably a good example of normalizing when it makes sense. I don't want to start pulling styles out of `ClassA` into a utility until we decide we also need it (imminently for something like buttons) in `ClassB`. Otherwise we'll get out of control in our Util or Base classes and we'll end up with single use or orphaned methods.

PR #337, opened 2021-03-03
Implement ApplicantData methods for reading and writing JSON

carolinedanzi:
### Description
1. Adds a `Path` class for holding a standard representation of JSON paths, without prefixes
2. Implement read and put methods for `Path`s

### Checklist

### Issue(s)
[Fixes] #<issue_number>


universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -2,55 +2,98 @@
...
+   * structure along the way, i.e., creates parent objects where necessary.
    *
-   * <p>The design suggestion for these methods is to have one per {@link
-   * services.question.ScalarType}
+   * @param path the {@link Path} with the fully specified path, e.g., "applicant.favorites.color"
+   *     or the equivalent "$.applicant.favorite.colors".
+   * @param value the value to place; values of type Map will create the equivalent JSON structure
    */
-  public Optional<String> readString(String path) {
+  public void put(Path path, Object value) {
```
bion, line 46: Suggestion (blocking): please make this method private and instead expose public methods with more specific typing for the value to be written than `Object` e.g. `writeString(Path path, String string)` and `writeDate(Path path, Date date)` . This will help prevent bugs where the wrong type could be written to a given path.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -2,55 +2,98 @@
...
+  /**
+   * Returns the value at the given path, if it exists; otherwise returns {@link Optional#empty}.
+   *
+   * @param path the {@link Path} for the desired scalar
+   * @param type the expected type of the scalar
+   * @param <T> the expected type of the scalar
+   * @return optionally returns the value at the path if it exists, or empty if not
+   * @throws JsonPathTypeMismatchException if the scalar at that path is not the expected type
+   */
+  public <T> Optional<T> read(Path path, Class<T> type) throws JsonPathTypeMismatchException {
```
bion, line 73: Suggestion (blocking): similarly, please make this method private and instead expose public methods with more specific typing for the value to be read. It will allow us to customize how types are serialized for types that aren't supported by JSON.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
carolinedanzi: Thanks for calling out why this matters! I added methods for String, Integer, and Objects (aka Maps) for now

I would love to use `Path` in Questions too - would that be an appropriate refactor?

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
carolinedanzi: I added only readString for now and would prefer to add others when we have question types that require values other than String

bion (approved):

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
bion: Sounds good!

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
bion: I think it would, but maybe it can be done as a follow-on PR? Up to you.

PR #339, opened 2021-03-03
Landing page redirects

bion:
Added redirect logic to the landing page as follows:

- if you're not logged in, redirect to the login form
- if you're logged in as an admin, redirect to the admin programs index
- if you're not logged in as an admin, redirect to the applicant programs index

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/UatProfile.java
+++ b/universal-application-tool-0.0.1/app/auth/UatProfile.java
@@ -64,6 +64,10 @@ public class UatProfile {
     return profileData.getRoles();
   }
 
+  public boolean isUatAdmin() {
+    return profileData.getRoles().contains(Roles.ROLE_UAT_ADMIN.toString());
+  }
+
   public String getId() {
     return profileData.getId();
   }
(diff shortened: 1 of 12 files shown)


universal-application-tool-0.0.1/app/auth/UatProfile.java
```diff
@@ -64,6 +64,10 @@ public String getClientName() {
     return profileData.getRoles();
   }
 
+  public boolean isUatAdmin() {
+    return profileData.getRoles().contains(Roles.ROLE_UAT_ADMIN.toString());
```
hmhsu, line 68: question (if-minor): Can this be an exact match instead of a contain check?

hmhsu (approved):

universal-application-tool-0.0.1/app/auth/UatProfile.java
bion: `getRoles()` returns a set of strings, so I don't think so

hmhsu: Ah cool. Thought the return type is a single string. Feel free to ignore this.

PR #347, opened 2021-03-03
Build the dev container on each merge, and pull it with run-dev.

ndmckinley:
### Description
This will mean we don't have to broadcast "make sure to update your container" every time it's needed.

### Checklist

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -14,6 +14,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Run build-dev.
+        id: build_dev
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.SECRET_ACCESS_KEY }}
+        run: bin/build-dev
+
       - name: Run build-and-push.
         if: "!contains(toJSON(steps.file_changes.outputs.files), 'infra/')"
         id: build_and_push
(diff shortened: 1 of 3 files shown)


bin/build-dev
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+
+set -e
+set +x
+export AWS_DEFAULT_REGION=us-west-2
+REGION=us-west-2
+
+aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/t1q6b4h2
+docker build -t uat-dev --cache-from public.ecr.aws/t1q6b4h2/universal-application-tool:latest .
```
hmhsu, line 9: Question (blocking): Does this need to be `public.ecr.aws/t1q6b4h2/uat-dev`?

.github/workflows/push.yaml
```diff
@@ -14,6 +14,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Run build-dev.
+        id: build_dev
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.SECRET_ACCESS_KEY }}
```
hmhsu, line 21: Question (non-blocking): Do we need to login to do the push?

hmhsu (approved):
Looks good. Thanks for doing this.

bin/build-dev
ndmckinley: Either would work but your suggestion will work better if / as the two diverge!  Changed.

.github/workflows/push.yaml
ndmckinley: Yep.

PR #351, opened 2021-03-04
Program always has one block.

kendrickt:
### Description

A Program now must always have one block. That means a new `Program` includes an empty block, and the `ProgramService` prevents the last block from being delete. 

This allows for:
* `POST /admin/program/:pid/blocks` to create new blocks, which used to be a `GET` method.
* The `Add Block` button is now a form that uses the new POST method.

Due to a Program now coming with an empty Block, the following changes were made:

* `DatabaseSeedController` now updates and adds questions to the default empty block, and then adds block 2.
* `ProgramService` throws a `ProgramNeedsABlockException` when trying to delete the last block.
* Tests had to be changed to account for the new default empty block.

### Checklist

### Approvals
Bion
one_of(natalie or caroline) for the `ApplicantProgramsControllerTest`.


bion (approved):
Definitely an improvement!

PR #370, opened 2021-03-06
Refactor services to use Path instead of String paths

carolinedanzi:
### Description
1) Uses the `Path` class instead of Strings for representing paths
2) Change the `Path` class to use a list for its underlying representation
3) Move `Path` from `services.applicant` to just `services`

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
@@ -19,6 +19,7 @@ import play.mvc.Controller;
 import play.mvc.Http.Request;
 import play.mvc.Result;
 import services.ErrorAnd;
+import services.Path;
 import services.question.InvalidPathException;
 import services.question.InvalidUpdateException;
 import services.question.QuestionDefinition;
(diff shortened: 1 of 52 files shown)


universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -87,7 +88,8 @@ public QuestionController(
         .thenApplyAsync(
             readOnlyService -> {
               try {
-                QuestionDefinition definition = readOnlyService.getQuestionDefinition(path);
+                QuestionDefinition definition =
+                    readOnlyService.getQuestionDefinition(Path.create(path));
```
bion, line 92: suggestion (non-blocking): optional, but you could instead add

```java
QuestionDefinition getQuestionDefinition(String stringPath) {
  return getQuestionDefinition(Path.create(stringPath));
}
```

bion (approved):

AWBaumann (approved):

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
carolinedanzi: Done! Great suggestion

PR #371, opened 2021-03-06
Turn off sbt scala doc link warnings

hmhsu:
### Description
sbt compiles doc with scaladoc and does not properly render javadoc links when referring to classes outside the package. Suppress the warning as we cannot do anything about it.

### Checklist

### Issue(s)
[Fixes] #<issue_number>


bion (approved):

hmhsu:
Turns out sbt compiles the doc into scaladoc and for some reason the javadoc linkage is broken. I cannot find an easy way to fix the linkings so let's turn off the warnings (for now).

PR #374, opened 2021-03-07
Stays on the edit page when question create or update fails

hmhsu:
### Description
Shows error message when a question create or update fails and stays on the edit page with user input fields populated.

### Checklist

### Issue(s)
[Fixes] #<issue_number>


universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -69,29 +70,40 @@ public QuestionController(
                   for (QuestionServiceError e : result.getErrors()) {
                     messageJoiner.add(e.message());
                   }
-                  errorMessage = messageJoiner.toString();
+                  String errorMessage = messageJoiner.toString();
+                  return withData(
+                      withMessage(redirect(routes.QuestionController.newOne()), errorMessage),
+                      questionForm.getData());
```
bion, line 76: Suggestion (blocking): When creation fails due validation failures, re-render the form instead of redirecting. You won't need to store the form data or validation messages in the flash this way.

Generally speaking, the flash should be used for simple string messages.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -66,6 +85,8 @@ private ContainerTag buildEditQuestionForm(QuestionDefinition definition) {
   private ContainerTag buildQuestionForm(QuestionForm questionForm) {
     ContainerTag formTag = form().withMethod("POST");
     formTag
+        .with(hiddenInputWithValue("questionId", questionForm.getQuestionIdString()))
```
bion, line 88: Question (blocking): the question ID is available via the URL path and is passed to the `update` method in the controller, it shouldn't need to also be included in the form contents

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -66,6 +85,8 @@ private ContainerTag buildEditQuestionForm(QuestionDefinition definition) {
...
+        .with(hiddenInputWithValue("questionVersion", questionForm.getQuestionVersionString()))
```
bion, line 89: Suggestion (blocking): recommend removing this because the version isn't necessary for updating the question

bion (approved):
Two comments then ready to merge!

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
hmhsu: Done.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
hmhsu: Done.

hmhsu:
> Two comments then ready to merge!

Please take another look and merge if appropriate.

bion (approved):

PR #386, opened 2021-03-09
Refactored inputs, textareas, and selects into utility classes and added some basic styling.

AWBaumann:
### Description
First pull request in the pipeline for updating UX. Mainly cleanup with a bit of styling/consistent component creation.

Sitcom Episode title: _Aaron Versus the Verbosity of Java_

### Screenshots
image

image

image


universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -0,0 +1,100 @@
...
+public class FieldWithLabel {
+  public static String[] CORE_FIELD_CLASSES = {
+    BaseStyles.FIELD_BACKGROUND_COLOR,
+    BaseStyles.FIELD_BORDER_COLOR,
+    Styles.BLOCK,
+    Styles.BORDER,
+    Styles.P_2,
+    Styles.W_FULL
+  };
+  public static String[] CORE_LABEL_CLASSES = {
```
bion, line 23: Suggestion (blocking): make `CORE_FIELD_CLASSES` and `CORE_LABEL_CLASSES` `private static final`.

AWBaumann: Done.

universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
```diff
@@ -0,0 +1,59 @@
...
+public class SelectWithLabel extends FieldWithLabel {
+
+  private ImmutableList<SimpleEntry<String, String>> options = ImmutableList.of();
+
+  public SelectWithLabel(String inputId) {
+    super(select(), inputId);
+  }
+
+  public SelectWithLabel setOptions(ImmutableList<SimpleEntry<String, String>> options) {
+    if (this.isRendered) {
```
bion, line 21: Question (blocking): should this throw instead of silently being a no-op?

universal-application-tool-0.0.1/test/app/BaseBrowserTest.java
```diff
@@ -165,4 +167,30 @@ protected void removeQuestionsToProgram(String programName, String... questions)
...
+  }
+
+  protected String getInputValue(String name) {
+    return browser.$("input", withName(name)).first().value();
+  }
+
+  protected void fillTextArea(String name, String text) {
+    WebElement element =
+        browser.getDriver().findElement(By.cssSelector("textarea[name=" + name + "]"));
+    element.clear();
```
natsid, line 182: question (non-blocking): Why is `fillTextArea` implemented differently from `fillInput`? Given their names I'd expect them to behave the same: either both clear before filling in, or both not.

universal-application-tool-0.0.1/test/app/BaseBrowserTest.java
```diff
@@ -165,4 +167,30 @@ protected void removeQuestionsToProgram(String programName, String... questions)
...
+    element.sendKeys(text);
+  }
+
+  protected String getTextAreaString(String name) {
```
natsid, line 186: nit: rename to `getTextAreaValue` to match input?

natsid (commented):
Mid-review, but so far, it looks great!

universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
AWBaumann: Discussed with Bion offline. Changing the code a little so that we can create similar Fields from the same "builder."

universal-application-tool-0.0.1/test/app/BaseBrowserTest.java
AWBaumann: I fought this a bunch yesterday. Basically `textarea` gets and sets the value differently from `input`. For input it's an `attribute` but for textarea it's a `property` so it does not play nicely with Selenium. I didn't feel like I was encountering issues with the fill portion of textarea, but was running into a ton of issues trying to pull the value back out. A bunch of searching suggested that this was the proper way to fill the textarea. It may not be necessary, and it didn't solve the problem of not being able to get the value, but I kept it this way since it does seem cleaner.

**TLDR:** I wrote it this way because I was trying to solve a related issue. It didn't solve the related issue but I kept it this way anyway.

universal-application-tool-0.0.1/test/app/BaseBrowserTest.java
AWBaumann: I wish! This doesn't get the value of the text area but returns the rendered html string, so I wanted to keep them different.

natsid (approved):

PR #402, opened 2021-03-10
Remove unused json annotations, and add in other places for clarity

hmhsu:
### Description

### Checklist

### Issue(s)
[Fixes] #<issue_number>


kendrickt (approved):
I wonder what changed. Some of these `get` methods were causing serialization problems without the annotations.

hmhsu:
> I wonder what changed. Some of these `get` methods were causing serialization problems without the annotations.

`ProgramDefinition` class is not serialized into JSON so these should not be needed. Maybe it was in the past?

PR #460, opened 2021-03-11
Add test that we actually save the user's real email address.

ndmckinley:
### Description
This just configured the fake oidc provider to return an email address so we can prove we save it.

### Checklist


ndmckinley:
It's not really javascript, just config.  Thanks!

PR #469, opened 2021-03-12
Extract locale and name from IDCS.

### LINKED ISSUE #384: Extract profile information from IDCS and ADFS
When a user authenticates to IDCS / ADFS, we should extract as much profile information as is useful to prevent them from having to answer questions related to their, e.g. name, email address, phone number.  Investigate what information is available and useful and extract it.  This will be done in `auth.AdfsProfileAdapter` and `auth.IdcsProfileAdapter`.

ndmckinley:
### Description
Extract name and locale from IDCS, if present.  Extract nothing additional from AD, since there's nothing else to extract without further scopes.

### Checklist

### Issue(s)
Fixes #384 - turns out there was less to get than we thought!

(note that the base here is a to-be-submitted PR.)


universal-application-tool-0.0.1/app/auth/IdcsProfileAdapter.java
```diff
@@ -28,6 +28,40 @@ protected String emailAttributeName() {
     return "user_emailid";
   }
 
+  @Override
+  public UatProfileData mergeUatProfile(UatProfile uatProfile, OidcProfile oidcProfile) {
+    String locale = oidcProfile.getAttribute("user_locale", String.class);
```
carolinedanzi, line 33: question: do you know what locale format these use? I'm hoping BCP 47, which is what Java Locales use

universal-application-tool-0.0.1/app/auth/IdcsProfileAdapter.java
```diff
@@ -28,6 +28,40 @@ protected String emailAttributeName() {
...
+              .thenApplyAsync(
+                  applicant -> {
+                    applicant.getApplicantData().setUserName(displayName);
+                    applicant.save();
+                    return null;
+                  }));
+    }
+    for (CompletionStage<Void> dbOp : dbOperations) {
+      dbOp.toCompletableFuture().join();
+    }
```
carolinedanzi, line 60: suggestion (non-blocking): If we don't care about being async here, we could complete the futures in the if statements instead

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -191,4 +195,36 @@ public int hashCode() {
...
+      case 3:
+        firstName = listSplit.get(0);
+        middleName = listSplit.get(1);
+        lastName = listSplit.get(2);
+        break;
+      case 1:
+        // fallthrough
+      default:
+        // Too many names - put them all in first name.
+        firstName = displayName;
```
carolinedanzi, line 218: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ :) My manager sent me this once and now I'm nervous whenever I need to write name logic

question: Do we need to split by first/middle/last, or could we just set their entire user/display name as their username?

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -191,4 +195,36 @@ public int hashCode() {
...
+    }
+    setUserName(firstName, middleName, lastName);
+  }
+
+  private void setUserName(
+      String firstName, @Nullable String middleName, @Nullable String lastName) {
+    // We don't have a question for this yet, so this is unimplemented right now.
```
carolinedanzi, line 225: question: If we're setting this on login, should we save it to the db? i.e. create a field on Applicant

carolinedanzi (commented):
Exciting to see this all coming together!

universal-application-tool-0.0.1/app/auth/IdcsProfileAdapter.java
ndmckinley: All I can say for sure is that the value that comes back for me is `en`, which is the right BCP 47 answer.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
ndmckinley: Yeah, I thought of that article while writing this.  In this case I am not omitting or disqualifying any names.  I am just doing some heuristic guessing - I don't think anyone will be insulted to have their display name split into first / last, or first/middle/last, by spaces.  This is inaccurate for some people - me, for example.  My legal name is `Nathan Douglas McKinley Henderson` - this code will give up and shove that into "first name".  But users like me will have the opportunity to change it if it's wrong, and it'll be right most of the time, so I think it's worth doing.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
ndmckinley: I don't really think so - I think it's more appropriate to have it be stored in applicant data once we have a "what is your name" question hardcoded in there.  You know more about how we're planning to store answers than me, though - is my theory about storing it in applicant data plausible?

universal-application-tool-0.0.1/app/auth/IdcsProfileAdapter.java
ndmckinley: This is *marginally* more parallel in that the first request can be completed while we're sending the second.  Barely worth it, but if we wind up with a ton of these for one of the identity providers, it might be a good pattern.

carolinedanzi: sounds good to me!

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
carolinedanzi: I think based on our discussion at standup Friday this is possible. Happy to go with this for now!

carolinedanzi (approved):

Issue #390, opened 2021-03-09
"Text" question type: implement simple validation criteria


neel-a-mehta:
Main issue for the Text question type is #388, so see that first.

This includes:
* Min length (in characters)
* Max length (in characters)

hmhsu:
This should be done now.

natsid:
Are you sure this is complete @hmhsu ? Before I went OOO last week I was still working on displaying the admin-configurable text validations to the applicant and did not complete it.

natsid:
Ohh, I just noticed #544 which I think covers it.

PR #497, opened 2021-03-13
Add validation predicates column to questions table

hmhsu:
### Description
- Add a `validation_predicates` column to `questions` table
- Store `validationPredicates` as a string in `QuestionDefinition`
- Add `ValidationPredicates` class to hold question-specific predicates with serialization and deserialization logic
- Add constructors to allow questions to have empty predicates
- Update `QuestionDefinitionBuilder`, `QuestionDefinition#equals` and tests

### Checklist

### Issue(s)
#390


bion:
I'm curious what the plan is for how this will support different types of predicates for different question types. Another option to consider would be to use object mapping with per-question type classes rather than a `Map`

hmhsu:
> I'm curious what the plan is for how this will support different types of predicates for different question types. Another option to consider would be to use object mapping with per-question type classes rather than a `Map`

We can certainly do that. We were thinking to have a free form string to hold the json and then I figured a map is better than a string. I am thinking having a single level map with all possible predicates is not too bad as long as we have clear naming. This eliminates the need for us to parse the json structure. What do you think?

hmhsu:
> I'm curious what the plan is for how this will support different types of predicates for different question types. Another option to consider would be to use object mapping with per-question type classes rather than a `Map`

Chatted offline. Will try out the Jackson inheritance feature!

bion (approved):

PR #510, opened 2021-03-15
Prevent path conflict on prefix

### LINKED ISSUE #476: Improve path conflicting validation logic
All data collected from applications in CiviForm is stored using a JSON path defined by the `QuestionDefinition` for the data. When new questions are created, we validate the path to ensure that it does not conflict with an existing `QuestionDefinition`'s path.

Right now path `path.one` would conflict with `path.onesomething`, but it shouldn't. Path validation logic should be able to distinguish path by key names rather than just prefix.

The validation is in [`QuestionRepository.java`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/repository/QuestionRepository.java#L36).

**Done when**
Path validation logic does not consider `path.one` and `path.onething` is conflicting.

Snackeroo:
### Description
Appended delimiter to distinguish paths by key names

### Checklist

### Issue(s)
Fixes #476

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/QuestionRepository.java
@@ -55,8 +55,8 @@ public class QuestionRepository {
     }
 
     static boolean pathConflicts(String path, String otherPath) {
-      path = path.toLowerCase();
-      otherPath = otherPath.toLowerCase();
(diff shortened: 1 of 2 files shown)


AWBaumann (approved):
Thanks!

hmhsu (approved):

PR #511, opened 2021-03-15
Standardize Element id formats

AWBaumann:
### Description
Minor refactor to add/standardize the use of ids in html elements.

### Checklist


universal-application-tool-0.0.1/app/views/J2HtmlDemoView.java
```diff
@@ -35,7 +35,8 @@ public Content render(String greeting, Request request) {
             form(
                     makeCsrfTokenInputTag(request),
                     FieldWithLabel.input()
-                        .setId("firstName")
+                        .setId("first-name-input")
```
natsid, line 38: suggestion (non-blocking): I tend to prefer classes to IDs in general. In this case specifically, I can imagine a scenario where we have two first name inputs on the same page (e.g., "you and your spouse's names).

Thoughts on using class instead of id throughout?

I realized after writing the above how annoying it might be given the way that J2Html overrides classes on each call to `withClass(es)` though 😕  So feel free to ignore if that hurdle seems too big.

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -90,15 +97,16 @@ public ContainerTag getContainer() {
     } else {
       fieldTag.withValue(this.fieldValue);
     }
-    fieldTag.withId(this.id).withName(id).withPlaceholder(this.placeholderText);
 
-    if (formId.length() > 0) {
-      fieldTag.attr("form", formId);
-    }
+    fieldTag
+        .withCondId(!Strings.isNullOrEmpty(this.id), this.id)
```
natsid, line 102: praise: Oh! This is great. We can use this `withCond` pattern here for `value` if/when we want to be able to pre-set a value in this component. We'll want that for applicant view inputs.

natsid (approved):
Thank you!! I'm psyched to have these more consistent - especially for testing!

PR #513, opened 2021-03-15
Store applications in the database as a snapshot, display the ID of the snapshot

### LINKED ISSUE #205: When applicant hits Submit, snapshot of their answers stored in database
Doesn't have to be exportable yet, but just make sure it gets saved.

The snapshot can then be rendered in PDF, CSV, or HTML in the frontend.

This is a static snapshot timestamped with the current time, so if the applicant's answer(s) change(s) later, the snapshot won't change at all.

Prerequisite for #191.

### LINKED ISSUE #209: Applicant gets Program Application ID Number when they submit program form
Dynamically generate an "application ID number" when the applicant hits submit. Show it on the "thank you" page once their answers are saved to the database.

ndmckinley:
### Description
Creates a new table - applications - which contains snapshots of a user's entire applicant object at the time that an application is submitted.  It links to a program and an applicant.

Moved some of the async logic around since now a request to the "next block" button might create this new DB object.  On that note, I'm sorry that this diff is borderline incomprehensible, I have _not_ figured out how to make java async code look good.  Suggestions welcome...

### Checklist

### Issue(s)
Fixes #205 
Fixed #209

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
+++ b/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
@@ -1,6 +1,8 @@
 package controllers.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.concurrent.CompletableFuture.failedFuture;
+import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -9,15 +11,18 @@ import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
+import models.Application;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.data.DynamicForm;
 import play.data.FormFactory;
 import play.i18n.MessagesApi;
 import play.libs.concurrent.HttpExecutionContext;
+import play.mvc.Call;
 import play.mvc.Controller;
 import play.mvc.Http.Request;
 import play.mvc.Result;
+import repository.ApplicationRepository;
 import services.applicant.ApplicantService;
 import services.applicant.Block;
 import services.applicant.ReadOnlyApplicantProgramService;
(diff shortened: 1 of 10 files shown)


universal-application-tool-0.0.1/app/models/Applicant.java
```diff
@@ -63,4 +69,8 @@ public Account getAccount() {
   public void setAccount(Account account) {
     this.account = account;
   }
+
+  public ImmutableList<Application> getApplications() {
```
natsid, line 73: question (non-blocking): Would it make sense to add a test for this? To make sure the mapping is how we expect it.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import static java.util.concurrent.CompletableFuture.supplyAsync;
+
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import javax.inject.Inject;
+import models.Applicant;
+import models.Application;
+import models.Program;
+
+public class ApplicationRepository {
```
natsid, line 12: suggestion (non-blocking): Seems like we could use a test suite for this class, given the relative complexity of `submitApplication` with the IDs.

natsid (approved):

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -118,48 +121,81 @@ public ApplicantProgramBlocksController(
                 Throwable cause = ex.getCause();
                 if (cause instanceof ProgramNotFoundException) {
                   return badRequest(cause.toString());
+                } else if (cause instanceof ProgramBlockNotFoundException) {
+                  logger.error("Exception while updating applicant data", cause);
+                  return badRequest("Unable to process this request");
```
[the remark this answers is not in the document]
ndmckinley: This is within an exceptionally() block, so the code is already executed asynchronously.

universal-application-tool-0.0.1/app/models/Applicant.java
ndmckinley: Definitely, added.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
ndmckinley: I kind of think the browser test covers it - I'm reading "non-blocking" here to mean you're okay with that.

Issue #465, opened 2021-03-11
Use the new "CiviForm" name throughout the codebase/GitHub


neel-a-mehta:
Replace "UAT" with "CiviForm" wherever it appears, both in code and elsewhere in this GitHub repo.

bion:
Blocked on this until after Jake has met with legal

PR #530, opened 2021-03-17
Add human readable names to AWS resources

### LINKED ISSUE #521: Add "name" tag to all AWS resources
From John Doke at Seattle IT:

> please ensure that the staff working in the AWS account give a “name” tag to all the resources they create? I’m sure this can be added to the CF templates. The name tag will add a visible clue when using the console as to what we are looking at. We are just using the standard AWS conventions: https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/other-aws-resource-types.html

image005
image003
image002

hmhsu:
### Description
Make AWS resource names more readable.

### Checklist

### Issue(s)
Fixes #521 
#465

### DIFF
--- a/bin/build-and-push
+++ b/bin/build-and-push
@@ -10,7 +10,7 @@ docker build -f prod.Dockerfile -t universal-application-tool --cache-from publi
 docker tag universal-application-tool:latest public.ecr.aws/t1q6b4h2/universal-application-tool:latest
 docker push public.ecr.aws/t1q6b4h2/universal-application-tool:latest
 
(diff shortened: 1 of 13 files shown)


hmhsu:
This change renames the whole stack, I likely need to delete the old stack and create a new one with the updated name.

ndmckinley (approved):
Definitely no way that'll apply cleanly, yeah.  I'd blow the old stack away, then create this one using `bin/deploy-prod`, then come back here and re-run the failed check.

If you have another plan, go for whatever seems good to you!

Issue #519, opened 2021-03-16
Add metadata for each update to applicant data


carolinedanzi:
**Is your feature request related to a problem? Please describe.**
We cannot accurately generate the `getCurrentBlockList()` without checking which blocks were completed in this session/program. See https://github.com/seattle-uat/civiform/issues/406 and https://github.com/seattle-uat/civiform/pull/399

**Describe the solution you'd like**
When `ApplicantData` is updated, we also need to record metadata for that update, which includes the timestamp and program ID for the context of the update. The metadata should look similar to [this](https://gist.github.com/carolinedanzi/67fdd1419e6b13859985ea6e05407cf3).

The solution here most likely involves adding methods to `ApplicantData` for updating metadata, as well as ensuring each `put` method also updates metadata (perhaps adding a `programId` parameter to those methods)

**Done when**
Each update to applicant data also records that update's metadata

carolinedanzi:
This was completed with #533

PR #533, opened 2021-03-17
Write metadata for every question an applicant answered

carolinedanzi:
### Main Updates and Files Changed
1. Add `updated_at` and `updated_in_program` metadata to questions themselves (i.e. they are scalars)
    1. _QuestionDefinition, NameQuestionDefinition, AddressQuestionDefinition, TextQuestionDefinition_
2. Change `ScalarType.INT` to `ScalarType.LONG` - we will represent all whole numbers as longs
    1. _ScalarType, ApplicantData (methods for int --> long)_
3. For each question path in updates for a block, write two pieces of metadata (updated time and program ID)
    1. _ApplicantServiceImpl + Test_
4. Remove `stageAndUpdateIfValid` method on `ApplicantService` that exposes internal `Update` details, and instead only expose the method which uses `ImmutableMap`
    1. _ApplicantService, ApplicantServiceImpl (reordered methods)_
5. Update `ApplicantData` methods to avoid writing the empty string
6. Update logic in `ApplicantQuestion` to check whether an applicant has answered a question, instead of checking for presence of a value
    1. _ApplicantQuestion + Test_

### Checklist

### Issue(s)
#519

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -4,10 +4,9 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
-import com.jayway.jsonpath.Configuration;
 import com.jayway.jsonpath.DocumentContext;
 import com.jayway.jsonpath.JsonPath;
-import com.jayway.jsonpath.Option;
+import com.jayway.jsonpath.PathNotFoundException;
 import com.jayway.jsonpath.spi.mapper.MappingException;
 import java.time.Instant;
 import java.util.HashMap;
@@ -22,12 +21,7 @@ import org.slf4j.LoggerFactory;
 import services.Path;
 
 public class ApplicantData {
-  // Suppress errors thrown by JsonPath and instead return null if a path does not exist in a JSON
-  // blob.
-  private static final Configuration CONFIGURATION =
-      Configuration.defaultConfiguration().addOptions(Option.SUPPRESS_EXCEPTIONS);
   private static final String EMPTY_APPLICANT_DATA_JSON = "{ \"applicant\": {}, \"metadata\": {} }";
-
   private static final Logger LOG = LoggerFactory.getLogger(ApplicantData.class);
   private static final Locale DEFAULT_LOCALE = Locale.US;
 
@@ -44,7 +38,7 @@ public class ApplicantData {
 
   public ApplicantData(Locale preferredLocale, String jsonData) {
     this.preferredLocale = preferredLocale;
-    this.jsonData = JsonPath.using(CONFIGURATION).parse(checkNotNull(jsonData));
+    this.jsonData = JsonPath.parse(checkNotNull(jsonData));
   }
 
   public Locale preferredLocale() {
@@ -53,6 +47,29 @@ public class ApplicantData {
 
   public void setPreferredLocale(Locale locale) {
     this.preferredLocale = locale;
+  }
+
+  public boolean hasPath(Path path) {
+    try {
+      this.jsonData.read(path.path());
+    } catch (PathNotFoundException e) {
+      return false;
+    }
+    return true;
+  }
+
+  /**
+   * Returns true if there is any value at the given {@link Path}; false otherwise.
+   *
+   * @param path the {@link Path} to check
+   * @return true if there is a non-null value at the given path; false otherwise
+   */
+  public boolean hasValueAtPath(Path path) {
+    try {
+      return read(path, Object.class).isPresent();
+    } catch (JsonPathTypeMismatchException e) {
+      return false;
+    }
   }
 
   public void putString(Path path, String value) {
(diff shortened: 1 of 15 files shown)


universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -23,28 +24,81 @@
...
+          long applicantId, long programId, long blockId, ImmutableMap<String, String> updateMap) {
+    ImmutableSet<Update> updates =
+        updateMap.entrySet().stream()
+            .map(entry -> Update.create(Path.create(entry.getKey()), entry.getValue()))
+            .collect(ImmutableSet.toImmutableSet());
+
+    boolean updatePathsContainReservedKeys =
+        updates.stream().anyMatch(u -> RESERVED_SCALAR_KEYS.contains(u.path().keyName()));
+    if (updatePathsContainReservedKeys) {
+      return CompletableFuture.completedFuture(
```
hmhsu, line 93: Suggestion (blocking): If this is not an expected user input error, please raise an exception instead.

```
return CompletableFuture.failedFuture(new IllegalArgumentException("Path contained reserved scalar key"));
```

Please handle the exception in the controller.

```
.exceptionally(
  ex -> {
    if (ex instanceof CompletionException) {
                Throwable cause = ex.getCause();
                if (cause ...) {
                   ...
                } else if (cause instanceof IllegalArgumentException) {
                  // handle this exception
                  return badRequest("Unable to process this request");
                }
                // ...
     }
  });
```

universal-application-tool-0.0.1/test/services/applicant/ApplicantServiceImplTest.java
```diff
@@ -160,6 +189,23 @@ public void stageAndUpdateIfValid_hasPathNotInBlockException() {
...
+  public void stageAndUpdateIfValid_hasIllegalArgumentExceptionForReservedScalarKeys() {
+    Applicant applicant = subject.createApplicant(1L).toCompletableFuture().join();
+    String reservedScalar = "applicant.name." + QuestionDefinition.METADATA_UPDATE_TIME_KEY;
+    ImmutableMap<String, String> updates = ImmutableMap.of(reservedScalar, "12345");
+
+    ErrorAnd<ReadOnlyApplicantProgramService, Exception> errorAnd =
+        subject
+            .stageAndUpdateIfValid(applicant.id, programDefinition.id(), 1L, updates)
+            .toCompletableFuture()
+            .join();
```
hmhsu, line 202: Suggestion (blocking): Please throw an exception in the method and catch it here. `join()` will throw an unchecked `CompletionException`.

universal-application-tool-0.0.1/app/services/question/QuestionDefinition.java
```diff
@@ -27,6 +27,9 @@
   private final ImmutableMap<Locale, String> questionHelpText;
   private final ValidationPredicates validationPredicates;
 
+  public static final String METADATA_UPDATE_TIME_KEY = "updated_at";
+  public static final String METADATA_UPDATE_PROGRAM_ID_KEY = "updated_in_program";
+
```
hmhsu, line 32: Nit (blocking): Please move static fields above non-static fields.

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -23,28 +24,81 @@
...
 public class ApplicantServiceImpl implements ApplicantService {
 
   private final ApplicantRepository applicantRepository;
   private final ProgramService programService;
+  private final Clock clock;
   private final HttpExecutionContext httpExecutionContext;
+  private static final ImmutableSet<String> RESERVED_SCALAR_KEYS =
+      ImmutableSet.of(
+          QuestionDefinition.METADATA_UPDATE_TIME_KEY,
+          QuestionDefinition.METADATA_UPDATE_PROGRAM_ID_KEY);
```
hmhsu, line 40: Nit(blocking): Please move static fields above non-static fields.

hmhsu (approved):
Looks good.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -54,16 +47,67 @@ public void setPreferredLocale(Locale locale) {
...
+    try {
+      this.jsonData.read(path.path());
+    } catch (PathNotFoundException e) {
+      return false;
+    }
+    return true;
+  }
+
+  /**
+   * Returns true if there is any value at the given {@link Path}; false otherwise.
```
bion, line 68: nit (blocking): "any value" is a little tricky in this context because `null` is a value and it's not clear without reading the code whether or not this can return null. Recommend clarifying that it will return `false` if the value at the path is `null`.

carolinedanzi: Great point, fixed the wording to "non-null"

bion (approved):

universal-application-tool-0.0.1/test/services/applicant/ApplicantServiceImplTest.java
carolinedanzi: Thanks! I believe this is fixed

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
carolinedanzi: Thanks Hung-Ming! I updated to throw instead of return ErrorAnd, mind double-checking my work?

hmhsu: Looks great! Thanks for doing this!

Issue #538, opened 2021-03-18
Browser tests don't reflect reality quite right


ndmckinley:
We have functional browser tests that seem to diverge from actually using the app in chrome.  Let's try swapping the tests to chrome, see if we can repro the errors we're getting in manually testing.  If that doesn't work, then we should probably find a new browser testing solution.

PR #539, opened 2021-03-18
Fix bug (but doesn't update tests because they weren't breaking 👎 )

AWBaumann:
### Description
The bug speaks for itself.


bion (approved):

AWBaumann:
The life of a programmer...

"Oof! A bug! Who wrote that?"
. o O (I wrote that)
"Why didn't you catch that in review?!?"

ndmckinley:
#538 to get the tests to catch this bug...

PR #541, opened 2021-03-18
Handle expected user input errors in program create and update

### LINKED ISSUE #362: Bug: invalid program form submission causes server error
**Describe the bug**
Creating a program with invalid params results in an 500-level server error.

**To Reproduce**
1. Login as admin
2. Click on 'Create new program'
3. Click 'Create' without entering anything into the form fields
4. See server error

**Expected behavior**
The form should render with friendly error messages describing the validation failures the user needs to correct

**Additional Context**
We want to add basic validation logic in [`ProgramServiceImpl.java`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java#L95) and return any errors as `ErrorAnd` like we do in [`QuestionServiceImpl.java`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java#L33). Then, we handle the errors in [`AdminProgramController.java`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java#L58).

Since we are changing the interface, we need to update multiple places where the interface is used. These are mostly test files today.

**Screenshots**
Screen Shot 2021-03-05 at 11 01 30 AM

hmhsu:
### Description
- Create and Update Program Definition return errors if input is invalid
- Errors are handled in admin program controller instead of throwing 5xx server errors
- Added tests

### Checklist

### Issue(s)
Fixes #362


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
```diff
@@ -74,10 +83,23 @@ public Result update(Request request, long id) {
...
     }
   }
+
+  private String joinErrors(ImmutableSet<ProgramServiceError> errors) {
+    StringJoiner messageJoiner = new StringJoiner(". ", "", ".");
+    for (ProgramServiceError e : errors) {
+      messageJoiner.add(e.message());
+    }
+    return messageJoiner.toString();
+  }
```
bion, line 104: Suggestion (blocking): nearly identical code appears in [QuestionController](https://github.com/seattle-uat/civiform/blob/aab99bb53941afbb527ab2f68435e163d4fc2106/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java#L68-L73). Recommend creating `class CiviFormController extends Controller` that implements `protected String joinErrors(...)` and make it a superclass of the controllers we define.

universal-application-tool-0.0.1/app/services/program/ProgramServiceError.java
```diff
@@ -0,0 +1,15 @@
...
+@AutoValue
+public abstract class ProgramServiceError {
+
+  public static ProgramServiceError of(String message) {
+    return new AutoValue_ProgramServiceError(message);
+  }
+
+  /** The failure message. */
+  public abstract String message();
+}
```
bion, line 15: Issue (non-blocking): this seems somewhat overly specific to program service given that it is just wrapping a string. Consider making this `services.CiviFormError`? It'd be nice to get re-use out of the code that consumes it when we can.

bion (approved):

universal-application-tool-0.0.1/app/services/program/ProgramServiceError.java
hmhsu: Done.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
hmhsu: Good idea! Done.

PR #549, opened 2021-03-19
Local s3 - Closes #482

### LINKED ISSUE #482: Set up localstack.cloud so our browser tests can upload files locally
Our local dev environment does not connect to a dev S3.

Integrate local dev environment with [localstack](https://github.com/localstack/localstack) so we have a dev S3 and store files locally.

**Additional Context**
[`AmazonS3Client.java`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/repository/AmazonS3Client.java) implements the logic to connect to a AWS S3 bucket. The client is only active when relevant [config](https://github.com/seattle-uat/civiform/blob/17cf42e64bb440d0ff783b5114cbc1a2bfb57df3/universal-application-tool-0.0.1/conf/application.conf#L387) is set. Local dev environment currently does not set this config.

We want to setup a fake S3 with localstack and point default local config to it.

**Done when**
Done when we have browser tests that verify that our app can store files locally.

syntapy:
### Description
Adds docker-compose file to set up basic localstack, as well as commands to create local s3 bucket

Defines a `DEV_ENV=1` environment var in dev docker-compose.yaml to inform app whether too look for local s3 buck and (attempts to) add java code in to hook up app backend to local s3 bucket

### Checklist
- Note: *I haven't been able to test whether the browser tests actually work though, since I can't seem to get a UI set up on my system. Will have to look at that further*
- [ ] Automate tests to demonstrate that browser tests store files locally across shutdown / startup of localstack
- [ ] Need to update the [contributing guide](https://github.com/seattle-uat/civiform/wiki/Contributing) to include localstack handling scripts
- [x] Make sure localstack handling commands are elegant enough according to dev preferenes

### Issue(s)
Fixes #482

### DIFF
--- a/bin/run-dev
+++ b/bin/run-dev
@@ -4,7 +4,18 @@
 $( dirname ${BASH_SOURCE[0]})/pull-image
 
 if [ -f ".secrets" ]; then
-	. .secrets
+    . .secrets
 fi
 
+# start localstack
+docker-compose up -d localstack
+
+# wait till localstack running
+# otherwise other services crash
+$( dirname ${BASH_SOURCE[0]})/localstack/wait
+
+# Ensure local s3 created
+$( dirname ${BASH_SOURCE[0]})/localstack/mk-s3
+
+# start everything else
 docker-compose up
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,17 @@ version: '3.1'
 
 services:
 
+  localstack:
+    image: localstack/localstack
+    ports:
+      - "4566:4566"
+    environment:
+      - SERVICES=s3
+      - DOCKER_HOST=unix:///var/run/docker.sock
+    volumes:
+      - "${TMPDIR:-/tmp/localstack}:/tmp/localstack"
+      - "/var/run/docker.sock:/var/run/docker.sock"
+
   db:
     image: postgres:12.5
     restart: always
@@ -10,6 +21,8 @@ services:
       - 5432:5432
     environment:
       POSTGRES_PASSWORD: example
+
+    depends_on: [localstack]
 
   civiform:
     image: civiform
@@ -23,8 +36,8 @@ services:
       - AWS_ACCESS_KEY_ID
       - AWS_SECRET_ACCESS_KEY
       - AWS_SESSION_TOKEN
-      - AWS_S3_REGION
-      - AWS_S3_BUCKET_NAME
+      - AWS_S3_REGION=us-west-2
+      - AWS_S3_BUCKET_NAME=civiform-local-s3
       - IDCS_CLIENT_ID
       - IDCS_SECRET
       - ADFS_CLIENT_ID
@@ -35,5 +48,7 @@ services:
       - ~/.coursier:/root/.coursier
       - ~/.sbt:/root/.sbt
       - ~/.ivy:/root/.ivy2
+
+    depends_on: [localstack]
 
     command: ~run
--- a/universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
+++ b/universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
@@ -3,17 +3,21 @@ package repository;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.typesafe.config.Config;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CompletableFuture;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import play.Environment;
 import play.inject.ApplicationLifecycle;
 import software.amazon.awssdk.core.ResponseBytes;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.S3ClientBuilder;
 import software.amazon.awssdk.services.s3.model.GetObjectRequest;
 import software.amazon.awssdk.services.s3.model.GetObjectResponse;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
@@ -23,18 +27,21 @@ import software.amazon.awssdk.services.s3.model.S3Exception;
 public class AmazonS3Client {
   public static final String AWS_S3_REGION = "aws.s3.region";
   public static final String AWS_S3_BUCKET = "aws.s3.bucket";
+  public static final String AWS_LOCAL_ENDPOINT = "aws.local.endpoint";
   private static final Logger log = LoggerFactory.getLogger("s3client");
 
   private final ApplicationLifecycle appLifecycle;
   private final Config config;
+  private final Environment environment;
   private Region region;
   private String bucket;
   private S3Client s3;
 
   @Inject
-  public AmazonS3Client(ApplicationLifecycle appLifecycle, Config config) {
+  public AmazonS3Client(ApplicationLifecycle appLifecycle, Config config, Environment environment) {
     this.appLifecycle = checkNotNull(appLifecycle);
     this.config = checkNotNull(config);
+    this.environment = checkNotNull(environment);
 
     log.info("aws s3 enabled: " + String.valueOf(enabled()));
     if (enabled()) {
@@ -103,9 +110,22 @@ public class AmazonS3Client {
 
   private void connect() {
     String regionName = config.getString(AWS_S3_REGION);
+    String endpoint;
     region = Region.of(regionName);
     bucket = config.getString(AWS_S3_BUCKET);
+    URI endpointOverride;
+    S3ClientBuilder s3ClientBuilder;
 
-    s3 = S3Client.builder().region(region).build();
+    endpoint = config.getString(AWS_LOCAL_ENDPOINT);
+    s3ClientBuilder = S3Client.builder().region(region);
+    if (environment.isDev()) {
+      try {
+        endpointOverride = new URI(endpoint);
+        s3ClientBuilder = s3ClientBuilder.endpointOverride(endpointOverride);
+      } catch (URISyntaxException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    s3 = s3ClientBuilder.region(region).build();
   }
 }
--- a/universal-application-tool-0.0.1/conf/application.conf
+++ b/universal-application-tool-0.0.1/conf/application.conf
@@ -386,3 +386,4 @@ ebean.default = "models.*"
 
 aws.s3.region=${?AWS_S3_REGION}
 aws.s3.bucket=${?AWS_S3_BUCKET_NAME}
+aws.local.endpoint="http://localstack:4566"
--- a/bin/localstack/mk-s3
+++ b/bin/localstack/mk-s3
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+aws --endpoint-url=http://localhost:4566 s3api create-bucket --bucket civiform-local-s3 --region us-west-2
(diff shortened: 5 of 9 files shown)


universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -103,9 +109,27 @@ private void getTestObject() {
...
     String regionName = config.getString(AWS_S3_REGION);
+    String endpoint;
     region = Region.of(regionName);
     bucket = config.getString(AWS_S3_BUCKET);
+    String dev_env = config.getString(DEV_ENV);
+    URI endpointOverride;
+
+    endpoint = config.getString(AWS_LOCAL_ENDPOINT);
+    try {
+        if (dev_env.equals("1")) {
```
hmhsu, line 120: Suggestion (blocking): would you leverage `play.Environment#isDev` instead of having a dedicated config for this?

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -103,9 +109,27 @@ private void getTestObject() {
...
+            endpointOverride = new URI(endpoint);
+            s3 = S3Client.builder().endpointOverride(endpointOverride).region(region).build();
 
-    s3 = S3Client.builder().region(region).build();
+        } else {
+            s3 = S3Client.builder().region(region).build();
+        }
+
+    } catch(URISyntaxException e) {
```
hmhsu, line 128: Suggestion (blocking): please have a narrower try-catch scope. It is okay to throw a RuntimeException here as this is only used in dev mode.

```
S3ClientBuilder s3ClientBuilder = S3Client.builder().region(region);
if (environment.isDev()) {
  try {
    URI localEndpoint = new URI(config.getString(AWS_LOCAL_ENDPOINT));
    s3ClientBuilder = s3ClientBuilder.endpointOverride(localEndpoint);
  } catch (URISyntaxException e) {
  }
}
s3 = s3ClientBuilder.build();
```

bin/run-dev
```diff
@@ -1,17 +1,31 @@
 #!/bin/bash
 
+set -e
+set +x
+export AWS_DEFAULT_REGION=us-west-2
+export REGION=$AWS_DEFAULT_REGION
+export AWS_S3_REGION=$AWS_DEFAULT_REGION
+export AWS_S3_BUCKET_NAME=civiform-local-s3
```
hmhsu, line 8: Suggestion (blocking): let's put this in application.config directly.

bin/localstack/start_localstack
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+docker-compose -f infra/localstack.yaml up -d
+
+./bin/localstack/wait_localstack_running
```
hmhsu, line 5: Suggestion (blocking): Would you inline all the waiting scripts here? I am not sure how much value they provide as separate scripts.

infra/localstack.yaml
```diff
@@ -0,0 +1,18 @@
...
+  localstack:
+    image: localstack/localstack
+    ports:
+      - "53:53"
+      - "443:443"
+      - "4510-4520:4510-4520"
+      - "4566-4620:4566-4620"
+      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"
+    environment:
+      - LOCALSTACK_API_KEY=$LOCALSTACK_API_KEY
```
hmhsu, line 12: Suggestion (blocking): Can we inline the api key here? If it's a secret, please also remove the `.env` file that you accidentally included in the PR.

bin/run-dev
syntapy: I put them in the `docker-compose.yaml`. Is this ok? I wasn't sure how to put them in the `application.config` file

infra/localstack.yaml
syntapy: I have idea how I ended up commiting that! Thanks for pointing it out

.gitignore
```diff
@@ -10,3 +10,6 @@ RUNNING_PID
 .secrets
 node_modules
 bundle.js
+.venv/
```
kendrickt, line 13: Do these need to be added to our gitignore? We aren't using python.

syntapy: Removed it

bin/localstack/wait-local-s3-exists
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo "Please make sure the local s3 bucket exists (run './bin/start-local-s3' from another terminal)"
```
hmhsu, line 3: Question (if-minor): Can we include the start-local-s3 command here? What is the benefit of having a separate terminal?

hmhsu (approved):
Looks good to me (with a few minor comments)! Nice work.

hmhsu:
Forget to ask what are the new steps to start a new dev server (from reading the PR, it looks like it might be changed). I'd like to document it if it will be changed from just running `bin/run-dev`. I think we prefer to have one command to start the server unless it is not possible.

bin/run-dev
hmhsu: The config in `application.config` is overridden in sequential order.

For example, you want

```
aws.s3.region=us-west-2
aws.s3.region=${?AWS_S3_REGION}
```

The config will pick up environment variable `AWS_S3_REGION` if it is set, or use `us-west-2` if not.

hmhsu:
The current state of the PR looks great to me. Is it ready to be merged? Or are you still working on it?

syntapy:
> Forget to ask what are the new steps to start a new dev server (from reading the PR, it looks like it might be changed). I'd like to document it if it will be changed from just running `bin/run-dev`. I think we prefer to have one command to start the server unless it is not possible.

After your suggestion, now 

- `bin/run-dev`: start everything (localstack + local s3 + local civiform server)
    - **only command really needed**
- `bin/stop-dev`: stop just the local civiform server, but not localstack.
    - dev can then run `bin/run-dev` again as needed

2nd `bin/run-dev` will be faster since it doesn't have to wait for localstack to startup

Everything associated with the localstack is in `bin/localstack/`

If desired, the dev can run
 -  `bin/localstack/start-localstack` to only start the local stack. 
     - **called by `bin/run-dev`**
 -  `bin/mk-s3` to only create the local s3
     - **called by `bin/run-dev`**
     - will fail if localstack not started yet
     - will **not** fail if local s3 already exists
 - `bin/localstack/purge-s3` to recursively remove all contents of local s3
     - **only for user to explicitly call; never called by other scripts**
     - I figured this might come in handy to someone for some reason. Who knows...
 - `bin/stop-localstack` to stop localstack, but doesn't try to stop civiform server
     - **Probably best after stopping local civiform server**

Questions: 

1. Do I need to do any modifications (e.g. file paths) for windows systems, or are they running linux shell emulators or whatnot?
2. Do you want me to automate any tests to make sure the setup / teardown of the local infrastructure works?
    - e.g. adding / removing from s3 bucket, different combinations of these commands 
3. How to verify that the civiform server actually interacts with the local s3?

-----------


*And the honor and distinction of the individual consist in this, that he among all the world’s multitudes should become a source of social good. Is any larger bounty conceivable than this, that an individual, looking within himself, should find that by the confirming grace of God he has become the cause of peace and well-being, of happiness and advantage to his fellow men?*

*- Abdul-Baha*

hmhsu:
Hey, sorry for the late reply. Thanks for the detailed explanation, those are very helpful.

I just tried this locally and found a few changes needed:

- I think it makes sense to combine `localstack.yaml` with our current `docker-compose.yml` so we can start everything together. This is necessary for the server to connect to the local s3 due to docker network setting. (otherwise it can't see localstack at all)
- This also has the benefit of simplifying the stack a bit. Now we can shut everything down together and won't forget having to run another shutdown command.
- The AWS profile doesn't work for me/ need extra setup. Can we just use default profile?

To answer your questions:
- Windows setup: good question - I don't see anything obvious that would break but we can only find out when we actually run on a windows machine. I am saying let's worry about that later, but if you have a windows box handy to test on then go for it.
- creating/deleting s3 buckets is not a common operation in our normal use case so I'd say not testing it is fine. We will only need one bucket in our current setup.
- glad you asked - the server automatically connects to s3 when we supply the environment variables. So running bin/run-dev and access any page at localhost:9000 will trigger the server to connect to s3 and create a test object named `file1`.

syntapy:
Ok, I will see if I can do it. Will let you know. Thank you

syntapy:
@hmhsu After your suggestion you all only need to worry about `bin/run-dev` and `bin/stop-dev`

Waaay more elegent than what I had before. Again, I was making things too complicated

docker-compose.yml
```diff
@@ -3,6 +3,22 @@ version: '3.1'
...
+  localstack:
+    image: localstack/localstack
+    ports:
+      - "53:53"
+      - "443:443"
+      - "4510-4520:4510-4520"
+      - "4566-4620:4566-4620"
+      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"
+    environment:
+      - DEBUG=1
```
hmhsu, line 15: Suggestion (blocking): we probably doesn't need `DEBUG`. Also would you set `SERVICES=s3` so that we activate s3 only?

docker-compose.yml
```diff
@@ -3,6 +3,22 @@ version: '3.1'
...
 services:
 
+  localstack:
+    image: localstack/localstack
+    ports:
+      - "53:53"
+      - "443:443"
+      - "4510-4520:4510-4520"
+      - "4566-4620:4566-4620"
+      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"
```
hmhsu, line 13: Suggestion (blocking): I think we can get away with only `4566:4566`.

docker-compose.yml
```diff
@@ -3,6 +3,22 @@ version: '3.1'
...
+    environment:
+      - DEBUG=1
+      - DATA_DIR=/tmp/localstack/data
```
hmhsu, line 16: Suggestion (non-blocking): We can probably start with non-persistent setup. Would you remove this variable? I don't feel strongly though, making this non-blocking.

universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -386,3 +386,4 @@ ebean.default = "models.*"
 
 aws.s3.region=${?AWS_S3_REGION}
 aws.s3.bucket=${?AWS_S3_BUCKET_NAME}
+aws.local.endpoint="http://localhost:4566"
```
hmhsu, line 389: Suggestion (blocking): I suspect this needs to be `http://localstack:4566`.

bin/run-dev
```diff
@@ -4,7 +4,18 @@
 $( dirname ${BASH_SOURCE[0]})/pull-image
 
 if [ -f ".secrets" ]; then
-	. .secrets
+    . .secrets
 fi
 
+# start localstack
+docker-compose up -d localstack
```
hmhsu, line 11: Praise: nice! I didn't we can do this.

bin/localstack/stop-localstack
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker-compose -f bin/localstack/localstack.yaml down
```
hmhsu, line 3: Nit (blocking): We don't use the file anymore, please remove it.

hmhsu (approved):
It looks real nice now. Thanks for your patience bearing with me.

universal-application-tool-0.0.1/conf/application.conf
syntapy: Oh interesting. Is it because of the docker networking going on it looks for services as urls instead?

hmhsu: yup, it uses the container names in docker-compose.yml as hostnames.

docker-compose.yml
syntapy: Yeah you're right, its faster when we add the ` - SERVICES=s3` to the `docker-compose.yaml`

syntapy:
Ohh yeah, got my first merge in. Thank you @hmhsu for help and guidance

hmhsu:
You're welcome and thanks for your contribution! If you are interested in getting more involved or simply taking a peek at our day-to-day dev discussion, feel free to join our [slack channel](civiform.slack.com). Totally okay if not.

PR #550, opened 2021-03-19
setup playwright browser testing

### LINKED ISSUE #542: Set up a browser test runner using Playwright and Jest
We would like to move our current browser testing from selenium to using [Playwright](https://github.com/microsoft/playwright) and [Jest](https://jestjs.io/).

**Background**

Writing and running tests in selenium has proven to be slow and we expect to benefit from playwright's faster execution speed and cross-platform testing features.

**Scope**
Set up a test runner that runs civiform+postgres in dockers containers and runs node/jest/playwright in a separate container. For this issue, the test should load the landing page and assert that it gets a correct response from the server.

Done when:
- [ ] a single playwright test passes loading the landing page
- [ ] the playwright test suite is integrated with CI by running in a github action

bion:
Closes #542


natsid (approved):

browser-test/src/other_page.test.ts
```diff
@@ -0,0 +1,10 @@
+import { chromium } from 'playwright'
+
+it('this is the other page test', async () => {
```
bion, line 3: This test is included temporarily.

PR #552, opened 2021-03-19
remove webpack dependency and generate tailwind css using tailwindcli

### LINKED ISSUE #475: Replace Webpack with Tailwind CLI for building stylesheets
Styles are currently built using [Webpack](https://webpack.js.org/) (#214 ) which is overkill CiviForm because we don't make heavy use of JS or client-side rendering.

The build process would likely be faster without webpack using the Tailwind CLI instead.

Read background about Tailwind and its CLI here: https://tailwindcss.com/docs/installation

Done when: webpack is no longer in the project and Tailwind stylesheet is generated via the tailwind CLI.

Snackeroo:
### Description
- Remove Webpack dependency and use Tailwind CLI to generate TailwindCSS instead by reconfiguring build definition files
- There is still one Webpack test in pr.yml that can be removed if this approach is approved 

### Timing sample
* old: webpack 5.24.2 compiled with 4 warnings in 8390 ms
* new: tailwindcss 2.0.4 building finished in 7.73 s

### Checklist

### Issue(s)
Fixes #475


bion:
I checked out the branch and am seeing tailwind styles loading correctly in dev:

Screen Shot 2021-03-19 at 5 38 36 PM

Tailwind weighs about 4MB without us doing anything to purge unused styles. That was already true with the webpack config and it's something we can follow up on as we get closer to production launch.

.gitignore
```diff
@@ -10,3 +10,4 @@ RUNNING_PID
 .secrets
 node_modules
 bundle.js
```
natsid, line 12: thought (non-blocking): We should be able to remove `bundle.js` from here now. But we'd probably want to make sure everyone deletes any `bundle.js` they still have, so as not to accidentally check it in. We can remove it in a follow-up PR after this one is merged to allow a small transition period.

universal-application-tool-0.0.1/project/Tailwindbuilder.scala
```diff
@@ -3,20 +3,14 @@ import sbt.File
...
       override def beforeStarted() = {
         process = Option(
-          Process("npx webpack", base).run()
-        )
-      }
-
-      override def afterStarted() = {
-        process = Option(
-          Process("npx webpack --watch", base).run()
+          Process("npx tailwindcss-cli@latest build ./app/assets/stylesheets/styles.css -o ./public/stylesheets/tailwind.css", base).run()
```
natsid, line 13: question (non-blocking): If we ever wanted an additional css file + output file, we'd need to add it to this line and the one below, right?

universal-application-tool-0.0.1/tailwindbuilder.sbt
```diff
@@ -0,0 +1,17 @@
+import scala.sys.process.Process
+
+lazy val tailwindCli = TaskKey[Unit]("run webpack when packaging the application")
+
+def runTailwindCli(file: File) = {
+  Process("npx tailwindcss-cli@latest build ./app/assets/stylesheets/styles.css -o ./public/stylesheets/tailwind.css", file) !
```
natsid, line 6: suggestion (non-blocking): It might be nice to define this command string in a constant somewhere so it can be shared with the one in `Tailwindbuilder.scala`, especially since both places would need to be updated if we ever made a change to this command. But I have no idea how dependencies between `.scala` and `.sbt` files work, so no worries if it's a pain.

universal-application-tool-0.0.1/project/Tailwindbuilder.scala
```diff
@@ -3,20 +3,14 @@ import sbt.File
 
 import scala.sys.process.Process
 
-object Webpack {
+object Tailwindbuilder {
```
natsid, line 6: suggestion (blocking): Use camel case here: `TailwindBuilder`. `TailwindCssBuilder` might be even better.

natsid (commented):
Good stuff! Left a few comments - only one blocking naming suggestion.

AWBaumann (approved):

universal-application-tool-0.0.1/project/Tailwindbuilder.scala
bion: or add it as a completely separate file in app/assets

AWBaumann:
> I checked out the branch and am seeing tailwind styles loading correctly in dev:
> 
> Screen Shot 2021-03-19 at 5 38 36 PM
> 
> Tailwind weighs about 4MB without us doing anything to purge unused styles. That was already true with the webpack config and it's something we can follow up on as we get closer to production launch.

With us building the CSS in Java I'm not sure how easy it'd be to prune styles.

kendrickt:
Just to get some numbers recorded:

* webpack 5.24.2 compiled with 4 warnings in 8390 ms
* tailwindcss 2.0.4 building finished in 7.73 s

universal-application-tool-0.0.1/project/Tailwindbuilder.scala
Snackeroo: Appreciate the comments, I will make changes accordingly.

universal-application-tool-0.0.1/tailwindbuilder.sbt
Snackeroo: I will publish further documents on this and add the constant as suggested.

.gitignore
Snackeroo: Sounds good!

universal-application-tool-0.0.1/project/Tailwindbuilder.scala
Snackeroo: @natsid As much as I understand the Build hook, that is correct. Tailwind CLI being used here to compile that styles.css, which uses @tailwind directives, to tailwind.css. If the CSS files are not utilizing TailwindCLI to compile, there is no need to add it to this line, you can just put inside app/assets as mentioned by Bion or use public directory since any js/css files under public package are automatically packaged and added to class path, typically, asset.jars. Please correct me if I'm wrong.

Issue #508, opened 2021-03-15
Form system has versioning


neel-a-mehta:
Each time the UAT admin updates a program form or question, the entire form system gets a new version. Initially this version is just a draft, but then the admin can choose to publish.

![image](https://user-images.githubusercontent.com/77128518/111206195-75a4e580-859e-11eb-9513-a4536f532600.png)

![image](https://user-images.githubusercontent.com/77128518/111206205-79386c80-859e-11eb-874a-db1d29e2d9fd.png)

Related to this:

#243 
#244 
#245

PR #553, opened 2021-03-20
Add notion of lifecycle stages to backend - predecessor to versioning

ndmckinley:
### Description
This makes the necessary database changes for applications, programs, and questions to have lifecycle states.  All three go `DRAFT` (editable but not visible to others) -> `ACTIVE` (visible to others, not editable) -> `OBSOLETE` (usable by in-progress or existing apps / programs, not visible to new ones) -> `DELETED` (no longer visible to anyone).

Also updates application submit logic to use these states, since that's the easiest of the three!  Doesn't do anything to UX yet.

### Checklist

### Issue(s)
Related to #508.

### DIFF
--- a/universal-application-tool-0.0.1/app/models/Application.java
+++ b/universal-application-tool-0.0.1/app/models/Application.java
@@ -1,12 +1,12 @@
 package models;
 
 import io.ebean.annotation.DbJson;
+import io.ebean.annotation.UpdatedTimestamp;
 import java.time.Instant;
 import javax.persistence.Entity;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 import play.data.validation.Constraints;
-import services.WellKnownPaths;
 import services.applicant.ApplicantData;
 
 @Entity
@@ -16,18 +16,22 @@ public class Application extends BaseModel {
 
   @ManyToOne private Program program;
 
+  @Constraints.Required private LifecycleStage lifecycleStage;
+
+  @UpdatedTimestamp private Instant submitTime;
+
   // used by generated code
   @SuppressWarnings("UnusedVariable")
   @Constraints.Required
   @DbJson
   private String object;
 
-  public Application(Applicant applicant, Program program, Instant submitTime) {
+  public Application(Applicant applicant, Program program, LifecycleStage lifecycleStage) {
     this.applicant = applicant;
     ApplicantData data = applicant.getApplicantData();
-    data.putString(WellKnownPaths.APPLICATION_SUBMITTED_TIME, submitTime.toString());
     this.object = data.asJsonString();
     this.program = program;
+    this.lifecycleStage = lifecycleStage;
   }
 
   public Applicant getApplicant() {
@@ -40,5 +44,17 @@ public class Application extends BaseModel {
 
   public ApplicantData getApplicantData() {
     return new ApplicantData(this.object);
+  }
+
+  public LifecycleStage getLifecycleStage() {
+    return this.lifecycleStage;
+  }
+
+  public Instant getSubmitTime() {
+    return this.submitTime;
+  }
+
+  public void setLifecycleStage(LifecycleStage stage) {
+    this.lifecycleStage = stage;
   }
 }
--- a/universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
@@ -33,8 +33,8 @@ public class ApplicationRepository {
       ApplicantRepository applicantRepository,
       EbeanConfig ebeanConfig,
       DatabaseExecutionContext executionContext) {
-    this.programRepository = programRepository;
-    this.applicantRepository = applicantRepository;
+    this.programRepository = checkNotNull(programRepository);
+    this.applicantRepository = checkNotNull(applicantRepository);
     this.ebeanServer = Ebean.getServer(checkNotNull(ebeanConfig).defaultServer());
     this.executionContext = checkNotNull(executionContext);
   }
--- a/universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
@@ -5,62 +5,168 @@ import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import io.ebean.Ebean;
 import io.ebean.EbeanServer;
-import java.time.Clock;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 import javax.inject.Inject;
 import models.Applicant;
 import models.Application;
+import models.LifecycleStage;
 import models.Program;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import play.db.ebean.EbeanConfig;
+import services.applicant.ApplicantNotFoundException;
+import services.program.ProgramNotFoundException;
 
 public class ApplicationRepository {
   private final ProgramRepository programRepository;
   private final ApplicantRepository applicantRepository;
-  private final Clock clock;
   private final EbeanServer ebeanServer;
   private final DatabaseExecutionContext executionContext;
+  private static final Logger LOG = LoggerFactory.getLogger(ApplicationRepository.class);
 
   @Inject
   public ApplicationRepository(
       ProgramRepository programRepository,
       ApplicantRepository applicantRepository,
-      Clock clock,
       EbeanConfig ebeanConfig,
       DatabaseExecutionContext executionContext) {
     this.programRepository = programRepository;
     this.applicantRepository = applicantRepository;
-    this.clock = clock;
     this.ebeanServer = Ebean.getServer(checkNotNull(ebeanConfig).defaultServer());
     this.executionContext = checkNotNull(executionContext);
   }
 
+  /**
+   * Submit an application, which will delete any in-progress drafts, obsolete any submitted
+   * applications to a program with the same name (to include past versions of the same program),
+   * and create a new application in the active state.
+   */
   public CompletionStage<Application> submitApplication(Applicant applicant, Program program) {
     return supplyAsync(
         () -> {
-          Application application = new Application(applicant, program, clock.instant());
-          application.save();
-          return application;
-        });
+          return submitApplicationInternal(applicant, program);
+        },
+        executionContext.current());
   }
 
   public CompletionStage<Optional<Application>> submitApplication(
       long applicantId, long programId) {
+    return this.perform(
+        applicantId,
+        programId,
+        (ApplicationArguments appArgs) ->
+            submitApplicationInternal(appArgs.applicant, appArgs.program));
+  }
+
+  private Application submitApplicationInternal(Applicant applicant, Program program) {
+    ebeanServer.beginTransaction();
+    try {
+      List<Application> oldApplications =
+          ebeanServer
+              .createQuery(Application.class)
+              .where()
+              .eq("applicant.id", applicant.id)
+              .eq("program.name", program.getProgramDefinition().name())
+              .findList();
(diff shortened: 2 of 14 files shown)


universal-application-tool-0.0.1/conf/evolutions/default/9.sql
```diff
@@ -0,0 +1,18 @@
+# --- Add versioning and lifecycle management.
+
+# --- !Ups
+
+alter table questions add constraint nameversion unique(name, version);
+alter table questions add column lifecycle_stage varchar;
+alter table programs add column lifecycle_stage varchar;
+alter table applications add column lifecycle_stage varchar;
+alter table applications add column submit_time timestamp;
+alter table programs add column version bigint;
```
hmhsu, line 10: Question: This column doesn't seem to be included in program model?

universal-application-tool-0.0.1/app/models/Application.java
```diff
@@ -16,18 +16,22 @@
...
 
+  @Constraints.Required private LifecycleStage lifecycleStage;
+
+  @UpdatedTimestamp private Instant submitTime;
+
   // used by generated code
   @SuppressWarnings("UnusedVariable")
   @Constraints.Required
   @DbJson
   private String object;
```
hmhsu, line 27: Question (non-blocking): Is the plan to have separate applications and a consolidated profile (bank of answers) for an applicant? Feel free to point me to any doc that I missed.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -5,62 +5,169 @@
...
+    return applicantDb
+        .thenCombineAsync(
+            programDb,
+            (applicantMaybe, programMaybe) -> {
+              if (applicantMaybe.isEmpty()) {
+                throw new RuntimeException(new ApplicantNotFoundException(applicantId));
+              }
+              if (programMaybe.isEmpty()) {
+                throw new RuntimeException(new ProgramNotFoundException(programId));
+              }
```
hmhsu, line 106: Thought(non-blocking): this makes me feel that we should raise these exceptions in the corresponding repositories instead of here.

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -115,7 +116,11 @@ public ProgramDefinition getProgramDefinition(long id) throws ProgramNotFoundExc
       return ErrorAnd.error(errors);
     }
     Program program =
-        programDefinition.toBuilder().setName(name).setDescription(description).build().toProgram();
+        programDefinition.toBuilder()
+            .setName(name)
+            .setDescription(description)
+            .build()
+            .toProgram(LifecycleStage.DRAFT);
```
hmhsu, line 123: Question (non-blocking): I am curious why we are not adding stages when we create new programs.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -5,62 +5,169 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ApplicationRepository.class);
 
   @Inject
   public ApplicationRepository(
       ProgramRepository programRepository,
       ApplicantRepository applicantRepository,
-      Clock clock,
       EbeanConfig ebeanConfig,
       DatabaseExecutionContext executionContext) {
     this.programRepository = programRepository;
```
hmhsu, line 36: Nit (blocking): `checkNotNull` for this field and below.

hmhsu (approved):
Mostly LGTM with a few questions and comments.

bion:
Question (blocking): I could use some help understanding how lifecycle stages on these three resources map to the semantics of the overall system and the versioning logic in particular. Given that all three are versioned together in 
lockstep, my initial impression is that the per-resource lifecycle is more granularity than is needed.

Are all instances of all resources for a given version always in the same lifecycle stage? When a new version is published, does that require updating the lifecycle stage for all instances of all resources in the previous version?

I'm wondering if it would be simpler to derive the lifecycle stage for a given `questions`, `programs`, or `applicants` based on the lifecycle stage of a new `application_versions` resource, my guess is that would require less bookkeeping.

> DRAFT (editable but not visible to others)
What does it mean for an applicant to be in this stage?

> OBSOLETE (usable by in-progress or existing apps / programs, not visible to new ones)
What does it mean for an applicant to be in this stage?

> DELETED (no longer visible to anyone)
When would any resource instance have this lifecycle stage?

ndmckinley:
> Question (blocking): I could use some help understanding how lifecycle stages on these three resources map to the semantics of the overall system and the versioning logic in particular. Given that all three are versioned together in
> lockstep, my initial impression is that the per-resource lifecycle is more granularity than is needed.

The number one point there is that it's not applicants, it's applications.  Named similar, unfortunate - I keep typing the wrong one - but I think I got it right in this case.

> Are all instances of all resources for a given version always in the same lifecycle stage?

Nope!  There's only one "ACTIVE"-state one at a time though, and probably only one DRAFT also unless someone shows a need for more.  

See the wireframes here: https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354952974840&cot=14 which seem to show "previously published versions", and current draft, but seems to imply only one draft at a time.

> When a new version is published, does that require updating the lifecycle stage for all instances of all resources in the previous version?

Yep!  See the new `ApplicationRepositoryTest` to see how that works.

> I'm wondering if it would be simpler to derive the lifecycle stage for a given `questions`, `programs`, or `applicants` based on the lifecycle stage of a new `application_versions` resource, my guess is that would require less bookkeeping.

I think that would be less easy to reason about, about the same amount of work to implement, and enable fewer features.  This design (I think) will be pretty clear pretty quickly.

> > DRAFT (editable but not visible to others)
> 
> What does it mean for an applicant to be in this stage?

Application - it means they haven't submitted it yet.  It's how we track whether an applicant is currently working on applying to a program, so that if a new version of the program is published, the applicant still has access to the old one.

> > OBSOLETE (usable by in-progress or existing apps / programs, not visible to new ones)
> 
> What does it mean for an applicant to be in this stage?

Application - it means that they submitted a new application and this one is no longer relevant.  It helps admins only view the most recent application per applicant.

> > DELETED (no longer visible to anyone)
> 
> When would any resource instance have this lifecycle stage?

For XX days after someone hits "delete", probably - you really want to have an "undelete" for anything that someone has spent any significant amount of time on, people hate accidentally deleting things permanently.



If you'd like to discuss further, let's meet - let me know!  I'll work on Hung-Ming's comments in the meantime.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -5,62 +5,169 @@
...
+
+  private Application submitApplicationInternal(Applicant applicant, Program program) {
+    ebeanServer.beginTransaction();
+    try {
+      List<Application> oldApplications =
+          ebeanServer
+              .createQuery(Application.class)
+              .where()
+              .eq("applicant.id", applicant.id)
+              .eq("program.name", program.getProgramDefinition().name())
```
bion, line 72: issue (blocking): afaik we don't do anything to ensure that `programs` with the same name are actually the same program, it's just a text field the admin specifies. What assumptions are being made about programs that share a name here?

ndmckinley: That they are different versions of the same program - I'll lock that field in / grey it out when clicking a "new version" button, and disallow v1s of programs with the same name as existing ones.

bion: Will this prevent programs from being renamed?

ndmckinley: Unless we do some extra work, yeah, a program with a new name restarts its identity at v1.  We could track them differently, I suppose, do a "previous version" id pointer, maybe, if that's not acceptable.

bion: chatted about this, resolved

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -384,11 +395,12 @@ private ProgramDefinition updateProgramDefinitionWithBlockDefinition(
     ImmutableList<BlockDefinition> updatedBlockDefinitions =
         ImmutableList.copyOf(mutableBlockDefinitions);
 
+    // Implicitly "draft" since we are editing.
     Program program =
         programDefinition.toBuilder()
             .setBlockDefinitions(updatedBlockDefinitions)
             .build()
-            .toProgram();
+            .toProgram(LifecycleStage.DRAFT);
```
bion, line 403: Suggestion (non-blocking, optional): alternatively could add `.toDraftProgram()`

universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -64,12 +62,8 @@ public Tag renderApplicationListItem(long programId, Application application) {
     String applicantName = applicantService.applicantName(application);
     String lastEditText;
     try {
-      lastEditText =
-          application
-              .getApplicantData()
-              .readString(WellKnownPaths.APPLICATION_SUBMITTED_TIME)
-              .get();
-    } catch (NoSuchElementException e) {
+      lastEditText = application.getSubmitTime().toString();
```
bion, line 65: good improvement

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -5,62 +5,169 @@
...
+        .thenCombineAsync(
+            programDb,
+            (applicantMaybe, programMaybe) -> {
+              if (applicantMaybe.isEmpty()) {
+                throw new RuntimeException(new ApplicantNotFoundException(applicantId));
+              }
+              if (programMaybe.isEmpty()) {
+                throw new RuntimeException(new ProgramNotFoundException(programId));
+              }
+              return new ApplicationArguments(programMaybe.get(), applicantMaybe.get());
```
bion, line 107: Suggestion (non-blocking, optional): instead of making an inner class for this could do
```suggestion
              return Pair.of(programMaybe.get(), applicantMaybe.get());
```

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
```diff
@@ -5,62 +5,169 @@
...
+    public Program program;
+    public Applicant applicant;
+
+    public ApplicationArguments(Program program, Applicant applicant) {
+      this.program = program;
+      this.applicant = applicant;
+    }
+  }
+
+  private Application createOrUpdateDraftApplicationInternal(Applicant applicant, Program program) {
```
bion, line 131: Suggestion (blocking): does this actually update the application if it already exists? I'd expect a method called `update...`  to take args specifying what attributes to update. Suggest renaming this and relatedly named methods to `ensureDraftApplicationInternal`

bion (approved):

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
ndmckinley: Gonna skip this one I think - Guava Pair.of() is not in the external Guava package, and is actually recommended against by the Guava folks, so we'd need to import someone else's Pair library.  When I tried that, I think I found that java type inference can't figure that out - it doesn't compile for some reason, at least.

universal-application-tool-0.0.1/app/repository/ApplicationRepository.java
ndmckinley: I think it does, yeah - I'll add a test to make sure it does.  The only thing to update is submission time, though.  Applicant and program are the args specifying which entry to update - they uniquely identify a draft.

PR #582, opened 2021-03-22
Handle user input error in admin program block edit page

hmhsu:
### Description
If program block has empty name or description, display error message in the edit view instead of throwing server errors.

### Checklist
- [x] Updated tests which fail without the change (if possible)
- [ ] Add browser tests with playwright (will do it in another PR to keep this PR small)


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
```diff
@@ -52,11 +54,17 @@ public Result index(long programId) {
...
-      long blockId = program.getLastBlockDefinition().id();
-      return redirect(routes.AdminProgramBlocksController.edit(programId, blockId).url());
+      ErrorAnd<ProgramDefinition, CiviFormError> result =
+          programService.addBlockToProgram(programId);
+      ProgramDefinition program = result.getResult();
+      BlockDefinition block = program.getLastBlockDefinition();
+      if (result.isError()) {
+        String errorMessage = joinErrors(result.getErrors());
+        return renderEditViewWithMessage(request, program, block, errorMessage);
+      }
```
carolinedanzi, line 66: suggestion(blocking): I think this error check needs to be higher - we try to get the result on line 61, which will throw a RuntimeException if there is no result

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -135,29 +135,34 @@ public ProgramDefinition getProgramDefinition(long id) throws ProgramNotFoundExc
 
   @Override
   @Transactional
-  public ProgramDefinition addBlockToProgram(long programId) throws ProgramNotFoundException {
```
carolinedanzi, line 138: question: Perhaps a bit out of scope of this change, but why does this method exist if blocks must have a name/description?

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -135,29 +135,34 @@ public ProgramDefinition getProgramDefinition(long id) throws ProgramNotFoundExc
...
+  public ErrorAnd<ProgramDefinition, CiviFormError> addBlockToProgram(long programId)
+      throws ProgramNotFoundException {
     ProgramDefinition programDefinition = getProgramDefinition(programId);
     String blockName = String.format("Block %d", getNextBlockId(programDefinition));
 
-    return addBlockToProgram(programId, blockName, "", ImmutableList.of());
+    return addBlockToProgram(programId, blockName, "no description", ImmutableList.of());
```
carolinedanzi, line 143: suggestion: maybe "default description"? It seems odd to me that we are writing blocks with placeholder descriptions

carolinedanzi (commented):
This is looking nice! A  few questions

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
hmhsu: This method is used in the admin UI when we click the add block button. It adds a block with pre-populated fields without asking you to input. I guess this is for a smooth experience that users can edit the name/description later if needed.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
hmhsu: The result actually includes an unmodified program definition in the error case. We need the program definition anyway to finish the rendering regardless whether the add succeeds or not so I figured I can return it with the errors. I can update the docstring to make that clearer. Or if you feel this is making it more confusing than the benefit of convenience, I can change it back to have a second call to `getProgramDefinition` in the error case.

hmhsu: Updated javadoc.

carolinedanzi: Thank you, that makes much more sense! Thank you for updating the documentation, too

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
carolinedanzi: Got it, thank you!

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
carolinedanzi: Maybe we can give some instructions to the admin, like "What is the purpose of this block? Add a description that summarizes the information collected"

That way they have a call to action

carolinedanzi (approved):
Thank you for answering my questions! Looks good

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
hmhsu: Can do. Done.

PR #583, opened 2021-03-23
Mount sbt cache volumes in dev container

bion:
Following advice from joshuasuereth:

> I think your caching issues are 100% related to your docker setup.  While I love docker, sbt relies on access to "global directories" where it caches things between invocations. If you can local-mount the following it should help:
> - `*/target directories`
> - `~/.coursier`
> - `~/.sbt`
> - `~/.ivy`


ndmckinley (approved):
LGTM as long as your local testing looked good!

PR #584, opened 2021-03-23
Rename getBlockDefinitionByIndex to distinguish getting by index from getting by id

hmhsu:
### Description
getBlockDefinition was overloaded by two signatures, one with an int argument (index) and one with a long argument (id). Rename getting by index to distinguish two semantics. This function is only used in tests.

### Checklist


carolinedanzi (approved):
Nice, thanks for clarifying this!

kendrickt:
lgmt!

kendrickt (commented):
lgmt!

PR #586, opened 2021-03-23
Add tests that fail if a new QuestionType enum is added

### LINKED ISSUE #581: Assert all QuestionTypes handled in tests for classes which switch on QuestionType
When adding a new `QuestionType`, it can be challenging to make sure you haven't forgotten to update a switch statement somewhere which switches on type

We can add `@AfterClass` methods in the test files that checks we have tested all `QuestionType`s, for example:

```java
// In each unit test, add the type it tests to this set.
private static Set<QuestionType> testedTypes = new HashSet<>();
...
@AfterClass
public void  assertAllQuestionTypesTested() {
  assertThat(testedTypes).containsAll(EnumSet.allOf(QuestionType.class));
}
```
This way we can guarantee we tested all types (handy for `QuestionBuilder`) and also now have tests that fail when you add a new type, so it is easier to not miss a file when adding a new question type.

**Done when**
`QuestionDefinitionBuilderTest`, `ApplicantQuestionRendererFactoryTest`, `QuestionConfigTest`, and `IconsTest` all have tests that fail if a new `QuestionType` enum value is added

carolinedanzi:
### Description
Adds tests that fail if a new `QuestionType` enum is added, so that it is harder to miss files that should be changed when adding a new question type.

### Checklist

### Issue(s)
Fixes #581 

Sample test failure (added a NEW_TYPE enum):
Screen Shot 2021-03-23 at 11 57 15 AM

### DIFF
--- a/universal-application-tool-0.0.1/build.sbt
+++ b/universal-application-tool-0.0.1/build.sbt
@@ -30,6 +30,9 @@ lazy val root = (project in file("."))
       "org.junit.jupiter" % "junit-jupiter-params" % "5.4.2" % Test,
       "com.h2database" % "h2" % "1.4.199" % Test,
 
+      // Parameterized testing
+      "pl.pragmatists" % "JUnitParams" % "1.1.0" % Test,
+
       // Testing libraries for dealing with CompletionStage...
       "org.assertj" % "assertj-core" % "3.14.0" % Test,
       "org.awaitility" % "awaitility" % "4.0.1" % Test,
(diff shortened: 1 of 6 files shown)


AWBaumann (approved):

AWBaumann (approved):
So when we want to add a new QuestionType, we just add it to the ignore list along-side REPEATER? 😅

carolinedanzi:
> So when we want to add a new QuestionType, we just add it to the ignore list along-side REPEATER? 😅

I'm hoping to get RepeatedQuestionDefinition in asap so there won't be an ignore list to add to soon :)

PR #591, opened 2021-03-23
Run browser tests in CI

bion:
- [x] run the playwright browser tests on CI for each PR
- [x] remove the webpack test since we're not using webpack anymore

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -14,6 +14,20 @@ jobs:
         run: docker build -t civiform ./
       - name: Run tests
         run: docker run -v /var/run/docker.sock:/var/run/docker.sock civiform test
+  run_browser_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: check out pr branch
+        uses: actions/checkout@v2
+      - name: Build test app container
+        run: docker build -t civiform ./
+      - name: Build browser testing container
+        run: bin/build-browser-tests
+      - name: Build and run the test stack
+        run: docker-compose -f browser-test/browser-test-compose.yml up -d
+      - name: Run browser tests
+        run: bin/run-browser-tests
+
   test_prod:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


bin/pull-image
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 if [ -z "$USE_LOCAL_CIVIFORM" ]; then
-	echo "Making sure we're up to date with the latest dev... set environment variable USE_LOCAL_CIVIFORM=1 to skip"
-	if ! docker pull public.ecr.aws/t1q6b4h2/civiform-dev:latest; then
-		echo "Failed to pull - possibly some stuck credentials.  De-authenticating then retrying..."
-		docker logout https://public.ecr.aws/t1q6b4h2/civiform-dev
-		docker pull public.ecr.aws/t1q6b4h2/civiform-dev:latest
-	fi
-	docker tag public.ecr.aws/t1q6b4h2/civiform-dev:latest civiform
```
bion, line 10: 💁 converting tabs to spaces

.github/workflows/pr.yaml
```diff
@@ -29,10 +43,6 @@ jobs:
         # Confirm that we get a response on port 9000.
         run: while ! docker run --network host appropriate/curl -v -s --retry-max-time 180 --retry-connrefused http://localhost:9000/ ; do sleep 5; done
         timeout-minutes: 3
-      - name: Specifically test that webpack assets are present.
-        # Confirm that we get a response on port 9000.
-        run: while ! docker run --network host appropriate/curl -v -s --retry-max-time 180 --retry-connrefused http://localhost:9000/assets/javascripts/bundle.js ; do sleep 5; done
-        timeout-minutes: 1
```
bion, line 35: We don't use webpack anymore, so I'm surprised this has been passing.

ndmckinley: Does the js file still wind up at bundle.js? That would explain it.

ndmckinley (approved):
Do you want to remove the existing Selenium tests, or migrate them slowly?

bion:
> Do you want to remove the existing Selenium tests, or migrate them slowly?

Migrate them piecemeal

.github/workflows/pr.yaml
bion: Maybe just cached? It was sitting around on my local dev file system. I deleted it and re-ran `docker build -f prod.Dockerfile -t civiform:prod ./` and `docker-compose -f infra/prod-simulator-compose.yml up` and it didn't regenerate.

ndmckinley: Hmmm... shouldn't be, but all right, seems like a pretty benign mystery.  I'll investigate while tests are running or I'm otherwise sitting around but won't freak out if I don't figure it out.

PR #593, opened 2021-03-23
Remove Path.parentPaths()

kendrickt:
The order of paths was very specific and felt dangerous to use, and it was only used in one place (ApplicantData.put) which can be done with recursion.

### Description
Path.parentPaths() is removed in favor for recursion, which is not dependent on the order of paths. 

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/services/Path.java
+++ b/universal-application-tool-0.0.1/app/services/Path.java
@@ -5,7 +5,6 @@ import com.google.auto.value.extension.memoized.Memoized;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
-import java.util.ArrayList;
 
 /**
  * Represents a path into the applicant JSON data. Stored as the path to data without the JsonPath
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -120,16 +120,10 @@ private void putNull(Path path) {
...
-        .forEach(
-            segmentPath -> {
-              if (!hasPath(segmentPath)) {
-                this.jsonData.put(
-                    segmentPath.parentPath().path(), segmentPath.keyName(), new HashMap<>());
-              }
-            });
-
-    this.jsonData.put(path.parentPath().path(), path.keyName(), value);
+    if (!hasPath(path.parentPath())) {
```
carolinedanzi, line 123: Neat!

I know "applicant" is pretty much in every path, but I think we should still include a base case here. If for some reason the path doesn't start with "applicant", eventually you'll get to Path.empty(), which will throw an error like "Json cannot be null or empty"

kendrickt: Ahh thanks! I think that may be why all the other tests are failing.

kendrickt: It was not! :(

universal-application-tool-0.0.1/app/services/Path.java
```diff
@@ -42,12 +42,22 @@ private static Path create(ImmutableList<String> pathSegments) {
    */
   public abstract ImmutableList<String> segments();
 
+  public boolean isEmpty() {
```
carolinedanzi, line 45: super nit: may as well `@Memoize` this too

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -120,7 +120,7 @@ private void putNull(Path path) {
    * @param value the value to place; values of type Map will create the equivalent JSON structure
    */
   private void put(Path path, Object value) {
-    if (!hasPath(path.parentPath())) {
+    if (!path.parentPath().isEmpty() && !hasPath(path.parentPath())) {
```
carolinedanzi, line 123: nice! suggestion (blocking): mind adding a quick test for this case, just to be safe?

carolinedanzi (approved):

PR #601, opened 2021-03-24
Add IDCS / ADFS client ids and secrets to our semi-prod env.

ndmckinley:
### Description
Add environment variables to hook 'prod' up to city idcs and adfs.

### Checklist

### DIFF
--- a/infra/containers.yaml
+++ b/infra/containers.yaml
@@ -109,6 +109,14 @@ Resources:
               Value: !Sub '${AWS::Region}'
             - Name: AWS_S3_BUCKET_NAME
               Value: !Ref 'S3BucketName'
+            - Name: IDCS_CLIENT_ID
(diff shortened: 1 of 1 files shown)


hmhsu (approved):

infra/containers.yaml
```diff
@@ -109,6 +109,14 @@ Resources:
               Value: !Sub '${AWS::Region}'
             - Name: AWS_S3_BUCKET_NAME
               Value: !Ref 'S3BucketName'
+            - Name: IDCS_CLIENT_ID
+              Value: '{{resolve:secretsmanager:dev-oidc-secrets:SecretString:IDCS_CLIENT_ID}}'
```
hmhsu, line 113: Thoughts (non-blocking): the names `dev-xxx` suggest that these are not for prod use, maybe we should create ones that are only used in prod?

ndmckinley: That is definitely correct - but we only have dev secrets right now, this is to-be-updated when we get prod secrets for real.  That will be after our security review.

Issue #589, opened 2021-03-23
Create a distinct `QuestionForm` class for each of the existing question types


natsid:
We currently have a single `QuestionForm` class for creating a question. In order to implement the solution proposed in [this doc](https://docs.google.com/document/d/1r9_NZYQWxzBTF1zwKjeeEzpoUEejtrmMITmnCw-XlLk/edit?usp=sharing), which allows admins to configure their own custom validations, we need to create a separate form class for each question type (`TextQuestionForm`, `AddressQuestionForm`, etc). Each of these form classes should extend the existing `QuestionForm`, which will have all of the common form fields for questions, and then add their own question type-specific fields (such as `minLength` and `maxLength`, for the text question type).

This is part of the implementation proposed in [this doc](https://docs.google.com/document/d/1r9_NZYQWxzBTF1zwKjeeEzpoUEejtrmMITmnCw-XlLk/edit?usp=sharing).

This work for this issue blocks #377.

**Done when** there is a form class for each of the currently existing question types (currently Text, Name, Address, and Number).

This task does not include utilizing and routing to these question form classes yet.

natsid:
Upon starting work on this, I realize I divided the tasks very awkwardly. This task will actually be done in parallel with #590.

carolinedanzi:
(Milestone 4 review): This is in progress, but will need to be moved to Milestone 5

PR #604, opened 2021-03-24
Create `TextQuestionForm extends QuestionForm`

natsid:
### Description
Add a new `TextQuestionForm` class that will be used to bind POST requests for creating **text** questions specifically. It's the same as a `QuestionForm`, but has additional text question-specific configurations.

More type-specific form classes to follow. In order to keep PRs small and manageable for review, this class is not currently used outside of its unit tests. That will come in following PRs.

### Checklist
- [x] Created tests which fail without the change 

### Issue(s)
Works toward #589

### DIFF
--- a/universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
@@ -19,7 +19,7 @@ public class TextQuestionForm extends QuestionForm {
   }
 
   public TextQuestionForm(TextQuestionDefinition qd) {
-    super();
+    super(qd);
     textMinLength = qd.getMinLength();
     textMaxLength = qd.getMaxLength();
   }
--- a/universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
@@ -1,0 +1,58 @@
+package forms;
+
+import java.util.OptionalInt;
+import services.question.InvalidQuestionTypeException;
+import services.question.QuestionDefinitionBuilder;
+import services.question.TextQuestionDefinition;
+
+public class TextQuestionForm extends QuestionForm {
+  private OptionalInt textMinLength;
+  private OptionalInt textMaxLength;
+
+  public TextQuestionForm() {
+    super();
+    textMinLength = OptionalInt.empty();
+    textMaxLength = OptionalInt.empty();
+    // TODO(https://github.com/seattle-uat/civiform/issues/590): Use QuestionType instead of String
+    //  for this?
+    setQuestionType("TEXT");
+  }
+
+  public TextQuestionForm(TextQuestionDefinition qd) {
+    super();
+    textMinLength = qd.getMinLength();
+    textMaxLength = qd.getMaxLength();
+  }
+
+  public OptionalInt getTextMinLength() {
+    return textMinLength;
+  }
+
+  public void setTextMinLength(int textMinLength) {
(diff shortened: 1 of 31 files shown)


universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
```diff
@@ -0,0 +1,58 @@
...
+    super();
+    textMinLength = OptionalInt.empty();
+    textMaxLength = OptionalInt.empty();
+    // TODO(https://github.com/seattle-uat/civiform/issues/590): Use QuestionType instead of String
+    //  for this?
+    setQuestionType("TEXT");
+  }
+
+  public TextQuestionForm(TextQuestionDefinition qd) {
+    super();
```
AWBaumann, line 22: Should this be `super(qd);`?

natsid: Yes - thank you!

natsid: done

natsid: Done, and added a test for this.

universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
```diff
@@ -0,0 +1,58 @@
...
+    super(qd);
+    textMinLength = qd.getMinLength();
+    textMaxLength = qd.getMaxLength();
+  }
+
+  public OptionalInt getTextMinLength() {
+    return textMinLength;
+  }
+
+  public void setTextMinLength(int textMinLength) {
```
hmhsu, line 31: Questions (non-blocking): I am not super familiar with html form so this question is just for my understanding. What are the types supported in the html form? I originally thought only text is supported so arguments here needs to be string as well.

hmhsu (approved):

universal-application-tool-0.0.1/app/forms/TextQuestionForm.java
natsid: You might actually be right... I thought that the Play Framework FormFactory stuff might give us some parsing out of the box (since you are right that the request just gets sent over as strings, as far as I understand).

Let me do some more research.

hmhsu: I think Play may actually do the conversion for us. My only concern there is that if the parsing fails, Play likely throws a runtime exception that results in a 5XX server error. But we'd like to handle that as expected user input errors and show an error message on the same page.

natsid: That is a good point!

So what do you think - safer to accept a string and manually parse? Then re-render with an error message?

natsid: It _might_ be overkill, since in the admin questions form, we can (and do) use an input field with `type="number" step="1"`, which makes it essentially impossible to enter a non-int. But it is still probably safer to check it ourselves.

hmhsu: Ah I see. Yeah, personally I'd manually parse it, but don't feel strongly.

natsid: It's not clear where an error thrown in `setMinLength` would surface - I'm guessing it would happen in the `form.bindFromRequest` call?

I'm going to leave it as is for now since I think it'd be a non-trivial effort to pipe it up correctly.

PR #608, opened 2021-03-24
Make browser testing nice and approachable

bion:


browser-test/bin/wait_for_server_start_and_run_tests.sh
```diff
@@ -18,4 +18,19 @@ echo detected server start
 
 ./bin/truncate_tables.sh
 
-yarn test $@
+debug=0
+for arg in "$@"; do
+    case $arg in
+        -d|--debug)
+        debug=1
+        shift # Remove --debug from processing
```
bion, line 26: does this mean --debug must come before other arguments?

natsid: Specifically the `shift` line? Hmm let me test this to double check. I wouldn't think so though.

natsid: You're right about `shift`. I fixed and manually tested it.

browser-test/bin/truncate_tables.sh
```diff
@@ -2,7 +2,15 @@
...
-cat <<EOF | PGPASSWORD=example psql -h db -p 5432 -U postgres 
+if [[ ${BASE_URL-not_set} == "http://localhost:9999" ]]; then
+    host=localhost
+    port=2345
+else
+    host=db
+    port=5432
+fi
+
+cat <<EOF | PGPASSWORD=example psql -h $host -p $port -U postgres
```
hmhsu, line 13: Thoughts (non-blocking): any chance we can run this in a docker container? `postgres` image has this CLI. My environment doesn't have psql installed that's why I asked 😛 .

bin/run-browser-tests-local
```diff
@@ -0,0 +1,7 @@
+#! /usr/bin/env bash
+
+pushd browser-test
```
hmhsu, line 3: Suggestions (non-blocking): this probably only works when run under main folder, consider using `$( dirname ${BASH_SOURCE[0]})/../browser-test` or something similar.

browser-test/.gitignore
```diff
@@ -0,0 +1,6 @@
+.DS_Store
+.idea
+*.log
+tmp/
+
+
```
hmhsu, line 6: Nit (non-blocking): extra line

universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -52,7 +52,7 @@ public Content render(Http.Request request, Optional<String> message) {
 
     // "defense in depth", sort of - this client won't be present in production, and this button
     // won't show up except when running locally.
-    if (request.host().startsWith("localhost:")) {
+    if (request.host().startsWith("localhost:") || request.host().startsWith("civiform:")) {
```
hmhsu, line 55: Thoughts (non-blocking): since we have this check in at least two places, maybe we can make it configurable somewhere. Not sure if that's secure though.

hmhsu (approved):

browser-test/.gitignore
bion: Done.

bin/run-browser-tests-local
bion: Done.

browser-test/bin/truncate_tables.sh
bion: Sure, I added some code to do this but it's a little tricky to test since I do have `psql` locally. Mind pulling it down and running `./bin/run-browser-tests-local` to see if it works for you?

hmhsu: It is failing saying some command/file is not found

```
% ./bin/run-browser-tests-local 
yarn run v1.22.10
$ ./bin/truncate_tables.sh && jest --forceExit --maxConcurrency=1
[FATAL tini (7)] exec -e failed: No such file or directory
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

hmhsu: Pushed a fix to branch. I can confirm this works for my setup.

hmhsu: Thanks for making the change!

bion: Oh great, thank you!

universal-application-tool-0.0.1/app/views/LoginForm.java
bion: Done.

PR #620, opened 2021-03-25
Add a knob to easily disable localstack/s3

hmhsu:
### Description
Add a few settings to easily disable aws and localstack.

To disable S3 connection from civiform:
- set aws.enable to false in application.conf

To disable localstack from our docker compose
- run `bin/stop-dev`
- set environment variable 
  ```shell
  export DISABLE_LOCALSTACK=1
  ```

### DIFF
--- a/bin/run-dev
+++ b/bin/run-dev
@@ -7,6 +7,13 @@ if [ -f ".secrets" ]; then
     . .secrets
 fi
 
+if [ "$DISABLE_LOCALSTACK" ]; then
+  echo "localstack is disabled, only starting civiform and db..."
+  docker-compose up db civiform
+  exit $?
+fi
+echo "Starting localstack... set environment variable DISABLE_LOCALSTACK=1 to skip"
+
 # start localstack
 docker-compose up -d localstack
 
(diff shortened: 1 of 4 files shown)


bin/run-dev
```diff
@@ -7,14 +7,21 @@ if [ -f ".secrets" ]; then
     . .secrets
 fi
 
+if [ "$DISABLE_LOCALSTACK" ]; then
```
ndmckinley, line 10: I'd prefer we find another approach here.  I think this is going to add to maintenance burden - we'll need to remember to add any new container to line 12 here, and I think we probably won't.  Is there any other way to do this?

hmhsu: Done.

docker-compose.yml
```diff
@@ -26,6 +26,7 @@ services:
     container_name: civiform
     links:
       - "db:database"
+      # comment below line and set environment variable DISABLE_LOCALSTACK=1 if you don't want localstack
       - "localstack"
```
ndmckinley, line 30: Can we remove this link?  "localstack" should be accessible at that name.  I'd love for DISABLE_LOCALSTACK to be the only thing we need.

ndmckinley (approved):

bin/run-dev
ndmckinley: Nice, good thought!

docker-compose.yml
hmhsu: I think so. The purpose of the link also serves as a declaration of dependency, but we are purposely waiting for localstack to be ready so that is not necessary, at least at the moment. I agreed that only having to set a shell variable is nice, so it seems a good tradeoff.

PR #623, opened 2021-03-29
Add a MultiOptionQuestionDefiniton abstract class

carolinedanzi:
### Description
1. Add a `MultiOptionQuestionDefinition` abstract class - this will be the parent class for dropdowns, radio buttons, toggle questions, and checkboxes
2. Add a field on the `QuestionType` enum so we can determine if a given type has multiple options
3. Add `DropdownQuestionDefinition` for use in testing - will fully implement in follow-up PR

Future changes: Implement validation and individual types (with rendering logic)

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
@@ -2,6 +2,7 @@ package services.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import java.util.Objects;
 import java.util.Optional;
@@ -10,6 +11,7 @@ import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import services.Path;
 import services.question.AddressQuestionDefinition;
+import services.question.MultiOptionQuestionDefinition;
 import services.question.NameQuestionDefinition;
 import services.question.NumberQuestionDefinition;
 import services.question.QuestionDefinition;
@@ -80,6 +82,10 @@ public class ApplicantQuestion {
     return new AddressQuestion();
   }
 
+  public MultiOptionQuestion getMultiOptionQuestion() {
+    return new MultiOptionQuestion();
+  }
+
   public TextQuestion getTextQuestion() {
     return new TextQuestion();
   }
@@ -96,6 +102,8 @@ public class ApplicantQuestion {
     switch (getType()) {
       case ADDRESS:
         return getAddressQuestion();
+      case MULTI_OPTION:
+        return getMultiOptionQuestion();
       case NAME:
         return getNameQuestion();
       case NUMBER:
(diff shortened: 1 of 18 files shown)


universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
```diff
@@ -582,6 +592,71 @@ public Path getNumberPath() {
     }
   }
 
+  public class MultiOptionQuestion implements PresentsErrors {
```
hmhsu, line 595: Question (blocking): MultiOptionQuestionDefinition is an abstract class while this class is concrete. Is the plan to have all multi-option questions to use this class, or did you mean DropdownQuestion here?

universal-application-tool-0.0.1/app/services/question/ScalarType.java
```diff
@@ -9,6 +10,7 @@
  * models.Applicant} JSON column and serialized using {@link services.applicant.ApplicantData}.
  */
 public enum ScalarType {
+  LIST(List.class),
```
hmhsu, line 13: Question (non-blocking): Do we ever want concrete `List` types? I am curious how this works.

universal-application-tool-0.0.1/conf/evolutions/default/3.sql
```diff
@@ -16,7 +16,8 @@ create table if not exists questions (
   description varchar,
   question_type varchar,
   question_text jsonb,
-  question_help_text jsonb
+  question_help_text jsonb,
+  question_options jsonb
```
hmhsu, line 20: Suggestion (blocking): the safest way to add a column would be to add a new script that adds the new column to the table.

universal-application-tool-0.0.1/app/services/question/MultiOptionQuestionDefinition.java
```diff
@@ -0,0 +1,109 @@
...
+      ImmutableListMultimap<Locale, String> options) {
+    super(
+        id,
+        version,
+        name,
+        path,
+        description,
+        questionText,
+        questionHelpText,
+        MultiOptionValidationPredicates.create());
```
hmhsu, line 35: Thoughts (non-blocking): Do we need specific validation predicates for each concrete multi-option question types? I presume it would be necessary but I am not 100% sure.

universal-application-tool-0.0.1/app/services/question/QuestionDefinitionBuilder.java
```diff
@@ -36,6 +40,11 @@ public QuestionDefinitionBuilder(QuestionDefinition definition) {
     questionHelpText = definition.getQuestionHelpText();
     questionType = definition.getQuestionType();
     validationPredicatesString = definition.getValidationPredicatesAsString();
+
+    if (definition.getQuestionType().isMultiOptionType()) {
+      MultiOptionQuestionDefinition singleSelect = (MultiOptionQuestionDefinition) definition;
```
hmhsu, line 45: Nit (blocking): the question may not always be single select even if it is now, and it doesn't seem to be necessary details here. Would you rename this variable to something like multiOption?

hmhsu (commented):
Nice work! Mostly looks good with a few quesitons.

universal-application-tool-0.0.1/app/services/question/ScalarType.java
carolinedanzi: Ah I fixed this in a different PR, will update here too. We will eventually want `LIST` for multi-select options (like checkboxes), but dropdowns for now are single-select. 

`ApplicantData` is set up to read/write `ImmutableList<String>` for when we do need it

universal-application-tool-0.0.1/app/services/question/MultiOptionQuestionDefinition.java
carolinedanzi: Great question! For now, the only [planned validation logic](https://docs.google.com/document/d/1HYVHBno6dk7fmS5w6V6Y3jxERHGfDoQKTtzBZr27Y6o/edit#heading=h.1i512171vbiq) across all four types is "min choices required" and "max choices allowed" - this is for multi-select questions like checkbox. For others, the default validation will be to set each of these to 1, so only 1 selection is valid

universal-application-tool-0.0.1/app/services/question/QuestionDefinitionBuilder.java
carolinedanzi: Oops, great catch! I once called this a "single select" question and forgot to change this variable name

universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
carolinedanzi: Great question - I believe all multi-option questions can use this class, at least for now. The ApplicantQuestion class provides methods for getting the applicant's answer if it exists, checking for validation errors, etc. - all of which at the moment are the same across multi-option questions.

We may want to have two different ones in the future - one for multi-answer questions (list) and one for single-answer questions (single string)

universal-application-tool-0.0.1/conf/evolutions/default/3.sql
carolinedanzi: Good to know! I think I did this, mind taking another look?

hmhsu: Looks good!

universal-application-tool-0.0.1/app/services/question/ScalarType.java
hmhsu: Thanks for the explanation. I was asking if we need `List<String>` or does `List` serve the need. This is just me learning Java 😅

universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
```diff
@@ -582,6 +592,71 @@ public Path getNumberPath() {
...
+    public boolean hasValue() {
+      return getSelectedOptionsValue().isPresent();
+    }
+
+    public Optional<ImmutableList<String>> getSelectedOptionsValue() {
+      if (selectedOptionsValue != null) {
+        return selectedOptionsValue;
+      }
+
+      selectedOptionsValue = applicantData.readList(getSelectionPath());
```
hmhsu, line 628: Question (blocking): Since we changed the scalar type, should this be updated as well?

hmhsu (approved):
Looks good with a suggestion.

universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
carolinedanzi: I think it makes sense to have a `SingleSelectQuestion` (dropdowns and radio buttons) and a `MultiSelectQuestion` (checkboxes). I updated to the new name and added a comment for implementing the other (linked to checkbox issue)

universal-application-tool-0.0.1/app/services/question/ScalarType.java
carolinedanzi: That's a good question! I think List will work for now, since we may have lists of different types (unlikely, but possible). If not, we could make a type reference for List<String> and get the class from that

universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
carolinedanzi: Done! I changed the class name to `SingleSelectQuestion` and added a TODO comment for checkboxes (i.e. `MultiSelectQuestion`)

Issue #555, opened 2021-03-22
enable civiform.com for our AWS instance


gcapiel:
Can we point civiform.com to our AWS instance.  I'm holding the domain under my personal GoDaddy account.  Please provide me the configuration that I need to set.

Gerardo

neel-a-mehta:
This will be essential for both user testing (we can point them to civiform.com) and helping community contributors see what work has been done already.

gcapiel:
we also need to put a warning across the top asking people to not enter personal/real data as this is just a demo/example system.  something like what mibridges does in their demo site referenced on github: [demo site](https://demo.michiganbenefits.org/).  maybe we can pick one of these tailwind formats: https://tailwindui.com/components/marketing/elements/banners

gcapiel:
because we can't transfer the domain to route 53 until 5/10, we will go with godaddy's domain forwarding to http://prod-civiform-publiclb-916393258.us-west-2.elb.amazonaws.com/ until then.  i tested it and it works, but reverted back to point to our github repo, because before we start sending  people there we should implement the banner.  we also need to show the login as admin button when the domains is *.amazonaws.com .

gcapiel:
Here's the banner CSS: https://play.tailwindcss.com/6yyHmMFEmG?size=540x720

hmhsu:
will work on adding the banner today/tomorrow.

gcapiel:
let's have the banner show on every page while the system is not officially in production even though it's on production servers.

PR #628, opened 2021-03-29
Show demo banner in every page

hmhsu:
### Description
Show demo banner in every page.

### Banner on admin pages
Screen Shot 2021-03-29 at 3 34 56 PM

### Issue(s)
#555

### DIFF
--- a/universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
+++ b/universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
@@ -1,13 +1,20 @@
 package views;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static j2html.TagCreator.div;
 import static j2html.TagCreator.document;
+import static j2html.TagCreator.p;
+import static j2html.TagCreator.span;
 
 import j2html.tags.ContainerTag;
 import j2html.tags.DomContent;
 import j2html.tags.Tag;
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
```diff
@@ -60,4 +69,88 @@ public String contentType() {
...
+      return addBgOpacity(Styles.BG_RED_600, Styles.ABSOLUTE, Styles.BOTTOM_2, Styles.W_FULL);
+    }
+    return new String[] {Styles.BG_RED_600};
+  }
+
+  private DomContent[] addDemoBanner(DomContent... domContents) {
+    if (isAdminPage()) {
+      return appendDomContent(domContents, demoBanner());
+    }
+    return prependDomContent(domContents, demoBanner());
```
hmhsu, line 141: Thoughts: this has an unfortunate side-effect of showing an un-styled banner before the tailwind styles are loaded. We probably want to move to appending all banners instead of prepending them.

AWBaumann: Let's take another pass at this. I don't love the branching logic, so I think we could probably simplify things if we come up with a style that we can use across the board.

hmhsu: Yeah, that looks way better. thanks!

PR #633, opened 2021-03-30
Add roles on merge correctly.

ndmckinley:
### Description
Correctly add roles to profiles when a profile already exists.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/UatProfileAdapter.java
+++ b/universal-application-tool-0.0.1/app/auth/UatProfileAdapter.java
@@ -1,7 +1,9 @@
 package auth;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 import java.util.Optional;
+import java.util.Set;
 import javax.inject.Provider;
 import models.Applicant;
 import org.pac4j.core.context.WebContext;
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/auth/UatProfileAdapter.java
```diff
@@ -42,11 +43,17 @@ public UatProfileAdapter(
...
+  protected abstract ImmutableSet<Roles> roles();
+
   /** Merge the two provided profiles into a new UatProfileData. */
   public UatProfileData mergeUatProfile(UatProfile uatProfile, OidcProfile oidcProfile) {
     String emailAddress = oidcProfile.getAttribute(emailAttributeName(), String.class);
     uatProfile.setEmailAddress(emailAddress).join();
     uatProfile.getProfileData().addAttribute(CommonProfileDefinition.EMAIL, emailAddress);
+    // Meaning: whatever you signed in with most recently is the role you have.
+    for (Roles role : roles()) {
+      uatProfile.getProfileData().addRole(role.toString());
```
hmhsu, line 55: Question: this is just for my understanding, is this something in the cookies? or is it stored in the db? I didn't quite understand the root cause of the issue.

hmhsu (approved):

universal-application-tool-0.0.1/app/auth/UatProfileAdapter.java
ndmckinley: Yep, all in the cookie - not persisted.  Clearing the cookie didn't fix it since it wasn't being set at all if the account already existed.

PR #635, opened 2021-03-31
Add styling to toast messages.

AWBaumann:
### Description
Add toast message support.

#### Creating toasts via Java
To display a toastMessage server-side, just attach a ToastMessage anywhere in the DOM.

```
ToastMessage.alert("This is my toast message.").setId("toast-id").getContainerTag();
```
creates the following div:
```
<div class="hidden cf-toast-data" id="toast-id" 
    canDismiss="true" canIgnore="false" toastDuration="3000" toastType="alert">
    This is my toast message.
</div>
```
When the page loads, the div will be turned into a rendered HTML message via javascript.

#### Creating toasts via JS

Toast messages can also be created client-side via javascript...

```
let newMessage = {
    id: "toast-id",
    canDismiss: true,
    canIgnore: false,
    content: "This is my toast message.",
    duration: 3000,
    type: "alert"
};
toastController.showToastMessage(newMessage);
``` 

image

#### TODOs
- Specify alert types when passing them to the views.


universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -54,6 +54,20 @@ function maybeShowWarning() {
   }
 }
 
+ /** If a toast message is present, make it visible for the specified duration. */
+function maybeShowToast() {
+  const toastMessage = document.querySelector('.cf-toast');
```
bion, line 59: Question (blocking): is there any risk of there being multiple toasts?

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -54,6 +54,20 @@ function maybeShowWarning() {
...
 
+ /** If a toast message is present, make it visible for the specified duration. */
+function maybeShowToast() {
+  const toastMessage = document.querySelector('.cf-toast');
+  if (toastMessage) {
+    toastMessage.classList.remove('opacity-0');
+    if (toastMessage.hasAttribute('duration')) {
+      const showDuration = toastMessage.getAttribute('duration');
+      setTimeout(function() {
+        toastMessage.classList.add('opacity-0');
```
bion, line 65: Issue (blocking): will this cause there to be an invisible element on the page that could cover and prevent interaction with other UI?

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
AWBaumann: Yes. I already have that in the TODO section, but perhaps I should just implement it now.

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
AWBaumann: Yes, it's possible. I'll fix that now along with the multiple toast portion (and convert the demo warning to a dismissible toast while I'm at it.)

universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -51,6 +44,11 @@ public Content render(Http.Request request, Optional<String> message) {
                     "Continue",
                     routes.CallbackController.callback(GuestClient.CLIENT_NAME).url())));
 
+    if (!message.orElse("").equals("login")) {
```
hmhsu, line 47: Nit: I think we need to negate the condition here.

hmhsu (approved):
NICE!

hmhsu:
My guess for the browser test failure is due to a timing issue. Giving a more specific selector (instead of just `html`) might fix it.

universal-application-tool-0.0.1/app/views/LoginForm.java
AWBaumann: Done

PR #637, opened 2021-03-31
Browser test typescript formatter

hmhsu:
### Description
- Reuse our formatter
- Overwrite command to only run tsfmt
- Updated wiki


browser-test/bin/fmt
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# find absolute path to local browser-test directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && cd .. && pwd )"
```
bion, line 4: Suggestion (blocking): this looks like it's intending to make the script runnable from any working directory, but also looks like it will change the current working directory to the project's base directory. Recommend either requiring the user to be in the project's working directory when they run the script or restoring their original directory after the script completes (`pushd` and `popd` may be helpful commands for the latter approach).

browser-test/bin/fmt
```diff
@@ -0,0 +1,12 @@
...
+
+# find relative path to typescript files
+FILES="$(cd $DIR; find src -name '*.ts' -not -path '*node_modules*' -not -path '*node-modules*' | xargs)"
```
bion, line 7: Question (blocking): why does `-not -path '*node_modules*'` appear twice?

bion (approved):
Suggestion (blocking): document this in the wiki

browser-test/bin/fmt
hmhsu: The change directory `cd` command is always run in a subshell, which doesn't change the directory of the current shell in script. Also I just tested, even if I explicitly change directly in the script, the script itself is run in a subshell so won't affect user's directory (if I understand correctly).

browser-test/bin/fmt
hmhsu: Removed.

hmhsu:
> Suggestion (blocking): document this in the wiki

Done.

PR #639, opened 2021-03-31
Minor browser tests fixes

hmhsu:
### Description
- rename run-test-env to run-browser-test-env
- add a few more question types
- close browser after test
- consolidate BASE_URL to one place

Reviewers:
One of @carolinedanzi and @bion


browser-test/src/support/index.ts
```diff
@@ -12,10 +12,16 @@ export const startSession = async () => {
...
+  browser.close()
+}
+
+export const logout = async (page: Page) => {
+  // TODO: add logout button to applicant page and use that
+  page.goto('/logout')
+}
 
 export const loginAsAdmin = async (page: Page) => {
-  await page.goto(BASE_URL + '/loginForm')
```
carolinedanzi, line 18: question: what will happen if we try  to login from a different page?

browser-test/src/support/admin_questions.ts
```diff
@@ -33,4 +54,52 @@ export class AdminQuestions {
...
+    await this.page.fill('text=Question help text', helpText)
+
+    await this.page.click('text=Create')
+
+    expect(await this.page.innerText('h1')).toEqual('All Questions')
+
+    const tableInnerText = await this.page.innerText('table')
+
+    expect(tableInnerText).toContain(questionName)
+    expect(tableInnerText).toContain(questionText)
```
carolinedanzi, line 79: suggestion (non-blocking): Could we extract these lines to a helper function, since they appear to be the same for all types of questions?

carolinedanzi (approved):

bion (approved):

browser-test/src/support/index.ts
hmhsu: If we are on a different page, this doesn't work. But we are trying to move away from using URLs to navigate in our browser tests and purely use the UI. As long as we properly log out (which is a prerequisite to log in as admin), we will be redirect to login page so I think removing this should be fine.

browser-test/src/support/admin_questions.ts
hmhsu: Sure. If you don't mind, I'd like to do it in a followup PR.

carolinedanzi: No worries! I'm actually trying to write a browser test now, and might take back my suggestion - as soon as we have types with custom config (like multi-option questions), the form-filling looks different. Not sure if it's worth it to have a helper function for asserting question exists or filling in the question form basics

hmhsu: Yeah, I am also thinking about having a helper just for common fields and each question type can augment as needed. And good idea for an assertion helper as well.

PR #647, opened 2021-04-01
Scripts that need to be in specific directories are `pushd`ed to the right place.

kendrickt:
### Description
Scripts that parsed PWD or BASH_SOURCE now use `pushd` with `git rev-parse --show-toplevel` and end with a `popd`.

A `set -e` was added to `bin/run-dev` to error early because that one has been prone to errors.

Tested by running the scripts from `universal-applicantion-tool_0.0.1/app`.

### Checklist

### Approvals
ALL

### DIFF
--- a/bin/build-and-push
+++ b/bin/build-and-push
@@ -1,4 +1,5 @@
 #! /bin/bash
+pushd $(git rev-parse --show-toplevel)
 
 set -e
 set +x
@@ -14,3 +15,5 @@ ECSSERVICE=$(aws cloudformation describe-stacks --region ${REGION} --stack-name 
 ECSCLUSTER=$(aws cloudformation describe-stacks --region ${REGION} --stack-name civiform | jq -r '.Stacks[0].Outputs[] | select(.OutputKey == "ECSCluster") | .OutputValue')
 
 aws ecs update-service --region=${REGION} --cluster "$ECSCLUSTER" --service "$ECSSERVICE" --force-new-deployment
+
+popd
--- a/bin/localstack/mk-s3
+++ b/bin/localstack/mk-s3
@@ -2,7 +2,7 @@
 pushd $(git rev-parse --show-toplevel)
 
 # find repo root name (currently civiform), also support any names for backward and forward compatibility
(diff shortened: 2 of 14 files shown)


kendrickt:
Is this dependency on git bad?

bin/localstack/mk-s3
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
+pushd $(git rev-parse --show-toplevel)
 
 # find repo root name (currently civiform), also support any names for backward and forward compatibility
-CIVIFORM_DIR=$(pwd | awk -F / '{print $(NF-2)}')
+CIVIFORM_DIR=${PWD##*/}
```
hmhsu, line 5: Suggestion (non-blocking): this is the same as `$(basename $PWD)`

bin/run-dev
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
+pushd $(git rev-parse --show-toplevel)
+
+set -e
 
 # run bin/pull-image
```
hmhsu, line 6: Suggestion (non-blocking): it is clear now the comment is no longer needed.

hmhsu (approved):
Thanks, definitely an improvement. Only one blocking concern: some of the scripts are run by GitHub in our CI/CD workflow. I don't know if this will break it or not. Please confirm with @ndmckinley that this is okay.

bin/build-and-push
```diff
@@ -1,4 +1,5 @@
 #! /bin/bash
+pushd $(git rev-parse --show-toplevel)
```
ndmckinley, line 2: niiiiice.

ndmckinley (approved):
Should work!  When you merge, watch the next PR and see if something goes badly wrong, but ... should work!

bin/run-dev
kendrickt: duh! i can delete these 😁

bin/localstack/mk-s3
kendrickt: done

PR #656, opened 2021-04-02
Move `ApplicantQuestion` inner classes to upper classes

### LINKED ISSUE #489: Refactor inner classes on ApplicantQuestion into their own separate classes
Currently, each question type has its own class on [`ApplicantQuestion`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java) with (for example) logic to check whether the applicant entered invalid information. These classes are getting hard to read, and should be pulled into their own separate files (perhaps in a subdirectory under `services/applicant/questions`)

Notes:
* The inner classes are not static - they share state with `ApplicantQuestion`, so refactoring not be as straightforward as separating code into files
* This will involve reworking implementation and test code to use the new, separate classes

**Done when**
Each ApplicantQuestion type is in its own file

natsid:
### Description
* This PR does not change the behavior of these classes - it simply refactors the old inner classes of `ApplicantQuestion` to be their own upper classes, and puts all of these classes in a new `services.applicant.question` package.
* But it also adds new and more extensive tests! Now we thoroughly test the validation logic of the applicant question type classes.

### Checklist

### Issue(s)
Fixes #489.


carolinedanzi:
FYI I will be working on this while Natalie is OOO

universal-application-tool-0.0.1/test/services/applicant/question/ApplicantQuestionTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    assertThat(question.errorsPresenter().hasTypeSpecificErrors()).isFalse();
+  }
+
+  private EnumSet<QuestionType> types() {
+    return EnumSet.complementOf(EnumSet.of(QuestionType.REPEATER));
+  }
+
+  @Test
+  public void getsExpectedQuestionType() {
```
kendrickt, line 101: (blocking) this could be a parameterized test.

kendrickt (approved):
(non-blocking): the questions used in the tests could go into the `TestQuestionBank` and be shared more widely. That could be a separate PR, but these questions would be great to share across our tests!

PR #667, opened 2021-04-02
Use Ebean history, Fixes #529

### LINKED ISSUE #529: Use Ebean updated and created timestamps for Applicants instead of custom created times
Currently, we are setting a custom `created_time` on Applicant metadata for use in comparing Applicants and merging profiles. We can (and should) instead use the Ebean built-in created or updated times (see [@WhenCreated and @WhenModified](https://ebean.io/docs/features/history))

**Done when**
We no longer set or read `created_time` in JSON and instead use fields on the `Applicant` model
`ApplicantData` does not include a parallel `metadata` JSON field on creation

syntapy:
### Description
Use Ebean implementation of history tables instead of manually entered timestamps

### Checklist
- [x] Ensure correct approach taken & refactorizations are OK

### Issue(s)
Fixes #529


universal-application-tool-0.0.1/app/models/Applicant.java
```diff
@@ -14,12 +17,15 @@
...
 
 /** The ebean mapped class that represents an individual applicant */
+@History
 @Entity
 @Table(name = "applicants")
 public class Applicant extends BaseModel {
   private static final long serialVersionUID = 1L;
   private ApplicantData applicantData;
 
+  @WhenCreated public Instant when_created;
```
hmhsu, line 27: Nit (blocking): please confirm to java variable naming convention and use `whenCreated`.

universal-application-tool-0.0.1/conf/evolutions/default/12.sql
```diff
@@ -0,0 +1,8 @@
+# --- Add answer options for multi-option questions
+# --- !Ups
+
+alter table applicants add when_created timestamp;
+
```
hmhsu, line 5: Suggestion (blocking): would you add a default value for existing entities so that we don't break by null pointer exception?
Something like `update applicants set whenCreated = '<default value here>' where whenCreated is null; ` would do it.

hmhsu (commented):
Looking good with a few minor comments.

universal-application-tool-0.0.1/app/auth/UatProfileData.java
```diff
@@ -18,19 +16,10 @@
...
 
-  public UatProfileData() {
-    this(Clock.system(ZoneId.of("America/Los_Angeles")));
-  }
-
-  public UatProfileData(Clock clock) {
-    super();
-    this.clock = Preconditions.checkNotNull(clock);
-  }
+  public UatProfileData() {}
```
carolinedanzi, line 20: suggestion (blocking): I think we still want to call `super()` here, so that the parent class constructor is called

universal-application-tool-0.0.1/app/auth/UatProfileData.java
```diff
@@ -18,19 +16,10 @@
...
 
-  public UatProfileData(Clock clock, Long accountId) {
-    this(clock);
+  public UatProfileData(Long accountId) {
```
carolinedanzi, line 22: suggestion (blocking): we'll want to either call `this()` (to call the other constructor) or `super()` (since that's all the other constructor does)

universal-application-tool-0.0.1/app/models/Applicant.java
```diff
@@ -14,12 +17,15 @@
 import services.applicant.ApplicantData;
 
 /** The ebean mapped class that represents an individual applicant */
+@History
```
carolinedanzi, line 20: question for @bion: do we want to maintain history for applicants? 

for @syntapy: I may have linked to a misleading doc in the issue, I apologize! I think here we just need the `@WhenCreated` below, but don't need to maintain history on the Applicant table. I'm not sure, though, so I'm looping in Bion

universal-application-tool-0.0.1/test/services/applicant/ApplicantServiceImplTest.java
```diff
@@ -215,6 +216,30 @@ public void createApplicant_createsANewApplicant() {
...
+
+    }
+
+    Applicant applicant = subject.createApplicant(1l).toCompletableFuture().join();
+
+    try {
+      Thread.sleep(5);
+    } catch (InterruptedException e) {
+
+    }
```
carolinedanzi, line 235: question: is it possible to avoid thread sleeping in the test? We could either try to
1. Just compare `Applicant.whenCreated` to make sure it is before `Instant.now()`
2. Create a fake `Clock`, then use it in [`now(clock)`](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#now-java.time.Clock-) to create our own version of now that we can assert on

carolinedanzi (commented):
This is looking really good, nice work! Mostly nit comments, with a question for our team lead

universal-application-tool-0.0.1/app/auth/ProfileFactory.java
```diff
@@ -35,17 +31,18 @@ public UatProfile wrapProfileData(UatProfileData p) {
   }
 
   private UatProfileData create(Roles role) {
-    UatProfileData p = new UatProfileData(clock);
+    // Not sure if this is bad to do
```
carolinedanzi, line 34: This should be good! We want to maintain behavior, which calling the new constructor does (once you add `super()`, see comment below)

universal-application-tool-0.0.1/app/models/Applicant.java
bion: Having timestamps for Applicants is good but we do not need history for applicants. Adding history increases scope for the work to be compliant for data retention policies so it's best to leave it out.

universal-application-tool-0.0.1/conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,9 @@
+# --- Add repeater id reference to questions
+
+# --- !Ups
+
+alter table questions add repeater_id bigint;
```
carolinedanzi, line 5: question: is this just git weirdness? I'm guessing this was added in a different PR

universal-application-tool-0.0.1/conf/evolutions/default/13.sql
```diff
@@ -1,9 +1,12 @@
-# --- Add repeater id reference to questions
-
+# --- Add answer options for multi-option questions
```
carolinedanzi, line 1: suggestion (blocking): please update this comment to describe what this script does

universal-application-tool-0.0.1/conf/evolutions/default/13.sql
```diff
@@ -1,9 +1,12 @@
...
 # --- !Ups
 
-alter table questions add repeater_id bigint;
+alter table applicants add when_created timestamp;
+
+update applicants set when_created = current_timestamp where when_created is null;
+
+update programs set export_definitions = '[]' where export_definitions is null;
```
carolinedanzi, line 8: question: should this be here? or is this from a different script?

carolinedanzi (commented):
This looks really good! A few questions about the .sql files but otherwise you're nearly good to go

universal-application-tool-0.0.1/app/models/Applicant.java
```diff
@@ -20,6 +22,8 @@
   private static final long serialVersionUID = 1L;
   private ApplicantData applicantData;
 
+  @WhenCreated public Instant whenCreated;
```
hmhsu, line 25: Suggestion (blocking): please make it private.

hmhsu (approved):
Looks good to me!

hmhsu:
Looks good! Merging.

PR #674, opened 2021-04-04
Use the new Path.join method to append to a path. (#638)

### LINKED ISSUE #638: Use the new Path.join method to append to a path.
There are many instances of path concatenation that looks like `path.toBuilder().append(something).build()`. There is a `Path.join(something)` method that should be used.

I think the entire `Path.Builder` can be removed. `Path.join` should be used, but may need to be updated because it is using the bulider.

**Additional context**

**Done when**
`Path.Builder` is removed, and `Path.join` is used.

davidrhallartificialeducation:
### Description
Please explain the changes you made here:

* Replaced `path.toBuilder().append(XYZ).build()` with `path.join(XYZ)`.
* All builder append usages were inside `ApplicantServiceImpl`.
* Added `pathJoin()` unit test to `PathTest`.
* Verified passing unit tests for `PathTest` & `ApplicantServiceImplTest`.
* Removed todo documentation above `Path.join()`.

### Checklist

### Issue(s)
Fixes #638.

### DIFF
--- a/universal-application-tool-0.0.1/test/services/PathTest.java
+++ b/universal-application-tool-0.0.1/test/services/PathTest.java
@@ -124,21 +124,6 @@ public class PathTest {
   }
 
   @Test
-  public void pathBuilder() {
-    Path path = Path.builder().setPath("applicant.my.path").build();
-    assertThat(path.path()).isEqualTo("applicant.my.path");
-
-    path = path.toBuilder().append("another").build();
-    assertThat(path.path()).isEqualTo("applicant.my.path.another");
-
-    path = path.toBuilder().append("part").build();
-    assertThat(path.path()).isEqualTo("applicant.my.path.another.part");
-
-    path = path.toBuilder().setPath("something.new").build();
-    assertThat(path.path()).isEqualTo("something.new");
-  }
-
-  @Test
   public void pathJoin() {
     Path path = Path.create("applicant.my.path");
     assertThat(path.path()).isEqualTo("applicant.my.path");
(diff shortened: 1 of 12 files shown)


kendrickt:
I see a lot more uses of `Path#toBuilder()` beyond `ApplicantServiceImpl`:

image

universal-application-tool-0.0.1/test/services/PathTest.java
```diff
@@ -137,4 +137,14 @@ public void pathBuilder() {
     path = path.toBuilder().setPath("something.new").build();
     assertThat(path.path()).isEqualTo("something.new");
   }
+
+  @Test
+  public void pathJoin() {
+    Path path = Path.empty();
+    path = path.join("applicant.my.path");
```
kendrickt, line 144: I think this test case reveals a bug in `Path#join`. The `"applicant.my.path"` segment is added as a single segment to the path.

So 
```
Path path = Path.create("path.start");
Path anotherPath = path.join("a.long.segment");
assertThat(anotherPath.parentPath()).isEqualTo(path);
```
But I think we'd really want `anotherPath.parentPath()` to be `Path.create(path.start.a.long)`

kendrickt: A fix would be to update the word `segment` to `path` in the docstring and implementation, and then fix `join` so that it properly joins paths.

You're welcome to open a separate PR to bugfix that, or I can make it an issue for another time.

davidrhallartificialeducation: I believe I've correctly searched for usages, let me know what you think of the newest commit

kendrickt: This bug was fixed in #680

kendrickt (approved):
Looks good to me. The `Path#Builder` was deleted, and it looks like everything still works.

PR #675, opened 2021-04-05
Checkbox question definition

### LINKED ISSUE #396: "Checkbox button set" question type: implement bare bones
This question type creates a list of checkboxes. Applicants may pick one or more choices, depending on how the admin configured the question.

image

Make it possible for UAT admins to create, and for applicants to fill out, questions of this type. Answers should be saved to a database. Configuration options and validation criteria aren't necessary yet.

See details at https://github.com/seattle-uat/universal-application-tool/wiki/Question-type-requirements-notes

carolinedanzi:
### Description
1. Implement a checkbox question and renderer
2. Enable admins to configure checkbox questions (`CheckboxQuestionForm`)
3. Update applicant data to allow putting single elements in a list, so that scalar path updates with array indices work as expected
    1. Example: `applicant.checkbox.selection[0]: chocolate` will correctly put the element "chocolate" in an array at scalar path `applicant.checkbox.selection`
    2. Note: this may put elements out of order (we don't check scalar array indices). For now, we don't need ordering, but we should revisit in the future if this changes

### Checklist

### Manual testing
1. Created a checkbox question on the admin side and verified editing kept the previous information
2. Filled out a checkbox question on the applicant side
3. Verified that seeing the same checkbox question again as an applicant pulls the previous data they entered and marks those options as checked

### Issue(s)
Fixes #396

Screen Shot 2021-04-05 at 7 17 27 PM
Screen Shot 2021-04-05 at 7 12 30 PM
Screen Shot 2021-04-05 at 7 18 03 PM


kendrickt (approved):
(blocking): the `CheckboxQuestionRenderer` shouldn't have to do path manipulation.

One option is to make the `Path` in the `CheckboxQuestionRenderer` end with "[]".

universal-application-tool-0.0.1/app/services/Path.java
```diff
@@ -175,8 +184,7 @@ private String keyNameWithoutArrayIndex() {
           .replace(matcher.start(ARRAY_SUFFIX_GROUP), matcher.end(ARRAY_SUFFIX_GROUP), "")
           .toString();
     }
-    throw new IllegalStateException(
-        String.format("This path %s does not reference an array element.", this));
+    return keyName();
```
kendrickt, line 187: (blocking): can this be removed if the checkbox question path has "[]" at the end?

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -189,10 +189,14 @@ private void stageUpdates(
       throws UnsupportedScalarTypeException, PathNotInBlockException {
     ImmutableSet.Builder<Path> questionPaths = ImmutableSet.builder();
     for (Update update : updates) {
+      // Check existence of the non-array path, which is how scalars are indexed.
       ScalarType type =
           blockDefinition
-              .getScalarType(update.path())
-              .orElseThrow(() -> new PathNotInBlockException(blockDefinition, update.path()));
+              .getScalarType(update.path().withoutArrayReference())
```
kendrickt, line 195: (blocking): can this be removed if checkbox question path ends with "[]"?

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -126,14 +126,37 @@ private void putNull(Path path) {
...
+      putAt(path, value);
+    }
+  }
+
+  /**
+   * Adds a JSON array at the given path, if it is not there already.
+   *
+   * @param path the path to the new array - must end with the array suffix [] or [index]
+   */
+  private void addArrayIfMissing(Path path) {
```
kendrickt, line 152: (blocking) this should be "put". "add" is for adding to an array.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -126,14 +126,37 @@ private void putNull(Path path) {
...
+    if (!path.isArrayElement()) {
```
kendrickt, line 153: (blocking): this method shouldn't check here. The `path` argument should be used as-is, and it is the caller's responsibility to pass it the right string. So in `put` it should use `putArrayIfMissing(path.withoutArrayReference())`.

This way this method is dumb and just puts lists if missing, to whatever path is passed in.

kendrickt (approved):

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
carolinedanzi: I pushed this into `BlockDefinition` - the path from the form will end with `[index]`, not `[]`, so we still need custom logic to handle this case whether or not the scalar path itself has `[]`

universal-application-tool-0.0.1/app/services/Path.java
carolinedanzi: I kept Path as-is and chose to handle this in `BlockDefinition` instead

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
carolinedanzi: Thanks, good catch! Updated

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
carolinedanzi: Got it! Updated

universal-application-tool-0.0.1/app/services/applicant/ApplicantQuestion.java
```diff
@@ -90,6 +90,10 @@ public TextQuestion getTextQuestion() {
     return new TextQuestion();
   }
 
+  public MultiSelectQuestion getMultiSelectQuestion() {
```
bion, line 93: nit: 'get' here may imply it's a singleton or the same thing each time it's called, recommend renaming to `make` or `create`

carolinedanzi: Ok! I'll call it `createMultiSelectQuestion` - I'm working on the refactor of `ApplicantQuestion` after this so it will be a reminder to name the other methods too

PR #678, opened 2021-04-05
Set all legacy entities' lifecycle_stage to 'active'

hmhsu:
### Description
This is to fix columns in db that have null lifecycle_stage.


hmhsu:
I am trying to find a way to test this locally, but meanwhile would like to get a pre-review to see if I am on the right path.

ndmckinley:
I think this is right / will work.  You can use `bin/deploy-prod` locally to see if it fixes it in prod.

hmhsu:
> I think this is right / will work. You can use `bin/deploy-prod` locally to see if it fixes it in prod.

Deployed and see it works in prod. Assume we want to check this in, right?

ndmckinley (approved):
Yes please!

Issue #468, opened 2021-03-12
`support.ProgramBuilder` can be used without a database.


kendrickt:
**Is your feature request related to a problem? Please describe.**
Currently, `support.ProgramBuilder` can only be used in tests with a database in scope because of `save` and `update`. It would be nice to gracefully handle `save/update` problems when there is no database to save to, and still return a properly formed `ProgramDefinition` for tests that don't have a database.

**Describe the solution you'd like**
`support.ProgramBuilder` methods will still work even if there is no database available. This may require a way to properly set program `id` values.

**Additional context**
Very similar to #467 

**Done when**
A test that does not use `withPostgresContainer` can use `support.ProgramBuilder` to make programs for testing.

PR #679, opened 2021-04-05
Make test support classes `ProgramBuilder` and `TestQuestionBank` available without postgres.

### LINKED ISSUE #467: `support.TestQuestionBank` can be used without a database
Currently, [`support.TestQuestionBank`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/test/support/TestQuestionBank.java) can only be used in tests with a database in scope because of `save`. It would be nice to gracefully handle `save` problems when there is no database to save to, and still return a properly formed `QuestionDefinition` for tests that don't have a database.

`support.TestQuestionBank` methods will still work even if there is no database available (tests that don't extend [`WithPostgresContainer`](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/test/repository/WithPostgresContainer.java)). This would require a way to set question `id` values.

**Additional context**
Very similar to #468 

**Done when**
A test that does not use `WithPostgresContainer` can use `support.TestQuestionBank` to make questions for testing.

kendrickt:
### Description
Add `maybeSave` methods to `ProgramBuilder` and `TestQuestionBank`. Then use these test helpers in some tests that do not use `WithPostgresContainer`:
1. `CsvExporterTest`
2. `BlockTest`
3. `ReadOnlyQuestionServiceImplTest`

### Checklist

### Issue(s)
Fixes #467 #468


ndmckinley (approved):

carolinedanzi (approved):
This looks like a nice improvement, thank you!

ndmckinley (approved):

PR #680, opened 2021-04-05
Added unit testing for pathJoin & fixed path.join segments

davidrhallartificialeducation:
### Description
Added unit testing for pathJoin & fixed path.join segments

### Checklist


kendrickt (approved):
Paired with David, and it looks good to me.

PR #682, opened 2021-04-06
Concurrent browser tests

hmhsu:
### Description
- Add helper functions for browser tests
- Make browser tests concurrent


browser-test/src/application_csv_download.test.ts
```diff
@@ -1,11 +1,15 @@
...
 describe('normal application flow', () => {
-  // If this times out, a likely cause is the .click() calls in
-  // support/admin_programs, which are called out as being asserts.
-  jest.setTimeout(25000);
+  // Our browser tests could be long-running (as of 2021-04-06, the longest
+  // takes 27 seconds). However, if your seletor fails to locate the HTML
+  // element, the test hangs as well. If you find the tests take unnaturally
+  // long, you want to verify that your selectors are working as expected.
+  // Because all tests are run concurrently, it could be that your selector
+  // selects a different entity from another test.
```
ndmckinley, line 9: FYI: This problem is well-addressed by setDefaultTimeout, which makes a hanging `.click` operation time out after 1 second.  You can probably set this timeout much lower.

ndmckinley (approved):
Spectacular - does this give us what we need to delete some of the old browser tests in java?

hmhsu:
> Spectacular - does this give us what we need to delete some of the old browser tests in java?

I plan to have another PR adding more coverage before removing Java tests, probably itself in yet another PR.

browser-test/src/application_csv_download.test.ts
hmhsu: Good point, will remove the comment. I think the timeout includes the time waiting for the page to load, so let's leave it as 1 second for now.

hmhsu: Move the comment to above setDefaultTimeout.

browser-test/src/application_csv_download.test.ts
```diff
@@ -1,35 +1,40 @@
 import { startSession, logout, loginAsGuest, loginAsAdmin, ApplicantQuestions, AdminQuestions, AdminPrograms, endSession } from './support'
 
 describe('normal application flow', () => {
-  // If this times out, a likely cause is the .click() calls in
-  // support/admin_programs, which are called out as being asserts.
-  jest.setTimeout(25000);
+  // Our browser tests could be long-running (as of 2021-04-06, the longest
+  // takes 27 seconds).
+  jest.setTimeout(200000);
```
bion, line 6: Question (blocking): can we set this globally rather than per-test file if we need it? Also can we set it to a lower value maybe just double the longest test?

bion (approved):

browser-test/src/application_csv_download.test.ts
hmhsu: Good call.

hmhsu: Done.

PR #685, opened 2021-04-06
Enable debugging in bin/run-dev.

ndmckinley:
### Description
Turns on listening for the debug port, but continues if no one connects to debug.  Instructions for setting up the debugger are at https://github.com/seattle-uat/civiform/wiki/Java-Debugging.

### Checklist

### Issue(s)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,6 +28,7 @@ services:
       - "db:database"
     ports:
       - 9000:9000
+      - 9999:9999
     environment:
       - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-test}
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
       - "db:database"
     ports:
       - 9000:9000
+      - 9999:9999
```
hmhsu, line 31: Thoughts (blocking): this will conflict with our browser testing env, could we choose a different port?

hmhsu (approved):

docker-compose.yml
ndmckinley: Ah shoot, sure thing.

bion: It's a popular port

PR #686, opened 2021-04-06
Set existing programs to have empty export

hmhsu:
### Description
Retrofit db column for existing programs.

### DIFF
--- a/universal-application-tool-0.0.1/conf/evolutions/default/12.sql
+++ b/universal-application-tool-0.0.1/conf/evolutions/default/12.sql
@@ -4,6 +4,8 @@
 
 alter table programs add export_definitions jsonb;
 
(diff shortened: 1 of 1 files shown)


hmhsu:
I wonder if this is only for testing purpose, do we need to persist it in db?

kendrickt:
@ndmckinley should the export definitions be persisted? they weren't in the DB before but it seems like they should be. Or are they transiently defined for each request from the admin UI?

ndmckinley (approved):

ndmckinley:
It's worth persisting them, but they're not currently configurable - I had intended to persist them in the PR where I made them configurable.

hmhsu:
SG. Thanks for the answers. Merging this.

PR #690, opened 2021-04-07
Enable the admin button in staging (for probers, when we have them).

ndmckinley:
### Description
This enables the admin button in staging.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/FakeAdminClient.java
+++ b/universal-application-tool-0.0.1/app/auth/FakeAdminClient.java
@@ -13,12 +13,17 @@ public class FakeAdminClient extends IndirectClient {
 
   public static final String CLIENT_NAME = "FakeAdminClient";
 
-  public static final ImmutableSet<String> ACCEPTED_LOCALHOSTS =
-      ImmutableSet.of("localhost", "civiform");
+  public static final ImmutableSet<String> ACCEPTED_HOSTS =
+      ImmutableSet.of("localhost", "civiform", "staging.seattle.civiform.com");
 
   private ProfileFactory profileFactory;
 
-  @Inject
+  public static boolean canEnable(String host) {
+      return FakeAdminClient.ACCEPTED_HOSTS.stream()
+              .anyMatch(acceptedHost -> host.equals(acceptedHost) || host.startsWith(acceptedHost + ":"));
+  }
+
+    @Inject
   public FakeAdminClient(ProfileFactory profileFactory) {
     this.profileFactory = Preconditions.checkNotNull(profileFactory);
   }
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -186,7 +186,7 @@ protected Config provideConfig(
     if (adClient != null) {
       clientList.add(adClient);
     }
-    if (URI.create(this.baseUrl).getHost().equals("localhost")) {
+    if (FakeAdminClient.canEnable(URI.create(baseUrl).getHost())) {
```
hmhsu, line 189: Thoughts (non-blocking): slightly surprised this is not enabled in our browser tests docker environment so far. What does it do? Why our browser tests don't seem to have problems with it?

hmhsu (approved):

hmhsu:
The browser test failure is annoying. It feels flaky to me (we don't get the same result every time and I've seen this failure and others before).

universal-application-tool-0.0.1/app/modules/SecurityModule.java
ndmckinley: I think it is enabled there - I think that's also `localhost`.

ndmckinley:
Ah, thanks!  I spent a lot of time trying to find the bug, haha.  Let me know when you're good to go there.

ndmckinley:
Yeah, these tests sometimes pass on my machine, so I'll rerun this until they succeed I suppose.

hmhsu:
#693 should fix the browser tests.

universal-application-tool-0.0.1/app/modules/SecurityModule.java
hmhsu: I see. `baseUrl` is from config not the request so it is still `localhost`. Make sense.

universal-application-tool-0.0.1/app/auth/FakeAdminClient.java
```diff
@@ -13,11 +13,16 @@
...
 
-  public static final ImmutableSet<String> ACCEPTED_LOCALHOSTS =
-      ImmutableSet.of("localhost", "civiform");
+  public static final ImmutableSet<String> ACCEPTED_HOSTS =
+      ImmutableSet.of("localhost", "civiform", "staging.seattle.civiform.com");
 
   private ProfileFactory profileFactory;
 
+  public static boolean canEnable(String host) {
+    return FakeAdminClient.ACCEPTED_HOSTS.stream()
```
kendrickt, line 22: (somewhat related comment) TIL `ImmutableSet` uses `AbstractCollection.contains` and just iterates.

PR #693, opened 2021-04-07
Deflake browser tests

hmhsu:
### Description
- Close browser after test
- Compare label string case insensitively when their style change case
- Click the right button to add questions to a program
- Fix a bug in awaiting

### DIFF
--- a/browser-test/src/admin_program_creation.test.ts
+++ b/browser-test/src/admin_program_creation.test.ts
@@ -1,8 +1,8 @@
-import { startSession, loginAsAdmin, AdminQuestions } from './support'
+import { startSession, loginAsAdmin, AdminQuestions, endSession } from './support'
 
 describe('create dropdown question with options', () => {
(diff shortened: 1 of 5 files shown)


ndmckinley (approved):
Wow, nice catches, thank you!

hmhsu:
That said, the tests time out, haha.

kendrickt:
(non-blocking) @hmhsu I don't think this is addressed by these changes: https://github.com/seattle-uat/civiform/pull/690/checks?check_run_id=2289905968

is it in scope?

hmhsu:
> (non-blocking) @hmhsu I don't think this is addressed by these changes: https://github.com/seattle-uat/civiform/pull/690/checks?check_run_id=2289905968
> 
> is it in scope?

Yes! There is a bug that we didn't wait for a function that asserts the question state so the next navigation creates a race condition that could caused the assertion to fail, which is what you see in the failure.

PR #701, opened 2021-04-08
Issue # 606

### LINKED ISSUE #606: Fix all usages of `services.Path.path()`.
A `Path.path()` method doesn't make sense. What is the path of a path?

In some places, what it really means is `Path.toString()`. e.g.

```
String.format("question %s updated", questionForm.getQuestionPath().path())
```

In other places, it just means `Path` but the string version of it. This is code debt because our classes should be using Path objects and not strings representing paths. 

Our application should use Path objects and not strings representing paths. The `Path.toString()` method may be necessary when interfacing with json data, and presenting to the UI. But otherwise, most places should be using `Path` objects.

In some cases (e.g. exception string formatting) the path object can be used as is, and the `toString()` method will be invoked automatically.

Cases:
1. JSON methods: use `toString()` explicitly.
2. String formatting: just use `Path` and let `toString()` be automatically invoked
3. Other (I don't think this exists): update our methods to take `Path` objects instead of strings representing paths.

**Additional context**

The most complicated behavior is in `QuestionRepository.PathConflictDetector`. For now, this can just use `toString()` and will be addressed in a separate PR.

**Done when**
`Path#path()` is deleted.

randomUser00:
### Description
I have changed the Path.path( ) to path.toString ( )

### Checklist

### Issue(s)
Fixes #606

### DIFF
--- a/universal-application-tool-0.0.1/app/services/program/PathNotInBlockException.java
+++ b/universal-application-tool-0.0.1/app/services/program/PathNotInBlockException.java
@@ -9,6 +9,6 @@ public class PathNotInBlockException extends Exception {
   public PathNotInBlockException(BlockDefinition block, Path path) {
     super(
         String.format(
-            "Block (ID %d) in program (ID ?) does not contain path %s", block.id(), path.path()));
+            "Block (ID %d) in program (ID ?) does not contain path %s", block.id(), path.toString()));
   }
 }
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/services/program/PathNotInBlockException.java
```diff
@@ -9,6 +9,6 @@
   public PathNotInBlockException(BlockDefinition block, Path path) {
     super(
         String.format(
-            "Block (ID %d) in program (ID ?) does not contain path %s", block.id(), path.path()));
+            "Block (ID %d) in program (ID ?) does not contain path %s", block.id(), path.toString()));
```
kendrickt, line 12: (blocking): the argument can just be `path`, and the `toString()` will be implicitly called by `String.format`.

kendrickt (commented):
(blocking) I see 42 usages of `Path.path()` and that doesn't match these changes. 

Also, `Path.path()` should be deleted. Please see the **Done when** section of #606

PR #703, opened 2021-04-08
Fix input label name for checkboxes

carolinedanzi:
### Description
We need the format to be `path.selection[]`, not `path.selection.[]`, which is what I accidentally did with my last change


hmhsu (approved):

kendrickt (approved):
Thanks for adding a test!

PR #722, opened 2021-04-08
Allow presigned url access to s3

hmhsu:
### Description
Only allow get object from s3 through our task role.

### Checklist

### DIFF
--- a/infra/filestorage.yaml
+++ b/infra/filestorage.yaml
@@ -70,7 +70,7 @@ Resources:
       Bucket: !Ref bucket
       PolicyDocument:
         Statement:
-          - Sid: Restrict access to VPC endpoint
+          - Sid: Restrict access to VPC endpoint only
             Action:
               - 's3:*'
             Effect: Deny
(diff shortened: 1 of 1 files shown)


infra/filestorage.yaml
```diff
@@ -79,10 +79,23 @@ Resources:
               - - 'arn:aws:s3:::'
                 - !Ref bucket
                 - /*
-            Principal: '*'
+            NotPrincipal:
+              AWS: 'arn:aws:iam::405662711265:role/AmazonECSTaskS3BucketRole'
```
hmhsu, line 83: Thoughts: this makes me realize our staging roles have the same permission as prod roles, although they not restricted by resource related permissions so might be okay.

ndmckinley: Ah, so we could reach the prod bucket from staging?  That would be worth fixing, haha.  :)  If that's true, we should fix it - if you think it might be true, can you make a github issue?  I had thought this role was restricted to the bucket in the same stack, but I don't know if that's right.

ndmckinley (approved):

infra/filestorage.yaml
hmhsu: I will follow up on this. Merging this PR.

PR #738, opened 2021-04-08
Add ReadOnlyQuestionService#getRepeaterQuestions()

bion:


universal-application-tool-0.0.1/test/support/TestQuestionBank.java
```diff
@@ -107,6 +106,21 @@ private static Question applicantFavoriteColor(QuestionEnum ignore) {
     return maybeSave(definition);
   }
 
+  private static Question applicantHouseholdMembers(QuestionEnum ignore) {
+    QuestionDefinition definition =
+        new RepeaterQuestionDefinition(
+            VERSION,
+            "applicant household members",
+            Path.create("applicant.household_members"),
```
kendrickt, line 114: (blocking, maybe?) I _think_ the repeater question path will need to include square brackets (`applicant.household_members[]`). This is yet another reason why `Path` shouldn't be a constructor argument. 

We may end up not needing the square brackets but they're being added to repeater questions, so this question should probably have it to match the real repeater questions.

hmhsu (approved):
Thanks for adding tests!

kendrickt (approved):
lgtm

universal-application-tool-0.0.1/test/support/TestQuestionBank.java
bion: Done.

Issue #742, opened 2021-04-09
Production checklist


bion:
Technical work needed before allowing applicant traffic to production.

- [x] server logs are sent to city of Seattle centralized logging service
- [x] deployment configures an appropriate number of server threads based on (v)CPU threads
- [x] automated database backups are configured
- [ ] the site offers some means of contacting technical support
- [x] non-city employees are unable to access the production database
- [x] basic traffic monitoring is available -- QPS, error ratios split out by 4xx and 5xx level errors
- [x] basic probers run in production (it may be that probers that mutate state are infeasible in prod)
- [x] domain and https cert are configured by the city.
- [x] complete security tasks (#363, #364, #365)
- [x] set up google analytics (#574, #575)
- [x] redirect http requests to https
- [ ] redirect `www.` to the bare domain or vice-versa

neel-a-mehta:
Are the security-related topics (#363, #364, #365) fair game for this too?

lianne-oy:
the site offers some means of contacting technical support - not needed until Phase 2 (public links working) - using pilot escalation path in phase 1 https://seattlegov.sharepoint.com/:w:/r/sites/IAC-EXT/AP/_layouts/15/Doc.aspx?sourcedoc=%7BF5ED3858-CD8F-4C7A-839B-FE6FB45CF8AB%7D&file=User%20Issue%20Escalation%20Path%20for%20CiviForm%20Pilot.docx&action=default&mobileredirect=true

automated database backups are configured - Still need to check

lianne-oy:
redirect www. to the bare domain or vice-versa - NICE TO HAVE; NOT NEEDED
the site offers some means of contacting technical support - STILL NEED.  Configuration setting for email address at bottom page / footer.  Need email address from City.

bion:
Closing this with one work item remaining and captured here: https://github.com/seattle-uat/civiform/issues/1308

PR #743, opened 2021-04-09
Only use https.

ndmckinley:
### Description
Remove the http path on port 80, leaving only https on port 443.

### Checklist

### Issue(s)
Related to #742.

### DIFF
--- a/bin/check-cfn-change
+++ b/bin/check-cfn-change
@@ -29,6 +29,10 @@ if aws cloudformation describe-change-set --region ${REGION}  --change-set-name 
 	sleep 5
 fi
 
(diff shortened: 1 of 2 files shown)


ndmckinley:
(failure unrelated - 500 over at maven's download website - will rerun)

hmhsu (approved):
Do we need to redirect the request or does it already handled by LB? Or do modern browsers auto-retry on https?

ndmckinley:
I _think_ this will be handled by Amazon but I will check!  :)

ndmckinley:
Thanks - good catch - it would have worked for modern browsers but older ones like to have the explicit redirect, which amazon does not add for you automatically.

PR #750, opened 2021-04-10
S3 retry logic. Fixes #215

### LINKED ISSUE #215: Add retry logic to AWS S3 when IO operations fail
The current behavior is that once an IO operation fails, the application throws a runtime error. We would like to at least retry a few times before we give up with exponential backup between retries.

**Done when**
This is done when S3 IO exceptions are properly handled with at least 3 retries.

syntapy:
### Description
Retry S3 connection 3x, each with larger `Thread.sleep` wait time, or until successful

### Checklist
- [x] Add a test making sure it works correctly on a failure

### Issue(s)
Fixes #215

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
+++ b/universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
@@ -117,7 +117,23 @@ public class AmazonS3Client {
     if (s3 != null) {
       return;
     }
-    connect();
+
+    int i = 0;
+    while (s3 == null && i < 3) {
+        connect();
+
+        if (s3 == null) {
+          log.warn("Failed to create S3 client on attempt "+i);
+          try {
+            Thread.sleep(50*(int) Math.pow(2, i));
+          } catch(InterruptedException e) {
+            log.info("Interrupt on Thread.sleep");
+          }
+        }
+
+        i++;
+    }
+
     if (s3 == null) {
       throw new RuntimeException("Failed to create S3 client");
     }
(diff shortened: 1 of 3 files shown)


hmhsu:
Hey, thanks for your interest in this issue. My apology on not being clear in the issue. I think we want to retry when the IO operations fail with S3, specifically on line 84 and 97 in AmazonS3Client.java.

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -117,10 +124,33 @@ private void ensureS3Client() {
...
-    connect();
+
+    int i = 0;
+    while (i < 3) {
+      connectS3();
+      connectPresigner();
+
+      if (s3 == null || presigner == null) {
+        try {
+          Thread.sleep(50 * (int) Math.pow(2, i));
```
hmhsu, line 135: Praise: Good job using exponential backoff.

universal-application-tool-0.0.1/app/repository/AmazonS3Client.java
```diff
@@ -117,10 +124,33 @@ private void ensureS3Client() {
     if (s3 != null) {
       return;
     }
-    connect();
+
+    int i = 0;
+    while (i < 3) {
```
hmhsu, line 129: Nit: This is equivalent to a for loop.

hmhsu:
Hey, are you still working on this PR? If not, I plan to close it this week.

hmhsu:
The change looks good, but we decided to go with a different approach with file upload. Now users upload files directly to S3 instead of proxying through CiviForm server. So this work is no longer necessary. Thanks again for your effort into this. I really appreciate it.

PR #758, opened 2021-04-12
Do not allow us to automatically apply db DOWNs.

ndmckinley:
### Description
Like it says in the comment - don't auto apply database downs (so we can't dump the prod db anymore).

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/conf/application.conf
+++ b/universal-application-tool-0.0.1/conf/application.conf
@@ -306,9 +306,7 @@ play.evolutions {
   db.default {
     useLocks = true
     autoApply = true
-    # This might break something in the prod database someday.  Let's remove it once we completely agree
-    # never to require any more evolution downs.
-    autoApplyDowns = true
+    autoApplyDowns = false
   }
 }
 


universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -306,9 +306,7 @@ play.evolutions {
   db.default {
     useLocks = true
     autoApply = true
-    # This might break something in the prod database someday.  Let's remove it once we completely agree
-    # never to require any more evolution downs.
-    autoApplyDowns = true
+    autoApplyDowns = false
```
hmhsu, line 309: Question (non-blocking): in what scenario would downs be applied when this is set to true? Could it happen when a server starts or does it only happen when the servers are shutdown?

hmhsu (approved):

universal-application-tool-0.0.1/conf/application.conf
hmhsu: Read the doc. I see. the evolutions are applied when the scripts are changed.

bion (approved):

universal-application-tool-0.0.1/conf/application.conf
hmhsu: I *think* some of the current ups are not idempotent though, like the ones that add a column.

ndmckinley: Hm, that's a good point - what's the equivalent to "if not exists" for adding a column.

ndmckinley: (It's not a big deal since we use evolution locks, but it would be nice to know!)

hmhsu: I *think* you can do

```sql
ALTER TABLE table_name ADD [COLUMN] IF NOT EXISTS column_name data_type;
```

ndmckinley: Ah, good eye!  I would do it to the old evolutions but as of this PR, we can't do that anymore, ha.

Issue #377, opened 2021-03-08
UAT admins can specify validation rules for questions


neel-a-mehta:
This doesn't mean that each individual validation rule has to be built out. It just means that UAT admins should have the ability to specify validation rules in general.

natsid:
Remaining (but in-progress) dev tasks for this: #589 and #590.

natsid:
This story is complete in a general respect.

That is, an admin is able to specify a handful of custom validation configs (min and max length for text questions, min and max value for number questions, and disallowing PO boxes for address questions).

natsid:
It is now relatively straightforward to add further validation configs (and other type-specific configs) for the types mentioned in the last comment, as well as all the other existing types.

lianne-oy:
This issue is closed; bugs will be added to future Milestone.

PR #761, opened 2021-04-13
Replace validation error messages with official copy

### LINKED ISSUE #634: `hasValue` in the `ApplicantQuestion` subclasses don't seem to WAI
We decided on a convention of storing `null` for "never answered" pieces of data, and `""` (the empty string) for "answered with a blank value".

However, it seems like the underlying spring form data binder binds empty inputs as `null` (at least that's what I guess is happening).

I ran into this issue while implementing min and max length validations for text questions. When there is no value entered in the applicant form, our code evaluates `hasValue()` to `false`, so we end up not running our `getQuestionErrors` validations.

Needs more investigation to conclude the root cause and obviously to fix the issue as well.

natsid:
### Description
* Replace the placeholder validation error messages with the official copy from [this doc](https://docs.google.com/document/d/1chNdJViEqadU8pBMWXAXc4dOmsIrplDC6vva0ScoSXM/edit?usp=sharing).
* Move the address question validation error messages to the `ValidationErrorMessage` class.
* Replace `hasValue` with `isAnswered` in `TextQuestion` (bug fix; will do the same fix for `NumberQuestion` and any other types that need it in a follow-up PR to fully fix #634).

### Checklist

### Issue(s)
Polish work toward #377 
Works toward #634


universal-application-tool-0.0.1/test/services/applicant/question/AddressQuestionTest.java
```diff
@@ -95,21 +95,22 @@ public void withInvalidApplicantData_missingRequiredFields() {
 
     assertThat(addressQuestion.hasTypeSpecificErrors()).isTrue();
     assertThat(addressQuestion.getStreetErrors())
-        .contains(ValidationErrorMessage.create("Street is required."));
+        .contains(ValidationErrorMessage.create("Please enter valid street name and number."));
```
carolinedanzi, line 98: question: Should we use the message constants here too?

universal-application-tool-0.0.1/test/services/applicant/question/AddressQuestionTest.java
```diff
@@ -95,21 +95,22 @@ public void withInvalidApplicantData_missingRequiredFields() {
...
 
   @Test
-  public void withInvalidApplicantData_invalidZipCode() {
+  @Parameters({"not a zip code", "123456789", "123ab"})
+  public void withInvalidApplicantData_invalidZipCode(String zipValue) {
     applicantData.putString(addressQuestionDefinition.getStreetPath(), "123 A St");
     applicantData.putString(addressQuestionDefinition.getCityPath(), "Seattle");
     applicantData.putString(addressQuestionDefinition.getStatePath(), "WA");
-    applicantData.putString(addressQuestionDefinition.getZipPath(), "not a zip code");
+    applicantData.putString(addressQuestionDefinition.getZipPath(), zipValue);
```
carolinedanzi, line 113: praise: good catch!

universal-application-tool-0.0.1/test/services/applicant/question/TextQuestionTest.java
```diff
@@ -93,9 +93,9 @@ public void withMinAndMaxLength_withValidApplicantData_passesValidation(String v
   @Parameters({
     // TODO(https://github.com/seattle-uat/civiform/issues/634): Uncomment the following line once
     //  this bug is fixed.
```
carolinedanzi, line 95: suggestion (blocking): please remove comment if no longer applicable

carolinedanzi (approved):
This looks nice!

universal-application-tool-0.0.1/test/services/applicant/question/AddressQuestionTest.java
natsid: Good call!

universal-application-tool-0.0.1/test/services/applicant/question/TextQuestionTest.java
natsid: yay - thanks for catching this!

PR #762, opened 2021-04-13
Change how path conflicts are caught, and add repeater id and path to question invariants.

kendrickt:
### Description
The `String.startsWith` cannot be used with repeated questions.

Question invariants are now: name, path, repeater id, and question type.

Updated the `TestQuestionBank` paths to what they should be based on their names.

Also got rid of an unused `Request` argument.

### DIFF
--- a/universal-application-tool-0.0.1/app/services/Path.java
+++ b/universal-application-tool-0.0.1/app/services/Path.java
@@ -212,7 +212,7 @@ public abstract class Path {
   private String stripArraySuffix(String segment, boolean strict) {
     Matcher matcher = ARRAY_INDEX_REGEX.matcher(segment);
     if (matcher.matches()) {
-      return new StringBuilder(keyName())
+      return new StringBuilder(segment)
           .replace(matcher.start(ARRAY_SUFFIX_GROUP), matcher.end(ARRAY_SUFFIX_GROUP), "")
           .toString();
     }
--- a/universal-application-tool-0.0.1/app/services/Path.java
+++ b/universal-application-tool-0.0.1/app/services/Path.java
@@ -198,7 +198,7 @@ public abstract class Path {
    * <p>For paths to non-array elements, {@code IllegalStateException} is thrown.
    */
   private String keyNameWithoutArrayIndex() {
-    return stripArraySuffix(keyName(), true);
+    return stripArraySuffix(keyName(), /* strict= */ true);
   }
 
   /**
--- a/universal-application-tool-0.0.1/app/services/Path.java
+++ b/universal-application-tool-0.0.1/app/services/Path.java
@@ -227,6 +227,6 @@ public abstract class Path {
 
   /** Non-strict version of {@link #stripArraySuffix(String, boolean)} */
   private String stripArraySuffix(String segment) {
-    return stripArraySuffix(segment, false);
+    return stripArraySuffix(segment, /* strict= */ false);
   }
 }
--- a/universal-application-tool-0.0.1/app/services/Path.java
+++ b/universal-application-tool-0.0.1/app/services/Path.java
@@ -167,19 +167,66 @@ public abstract class Path {
   }
 
   /**
+   * Returns true if this path starts with the other path, ignoring array index suffixes. that is,
+   * "a.b[].c[].d" starts with "a.b.c".
+   */
+  public boolean startsWith(Path other) {
+    ImmutableList<String> thisSegments =
+        segments().stream().map(this::stripArraySuffix).collect(ImmutableList.toImmutableList());
+    ImmutableList<String> otherSegments =
+        other.segments().stream()
+            .map(this::stripArraySuffix)
+            .collect(ImmutableList.toImmutableList());
+
+    // This can't start with something that is longer than it.
+    if (otherSegments.size() > thisSegments.size()) {
+      return false;
+    }
+
+    for (int i = 0; i < otherSegments.size(); i++) {
+      if (!thisSegments.get(i).equals(otherSegments.get(i))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
    * Returns the path's key name without an array index suffix. e.g. {@code a.b[1].c[3]} returns
    * "c".
    *
    * <p>For paths to non-array elements, {@code IllegalStateException} is thrown.
    */
   private String keyNameWithoutArrayIndex() {
-    Matcher matcher = ARRAY_INDEX_REGEX.matcher(keyName());
+    return stripArraySuffix(keyName(), true);
+  }
+
+  /**
+   * Returns the path segment without an {@link #ARRAY_SUFFIX}.
+   *
+   * @param segment a path segment to strip of {@link #ARRAY_SUFFIX}
+   * @param strict if true, throws an {@link IllegalArgumentException} if segment does not have
+   *     anything to strip
+   * @return segment, without an {@link #ARRAY_SUFFIX}
+   */
+  private String stripArraySuffix(String segment, boolean strict) {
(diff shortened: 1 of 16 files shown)


bion (changes requested):
> Paths only conflict if they are exactly the same

I don't think this is true. If a new path is prefixed with the path of an existing scalar or parent, it is invalid. E.g.

```python
# existing question
"applicant.name"

# which implies
"applicant.name.first_name"
"applicant.name.middle_name"
"applicant.name.last_name"

# making all of these invalid for new paths
"applicant.name"
"applicant.name.foo"
"applicant.name.first_name"
"applicant.name.first_name.foo"
```

kendrickt:
@bion I can add those back in and ignore REPEATER questions since they'll be prefixes for questions.

In theory, because we're generating the paths from the path of a repeater and a formatted name string, we'll never have conflicts, but we can keep the checks.

hmhsu:
> In theory, because we're generating the paths from the path of a repeater and a formatted name string, we'll never have conflicts, but we can keep the checks.

I think we need to check to prevent admin to create two questions of the same name. I think you are right that we just need to add an exception for repeater type.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -99,43 +106,64 @@ public boolean addTranslation(
...
+                  questionDefinition.getPath(), conflict.id)));
     }
     return ImmutableSet.of();
   }
 
+  /**
+   * Validates that a question's updates do not change its invariants.
+   *
+   * <p>Question invariants are: name, repeater id, path, and type.
+   */
```
bion, line 131: Suggestion (blocking): I think 'invariant' is a more general / somewhat different term than the way it's being used here. Recommend calling these 'immutable state' or 'immutable members' instead

universal-application-tool-0.0.1/app/services/Path.java
```diff
@@ -166,20 +166,67 @@ public Path atIndex(int index) {
...
+
   /**
    * Returns the path's key name without an array index suffix. e.g. {@code a.b[1].c[3]} returns
    * "c".
    *
    * <p>For paths to non-array elements, {@code IllegalStateException} is thrown.
    */
   private String keyNameWithoutArrayIndex() {
-    Matcher matcher = ARRAY_INDEX_REGEX.matcher(keyName());
+    return stripArraySuffix(keyName(), true);
```
bion, line 201: Suggestion (non-blocking):
```suggestion
    return stripArraySuffix(keyName(), /* strict= */ true);
```

universal-application-tool-0.0.1/app/services/Path.java
```diff
@@ -166,20 +166,67 @@ public Path atIndex(int index) {
...
+      throw new IllegalStateException(
+          String.format("This path %s does not reference an array element.", this));
+    }
+
+    return segment;
+  }
+
+  /** Non-strict version of {@link #stripArraySuffix(String, boolean)} */
+  private String stripArraySuffix(String segment) {
+    return stripArraySuffix(segment, false);
```
bion, line 230: Suggestion (non-blocking): 
```suggestion
    return stripArraySuffix(segment, /* strict= */ false);
```

universal-application-tool-0.0.1/app/services/question/ReadOnlyQuestionService.java
```diff
@@ -56,7 +56,10 @@
    */
   QuestionDefinition getQuestionDefinition(long id) throws QuestionNotFoundException;
 
-  /** Checks whether a specific path is valid. */
+  /**
+   * Checks whether a specific path is valid. A path is valid in this context if it represents a
+   * real path to a value.
+   */
```
bion, line 62: Thanks for adding docs!

bion (approved):

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
kendrickt: changed to immutable members

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -99,43 +106,64 @@ public boolean addTranslation(
...
+        questionRepository.findPathConflictingQuestion(questionDefinition);
     if (maybeConflict.isPresent()) {
-      Question question = maybeConflict.get();
+      Question conflict = maybeConflict.get();
       return ImmutableSet.of(
           CiviFormError.of(
               String.format(
-                  "path '%s' conflicts with question: %s", newPath.path(), question.getPath())));
+                  "path '%s' conflicts with question id: %s",
+                  questionDefinition.getPath(), conflict.id)));
```
hmhsu, line 122: Suggestion (non-blocking): I feel path is more readable than an id as this error message is shown to users. Also I am not completely sure integer works with %s.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -99,43 +106,64 @@ public boolean addTranslation(
...
+              String.format(
+                  "question names mismatch: %s does not match %s",
+                  questionDefinition.getName(), toUpdate.getName())));
+    }
+
+    if (!questionDefinition.getRepeaterId().equals(toUpdate.getRepeaterId())) {
+      errors.add(
+          CiviFormError.of(
+              String.format(
+                  "question repeater ids mismatch: %s does not match %s",
```
hmhsu, line 148: Nit (blocking): ids -> id.

universal-application-tool-0.0.1/test/controllers/admin/QuestionControllerTest.java
```diff
@@ -183,18 +185,22 @@ public void newOne_returnsFailureForInvalidQuestionType() {
...
+        .put("questionText", "question text updated")
+        .put("questionHelpText", "a new help text");
     RequestBuilder requestBuilder = addCSRFToken(Helpers.fakeRequest().bodyForm(formData.build()));
 
-    Result result = controller.update(requestBuilder.build(), question.id, "text");
+    Result result =
+        controller.update(
+            requestBuilder.build(),
+            nameQuestion.getId(),
+            nameQuestion.getQuestionType().toString());
```
hmhsu, line 203: Question (non-blocking): what's the difference between `#name` and `#toString`? Should we have only one?

universal-application-tool-0.0.1/test/support/TestQuestionBank.java
```diff
@@ -121,6 +127,22 @@ private static Question applicantHouseholdMembers(QuestionEnum ignore) {
     return maybeSave(definition);
   }
 
+  private static Question applicantHouseholdMemberName(QuestionEnum ignore) {
+    Question householdMembers = applicantHouseholdMembers();
+    QuestionDefinition definition =
+        new NameQuestionDefinition(
+            VERSION,
+            "household members name",
```
hmhsu, line 135: Nit: add applicant in the name to make it consistent with path.

universal-application-tool-0.0.1/test/services/question/ReadOnlyQuestionServiceImplTest.java
```diff
@@ -133,16 +135,16 @@ public void isValid_returnsFalseForInvalid() {
 
   @Test
   public void isValid_returnsFalseWhenEmpty() {
```
hmhsu, line 137: Nit: This probably could have a better name `returnsFalseWhenNonExistent`

hmhsu (commented):
LGTM

PR #768, opened 2021-04-13
Support localizations for program name and description

### LINKED ISSUE #766: Allow translations for program names and descriptions
Currently, we only store a single name and description for programs. To support more languages, we should change these to `ImmutableMap<Locale, String>` so we can display the text in an applicant's preferred locale.

**Done when**
1. `ProgramDefinition` uses `ImmutableMap<Locale, String>` for `name` and `description`
2. The applicant program list view uses strings in their preferred locale

carolinedanzi:
### Description
Programs should have names and descriptions that are localized to all supported languages. To do this, the first step is adding `localizedName` and `localizedDescription` fields to `ProgramDefinition`

1. `name` for programs is now an immutable field - the admin sets this once, and it is later used to (for example) get all versions of a program with name x
2. Adds a `localizedName` field to `ProgramDefinition` that contains an `ImmutableMap<Locale, String>` with the localized, applicant-viewable names of programs
3. Adds a `localizedDescription` field to `ProgramDefinition` similar to the above, though the description is both for admin and applicant views

### Checklist

### Issue(s)
Fixes #766 
#777


universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -113,7 +115,11 @@ public ProgramDefinition getProgramDefinition(long id) throws ProgramNotFoundExc
       return ErrorAnd.error(errors);
     }
     Program program =
-        programDefinition.toBuilder().setName(name).setDescription(description).build().toProgram();
+        programDefinition.toBuilder()
+            .setLocalizedName(ImmutableMap.of(Locale.US, name))
```
hmhsu, line 119: Thought (blocking): this would not update the admin description, which may not be what we want?

universal-application-tool-0.0.1/conf/evolutions/default/15.sql
```diff
@@ -0,0 +1,11 @@
+# --- Add localized name and description to the Programs table.
+
+# --- !Ups
+
+alter table programs add localized_name jsonb;
+alter table programs add localized_description jsonb;
+
```
hmhsu, line 7: Thought (blocking): is it okay for jsonb to be null? If not please provide default value for backward compatibility.

```sql
update programs set localized_name = '{"en_US" : "' || name || '"}' where localized_name is null;
```

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
```diff
@@ -126,6 +127,23 @@ public void index_usesMessagesForUserPreferredLocale() {
...
 
+  @Test
+  public void index_missingTranslationForProgram_defaultsToEnglish() {
+    resourceCreator.insertProgram(
+        Locale.forLanguageTag("es-US"), "Programa en español", LifecycleStage.ACTIVE);
+    resourceCreator.insertProgram("English program", LifecycleStage.ACTIVE); // Missing translation
+
+    // Set the PLAY_LANG cookie
+    Http.Request request =
+        fakeRequest().langCookie(Locale.forLanguageTag("es-US"), stubMessagesApi()).build();
```
hmhsu, line 138: Thought (blocking): it might be more obvious to use a non-US locale here for the test.

hmhsu (approved):
In general looks good. Do we want to include input field for admin to set localized name/description? It's a bit confusing to me as we seem to set admin name/description on create but then on update we change the localized name/description.

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: Good point! I changed it to `Locale.FRANCE`

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
carolinedanzi: That's a good point - I'm not sure what the behavior should be for now. Eventually, these name/descriptions will be treated separately - adminName will always be immutable, but the others can change. I will leave this for now but will update in #777

universal-application-tool-0.0.1/conf/evolutions/default/15.sql
carolinedanzi: Good catch, thank you!  Added defaults for localized_name and localized_description

carolinedanzi:
> In general looks good. Do we want to include input field for admin to set localized name/description? It's a bit confusing to me as we seem to set admin name/description on create but then on update we change the localized name/description.

Agree it's confusing! The next step is to allow the admin to add localized names and descriptions for the applicant display cards: #777

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramsControllerTest.java
carolinedanzi: Never mind, I forgot the locale needs to be one of our supported locales, which so far only include `es-US` and `en-US`

PR #769, opened 2021-04-13
Bug fix: Replace `hasValue` with `isAnswered` methods throughout applicant question classes

### LINKED ISSUE #634: `hasValue` in the `ApplicantQuestion` subclasses don't seem to WAI

natsid:
### Description
Fixes a bug. In some cases, we were depending on `hasValue` to determine whether or not a question had an attempted answer, which would then help determine whether or not the question had any errors. However, `hasValue` would return false, even if there was an attempted answer, but it was empty. Instead, we use an `isAnswered` method that checks whether the JSON path exists in the applicant data. Now, even if the answer is empty, `isAnswered` will return true if the path exists.

### Note
This still doesn't work perfectly. See https://github.com/seattle-uat/civiform/issues/783 for why. Linked TODOs where relevant.

### Checklist

### Issue(s)
Fixes #634


universal-application-tool-0.0.1/app/services/applicant/question/MultiSelectQuestion.java
```diff
@@ -61,6 +61,10 @@ public boolean hasValue() {
     return getSelectedOptionsValue().isPresent();
   }
 
+  public boolean isAnswered() {
```
kendrickt, line 64: (blocking) Should this be added to the interface? or should there be an additional interface if `PresentsErrors` isn't the right scope?

It seems like if every `SomethingQuestion` has a public `isAnswered` method it should be available generically.

natsid: Good point.... I think that sounds good. The name that comes to mind is `Answerable`. LMK if you think of something better.

hmhsu (approved):

universal-application-tool-0.0.1/app/services/applicant/question/MultiSelectQuestion.java
natsid: Hmm, I guess one awkward aspect of it is that multi-field question types, like `AddressQuestion` have multiple `Answered` methods, like `streetAnswered`, `cityAnswered`. We could have an overall `isAnswered` method that takes the union of those though.

I'm trying to think of what value it adds to have an interface and the overall `isAnswered` method. The only place we really use it is to determine whether or not to check for errors. With that in mind, maybe it does make the most sense to have `isAnswered` be part of the `PresentsErrors` interface.

natsid: Updated. Let me know what you think of the change, @kendrickt .

natsid:
@kendrickt and I chatted offline yesterday.

As a result, I created https://github.com/seattle-uat/civiform/issues/783 and linked this issue in TODOs throughout these applicant question classes to use the "hydrated" paths instead of the generic question definition paths.

kendrickt (approved):
lgtm

PR #781, opened 2021-04-14
Admin can create repeated questions

### LINKED ISSUE #671: Enable creating Repeated question definitions.
Make sure when constructing new question definitions that have a "parent path" in the QuestionForm referencing a repeater question, the repeated question definition needs to have the repeaterId reference set correctly.

**NOTE**: make sure this works correctly with Question Versioning #641 .

**Additional context**
The `repeaterId` reference is created in #669 

**Done when**
Admin can create repeated question definitions. Nested repeater questions should also be possible.

This should have a browser test that creates a repeater question, and then creates repeated questions with that repeater. Not sure how a browser test can assert the correct results in the DB.

kendrickt:
### Description

Update `QuestionController`, `QuestionForm`, and `QuestionEditView` to enable defining repeated questions in the Admin UI.

Doubly nested repeated question from `/dev/seed`:
```
{
  "id" : 12,
  "version" : 1,
  "name" : "hm job address",
  "path" : {
    "segments" : [ "applicant", "household[]", "jobs[]", "hm_job_address" ],
    "empty" : false,
    "arrayElement" : false
  },
  "repeaterId" : 11,
....[omitted a bunch of other things]
}
```

Screenshot: repeater id options available when creating a new question
new-question-has-repeater-options

Screenshot: repeater id is an immutable member of question so only the question's repeater id is available when editing a question (across versions). 
editing-question_cant-change-repeater

### Open question
What text should be presented in the dropdown selector? Is path appropriate or nah?

### Checklist

### Issue(s)
Fixes #671


kendrickt:
This is blocked because of how `Path` is used by the question definition builder and each question form. Opening a separate PR to fix that first.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -38,31 +40,47 @@
 public final class QuestionEditView extends BaseHtmlView {
   private final AdminLayout layout;
 
+  private static final String NO_REPEATER_DISPLAY_STRING = "applicant";
```
bion, line 43: Suggestion (blocking): 
```suggestion
  private static final String NO_REPEATER_DISPLAY_STRING = "does not repeat";
```

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -243,6 +279,52 @@ private DomContent formQuestionTypeSelect(QuestionType selectedType) {
...
+      QuestionForm questionForm,
+      ImmutableList<RepeaterQuestionDefinition> repeaterQuestionDefinitions) {
+    ImmutableList.Builder<SimpleEntry<String, String>> optionsBuilder = ImmutableList.builder();
+    optionsBuilder.add(new SimpleEntry<>(NO_REPEATER_DISPLAY_STRING, NO_REPEATER_ID_STRING));
+    optionsBuilder.addAll(
+        repeaterQuestionDefinitions.stream()
+            .map(
+                repeaterQuestionDefinition ->
+                    new SimpleEntry<>(
+                        repeaterQuestionDefinition.getPath().toString(),
```
bion, line 310: Suggestion (blocking): use the question name rather than the path for the option display value. Rationale is that we don't expose the path anywhere else to the admin so it's unlikely they'll have any means of interpreting it if we display it here.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -243,6 +279,52 @@ private DomContent formQuestionTypeSelect(QuestionType selectedType) {
...
+  /**
+   * Generate a {@link SelectWithLabel} fixed repeater selector based on the question definition's
+   * repeater.
+   */
+  private SelectWithLabel repeaterOptionWithQuestionDefinition(
+      QuestionDefinition questionDefinition) {
+    Path parentPath = questionDefinition.getPath().parentPath();
+    SimpleEntry<String, String> repeaterPathAndId =
+        new SimpleEntry<>(
+            parentPath.isEmpty() ? NO_REPEATER_DISPLAY_STRING : parentPath.toString(),
```
bion, line 291: Suggestion (blocking): use the question name rather than the path for the option display value. Rationale is that we don't expose the path anywhere else to the admin so it's unlikely they'll have any means of interpreting it if we display it here.

bion (approved):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: Agreed that it would be confusing to admin.

I'm conflicted about using just the name. The name may not be clear for repeated entities that exist with different contexts (e.g. applicant's jobs vs jobs for each household member). It also won't sufficiently describe nested repeated entities very clearly.

natsid (approved):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -243,6 +279,52 @@ private DomContent formQuestionTypeSelect(QuestionType selectedType) {
         .withClasses(Styles.HIDDEN);
   }
 
+  /**
+   * Generate a {@link SelectWithLabel} fixed repeater selector based on the question definition's
+   * repeater.
+   */
+  private SelectWithLabel repeaterOptionWithQuestionDefinition(
```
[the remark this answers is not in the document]
kendrickt: I wanted to make it clear that this only gives one option. I can update the docstring to say that.

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -224,4 +244,22 @@ private QuestionForm createQuestionForm(Request request, String type)
...
+    try {
+      Path path =
+          roService.makePath(
+              questionForm.getRepeaterId(),
+              questionForm.getQuestionName(),
+              questionForm.getQuestionType().equals(QuestionType.REPEATER));
+      return questionForm.getBuilder(path);
+    } catch (QuestionNotFoundException | InvalidQuestionTypeException e) {
+      throw new RuntimeException(
+          "Failed to created a question definition builder because of invalid repeater id"
```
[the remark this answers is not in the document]
kendrickt: fixed!

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
carolinedanzi: Could we do something like "applicant -> child -> name"? Maybe something more sophisticated than arrows, but some indication of the chain of references

carolinedanzi (approved):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -71,32 +89,39 @@ private Content renderNewQuestionForm(
...
   }
 
   public Content renderEditQuestionForm(
-      Request request, long id, QuestionForm questionForm, String message) {
+      Request request,
+      long id,
+      QuestionForm questionForm,
+      QuestionDefinition questionDefinition,
+      String message) {
+    return renderEditQuestionForm(
```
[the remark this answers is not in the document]
kendrickt: yeah a question's name, type, repeater id reference, and path cannot change in an update or by creating a new draft of it.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: i'm just doing `name` for now. we can make this fancier in the future.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: i'm just doing `name` for now. we can make this fancier in the future.

PR #782, opened 2021-04-14
Get rid of "questionParentPath" from `QuestionForm`.

kendrickt:
### Description
Remove 'questionParentPath' from `QuestionForm`.

Adds an explicit `Path` argument to `QuestionForm.getBuilder` with a temporary method that generates a path based on just the name.

This is some pre-work for #781

### DIFF
--- a/universal-application-tool-0.0.1/app/forms/QuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/QuestionForm.java
@@ -4,7 +4,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.Locale;
-import java.util.Optional;
 import services.Path;
 import services.question.exceptions.TranslationNotFoundException;
 import services.question.types.QuestionDefinition;
@@ -14,7 +13,6 @@ import services.question.types.QuestionType;
 public class QuestionForm {
   private String questionName;
   private String questionDescription;
-  private Optional<Long> repeaterId;
   private QuestionType questionType;
   private String questionText;
   private String questionHelpText;
@@ -24,7 +22,6 @@ public class QuestionForm {
   public QuestionForm() {
     questionName = "";
     questionDescription = "";
-    repeaterId = Optional.empty();
     questionType = QuestionType.TEXT;
     questionText = "";
     questionHelpText = "";
@@ -33,7 +30,6 @@ public class QuestionForm {
   public QuestionForm(QuestionDefinition qd) {
     questionName = qd.getName();
     questionDescription = qd.getDescription();
-    repeaterId = qd.getRepeaterId();
     questionType = qd.getQuestionType();
 
     try {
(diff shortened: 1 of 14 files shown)


universal-application-tool-0.0.1/app/forms/QuestionForm.java
```diff
@@ -114,10 +98,20 @@ public QuestionDefinitionBuilder getBuilder() {
...
-            .setPath(getQuestionPath())
+            .setPath(path)
             .setDescription(questionDescription)
             .setQuestionText(questionTextMap)
             .setQuestionHelpText(questionHelpTextMap);
     return builder;
   }
+
+  /** This is a temporary fix until Path is properly handled. */
+  public Path getPath() {
```
carolinedanzi, line 109: question: why is this public? Can't we just call `getPath` in `getBuilder`?

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
```diff
@@ -49,29 +49,4 @@ public void getBuilder_returnsCompleteBuilder() throws Exception {
 
     assertThat(actual).isEqualTo(expected);
   }
-
```
carolinedanzi, line 52: suggestion: can we keep these test cases since this logic still exists?

carolinedanzi (commented):
I think I'm missing the big picture - why do we need to do this?

universal-application-tool-0.0.1/app/forms/QuestionForm.java
kendrickt: I wanted `getBuilder` to take a path argument, since we'll need to be passing a path somehow. One alternative is `getBuilder` returns a builder that doesn't have path set, and `builder.setPath(path)` needs to be called afterwards each time. I think it makes more sense to pass it in as an argument.

Right now, the `QuestionForm#getPath` method is used, but as soon as repeated questions are added the paths will be coming from `QuestionService#makePath`.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
kendrickt: we could, but the main method for making paths is `QuestionService#getPath` and the temporary path method is going to be deleted in a follow-up PR.

universal-application-tool-0.0.1/app/forms/QuestionForm.java
natsid: Does that mean we'll remove `QuestionForm#getPath` once we start using `ReadOnlyQuestionService#makePath`? I don't fully understand why we don't use that now since it already exists - think I'm just missing context from the complexity of repeater(d)s. :)

carolinedanzi: Agree that passing as an argument makes more sense since callers can't forget to set it

carolinedanzi (approved):
Thanks for clarifying offline!

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
natsid: That should say `ReadOnlyQuestionService#makePath`, right?

universal-application-tool-0.0.1/app/forms/QuestionForm.java
kendrickt: The `QuestionForm` doesn't have access to the `ReadOnlyQuestionService` so it needs to be passed in as an argument.

universal-application-tool-0.0.1/test/forms/QuestionFormTest.java
kendrickt: yup! edited the message correct it!

universal-application-tool-0.0.1/app/forms/QuestionForm.java
natsid: Oh, cool. #781 is helpful context!

natsid (approved):

PR #784, opened 2021-04-14
Faster prod image build

hmhsu:
### Description
- Match prod image build to dev build to leverage cache
- Also remove a dummy browser test


ndmckinley (approved):
Oh nice, it totally did not occur to me that we could use a cache for the prod build if we include those steps!  I removed them for speed (since they are not needed) but wound up decreating speed (since they ensure cache hits).  Nice!

Issue #759, opened 2021-04-12
Civiform admin: switch to "synchronized" versioning


neel-a-mehta:
- All program forms and questions should be versioned together with a monotonically increasing version number that the entire Civiform instance shares. [@bion has a doc for this](https://drive.google.com/file/d/1OJhbwpCnqB8sS-e7gAKdlGxsRvb8IBBg/view?usp=sharing).
- A separate page is created that lists all the past versions (avoid the endless sideways scrolling of the [mocks](https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457356370156646&cot=14), but the concept is right). Admins can click over to it to view all past versions.
- Click on a past version to overwrite the current draft with a copy of that past version.
- Overall, use the [light theme](https://miro.com/app/board/o9J_lXULxwE=/?moveToWidget=3074457354953282475&cot=14) mocks rather than the nearby dark theme ones.

neel-a-mehta:
Doing story acceptance testing.

* Logging in as admin on staging.seattle.civiform.com
* I just published a version before, so I go to the Versions tab and see Version 140 as active (with my most recent changes) and Version 141 is draft (with nothing inside, which makes sense).
* I see a helpful list of all past versions, from 132 to 139. I assume there's a reason we start at 132.
* I update my test program (Coffee program) to include the Favorite Ice Cream question. This requires making a new version.
* Go back to Versions. I see my updated coffee program is in the draft. Shows me the last updated time as 5 minutes ago, which I don't think is accurate. see the time in the bottom right of my screenshot (except I'm on eastern time, so +3 hours).
* 
![image](https://user-images.githubusercontent.com/77128518/116263066-131b3980-a747-11eb-8991-177979bb7c47.png)
* I go to edit the Favorite Ice Cream question. Add chocolate as an option alongside vanilla and strawberry.
* Go back to Questions list. I see the favorite Ice Cream question has a new draft.
* Go back to Versions. The last updated date hasn't changed, and the favorite ice cream doesn't show up there.
* Hm, I'll just publish what i have right now.
* Go back to Versions. I see version 141 includes the coffee program and a few other programs (idk why). shows the right last updated time. the draft shows the same last updated time, except this one is version 142 and is empty.
* Go back to edit the Ice Cream question again. Add another option (Rocky Road).
[...]

PR #788, opened 2021-04-15
Update existing versioning logic to do synchronized versioning.

ndmckinley:
### Description
All programs and questions increment in lock-step.

### Checklist

View example here:
Screen Shot 2021-04-15 at 11 42 01 AM

### Issue(s)
Related to #759 - doesn't close it yet since we don't have "go back in time" yet.
We should really consider doing #806 later - see discussion.


universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+                  ebeanServer
+                      .find(Program.class)
+                      .where()
+                      .raw("active.name = name")
+                      .eq("lifecycle_stage", LifecycleStage.DRAFT)
+                      .query())
+              .findList()) {
+        uneditedActive.setLifecycleStage(LifecycleStage.OBSOLETE);
+        uneditedActive.save();
+        Program newVersion = new Program(uneditedActive.getProgramDefinition());
```
bion, line 58: Suggestion (non-blocking): L41 has a variable 'version' that's bound to a `long`, whereas `newVersion` here is bound to a program. Recommend renaming `newVersion` to `newProgramVersion`.

bion: Alternatively, rename `version` to `versionNumber`

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+      ebeanServer.commitTransaction();
+    } finally {
+      ebeanServer.endTransaction();
+    }
+  }
+
+  public long getNextVersion() {
+    Optional<Program> draft =
+        ebeanServer
+            .find(Program.class)
```
bion, line 85: Question (blocking): it looks like a new draft version is created when a question or program is updated or created and a draft version hasn't been created yet. Is that the case? If so, will this logic still work if the version is created as a result of activity with a question that is not referenced by any programs?

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+            .find(Question.class)
+            .where()
+            .eq("lifecycle_stage", LifecycleStage.DRAFT)
+            .isNotNull("version")
+            .setMaxRows(1)
+            .findOneOrEmpty();
+    if (draftQuestion.isPresent()) {
+      return draftQuestion.get().getVersion();
+    }
+    // If there are no drafts, add one to any active program
```
bion, line 120: Suggestion (blocking): 
```suggestion
    // If there are no drafts, add one to any active question
```

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+                draft.setVersion(version);
+                this.publishProgram(draft, transaction);
+              });
+      ebeanServer.commitTransaction();
+    } finally {
+      ebeanServer.endTransaction();
+    }
+  }
+
+  public long getNextVersion() {
```
bion, line 82: Suggestion (blocking): document this public method

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+  private final ProgramRepository programRepository;
+  private final Logger LOG = LoggerFactory.getLogger(VersionRepository.class);
+
+  @Inject
+  public VersionRepository(EbeanConfig ebeanConfig, ProgramRepository programRepository) {
+    this.ebeanServer = Ebean.getServer(checkNotNull(ebeanConfig).defaultServer());
+    this.programRepository = checkNotNull(programRepository);
+  }
+
+  public void publishNewSynchronizedVersion() {
```
bion, line 36: suggestion (blocking): document this public method

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+import models.Program;
+import models.Question;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import play.db.ebean.EbeanConfig;
+import services.program.BlockDefinition;
+import services.program.ProgramDefinition;
+import services.program.ProgramQuestionDefinition;
+
+public class VersionRepository {
```
bion, line 24: Suggestion (blocking): document this public class

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,328 @@
...
+            .setMaxRows(1)
+            .findOneOrEmpty();
+    if (activeQuestion.isPresent()) {
+      return activeQuestion.get().getVersion() + 1;
+    }
+
+    return 1L;
+  }
+
+  private Optional<Question> getLatestVersionOfQuestion(Transaction transaction, long questionId) {
```
bion, line 136: Suggestion (non-blocking): it'd be good to make clear somewhere in documentation and/or comments that questions are linked via their immutable name.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
kendrickt: Should this be adding one to any active program? It seems like `Program` has precedence over `Question` in the first half of this code, which doesn't match basing the next version number on an active `Question` and not an active `Program` first.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
kendrickt: if there are no DRAFT programs, this method would continue copying all programs. That should be clear in the doc string.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -39,14 +41,31 @@ public Content render(ImmutableList<ProgramDefinition> programs, Http.Request re
...
+                            new GroupByKeyCollector<>(
+                                programDefinition -> programDefinition.name())),
+                    e -> this.renderProgramListItem(e.getKey(), e.getValue(), request)),
+                renderNewProgramButton(),
+                maybeRenderPublishButton(programs, request));
 
     return layout.render(head(layout.tailwindStyles()), body(contentDiv));
   }
 
+  private Tag maybeRenderPublishButton(
```
kendrickt, line 54: (blocking) it would be helpful to state this only renders if there are DRAFT programs.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Went with `newActiveProgram`.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Yes - I believe that's what's tested in VersionRepositoryTest.publishQuestion_simple.  If that's not what you're looking at could you construct a scenario that I could add to the tests?

kendrickt (commented):
For the most part, looks good to me. 

(blocking) can you add screenshots of view in the PR description?

(non-blocking, just an idea):
It is complicated to keep everything in sync across programs and questions, and differentiating between the one active version and one draft version. And over time there's bloat in making copies of everything even if there's just one small change.

Would it be easier to maintain if there was a `Version` table that was just a list of the versions and whether they were active, obsolete, or draft; and a `QuestionVersions` and `ProgramVersions` that just had foreign key references? I think that would help clarify what happens to programs and their questions when a new version is published. 

This would be yet-another-large-refactor, and is a premature optimization, and I def don't want this to affect this PR.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Done.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Done.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Done.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: Commented in the model and service classes.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
ndmckinley: Added.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
bion: See Kendrick's question

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -0,0 +1,338 @@
...
+import models.Program;
+import models.Question;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import play.db.ebean.EbeanConfig;
+import services.program.BlockDefinition;
+import services.program.ProgramDefinition;
+import services.program.ProgramQuestionDefinition;
+
+/** A repository object for dealing with versioning of questions or programs. */
```
bion, line 24:
```suggestion
/** A repository object for dealing with versioning of questions and programs. */
```

ndmckinley:
@kendrickt Yeah, absolutely - I thought about whether to do that, looked into ebean's mapping facilities, and you're 100% right that that'd be simpler, but I think it'd take longer.  The thing that pushed me over the edge is that we can't just drop the database and start from scratch anymore, and just testing the migration from the existing system to the new one would be the better part of a day.  I decided it wasn't a great trade - what do you think?  Do you think that's worth doing?

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: This part of the code is only in here so that we can correctly calculate versions in the specifically constructed case where we have inserted questions in the active state during testing.  You're right that programs have precedence, this comment is below the part where we have checked on programs.  We are basing this on a question first.

ndmckinley:
@kendrickt The more I think about it the more I think you're right that that's much, much better, though - I think I underestimated how much corner I was cutting.  Still not sure whether that's worth doing from here.  Thoughts?

bion (approved):

bion:
@ndmckinley IMO it's at your discretion to change/simplify the implementation. If you have confidence that it will work as-is and not cause undue barriers to continued work on the other vital features, then I think it's fine to submit as-is. We have a lot to get done and we're at the point of needing to make these kinds of trade-offs. Repeaters and i18n are both shaping up to be very big efforts and both are prereqs to piloting the first program.

kendrickt:
@ndmckinley can you create a short ticket describing the more flushed out solution, and reference it in this PRs description, and throw it in the void that is our ice box? 

That might help give more context to whoever ends up picking up that ticket in the future.

kendrickt (approved):
@ndmckinley can you create a short ticket describing the more flushed out solution, and reference it in this PRs description, and throw it in the void that is our ice box?

That might help give more context to whoever ends up picking up that ticket in the future.

ndmckinley:
Yeah - absolutely.

PR #789, opened 2021-04-15
Make `QuestionForm` abstract

### LINKED ISSUE #589: Create a distinct `QuestionForm` class for each of the existing question types

### LINKED ISSUE #590: Use different form class depending on question type
This work is blocked by #589, which creates the different form classes for each question type.

This task is about _using_ the correct form class given the question type selected. See the proposed solution in [this doc](https://docs.google.com/document/d/1r9_NZYQWxzBTF1zwKjeeEzpoUEejtrmMITmnCw-XlLk/edit?usp=sharing) for more details on how the question type should be passed.

**Done when** selecting to create a new text question presents the text question form and uses the `TextQuestionForm` to persist the data, and so on for each of the currently existing question types.

This issue blocks #299.

natsid:
### Description
* Make `QuestionForm` an abstract class. This means that from now on for any new question type added, there must also be a new question form class that extends `QuestionForm`.
* Clean up the way question type is used and computed throughout. Now each `QuestionForm` subclass is responsible for implementing its own `getQuestionType` method.
* Implement `NameQuestionForm` and `RepeaterQuestionForm`, the two remaining types that didn't have their own form classes.

### Checklist

### Issue(s)
Fixes #589
Fixes #590


hmhsu (approved):
Looks good! Thanks for completing this!

PR #805, opened 2021-04-15
Make empty number value inputs pass validation, unless there is a configured minimum or maximum value

### LINKED ISSUE #409: "Number" question type: implement simple validation criteria
Let the admin specify:

- Minimum value
- Maximum value

When building a question of this type.

See #392, the parent issue for this, and [this wiki page](https://github.com/seattle-uat/universal-application-tool/wiki/Question-type-requirements-notes) for details.

natsid:
### Description
#769 inadvertently created a new bug where an exception is thrown when a user attempts to submit an empty value for questions of type `number`. This PR makes it so that number fields are not required by default. In other words, the field will pass validation on the applicant side if it is left blank.

However, if the admin has configured a minimum or maximum value on the `number` question, we now *do* require a value, and return a `"Must be at least %d"` or `"Must be at most %d"` validation error message if it is left blank.

### Checklist

### Issue(s)
Fixes #409

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/question/NumberQuestion.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/NumberQuestion.java
@@ -29,8 +29,10 @@ public class NumberQuestion implements PresentsErrors {
 
     NumberQuestionDefinition questionDefinition = getQuestionDefinition();
 
-    // If there is no minimum value set, accept a blank answer.
-    if (getNumberValue().isEmpty() && questionDefinition.getMin().isEmpty()) {
+    // If there is no minimum or maximum value configured, accept a blank answer.
+    if (getNumberValue().isEmpty()
+        && questionDefinition.getMin().isEmpty()
+        && questionDefinition.getMax().isEmpty()) {
       return ImmutableSet.of();
     }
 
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/services/applicant/question/NumberQuestion.java
```diff
@@ -27,20 +27,27 @@ public boolean hasQuestionErrors() {
...
       }
     }
 
-    if (definition.getMax().isPresent()) {
-      long max = definition.getMax().getAsLong();
-      if (answer > max) {
+    if (questionDefinition.getMax().isPresent()) {
+      long max = questionDefinition.getMax().getAsLong();
+      // If the value is empty, do not consider it to be "greater than the maximum".
+      if (getNumberValue().isPresent() && getNumberValue().get() > max) {
```
ndmckinley, line 50: I think I would invert this - consider empty similar to the way that NaN works, which is to say, it fails any test you give it - it's both below the minimum and above the maximum.

natsid: Here's my thinking on it: If someone leaves a number answer blank, it implies "none". None is below the maximum (unless then max is negative...).

I can see either way though.

hmhsu (approved):
Looks good to me.

universal-application-tool-0.0.1/app/services/applicant/question/NumberQuestion.java
natsid: OK, updated!

Issue #672, opened 2021-04-03
Admin can create and edit repeater and repeated blocks.


kendrickt:
**Is your feature request related to a problem? Please describe.**
Enable admin creation of a Repeater Block, and subsequently Repeated Blocks. 

Question bank needs to follow rules for adding repeater and repeated questions to blocks.

Rules:
1. A repeater question can only be added if it is the only question. 
2. No other questions can be added after a repeater question is added to a block.
3. Repeated questions for a repeated entity are only available when editing a repeated block that is created from a repeater block with a repeater question for that entity.

**Describe the solution you'd like**

**Additional context**

**Done when**
Repeated and Repeater blocks can be created by the admin.

PR #810, opened 2021-04-16
Question bank changes based on the block definition being edited

kendrickt:
### Description
Questions available in the question bank are based on the block definition being edited:

1. Repeater questions are only available as the first question of a block definition
2. Only questions with the block definition's `repeaterBlockId` are available. This field isn't actually set yet.
3. All questions are unavailable for a repeater block
4. Questions already in the program are not shown (this is not new)

The `admin_dropdown_create_and_edit` test is moved as-is from `admin_program_creation` test because the name didn't match, and I wanted the name for the test I was adding.

### Screenshots

After adding a question, repeater questions are unavailable
no repeater question

After adding a repeater question, all other questions are unavailable.
no questions

### Checklist

### Issue(s)
Progress towards #672


universal-application-tool-0.0.1/app/views/components/QuestionBank.java
```diff
@@ -140,4 +149,44 @@ private ContainerTag renderQuestionDefinition(QuestionDefinition definition) {
                 addButton);
     return questionDiv.with(icon, content);
   }
+
+  /**
+   * Used to filter questions in the question bank.
```
hmhsu, line 154: praise: thanks for the java docstring, those are helpful.

browser-test/src/admin_program_creation.test.ts
```diff
@@ -1,64 +1,39 @@
...
-    await adminQuestions.expectDraftQuestionExist(questionName);
-
-    // Edit the question
-    await adminQuestions.gotoQuestionEditPage(questionName);
-    var questionSettingsDiv = await page.innerHTML('#question-settings');
-    expect(questionSettingsDiv.match(/<input/g)).toHaveLength(2);
+    // Remove the non-repeater question and add a repeater question. All options should go away..
+    await page.click('button:text("apc-name")');
+    await page.click('button:text("apc-repeater")');
+    expect(await page.innerText('id=question-bank-questions')).toBe('Question bank');
```
hmhsu, line 35: Suggestion (if-minor): could we test repeated questions in the repeater?

hmhsu (approved):
Looks good with a suggestion.

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
```diff
@@ -89,7 +96,7 @@ public boolean isRepeater() {
    */
   @JsonIgnore
   public boolean isRepeated() {
-    return repeaterId().isPresent();
+    return repeaterId().isPresent() && repeaterQuestionId().isPresent();
```
bion, line 99: Question (blocking): why is it not repeated if only `repeaterId().isPresent()`?

universal-application-tool-0.0.1/app/views/components/QuestionBank.java
```diff
@@ -140,4 +149,44 @@ private ContainerTag renderQuestionDefinition(QuestionDefinition definition) {
...
+   *
+   * <ul>
+   *   <li>If there is at least one question in the current block, all repeater questions
+   *   <li>If this is a repeated block only show the appropriate repeated questions
+   *   <li>Questions already in the program
+   * </ul>
+   */
+  private ImmutableList<QuestionDefinition> filterQuestions() {
+    // A repeater block cannot add any more questions
+    if (blockDefinition.isRepeater()) {
```
bion, line 166: Suggestion (non-blocking): it seems like this would be the better place to check if the block has a repeater question yet or not

bion (approved):

browser-test/src/admin_program_creation.test.ts
kendrickt: I need some more clarification on this.

If you mean checking that the repeated questions are available in repeated blocks, there's no way to create repeated blocks yet. That's next!

If you mean repeated questions in the repeater block, that isn't allowed. A repeater block can only have one question and it has to be a repeater question.

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
kendrickt: I thought about this more last night and I think we can get away with not having `repeaterQuestionId` at all.

universal-application-tool-0.0.1/app/views/components/QuestionBank.java
kendrickt: I don't understand the comment. Isn't this the place that is checking if the block has a repeater question?

bion: If you're not using `repeaterQuestionId` this can be ignored!

browser-test/src/admin_program_creation.test.ts
hmhsu: Yes I meant repeated questions in repeated blocks. Looking forward to it!

PR #811, opened 2021-04-16
Clean up exceptions and `badRequest` results in `QuestionController`

natsid:
### Description
* Refactor some `try`/`catch` blocks
* Make bad request messages more consistent for the same exceptions

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
@@ -90,7 +90,8 @@ public class QuestionController extends CiviFormController {
               try {
                 return ok(editView.renderViewQuestionForm(questionDefinition));
               } catch (InvalidQuestionTypeException e) {
-                return badRequest(e.toString());
+                return badRequest(
+                    invalidQuestionTypeMessage(questionDefinition.getQuestionType().toString()));
               }
             },
             httpExecutionContext.current());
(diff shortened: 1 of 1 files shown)


hmhsu (approved):
Thanks for doing this!

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -90,41 +90,43 @@ public QuestionController(
...
-      ImmutableList<RepeaterQuestionDefinition> repeaterQuestionDefinitions =
-          service
-              .getReadOnlyQuestionService()
-              .toCompletableFuture()
-              .join()
-              .getUpToDateRepeaterQuestions();
       return ok(editView.renderNewQuestionForm(request, questionType, repeaterQuestionDefinitions));
     } catch (UnsupportedQuestionTypeException e) {
       return badRequest(e.getMessage());
-    } catch (IllegalArgumentException e) {
```
carolinedanzi, line 112: question: what is throwing the `IllegalArgumentException`? Is it something in the view that still needs to be caught?

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -201,9 +203,7 @@ public Result update(Request request, Long id, String questionType) {
               .setLifecycleStage(LifecycleStage.DRAFT)
               .build();
     } catch (UnsupportedQuestionTypeException e) {
-      throw new RuntimeException(
-          "Failed while trying to update a question that was already created for question type "
```
carolinedanzi, line 205: suggestion (non-blocking): Can we move this message into a comment, so we still have info for when this case happens?

carolinedanzi (approved):
Thanks for doing this cleanup!

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
natsid: Good idea

universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
natsid: Good question. I changed a line above from `QuestionType.valueOf` (which can throw an `IllegalArgumentException`), to using `QuestionType.of`, which is a method on `QuestionType` that does some nicer handling and potentially throws an `InvalidQuestionTypeException`. So this should be fine.

PR #819, opened 2021-04-16
[BUGFIX-817] Set checkbox input value to true.

### LINKED ISSUE #817: Creating an address question with "disallow PO boxes" checked returns an error
Error looks like: "7jdfaak98".

Needs investigation.

kendrickt:
### Description
The checkbox input needs value=true.

### Screenshots
From /dev/seed, an address question without pobox checkbox checked
pobox-false

From /dev/seed, an address question with pobox checkbox checked
pobox-true

### Issue(s)
Fixes #817

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
+++ b/universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
@@ -113,12 +113,6 @@ public class QuestionConfig {
             .setLabelText("Disallow post office boxes")
             .setValue("true")
             .setChecked(addressQuestionForm.getDisallowPoBox())
-            .getContainer(),
-        FieldWithLabel.checkbox()
-            .setId("address-question-include-none-checkbox")
-            .setFieldName("noAddress")
-            .setLabelText("Include \"No address\" option")
-            .setValue("true")
             .getContainer());
     return this;
   }
--- a/universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
+++ b/universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
@@ -111,12 +111,14 @@ public class QuestionConfig {
             .setId("address-question-disallow-po-box-checkbox")
             .setFieldName("disallowPoBox")
             .setLabelText("Disallow post office boxes")
+            .setValue("true")
             .setChecked(addressQuestionForm.getDisallowPoBox())
             .getContainer(),
         FieldWithLabel.checkbox()
             .setId("address-question-include-none-checkbox")
             .setFieldName("noAddress")
             .setLabelText("Include \"No address\" option")
+            .setValue("true")
             .getContainer());
     return this;
   }


kendrickt:
The other checkbox for address question is not supported in the form, or in `AddressQuestionDefinition`. Should this be supported now, or is there a ticket for that question config?

hmhsu:
> The other checkbox for address question is not supported in the form, or in AddressQuestionDefinition. Should this be supported now, or is there a ticket for that question config?

Tagging @natsid for the second checkbox.

hmhsu (approved):
Could you add a browser test to avoid regression?

kendrickt:
@hmhsu I think we need more robust end-to-end browser testing for each question. I'm going to bring it up at stand-up on Monday and probably make tickets.

hmhsu:
Sounds good. Feel free to create tickets and merge this in as is.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
```diff
@@ -111,12 +111,14 @@ private QuestionConfig addAddressQuestionConfig(AddressQuestionForm addressQuest
             .setId("address-question-disallow-po-box-checkbox")
             .setFieldName("disallowPoBox")
             .setLabelText("Disallow post office boxes")
+            .setValue("true")
```
natsid, line 114: I think this should be `setValue(addressQuestionForm.getDisallowPoBox())`.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
```diff
@@ -111,12 +111,14 @@ private QuestionConfig addAddressQuestionConfig(AddressQuestionForm addressQuest
...
             .setChecked(addressQuestionForm.getDisallowPoBox())
             .getContainer(),
         FieldWithLabel.checkbox()
             .setId("address-question-include-none-checkbox")
             .setFieldName("noAddress")
             .setLabelText("Include \"No address\" option")
+            .setValue("true")
```
natsid, line 121: Should we add this now or wait till we have implemented the "no address" option?

natsid:
Agreed that we need robust end-to-end browser testing for each question type.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
kendrickt: I think this should be updated now, whether or not we have implemented the option. Alternatively, we could get rid of the checkbox entirely.

What do you think?

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
kendrickt: I don't actually know the details, but it looks like this is working as intended (see the screenshots in the description). The form data in the request object does not include `disallowPoBox` at all when the box isn't checked, and includes it with `"true"` when it is checked.

kendrickt:
I opened #825 to track browser testing for address question.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
kendrickt: I deleted the checkbox option entirely.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionConfig.java
natsid: Thank you for explaining! That makes sense.

PR #834, opened 2021-04-19
Separate ECS task roles between staging and prod

hmhsu:
### Description
Use separate roles for staging and prod so they cannot access each other's resources, specifically S3 buckets.

### DIFF
--- a/infra/containers.yaml
+++ b/infra/containers.yaml
@@ -6,6 +6,11 @@ Mappings:
       image: public.ecr.aws/t1q6b4h2/universal-application-tool:prod
     staging:
       image: public.ecr.aws/t1q6b4h2/universal-application-tool:latest
+  S3TaskRoles:
+    prod:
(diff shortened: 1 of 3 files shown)


infra/containers.yaml
```diff
@@ -6,6 +6,11 @@ Mappings:
       image: public.ecr.aws/t1q6b4h2/universal-application-tool:prod
     staging:
       image: public.ecr.aws/t1q6b4h2/universal-application-tool:latest
+  S3TaskRoles:
```
ndmckinley, line 9: This is a heck of a nitpick - but would you mind putting this at the root level and passing it in as an argument to each of these substacks, so that we can have it in only one place?

hmhsu: Yeah, agreed that's better. Done.

ndmckinley (approved):
Thanks!  :)

PR #812, opened 2021-04-16
Add `REPEATER` to `errorsPresenter` switch statement

natsid:
### Description
* #697 added the `RepeaterQuestion` class. This PR just adds the repeater question type to a switch statement and updates a test to reflect that.
* Also cleans up ApplicantQuestionTest a little more.

### Checklist
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)
Works toward #405

PR #839, opened 2021-04-21
Bug fix: Update TestQuestionBank to use non-static methods

natsid:
### Description
There was a conflict in tests that used `TestQuestionBank`. The source of the problem was the `maybeSave` method, which attempted to `save` and catch errors if there was no database present. For some reason, this wasn't working well and caused tests to fail with a strange error when a test *with* a database was run sequentially after a test *without* a database.

The fix is to specify upon construction whether the `TestQuestionBank` `canSave` questions.

This bug was discovered in #812.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/test/repository/WithPostgresContainer.java
+++ b/universal-application-tool-0.0.1/test/repository/WithPostgresContainer.java
@@ -25,9 +25,9 @@ public class WithPostgresContainer {
 
   protected static Materializer mat;
 
-  private static ResourceCreator resourceCreator;
+  protected static ResourceCreator resourceCreator;
 
-  private static TestQuestionBank testQuestionBank = new TestQuestionBank(true);
+  protected static TestQuestionBank testQuestionBank = new TestQuestionBank(true);
 
   @BeforeClass
   public static void startPlay() {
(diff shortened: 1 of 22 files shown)


universal-application-tool-0.0.1/test/repository/WithPostgresContainer.java
```diff
@@ -25,7 +25,9 @@
 
   protected static Materializer mat;
 
-  protected static ResourceCreator resourceCreator;
+  private static ResourceCreator resourceCreator;
+
+  private static TestQuestionBank testQuestionBank = new TestQuestionBank(true);
```
kendrickt, line 30: (nit) this could be a protected member. Not sure if there's much difference between using `testQuestionBank()` method or directly using `testQuestionBank` member.

kendrickt (approved):
Looks like the resource creator and test question bank are accessed through methods. I would lean towards direct member access in the tests, but that might be bad java.

universal-application-tool-0.0.1/test/repository/WithPostgresContainer.java
natsid: Agreed... I saw there was the protected `resourceCreator` method, so I thought I'd do the same thing for `testQuestionBank`.

Just went back and removed both "getters" and made the members `protected`.

natsid:
No clue if it's bad Java either 😆  Just made them used via the member vars instead.

hmhsu (approved):

Issue #837, opened 2021-04-20
Address security scan issues


bion:
Address security scan issues:

- [ ] HSTS – this can be applied at the application, F5 or sever level. Any one of those is fine.
- [ ] Weak Ciphers Enabled – this appears to be coming from the web server. There is a list of weak ciphers that are used and what is considered strong as replacements.
- [x] Cookie Not Marked as HttpOnly – mark cookie as HttpOnly. This will add an extra layer of defense against XSS.
- [x] Cookie Not Marked Secure – Mark all cookies as secure in the application.
- [ ] TLS 1.0 – this looks to be coming from the web server.

Scan details: https://drive.google.com/file/d/1zS_LOiDlI5-CmYhiCTnmIgd3LWAjb5yd/view?usp=sharing

PR #840, opened 2021-04-21
Use secure cookies in prod

bion:
Fixing cookie related items in #837

- [x] Specify HTTP only for all cookies
- [x] Specify SSL required for all cookies in prod
- [x] introduce dev application config overrides, test config inherits from it

### DIFF
--- a/universal-application-tool-0.0.1/conf/application.conf
+++ b/universal-application-tool-0.0.1/conf/application.conf
@@ -102,10 +102,10 @@ play.i18n {
   langs = [ "en-US", "es-US" ]
 
   # Whether the language cookie should be secure or not
-  #langCookieSecure = true
+  langCookieSecure = true
 
   # Whether the HTTP only attribute of the cookie should be set to true
-  #langCookieHttpOnly = true
+  langCookieHttpOnly = true
 }
 
 ## Play HTTP settings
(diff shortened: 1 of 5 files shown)


universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -142,12 +142,12 @@ play.http {
...
   session {
     # Sets the cookie to be sent only over HTTPS.
-    #secure = true
+    secure = true
 
     # Sets the cookie to be accessed only by the server.
-    #httpOnly = true
+    httpOnly = true
 
-    sameSite = null
```
ndmckinley, line 150: I believe that this will change the behavior of login - if you want to do this, you'll need to check both login providers in staging.

ndmckinley (approved):

universal-application-tool-0.0.1/conf/application.conf
bion: Good call, I'll test it after merging. My intention is not to change behavior but to just be more explicit. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#values) the default value is 'Lax' when unspecified.

ndmckinley: Yeah, I recall finding that wasn't true when I added this line for the first time.

hmhsu (approved):

universal-application-tool-0.0.1/conf/application.conf
bion: Leaving out this change for this PR

PR #851, opened 2021-04-22
Improve the repeater question renderer

### LINKED ISSUE #705: Make `RepeaterQuestionRenderer` render something, but not functionally complete.
Right now the `RepeaterQuestionRenderer` doesn't do anything. 

The `RepeaterQuestionRenderer` should visually be good. Functionally, it may not do anything yet because none of the applicant controllers and applicant service can't really handle repeated entity keys.

**Additional context**

**Done when**
The `RepeaterQuestionRenderer` can be used on the admin side for viewing what they question looks like. Functionality can come later when the applicant services are capable of handling repeated entity keys.

kendrickt:
### Description
Update `RepeaterQuestionRenderer` and add a typescript function for clicking the add field button.

image

The html form for the repeater question should have two parts: an array of string values for repeated entity name, and an array of repeated entities to delete, which is quite different from all of our other question types.

### Checklist

### Issue(s)
Fixes #705

### DIFF
--- a/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
+++ b/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
@@ -64,16 +64,20 @@ function removeQuestionOption(event: Event) {
 
 /** In the repeater form - add a new input field for a repeated entity. */
 function addNewRepeaterField(event: Event) {
-  // Copy the answer template and remove ID and hidden properties.
-  const newField = document.getElementById("repeater-field-template").cloneNode(true) as HTMLElement;
-  newField.classList.remove("hidden");
-  newField.removeAttribute("id");
+  // Copy the first repeater field
+  const newField = document.querySelector(".repeater-field").cloneNode(true) as HTMLElement;
+
+  // Remove the text
+  const newTextField = newField.querySelector("[type=text]") as HTMLInputElement;
+  newTextField.value = null;
 
   const repeaterFields = document.getElementById("repeater-fields");
-
-  // Set new checkbox value to the index
   const index = repeaterFields.querySelectorAll(".repeater-field").length;
-  newField.querySelector("[type=checkbox]").setAttribute("value", index.toString());
+
+  // Set new checkbox value to the index and uncheck it
+  const newCheckbox = newField.querySelector("[type=checkbox]") as HTMLInputElement;
+  newCheckbox.setAttribute("value", index.toString());
+  newCheckbox.checked = false;
 
   repeaterFields.appendChild(newField);
 }
--- a/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
+++ b/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
@@ -62,6 +62,22 @@ function removeQuestionOption(event: Event) {
   optionDiv.parentNode.removeChild(optionDiv);
 }
 
+/** In the repeater form - add a new input field for a repeated entity. */
+function addNewRepeaterField(event: Event) {
+  // Copy the answer template and remove ID and hidden properties.
+  const newField = document.getElementById("repeater-field-template").cloneNode(true);
+  newField.classList.remove("hidden");
+  newField.removeAttribute("id");
+
+  const repeaterFields = document.getElementById("repeater-fields");
+
+  // Set new checkbox value to the index
+  const index = repeaterFields.querySelectorAll(".repeater-field").length;
+  newField.querySelector("[type=checkbox]").setAttribute("value", index.toString());
+
+  repeaterFields.appendChild(newField);
+}
+
 function init() {
   attachDropdown("create-question-button");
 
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -20,10 +32,14 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
 
   @Override
   public Tag render() {
+    ContainerTag repeaterFields = div().withId(REPEATER_FIELDS_ID);
+    // TODO: add each answer as a repeaterField
```
carolinedanzi, line 36: suggestion (non-blocking): Can we also add a TODO for making the placeholder name an admin-configurable field, with a link to an issue?

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -20,10 +32,14 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
+    repeaterFields.with(repeaterField("placeholder", Optional.empty(), 0));
+
     return div()
-        .withId(question.getPath().path())
         .withClasses(Styles.MX_AUTO, Styles.W_MAX)
         .with(
+            repeaterFieldTemplate(),
```
carolinedanzi, line 42: I think adding the template here will cause it to be submitted along with the form, which I'm not sure is what we want (it will bind something like `entity[]: ""`). I think we need to put it somewhere on the page for a block, but outside the block form - @bion has a better understanding of this

carolinedanzi (commented):
This looks good, I just have one concern on the template

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: Ahh yeah there is no "form" here. It might make sense to render repeater blocks in a very special way, that is not like how we render the other blocks, i.e. the form to be part of the repeater renderer since it doesn't have to be generically handled like blocks with other questions.

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: Added a TODO for #859.

carolinedanzi: Thanks!

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
carolinedanzi: I think I may  be missing some context on how repeated blocks can be handled differently. For now this seems ok, though it will appear in the form from `ApplicantBlockEditView`

carolinedanzi (approved):

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: There's gonna be special cases for handling enumeration questions and enumeration blocks. So I was assuming the `ApplicantProgramBlockEditView` would do things a bit differently. But maybe not.

I'm gonna make a change to not use a template and just copy the last element in the `repeater_fields` with all the right things reset.

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -62,6 +62,26 @@ function removeQuestionOption(event: Event) {
   optionDiv.parentNode.removeChild(optionDiv);
 }
 
+/** In the repeater form - add a new input field for a repeated entity. */
+function addNewRepeaterField(event: Event) {
+  // Copy the first repeater field
```
carolinedanzi, line 67: praise: Clever!

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -20,8 +35,11 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
 
   @Override
   public Tag render() {
+    ContainerTag repeaterFields = div().withId(REPEATER_FIELDS_ID);
+    // TODO: add each answer as a repeaterField
+    repeaterFields.with(repeaterField(PLACEHOLDER, Optional.empty(), 0));
```
carolinedanzi, line 40: question (blocking): if an applicant initially had no entities (ex: they have no children) but later need to add one, how will this work if there is no initial field? Or will we always display a blank one?

carolinedanzi (approved):
Looks good, have one question

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: I wasn't sure what we wanted the UX to be.

I think showing a blank one makes sense. I think this would also make sense for the `MultiOptionQuestionDefinition` piece too since the admin wouldn't ever create a question without giving it at least one option.

Not sure how `value=null` shows up in the request object, but we'd be able to handle things on the server side.

carolinedanzi: Thanks for the explanation!

Issue #777, opened 2021-04-13
Civiform admin can translate (internationalize) program name/description


neel-a-mehta:

neel-a-mehta:
Doing story acceptance review per @carolinedanzi 's request.

* Logging into admin view on staging.seattle.civiform.com.
* Creating a brand new program from scratch. Calling it "Seattle Coffee Program."
* Program is now a draft.
* Hitting Manage Translations.
* It appears I can edit English and Spanish names independently.
![image](https://user-images.githubusercontent.com/77128518/116261236-6f7d5980-a745-11eb-9bec-c84d5c59c7e0.png)
* I set a Spanish name for the program.
* Click away.
* Return to program -- I see Spanish name change went through.
* I publish my changes.
* I log out and re-log in as a guest.
* Set language to Spanish during onboarding flow.
* I see my program name and description in Spanish![image](https://user-images.githubusercontent.com/77128518/116261958-237ee480-a746-11eb-994e-007566df35ba.png)

This works as intended! Good to go. I'll close this.

PR #852, opened 2021-04-22
Add `getSupportedLanguages` to `ProgramDefinition`

carolinedanzi:
### Description
Adds a method to program definitions for getting locales for which this program is fully translated. A program fully supports a locale if
1. The program display name is localized
2. The program description is localized
3. For every question in the program form, the question fully supports that locale

### Checklist

### Issue(s)
#777

### DIFF
--- a/universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
@@ -1,12 +1,17 @@
 package services.program;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
 import com.google.auto.value.AutoValue;
+import com.google.auto.value.extension.memoized.Memoized;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.Set;
 import models.LifecycleStage;
 import models.Program;
 import services.LocalizationUtils;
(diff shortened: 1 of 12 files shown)


universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -103,6 +107,29 @@ public String getLocalizedDescription(Locale locale) throws TranslationNotFoundE
...
+
+    Set<Locale> intersection =
+        Sets.intersection(localizedName().keySet(), localizedDescription().keySet());
+    for (ImmutableSet<Locale> set : questionLocales) {
+      intersection = Sets.intersection(intersection, set);
+    }
+    return ImmutableSet.copyOf(intersection);
+  }
+
+  public ImmutableSet<QuestionDefinition> getQuestionDefinitions() {
```
natsid, line 124: question (non-blocking): Should we make this private until/unless we end up with a need for it to be public?

natsid (approved):
Good stuff!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
kendrickt: (blocking) this should return a stream since that's how it is used everywhere. It could be renamed `getQuestionDefinitionStream`. And I agree with @natsid that this should be private.

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -103,6 +107,29 @@ public String getLocalizedDescription(Locale locale) throws TranslationNotFoundE
     }
   }
 
+  public ImmutableSet<Locale> getSupportedLocales() {
```
kendrickt, line 110: (blocking) this should have a docstring describing what it means to be "supported"

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: Good catch, thank you!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: Thank you for catching that! Added

PR #861, opened 2021-04-23
Fix i18n bug #856 - do not throw when locale not found

### LINKED ISSUE #856: Text should default to Locale.US if a translation is not found
**Describe the bug**
I ran into the following error when I had my language set to `es-US`:
```
[CompletionException: java.lang.RuntimeException: java.lang.RuntimeException: services.question.exceptions.TranslationNotFoundException: Translation not found for Question at path: applicant.color Locale: es_US]
```

**To Reproduce**
1. Sign in as guest
2. Select "Español" as preferred language
3. Apply to a program

**Expected behavior**
If a translation does not exist, it should default to English instead of throwing

carolinedanzi:
### Description
Currently, we throw an error if a question cannot be displayed in an applicant's preferred locale. This changes this behavior by:

1. Set a global default locale for CiviForm and use across app
2. Add methods to `QuestionDefinition` to fall back to the default locale text
3. Update `ApplicantQuestion` to use new fallback methods
4. Add a method to `ReadOnlyApplicantService` to determine whether the given program fully supports the applicant's preferred locale
4. Display a toast message to applicants when they apply to a program that is not fully localized to their preferred language

### Checklist

### Issue(s)
Fixes #856 

Screen Shot 2021-04-23 at 9 45 15 AM

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramController.java
@@ -5,7 +5,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import auth.Authorizers;
 import controllers.CiviFormController;
 import forms.ProgramForm;
-import java.util.Locale;
 import javax.inject.Inject;
 import org.pac4j.play.java.Secure;
 import play.data.Form;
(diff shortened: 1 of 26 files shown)


hmhsu (approved):
Thanks for adding the tests!

PR #872, opened 2021-04-23
Add HSTS to production only.

ndmckinley:
### Description
Adding HSTS via a filter in the application.

### Checklist

### Issue(s)
Addresses point 1 of https://github.com/seattle-uat/civiform/issues/837.

### DIFF
--- a/universal-application-tool-0.0.1/app/filters/HSTSFilter.java
+++ b/universal-application-tool-0.0.1/app/filters/HSTSFilter.java
@@ -1,0 +1,32 @@
+package filters;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.concurrent.Executor;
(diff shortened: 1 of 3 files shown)


bion:
Looks like this can be set using application.conf instead of a custom filter? https://www.playframework.com/documentation/2.8.x/RedirectHttpsFilter#Strict-Transport-Security

ndmckinley:
@bion Could do - but that requires the redirect HTTPS filter, which we do at the infrastructure layer.  It's a pretty simple filter, would this be okay as-is?

bion:
> @bion Could do - but that requires the redirect HTTPS filter, which we do at the infrastructure layer. It's a pretty simple filter, would this be okay as-is?

Ah got it, in that case LGTM

bion (approved):

PR #878, opened 2021-04-24
Fix dev/seed.

ndmckinley:
### Description
Doing version truncation causes some issues, and the Version table doesn't have foreign key references, so deletes don't cascade.

### Checklist

### Issue(s)

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/dev/DatabaseSeedController.java
+++ b/universal-application-tool-0.0.1/app/controllers/dev/DatabaseSeedController.java
@@ -21,6 +21,7 @@ import play.Environment;
 import play.db.ebean.EbeanConfig;
 import play.mvc.Http.Request;
 import play.mvc.Result;
+import repository.VersionRepository;
 import services.Path;
 import services.program.ProgramDefinition;
 import services.program.ProgramQuestionDefinition;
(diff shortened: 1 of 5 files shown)


universal-application-tool-0.0.1/app/controllers/dev/DatabaseSeedController.java
```diff
@@ -253,6 +257,8 @@ private ProgramDefinition insertProgramWithBlocks(String name) {
   }
 
   private void truncateTables() {
+    versionRepository.getActiveVersion().empty(this);
+    versionRepository.getDraftVersion().empty(this);
```
hmhsu, line 261: Suggestion (blocking): I'd lean toward truncating version table as well as all the old version would not work, i.e. cannot be revive, after all programs and questions being deleted.

ndmckinley: I'll give that a try.

universal-application-tool-0.0.1/app/controllers/dev/DevController.java
```diff
@@ -19,7 +19,7 @@ public DevController(Environment environment, Config configuration) {
   }
 
   public boolean isDevEnvironment() {
-    if (environment.isDev()) {
+    if (environment.isDev() || environment.isTest()) {
```
hmhsu, line 22: Thoughts: I think we might not need this, the only test that touches this set its environment to dev already.

ndmckinley: Yeah, those tests are currently giving me trouble ... suppose I could try another tack.

hmhsu (approved):
Looks good. Hope the tests pass!

PR #879, opened 2021-04-24
Dev seed to only include current programs

hmhsu:
### Description
- Only show current draft and active programs in dev seed controller.
- Ignore getSupportedLocales field because it cannot be serialized without further loading question definitions.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/views/dev/DatabaseSeedView.java
+++ b/universal-application-tool-0.0.1/app/views/dev/DatabaseSeedView.java
@@ -41,16 +41,8 @@ public class DatabaseSeedView extends BaseHtmlView {
       ImmutableList<QuestionDefinition> questionDefinitions,
       Optional<String> maybeFlash) {
 
-    ImmutableList<ProgramDefinition> draftPrograms =
-        activeAndDraftPrograms.getProgramNames().stream()
-            .filter(name -> activeAndDraftPrograms.getDraftProgramDefinition(name).isPresent())
-            .map(name -> activeAndDraftPrograms.getDraftProgramDefinition(name).get())
-            .collect(ImmutableList.toImmutableList());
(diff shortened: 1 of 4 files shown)


universal-application-tool-0.0.1/app/views/dev/DatabaseSeedView.java
```diff
@@ -36,11 +37,23 @@ public DatabaseSeedView(BaseHtmlLayout layout, ObjectMapper objectMapper) {
...
       ImmutableList<QuestionDefinition> questionDefinitions,
       Optional<String> maybeFlash) {
 
-    String prettyPrograms = getPrettyJson(programDefinitions);
+    ImmutableList<ProgramDefinition> draftPrograms =
+        activeAndDraftPrograms.getProgramNames().stream()
+            .filter(name -> activeAndDraftPrograms.getDraftProgramDefinition(name).isPresent())
+            .map(name -> activeAndDraftPrograms.getDraftProgramDefinition(name).get())
+            .collect(ImmutableList.toImmutableList());
+    ImmutableList<ProgramDefinition> activePrograms =
```
ndmckinley, line 49: I guess I'd probably add this as a method on ActiveAndDraftPrograms, but it's all the same to me.

ndmckinley (approved):

universal-application-tool-0.0.1/app/views/dev/DatabaseSeedView.java
ndmckinley: uh, sorry, in our normal language that would be:
nit(nonblocking): this might be good as a method on ActiveAndDraftPrograms (maybe ActiveAndDraftQuestions also).

hmhsu: Cool, can do.

PR #880, opened 2021-04-24
dev/seed lists too many programs - make it list only active and draft ones

ndmckinley:
### Description
It's not beautiful but it does work!  It also fixes a sort of incipient problem - there's no transactionality around creating new drafts, so we need to be able to handle multiple drafts accidentally existing.  The first fix exposed that problem so I fixed that also.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/VersionRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/VersionRepository.java
@@ -9,6 +9,7 @@ import io.ebean.EbeanServer;
 import java.util.List;
 import java.util.Optional;
 import javax.inject.Inject;
+import javax.persistence.NonUniqueResultException;
 import models.LifecycleStage;
 import models.Program;
 import models.Question;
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -101,6 +108,31 @@ public Version getDraftVersion() {
     }
   }
 
+  private void handleMultipleDrafts() {
```
hmhsu, line 111: Thoughts (non-blocking): I wonder if these should be guarded by a single transaction.

hmhsu (approved):

universal-application-tool-0.0.1/app/repository/VersionRepository.java
ndmckinley: It's hard because the transactions are global - they probably SHOULD but you can't be sure you haven't arrived here inside another transaction, and `commitTransaction` might prematurely commit the parent transaction.

I _think_.  no idea if my experiments really prove this at all.

Issue #862, opened 2021-04-23
Add browser tests for i18n


carolinedanzi:
**Describe the solution you'd like**
We should add browser tests for the following cases:
1. An admin adds translations for program display name and description
2. An admin adds translations for question text
3. An applicant views an application in their preferred non-English locale
4. An applicant sees a toast message stating that we cannot display the program in their preferred locale (#856)

**Done when**
The above test cases are covered by browser tests and there are helper/support functions for adding localizations as an admin

PR #889, opened 2021-04-26
Add browser test for translated programs

carolinedanzi:
### Description
1. Adds `admin_translations.ts` with helper methods for translating programs (and later questions)
2. Adds `admin_program-translation.test.ts` with two tests - one for translating a program's name and description, and one for viewing a toast message when a program is not fully translated

### Checklist

### Issue(s)
#862


universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
```diff
@@ -150,6 +150,7 @@ private ContainerTag programCard(
             .withText(applyText)
             .withId(baseId + "-apply")
             .withClasses(
+                "apply-button",
```
hmhsu, line 153: Nit (non-blocking): You could define this constant in `view/style/ReferenceClasses.java`.

browser-test/src/support/admin_translations.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class AdminTranslations {
+  public page!: Page
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  async selectLanguage(language: string) {
+    await this.page.click(`.language-link:has-text("${language}")`);
```
hmhsu, line 11: Suggestion (blocking): please prefer `:text("search-text")` over `:has-text("search-text")` as the latter returns any parents that contain an element that has the matching text, while the former returns the immediate element that contains the matching text.

hmhsu (approved):
Thanks for adding the tests.

universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
carolinedanzi: Great suggestion! Updated

browser-test/src/support/admin_translations.ts
carolinedanzi: Ah that's good to know, thank you!

carolinedanzi:
> Thanks for adding the tests.

Thank you for the review!

browser-test/src/support/admin_translations.ts
carolinedanzi: Done!

Issue #778, opened 2021-04-13
Enable translation/i18n of questions directly within the Civiform admin UI


neel-a-mehta:
#776 describes how admins can download a list of *all* strings required to be translated, translate in a new column, and re-upload. A more graphical (and maybe easier to implement) alternative is for Civiform admins to be able to edit question titles, help texts, and options (for checkboxes, radio buttons, etc.) directly within the Civiform admin UI.

neel-a-mehta:
This is going to be more important than #776; we want admins to be able to edit translations from directly within the UI.

#829 requests UX help for this.

neel-a-mehta:
All right, going to list some product requirements and rough wireframes here -- more as proofs of concept rather than proper UI/UX explorations. Hopefully, everything here will be agnostic to the specifics of UI/UX implementation.

- Every question tracks which language(s) it has been translated into. By default, every question starts off with just English.
- The set of languages that any question is translated into is shown in the question bank, like so. We can use either the full language names (English, Spanish, French) or their [2-letter language codes](https://www.sitepoint.com/iso-2-letter-language-codes/) (EN, ES, FR).
![image](https://user-images.githubusercontent.com/77128518/115302468-75b17b80-a130-11eb-99d0-d54cbb1e8309.png)
- In the Question editor page, there's one tab for every translation that a given question has. The tab bar is placed under the fields that are invariant to translation (namely, the fields that only the admin sees, like the question name and description). But all the fields that change per translation (such as the user-facing text) are within the tabbed pane and change when the user switches tabs. ([Tailwind snippet](https://play.tailwindcss.com/ieqjtUpoGn))
![image](https://user-images.githubusercontent.com/77128518/115303314-77c80a00-a131-11eb-9666-9cfbc611f7b3.png)

neel-a-mehta:
- Note that free-text configuration options are also in the tabbed pane, so the admin has to specify different versions of each for each language. (A pitfall of this design is that configuration options that presumably should always be the same -- such as the min or max number of options the user can choose -- also change between tabbed panes, meaning the admin has to copy the same values every time they add a new language. But I think that's OK for now.)
![image](https://user-images.githubusercontent.com/77128518/115303736-fcb32380-a131-11eb-9dc5-aab6064ae9be.png)
- When the user hits "+ Add Language" in the tab bar, a text input field appears. They then hit a "+" button to create a translation for that language. ([Tailwind snippet](https://play.tailwindcss.com/9kIBRPcSDW))
[...]

Issue #211, opened 2021-02-22
UAT admin can localize/translate forms & questions


neel-a-mehta:
UAT admin can see a list of languages they can localize forms into on a dedicated tab. Then they can specify localized strings for all questions in the form.

It would look something like this on the question page:
![image](https://user-images.githubusercontent.com/77128518/108777687-31d83680-7532-11eb-9551-69f08121ea1b.png)

carolinedanzi:
This seems to be a duplicate of #778, which is nearly ready for story acceptance

PR #891, opened 2021-04-27
[WIP] Add backend support for localizing questions

carolinedanzi:
### Description
1. Adds two new routes for editing and updating question translations
2. Implements two controller methods for the above
3. Adds an empty view for managing translations (this is the next step)
4. Removes an unused `addTranslations` method from `QuestionService`. Instead, we will use update, since some questions need to translate more than just the question text and help text (ex: answer options)

### Checklist

### Issue(s)
#778, #211


AWBaumann (approved):
Thanks for all of the i18n work!

carolinedanzi:
> Thanks for all of the i18n work!

Thanks Aaron!

PR #897, opened 2021-04-27
Remove `"metadata"` from initial applicant data and put `"applicant"` in a constant

natsid:
### Description

Initially we had applicant data JSON structured like:

```
{
  "applicant": {},
  "metadata": {}
}
```

However a recent change made metadata live alongside applicant data answers (i.e., within the `"applicant"` object), so now this top-level `"metadata"` object is no longer needed.

Also create new constants that refers to the "applicant" path in the applicant data JSON.

### Checklist

Paired with @kendrickt on this.


carolinedanzi (approved):
Thank you for doing this!

PR #905, opened 2021-04-28
Improve language select in browser tests

carolinedanzi:
### Description
Assert page by URL, rather than elements that may or may not be there immediately

Ran @kendrickt's script and it passed 10/10 times


kendrickt (approved):
Looks like these are the only places where `page.$` is used.

PR #918, opened 2021-04-29
Recursively generate `Block`s in `ReadOnlyApplicantProgramServiceImpl`

### LINKED ISSUE #783: Implement "contextualized" paths for repeated questions
In order to get the fully specified JSON path to applicant data that has a repeater as a parent, we need the path with indices into any repeaters.

Currently, we just use the path from the QuestionDefinition to look up applicant data. However, that path is "unhydrated" and is a generic path like `applicant.household_members[].address`. We need the path with indices, like `applicant.household_members[0].address`, in order to actually reference the applicant data to prefill forms and to update applicant data.

The solution must also work for nested repeaters. For example, `applicant.household_members[1].employers[0].phone_number`.

### More thoughts about the potential solution

The solution for this will likely involve creating a new `RepeaterContext` class that has a list of indices into any repeaters as well as the full path to the repeater question - this may either be a path or a question definition... Haven't thought it through enough to know yet.

For example:

```java
public class RepeaterContext {
  public ?? repeater;
  public ImmutableList<Integer> repeaterIndices;

  public RepeaterContext(?? repeater, ImmutableList<Integer> repeaterIndices) {
    this.repeater = repeater;
    this.repeaterIndices = repeaterIndices;
  }

  public Path hydratedPath(Path path) throws NotInTheRightContextException {
    // Combines repeater, repeaterIndices, and path to compute the hydrated path.
  }

  // Other helpers?
}


?? repeater = something that contains "applicant.household_members[].employers[]"

RepeaterContext repeaterContext = new RepeaterContext(repeater, ImmutableList.of(0,0));
```

kendrickt:
### Description
While building the `Block`s list, use recursion on a repeater block for all of its repeated blocks.

The variables used in recursion are:
* `blockIdSuffix`, which gets the repeated entity index appended to it
* `contextualPath`, which gets the repeater question path segment appended to it

### Checklist

### Issue(s)
Fixes #783


universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
```diff
@@ -87,6 +87,13 @@ public boolean isRepeated() {
     return repeaterId().isPresent();
   }
 
+  public QuestionDefinition getEnumerationQuestionDefinition() {
```
natsid, line 90: sweeeet

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
```diff
@@ -87,6 +87,13 @@ public boolean isRepeated() {
...
+    if (isRepeater()) {
+      return getQuestionDefinition(0);
+    }
+    throw new RuntimeException("Only a repeater block can have an enumeration question definition.");
```
natsid, line 94: nit/question: Should we say "Only a **enumeration** block can.."? Or should we save that update till we rename the methods for like `isEnumeration` etc?

natsid (commented):
LGTM, but I'll wait till test failures are resolved

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
kendrickt: I think we save this for one big PR later.

natsid (approved):
Wohoo!!!

Issue #673, opened 2021-04-03
QuestionDefinition Path should not be a constructor argument


kendrickt:
**Is your feature request related to a problem? Please describe.**
A QuestionDefinition's path should be defined by its name formatted, preceded by dot separated formatted names of the repeated entities it is being asked of. For example, a question asking the favorite color of each pet of each household member should have a path of `applicant.household_members[].pets[].favorite_color`.

Right now, the path is created by the `QuestionForm` and then passed in to `QuestionDefinition` constructors.

The `QuestionForm` shouldn't be responsible for defining the Path of QuestionDefinitions, and it should not be a constructor argument. It should be defined by properties of the QuestionDefinition itself.

**Describe the solution you'd like**
Issue #669 adds an explicit reference to the repeater question's ID for each repeated question. This reference should be used to define the question's path.

The question definition name is also being used to define path, but this may be prone to change by the admin. Instead, the `QuestionDefinition`s should have a special admin defined node name that cannot be updated and uniquely defines a question (sans versioning). This has the benefit that question versioning can be based on something less fickle than the question's display name (#641).

**Additional context**

**Done when**
Path is not a constructor argument, but rather something recursively defined by the repeaterId's path and a unique and invariant question name field. If repeaterId is not present, then the path is `applicant.<unique_and_invariant_quesiton_name>`.

davidrhallartificialeducation:
Picking this up

kendrickt:
This is actually almost done. Blocked by #665.

PR #920, opened 2021-04-29
Start question path cleanup

kendrickt:
### Description
* Instead of using `QuestionDefinition#getPath` to figure out what to display for repeater option in question edit form, use the question service.

* No longer care about path conflicts for questions. Instead rely on `QuestionDefinition#getQuestionPathSegment` and `QuestionDefinition.getRepeaterId` for conflicts.

* Question path is no longer an immutable member.  `QuestionDefinition#getQuestionPathSegment` is immutable instead.

Remaining uses of question path:
* the `getXPath` method in all the different `QuestionDefinition` subclasses, which is blocked from deletion because they are being used in CSV export.
* A couple places that exist because we need path elsewhere, marked with comments

### Issue(s)
Progress towards #673


universal-application-tool-0.0.1/test/repository/QuestionRepositoryTest.java
```diff
@@ -69,100 +68,61 @@ public void findPathConflictingQuestion_noConflicts_ok() throws Exception {
...
-      throws Exception {
-    Question householdMembers = testQuestionBank.applicantHouseholdMembers();
     QuestionDefinition newQuestionDefinition =
-        new QuestionDefinitionBuilder(householdMembers.getQuestionDefinition())
+        new QuestionDefinitionBuilder(applicantAddress.getQuestionDefinition())
             .clearId()
-            .setPath(householdMembers.getQuestionDefinition().getPath().withoutArrayReference())
-            .setQuestionType(QuestionType.TEXT)
+            .setName("applicant address!")
+            .setPath(Path.create("fake"))
```
hmhsu, line 83: Question (non-blocking): is this necessary or used at all?

hmhsu (approved):

universal-application-tool-0.0.1/test/repository/QuestionRepositoryTest.java
kendrickt: are you talking about this test case, or `setPath(Path.create("fake"))`?

The fake path is not useful. Hopefully very soon we can delete all paths from questions, including the `QuestionDefinitionBuilder#setPath` and so many other things. But for now, I think the path needs to be set or the builder won't work.

hmhsu: Ah I see. Thanks!

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -284,10 +298,12 @@ private DomContent formQuestionTypeSelect(QuestionType selectedType) {
    * question definition's repeater id.
    */
   private SelectWithLabel repeaterOptionFromQuestionDefinition(
-      QuestionDefinition questionDefinition) {
+      QuestionDefinition questionDefinition, Optional<String> maybeRepeaterName) {
```
natsid, line 301: It seems a little funny to me that this method ultimately still needs the questionDefinition _and_ the maybeRepeaterName. Makes me think we should should compute the name here instead of piping it through.

Do we do this because getting the name requires access to the service?

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -284,10 +298,12 @@ private DomContent formQuestionTypeSelect(QuestionType selectedType) {
    * question definition's repeater id.
```
natsid, line 298: Question (blocking): Should this field be disabled? This is the case where someone is editing a question and can't actually change the repeater, right? If this requires a bigger change than makes sense for this PR, just add a TODO.

natsid (approved):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: Computing the name requires access to the service, which is why it is computed way high up, and this optional string is passed through.

Wec could do the `String displayString = maybeRepeaterName.orElse(NO_REPEATER_DISPLAY_STRING)` earlier.

natsid: Cool cool. The suggested change makes sense to me, but don't feel strongly.

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
kendrickt: I'll do this in a follow-up PR.

natsid (approved):
Finally reloaded properly to see all your changes. Looks good!

PR #926, opened 2021-04-29
Disable dev file upload in staging temporarily.

hmhsu:
### Description
Temporarily disable fileupload probe in staging for emergent prod promotion.

### Checklist

### DIFF
--- a/browser-test/src/dev_file_upload.test.ts
+++ b/browser-test/src/dev_file_upload.test.ts
@@ -38,8 +38,9 @@ describe('the dev file upload page', () => {
         break;
       default:
         // Download the file and verify content in all other cases.
-        const fileContent = await downloadFile(page, 'file.txt');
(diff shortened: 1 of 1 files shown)


carolinedanzi (approved):

ndmckinley:
It's not a super bad emergency - would you mind taking an extra sec to disable it only in staging, I think we have a function for checking if our base_url is localhost... or is it broken for dev also?

ndmckinley (approved):

hmhsu:
> It's not a super bad emergency - would you mind taking an extra sec to disable it only in staging, I think we have a function for checking if our base_url is localhost... or is it broken for dev also?

Yes, it's a switch now. Because local/browser-env/ and staging all behave differently (ugh). I am only disabling the staging env.

ndmckinley:
🤦  I should have looked three lines above your code, yes, i just saw that.  ha.

PR #929, opened 2021-04-29
pluralize table so it properly deletes.

ndmckinley:
### Description
I changed the name of this table *almost* everywhere.  🤦 

This might be a problem since staging and prod might try (and refuse) to apply this down statement, due to the "never delete anything" posture we adopted a few weeks ago.  We'll see!

### Checklist

### Issue(s)


carolinedanzi (approved):
Thank you so much!

hmhsu (approved):

Issue #844, opened 2021-04-21
"Address" question type: implement Address Line 2


neel-a-mehta:
#398 added the bare bones of the Address question type; the only part left out was Address Line 2. Let's add that to the data model and the UI. See that issue for more context.

natsid:
Thanks for making this @neel-a-mehta !

PR #939, opened 2021-04-30
Add new `Scalar.LINE2` for Address question type and ability to read and write its value

natsid:
… value in QuestionAnswerer and read its value in AddressQuestion.

### Description
* Add ability to set the address line 2 value in `QuestionAnswerer`
* Add ability to read the address line 2 value in `AddressQuestion`

To satisfy #844, we still need to make changes to the address question renderer and form binding, as well as include it in the PO box validation check (see TODO in `AddressQuestion`).

### Checklist

### Issue(s)
Works toward #844

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/question/AddressQuestion.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/AddressQuestion.java
@@ -15,6 +15,7 @@ public class AddressQuestion implements PresentsErrors {
 
   private final ApplicantQuestion applicantQuestion;
   private Optional<String> streetValue;
+  private Optional<String> line2Value;
   private Optional<String> cityValue;
   private Optional<String> stateValue;
   private Optional<String> zipValue;
@@ -39,6 +40,8 @@ public class AddressQuestion implements PresentsErrors {
 
     if (definition.getDisallowPoBox() && getStreetValue().isPresent()) {
       Pattern poBoxPattern = Pattern.compile(PO_BOX_REGEX);
+      // TODO(https://github.com/seattle-uat/civiform/issues/844): Compare PO_BOX_REGEX against
+      // getLine2Value() as well.
       Matcher poBoxMatcher = poBoxPattern.matcher(getStreetValue().get());
 
       if (poBoxMatcher.matches()) {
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/services/applicant/question/AddressQuestion.java
```diff
@@ -200,6 +220,10 @@ private boolean isZipAnswered() {
    */
   @Override
   public boolean isAnswered() {
-    return isStreetAnswered() || isCityAnswered() || isStateAnswered() || isZipAnswered();
+    return isStreetAnswered()
+        || isLine2Answered()
```
hmhsu, line 224: Question (non-blocking): Does this mean the line 2 is required? Usually it is optional, I think.

hmhsu (approved):
Looks good. Thanks for making this change!

universal-application-tool-0.0.1/app/services/applicant/question/AddressQuestion.java
carolinedanzi: +1, will make this blocking - I think Hung-Ming is right that it should be optional

carolinedanzi (approved):

universal-application-tool-0.0.1/app/services/applicant/question/AddressQuestion.java
natsid: This does not make it required! I think this method might be kind of confusing... But basically, it's used to determine whether a question has been answered at all. Because we `||` (OR) all the `isAnswered` methods together, adding in `isLine2Answered()` doesn't actually effectively change the behavior here. All it does is say that if line 2 is answered at any point, this method will now evaluate to `true`, even if all other fields are not answered.

AlaseI: Does that take away the requiredness of other parts viz street, city and zip?

carolinedanzi: Thanks for the explanation! That does help a lot - looks good to me!

natsid: Great question @AlaseI . `isAnswered` is not used to determine validation for required fields at all. I know this is sort of confusing. 😛 

`getAllTypeSpecificErrors` is the method that is used to check for those errors. `getAllTypeSpecificErrors` then checks for errors for each of the required fields. But notice that it doesn't check for line 2 to be answered.

AlaseI (approved):
Thanks!

PR #948, opened 2021-04-30
"Address line 2": add prompt in applicant view and save answer

### LINKED ISSUE #844: "Address" question type: implement Address Line 2

natsid:
### Description
This is the last step to make Address line 2 functional.

One final piece is to include Address line 2 when validating that an address is not a PO Box. That will be done in a follow-up PR.

### Checklist

### Issue(s)
Works toward #844


universal-application-tool-0.0.1/conf/messages
```diff
@@ -53,14 +53,16 @@ content.description=CiviForm lets you apply for many benefits at once by reusing
 #----------------------------------------------------------#
 
 # Address question labels and placeholder text
-label.street=Street address
-placeholder.street=Enter your street address
+label.street=Address line 1
+placeholder.street=Street address
+label.line2=Address line 2
```
kendrickt, line 58: (nit) idk what the standard here is. Could we add `address` in here somewhere so its really obvious what line2 refers to?

natsid: That's a good point - it's really unclear as is. Will improve.

hmhsu (approved):

PR #950, opened 2021-04-30
Allow translation of validation errors for applicants

### LINKED ISSUE #820: Multi-option questions are missing field errors
**Describe the bug**
We should use `fieldErrors` in question renderers, but I forgot to include these for multi-option questions

**Expected behavior**
Multi-option questions (right now, only checkboxes can have errors) should display validation errors to the applicant

### LINKED ISSUE #921: Audit the applicant pages for untranslated strings
The `conf/messages` files are used to provide translations of app text, such as buttons and titles. We should audit our applicant views to make sure all public strings use the message file rather than hard-coded strings.

**Done when**
All applicant views use `Messages` rather than hardcoded strings for publicly-visible text

carolinedanzi:
### Description
1. Transfer the validation text to messages files
2. Add a `MessageKey` enum that has constants for the keys into messages files, with a TODO to use these elsewhere
3. Update `ValidatioErrorMessage` to contain a `MessageKey` and any message arguments. We do not have access to `Messages` at the time validation errors are created, so we use a wrapper class to contain info needed to build a message in the renderer later
4. Add all error-related methods to the `PresentsErrors` interface and update applicant questions
5. Adds missing `fieldErrors` to multi-option question renderers

### Checklist

### Issue(s)
Fixes #921 
Fixes #820


hmhsu (approved):
Nice work!

Issue #405, opened 2021-03-10
"Repeater" question type: implement bare bones


neel-a-mehta:
The Repeater Question type will be used to ask applicants for an arbitrary number of repeated entities (children, household members, jobs, cars, etc.). Admins will be able to create Repeated Questions for each Repeater Question (#612), and add these to program blocks. Applicants will be asked Repeated Questions for each repeated entity the applicant lists (#613).

![image](https://user-images.githubusercontent.com/77128518/110679068-0b0e3700-81a5-11eb-97e2-13dd85759eaa.png)

In the future it’d be great to import the text typed into these fields into questions in the repeated blocks -- for instance, applicants type their children’s first names into the multi-text question, and those first names auto-fill into the Name questions when it comes time to fill in details about each child. But that adds a lot of complexity to the UAT admin experience (they have to drill deep into the data model to show what field the text inputs correlate with), so it’s better to leave off for v1. 

(Also, there are edge cases where two household members have the same name, so auto-filling information might get messy.)

Make it possible for UAT admins to create, and for applicants to fill out, questions of this type. Answers should be saved to a database. Just implement the sub-fields; configuration options, validation criteria, and custom functionality aren't necessary yet.

See details at https://github.com/seattle-
uat/universal-application-tool/wiki/Question-type-requirements-notes

See more details at 
https://docs.google.com/document/d/1jkbFPzG_whZ1AqZVmxmkGuoqR5wRdG817COX5usPAWo/edit#heading=h.vkhepsgmicsf

neel-a-mehta:
Subfields: 

Nickname fields (required)
- Implement each as a required text input field.
- Applicants can add/delete fields. There can never be less than 1 field; if the applicant has 0 items to report, they can just hit “skip” instead.

neel-a-mehta:
[See this wiki page](https://github.com/seattle-uat/civiform/wiki/Adding-a-new-question-type) to learn how to add a new question type to the codebase!

lianne-oy:
In progress, moving to M5.

Issue #919, opened 2021-04-29
Rename `Repeater` to `Enumerator` everywhere.


kendrickt:
`Repeater` is used in a lot of places, which is easily confused with `Repeated`.

Solution: use `Enumerator` (e.g. `EnumeratorQuestion`, `QuestionType.ENUMERATOR`), and continue to use `Repeated` (e.g. `isRepeated()`).

PR #951, opened 2021-05-01
Enable simple end-to-end enumerator functionality.

### LINKED ISSUE #613: Handle repeated questions in ApplicantProgramBlockController.
Repeated questions are questions that are children of a repeated entity. The repeated questions need to be properly orchestrated to the Applicant.

**Additional context**
Repeater questions will be enabled in #405 and repeated questions will be admin configurable in #612. 

**Done when**
The applicant can answer repeated questions for each repeated entity defined by a repeater question.

### LINKED ISSUE #952: be able to delete repeated entities
implement the delete functionality on form submission with the delete checkboxes for enumerator questions.

kendrickt:
### Description
Enable simple end-to-end enumerator and repeated entities functionality.

Changes to the enumerator question renderer:
* Old answers are now differentiated from new enumerator fields
  * old answers have a checkbox field to delete them on form submission
  * new enumerator fields have an `x` button to delete them dynamically
* add a public method to generate a template with the right contextualized path
* added a hidden div for placeholder text so it can be added to new enumerator fields

Split apart `ApplicantServiceImpl#stageUpdates` to update enumerator answers differently.

Fix `ApplicantQuestion#getUpdatedInProgramMetadata` bug for enumerator metadata. 

Memoize `ReadOnlyApplicantProgramServiceImpl#getAllBlocks`. 

Add `ApplicantData#deleteRepeatedEntity` that deletes indices in reverse sorted order.

### Screenshot
Differentiating previously answered vs new enumator answers.
image

### Issue(s)
Fixes #613, Fixes #952, Progress towards #405


browser-test/src/support/admin_questions.ts
```diff
@@ -305,37 +315,19 @@ export class AdminQuestions {
...
 
-  async addRepeaterQuestion(questionName: string,
-    description = 'repeater description',
-    questionText = 'repeater question text',
-    helpText = 'repeater question help text') {
+  async addEnumeratorQuestion(questionName: string,
+    description = 'enumerator description',
+    questionText = 'enumerator question text',
+    helpText = 'enumerator question help text',
+    enumeratorName = 'does not repeat') {
```
carolinedanzi, line 322: question: why does this default to 'does not repeat'? Does that mean the other questions that are by default added to the 'does not repeat' enumerator will instead be added to this one if it exists?

carolinedanzi: Never mind, I understand now! This enumerator by default does not repeat - maybe we could add a comment to explain this one?

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -379,6 +375,38 @@ private void putParentArray(Path path) {
     return readString(path);
   }
 
+  /**
+   * Delete the ENTIRE repeated entity for each entity specified by the path to an array element and
+   * list of indices.
```
carolinedanzi, line 380: nit: please add comment explaining return value (under what cases is it true/false)

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -179,6 +181,104 @@ private void stageUpdates(
...
+                        .equals(Path.empty().join(Scalar.DELETE_ENTITY)))
+            .collect(ImmutableSet.toImmutableSet());
+
+    // If there are more updates than there are adds/changes and deletes, throw
+    Set<Update> unknownUpdates = Sets.difference(updates, Sets.union(addsAndChanges, deletes));
+    if (!unknownUpdates.isEmpty()) {
+      throw new PathNotInBlockException(block.getId(), unknownUpdates.iterator().next().path());
+    }
+
+    for (Update update : addsAndChanges) {
```
carolinedanzi, line 220: nit: could we add a comment reminder that changes need to happen first,  before deletions possibly change the original index?

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -179,6 +181,104 @@ private void stageUpdates(
...
+                                .getContextualizedPath()
+                                .withoutArrayReference()))
+            .collect(ImmutableSet.toImmutableSet());
+
+    // Early return if it is empty.
+    if (entityUpdates.isEmpty()) {
+      return entityUpdates;
+    }
+
+    // Check that the entity updates have unique and consecutive indices
```
carolinedanzi, line 259: question: why do they need consecutive indices? What if I have 5 children and need to update only the names for #2 and #5?

universal-application-tool-0.0.1/app/services/applicant/Block.java
```diff
@@ -66,6 +66,20 @@ public String getDescription() {
     return blockDefinition.description();
   }
 
+  /** This block is an enumerator block if its {@link BlockDefinition} is an enumerator. */
+  public boolean isEnumerator() {
```
carolinedanzi, line 70: question: is there a rename in the future to standardize "repeater"/"enumerator"?

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -17,6 +17,8 @@
   /**
    * Get the {@link Block}s this applicant needs to fill out or has filled out for this program.
    *
+   * <p>All blocks where {@link Block#isEnumerator()} are <b>always</b> in-progress.
```
carolinedanzi, line 20: question: why is this the case? Couldn't we just check whether they've seen this question before, and then show them that question in the review page to make sure they don't need to change anything?

universal-application-tool-0.0.1/app/services/applicant/ValidationErrorMessage.java
```diff
@@ -22,6 +22,8 @@
   public static String INVALID_ZIP = "Please enter valid 5-digit ZIP code.";
   public static String NO_PO_BOX = "Please enter a valid address. We do not accept PO Boxes.";
 
+  public static String ENTITY_NAME_REQUIRED = "Please enter a value for each entity.";
```
carolinedanzi, line 25: suggestion (non-blocking): "entity" may be confusing for applicants - we can instead change this to the placeholder text , like `String.format("Please enter a value for each %s", placeholder)`

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
```diff
@@ -93,11 +93,28 @@ public boolean hasErrors() {
   }
 
   public Optional<Long> getUpdatedInProgramMetadata() {
-    return applicantData.readLong(getContextualizedPath().join(Scalar.PROGRAM_UPDATED_IN));
+    Path contextualizedMetadataPath = getContextualizedPath().join(Scalar.PROGRAM_UPDATED_IN);
+
+    // Metadata for enumerators is stored for each JSON array element, but we rely on metadata for
+    // the first one.
```
carolinedanzi, line 99: question: to clarify, this is *just* for the enumerator question, correct? A repeated question will have its own metadata separate from the entity?

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -17,16 +20,17 @@
...
+  private static final String PLACEHOLDER_ID = "enumerator-placeholder-text";
+  private static final String ENUMERATOR_FIELD_CLASS = "enumerator-field";
 
   // TODO(#859): make this admin-configurable
   private static final String PLACEHOLDER = "Placeholder";
 
-  private static final String REPEATER_FIELD_CLASSES =
-      StyleUtils.joinStyles(REPEATER_FIELD_CLASS, Styles.FLEX, Styles.FLEX_ROW, Styles.MB_4);
+  public static final String ENUMERATOR_FIELD_CLASSES =
+      StyleUtils.joinStyles(ENUMERATOR_FIELD_CLASS, Styles.FLEX, Styles.FLEX_ROW, Styles.MB_4);
```
carolinedanzi, line 33: nit: ENUMERATOR_FIELD_CLASS could go in `ReferenceClasses`

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +40,68 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
   @Override
   public Tag render(Messages messages) {
-    ContainerTag repeaterFields = div().withId(REPEATER_FIELDS_ID);
-    // TODO: add each answer as a repeaterField
-    repeaterFields.with(repeaterField(PLACEHOLDER, Optional.empty(), 0));
+    RepeaterQuestion enumeratorQuestion = question.createEnumeratorQuestion();
+    ImmutableList<String> entityNames = enumeratorQuestion.getEntityNames();
+
+    ContainerTag enumeratorFields = div().withId(ENUMERATOR_FIELDS_ID);
+    int index;
```
carolinedanzi, line 47: super nit: any reason why index can't be declared inside the for loop? I don't see it used afterwards

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +40,68 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
+      enumeratorFields.with(existingEnumeratorField(Optional.of(entityNames.get(index)), index));
+    }
 
     return div()
         .withClasses(Styles.MX_AUTO, Styles.W_MAX)
         .with(
             div()
                 .withClasses(ReferenceClasses.APPLICANT_QUESTION_TEXT)
                 .withText(question.getQuestionText()),
+            div().withId(PLACEHOLDER_ID).withClass(Styles.HIDDEN).withText(PLACEHOLDER),
```
carolinedanzi, line 58: question: not sure I understand this, why is the placeholder hidden?

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +40,68 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
+    return div().withClasses(ENUMERATOR_FIELD_CLASSES).with(optionInput, removeOptionBox);
+  }
+
+  /**
+   * Create an enumerator field template for new entries. These come with a button to delete itself.
+   */
+  public static Tag newEnumeratorFieldTemplate(Path contextualizedPath) {
+    ContainerTag optionInput =
+        FieldWithLabel.input()
+            .setFieldName(contextualizedPath.toString())
```
carolinedanzi, line 98: question: do we still want the placeholder here?

carolinedanzi (commented):
This is looking good so far! I'm behind on the enumerator work so I have a bunch of questions - happy to meet to discuss if that's easier

Will review tests next

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
kendrickt: The form will submit all the entity names, so we expect consecutive indices from `0` to `len-1`. Is there a better way to do this?

universal-application-tool-0.0.1/app/services/applicant/Block.java
kendrickt: yes! #919

browser-test/src/support/admin_questions.ts
kendrickt: I can add a small comment here, but these arguments match all the other `addXQuestion` methods and I don't think they all need a small comment.

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
kendrickt: done

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramService.java
kendrickt: This was a UX thing I thought we were doing... but maybe I just assumed it was.

universal-application-tool-0.0.1/app/services/applicant/ValidationErrorMessage.java
kendrickt: dibe!

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: done!

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: This is hacky:

1. This is a hidden div that holds the placeholder text.
2. The template is created outside of the scope of this instance, so the method that makes the template is `static` and has no reference to question-specific admin-configurable placeholder text.
3. The main.ts method that copies the template finds the placeholder div and puts it in the template.

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: No, the placeholder gets set in the `main.ts` method. See the other comment for why I did it this way.

Super happy to change it if there's a more straightforward way to do this.

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
kendrickt: yes

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: it can be! originally this index was used to create new enumerator fields but that's changed now.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
kendrickt: done

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -379,6 +375,40 @@ private void putParentArray(Path path) {
...
+
+    // Make sure there are enough things to delete by checking the first index in the reverse sorted
+    // list
+    if (!hasPath(path.atIndex(reverseSortedIndices.get(0)))) {
+      return false;
+    }
+
+    // Delete in reverse sorted order because deletion is index based, and indices would need to be
+    // decremented
+    // for each deletion if it wasn't reverse sorted.
```
natsid, line 405:
```suggestion
    // decremented for each deletion if it wasn't reverse sorted.
```

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramService.java
natsid: I don't recall if/what the UX decision was for this. I could see either way...

universal-application-tool-0.0.1/app/services/applicant/question/RepeaterQuestion.java
```diff
@@ -1,20 +1,26 @@
...
+import services.applicant.ValidationErrorMessage;
 import services.question.types.QuestionType;
 import services.question.types.RepeaterQuestionDefinition;
 
 public class RepeaterQuestion implements PresentsErrors {
 
   private final ApplicantQuestion applicantQuestion;
 
+  // TODO(#859): make this admin-configurable
+  private final String PLACEHOLDER = "placeholder";
```
natsid, line 14: Since this isn't implemented yet (and it's possible we won't get to it), I think it'd be better to leave this as an empty string for now. The string "placeholder" doesn't add any value and might just be more confusing than the empty string.

universal-application-tool-0.0.1/app/services/applicant/question/RepeaterQuestion.java
```diff
@@ -45,4 +59,15 @@ public boolean isAnswered() {
         .getApplicantData()
         .hasPath(applicantQuestion.getContextualizedPath().atIndex(0).join(Scalar.ENTITY_NAME));
   }
+
+  /** Return the repeated entity names associated with this enumerator question. */
+  public ImmutableList<String> getEntityNames() {
```
natsid, line 64: (blocking) This class is complex enough now that we should add a test class to it like [the other question classes have](https://github.com/seattle-uat/civiform/tree/main/universal-application-tool-0.0.1/test/services/applicant/question).

universal-application-tool-0.0.1/app/services/applicant/Block.java
```diff
@@ -66,6 +66,20 @@ public String getDescription() {
...
+    return blockDefinition.isRepeater();
+  }
+
+  /** Get the enumerator {@link ApplicantQuestion} for this enumerator block. */
+  public ApplicantQuestion getEnumeratorQuestion() {
```
natsid, line 75: Add a couple of tests for this method

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +37,70 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
 
     return div()
         .withClasses(Styles.MX_AUTO, Styles.W_MAX)
         .with(
             div()
                 .withClasses(ReferenceClasses.APPLICANT_QUESTION_TEXT)
                 .withText(question.getQuestionText()),
+            div()
+                .withId(PLACEHOLDER_ID)
+                .withClass(Styles.HIDDEN)
```
natsid, line 56: question: What is this div for and why is it hidden?

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +37,70 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
                     Styles.TEXT_BASE,
                     Styles.FONT_THIN,
                     Styles.MB_2)
                 .withText(question.getQuestionHelpText()),
-            repeaterFields,
-            button(ADD_ELEMENT_BUTTON_ID, messages.at("button.addRepeaterEntity")))
-        .withType("button");
+            enumeratorFields,
+            button(ADD_ELEMENT_BUTTON_ID, messages.at("button.addEnumeratorEntity"))
+                .withType("button"),
```
natsid, line 67: I don't think `withType` is necessary here since [the underlying `button` method](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/views/BaseHtmlView.java#L50) sets the type.

I'm curious if this `withType` call was here for some other purpose though, since it looks like it was previously modifying a div rather than this button, based on the parens...

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -36,43 +37,70 @@ public RepeaterQuestionRenderer(ApplicantQuestion question) {
...
+  /**
+   * Create an enumerator field template for new entries. These come with a button to delete itself.
+   */
+  public static Tag newEnumeratorFieldTemplate(Path contextualizedPath) {
+    ContainerTag optionInput =
+        FieldWithLabel.input()
+            .setFieldName(contextualizedPath.toString())
+            .getContainer()
+            .withClasses(Styles.FLEX, Styles.ML_2);
+    Tag removeFieldButton = button("x").withType("button").withClasses(Styles.FLEX, Styles.ML_4);
```
natsid, line 100: For a11y reasons, I'd suggest using "remove" or "delete" here instead of "x". Alternatively, you could continue using an "icon" (though I think ideally we'd use an actual icon rather than the letter x), and add an `aria-label`. I think in j2html that'll probably just be something like `.withAttr("aria-label", "Remove entity")`. 

In general we need to do a big sweep to add aria-labels and alt-text to our elements - sorry for picking on this element 😄  Just noticed an easy spot to improve things now.

browser-test/src/support/admin_questions.ts
```diff
@@ -19,12 +19,14 @@ export class AdminQuestions {
   async fillInQuestionBasics(questionName: string,
     description: string,
     questionText: string,
-    helpText: string) {
+    helpText: string,
+    enumeratorName: string) {
```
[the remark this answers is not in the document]
kendrickt: All the `addXQuestion` methods have default values and they all call this underlying one. I think I'll make the default values constants that are shared across the `addXQuestion` methods, and use a default again here.

universal-application-tool-0.0.1/app/services/applicant/question/RepeaterQuestion.java
kendrickt: fair enough!

browser-test/src/support/admin_questions.ts
carolinedanzi: No worries! Sounds good

universal-application-tool-0.0.1/app/services/applicant/Block.java
carolinedanzi: Awesome!

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
carolinedanzi: Thanks!

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: fixed!

Really not sure where this came from 😦

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: Fixed!!! This is much more straightforward now.

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: Fixed!!! This is much more straightforward now.

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
kendrickt: Fixed!!! This is much more straightforward now.

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
kendrickt: Added a comment to clarify this.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    await adminQuestions.addNameQuestion('enumerator-ete-name');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-enumerator');
+    await adminQuestions.addNameQuestion('enumerator-ete-enumerator-name', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-enumerator-enumerator', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
+    await adminQuestions.addTextQuestion('enumerator-ete-enumerator-enumerator-text', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator-enumerator');
+
+    await adminPrograms.addProgram(programName);
+    await adminPrograms.editProgramBlock(programName, 'ete enumerator program description');
+
+    // All non-repeated questions should be available in the question bank.
```
carolinedanzi, line 23: praise: thank you for these comments!

carolinedanzi (commented):
This is looking great! Just need to add more tests and then it's good to go

universal-application-tool-0.0.1/app/services/applicant/question/RepeaterQuestion.java
kendrickt: adding tests!

universal-application-tool-0.0.1/app/services/applicant/Block.java
kendrickt: adding!

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramService.java
kendrickt: ended up getting rid of this. not hard to add it back later.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  it('Create nested enumerator and repeated questions', async () => {
+    const { browser, page } = await startSession();
+    page.setDefaultTimeout(4000);
+
+    await loginAsAdmin(page);
+    const adminQuestions = new AdminQuestions(page);
+    const adminPrograms = new AdminPrograms(page);
+
+    await adminQuestions.addNameQuestion('enumerator-ete-name');
```
AWBaumann, line 14: Nit. Building off of this string would probably be more readable.

```
const enumeratorBase = 'enumerator-ete';
const enumerator = enumeratorBase + '-enumerator';
const nestedEnumerator = enumerator + '-enumerator';
...etc
```

universal-application-tool-0.0.1/conf/messages.en-US
```diff
@@ -45,3 +46,4 @@ validation.stateRequired=Please enter state.
 validation.zipcodeRequired=Please enter valid ZIP code.
 validation.invalidZipcode=Please enter valid 5-digit ZIP code.
 validation.noPoBox=Please enter a valid address. We do not accept PO Boxes.
+validation.entityNameRequired=Please enter a value for each {0}.
```
natsid, line 49: (non-blocking): Why not add the es-US version of this one?

universal-application-tool-0.0.1/conf/messages.en-US
```diff
@@ -3,6 +3,7 @@
 button.logout=Logout
 button.apply=Apply
 button.nextBlock=Save and continue
+button.deleteEntity=Delete entity
```
natsid, line 6: I don't see this one in the default `messages` file. Does it need to be added? Also, why are the deleteEntity and addEnumeratorEntity button messages separated - should they be grouped together?

natsid (approved):

universal-application-tool-0.0.1/app/views/questiontypes/RepeaterQuestionRenderer.java
```diff
@@ -54,26 +60,53 @@ public Tag render(ApplicantQuestionRendererParams params) {
...
+            .getContainer()
+            .withClasses(Styles.FLEX, Styles.ML_2);
+    ContainerTag icon =
+        Icons.svg(Icons.TRASH_CAN_SVG_PATH, 24)
+            .withClasses(Styles.FLEX_SHRINK_0, Styles.H_12, Styles.W_6);
+    Tag removeEntityButton =
+        TagCreator.button(icon)
+            .withType("button")
+            .withClasses(Styles.FLEX, Styles.ML_4)
+            .attr("aria-label", messages.at("button.deleteEntity"));
```
carolinedanzi, line 106: praise: thank you for making this i18n-friendly!

universal-application-tool-0.0.1/app/views/style/ReferenceClasses.java
```diff
@@ -3,24 +3,26 @@
 /** Non-styled classes that point to different components. */
 public final class ReferenceClasses {
```
carolinedanzi, line 4: praise: thank you for alphabetizing these!

universal-application-tool-0.0.1/conf/messages
```diff
@@ -17,6 +17,9 @@ button.apply=Apply
 # The button text for saving the current applicant-entered data and continuing to the next block in the form.
 button.nextBlock=Save and continue
 
+# The aria text on the button used by an applicant to mark an entity for deletion.
+button.deleteEntity=Delete entity
```
carolinedanzi, line 21: suggestion (non-blocking): Could this just be "Delete"? If we change the placeholder eventually to not be "entity", it might be odd for an applicant to see this word only for the delete aria text

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    await applicantQuestions.saveAndContinue();
+
+    // Answer two nested repeated text questions
+    await applicantQuestions.answerTextQuestion("hello");
+    await applicantQuestions.saveAndContinue();
+    await applicantQuestions.answerTextQuestion("world");
+    await applicantQuestions.saveAndContinue();
+
+    // Make sure the enumerator answers are in the review page
+    expect(await page.innerText("body")).toContain("first name");
```
AWBaumann, line 104: Nit. I feel like we should have a helper function for this. (I'm surprised jest doesn't already.)

`expectContains(body, ["first name", "last name"...]);`

AWBaumann (approved):

universal-application-tool-0.0.1/test/services/applicant/ApplicantServiceImplTest.java
```diff
@@ -161,6 +161,40 @@ public void stageAndUpdateIfValid_rawUpdatesContainMultiSelectAnswers_isOk() {
...
+            .toCompletableFuture()
+            .join();
+
+    assertThat(errorAnd.isError()).isFalse();
+    assertThat(errorAnd.getResult()).isInstanceOf(ReadOnlyApplicantProgramService.class);
+
+    ApplicantData applicantDataAfter =
+        userRepository.lookupApplicantSync(applicant.id).get().getApplicantData();
+
+    assertThat(applicantDataAfter.readRepeatedEntities(enumeratorPath)).containsExactly("second");
```
carolinedanzi, line 195: question: I think I'm just misunderstanding, but why do we only expect "second"? I thought 1 and 0 are deleted, which just leaves "third"

universal-application-tool-0.0.1/test/services/applicant/question/EnumeratorQuestionTest.java
```diff
@@ -0,0 +1,110 @@
...
+            enumeratorQuestionDefinition, applicantData, ApplicantData.APPLICANT_PATH);
+    QuestionAnswerer.answerEnumeratorQuestion(
+        applicantData, applicantQuestion.getContextualizedPath(), ImmutableList.of(value));
+
+    RepeaterQuestion enumeratorQuestion = new RepeaterQuestion(applicantQuestion);
+
+    assertThat(enumeratorQuestion.isAnswered()).isTrue();
+    assertThat(enumeratorQuestion.getEntityNames()).contains(value);
+    assertThat(enumeratorQuestion.hasTypeSpecificErrors()).isFalse();
+    assertThat(enumeratorQuestion.hasQuestionErrors()).isTrue();
```
carolinedanzi, line 90: nit: consider checking the error message, too

carolinedanzi (approved):
Thank you for adding tests! This looks great!

browser-test/src/end_to_end_enumerators.test.ts
kendrickt: Since the strings are so important in browser tests, I like keeping them as raw strings. Then there's no logic involved in trying to figure out what the expected strings should be, or how the questions are linked to their enumerators.

browser-test/src/end_to_end_enumerators.test.ts
kendrickt: There's https://github.com/playwright-community/expect-playwright#why-do-i-need-it but I don't want to add dependencies...

universal-application-tool-0.0.1/conf/messages.en-US
kendrickt: This is online 21 in `conf/messages`. 

There's gonna be some clean up to organize all the messages, so I'm just going to leave the messages where they are for now.

universal-application-tool-0.0.1/conf/messages.en-US
kendrickt: done!

universal-application-tool-0.0.1/conf/messages
kendrickt: Changed. I also made the `placeholder` a `Map<Locale, String>` so we can i18n it more easily.

universal-application-tool-0.0.1/test/services/applicant/question/EnumeratorQuestionTest.java
kendrickt: done

universal-application-tool-0.0.1/test/services/applicant/ApplicantServiceImplTest.java
kendrickt: The delete array will always have indices 0...n where n is the number of things the applicant wants to delete, and we use the values of the array to determine what index to actually delete.

So `delete = [2, 0]` (on line 180 and 181), which means we delete whatever is at index 0 and index 2, which leaves "second" at index one.

PR #969, opened 2021-05-04
Remove dev_file_upload download test

hmhsu:
### Description
This test is flaky and it is testing a dev page that is not critical. Since it is causing more harm than providing value, removing the test. We will have the actual file upload question ready soon so this is fine.

### Checklist


ndmckinley (approved):

hmhsu:
hm, the test failure is due to having 2 draft programs issue I think.

PR #985, opened 2021-05-05
Delete Path#path.

### LINKED ISSUE #606: Fix all usages of `services.Path.path()`.

kendrickt:
### Description
No more `Path#path()`. Use `Path#toString` for JSON stuff, and just use `Path` and let string formatting call `toString`. 

### Issue(s)
Fixes #606

### DIFF
--- a/universal-application-tool-0.0.1/app/models/Question.java
+++ b/universal-application-tool-0.0.1/app/models/Question.java
@@ -149,7 +149,7 @@ public class Question extends BaseModel {
 
     // TODO(https://github.com/seattle-uat/civiform/issues/673): delete this when questions don't
     //  need paths
-    path = questionDefinition.getPath().path();
+    path = questionDefinition.getPath().toString();
 
     repeaterId = questionDefinition.getRepeaterId().orElse(null);
     name = questionDefinition.getName();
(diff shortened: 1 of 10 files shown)


carolinedanzi (approved):
Thanks Kendrick!

PR #987, opened 2021-05-05
List the incomplete translation messages!

### LINKED ISSUE #983: Add a script to check which messages need translation
It would be nice if we could create a bash script that analyzes all the `messages` files and logs which message keys need translations in each language

**Done when**
We have a way to quickly check which messages are missing translations

kendrickt:
### Description
Add a bash script to diff the message translation files. 

Without any arguments, this script will generate a diff of each localized message
file with the default `messages` file. Missing translations will be presented
first, followed by weird stuff. The weird stuff may be
  1. message keys in the localized file that do not exist in the default file
  2. duplicated message keys

You can run this script with a single argument like "en-US" which will generate the
diff of just that one translation.

Example output:

List all the missing messages
image

List missing messages for one localization
image

File not found message
image

### Issue(s)
Fixes #983

### DIFF
--- a/bin/list-incomplete-messages
+++ b/bin/list-incomplete-messages
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Without any arguments, this script will generate a diff of each localized message
+# file with the default `messages` file. Missing translations will be presented
+# first, followed by weird stuff. The weird stuff may be
+#   1. message keys in the localized file that do not exist in the default file
+#   2. duplicated message keys
+#
+# You can run this script with a single argument like "en-US" which will generate the
+# diff of just that one translation.
+#
+# Example output:
+#  There are no translations in universal-application-tool-0.0.1/conf/messages.am
+#  Missing translations in universal-application-tool-0.0.1/conf/messages.en-US:
+#    content.selectLanguage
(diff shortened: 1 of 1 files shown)


ndmckinley (approved):

carolinedanzi (approved):
This is awesome! Thanks Kendrick!

bin/list-incomplete-messages
```diff
@@ -0,0 +1,73 @@
...
+
+get_messages() {
+  regex="^[a-z][^=]*"
+  echo "$(grep -o ${regex} $1 | sort)"
+}
+
+diff_messages() {
+  if [[ ! -z $2 ]]
+  then
+    echo "Missing translations in $3:";
```
hmhsu, line 34: Suggestion (non-blocking): maybe you can check if difference is empty and print there is nothing missing here.

hmhsu (approved):
This is great! Thanks for doing this.

bin/list-incomplete-messages
kendrickt: I was having trouble getting an "empty diff" to really show up as empty when the file was empty.

This works and does so I'm just going to leave it.

Issue #578, opened 2021-03-22
Basic styling for Applicant experience


natsid:
Add some styling to the Applicant pages. We should just do a basic initial pass to make the pages usable, even without any high fidelity UX mocks.

The pages that come to mind right now that need styling are:
* program list
* program block forms

**Done when** the Applicant experience is understandable and usable by a random outside user.

lianne-oy:
Made some progress; but not complete.  Moving to M5.

lianne-oy:
Still need to define e2e applicant experience.

neel-a-mehta:
Here's a mock of what the applicant experience can look like with Seattle styling. May want to coordinate with Aaron to settle on final styling: April Showers Bring June Launches (tm)

PR #991, opened 2021-05-05
Applicant UI - Update checkbox renderers

AWBaumann:
### Description
Updated the checkbox renderer to reflect general styling. Also fixed an issue where the user couldn't unsave checked options.

### Checklist

### Issue(s)
Part of #578


universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -167,7 +167,9 @@ public void putRepeatedEntities(Path path, ImmutableList<String> entityNames) {
   }
 
   private void putNull(Path path) {
-    put(path, null);
+    if (!path.isArrayElement()) {
```
hmhsu, line 170: Thoughts (blocking): would this make more sense to check ifNull at line 192 and only add the value if not null there? The thinking I have is that putNull is an indication that the user answers the question with an empty answer, v.s. the user did not answer the question.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -204,6 +206,13 @@ private void putArrayIfMissing(Path path) {
     }
   }
 
+  /** Clears an array in preparation of updates. */
+  public void maybeClearArray(Path path) {
+    if (path.isArrayElement()) {
```
hmhsu, line 211: Thoughts (blocking): should we add `checkLocked()` here as well?

hmhsu: nvm, `put()` has it already.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
AWBaumann: I think there may be valid uses for putting null in the applicant data, but for arrays we should have empty arrays instead.

hmhsu (approved):

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
hmhsu: Yeah, I meant it might still be useful to insert an empty array `[]` here. The behavior of the code would be to do nothing, which doesn't distinguish between answering with an empty answer or not answering at all.

PR #992, opened 2021-05-05
Simplify how metadata conflicts are caught.

kendrickt:
### Description
It is much more straightforward now.

Also, I organized `Scalar` enum and put a comment on it so hopefully people only append to it.


natsid (approved):
Awesome update!! Thanks @kendrickt !

PR #1017, opened 2021-05-05
Allow admin translation of multi-option question answers

carolinedanzi:
### Description
1. Refactored translation-related forms into `forms/translation/` package
2. Added a `MultiOptionQuestionTranslationForm` that extends `QuestionTranslationForm` and adds a field for question option translations
2. Updates `QuestionTranslationEditView` to include fields for multi-option questions, if they exist
3. Removes the rule that question options must have the same supported locales for each option. This is because an admin could (for example) add a new option that has not yet been translated - we should allow this, even though in this state we may have 3 options with full translation, and then a new option with only the default locale.

NOTE: If an admin edits the question options for an existing question, we keep existing translations for strings that are unchanged. If an option was changed, there are no translations, but they will default to English in the applicant view

### Checklist

### Issue(s)
#778 

Screen Shot 2021-05-18 at 1 52 24 PM


universal-application-tool-0.0.1/test/services/question/types/MultiOptionQuestionDefinitionTest.java
```diff
@@ -99,4 +123,30 @@ public void getOptionsForLocale_returnsAllTranslations() throws Exception {
             LocalizedQuestionOption.create(1L, "one", Locale.US),
             LocalizedQuestionOption.create(2L, "two", Locale.US));
   }
+
+  @Test
+  public void getOptionsForLocaleOrDefault_returnsBothLocalizedAndDefault() throws Exception {
```
natsid, line 128: praise: Good test and good behavior here!

universal-application-tool-0.0.1/app/forms/translation/MultiOptionQuestionTranslationForm.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import java.util.Locale;
+import services.LocalizedStrings;
+import services.question.QuestionOption;
+import services.question.exceptions.UnsupportedQuestionTypeException;
+import services.question.types.MultiOptionQuestionDefinition;
+import services.question.types.QuestionDefinition;
+import services.question.types.QuestionDefinitionBuilder;
+
+public class MultiOptionQuestionTranslationForm extends QuestionTranslationForm {
```
natsid, line 14: praise: Nice work with QuestionTranslationForm and this extension! This makes a lot of sense to me.

natsid (approved):
Great work, @carolinedanzi !

universal-application-tool-0.0.1/app/forms/translation/MultiOptionQuestionTranslationForm.java
carolinedanzi: Thanks Natalie! I modeled it off your work on the QuestionForms

Issue #404, opened 2021-03-10
"File upload" question type: implement bare bones


neel-a-mehta:
This generates a widget that lets applicants upload files from their devices or from camera rolls. Upload from cloud storage services will be considered out of scope for v1.

![image](https://user-images.githubusercontent.com/77128518/110679203-3729b800-81a5-11eb-8309-abf464af980c.png)

Make it possible for UAT admins to create, and for applicants to fill out, questions of this type. Answers should be saved to a database. Just implement the sub-fields; configuration options, validation criteria, and custom functionality aren't necessary yet.

See details at https://github.com/seattle-uat/universal-application-tool/wiki/Question-type-requirements-notes

neel-a-mehta:
[See this wiki page](https://github.com/seattle-uat/civiform/wiki/Adding-a-new-question-type) to learn how to add a new question type to the codebase!

neel-a-mehta:
This looks good. Small follow-up is #1067

PR #1039, opened 2021-05-06
Make fileupload question in its own block

hmhsu:
### Description
Ensure admin can only add fileupload question in its own block.

### Checklist


universal-application-tool-0.0.1/app/views/components/QuestionBank.java
```diff
@@ -160,14 +159,13 @@ private ContainerTag renderQuestionDefinition(QuestionDefinition definition) {
    * <p>Questions that are filtered out:
    *
    * <ul>
-   *   <li>If there is at least one question in the current block, all enumerator questions
+   *   <li>If there is at least one question in the current block, all single-block questions
    *   <li>If this is a repeated block only show the appropriate repeated questions
    *   <li>Questions already in the program
```
kendrickt, line 164: (blocking) Is this missing a case?

`If there is single block question, all questions are filtered.`

universal-application-tool-0.0.1/app/views/components/QuestionBank.java
```diff
@@ -176,6 +174,11 @@ private ContainerTag renderQuestionDefinition(QuestionDefinition definition) {
     return questions.stream().filter(filter).collect(ImmutableList.toImmutableList());
   }
 
+  /** If a block already contains a single-block question, no more questions can be added. */
+  private boolean noMoreQuestionsAllowed() {
```
kendrickt, line 178: (blocking) rename `noMoreQuestionsAllowed` to `containsSingleBlockQuestion` to keep re-using the new "single block question" term.

browser-test/src/ti_group.test.ts
```diff
@@ -1,19 +1,19 @@
-import { startSession, loginAsAdmin, endSession, AdminTIGroups } from './support'
```
kendrickt, line 1: Is this a carriage return diff?

kendrickt (changes requested):
(blocking) I think there are a couple files that are just whitespace changes. Not sure what's going on but can you not include them in this PR? If we need to push those changes it should be in a separate PR.

(blocking maybe?) do we need tests for this change?

browser-test/src/ti_group.test.ts
hmhsu: Yes, they are formatting changes only.

universal-application-tool-0.0.1/app/views/components/QuestionBank.java
hmhsu: Done.

universal-application-tool-0.0.1/app/views/components/QuestionBank.java
hmhsu: Done.

hmhsu:
> I think there are a couple files that are just whitespace changes. Not sure what's going on but can you not include them in this PR? If we need to push those changes it should be in a separate PR.
Removed

> Do we need tests for this change?
I think the existing test for enumerator question/block is sufficient as they use exactly the same logic.

kendrickt (approved):
lgtm

hmhsu:
Linking - this works toward #404.

Issue #846, opened 2021-04-21
Support deep-linking directly into a specific program application


gcapiel:
Users will mostly enter into CiviForm when the apply for a specific program from affordableseattle.com.  We need to provide the program admin the URL that they should link to for that specific program. CiviForm should support guest entry into a specific application and language selection before they start the specific application.

neel-a-mehta:
Connected to #317 (except this one is on our side, while the other one is for other websites).

neel-a-mehta:
The main type of deep-linking we have to support is a link that sends the user to a program's "info" or "detail" page, which is the screen that shows up right before the applicant actually starts filling things out. Something like the below.

The URL structure would be something like `civiform.seattle.gov/programs/seattle-parks-and-recreation-scholarship`. The last bit of the URL could be a slugified version of the program name. If it's easier, we could use the program ID, so like `civiform.seattle/gov/programs/32`, but that's less readable.

![Mobile - application information](https://user-images.githubusercontent.com/77128518/117357030-719c9200-ae82-11eb-8880-3cb2fa5ff6af.png)


cc @ndmckinley

elisekalstad:
Is it possible to have the deep link go directly into the specific program? The deep link currently takes users to the "All Programs" page of CiviForm and they have to hunt around for the right program in order to apply. The ideal state would have users click on the deep link >> login >> select language >> directs right into the specific program they originally intended.
[Deep links not program specific.pdf](https://github.com/seattle-uat/civiform/files/7087265/Deep.links.not.program.specific.pdf)

PR #1041, opened 2021-05-06
Beginning of deep linking - creates deep links, but not landing pages.

ndmckinley:
### Description
Create a new deeplinking controller whose sole job is to redirect to the correct place.  You do need to be logged in to hit the deep link, but if you aren't, you're redirected to the login page.  

Two follow-ups:
- We need to get the login-page redirects to work the way they're supposed to - at the moment they're broken and you're always redirected to the home page.
- We need to build the details/landing page.

### Checklist

### Issue(s)
Towards #846.

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
@@ -9,6 +9,7 @@ import io.ebean.Ebean;
 import io.ebean.EbeanServer;
 import io.ebean.TxScope;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
 import javax.inject.Provider;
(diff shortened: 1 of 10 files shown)


universal-application-tool-0.0.1/app/repository/ProgramRepository.java
```diff
@@ -110,4 +112,17 @@ public Program createOrUpdateDraft(Program existingProgram) {
       }
     }
   }
+
+  public CompletableFuture<Program> getForSlug(String slug) {
```
kendrickt, line 116: (blocking) this should be stored in a column in the table so we can do an indexed search on this.

kendrickt (approved):
`slug` should be in the database for an indexed search. 

The annoying part would be populating the slug column for old programs. Maybe that's a separate follow-up issue?

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
ndmckinley: yeah, but it will really be a hassle, and the follow-up issue is going to sit in the icebox forever... let me see if I can figure out how to do this.

ndmckinley: Okay, looks like I can make that work...  I'll watch the prober when this goes in.

PR #1047, opened 2021-05-07
Deeplinking: only return active programs.

ndmckinley:
### Description
This prevents people from guessing draft program slugs or getting errors when several versions of a program exist.

The test is a little weird but it's meant to simulate the migration from v19 to v20.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
@@ -8,6 +8,7 @@ import com.google.common.collect.ImmutableList;
 import io.ebean.Ebean;
 import io.ebean.EbeanServer;
 import io.ebean.TxScope;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/repository/ProgramRepository.java
```diff
@@ -121,8 +122,14 @@ public Program createOrUpdateDraft(Program existingProgram) {
             program.getSlug();
             program.save();
           }
+          ImmutableList<Program> activePrograms =
+              versionRepository.get().getActiveVersion().getPrograms();
+          List<Program> programsMatchingSlug =
+              ebeanServer.find(Program.class).where().eq("slug", slug).findList();
```
kendrickt, line 128: (question): in the Program model it says there's a many to many relationship between versions. Doesn't each program just get one version?

I'm wondering if it'd be more effective to use the joined table instead of streaming through all active programs.

kendrickt (approved):

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
ndmckinley: Each program has many versions, yep - if a program is unchanged, the same program id will be in two different versions.  There is only one active and one draft version at a time, though.

ndmckinley: It might be more efficient but it's also harder to write in ebean's query DDL - and we're talking about a database query returning a few dozen rows.  At this scale we can just do the easy thing.

Issue #1051, opened 2021-05-07
Ensure that program admins can only access programs they administer


neel-a-mehta:
A follow-on to #772. Here, please ensure that a program admin can *only* access programs that are assigned to them. My guess is that, since CiviForm admins are the ones creating the programs, we'll need to add some kind of configuration field where the CiviForm admin can specify which program admin(s) is/are responsible for each program.

neel-a-mehta:
cc @ndmckinley

ndmckinley:
It is a many-to-many mapping.

PR #1058, opened 2021-05-07
Create a table to map programs to accounts 

carolinedanzi:
### Description
Add program admin data:
* programs can be administered by multiple Accounts
* accounts can administer multiple programs

### Checklist

### Issue(s)
#1051


universal-application-tool-0.0.1/app/models/Account.java
```diff
@@ -64,6 +70,14 @@ public void setManagedByGroup(TrustedIntermediaryGroup group) {
     return Optional.ofNullable(this.managedByGroup);
   }
 
+  public ImmutableList<Program> getAdministeredPrograms() {
+    return ImmutableList.copyOf(this.administeredPrograms);
+  }
+
+  public void setAdministeredPrograms(ImmutableList<Program> administeredPrograms) {
```
ndmckinley, line 77: Can you do "addAdministeredProgram" (maybe in addition?)

universal-application-tool-0.0.1/conf/evolutions/default/21.sql
```diff
@@ -0,0 +1,11 @@
+# --- Create a table that links programs to the program admin accounts that administer them.
+
+# --- !Ups
+create table program_accounts (
+  program_id bigint not null,
+  account_id bigint not null,
+  primary key (program_id, account_id)
```
ndmckinley, line 7: You probably want a foreign key constraint here.

ndmckinley: (plus ebean wants these to be plural)

universal-application-tool-0.0.1/app/models/Program.java
```diff
@@ -54,6 +54,10 @@
   @OneToMany(mappedBy = "program")
   private List<Application> applications;
 
+  @ManyToMany
```
ndmckinley, line 57: I think you want this side of it to be "mappedBy" in the `ManyToMany` rather than having both sides hit the `JoinTable` (as with Versions and Programs).  The choice of ownership side is, I think, arbitrary.

universal-application-tool-0.0.1/app/models/Account.java
carolinedanzi: Done!

universal-application-tool-0.0.1/app/models/Program.java
carolinedanzi: You're right! Thanks for catching that!

universal-application-tool-0.0.1/conf/evolutions/default/21.sql
carolinedanzi: Thanks! Added the constraint

ndmckinley (approved):

carolinedanzi:
We no longer wish to do this - instead, we will have an array of program names on `Account`, since the same program admins should administer the same programs across versions (which are matched on program name)

PR #1071, opened 2021-05-10
Remove unnecessary JsonIgnore

kendrickt:
### Description
We don't need @JsonIgnore on properties that Jackson doesn't automatically detect as getters.

### DIFF
--- a/universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
@@ -18,7 +18,6 @@ public abstract class ProgramQuestionDefinition {
   @JsonProperty("id")
   public abstract long id();
 
-  @JsonIgnore
   abstract Optional<QuestionDefinition> questionDefinition();
 
   @JsonIgnore
--- a/universal-application-tool-0.0.1/app/services/question/QuestionOption.java
+++ b/universal-application-tool-0.0.1/app/services/question/QuestionOption.java
@@ -1,7 +1,6 @@
 package services.question;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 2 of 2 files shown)


hmhsu (commented):
I thought it might be good for documentation purpose though, as it's a bit confusing what are considered serializable fields in Jackson (I think to anyone at first sight). But approved if you think removing is more appropriate.

hmhsu (approved):

universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
```diff
@@ -18,7 +18,6 @@
   @JsonProperty("id")
   public abstract long id();
 
-  @JsonIgnore
   abstract Optional<QuestionDefinition> questionDefinition();
```
hmhsu, line 21: This one might actually be necessary, I think.

hmhsu: I see. TIL.

I know it takes `getXXX` and `isXXX`. and I thought `XXX` itself (with no arguments) is considered a field.

hmhsu: A good way to verify is to see /dev/seed. Would you give it a quick check?

kendrickt: This is what I when I `bin/run-dev` and check `dev/seed`. Nothing but the `id` is stored for each question here, which is expected.
```
  "blockDefinitions" : [ {
    "id" : 1,
    "name" : "Block 1",
    "description" : "name and favorite color",
    "repeaterId" : null,
    "hidePredicate" : null,
    "optionalPredicate" : null,
    "questionDefinitions" : [ {
      "id" : 1
    }, {
      "id" : 2
    } ]
    ```

hmhsu: Nice! Thanks for checking.

PR #1074, opened 2021-05-10
Fix 22.sql down.

kendrickt:
### Description
Get rid of the column types in the column drop statements.

### DIFF
--- a/universal-application-tool-0.0.1/conf/evolutions/default/22.sql
+++ b/universal-application-tool-0.0.1/conf/evolutions/default/22.sql
@@ -16,14 +16,14 @@ alter table programs add localized_description jsonb;
 
 # --- !Downs
 
-alter table questions drop column question_text jsonb;
-alter table questions drop column question_help_text jsonb;
-alter table questions drop column question_options jsonb;
+alter table questions drop column question_text;
+alter table questions drop column question_help_text;
+alter table questions drop column question_options;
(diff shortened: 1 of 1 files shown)


universal-application-tool-0.0.1/conf/evolutions/default/22.sql
```diff
@@ -16,14 +16,14 @@ alter table programs add localized_description jsonb;
...
+alter table questions drop column question_help_text;
+alter table questions drop column question_options;
 alter table questions rename column legacy_question_text to question_text;
 alter table questions rename column legacy_question_help_text to question_help_text;
 alter table questions rename column legacy_question_options to question_options;
 
-alter table programs drop column localized_name jsonb;
-alter table programs drop column localized_description jsonb;
+alter table programs drop column localized_name;
+alter table programs drop column localized_description;
```
hmhsu, line 27:
```suggestion
alter table programs drop column if exists localized_name;
alter table programs drop column if exists localized_description;
```

hmhsu (approved):
(non-blocking) while you are at it, would you add `if exists` to the statements?

universal-application-tool-0.0.1/conf/evolutions/default/22.sql
kendrickt: fixed

PR #1076, opened 2021-05-10
Admin UI refactor

AWBaumann:
### Description
Added an `HtmlBundle` class that allows us to insert content into different portions of the page instead of the flat J2HTML order. It also has utility functions for adding ToastMessages to the header, setting page titles, and adding style and script tags. This PR currently only uses it in the admin pages, but a follow-up will also add it to the Applicant pages.

### TODOS
- Use `HtmlBundle` on applicant pages.
- Cleanup titles and add translations for all page titles/headers.
- Fix the BaseHtmlLayout/BaseHtmlView inheritance nastiness.

### Checklist


universal-application-tool-0.0.1/app/views/HtmlBundle.java
```diff
@@ -0,0 +1,141 @@
...
+  ArrayList<Tag> footerContent = new ArrayList<Tag>();
+  ArrayList<String> footerStyles = new ArrayList<String>();
+
+  ArrayList<Tag> headerContent = new ArrayList<Tag>();
+  ArrayList<String> headerStyles = new ArrayList<String>();
+
+  ArrayList<Tag> mainContent = new ArrayList<Tag>();
+  ArrayList<String> mainStyles = new ArrayList<String>();
+
+  public HtmlBundle addBodyStyles(String... styles) {
```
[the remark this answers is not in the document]
AWBaumann: Done.

universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
```diff
@@ -27,85 +16,38 @@ public AdminLayout(ViewUtils viewUtils) {
...
-            .with(mainContent)
-            .with(viewUtils.makeLocalJsTag("main"))
-            .withClasses(
-                BaseStyles.BODY_GRADIENT_STYLE,
-                Styles.BOX_BORDER,
-                Styles.H_SCREEN,
-                Styles.W_SCREEN,
-                Styles.OVERFLOW_HIDDEN,
-                Styles.FLEX));
+  private final String BODY_STYLES =
```
natsid, line 19: suggestion (non-blocking): Consider moving these to `AdminStyles.java` and removing the `_STYLES` suffix from them. That way you can reference them like `AdminStyles.BODY` (which will match the pattern I have going on in #1077). It also has the benefit of putting the definitions of these constants elsewhere, so that this class can focus on the other stuff.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -30,17 +29,20 @@ public ProgramApplicationListView(AdminLayout layout) {
   }
 
   public Content render(long programId, ImmutableList<Application> applications) {
+    // TODO: [i18n].
```
natsid, line 32: (non-blocking): I'm not sure we need to internationalize admin-only strings, if only because we don't actually have a way for admins to choose a different language at this point. @carolinedanzi can you confirm?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -30,17 +29,20 @@ public ProgramApplicationListView(AdminLayout layout) {
...
             .with(
-                h1("All Applications").withClasses(Styles.MY_4),
+                h1(title).withClasses(Styles.MY_4),
                 each(
                     applications,
                     application -> this.renderApplicationListItem(programId, application)),
                 renderDownloadButton(programId));
 
-    return layout.render(head(layout.tailwindStyles()), body(contentDiv));
+    HtmlBundle htmlBundle = getHtmlBundle().setTitle(title).addMainContent(contentDiv);
```
natsid, line 44: praise: Nice!  I like the way this bundle/builder works...

universal-application-tool-0.0.1/app/views/admin/programs/ProgramNewOneView.java
```diff
@@ -21,37 +21,25 @@ public ProgramNewOneView(AdminLayout layout) {
...
-                        programForm.getAdminName(),
-                        programForm.getAdminDescription(),
-                        programForm.getLocalizedDisplayName(),
-                        programForm.getLocalizedDisplayDescription(),
-                        false)
-                    .with(makeCsrfTokenInputTag(request))
-                    .withAction(controllers.admin.routes.AdminProgramController.create().url())));
-    if (message.length() > 0) {
-      bodyContent.with(ToastMessage.error(message).setDismissible(false).getContainerTag());
+    String title = "New program";
```
natsid, line 28: discussion topic that doesn't need to be solved in this PR (non-blocking): We should eventually standardize on either using "Sentence case" or "Title Case" for our page titles.

"Sentence case" (i.e., only capitalizing the first letter of the first word rather than capitalizing the first letter of every word) is what we use on my other team, and I personally think it looks nice. But most important is definitely just consistency across CiviForm.

universal-application-tool-0.0.1/app/views/admin/ti/TrustedIntermediaryGroupListView.java
```diff
@@ -39,24 +39,29 @@ public TrustedIntermediaryGroupListView(AdminLayout layout) {
...
-        body(
-            renderHeader("Create New Trusted Intermediary").withClass(Styles.MT_8),
-            renderAddNewButton(request),
-            renderHeader("Existing Trusted Intermediaries"),
-            renderTiGroupCards(tis, request));
+    String title = "Create New Trusted Intermediary";
+
+    HtmlBundle htmlBundle =
+        getHtmlBundle()
+            .setTitle(title)
```
natsid, line 46: praise: Thanks for setting the title in both `title` and `header`!!

natsid (approved):
Looks good! Handful of comments.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
carolinedanzi: Good question! We don't have a way to do this yet, though it will be similar to the applicant side. I think eventually we can internationalize the admin side but for now the assumption is they use the default of English

PR #1082, opened 2021-05-11
delete question path.

### LINKED ISSUE #673: QuestionDefinition Path should not be a constructor argument

kendrickt:
### Description
Delete `path` from `question`.

This should not affect storage. The column can be dropped in a follow-up PR.

### Issue(s)
Fixes #673

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
@@ -127,7 +127,7 @@ public class QuestionController extends CiviFormController {
 
     QuestionDefinition questionDefinition;
     try {
-      questionDefinition = getBuilderWithQuestionPath(Optional.empty(), questionForm).build();
+      questionDefinition = getBuilder(Optional.empty(), questionForm).build();
     } catch (UnsupportedQuestionTypeException e) {
       // Valid question type that is not yet fully supported.
       return badRequest(e.getMessage());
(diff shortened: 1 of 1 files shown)


universal-application-tool-0.0.1/app/controllers/admin/QuestionController.java
```diff
@@ -239,30 +236,14 @@ private Result withMessage(Result result, String message) {
   }
 
   private QuestionDefinitionBuilder getBuilderWithQuestionPath(
```
hmhsu, line 238: Suggestion (blocking): maybe remove `Path` from function name? like `getUpdatedQuestionDefinition`?

kendrickt: nice catch. done!

kendrickt:
> I guess I missed the PR that migrated the use, but this LGTM.

@hmhsu the "contextualized path" was added in #783, and there was a follow up clean up in #920.

PR #1086, opened 2021-05-11
Remove list all programs

hmhsu:
### Description
With versioning, list all programs are not useful. Getting all programs regardless versions is at best misleading, if not causing any harm. The method is only used in test so it is safe to remove.

### Checklist

### Issue(s)


kendrickt (approved):
nice

PR #1110, opened 2021-05-12
Change the way we delete existing enumerator entities so it looks the same as new deletes

carolinedanzi:
### Description
1. Changes the "mark for deletion" checkbox in `EnumeratorQuestionRenderer` to a button with an ID of the entity to delete's index
2. When the existing entity delete button is clicked, we hide the entire field and add a hidden input field that will handle adding the `deleteEntity[]: <index>` to the HTTP form body
3. Consolidates logic for new and existing enumerator fields, since now the only difference is adding an ID to the delete button and reference class

### Checklist


kendrickt (approved):
LGTM

kendrickt:
Problem: Applicant cannot un-delete an entity. 
Solution: add a warning to verify with applicant that they want to delete it. This can include a message about needing to click `save and continue` to actually persist these changes.

carolinedanzi:
> Problem: Applicant cannot un-delete an entity.
> Solution: add a warning to verify with applicant that they want to delete it. This can include a message about needing to click `save and continue` to actually persist these changes.

Filed #1117 for adding the delete confirmation dialog

PR #1120, opened 2021-05-13
View refactor.

AWBaumann:
### Description
Update Views/Layouts in the app.

### Introducing `HtmlBundle`!
- Allows us to set page title
- Allows us to add scripts and CSS
- Allows us to add tags to each section (header, main, footer)
- Returns a well organized page with `getContent()`
```
<html>
      <head>
            <title>Whee!</title>
      </head>
      <body>
            <header></header>
            <main></main>
            <footer></footer>
      </body>
</html>
```


universal-application-tool-0.0.1/app/views/style/AdminStyles.java
```diff
@@ -51,4 +51,28 @@
...
+
+  public static final String BODY =
+      StyleUtils.joinStyles(
+          BODY_GRADIENT_STYLE,
+          Styles.BOX_BORDER,
+          Styles.H_SCREEN,
+          Styles.W_SCREEN,
+          Styles.OVERFLOW_HIDDEN,
+          Styles.FLEX);
+
```
natsid, line 63: suggestion (non-blocking): Can you add a brief comment on CENTERED and FULL for what kind of elements they should be used on and roughly what they do visually? Maybe rename them `MAIN_CENTERED` and `MAIN_FULL` if they should only be used on a `main` tag.

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
```diff
@@ -36,38 +35,47 @@ public ApplicantLayout(ViewUtils viewUtils, ProfileUtils profileUtils) {
...
-        body()
-            .with(renderNavBar(profile, messages))
-            .with(mainDomContents)
-            .with(viewUtils.makeLocalJsTag("main"))
-            .withClasses(ApplicantStyles.BODY_BG_COLOR));
+
+  @Override
+  public HtmlBundle getBundle(HtmlBundle bundle) {
+    return super.getBundle(bundle);
+  }
```
natsid, line 46: (blocking) I don't think we need to override these methods since they're identical to those of `BaseHtmlLayout`

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
```diff
@@ -36,38 +35,47 @@ public ApplicantLayout(ViewUtils viewUtils, ProfileUtils profileUtils) {
...
+  public Content render(HtmlBundle bundle) {
+    bundle.addBodyStyles(ApplicantStyles.BODY_BG_COLOR);
+    String currentTitle = bundle.getTitle();
+    if (currentTitle != null && !currentTitle.isEmpty()) {
+      bundle.setTitle(currentTitle + " - CiviForm");
+    }
+    return super.render(bundle);
+  }
+
+  public Content renderWithNav(Http.Request request, Messages messages, HtmlBundle bundle) {
```
natsid, line 58: question: When do we ever want to render this layout _without_ the nav bar?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -52,7 +52,8 @@ public Content render(
                             request,
                             profile)));
 
-    return layout.render(layout.headContent(), body(contentDiv));
+    HtmlBundle htmlBundle = layout.getBundle().setTitle("All Programs").addMainContent(contentDiv);
```
natsid, line 55: nit: Just cuz you did it everywhere else... :) "All programs"

universal-application-tool-0.0.1/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -13,6 +15,25 @@
  */
 public class ProgramFormBuilder extends BaseHtmlView {
 
+  public static ContainerTag buildProgramForm(ProgramForm program, boolean editExistingProgram) {
```
natsid, line 18: (non-blocking): I'm not super familiar with this class. If it's not too much trouble, can you add a short docstring to each of these `buildProgramForm` methods explaining when you'd want to use each of them? It's not clear to me why we need so many different public method signatures at first glance.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -75,30 +75,34 @@ public Content render(
...
-                .with(
-                    blockEditPanel(
-                        program,
-                        blockId,
-                        blockForm,
-                        blockQuestions,
-                        blockDefinition.isEnumerator(),
-                        csrfTag))
-                .with(questionBankPanel(questions, program, blockDefinition, csrfTag)));
+    String title = "Block edit view";
```
natsid, line 78: suggestion: Rename this title to "Edit block".

universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -30,17 +30,19 @@ public ProgramApplicationListView(AdminLayout layout) {
   }
 
   public Content render(long programId, ImmutableList<Application> applications) {
+    String title = "All Applications";
```
natsid, line 33: nit: "All applications" to match the rest of the titles

universal-application-tool-0.0.1/app/views/admin/programs/ProgramAdministratorProgramListView.java
```diff
@@ -32,11 +31,12 @@ public ProgramAdministratorProgramListView(AdminLayout layout) {
   }
 
   public Content render(ActiveAndDraftPrograms programs, List<String> authorizedPrograms) {
+    String title = "Your Programs";
```
natsid, line 34:
```suggestion
    String title = "Your programs";
```

universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
```diff
@@ -26,6 +24,35 @@ public AdminLayout(ViewUtils viewUtils) {
     super(viewUtils);
   }
 
+  @Override
+  public HtmlBundle getBundle() {
+    return getBundle(new HtmlBundle());
+  }
```
natsid, line 30: Same question as in ApplicantLayout: Do we need to override this?

universal-application-tool-0.0.1/app/views/HtmlBundle.java
```diff
@@ -0,0 +1,180 @@
...
+
+    if (footerStyles.size() > 0) {
+      footerTag.withClasses(footerStyles.toArray(new String[0]));
+    }
+
+    return footerTag;
+  }
+
+  /**
+   * The page head contains: - page title - page metadata - CSS styles - javascript that needs to
```
natsid, line 127: (non-blocking): I think you can format these as actual lists like here, e.g., https://github.com/seattle-uat/civiform/blob/fa6d6e42e3b7e9466ac173bb0bc2bdd86de362e3/universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java#L58

universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
```diff
@@ -1,25 +1,24 @@
...
 import play.twirl.api.Content;
 import views.components.ToastMessage;
 
 /**
  * Base class for all layout classes.
  *
- * <p>A layout class should describe the DOM contents of the head, header, nav, and footer. It
- * should have a `render` method that takes the DOM contents for the main tag.
+ * <p>A layout class should describe the DOM contents of the head, header, nav, and footer. It acts
+ * on an HtmlBundle and returns a rendered page.
```
natsid, line 17: "and returns a rendered page" sounds a little confusing to me since the class itself doesn't return something, but its methods do. Maybe call out which method does/should return the rendered page?

universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
```diff
@@ -31,49 +30,69 @@ public BaseHtmlLayout(ViewUtils viewUtils) {
...
-    return new HtmlResponseContent(contents.toArray(new DomContent[0]));
+  /** Creates a new {@link HtmlBundle} with default css, scripts, and toast messages. */
+  public HtmlBundle getBundle() {
+    return getBundle(new HtmlBundle());
   }
 
   /**
-   * Returns a script tag that loads Tailwindcss styles and configurations common to all pages in
-   * the CiviForm.
+   * The reason for two getBundle methods here is that order occasionally matters, but this method
```
natsid, line 39: I feel like this docstring should also call out that the subclasses of BaseHtmlLayout should probably almost always, if not just always call this super class version since it adds the tailwind styles and the google analytics, etc.

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
AWBaumann: Never that I'm aware of?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramFormBuilder.java
AWBaumann: I'm not super familiar with this either. It looks like you can build the form using data either from a `ProgramForm` or a `ProgramDefinition`. The former is passed when a user updates a program and stays on the page (and may include update errors). The latter is used when reading data from the database.

It's mainly just me relocating the mapping from `ProgramEditView` and `ProgramNewOneView` into `ProgramFormBuilder`

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
natsid: In that case, I was wondering if we should remove this as a public method, and instead just use `render` and always add that `addHeaderContent` line, if that makes sense.

AWBaumann: It requires `request` and `messages` though, so it needs more context than the typical render method and I didn't want to overload render with additional data so the method is kind of self-documenting

I think I was initially going to have an `ApplicantView`class that extended `BaseHtmlView` and had this functionality, but it felt like overkill for a single method. It might be worth looking into `BaseHtmlView` and extending that as the class is pretty worthless/unused right now.

natsid: Yeah... `BaseHtmlView` is pretty unused as a super class right now.

Hmm, OK. But I wonder if, in that case, we should somehow discourage people to call the `render` method on ApplicantLayout...  We always want the caller to use `renderWithNav`

universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
natsid: What's the verdict here? We need to override it?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramFormBuilder.java
natsid: But does the the third version below these two need to be public? Or is it only ever called from within this class?

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
AWBaumann: I could override `render` to throw an error?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramFormBuilder.java
AWBaumann: The third method should be private now.

universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
AWBaumann: This should be removed with my latest changes.

universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
natsid: Might not be a bad temporary solution. Your call.

natsid (approved):
Sweet! Sorry - I must not have been looking at your most recent changes for some reasons.

Issue #193, opened 2021-02-19
Program admins get emails when one of their program forms gets filled out


neel-a-mehta:
Emails contain links to pages where they can download PDFs/CSVs (#191 or eventually #192)

neel-a-mehta:
OK this clearly works in staging. I have no way of knowing if it works in prod, but we'll have to wait until the program admins start getting applications (hopefully that's soon!). At that point we can ask them if they're getting emails and, if necessary, revisit this.

PR #1130, opened 2021-05-14
Notify program admins

hmhsu:
### Description
- Notify real program admins instead of a static mailing list
- Notify global admin if no program admins are configured
- Use @seattle.gov email domain in prod

### Checklist
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)
#193

PR #1132, opened 2021-05-14
Return correct email addresses for the emails we need to send.

ndmckinley:
### Description
Send emails to program admins if there are any, and to global admins if there aren't any program admins.

### Checklist

### Issue(s)
#193


ndmckinley:
Hmm... this is going to start sending some of us emails from staging every time we run the prober.  Is that okay?  It seems like we might want this to run only in prod.

hmhsu (approved):

hmhsu:
> Hmm... this is going to start sending some of us emails from staging every time we run the prober. Is that okay? It seems like we might want this to run only in prod.

That was part of the reasons I intended to stick with the mailing list for now. I happened to open #1130, but feel free to check in yours.

ndmckinley:
Cool, nope, all yours!

PR #1138, opened 2021-05-14
Enable ability to breakglass access AWS database

hmhsu:
### Description
- The database is in private subnets so we need to create an EC2 instance in a public subnet and relay our access to it
- We create a new security group allowing only our own IP and attach it to the DB security group ingress
- Create a new key pair for the EC2 instance and store private key locally
- Launch the instance in the public subnet with the new security group, wait for it, and connect to it
- Install `psql` client in the new instance and use it to access the database
- Clean up everything after the access
- The script connects to staging right now, not prod

### Checklist

### DIFF
--- a/bin/breakglass-db-access
+++ b/bin/breakglass-db-access
@@ -1,17 +1,21 @@
 #! /bin/bash
 
+set -e
+
 export AWS_DEFAULT_REGION=us-west-2
 TIMESTAMP=$(date +%s)
+ENVIRONMENT=staging
+STAGING_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:staging-databasepassword-rh1wfF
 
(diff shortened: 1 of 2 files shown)


bin/breakglass-db-access
```diff
@@ -0,0 +1,118 @@
+#! /bin/bash
+
```
ndmckinley, line 2: Consider `set -e` so that errors are easy to detect and stop the script from running.  Also perhaps `set -x` so that we can find out what's running.

ndmckinley (approved):
This is amazing, wow.

bin/breakglass-db-access
hmhsu: Added `set -e`. I printed most relevant information (except DB password) so `set -x` may not be necessary. It can always be added to debug.

Issue #624, opened 2021-03-29
ApplicantData.put methods are too recursive


kendrickt:
**Is your feature request related to a problem? Please describe.**
Currently there are no rules to what data can be added to `ApplicantData`. For example, an empty applicant data can have the following data added:

```applicant.children[10].jobs[5].employer.text = "Google"```

There are rules in place that can make adding data to `ApplicantData` more robust:

1. Paths to answers look like `{parent_path}.{question_name}.{scalar}`. The `question_name` JSON object shouldn't be put at `{parent_path}` if the parent path doesn't already exist.

2. Paths to repeated entities look like `{parent_path}.{repeated_entity_name}[{index}]`. The `repeated_entity_name` JSON array shouldn't be put at `{parent_path}` if the parent path doesn't already exist.

3. Paths to repeated entities look like `{parent_path}.{repeated_entity_name}[{index}]`. For `N > 0`, the JSON object at index `N` shouldn't be added to the JSON array at `repeated_entity_name` if the JSON object at index `N-1` doesn't already exist.

Rules 2 and 3 depend on being able to specify the order of updates in a single `ApplicantService.stageAndUpdateIfValid`. It currently uses a `Set` and we don't want to be dependent on order.

**Describe the solution you'd like**
In `ApplicantData`, the `putParentIfMissing` and `putParentArray` methods recursively call `putParentIfMissing` (this issue is linked in TODOs). Get rid of the recursion, and then fix all the tests that break because tests are assuming they can add applicant data anywhere.

**ALTERNATIVE SOLUTION**: allow `ApplicantData.put` methods to be flexible, and enforce the rules in `ApplicantService`. 

**Additional context**
#625 PR adds recursive for adding parent arrays.

**Done when**
Rules are enforced, tests are added for the rules, and tests pass.

If this work isn't going to be done, please remove the TODO comments marked with #624 before closing this issue.

kendrickt:
We may not be able to enforce update ordering in the ApplicantService. If not, then we may need to enable adding `something[3]` before `something[2]`.

carolinedanzi:
Originally, Natalie and I decided that `ApplicantData` should be "dumb", in that it blindly adds whatever we tell it to, and it's up to callers to enforce logic around valid updates

bion:
> Originally, Natalie and I decided that `ApplicantData` should be "dumb", in that it blindly adds whatever we tell it to, and it's up to callers to enforce logic around valid updates

[...]

PR #1155, opened 2021-05-15
Mark enumerator as complete if applicant did not answer

### LINKED ISSUE #1141: We should have a way to mark an enumerator as completed, even if it is empty.
An applicant may leave an enumerator empty. This should mean it was completed, but there's no way to indicate that right now.

We should have a way to indicate an intentionally empty enumerator in `ApplicantData`.

**Additional context**

**Done when**
`Block#isCompleteWithoutErrors` returns true for an intentionally empty enumerator block.

carolinedanzi:
### Description
If an applicant chose not to add any enumerator entities, we still showed them the question as if they had not completed it. This fixes this issue by ensuring the path exists in `ApplicantData` even if the applicant has no data to write for an enumerator

### Checklist

### Issue(s)
Fixes #1141

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
@@ -7,6 +7,8 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
+import com.typesafe.config.Config;
+import java.net.URI;
 import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Comparator;
@@ -34,8 +36,7 @@ import services.question.exceptions.UnsupportedScalarTypeException;
 import services.question.types.ScalarType;
 
 public class ApplicantServiceImpl implements ApplicantService {
-  // TODO: use program admin emails instead
-  private static final String PROGRAM_ADMIN_NOTIFICATION_MAILING_LIST =
+  private static final String STAGING_PROGRAM_ADMIN_NOTIFICATION_MAILING_LIST =
       "seattle-civiform-program-admins-notify@google.com";
 
   private final ApplicationRepository applicationRepository;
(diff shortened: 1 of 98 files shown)


hmhsu:
Could we do something like this https://github.com/seattle-uat/civiform/pull/1157? I really think this might be cleaner, but I fear I am missing some important pieces.

kendrickt (approved):
I removed the hack I put into the `ApplicantProgramBlocksController` and fixed the test.

This looks great to me. Two comments:

1. would make sense for `readRepeatedEntities` to return an Optional since that's what the other read methods return?
2. Using `putArrayIfMissing` is fine because we have lots of structure in the order of updates, but this reminds me of https://github.com/seattle-uat/civiform/issues/624. This won't work if the update is something like `applicant.household_members[3].job -> []`, which can't really happen, but I think `ApplicantData` should allow it.

carolinedanzi:
> Could we do something like this #1157? I really think this might be cleaner, but I fear I am missing some important pieces.

Oh neat, that's a nice improvement! I think you're right that something similar would work - I updated to use `putRepeatedEntities` instead with an empty list

carolinedanzi:
> I removed the hack I put into the `ApplicantProgramBlocksController` and fixed the test.
> 
> This looks great to me. Two comments:
> 
> 1. would make sense for `readRepeatedEntities` to return an Optional since that's what the other read methods return?
> 2. Using `putArrayIfMissing` is fine because we have lots of structure in the order of updates, but this reminds me of #624. This won't work if the update is something like `applicant.household_members[3].job -> []`, which can't really happen, but I think `ApplicantData` should allow it.

Thanks Kendrick! For #1, I think it makes more sense for `readRepeatedEntities` to return an empty list as it currently does - the optionals were due to other types not having a good empty/null default value. For lists, an empty list makes more sense

I'm not sure I understand #2, but I updated to use `putRepeatedEntities` with an empty list, is that better?

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -258,6 +258,15 @@ private void stageEnumeratorUpdates(
       UpdateMetadata updateMetadata,
       ImmutableSet<Update> updates)
       throws PathNotInBlockException {
+
+    // The applicant has seen this question, but has not supplied any entities.
+    // We need to still write this path so we can tell they have seen this question.
+    if (updates.isEmpty()) {
```
kendrickt, line 264: (blocking) I'd rather see `if !applicantData.hasPath(enumeratorPath) { put empty array; }

carolinedanzi: Good point! Done

PR #1162, opened 2021-05-15
Add question validation check for $this in question text and help text for repeated questions.

kendrickt:
### Description
Repeated questions are now required to have "$this" in their text and help text, for all translations.

### DIFF
--- a/browser-test/src/end_to_end_enumerators.test.ts
+++ b/browser-test/src/end_to_end_enumerators.test.ts
@@ -12,46 +12,46 @@ describe('End to end enumerator test', () => {
     const adminPrograms = new AdminPrograms(page);
 
     await adminQuestions.addNameQuestion('enumerator-ete-name');
-    await adminQuestions.addEnumeratorQuestion('enumerator-ete-enumerator', 'desc', 'enumerator-ete-question');
-    await adminQuestions.addNameQuestion('enumerator-ete-enumerator-name', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
-    await adminQuestions.addEnumeratorQuestion('enumerator-ete-enumerator-enumerator', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
-    await adminQuestions.addTextQuestion('enumerator-ete-enumerator-enumerator-text', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator-enumerator');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-householdmembers', 'desc', 'Household members', 'list household members');
+    await adminQuestions.addNameQuestion('enumerator-ete-repeated-name', 'desc', 'Name for \$this', 'full name for \$this', 'enumerator-ete-householdmembers');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-repeated-jobs', 'desc', 'Jobs for \$this', '\$this\'s jobs', 'enumerator-ete-householdmembers');
+    await adminQuestions.addNumberQuestion('enumerator-ete-repeated-jobs-income', 'desc', 'Income for \$this.parent\'s job at \$this', 'Monthly income at \$this', 'enumerator-ete-repeated-jobs');
 
     await adminPrograms.addProgram(programName);
     await adminPrograms.editProgramBlock(programName, 'ete enumerator program description');
 
     // All non-repeated questions should be available in the question bank.
     expect(await page.innerText('id=question-bank-questions')).toContain('enumerator-ete-name');
-    expect(await page.innerText('id=question-bank-questions')).toContain('enumerator-ete-enumerator');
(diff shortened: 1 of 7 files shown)


browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -12,46 +12,46 @@ describe('End to end enumerator test', () => {
...
-    await adminQuestions.addNameQuestion('enumerator-ete-enumerator-name', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
-    await adminQuestions.addEnumeratorQuestion('enumerator-ete-enumerator-enumerator', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator');
-    await adminQuestions.addTextQuestion('enumerator-ete-enumerator-enumerator-text', 'desc', 'text', 'helptext', 'enumerator-ete-enumerator-enumerator');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-householdmembers', 'desc', 'Household members', 'list household members');
+    await adminQuestions.addNameQuestion('enumerator-ete-repeated-name', 'desc', 'Name for \$this', 'full name for \$this', 'enumerator-ete-householdmembers');
+    await adminQuestions.addEnumeratorQuestion('enumerator-ete-repeated-jobs', 'desc', 'Jobs for \$this', '\$this\'s jobs', 'enumerator-ete-householdmembers');
+    await adminQuestions.addNumberQuestion('enumerator-ete-repeated-jobs-income', 'desc', 'Income for \$this.parent\'s job at \$this', 'Monthly income at \$this', 'enumerator-ete-repeated-jobs');
```
bion, line 18: Question (non-blocking): why revert to `repeated` here and other places?

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -12,46 +12,46 @@ describe('End to end enumerator test', () => {
     const adminPrograms = new AdminPrograms(page);
 
     await adminQuestions.addNameQuestion('enumerator-ete-name');
```
carolinedanzi, line 14: praise: thank you, this is much more readable and easier to understand!

universal-application-tool-0.0.1/app/services/question/types/QuestionDefinition.java
```diff
@@ -244,4 +249,13 @@ public boolean equalsIgnoreId(Object other) {
...
     return false;
   }
+
+  private boolean questionTextAndHelpTextContainsRepeatedEntityNameFormatString() {
+    boolean textMissingFormatString =
+        questionText.translations().values().stream().anyMatch(text -> !text.contains("$this"));
+    boolean helpTextMissingFormatString =
+        questionHelpText.translations().values().stream()
+            .anyMatch(helpText -> !helpText.contains("$this"));
+    return !textMissingFormatString && !helpTextMissingFormatString;
```
carolinedanzi, line 259: I think I agree with you, it would be more flexible for this to be "or", but we may need to change `anyMatch` to `allMatch`. I'm worried about a case where a translation is missing a reference to $this and we don't catch it

carolinedanzi (approved):
This looks nice! Thanks Kendrick!

universal-application-tool-0.0.1/app/services/question/types/QuestionDefinition.java
```diff
@@ -244,4 +249,13 @@ public boolean equalsIgnoreId(Object other) {
...
+  }
```
bion, line 260: Issue (blocking): this will return an error even if the help text is missing altogether, but the error message says it's ok for the help text to be blank

browser-test/src/end_to_end_enumerators.test.ts
kendrickt: It is because they are repeated questions. Before the questions weren't asking real questions so i just chained the names (e.g. `enumerator-enumerator-text`). Now that the questions are real questions, they have real names. e.g. `repeated-jobs-income` is the repeated question asking about job's income.

universal-application-tool-0.0.1/app/services/question/types/QuestionDefinition.java
kendrickt: The text is missing the format string if any translations do not contain "$this". 

I think that's what is implemented. I'm adding a test to make sure though.

universal-application-tool-0.0.1/app/services/question/types/QuestionDefinition.java
kendrickt: I'm not seeing this when I run it locally. 

Error when help text is added and does not include $this:
image

When I get rid of the help text, the question gets created.

kendrickt: there's also a unit test that checks when the help text is not present that no error is returned.

PR #1174, opened 2021-05-17
Don't throw a runtime exception when a multi-option question has no options

kendrickt:
### Description
Admins were being sent a 500 if they didn't add options to multi option questions.

Now they get a validation toast if a multi-option question has no options, or it has blank ones.


carolinedanzi (approved):
Thanks for fixing this, Kendrick!

PR #1196, opened 2021-05-18
Add scrollbar when program description is too long for program card

### LINKED ISSUE #1180: Handle program description text overflow
Do something more graceful than this:

Screen Shot 2021-05-17 at 4 49 52 PM

Add scrolling ? Though this solution probably won't work well on mobile.

natsid:
@AWBaumann - I saw your comment on using line clamp in the issue. Please see issue linked below for my reply.

### Screenshots

Desktop:
Screen Shot 2021-05-18 at 3 48 09 PM

Mobile (not the greatest experience as a scrollbar... but OK for now):
Screen Shot 2021-05-18 at 4 23 09 PM

Old behavior:
Screen Shot 2021-05-18 at 4 27 31 PM

### Issue(s)
Fixes #1180

### DIFF
--- a/universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
+++ b/universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
@@ -158,7 +158,7 @@ public class ProgramIndexView extends BaseHtmlView {
     ContainerTag applyDiv =
         div(applyButton)
             .withClasses(
-                Styles.W_FULL, Styles.MT_4, Styles.MB_6, Styles.SELF_END, Styles.FLEX_GROW);
+                Styles.W_FULL, Styles.MB_6, Styles.SELF_END, Styles.FLEX_GROW);
     return div()
         .withId(baseId)
         .withClasses(ReferenceClasses.APPLICATION_CARD, ApplicantStyles.PROGRAM_CARD)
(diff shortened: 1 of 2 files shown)


AWBaumann:
I'm not sure this is a good approach. Scrollbars on inner divs are havoc on touch screens.

I really think we should just add line clamps and call it good. I wouldn't worry about making the div expand. I was thinking something more like below. Once we have a program landing page we can show the entire description there more easily.

**Raw CSS would be...**
```
-webkit-line-clamp: 7;
display: -webkit-box;
-webkit-box-orient: vertical;
overflow: hidden;
```

Tailwind has a plugin for it so we may just need 

`line-clamp-7` and
```
// tailwind.config.js
module.exports = {
  // ...
  plugins: [
    // ...
    require('@tailwindcss/line-clamp'),
  ],
}
```

image

bion (approved):

natsid:
@AWBaumann , I understand the concern. We still need to figure out an affordance for how the text will expand beyond the line clamp. I don't think it makes sense just to clamp it and never be able to see the rest of the description.

I agree that using line clamp will eventually be better. But I'd like to have a design for the full experience before implementing this.

I also think this PR adds some other improvements beyond just the scroll bar, by using `flex`, `gap`, and related classes within the program cards. Makes the layout easier to work with.

PS we can also add custom line clamp classes so that we can use a longer line clamp if desired. It's something we can add to the `extend` block in `tailwind.config.js`.

AWBaumann (approved):

PR #1197, opened 2021-05-18
Help text should be optional when translating

### LINKED ISSUE #1186: Optional help text breaks translations
**Describe the bug**
Question help text is optional - an admin does not need to provide it. However, if they do not provide help text, the question table incorrectly states that English is not a supported language for that question. Additionally, trying to reach the question manage translations page throws this error: `CompletionException: java.lang.RuntimeException: services.TranslationNotFoundException: No translation was found for locale en_US`

**To Reproduce**
1. Create a new question, but do not provide help text
2. Note that the list of supported locales for that question is empty in the `/admin/questions` view
3. Click "Manage Translations" for that question and note the exception that is thrown

**Additional context**
This should be an easy fix:
1. `getSupportedLocales` in `QuestionDefinition` should ignore help text if it is empty
2. Do not add an input for help text in `QuestionTranslationsView` if it is empty

carolinedanzi:
### Description
1. Only shows the help text localization input if there is help text to localize
2. A question is considered to support a locale if
    a. It has help text and both question text and help text support the locale OR
    b. It does not have help text and the question text supports the locale

### Checklist

### Issue(s)
Fixes #1186


natsid (approved):
Excellent !

PR #1199, opened 2021-05-19
Auto logout if account is deleted

hmhsu:
### Description
This mostly only matters in dev and staging where db is frequently wiped out. You will be auto logout if your previous account is deleted.

### Checklist


ndmckinley:
Hmm... I think this is going to be a little messy to do upkeep on / to maintain, since it needs to sit in front of every controller.  What would you say instead to a Filter, which can run in parallel with every request, and checks if the account is valid, and if it isn't, redirects to a logout page.  Should be easier to do upkeep on, plus it can be disabled with tags in the `routes` file.

hmhsu:
> Hmm... I think this is going to be a little messy to do upkeep on / to maintain, since it needs to sit in front of every controller. What would you say instead to a Filter, which can run in parallel with every request, and checks if the account is valid, and if it isn't, redirects to a logout page. Should be easier to do upkeep on, plus it can be disabled with tags in the `routes` file.

Ah nice I didn't know that is possible. Do you have an example? totally fine if not, I can try to find one.

ndmckinley:
Yeah, you can take a look in the `filters/` directory - the `LoggingFilter` is ... well, it's less complicated than this, for sure, but it shows how the interface works.

hmhsu:
Close in favor of #1201

PR #1201, opened 2021-05-19
Auto logout if account is nonexistent

hmhsu:
### Description
- If the account referenced in cookie is nonexistent, auto redirect the user to log out

### Checklist


kendrickt (approved):

ndmckinley (approved):
LGTM!  have you tested it?  It seems like it would work right.

hmhsu:
> LGTM! have you tested it? It seems like it would work right.

Tested it locally. That's how I realized I can't redirect on logout page.

PR #1209, opened 2021-05-19
Make greps for finding missing and weird translations unix compatible.

kendrickt:
Not sure why but unix doesn't like the regex used.

In unix, 
`echo " 1111" | grep -o 1*` doesn't work, but `echo " 1111" | grep -o 11*` does work. :confused: 

So the regex is changed from `[^<]*` to `[^<][a-zA-Z.]*`


hmhsu (approved):
[nit] please add PR description.

PR #1214, opened 2021-05-19
Add a regex for \$this and \$this.parent etc etc to highlight it in preview.ts

### LINKED ISSUE #1198: When admin is creating enumerated question, highlight `$this` in preview pane
When I'm writing an enumerated question, I can write things like `$this` and `$this.parent` in the question text or description. Those strings also show up in the preview pane on the right. Is it possible to highlight `$this` and `$this.parent`, etc, in the preview pane (perhaps by making the text appear a different color or underlining it with a tooltip showing that the text will be replaced)? This would help CiviForm admins know that they are doing it right.

See #1049 for the original issue for this feature.

kendrickt:
### Description
Add regex `/(\$this(\.parent)*)/g` and use it to highlight matching strings in the preview text and help text divs. It does not use `innerHTML` so raw html won't get rendered.

image

### Issue(s)
Fixes #1198

### DIFF
--- a/universal-application-tool-0.0.1/app/assets/javascripts/preview.ts
+++ b/universal-application-tool-0.0.1/app/assets/javascripts/preview.ts
@@ -26,7 +26,7 @@ class PreviewController {
       textInput.addEventListener('input', PreviewController.onTextChanged, false);
       let text = (<HTMLInputElement>textInput).value;
       if (text.length > 0) {
-        PreviewController.setChildrenTextNodes(PreviewController.QUESTION_TEXT_CLASS, text);
+        PreviewController.setTextAndHighlightEnumeratorReferences(PreviewController.QUESTION_TEXT_CLASS, text);
       }
     }
     const helpTextInput =
@@ -35,7 +35,7 @@ class PreviewController {
       helpTextInput.addEventListener('input', PreviewController.onHelpTextChanged, false);
       let helpText = (<HTMLInputElement>helpTextInput).value;
       if (helpText.length > 0) {
-        PreviewController.setChildrenTextNodes(PreviewController.QUESTION_HELP_TEXT_CLASS, helpText);
+        PreviewController.setTextAndHighlightEnumeratorReferences(PreviewController.QUESTION_HELP_TEXT_CLASS, helpText);
       }
     }
     const enumeratorSelector = document.getElementById(PreviewController.QUESTION_ENUMERATOR_INPUT_ID);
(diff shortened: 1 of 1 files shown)


kendrickt:
The bad thing about this is that it allows the admins to type html into the inputs and it'll render it on the right hand side. Is there a cleaner way to do this?

AWBaumann (approved):

neel-a-mehta:
Is it possible to auto-escape the user's inputted text (so `<>` becomes `&lt;&gt;`, etc) before applying the regex to apply highlighting formatting?

natsid:
Just for context for anyone reading this PR: @kendrickt 's comment about allowing the admins to type html into the inputs is no longer relevant or an issue!

universal-application-tool-0.0.1/app/assets/javascripts/preview.ts
```diff
@@ -15,14 +15,16 @@ class PreviewController {
   static readonly DEFAULT_QUESTION_HELP_TEXT = "Sample question help text";
   static readonly DEFAULT_ENTITY_TYPE = "Sample repeated entity type";
 
+  static readonly THIS_REGEX = /(\$this(?:\.parent)*)/g;
```
natsid, line 18: maybe add a brief comment explaining the function of this regex?

kendrickt: added

universal-application-tool-0.0.1/app/assets/javascripts/preview.ts
```diff
@@ -98,6 +100,30 @@ class PreviewController {
...
 
+  /**
+   * Sets the child nodes of the selected div as text or span nodes.
+   * This will highlight text matching PreviewController.THIS_REGEX
+   * in span nodes.
+   *
+   * This will only work when the selected div is only supposed to contain
+   * text and has no other child nodes.
+   */
+  static setChildrenTextNodes(selector: string, text: string) {
```
natsid, line 113: nit: `setChildrenTextNodes` is kinda vague. Maybe rename to `highlightEnumeratorReferences` or something.

natsid (approved):

AWBaumann (approved):

universal-application-tool-0.0.1/app/assets/javascripts/preview.ts
kendrickt: renamed to `setTextAndHighlightEnumeratorReferences`.

PR #1257, opened 2021-05-22
Add ApplicantData#evalPredicate

bion:
### Description
Add `ApplicantData#evalPredicate()`

### Checklist


universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -32,20 +32,17 @@
...
-  private boolean locked = false;
-
-  private Optional<Locale> preferredLocale;
   private final DocumentContext jsonData;
-
   private final Logger logger = LoggerFactory.getLogger(this.getClass());
-
-  public static final Path APPLICANT_PATH = Path.create(APPLICANT);
+  private boolean locked = false;
+  private Optional<Locale> preferredLocale;
```
bion, line 45: the autoformatter made a bunch of these changes

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
```diff
@@ -0,0 +1,22 @@
+package services.applicant;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/** Holds a JsonPath (https://github.com/json-path/JsonPath) path containing a query predicate. */
+public class JsonPathPredicate {
```
carolinedanzi, line 6: question (blocking): any reason not to use `@AutoValue` here? Seems like a data class that would benefit from the auto-generated AutoValue features

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
```diff
@@ -0,0 +1,22 @@
+package services.applicant;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/** Holds a JsonPath (https://github.com/json-path/JsonPath) path containing a query predicate. */
```
carolinedanzi, line 5: design question (non-blocking): Will this hold the full JsonPath  predicate, including the question paths?

carolinedanzi, line 5: praise: thank you for linking to the documentation!

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
```diff
@@ -0,0 +1,22 @@
+package services.applicant;
```
carolinedanzi, line 1: suggestion (non-blocking): Since we have so many predicate-related classes planned, I wonder if it would make sense to create a `services.applicant.predicate` package. What do you think?

carolinedanzi (approved):
Nice work, this looks really good! Glad to see JsonPath behaves as expected and will fit our needs

universal-application-tool-0.0.1/test/services/applicant/ApplicantDataTest.java
```diff
@@ -506,4 +506,68 @@ public void clearArray_ignoresScalar() {
     data.maybeClearArray(Path.create("applicant.cars"));
     assertThat(data.asJsonString()).isEqualTo(expected);
   }
+
```
carolinedanzi, line 509: praise: nice tests! 

question (blocking): Can we add a test for a multi-part expression with `&&` or `||`? I want to understand if it is possible

Maybe something like:
```
{
  "applicant": {
    "favorite_color": "blue"
    "address": {
      "zip": "12345"
    }
  }
}

$.applicant[?(@.favorite_color != 'green' && @.address.zip in ['12345', '45678'])]
```

carolinedanzi: Or something like this one from the examples, where we compare against another value in the json structure: `$..book[?(@.price <= $['expensive'])]`

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
bion: Done

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
bion: Done

universal-application-tool-0.0.1/app/services/applicant/JsonPathPredicate.java
bion: I don't think so, at the moment I think it will just hold this string value. I added it to avoid primitive obsession and to ensure JsonPathPredicate as a first class domain concept in the system.

universal-application-tool-0.0.1/test/services/applicant/ApplicantDataTest.java
bion: We plan to implement `AND` and `OR` in java rather than JsonPath, so skipped that. Added two test assertions for referencing other values in applicant data for the right-hand comparison value.

PR #1265, opened 2021-05-24
Bugfix/1261 empty helptext

### LINKED ISSUE #1261: Editing a question with no help text makes it have empty string as help text
**Describe the bug**
The help text is optional. If it is not provided, it is "empty". However, after editing the question, even if leaving the help text field empty, the help text is stored as the empty string.

This might happen on question load, or after edit.

This causes problems if we have code relying on `LocalizedStrings.isEmpty()`. It messes up the check for `"$this"` for repeated questions because we only require `"$this"` when the help text is not empty.

**To Reproduce**
1. Create a question with no help text
2. Edit the question, keep help text empty.
3. In `/dev/seed` the help text now has a blank translation.

**Expected behavior**
If no help text has been configured by the admin, the help text should be empty, and not contain an empty string.

**Screenshots**
question upon creation. the help text is blank.
image

question after edit. the help text has a default translation of the empty string.
image

kendrickt:
### Description
Question help text deleted in the question edit form now deletes all translations of it. This is the same behavior that happens when admins delete an option from a multi option question.

Two browser tests were added:
1. `admin_program_translation.test.ts` tests that translations are deleted when an admin updates the question help text to be blank.
2. `question_lifecycle.test.ts` tests that updating a repeated question does not require "$this" in the help text because it knows it is an empty set of localized strings. Before these changes, it behaved as if there was a help text, and then required "$this", even though the help string was empty.
    * Upon editing the `question_lifecycle.test.ts` it was clear that this test didn't create new versions of all the questions, so the test was fixed. 

This is what used to happen. Help text was set to empty string for default locale, but was kept for other locales. 
after deleting help text

Now the translations are empty, instead of containing the empty string and old translations for other locales.
image

### Checklist

### Issue(s)
Fixes #1261

### DIFF
--- a/browser-test/src/admin_program_translation.test.ts
+++ b/browser-test/src/admin_program_translation.test.ts
@@ -164,6 +164,37 @@ describe('Admin can manage translations', () => {
     await endSession(browser);
   });
 
+  it('deleting help text in question edit view deletes all help text translations', async () => {
+    const { browser, page } = await startSession();
+
+    await loginAsAdmin(page);
+    const adminQuestions = new AdminQuestions(page);
+
+    // Add a new question with help text
+    const questionName = 'translate-help-text-deletion';
+    await adminQuestions.addNumberQuestion(questionName);
+
+    // Add a translation for a non-English language.
+    await adminQuestions.goToQuestionTranslationPage(questionName);
+    const adminTranslations = new AdminTranslations(page);
+    await adminTranslations.selectLanguage('Spanish');
+    await adminTranslations.editQuestionTranslations('something different', 'help text different');
+
+    // Edit the question and delete the help text.
+    await adminQuestions.changeQuestionHelpText(questionName, "");
+
+    // Edit the question and add help text back
+    await adminQuestions.changeQuestionHelpText(questionName, "a new help text");
+
+    // View the question translations and check that the Spanish translations for question help text are gone.
+    await adminQuestions.goToQuestionTranslationPage(questionName);
+    await adminTranslations.selectLanguage('Spanish');
+    expect(await page.getAttribute('#localize-question-text', 'value')).toContain('something different');
+    expect(await page.getAttribute('#localize-question-help-text', 'value')).toEqual('');
+
+    await endSession(browser);
+  });
+
   it('Applicant sees toast message warning translation is not complete', async () => {
     const { browser, page } = await startSession();
 
--- a/browser-test/src/admin_program_translation.test.ts
+++ b/browser-test/src/admin_program_translation.test.ts
@@ -181,10 +181,11 @@ describe('Admin can manage translations', () => {
     await adminTranslations.editQuestionTranslations('something different', 'help text different');
 
     // Edit the question and delete the help text.
-    await adminQuestions.changeQuestionHelpText(questionName, "");
+    await adminQuestions.changeQuestionHelpText(questionName, '');
 
     // Edit the question and add help text back
-    await adminQuestions.changeQuestionHelpText(questionName, "a new help text");
+    await adminQuestions.changeQuestionHelpText(questionName, 'a new help text');
+
 
     // View the question translations and check that the Spanish translations for question help text are gone.
     await adminQuestions.goToQuestionTranslationPage(questionName);
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -14,8 +14,9 @@ describe('normal question lifecycle', () => {
     const repeatedQuestion = 'qlc-repeated-number';
     await adminQuestions.addNumberQuestion(
       repeatedQuestion, 'description', '$this\'s favorite number', '', 'qlc-enumerator');
-    const allQuestions = questions.concat(singleBlockQuestions);
 
+    // Combine all the questions that were made so we can update them all together.
+    const allQuestions = questions.concat(singleBlockQuestions);
     // Add to the front of the list because creating a new version of the enumerator question will
     // automatically create a new version of the repeated question.
     allQuestions.unshift(repeatedQuestion);
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -11,22 +11,30 @@ describe('normal question lifecycle', () => {
 
     const questions = await adminQuestions.addAllNonSingleBlockQuestionTypes('qlc-');
     const singleBlockQuestions = await adminQuestions.addAllSingleBlockQuestionTypes('qlc-');
+    const repeatedQuestion = 'qlc-repeated-number';
+    await adminQuestions.addNumberQuestion(
+      repeatedQuestion, 'description', '$this\'s favorite number', '', 'qlc-enumerator');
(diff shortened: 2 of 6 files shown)


kendrickt:
This is not ready to merge. There's a strange interaction with deleting the "optional" question help text.

browser-test/src/question_lifecycle.test.ts
```diff
@@ -11,30 +11,41 @@ describe('normal question lifecycle', () => {
 
     const questions = await adminQuestions.addAllNonSingleBlockQuestionTypes('qlc-');
     const singleBlockQuestions = await adminQuestions.addAllSingleBlockQuestionTypes('qlc-');
+    const repeatedQuestion = 'qlc-repeated-number';
+    await adminQuestions.addNumberQuestion(
+      repeatedQuestion, 'description', '$this\'s favorite number', '', 'qlc-enumerator');
+    const allQuestions = questions.concat(singleBlockQuestions);
```
carolinedanzi, line 17: nit (non-blocking): Could we move this line right below the single block question creation? It confused me that we create the repeated question but don't add it until later

browser-test/src/question_lifecycle.test.ts
```diff
@@ -11,30 +11,41 @@ describe('normal question lifecycle', () => {
 
     const questions = await adminQuestions.addAllNonSingleBlockQuestionTypes('qlc-');
     const singleBlockQuestions = await adminQuestions.addAllSingleBlockQuestionTypes('qlc-');
+    const repeatedQuestion = 'qlc-repeated-number';
+    await adminQuestions.addNumberQuestion(
+      repeatedQuestion, 'description', '$this\'s favorite number', '', 'qlc-enumerator');
+    const allQuestions = questions.concat(singleBlockQuestions);
 
-    await adminQuestions.updateAllQuestions(questions);
+    // Add to the front of the list because creating a new version of the enumerator question will
```
carolinedanzi, line 19: praise: thank you for this comment!

browser-test/src/support/admin_programs.ts
```diff
@@ -125,7 +125,27 @@ export class AdminPrograms {
...
     await this.page.click('#add-block-button');
 
-    await this.page.type('textarea', blockDescription);
+    await this.page.fill('textarea', blockDescription);
+    await this.page.click('#update-block-button');
+
+    for (const questionName of questionNames) {
+      await this.page.click(`button:text("${questionName}")`);
+    }
+    return await this.page.$eval('#block-name-input', el => (el as HTMLInputElement).value);
```
hmhsu, line 134: Suggestion (non-blocking): `await this.page.textContent('#block-name-input')` may be more readable.

browser-test/src/admin_program_translation.test.ts
```diff
@@ -164,6 +164,37 @@ describe('Admin can manage translations', () => {
     await endSession(browser);
   });
 
+  it('deleting help text in question edit view deletes all help text translations', async () => {
```
carolinedanzi, line 167: praise: thank you for adding a browser test!

browser-test/src/admin_program_translation.test.ts
```diff
@@ -164,6 +164,37 @@ describe('Admin can manage translations', () => {
...
+    await adminQuestions.goToQuestionTranslationPage(questionName);
+    const adminTranslations = new AdminTranslations(page);
+    await adminTranslations.selectLanguage('Spanish');
+    await adminTranslations.editQuestionTranslations('something different', 'help text different');
+
+    // Edit the question and delete the help text.
+    await adminQuestions.changeQuestionHelpText(questionName, "");
+
+    // Edit the question and add help text back
+    await adminQuestions.changeQuestionHelpText(questionName, "a new help text");
```
carolinedanzi, line 187: question: why add it back?

browser-test/src/question_lifecycle.test.ts
```diff
@@ -11,30 +11,41 @@ describe('normal question lifecycle', () => {
 
     const questions = await adminQuestions.addAllNonSingleBlockQuestionTypes('qlc-');
     const singleBlockQuestions = await adminQuestions.addAllSingleBlockQuestionTypes('qlc-');
+    const repeatedQuestion = 'qlc-repeated-number';
```
carolinedanzi, line 14: suggestion: Mind mentioning fixing this test in the PR description for future reference? I was confused why this was in here

carolinedanzi (approved):
Thank you Kendrick, .this is super helpful!

browser-test/src/admin_program_translation.test.ts
hmhsu, line 187: Nit (blocking): please use single quotes to be consistent with the rest of the file.
```suggestion
    await adminQuestions.changeQuestionHelpText(questionName, '');

    // Edit the question and add help text back
    await adminQuestions.changeQuestionHelpText(questionName, 'a new help text');

```

hmhsu (approved):

universal-application-tool-0.0.1/app/controllers/admin/AdminQuestionController.java
```diff
@@ -264,11 +264,18 @@ private void updateDefaultLocalizations(
...
             .updateTranslation(LocalizedStrings.DEFAULT_LOCALE, questionForm.getQuestionText()));
-    updated.setQuestionHelpText(
-        existing
-            .getQuestionHelpText()
-            .updateTranslation(
-                LocalizedStrings.DEFAULT_LOCALE, questionForm.getQuestionHelpText()));
+
+    // Question help text is optional. If the admin submits an empty string, delete
+    // all translations of it.
+    if (questionForm.getQuestionHelpText().isEmpty()) {
```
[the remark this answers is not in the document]
kendrickt: good call!

browser-test/src/admin_program_translation.test.ts
kendrickt: Its a way to check that the translations for the old help text are properly deleted. Otherwise there may be a case where:

1. create question with help text "aaa"
2. add spanish translation "aaa"
3. delete help text
4. add help text "bbb"

At this point we expect no translations to be present. It would be a bug if it preserved the "aaa" translation.

browser-test/src/support/admin_programs.ts
kendrickt: This wasn't working 🤷 . I can try it again.

browser-test/src/admin_program_translation.test.ts
carolinedanzi: Thanks for the explanation!

browser-test/src/question_lifecycle.test.ts
kendrickt: I added some white space and a comment about what `allQuestions` is.

browser-test/src/question_lifecycle.test.ts
kendrickt: 👍

browser-test/src/support/admin_programs.ts
kendrickt: It don't work 😦 . I'm gonna keep what I have.

Issue #775, opened 2021-04-13
TIs can search through their list of clients


neel-a-mehta:
Once they find a client, they can visit that client's profile page (#249). 

See [tiny.cc/cvf-ux-tracker](tiny.cc/cvf-ux-tracker) for more details.

neel-a-mehta:
@gcapiel says: "For user testing, we don’t need to have search, but given the list of applicants will grow large for launch the org user needs to be able to search for an applicant by first and/or last name."

PR #1266, opened 2021-05-24
Implement search and pagination for TI clients.

ndmckinley:
### Description
TIs can search through their list of clients.

Screen Shot 2021-05-24 at 3 58 27 PM
Screen Shot 2021-05-24 at 3 58 37 PM
Screen Shot 2021-05-24 at 4 00 28 PM
Screen Shot 2021-05-24 at 4 00 41 PM

### Checklist

### Issue(s)
Implements #775

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
+++ b/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
@@ -66,15 +66,15 @@ public class TrustedIntermediaryController {
     ImmutableList<Account> managedAccounts =
         trustedIntermediaryGroup.get().getManagedAccounts(search);
     int endOfListIndex = page.get() * PAGE_SIZE;
-    boolean hasNextTab = true;
-    if (managedAccounts.size() <= page.get() * PAGE_SIZE) {
+    int totalPageCount = (int) Math.ceil((double) managedAccounts.size() / PAGE_SIZE);
+    if (managedAccounts.size() <= endOfListIndex) {
       endOfListIndex = managedAccounts.size();
-      hasNextTab = false;
     }
     if (managedAccounts.size() <= (page.get() - 1) * PAGE_SIZE) {
       managedAccounts = ImmutableList.of();
       if (managedAccounts.size() == 0) {
-        hasNextTab = false;
+        // Display 1 page (which is empty)
+        totalPageCount = 1;
       } else {
         // If for some reason we're way past the end of the list, make sure the "previous"
         // button goes to the end of the list.
@@ -88,7 +88,7 @@ public class TrustedIntermediaryController {
             trustedIntermediaryGroup.get(),
             uatProfile.get().getApplicant().join().getApplicantData().getApplicantName(),
             managedAccounts,
-            hasNextTab,
+            totalPageCount,
             page.get(),
             search,
             request,
--- a/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
+++ b/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
@@ -10,9 +10,11 @@ import auth.ProfileUtils;
 import auth.UatProfile;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import forms.AddApplicantToTrustedIntermediaryGroupForm;
 import java.util.Optional;
 import javax.inject.Inject;
+import models.Account;
 import models.TrustedIntermediaryGroup;
 import org.pac4j.play.java.Secure;
 import play.data.Form;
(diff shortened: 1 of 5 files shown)


universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -26,6 +28,7 @@
 
 public class TrustedIntermediaryController {
 
+  private static final int PAGE_SIZE = 10;
```
bion, line 31: suggestion (non-blocking): bigger page size? maybe 30? i guess as a user i usually crank the page size up to whatever the max setting is

ndmckinley: The page is already super busy and with the search feature I figure it's probably better to make this window small - there's stuff below it, still.  If, given that, you still want it bigger, lmk and I'll do it.

universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -57,10 +63,34 @@ public Result dashboard(Http.Request request) {
...
     }
+    ImmutableList<Account> managedAccounts =
+        trustedIntermediaryGroup.get().getManagedAccounts(search);
+    int endOfListIndex = page.get() * PAGE_SIZE;
+    boolean hasNextTab = true;
+    if (managedAccounts.size() <= page.get() * PAGE_SIZE) {
+      endOfListIndex = managedAccounts.size();
+      hasNextTab = false;
+    }
+    if (managedAccounts.size() <= (page.get() - 1) * PAGE_SIZE) {
```
AWBaumann, line 74: It'd probably be useful to calculate the number of pages instead of infer them here. Then in the UI we can either show **Page 2 of 14** or **Results 11 - 20 of 135**

ndmckinley: Done!
Screen Shot 2021-05-25 at 11 25 27 AM
Screen Shot 2021-05-25 at 11 25 35 AM

universal-application-tool-0.0.1/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -80,18 +89,67 @@ public Content render(
...
+                        each(
+                            managedAccounts.stream()
+                                .sorted(Comparator.comparing(Account::getApplicantName))
+                                .collect(Collectors.toList()),
+                            account -> renderApplicantRow(account)))))
+            .withClasses(Styles.MB_16);
+    ContainerTag div = div();
+    if (page <= 1) {
+      div.with(new LinkElement().setText("∅").asButton());
+    } else {
```
AWBaumann, line 131: What does the `∅` button do? Perhaps just omit next/previous buttons if it doesn't apply? 

```suggestion
    if (page > 1) {
```

If you want to keep the button so that buttons are in a consistent location I understand that, too, but then we should just preserve the **←** text and disable the buttons. (We can do that in a follow-up if you don't want to bother with it now since we don't have disabled button styling yet - should be coming soon.)

universal-application-tool-0.0.1/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -80,18 +89,67 @@ public Content render(
...
+      div.with(
+          new LinkElement()
+              .setText("←")
+              .setHref(
+                  routes.TrustedIntermediaryController.dashboard(search, Optional.of(page - 1))
+                      .url())
+              .asButton());
+    }
+    div.with(new LinkElement().setText("Page " + page + " of " + totalPageCount).asButton());
```
AWBaumann, line 140: Can we convert this to plain text?

AWBaumann (approved):
Awesome work, just a few UI nits.

universal-application-tool-0.0.1/app/views/applicant/TrustedIntermediaryDashboardView.java
ndmckinley: Yep, just to keep things in the same place - but the buttons are actually links, right?  They're not `button`s, they're `a`s.  And disabling a "button" that is actually a link is super annoying / hard.  Once we get disabled button styling we can do that.

universal-application-tool-0.0.1/app/views/applicant/TrustedIntermediaryDashboardView.java
ndmckinley: I tried - the alignment goes all screwy and I can't figure out how to fix it.  This is the only trick I could figure out (even though it looks silly).  What classes would you apply to a `div` (or a `p`?  something else?) to get it to flow / align like one of these buttons?  I could give that a shot.

Issue #995, opened 2021-05-05
Bug: when Civiform admin clicks to add/remove question, changes to block name and description are thrown out


neel-a-mehta:
If a Civiform admin is working on a form and changes the block name or description and then clicks to add/remove a question, the changes they made to the block name or description are thrown out. We should ensure that the block changes get saved/updated when the admin adds or removes a question to the block.

kendrickt:
Related to #291

kendrickt:
Proposed Solution: 

Every action on this page is an atomic form submission, so we'll make the block name/description edit atomic by using a modal that forces the admin to save it. Then everything on this view will behave the same, and changes cannot be thrown out because 
block name/description changes are saved before any other action is allowed.

neel-a-mehta:
That works!

lianne-oy:
Ready for story acceptance @neel-a-mehta

neel-a-mehta:
The block metadata editor is now modal, so this is no longer a problem!

![image](https://user-images.githubusercontent.com/77128518/121080689-bda46480-c7a9-11eb-9854-d2a63d492e90.png)

PR #1269, opened 2021-05-25
Introduce Modals.

kendrickt:
### Description
Add the concept of a Modal to the HtmlBundle, and convert the admin block name/description form into a modal.

### Issue(s)
Story Acceptance for #995


universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -159,11 +159,35 @@ function attachLineClampListeners() {
   applicationCardDescriptions.forEach(el => el.addEventListener("click", removeLineClamp));
 }
```
AWBaumann, line 161: Can you put the modal specific scripts in its own class? Now that we're piling on the JS we should move away from using main.ts.

```
class ModalController {
    attachListeners() {
       ...
    }
    
    dismissModal(modal) {
       ...
    }
}
```

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -159,11 +159,35 @@ function attachLineClampListeners() {
...
 
+function attachModalListeners() {
+  const modalContainer = document.querySelector('#modal-container');
+  const modalGlassPane = document.querySelector('#modal-glass-pane');
+  // Find and connect each modal to its button
+  const modals = Array.from(document.querySelectorAll('.cf-modal'));
+  modals.forEach(modal => {
+    const modalButton = document.querySelector(`#${modal.id}-button`);
+    modalButton.addEventListener("click", function() {
+      modalContainer.classList.toggle('hidden');
```
AWBaumann, line 170: You probably don't need to worry about hiding the container, glassPane, and the modal here. Since there can be multiple hidden modals, you'll have to hide whichever one is visible, and you'll have to hide the container, but not the pane.

Alternatively, you could store the non-visible modals in a separate hidden div and just overwrite the content of modal when you need to show a new one. The only problem with that approach is you'll experience a bad UI redraw if you're chaining modals, but it does mean that you can have only one close button.

universal-application-tool-0.0.1/app/views/HtmlBundle.java
```diff
@@ -174,6 +185,24 @@ private ContainerTag renderMain() {
...
+    modals.forEach(
+        modal ->
+            modalContainer.with(
+                div()
+                    .with(
+                        button("x")
+                            .withId(modal.modalId() + "-close")
+                            .withClasses(BaseStyles.MODAL_CLOSE_BUTTON))
+                    .with(modal.getContainerTag())));
+    return modalContainer;
```
AWBaumann, line 203: Yeah, if you used the alternative method you'd just have this instead...

```suggestion
    ContainerTag modalContent = div.withId("modal-content");
    ContainerTag hiddenModals = div().hidden();
    modals.forEach(modal -> hiddenModals.with(modal.getContainerTag()));
    modalContainer.with(
                div()
                    .with(
                        button("x")
                            .withId(modal.modalId() + "-close")
                            .withClasses(BaseStyles.MODAL_CLOSE_BUTTON))
                    .with(modalContent);
    return modalContainer;
```

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
kendrickt: got it!

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
kendrickt: The glass pane is always shown, but the modal container is hidden. And the close button closes only the modal it is part of.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -357,6 +386,41 @@ private ContainerTag questionBankPanel(
...
+                .withClasses(
+                    Styles.MX_4,
+                    Styles.MY_1,
+                    Styles.INLINE,
+                    Styles.OPACITY_100,
+                    StyleUtils.disabled(Styles.OPACITY_50))
+                .attr("disabled", ""));
+    return Modal.builder("block-description-modal", blockDescriptionForm)
+        .setModalTitle("Block Name and Description")
+        .setButtonText("Edit Name and Description")
```
AWBaumann, line 420: Can you remove strings from the code flow? I know that we currently don't plan on localizing the admin view, but if we do, or if we want to go back and change a string value, it's a lot more manageable if these are static strings or at least declared at the top of the method body.

AWBaumann (approved):

universal-application-tool-0.0.1/app/views/HtmlBundle.java
```diff
@@ -115,7 +124,11 @@ public HtmlBundle setTitle(String title) {
 
   /** The page body contains: - header - main - footer */
   private ContainerTag renderBody() {
-    ContainerTag bodyTag = j2html.TagCreator.body(renderHeader(), renderMain(), renderFooter());
+    ContainerTag bodyTag = j2html.TagCreator.body().with(renderHeader()).with(renderMain());
+    if (modals.size() > 0) {
```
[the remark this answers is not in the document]
kendrickt: fixed

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockEditView.java
kendrickt: fixed

PR #1274, opened 2021-05-25
Language select redirect fix.

AWBaumann:
### Description
Fixed broken language select redirects.

I was using the wrong applicant id so the uri was resolving to another user and giving me 401 errors. I also modified the POST so that the redirect is part of the form data instead of overloading the route.

### DIFF
--- a/universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
+++ b/universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
@@ -3,6 +3,8 @@ package views.applicant;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static j2html.TagCreator.a;
 import static j2html.TagCreator.div;
+import static j2html.TagCreator.form;
+import static j2html.TagCreator.input;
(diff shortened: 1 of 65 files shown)


universal-application-tool-0.0.1/app/views/applicant/ApplicantLayout.java
```diff
@@ -90,21 +90,19 @@ private ContainerTag renderNavBar(Http.Request request, String userName, Message
   private ContainerTag getLanguageForm(Http.Request request, Optional<UatProfile> profile) {
     ContainerTag languageForm = div();
     if (profile.isPresent()) { // Show language switcher.
-      long userId = Long.parseLong(profile.get().getId());
+      long userId = profile.get().getApplicant().join().id;
```
hmhsu, line 93: Suggestion (non-blocking, if-minor): Alternatively, you could also pass in the actual applicant id from the controller. This ensures that the user, although very unlikely, doesn't inadvertently switch applicants within their account (because they should always be using the latest one and that is the one returned from profile).

hmhsu (approved):
Looks good!

Issue #664, opened 2021-04-02
Once file uploads enabled, output file URL in exported CSVs


neel-a-mehta:
#650 added CSV export functionality, but since file uploads weren't in place yet, didn't include functionality to embed file URLs/names in the outputted CSV.

So once file uploads are enabled (#404 etc), [see the requirements doc](https://docs.google.com/document/d/1zhs65i5PzxfkiZXPniPAxEDD257Iz4IP1JfxHY36hR8/edit?resourcekey=0-ByzGbFZlzPNxZTxRQCbVDQ#) and find a way to embed details about uploaded files in the CSV.

neel-a-mehta:
@bion was saying to avoid making users download ZIPs... and to avoid making users download data to their machine in the first place. Instead, for uploaded files, link to a secure URL where the user can view them. So this approach is actually quite good.

neel-a-mehta:
I downloaded a CSV as a test and I vaguely recall that it included columns with file URLs. Need to test that again.

neel-a-mehta:
A, I see -- the file key is put in the CSV, which is an OK start, but it's not actionable -- the admin wouldn't know where to actually get the file. We should convert those into S3 links (or wherever the file is stored).

neel-a-mehta:
Yup, file URL is correctly uploaded!  Here's the CSV download from the program admin's point of view. Click on it and you can download the original files.

![image](https://user-images.githubusercontent.com/77128518/123019760-f6d1fc80-d385-11eb-8810-5b5eb71215bd.png)

PR #1275, opened 2021-05-26
Use a file controller to relay file download

hmhsu:
### Description
Add a file controller to relay file download access so only authorized user can access those files.

### Checklist

### Issues
Works toward #664


universal-application-tool-0.0.1/app/controllers/FileController.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Secure(authorizers = Authorizers.Labels.ANY_ADMIN)
+  public Result adminShow(Request request, long programId, String fileKey) {
+    try {
+      ProgramDefinition program = programService.getProgramDefinition(programId);
+      checkProgramAdminAuthorization(profileUtils, request, program.adminName()).join();
+      return redirect(amazonS3Client.getPresignedUrl(fileKey).toString());
+    } catch (ProgramNotFoundException e) {
+      return notFound(e.toString());
+    } catch (CompletionException e) {
```
bion, line 65: question (blocking): why is a `CompletionException` unauthorized? seems like a broad error category mapped to a very specific response message

bion (approved):

universal-application-tool-0.0.1/app/controllers/FileController.java
hmhsu: The `CompletionException` can only be thrown with one reason here when checkProgramAdminAuthorization fails. I will add some comments.

Issue #1001, opened 2021-05-05
Let Civiform admins specify questions/blocks as optional or required


neel-a-mehta:
I'm not sure if we've agreed whether or not questions/blocks should be optional or required by default. Either is fine as long as we're consistent. Going required by default may help reduce the chance of mistakes on the Civiform admins' part, though, since most questions will probably be required.

kendrickt:
related to #322

lianne-oy:
Should be a simple function - optional / required for just questions, not blocks.  Moving to icebox; will open new issue for questions only.

kendrickt:
This is a duplicate of #1307 for questions, and #322 for blocks. Closing this.

PR #1276, opened 2021-05-26
Create PredicateDefinition and use in BlockDefinition

carolinedanzi:
### Description
1. Moves most of the predicate classes to `services.program.predicate` to avoid the program service depending on the applicant service. Applicant service now has classes pertaining to generating and evaluating predicates(requires applicant data), whereas the program service has classes used to define predicates. 
2. Moved the logic for converting a `LeafOperatorExpressionNode` to a `JsonPathPredicate` to a new class `JsonPathPredicateGenerator` - this will be expanded in the next PR to actually generate predicates. For now, it  just formats them
2. Creates `PredicateDefinition` and uses it in place of `Predicate` in `BlockDefinition`
3. Changes `PredicateExpressionNode` to an `AutoValue` class
4. Adds Jackson annotations to predicate classes and tests serialization/deserialization in `ProgramTest`

### Checklist

### Issue(s)
#1001


bion (approved):

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
```diff
@@ -111,13 +112,19 @@ public boolean isFileUpload() {
         .anyMatch(questionType -> questionType.equals(QuestionType.FILEUPLOAD));
   }
 
-  /** A {@link Predicate} that determines whether this is hidden or shown. */
+  /** A {@link PredicateDefinition} that determines whether this is hidden or shown. */
   @JsonProperty("hidePredicate")
-  public abstract Optional<Predicate> hidePredicate();
+  public abstract Optional<PredicateDefinition> predicate();
```
natsid, line 117: question: Should the @JsonProperty name match this method name?

Related suggestion/question: Maybe we should keep the `hide` prefix (or change it to `show`? depending how we are doing it. I guess we haven't thought through the details of how we'll offer both "show if" and "hide if" to the admin user...). Removing the prefix allows us some flexibility (i.e., if `PredicateDefinition` has some bool or enum like `HIDE_IF`/`SHOW_IF`, then maybe this predicate *isn't* one or the other, but could represent either one). This is not a well-formed suggestion... Just bringing it up as a discussion/thought.

natsid (approved):
One comment/question :)

universal-application-tool-0.0.1/app/services/program/BlockDefinition.java
carolinedanzi: Thanks Natalie, really good questions/comments! First, as far as I know we have to keep the "hidePredicate" json property because existing questions in the database use this field name - for example, this is a serialized block I pulled from staging's `/dev/seed`:

```json
"blockDefinitions" : [ {
    "id" : 1,
    "name" : "Block 1",
    "description" : "not translated",
    "repeaterId" : null,
    "hidePredicate" : null,
    "optionalPredicate" : null,
    "questionDefinitions" : [ {
      "id" : 6
    } ]
  } ],
```

For the method name, what would you think of something like `visibilityPredicate`? We currently have a plan to offer both `SHOW_BLOCK` and `HIDE_BLOCK` actions in `PredicateDefinition`, so this could capture both actions.

PR #1278, opened 2021-05-26
Disable caching.

### LINKED ISSUE #1135: Enumerator question deletions aren't staged properly
**Describe the bug**
In `ApplicantService#stageAndUpdateIfValid`, enumerator question deletions are not staged properly. This means if it is invalid, the deletions are lost to the void, and then very unexpected behavior can happen.

**To Reproduce**
1. Publish a program with an enumerator question, and a repeated question. (the ete enumerator program from browser tests can be used)
2. Create an "a" and "b" repeated entity, and then answer the repeated question with a different answer for "a" and "b".
3. Go back to edit the enumerator block
4. click the delete button for "a", > delete the string in the input field for "b" > save and continue
5. You should still be in the enumerator block, but there's an error that says "please fill in values for each row". Now fill in "c", and save and continue.
6. Make your way back to the summary review page

**Expected behavior**
Based on the user input, "a" should be deleted, and "b" should be renamed to "c".

However, the result is "a" is renamed to "c".

**Problem**
Enumerator deletions are not "staged", which means the state of the repeated entities after an invalid staging does not match what is stored, which means when the updates are finally valid and persisted, the outcome will not match what the user expects.

AWBaumann:
### Description
Disabling caching should eliminate the issue where the applicant hits the back button and the form is in a weird state because the underlying data has changed but wasn't refetched by the server. This will, however, result in more "Confirm form resubmission" warnings if a previous page relied on post data (like when we stage invalid changes).

### Issues
Fixes #1135


universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -227,8 +227,9 @@ play.cache {
 # to give Play greater security.
 #
 play.filters {
-  enabled += filters.LoggingFilter
+  enabled += filters.DisableCachingFilter
```
carolinedanzi, line 230: praise: thanks for keeping these alphabetical!

universal-application-tool-0.0.1/app/filters/DisableCachingFilter.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.concurrent.Executor;
+import javax.inject.Inject;
+import play.mvc.EssentialAction;
+import play.mvc.EssentialFilter;
+
+/** A filter to disable caching our responses. */
+public class DisableCachingFilter extends EssentialFilter {
```
carolinedanzi, line 11: question: I know nothing about Play's caching - why do we need a filter do turn off caching? Is it not just a setting somewhere?

carolinedanzi (approved):
Looks ok to me, though I really don't know anything about caching

universal-application-tool-0.0.1/app/filters/DisableCachingFilter.java
AWBaumann: This is to control the local cache. It puts a few properties in the response header to tell the client not to use a cached version of the page.

PR #1279, opened 2021-05-26
Evaluate predicates in applicant service

carolinedanzi:
### Description
1. Updates `JsonPathPredicateGenerator` to generate predicates based on the current list of `QuestionDefinitions` for a given program
2. Determines the `RepeatedEntity` context of the target predicate question (i.e. which question are basing this predicate on?)
2. Creates `PredicateEvaluator`, which uses the current `ApplicantData` and `JsonPathPredicateGenerator` to evaluate the root predicate node (or other nodes)

Next: Use the evaluator in `ReadOnlyApplicantProgramService` to determine which blocks to show based on predicates

### Checklist

### Issue(s)
#1001


universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
```diff
@@ -1,20 +1,45 @@
...
+import services.applicant.exception.InvalidPredicateException;
+import services.applicant.question.ApplicantQuestion;
 import services.program.predicate.LeafOperationExpressionNode;
 
+/** Generates {@link JsonPathPredicate}s based on the current applicant filling out the program. */
 public class JsonPathPredicateGenerator {
 
+  private final ImmutableMap<Long, ApplicantQuestion> questionsById;
+
+  public JsonPathPredicateGenerator(ImmutableList<ApplicantQuestion> programQuestions) {
```
kendrickt, line 16: (nit): the list _could_ have duplicates, which I think will break the collection into a map. Could this be a `set` object to guarantee uniqueness?

kendrickt: (blocking) for repeated questions, each `ApplicantQuestion` will share the same question definition id. Will that cause problems?

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
```diff
@@ -0,0 +1,49 @@
...
+import services.program.predicate.LeafOperationExpressionNode;
+import services.program.predicate.PredicateExpressionNode;
+
+/** Evaluates complex predicates based on the given {@link ApplicantData}. */
+public class PredicateEvaluator {
+
+  private final ApplicantData applicantData;
+  private final JsonPathPredicateGenerator predicateGenerator;
+
+  public PredicateEvaluator(
```
kendrickt, line 14: (non-blocking) the `JsonPathPredicateGenerator` has `ApplicantQuestion`s, which have `ApplicantData`, and this class also has `ApplicantData`. They may not be the same. Is that weird or nah?

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
```diff
@@ -0,0 +1,49 @@
...
+      ApplicantData applicantData, JsonPathPredicateGenerator predicateGenerator) {
+    this.applicantData = applicantData;
+    this.predicateGenerator = predicateGenerator;
+  }
+
+  /**
+   * Evaluate an expression tree rooted at the given {@link PredicateExpressionNode}. Will return
+   * true if and only if the entire tree evaluates to true based on the {@link ApplicantData} used
+   * to create this generator.
```
kendrickt, line 23: (nit) I think "this generator" should be "this evaluator".

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
kendrickt, line 16: (idea): if this object was constructed with the `Block`'s `RepeatedEntity` and the `QuestionDefinition`s, then the `Path` for the predicate could be generated by `repeatedEntity.getContextualizedPath().join(questionDefinition.getFormattedName()).

We'd have to "walk up" the `RepeatedEntity` chain to find the repeated entity whose enumerator id matches the question's enumerator id.

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
carolinedanzi: Good catch! Fixed

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
carolinedanzi: I think this should be fine - we're in the context of a single applicant, so the `ApplicantData` should be the same. Additionally, all we needed `ApplicantData` for in the generator is to get the path of the predicate question, which doesn't actually depend on `ApplicantData`.

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
carolinedanzi: Thanks Kendrick! I added this logic and unit tests - let me know if there are other tests you think would be helpful here

carolinedanzi: This should be fixed now - we use a list of `QuestionDefinition`s instead

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
```diff
@@ -1,20 +1,90 @@
...
+          String.format(
+              "Tried to apply a predicate based on question %d, which is not found in this"
+                  + " program.",
+              node.questionId()));
+    }
+
+    QuestionDefinition targetQuestion = questionsById.get(node.questionId());
+    Optional<RepeatedEntity> predicateContext;
+
+    if (targetQuestion.getEnumeratorId().isEmpty()) {
```
kendrickt, line 60: (maybe blocking): I don't think we need to treat this as a special case. Using `predicateContext.parent()` until it is an empty optional would eventually match `targetQuestion.getEnumeratorId()` if it is an empty optional. 

Getting rid of this as a special case might make the code simpler/easier to maintain. But if this is more readable, then please stick with this!

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
```diff
@@ -1,20 +1,90 @@
...
+      predicateContext = Optional.empty();
+    } else {
+      // Walk up the RepeatedEntity ancestors to find the right context. We need the context of the
+      // question in the predicate definition - that is, the one where the predicate question's
+      // enumerator ID matches the context's enumerator ID.
+      long enumeratorId = targetQuestion.getEnumeratorId().get();
+      predicateContext = this.currentRepeatedContext;
+      while (predicateContext.isPresent()
+          && predicateContext.get().enumeratorQuestionDefinition().getId() != enumeratorId) {
+        predicateContext = predicateContext.get().parent();
```
kendrickt, line 71: (maybe blocking): I'm concerned about the case where `this.currentRepeatedContext` doesn't contain a parent that matches `targetQuestion.getEnumeratorId()`. Then the loop ends with because `!predicateContext.isPresent()`, and this will behave as if the context should be the empty context, which is not correct.

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
kendrickt: (blocking): if that's the case, can we use `new ApplicantData()` as a useless placeholder in the `JsonPathPredicateGenerator`?

kendrickt (changes requested):

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
carolinedanzi: Good idea! Added

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
carolinedanzi: Good callout - I like early exits, which is why I originally put it in. Now, I think based on your other comment that this is necessary to differentiate between the "this is a top-level question with an empty context" case and the "this is an incorrectly configured predicate that led to an empty context" case. Thoughts?

universal-application-tool-0.0.1/app/services/applicant/predicate/JsonPathPredicateGenerator.java
carolinedanzi: Hopefully that would never happen, since I think that would mean the admin misconfigured the predicate. I threw an `InvalidPredicateException` for this case, mind taking a look?

kendrickt (approved):

universal-application-tool-0.0.1/app/services/applicant/predicate/PredicateEvaluator.java
carolinedanzi: Chatted offline - we're using the new `QuestionDefinition#getContextualizedPath` method here

Issue #940, opened 2021-04-30
Program admins can see a filterable, paginated list of applications their program has received


neel-a-mehta:
As specified [in this doc](https://docs.google.com/document/d/1xruNGvnI5FuvmoibzUujPvcIXkRGi7Kx1gSx6IpAd8Y/edit?usp=sharing), program admins should be able to see a list of all applications to their programs in a two-pane view. They should be able to filter viewable programs using a few filters specified in the doc. In addition, they should only see a limited number of applications on one page (50?) and have to click to other pages to see the rest. This will be become important as the admins get 1000s of applications.

ndmckinley:
That doc doesn't specify a two-pane view.  Do you see that as mandatory?

bion:
LGTM

PR #1281, opened 2021-05-26
Allow users to see old applications, add search + pagination for apps.

ndmckinley:
### Description
Screen Shot 2021-05-26 at 3 10 41 PM
Screen Shot 2021-05-26 at 3 10 51 PM

### Checklist

### Issue(s)
Implements #940.

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
@@ -11,10 +11,12 @@ import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import javax.inject.Inject;
 import models.Application;
+import models.Program;
 import org.pac4j.play.java.Secure;
 import play.mvc.Http;
 import play.mvc.Result;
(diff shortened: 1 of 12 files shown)


universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
```diff
@@ -37,6 +39,7 @@
   private final ExporterService exporterService;
   private final ProfileUtils profileUtils;
   private final Clock clock;
+  private static final int PAGE_SIZE = 10;
```
AWBaumann, line 42: Nit. Do we have a constants class anywhere?

AWBaumann (approved):

universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
ndmckinley: We have a handful but no generic ones I think - but it's a constant specific to the class that might want to be tweaked independently, so I would probably leave it here even so.

PR #1286, opened 2021-06-01
[WIP] Add a method to determine whether to show a block based on its predicate

carolinedanzi:
### Description
1. (Minor cleanup): Moves `summaryData` method to be next to the other overridden methods in the file
2. Adds a `showBlock` method that determines whether to show a block based on evaluating its predicate
2. This method is `protected` for now (for unit testing), but will change to `private` once it is used to determine lists of blocks for a given applicant

### Checklist

### Issue(s)
#1001


carolinedanzi:
Messed up git bases; reopened as #1288

PR #1292, opened 2021-06-02
Expire the cookie and jwt after 12 hours.

ndmckinley:
### Description
Helps make sure that people on guest computers don't wind up leaking information.

### Checklist

### Issue(s)


hmhsu (approved):

universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -155,11 +155,9 @@ play.http {
 
     sameSite = null
 
-    # Sets the max-age field of the cookie to 5 minutes.
-    # NOTE: this only sets when the browser will discard the cookie. Play will consider any
-    # cookie value with a valid signature to be a valid session forever. To implement a server side session timeout,
-    # you need to put a timestamp in the session and check it at regular intervals to possibly expire it.
-    #maxAge = 300
+    # 12 hours - auto-log-out after that long.
```
natsid, line 158: I wonder if this should be shorter - maybe something more like 30 min. Let's submit like this for now and we can discuss at standup today if we want to play with this length.

natsid (approved):

PR #1297, opened 2021-06-02
path column no longer needed.

### LINKED ISSUE #1085: Delete the `questions.path` column from the database.
#673 deleted it from our code paths. Now we can delete it from storage.

ndmckinley:
### Description
Fixes #1085.  tested by building things before this change, then making this change, then clicking around in `run-dev` - should work.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/conf/evolutions/default/27.sql
+++ b/universal-application-tool-0.0.1/conf/evolutions/default/27.sql
@@ -1,0 +1,5 @@
+# --- !Ups
+alter table questions drop column path;
+
+# --- !Downs
+alter table questions add column path varchar;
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/conf/evolutions/default/27.sql
```diff
@@ -0,0 +1,5 @@
+# --- !Ups
+alter table questions drop column path;
```
hmhsu, line 2: Suggestion (non-blocking): please add `if exists` after `drop column` so it is less prone to error in unexpected situation.

hmhsu (approved):

PR #1300, opened 2021-06-03
Deflake program admin browser test

hmhsu:
### Description
My current theory is that the selector is too generic that when the page loads, before styles are applied (we've seen this happened elsewhere), the selector selects the hidden template (wrong element) and thus fails the assertion. Using a more specific selector appears to work. Tested in staging with 10 successful runs.

### Checklist

### Issue(s)


AWBaumann (approved):

ndmckinley:
Thank you!

Issue #1228, opened 2021-05-20
Let CiviForm admin rearrange blocks in Form Builder


neel-a-mehta:
RIght now there's no way to move blocks around. You'd have to delete and recreate a block from scratch, which is very painful.

neel-a-mehta:
Just created a program on staging to play around with it. Logic and functionality appears to work!

![image](https://user-images.githubusercontent.com/77128518/122286565-d9dc8b80-cea4-11eb-8a6d-aa7d18a44e31.png)

neel-a-mehta:
Another screenshot 

![image](https://user-images.githubusercontent.com/77128518/122286711-ff699500-cea4-11eb-9c74-dbd50badc5c5.png)

PR #1321, opened 2021-06-08
Order program block definitions when reading and writing programs to DB

kendrickt:
### Description
`ProgramDefinition` has two new methods:

1. `hasOrderedBlockDefinitions` is a quick check to see if its block definitions have the right order (this is already true)
2. `orderBlockDefinitions` returns a program definition with blocks in the right order - either a new one, or the same one if blocks are already ordered.

In `models.Program`, we try to reorder the blocks before updates and saves. However, we cannot reorder blocks if the questions have not yet been loaded in, and there are lots of places where this is true. For example:
* In `models.Program#loadProgramDefinition`
* `ProgramRepository#createOrUpdateDraft`
* In lots of the `VersionRepository`

If the questions are not yet loaded, then we don't really need to check for order. Methods that update programs will have questions loaded, and will be saved with proper block order. Methods that require blocks in proper order (e.g. methods that move blocks around) should explicitly check for proper block order since its a pretty cheap operation.

### Checklist

### Issue(s)
Progress towards #1228


universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -47,12 +50,77 @@ public static Builder builder() {
...
+          toBuilder()
+              .setBlockDefinitions(reorderBlockDefinitionsInner(getNonRepeatedBlockDefinitions()))
+              .build();
+      orderedProgramDefinition.hasOrderedBlockDefinitionsMemo = true;
+      return orderedProgramDefinition;
+    }
+    return this;
+  }
+
+  private ImmutableList<BlockDefinition> reorderBlockDefinitionsInner(
```
natsid, line 77: nit: rename to match the new name of the public method? (reoder --> order). And maybe same with models.Program.java

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -47,12 +50,77 @@ public static Builder builder() {
...
   public abstract LocalizedStrings localizedDescription();
 
-  /** The list of {@link BlockDefinition}s that make up the program. */
   public abstract ImmutableList<BlockDefinition> blockDefinitions();
 
   /** The list of {@link ExportDefinition}s that make up the program. */
   public abstract ImmutableList<ExportDefinition> exportDefinitions();
 
+  /**
+   * Returns a program definition with block definitions such that each enumerator block is
```
natsid, line 59: praise: I can't remember if this is what the comment said before we chatted, but I think the docstring is pretty dang good!

natsid (approved):

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
kendrickt: done!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -47,12 +50,77 @@ public static Builder builder() {
   /** A human readable description of a Program, localized for each supported locale. */
   public abstract LocalizedStrings localizedDescription();
 
-  /** The list of {@link BlockDefinition}s that make up the program. */
```
hmhsu, line 50: Question (blocking for an answer): Did you mean to remove this docstring?

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -47,12 +50,77 @@ public static Builder builder() {
...
+   * This method should be treated as VISIBLE FOR TESTING.
+   *
+   * <p>True indicates that each enumerator block in {@link #blockDefinitions()} is immediately
+   * followed by all of its repeated and nested repeated blocks.
+   */
+  public boolean hasOrderedBlockDefinitions() {
+    if (hasOrderedBlockDefinitionsMemo == null) {
+      Deque<Long> enumeratorIds = new ArrayDeque<>();
+
+      // Walk through the list of block definitions, checking that repeated blocks are
```
hmhsu, line 100: Suggestion (blocking): Please finish the comment.

universal-application-tool-0.0.1/test/services/program/ProgramServiceImplTest.java
```diff
@@ -699,4 +704,86 @@ public void getNotificationEmailAddresses() {
...
+                    .setDescription("description")
+                    .addQuestion(
+                        ProgramQuestionDefinition.create(
+                            testQuestionBank.applicantName().getQuestionDefinition()))
+                    .build())
+            .build();
+    ObjectMapper mapper =
+        new ObjectMapper().registerModule(new GuavaModule()).registerModule(new Jdk8Module());
+    String updateString =
+        String.format(
```
hmhsu, line 780: Suggestion (if-minor): is there a specific reason that we can't utilize `program.save()` here? I think it would be more robust if we can use that.

hmhsu (approved):

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
kendrickt: nope! thanks for catching!!!!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
kendrickt: done!

universal-application-tool-0.0.1/test/services/program/ProgramServiceImplTest.java
kendrickt: Block order is fixed on saves, so I think we need to do a direct sql thing.

kendrickt: added comments about this, and added `OnRead` to test name.

PR #1324, opened 2021-06-09
Use the same wording for returning to applicant program page

hmhsu:
### Description
Use the same wording for the link of returning to applicant program page so the browser test passes.


natsid (approved):
Thanks @hmhsu ! One question: What's with the empty `dummy` file added?

hmhsu:
> Thanks @hmhsu ! One question: What's with the empty `dummy` file added?

Chatted offline, that was added for an irrelevant change and is not needed. Removed.

Issue #322, opened 2021-03-02
Civiform admin can add show/hide logic in form builder


neel-a-mehta:
On each block, they'd be able to specify when the block should be shown to the applicant: always, or only if certain criteria are met.

If they specify that certain criteria should be met, we will open up an area where they can define the criteria. Criteria would be of one of these formats:

* If `answer X` is less than, equal to, not equal to, or greater than `value Y` (for numbers)
* If `answer X` contains, doesn't contain, equals, or does not equal `value Y` (for text)
* If `condition A` and/or `condition B`

For example:

* If `income` is less than `$50,000`
* If `city` is not `Seattle`
* If `age` is greater than `5` _and_ `number of family members` is equal to `3`

We'd probably need to have a series of dropdowns to let people build these expressions, and some way for people to create multiple rows or dropdowns to specify combinations like "and"s and "or"s. It's going to be quite messy, unfortunately.

## Examples / proofs of concept

We're going to have to build a Boolean Logic engine, basically. The good news is that there are several examples of good UI/UX for this, such as this one: https://ux.stackexchange.com/questions/1737/intuitive-interface-for-composing-boolean-logic

lianne-oy:
On track to be done on time (mid next week).

carolinedanzi:
### Test Cases

**Applicant View**
1. Applicant answers a block so that a later block is hidden
2. Applicant answers a block so that a later block is shown
3. Applicant answers a block so a predicate block is shown, then goes back and modifies their answer so the block is then hidden. The hidden, answered block should not show in the applicant summary nor be submitted with their application
4. Applicant answers a block so a predicate block is hidden, then goes back and modifies their answer so the block is now shown. They must answer this block in order to submit their application

**Admin View**
1. An admin can add a simple predicate for a block that depends on an earlier question
2. An admin can add a complex predicate using and/or
3. If an admin reorders blocks, they get a warning if a block with a predicate moves to an illegal state (i.e. the question it depends on no longer exists in an earlier block)
4. If an admin removes a question from a block that a later predicate depends on, they see a warning or cannot perform the question removal

natsid:
Remaining work for simple predicates:

- [ ] Disallow question removal from a program if the program has a block with a predicate that depends on that question (should look similar to the reordering block warning - error toast)
- [ ] Display the current predicate on the Block Edit view
  - [ ] Would be good to have a pretty-printing function that can be used on both that page and on the predicate page itself
[...]

PR #1327, opened 2021-06-09
Build out more of the Admin show/hide block predicate UI

natsid:
### Description
Build out more of the Admin show/hide block predicate UI.
* Make the list of question definition options a list of buttons that each open a modal with a form for creating a single predicate.

Other related changes in this PR:
* Update Modal to take custom trigger button content instead of just text. This is so that we can have richer button content with an icon and specific layout.
* Add a height attribute on input and select form fields so that they have uniform height with respect to each other.

Works that still remains:
* Notice there is no submit button yet... The form needs an endpoint to POST to that actually processes and saves the form data.
* The form should have different dropdown options depending on the question type and scalar type.
* Once the above are done, we can add a link to this page from the block edit page.

### Screenshot
Screen Shot 2021-06-09 at 4 27 40 PM

### Checklist

### Issue(s)
Works toward #322


universal-application-tool-0.0.1/app/views/components/Modal.java
```diff
@@ -60,7 +72,9 @@ public static ModalBuilder builder(String modalId, Tag content) {
 
     // Optional fields. See #setOptionalFields().
     private String modalTitle;
-    private String buttonText;
+    private String triggerButtonText;
+
+    private Tag triggerButtonContent = null;
```
bion, line 77: Suggestion (blocking):
```suggestion
    private Optional<Tag> triggerButtonContent = Optional.empty();
```

bion (approved):

carolinedanzi (approved):
This looks really nice!

universal-application-tool-0.0.1/app/views/components/Modal.java
natsid: good call

PR #1331, opened 2021-06-09
ProgramService methods that add blocks add them in the right place.

kendrickt:
### Description
Add a helper method for `ProgramServiceImpl#addBlockDefinition` and `ProgramServiceImpl#addRepeatedBlockDefinition` so they get added in the right place.

### Checklist

### Issue(s)
Progress towards #1228

### DIFF
--- a/universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
+++ b/universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
@@ -7,8 +7,11 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import forms.BlockForm;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -240,16 +243,77 @@ private void validateProgramText(
...
+    }
+
+    // At this point, insertIndex is AFTER the last repeated or nested repeated block of the
+    // enumerator. This might be off the end of the list.
+    ImmutableList.Builder<BlockDefinition> newBlockDefinitionsBuilder = ImmutableList.builder();
+    orderedProgramDefinition.blockDefinitions().stream()
+        .limit(insertIndex)
+        .forEach(blockDefinition -> newBlockDefinitionsBuilder.add(blockDefinition));
+    newBlockDefinitionsBuilder.add(newBlockDefinition);
+    orderedProgramDefinition.blockDefinitions().stream()
```
natsid, line 308: question: Not sure why this is so hard for me to understand 😆  but why do we need this line with the `skip`? I feel like line 304 already takes care of adding the other block defs?

natsid (approved):

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
kendrickt: `limit` adds the first N blocks. Then the new block is added. And then the remaining blocks are added with `skip`.

PR #1334, opened 2021-06-10
CSV export gets answer data through readonly applicant program service

hmhsu:
### Description
Use answer retrieved from readonly applicant program service in CSV export. This allows us to further process the raw value before exporting. Also added all question types to CSV export test

### Checklist

### Issue(s)
#664 
#1371


universal-application-tool-0.0.1/app/services/export/CsvExporter.java
```diff
@@ -50,24 +57,30 @@ private void writeHeadersOnFirstExport(CSVPrinter printer) throws IOException {
...
    */
-  public void export(Application application, Writer writer) throws IOException {
+  public void export(
+      Application application, ReadOnlyApplicantProgramService roApplicantService, Writer writer)
+      throws IOException {
     CSVPrinter printer = new CSVPrinter(writer, CSVFormat.DEFAULT.withFirstRecordAsHeader());
 
     this.writeHeadersOnFirstExport(printer);
 
+    ImmutableMap<String, AnswerData> answerMap = roApplicantService.getSummaryDataMap();
```
ndmckinley, line 67: Why can't this be done for the demographic export again?

ndmckinley (commented):
Looks good, but I don't follow why the deviation is needed.

universal-application-tool-0.0.1/app/services/export/CsvExporter.java
hmhsu: It probably can. The key, however, is highly dependent on the program (basically it's block id + question index in the block). So for each question I will have to find the programs they are in, and generate the correct key based on the application. I feel that's too much of a trouble than doing this, but maybe it's not that hard.

hmhsu: I can actually just make all answers in the application a map and select by Path.
hmm I guess it's doable. I will give it a try.

ndmckinley: If you can that would make me happy because it makes some of my job a lot easier, ha.  Thanks for giving it a shot!

hmhsu: Done.

universal-application-tool-0.0.1/app/services/export/CsvExporter.java
```diff
@@ -50,17 +54,22 @@ private void writeHeadersOnFirstExport(CSVPrinter printer) throws IOException {
...
+        roApplicantService.getSummaryData().stream()
+            .flatMap(data -> data.scalarAnswersInDefaultLocale().entrySet().stream())
+            .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
     for (Column column : getColumns()) {
       switch (column.columnType()) {
         case APPLICANT:
-          Optional<String> value =
-              application.getApplicantData().readAsString(column.jsonPath().orElseThrow());
-          printer.print(value.orElse(EMPTY_VALUE));
+          String value = getValueFromAnswerMap(column, answerMap);
```
ndmckinley, line 71: Can you do the same thing for APPLICANT_OPAQUE, below?

ndmckinley (approved):
heck yeah.  thank you!!

universal-application-tool-0.0.1/app/services/export/CsvExporter.java
hmhsu: Sure thing.

hmhsu:
Oh, ha. The readonly applicant service only knows about active programs/questions it looks like. I think we need it to support past programs and questions.

ndmckinley:
Hm, we shouldn't have to - you can just filter out any question that isn't active.

hmhsu:
The browser test is blocked on #1368

neel-a-mehta:
This fixes #1371 btw

Issue #931, opened 2021-04-29
If I don't select any checkboxes, it thinks I didn't answer the question


carolinedanzi:
**Describe the bug**
When an applicant is presented with a checkbox question and they don't choose any of the options, we should be able to detect this and treat it as their answer (i.e. none of the checkboxes apply to them). Instead, we think they haven't answered the question

**To Reproduce**
1. Seed the database with /dev/seed
2. Apply to the mock program entirely
3. Note that it says the application was complete
4. Try to apply again - it jumps to the block with the checkbox, since it thinks the question was not answered

**Additional context**
Not selecting any checkboxes is counted as an answer

carolinedanzi:
@syntapy is handling this one (thank you!)

carolinedanzi:
This is an interesting interaction between "applicant answered this question" (which in our system means their data has the path of the question) and "applicant has never seen this question" (which means the question path does not appear in their data). 

The logic for checking whether a checkbox question has been answered can be found in the `isAnswered` https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/services/applicant/question/MultiSelectQuestion.java

In this case, we don't send any form data, so we don't know the checkbox question is present in the form. Instead, we should include a hidden input field that is submitted with the form indicating that a checkbox question is present, so that we know to write the checkbox question path to `ApplicantData`

carolinedanzi:
This was closed by #1335

PR #1335, opened 2021-06-10
Non-selected checkbox questions are counted as answered

syntapy:
### Description
When an applicant does not select any items in a checkbox question, it counts as being answered

### Checklist
- [x] Modify browser test that skips a checkbox question

### Issue(s)
#931

### DIFF
--- a/browser-test/src/application_review.test.ts
+++ b/browser-test/src/application_review.test.ts
@@ -13,6 +13,8 @@ describe('normal application flow', () => {
     await adminQuestions.addEmailQuestion('email-q');
     await adminQuestions.addDropdownQuestion('ice-cream-q', ['chocolate', 'banana', 'black raspberry']);
     await adminQuestions.addCheckboxQuestion('favorite-trees-q', ['oak', 'maple', 'pine', 'cherry']);
+    await adminQuestions.addCheckboxQuestion('favorite-rats-q', ['sewage', 'laboratory', 'bubonic', 'giant']);
+    await adminQuestions.addCheckboxQuestion('scared-of-q', ['dogs', 'bees', 'spiders', 'the dark', 'clowns']);
     await adminQuestions.addAddressQuestion('address-q');
     await adminQuestions.addFileUploadQuestion('fileupload-q');
     await adminQuestions.addNameQuestion('name-q');
@@ -25,6 +27,7 @@ describe('normal application flow', () => {
     await adminPrograms.editProgramBlock(programName, 'block description', ['date-q', 'address-q', 'name-q', 'radio-q', 'email-q']);
     await adminPrograms.addProgramBlock(programName, 'another description', ['ice-cream-q', 'favorite-trees-q', 'number-q', 'text-q']);
     await adminPrograms.addProgramBlock(programName, 'third description', ['fileupload-q']);
+    await adminPrograms.addProgramBlock(programName, 'fourth description', ['scared-of-q', 'favorite-rats-q']);
 
     await adminPrograms.gotoAdminProgramsPage();
     await adminPrograms.expectDraftProgram(programName);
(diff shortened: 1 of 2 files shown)


browser-test/src/application_review.test.ts
```diff
@@ -74,18 +79,23 @@ describe('normal application flow', () => {
...
     await applicantQuestions.answerFileUploadQuestion('file key');
     await applicantQuestions.clickNext();
 
-    // Applicant submits answers from review page.
+    // fill 4th application block. 
+    // skip one checkbox question.
+    await applicantQuestions.answerCheckboxQuestion(['clowns']);
+    await applicantQuestions.clickNext();
+
+    // submit
```
syntapy, line 98: Do I need to modify the test to verify that if it skips a non-checkbox question that it results in a review page with 'continue' on it? i.e. to make sure that me adding stuff didn't make it artificially result in 'submit'

carolinedanzi (commented):
Hi Sami! Nice work on this, it's a tricky issue! While reviewing your PR I noticed it looks like someone else has added logic in `ApplicantServiceImpl#stageNormalUpdates` since the last time we talked that partially addresses this issue - basically, it adds an empty array to "clear" the existing data before updating. However, it doesn't quite work for checkbox questions - if I don't answer it, it won't let me submit. But if I go back to edit the answer and still don't answer the checkbox question, it does register as "answered" and then lets me submit. It's a weird issue - mind looking into this and updating so that it considers it "answered" the first time I don't answer? 

I like that you added browser tests - let's definitely keep those. To build off the newest code, we should leverage the new hidden input field added in `CheckboxQuestionRenderer`. Are you willing to update your approach?

Happy to chat in open office hours Friday!

syntapy:
@carolinedanzi Oh yeah that makes sense. IYeah doing similar things in two different places (without the change in approach) could turn ugly 6-months from now when someone's trying to change something or whatnot...that is true in general, no?

I had noticed that other behavior you mentioned but it didn't register as something I should pay attention to

carolinedanzi:
> @carolinedanzi Oh yeah that makes sense. Yeah doing similar things in two different places (without the change in approach) could turn ugly 6-months from now when someone's trying to change something or whatnot...that is true in general, no?
> 
> I had noticed that other behavior you mentioned but it didn't register as something I should pay attention to

Yes, great point! Maintenance is harder if we have multiple related special cases. Thank you for being willing to modify your approach and reduce the maintenance burden!

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -233,11 +233,13 @@ public void maybeClearArray(Path path) {
 
   private void putAt(Path path, Object value) {
     checkLocked();
+    putParentIfMissing(path);
     jsonData.put(path.parentPath().toString(), path.keyName(), value);
   }
 
   private void addAt(Path path, Object value) {
     checkLocked();
+    putParentIfMissing(path);
```
syntapy, line 242: Should I add this one too? Not relevent to the checkbox thing I guess but seemed right based on concistency

carolinedanzi: Good catch, I think that is correct. Thanks for updating!

carolinedanzi (approved):

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
hmhsu: Hey, this looks good. I have one request though, reading the `putParentIfMissing` function, there is a desire to simplify the function and remove the recursion. These two functions are called by the `putParentIfMissing` function and calling `putParentIfMissing` here further complicates the recursion tree. Would you instead add the function call in `maybeClearArray` directly before `putAt`? I think that will work, too, and leave a less complicated codebase for anyone who aspires to remove the recursion in the future.

syntapy: Ok I did it

hmhsu: Thanks!

hmhsu (approved):

PR #1339, opened 2021-06-10
Pre application summary

AWBaumann:
### Description
Uses the review screen as a program preview screen. Cards now have "Program info (internal)" and "Program details (external)" links. The `apply` button on the program index screen takes you directly to this page. 

##### Program Cards
Program cards now contain 2-3 click targets. The `Apply` button goes to the **Program Preview** page. The `Program details` link goes to the internal program info page. The `External Site` link is only visible if an external url was provided by a CiviForm admin during program creation.
 
image

###### Program Preview
The program preview page looks like the program review page except you can only jump to the next unanswered question. (We may want this to be functionally identical to the review page, but I'd just like to get this in and we can iterate after we've had a chance to play with it a little.)

image

### Checklist

### Issue(s)


browser-test/src/admin_program_translation.test.ts
```diff
@@ -63,7 +63,7 @@ describe('Admin can manage translations', () => {
 
 
     // Expect program details link to contain 'Detalles del programa' with link to 'http://seattle.gov'
-    expect(await page.innerText('.cf-application-card a[href="http://seattle.gov"]')).toContain('Detalles del programa');
+    expect(await page.innerText('.cf-application-card a[href="http://seattle.gov"]')).toContain('External Site');
```
natsid, line 66: question: Why did this change to english?

browser-test/src/support/applicant_questions.ts
```diff
@@ -82,7 +82,7 @@ export class ApplicantQuestions {
 
   async submitFromReviewPage(programName: string) {
     // Assert that we're on the review page.
-    expect(await this.page.innerText('h1')).toContain('Application summary');
+    expect(await this.page.innerText('h1')).toContain('Program review');
```
natsid, line 85: question: I see "Program preview" rather than "review" in the screenshot on the PR description.  Is this testing the last page rather than the first page version?

browser-test/src/admin_program_translation.test.ts
natsid: Maybe we just need to add an "External site" Spanish translation to messages.es-us.

universal-application-tool-0.0.1/conf/messages
```diff
@@ -85,6 +85,18 @@ button.submit=Submit
...
 button.continue=Continue
 
+# The text that appears next to the first unanswered question that the applicant can click on to continue filling out the program.
+link.begin=Start here
+
+# The text that appears next to an answered question that the applicant can click on to modify the response.
+link.edit=Edit
+
+# Program Prewiew Title
+title.programPreview=Program preview
```
natsid, line 95: suggestion: Thoughts on using "Program application preview" and "Program application review"? Since these pages are showing the _application_ information rather than info about the program itself.

browser-test/src/support/applicant_questions.ts
natsid: Nvm - saw the titles in messsages

browser-test/src/admin_program_translation.test.ts
AWBaumann: Yeah, it's in English because the Spanish translation isn't available. Will update.

universal-application-tool-0.0.1/conf/messages
AWBaumann: TBH, I never understood why they're referred to as _Programs_ and prefer the term _Application_, so I'm happy to make the change.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -153,36 +163,55 @@ private ContainerTag renderQuestionSummary(AnswerData data, Long applicantId) {
...
+              : routes.ApplicantProgramBlocksController.review(
+                      applicantId, data.programId(), data.blockId())
+                  .url();
+
+      ContainerTag editAction =
+          new LinkElement()
+              .setHref(editLink)
+              .setText(editText)
+              .setStyles(
+                  Styles.ABSOLUTE,
```
natsid, line 190: I'm pretty hesitant about absolute style 😛  What are your thoughts on using flexbox or grid to put this Edit button here? Non-blocking, because it looks like it works and we don't need to change this just for the heck of it! But I often find flexbox and grid very delightful to work with.

natsid: Ohh, maybe I misunderstood how absolute is being used here because I see flex used above and below. Disregard if so.

natsid (approved):

universal-application-tool-0.0.1/conf/messages
natsid: Ohh cool! I noticed that you changed the name from "Application summary" to "Program review", so I wasn't sure which one we were going with.

PR #1345, opened 2021-06-11
Admin predicate form: Display the right scalars for question type

natsid:
Also:
* Fix the "Value" field for multi-option questions. Now the admin will see a dropdown of options for the "Value" instead of a freeform input box.
* Don't include enumerator questions in the list of available predicate question definitions. This is because enumerators do not have any "scalar" values associated with them. If we want to support predicates based on enumerator questions, it will take a bit more design and special case implementation. This is not in scope for the current goal.

### Screenshots
Screen Shot 2021-06-11 at 1 40 00 PM
Screen Shot 2021-06-11 at 1 40 22 PM

### Checklist
  - Will add browser tests for this flow once it's working e2e. 

### Issue(s)
Works toward #322

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
@@ -41,6 +41,42 @@ public enum Scalar {
   UPDATED_AT,
   PROGRAM_UPDATED_IN;
 
+  @Override
+  public String toString() {
+    switch (this) {
+      case CITY:
+        return "city";
+      case DATE:
+        return "date";
+      case EMAIL:
+        return "email";
+      case FILE_KEY:
+        return "file key";
+      case FIRST_NAME:
+        return "first name";
+      case LAST_NAME:
+        return "last name";
+      case LINE2:
+        return "address line 2";
+      case MIDDLE_NAME:
+        return "middle name";
+      case NUMBER:
+        return "number";
+      case SELECTION:
+        return "selection";
+      case STATE:
+        return "state";
+      case STREET:
+        return "street";
+      case TEXT:
+        return "text";
+      case ZIP:
+        return "ZIP code";
+      default:
+        return this.toString();
+    }
+  }
+
   private static final ImmutableMap<Scalar, ScalarType> ADDRESS_SCALARS =
       ImmutableMap.of(
           STREET, ScalarType.STRING,
--- a/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
@@ -18,63 +18,38 @@ import services.question.types.ScalarType;
  * <p>EXISTING SCALARS SHOULD NOT BE MODIFIED. The Scalar enum should be treated as append-only.
  */
 public enum Scalar {
-  CITY,
-  DATE,
-  EMAIL,
-  FILE_KEY,
-  FIRST_NAME,
-  LAST_NAME,
-  LINE2,
-  MIDDLE_NAME,
-  NUMBER,
-  SELECTION,
-  STATE,
-  STREET,
-  TEXT,
-  ZIP,
+  CITY("city"),
+  DATE("date"),
+  EMAIL("email"),
+  FILE_KEY("file key"),
+  FIRST_NAME("first name"),
+  LAST_NAME("last name"),
+  LINE2("address line 2"),
+  MIDDLE_NAME("middle name"),
+  NUMBER("number"),
+  SELECTION("selection"),
+  STATE("state"),
+  STREET("street"),
+  TEXT("text"),
+  ZIP("ZIP code"),
 
   // Special scalars for Enumerator updates
-  DELETE_ENTITY, // This is used for deleting enumerator entries
-  ENTITY_NAME, // This is used for adding/updating enumerator entries
+  DELETE_ENTITY("delete entity"), // This is used for deleting enumerator entries
+  ENTITY_NAME("entity name"), // This is used for adding/updating enumerator entries
 
   // Metadata scalars
-  UPDATED_AT,
-  PROGRAM_UPDATED_IN;
+  UPDATED_AT("updated at"),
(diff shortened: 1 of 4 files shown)


universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -273,12 +273,13 @@ public boolean hasEnumerator(long enumeratorId) {
    * Returns a list of the question definitions that may be used to define predicates on the block
    * definition with the given ID.
    *
-   * <p>The available question definitions for predicates satisfy BOTH of the following:
+   * <p>The available question definitions for predicates satisfy ALL of the following:
```
carolinedanzi, line 276: praise: thank you for keeping the docstrings up-to-date!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -105,6 +111,39 @@ private Modal predicateFormModal(
 
   private Tag renderPredicateForm(
       String blockName, QuestionDefinition questionDefinition, Tag csrfTag) {
+
+    ImmutableMap<Scalar, ScalarType> scalars;
+    try {
+      scalars = Scalar.getScalars(questionDefinition.getQuestionType());
+    } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
+      return div()
```
carolinedanzi, line 119: suggestion (non-blocking): Could we add a comment here that this should never happen?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -105,6 +111,39 @@ private Modal predicateFormModal(
...
+          .withText("Sorry, you cannot create a show/hide predicate with this question type.");
+    }
+
+    ImmutableList.Builder<SimpleEntry<String, String>> scalarOptionsBuilder =
+        ImmutableList.builder();
+    scalars.forEach(
+        (scalar, type) -> {
+          scalarOptionsBuilder.add(new SimpleEntry<>(scalar.toString(), scalar.toString()));
```
carolinedanzi, line 127: question (blocking): If the html value is the human-readable string, how do we plan to convert back to the enum on form submission?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -105,6 +111,39 @@ private Modal predicateFormModal(
...
+        });
+
+    ContainerTag valueField;
+    if (questionDefinition.getQuestionType().isMultiOptionType()) {
+      ImmutableList<SimpleEntry<String, String>> valueOptions =
```
carolinedanzi, line 132: thought: I'm not a fan of `SimpleEntry`. I might change `SelectWithLabel` to use a map instead, or generalize to just `Entry`. Perhaps this will be my late Friday code cleanup :)

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -105,6 +111,39 @@ private Modal predicateFormModal(
...
+
+    ImmutableList.Builder<SimpleEntry<String, String>> scalarOptionsBuilder =
+        ImmutableList.builder();
+    scalars.forEach(
+        (scalar, type) -> {
+          scalarOptionsBuilder.add(new SimpleEntry<>(scalar.toString(), scalar.toString()));
+        });
+
+    ContainerTag valueField;
+    if (questionDefinition.getQuestionType().isMultiOptionType()) {
```
carolinedanzi, line 131: suggestion (blocking): Mind writing a comment explaining why this is necessary?

universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
```diff
@@ -41,6 +41,42 @@
   UPDATED_AT,
   PROGRAM_UPDATED_IN;
 
+  @Override
+  public String toString() {
```
carolinedanzi, line 45: question: Thoughts on making this a `asHumanReadableString` method rather than overriding? Not sure it matters too much, but might be helpful for clarity

carolinedanzi (approved):
Nice work! I thought this would be a nightmare, but your solution is so clean! Way to go!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: yes!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: Yeah, good point. I think this is the wrong way to do this and that I need to use scalar.name() for the "key" in the option, even if we use a display string for the name. That should work right?

Related: I'm gonna update the `Scalar` enum. Instead of having that toString method with the case statements, I'll make it look more like the `Operator` enum by having it take a displayString.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: That would be awesome!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: good idea

universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
natsid: Let me know what you think of the new `Scalar` implementation.

Issue #1326, opened 2021-06-09
Don't display unreachable top navigation links to program admin


bion:
Program admins can't view most of the top navigation links and get a forbidden error when they try. Remove the ones they can't reach.

neel-a-mehta:
Here's a screenshot -- hide the Questions, Programs, Versions, and Intermediaries links when the user is logged in as a Program Admin.

![image](https://user-images.githubusercontent.com/77128518/121432434-aa7bcb00-c948-11eb-9b65-6e1d2a6b3544.png)

neel-a-mehta:
Yup, top links no longer render for program admins.

![image](https://user-images.githubusercontent.com/77128518/121930025-e380ba80-cd0f-11eb-9438-1343b9d44526.png)

PR #1349, opened 2021-06-11
Don't render nav links for program admin

bion:
see #1326

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
+++ b/universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
@@ -21,11 +21,26 @@ import views.style.Styles;
 
 public class AdminLayout extends BaseHtmlLayout {
 
+  public enum AdminType {
+    CIVI_FORM_ADMIN,
+    PROGRAM_ADMIN
+  }
+
   private static final String[] FOOTER_SCRIPTS = {"preview", "questionBank"};
+
+  private AdminType adminType = AdminType.CIVI_FORM_ADMIN;
 
   @Inject
   public AdminLayout(ViewUtils viewUtils, Config configuration) {
     super(viewUtils, configuration);
+  }
+
+  /**
+   * Sets this layout's admin type to PROGRAM_ADMIN, used to determine which navigation to include.
+   */
+  public AdminLayout setProgramAdminType() {
+    adminType = AdminType.PROGRAM_ADMIN;
+    return this;
   }
 
   public Content renderCentered(HtmlBundle bundle) {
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
```diff
@@ -73,15 +87,24 @@ private ContainerTag renderNavBar() {
...
+      return adminHeader.with(headerLink("Logout", logoutLink, Styles.FLOAT_RIGHT));
+    }
+
+    String questionLink = controllers.admin.routes.AdminQuestionController.index().url();
+    String programLink = controllers.admin.routes.AdminProgramController.index().url();
+    String versionLink = routes.AdminVersionController.index().url();
+    String intermediaryLink = routes.TrustedIntermediaryManagementController.index().url();
+
+    return adminHeader
+        .with(headerLink("Questions", questionLink))
```
hmhsu, line 103: Suggestion (non-blocking): while you are at it, would you move program to the first link since it seems more natural that way in terms of user flow (also according to recent admin UX mock)

universal-application-tool-0.0.1/app/views/admin/programs/ProgramAdministratorProgramListView.java
```diff
@@ -31,7 +31,7 @@
 
   @Inject
   public ProgramAdministratorProgramListView(AdminLayout layout, Config config) {
-    this.layout = checkNotNull(layout);
+    this.layout = checkNotNull(layout).setProgramAdminType();
```
ndmckinley, line 34: This can be shown to a global admin too, actually, since the global admin is the admin of every program with no admin.

hmhsu (approved):

universal-application-tool-0.0.1/app/views/admin/AdminLayout.java
bion: Good call!

universal-application-tool-0.0.1/app/views/admin/programs/ProgramAdministratorProgramListView.java
hmhsu: Agree it will be better if we can switch based on roles.

I also wonder if this is a change across different admins, as layout is not re-created for each request.

bion: > I also wonder if this is a change across different admins, as layout is not re-created for each request.

I'm not following this part of our comment

bion: Resolved in chat, layout objects are not re-used between requests.

ndmckinley (approved):

PR #1352, opened 2021-06-12
Added Dropdown null state

### LINKED ISSUE #1244: In applicant view, dropdown questions don't have a "null" (default) state
Ideally these dropdowns would have a "null" state like "Choose an option:" before any of the real options show up. This lets the applicant see when they haven't answered the question, instead of defaulting to the first option.

In this screenshot, I would expect to see "Choose an option:" before "abc", "def", and "ghi".

image

image

WaterLily:
### Description
I added an option to the dropdown with text "Choose an option:" that will only show when nothing is selected.  Once a user has selected an option, the "choose" option will be not be visible or selectable any more.  The user is able to skip the dropdown question and submit the application if they leave the "choose" option selected.

### Checklist

### Issue(s)
Fixes #1244 

Capture

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
+++ b/universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
@@ -72,6 +72,11 @@ public class SelectWithLabel extends FieldWithLabel {
 
   @Override
   public ContainerTag getContainer() {
+    Tag placeholder = option(placeholderText).withValue("").attr(Attr.HIDDEN);
+    if (this.fieldValue.isEmpty()) {
+      placeholder.attr(Attr.SELECTED);
+    }
+    ((ContainerTag) fieldTag).with(placeholder);
     for (SimpleEntry<String, String> option : this.options) {
       Tag optionTag = option(option.getKey()).withValue(option.getValue());
       if (option.getValue().equals(this.fieldValue)) {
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
```diff
@@ -72,6 +72,11 @@ public SelectWithLabel setDisabled(boolean disabled) {
 
   @Override
   public ContainerTag getContainer() {
+    Tag placeholder = option(placeholderText).withValue("").attr(Attr.HIDDEN);
+    if (this.fieldValue.isEmpty()) {
+      placeholder.attr(Attr.SELECTED);
```
hmhsu, line 77: Question: what does it mean when a hidden element is selected?

universal-application-tool-0.0.1/test/views/questiontypes/DropdownQuestionRendererTest.java
```diff
@@ -0,0 +1,77 @@
+package views.questiontypes;
```
hmhsu, line 1: Praise: thanks for adding tests!

hmhsu (approved):
Looks really good. Nice work!

universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
WaterLily: It behaves as if it's selected, but when you open the dropdown it's hidden in the list itself.  So that combination makes it show in the form initially but keeps the user from being able to select it deliberately.

PR #1353, opened 2021-06-12
Add more columns to demographics export.

ndmckinley:
### Description
* Enable "creation time" for applications.
  * in doing this, discovered that the logic around draft applications got deleted at some point, re-enabled it.
* Enable TI group name
* obfuscated email of submitter
* stop obfuscating empty values for non-applicant columns - no PII being leaked there.
* unrelated, but also stop logging so much auth stuff.  it's not useful and it means the logs might contain PII depending on what the auth server sends back, and that might lead to a mess later on.

### Checklist


hmhsu (approved):
Looks good. Do we need some unit tests?

ndmckinley:
I think the browser tests should cover it.

Issue #1355, opened 2021-06-14
Any file can be downloaded by any authenticated user


bion:
In FileController, we're authorizing that the requester is allowed to access the applicant ID in the URL path, but never check that they are allowed to access the file key or that the file key is owned by the applicant identified in the URL. This means an attacker could login, put their own applicant ID in the URL but the file key for someone else's file key and access it.

See: https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/controllers/FileController.java#L39

Done when: logged in user's can they are authorized to see.

neel-a-mehta:
Testing steps:

1. Make a new program that has a file input
2. Log in as a guest applicant
3. Apply and upload my file
4. Link is https://staging.seattle.civiform.com/applicants/10389/files/applicant-10389%2Fprogram-4243%2Fblock-1 (copied from application Review page)
5. Open link in incognito, expect to be auth-walled -- PASS
6. Log in as new guest, get id number (id 10390)
7. Sub that number into the URL: https://staging.seattle.civiform.com/applicants/10390/files/applicant-10389%2Fprogram-4243%2Fblock-1, expect to get an error -- PASS
8. Visit the original URL, expect to get an error -- PASS
9. Apply for the program with my new account (10390)
10. Visit https://staging.seattle.civiform.com/applicants/10390/files/applicant-10390%2Fprogram-4243%2Fblock-1 (numbers substituted in fully), expect to get my original file -- PASS

I might be testing this wrong, but I'm pretty sure it works!

PR #1360, opened 2021-06-14
Restrict file access to the explicit applicant or program

hmhsu:
### Description
- Only file keys that contain the explicit applicant id or program id can be accessed through file controller.

### Checklist

### Issue(s)
#1355

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/FileController.java
+++ b/universal-application-tool-0.0.1/app/controllers/FileController.java
@@ -39,6 +39,10 @@ public class FileController extends CiviFormController {
     return checkApplicantAuthorization(profileUtils, request, applicantId)
         .thenApplyAsync(
             v -> {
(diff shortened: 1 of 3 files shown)


bion (approved):

universal-application-tool-0.0.1/app/controllers/FileController.java
```diff
@@ -39,6 +39,10 @@ public FileController(
     return checkApplicantAuthorization(profileUtils, request, applicantId)
         .thenApplyAsync(
             v -> {
+              // Ensure the file being accessed indeed belongs to the applicant.
```
bion, line 42: Suggestion (non-blocking): link to the code where this is set on create

hmhsu: Done.

PR #1362, opened 2021-06-15
Only show relevant `Operator`s in predicate builder

carolinedanzi:
### Description
1. Allow setting custom options in `SelectWithLabel`
2. Add a `LIST_OF_STRINGS` scalar type - this is only used by multi-select questions but helps when filtering operators to just those that operate on lists
2. Add the `ScalarType` as a data attribute to each `Scalar`. When a scalar is selected, show only the operators that can act on that scalar's type
3. Add two operators for dates - we compare timestamps, so in reality they are just < and >, but are clearer to use for the admin

### Checklist

### Issue(s)
#322

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
@@ -205,7 +205,7 @@ public class ProgramBlockPredicatesEditView extends BaseHtmlView {
                     option(e.getKey().toDisplayString())
                         .withValue(e.getKey().name())
                         // Add the scalar type as data so we can determine which operators to allow.
-                        .withData("type", e.getValue().name()))
+                        .withData("type", e.getValue().name().toLowerCase()))
             .collect(toImmutableList());
 
     return new SelectWithLabel()
@@ -220,15 +220,15 @@ public class ProgramBlockPredicatesEditView extends BaseHtmlView {
         Arrays.stream(Operator.values())
             .map(
                 operator -> {
-                  // Add this operator's allowed scalar types as classes, so that we can determine
+                  // Add this operator's allowed scalar types as data, so that we can determine
                   // whether to show or hide each operator based on the current type of scalar
                   // selected.
-                  return option(operator.toDisplayString())
-                      .withValue(operator.name())
-                      .withClasses(
-                          operator.getOperableTypes().stream()
-                              .map(ScalarType::name)
-                              .toArray(String[]::new));
+                  ContainerTag option =
+                      option(operator.toDisplayString()).withValue(operator.name());
+                  operator
+                      .getOperableTypes()
+                      .forEach(type -> option.withData(type.name().toLowerCase(), type.name()));
+                  return option;
                 })
             .collect(toImmutableList());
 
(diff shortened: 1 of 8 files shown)


universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -208,4 +196,46 @@ private ContainerTag renderQuestionDefinitionBox(QuestionDefinition questionDefi
...
+        .addReferenceClass(ReferenceClasses.PREDICATE_SCALAR_SELECT)
+        .getContainer();
+  }
+
+  private static ContainerTag createOperatorDropdown() {
+    ImmutableList<ContainerTag> operatorOptions =
+        Arrays.stream(Operator.values())
+            .map(
+                operator -> {
+                  // Add this operator's allowed scalar types as classes, so that we can determine
```
natsid, line 223: question: Did using the `data-` style attributes end up not working as well? The only reason I ask is because this is a non-standard use of classes - might be nice to avoid it if possible, but only if minor.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -165,19 +162,10 @@ private Tag renderPredicateForm(
...
-                .with(
-                    new SelectWithLabel()
-                        .setLabelText("Operator")
-                        // TODO(#322): Display the right operators for the given scalar type
-                        //  (requires javascript).
-                        .setOptions(operatorOptions)
-                        .getContainer())
+                .withClasses(
+                    ReferenceClasses.PREDICATE_OPTIONS, Styles.FLEX, Styles.FLEX_ROW, Styles.GAP_1)
+                .with(createScalarDropdown(scalars))
```
natsid, line 167: praise: Thanks for factoring a lot of this out into separate methods! Looks much better.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
carolinedanzi: Whoops, completely forgot about that! Thanks for calling it out - I'll try using the data attributes after lunch, you're right that this is a weird use of classes

universal-application-tool-0.0.1/app/views/components/SelectWithLabel.java
```diff
@@ -31,6 +33,15 @@ public SelectWithLabel setOptions(ImmutableMap<String, String> options) {
     return this;
   }
 
+  /**
+   * If you want more flexibility over your options (for example, if you want to add individual
```
natsid, line 37: praise: nice addition here

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: sounds good! other than that, lgtm.

natsid (approved):
Looks good aside from trying to use a `data-` attribute instead of `class`. :)

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
carolinedanzi: Fixed! Example option: `<option value="EQUAL_TO" data-long data-string>is equal to</option>`

PR #1363, opened 2021-06-15
Block visibility predicate form handling

natsid:
### Description
* Create POST endpoint for block visibility predicate form
* Add some basic validation to the form
* Display error toast when validation fails
* Display success toast when POST is successful
* Show the current visibility condition if there is one (needs work to look more human-readable)

### Screenshots
Screen Shot 2021-06-15 at 1 55 10 PM
Screen Shot 2021-06-16 at 10 11 51 AM
Screen Shot 2021-06-16 at 10 12 08 AM
Screen Shot 2021-06-16 at 10 12 25 AM

### Checklist
- x] Created tests which fail without the change (if possible)
  - Will add browser tests as a follow-up to this PR in order to more quickly unblock parallel work 

### Issue(s)
Works toward #322


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -45,4 +60,47 @@ public Result edit(Request request, long programId, long blockDefinitionId) {
           String.format("Block ID %d not found for Program %d", blockDefinitionId, programId));
     }
   }
+
+  @Secure(authorizers = Authorizers.Labels.UAT_ADMIN)
+  public Result update(Request request, long programId, long blockDefinitionId) {
+    Form<BlockVisibilityPredicateForm> predicateFormWrapper =
+        formFactory.form(BlockVisibilityPredicateForm.class).bindFromRequest(request);
+
+    if (predicateFormWrapper.hasErrors()) {
```
carolinedanzi, line 69: praise: nice error handling!

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -45,4 +60,47 @@ public Result edit(Request request, long programId, long blockDefinitionId) {
...
+      predicateFormWrapper
+          .errors()
+          .forEach(error -> errorMessageBuilder.append(String.format("\n• %s", error.message())));
+
+      return redirect(
+              routes.AdminProgramBlockPredicatesController.edit(programId, blockDefinitionId))
+          .flashing("error", errorMessageBuilder.toString());
+    } else {
+      BlockVisibilityPredicateForm predicateForm = predicateFormWrapper.get();
+      PredicateDefinition predicateDefinition =
```
carolinedanzi, line 81: suggestion (blocking): Can we add a TODO here to implement more complex predicates?

universal-application-tool-0.0.1/app/forms/BlockVisibilityPredicateForm.java
```diff
@@ -0,0 +1,99 @@
+package forms;
+
+import play.data.validation.Constraints;
+import play.data.validation.Constraints.Validatable;
+import services.applicant.question.Scalar;
+import services.program.predicate.Operator;
+
+public class BlockVisibilityPredicateForm implements Validatable<String> {
+  @Constraints.Required(message = "Must select 'hidden if' or 'shown if'.")
```
carolinedanzi, line 9: whoa! I've never seen this validation pattern, it's really cool. We should probably add this in our other forms, too

universal-application-tool-0.0.1/app/forms/BlockVisibilityPredicateForm.java
```diff
@@ -0,0 +1,99 @@
...
+    Operator operator = Operator.valueOf(getOperator());
+    Scalar scalar = Scalar.valueOf(getScalar());
+
+    // This should never happen since we only expose the usable operators for the given scalar.
+    if (!operator.getOperableTypes().contains(scalar.toScalarType())) {
+      return String.format(
+          "Cannot use operator \"%s\" on scalar \"%s\".",
+          operator.toDisplayString(), scalar.toDisplayString());
+    }
+    return null;
```
carolinedanzi, line 57: question: why return null here?

universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
```diff
@@ -18,40 +18,50 @@
...
 
   public String toDisplayString() {
     return this.displayString;
   }
 
+  public ScalarType toScalarType() {
+    return this.scalarType;
+  }
+
+  // TODO(natsid): Refactor the following (ADDRESS_SCALARS, etc) to take advantage of the fact that
```
carolinedanzi, line 62: this would be a very nice refactor!

universal-application-tool-0.0.1/app/services/program/predicate/PredicateDefinition.java
```diff
@@ -19,4 +19,7 @@ public static PredicateDefinition create(
 
   @JsonProperty("action")
   public abstract PredicateAction action();
+
+  // TODO(#322): Override toString method and/or add a different getDisplayString method to pretty
```
carolinedanzi, line 23: nit:

```suggestion
  // TODO(https://github.com/seattle-uat/civiform/issues/322): Override toString method and/or add a different getDisplayString method to pretty
```

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -197,7 +220,35 @@ private ContainerTag renderQuestionDefinitionBox(QuestionDefinition questionDefi
...
+        Arrays.stream(PredicateAction.values())
+            .collect(toImmutableMap(PredicateAction::toDisplayString, PredicateAction::name));
+    return new SelectWithLabel()
+        .setFieldName("predicateAction")
+        .setLabelText(String.format("%s should be", blockName))
+        .setOptions(actionOptions)
+        .getContainer();
+  }
+
+  private Tag createHiddenQuestionDefinitionInput(QuestionDefinition questionDefinition) {
```
carolinedanzi, line 234: suggestion: can we add a comment explaining why this is required?

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -197,7 +220,35 @@ private ContainerTag renderQuestionDefinitionBox(QuestionDefinition questionDefi
...
+        .withValue(String.valueOf(questionDefinition.getId()));
+  }
+
+  private ContainerTag createScalarDropdown(QuestionDefinition questionDefinition) {
+    ImmutableMap<Scalar, ScalarType> scalars;
+    try {
+      scalars = Scalar.getScalars(questionDefinition.getQuestionType());
+    } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
+      // This should never happen since we filter out Enumerator questions before this point.
+      return div()
```
carolinedanzi, line 248: question (blocking): This used to return a div instead of the form entirely if we hit this case. Is it ok to change the behavior so the div just appears in place of the scalar dropdown instead? 

Theoretically this should never happen, but if it did it might look a little funky

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -233,9 +285,37 @@ private static ContainerTag createOperatorDropdown() {
...
 
     return new SelectWithLabel()
+        .setFieldName("operator")
         .setLabelText("Operator")
         .setCustomOptions(operatorOptions)
         .addReferenceClass(ReferenceClasses.PREDICATE_OPERATOR_SELECT)
         .getContainer();
   }
+
+  private ContainerTag createValueField(QuestionDefinition questionDefinition) {
```
carolinedanzi, line 295: praise: thanks for separating this out!

carolinedanzi (approved):
This looks really nice!

universal-application-tool-0.0.1/app/forms/BlockVisibilityPredicateForm.java
natsid: I know... I found this a few weeks ago. Was excited to finally use it 😅  Another cool thing is that you can put something like `message = "error.operatorRequired" that references `conf/messages`! However, I wasn't sure how to parameterize it / how to pass in the args for a parameterized message here, so I just did it like this for now.

universal-application-tool-0.0.1/app/forms/BlockVisibilityPredicateForm.java
natsid: This is the pattern that means "passes validations". See docs here: https://www.playframework.com/documentation/2.8.x/JavaForms#Advanced-validation

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: Yes... good idea.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
natsid: I actually think this is preferable to only replacing the scalar dropdown because we don't want there to be a form at all in this case (don't want the submit functionality, and no need to have the other form fields there in that case). Let me know if you feel strongly another way! I feel comfortable with this especially because like you said, this should never happen.

Issue #1220, opened 2021-05-19
Record applicant's language in the downloaded CSV


neel-a-mehta:
One program admin requested this feature. When an applicant applies for a program, they have to specify a language that they'd like to fill it out in: English or some other language.

The program admin wants this chosen language to be shown in the downloaded CSV somewhere (and, I would assume, in the list of submitted applications too). Can we add a column for "chosen language for this program"?

lianne-oy:
Should be easy per bion - adding to milestone 9

hmhsu:
Reopen for story acceptance.

neel-a-mehta:
It works now!

To test, I created a program that was fully translated into Spanish and used only questions that were translated into Spanish. Then I applied as a guest with Spanish chosen as my language, then again with English. Looks like both my language choices were shown:

![image](https://user-images.githubusercontent.com/77128518/122267531-db4f8900-ce8f-11eb-86fd-389f8ce6e52f.png)

neel-a-mehta:
So everything works when translated!

PR #1330, opened 2021-06-09
Record applicant's language in CSV export

hmhsu:
### Description
Record applicant's language in CSV export

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1366, opened 2021-06-15
Application records preferred language

hmhsu:
### Description
While applicant has preferred lang field, they could change it anytime they like. This change locks in the language an applicant set while they applied for a program.

### Checklist

### Issue(s)
#1220


ndmckinley (approved):
You might aim to output this in the CSVs as well.

hmhsu:
> You might aim to output this in the CSVs as well.

I did in #1330 and then broke it in #1350. Basically I was using `application.getApplicant().getApplicantData().getPreferredLocale()` which works. When I changed to `application.getApplicantData().getPreferredLocale()`, this becomes missing all the time, hence this PR to fill it.

Issue #1307, opened 2021-06-07
Let CiviForm admins specify questions as optional or required


neel-a-mehta:
Spinoff of #1001 . We want to let CiviForm admins, when building program forms, specify individual questions as Required or Optional. Don't have any fancy logic here -- just store a simple true/false flag of whether or not the question is required.

Tasks:
1. Make all questions capable of recording a `visited but unanswered` state. Namely, the file upload question needs this.
2. Create a `bool optional` field in `ProgramQuestionDefinition` and let admins set that field on questions in blocks.
3. Update all `isComplete` methods to use the new `bool optional` field.

kendrickt:
With this feature enabled we'll have five states for each answer:
1. unvisited
2. visited but unanswered in the past (for optional questions)
3. visited and unanswered in current program (for optional questions)
4. answered in the past
5. answered in current program

How will `visited but unanswered` be treated in determining whether blocks are complete? 

Proposal: treat `visited and unanswered in previous programs` as incomplete, and `visited and unanswered in current program` as complete.

neel-a-mehta:
> Proposal: treat `visited and unanswered in previous programs` as incomplete, and `visited and unanswered in current program` as complete.

I'm good with this.

lianne-oy:
Ensuring all question types are treated consistently but will not turn on before fellowship ends b/c has potential to create bugs; better for next team to turn on.

neel-a-mehta:
#1452 built this feature but hid it behind a feature flag. A future team will want to enable the feature and check for bugs or unintended behavior. Leaving open until then.

neel-a-mehta:
#1423 does the applicant-facing UX for this.

kendrickt:
Once #1417 is done for required questions, we can move on to optional questions:

kendrickt:
For enumerator and multi-select questions, required (default) means:
* enumerator questions need at least one repeated entity
* multi select questions must have at least one selection

leonwong:
Just capturing what we discussed in standup - next steps here are to figure out what's broken/missing so we can discuss priority in triage and farm out more focused tasks to the rest of the team. From other discussions, some combination of the following might be issues:

- Client-side optional/required validation not aligned with server-side 
- No indication to the applicant which questions are required vs. optional
- Admin controls over whether questions are optional/required are missing

I believe some of this code is written (e.g., see #1550) but not merged since the e2e wasn't complete, so it might be helpful to assess where we stand both with and without this PR merged.

[...]

PR #1369, opened 2021-06-16
Add `required` field to `ProgramQuestionDefinition` 

kendrickt:
### Description
This change introduces the `optional` field and allows admins to configure it for each program question definition. The field **does not affect anything** yet.

* Add `optional` field to `ProgramQuestionDefinition`. Old program question definitions will default to required.
* Program service method added to toggle `optional` for program question definitions
* Controller method added to toggle `optional` for program question definitions. 
  * changed controller method to delete program question definitions to set `question definition id` in the route, and not in the form. This controller method doesn't need a form anymore.
* Added optional/required toggle to program block edit views for each program question definition
  * small change to this view to have an explicit `delete program question definition` button

### Screenshots

First question is required, second question is optional.

optional required question toggle 1

After toggling both questions, the first question is optional and second question is required.

optional required question toggle 2

### Checklist

### Issue(s)
Progress towards #1307


universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
```diff
@@ -18,6 +18,15 @@
   @JsonProperty("id")
   public abstract long id();
 
+  /**
+   * True if this program question definition is required. Otherwise it is optional.
+   *
+   * <p>This field was added in June. Program question definitions created before this field will
```
carolinedanzi, line 24: super nit:

```suggestion
   * <p>This field was added in June 2021. Program question definitions created before this field will
```

universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
```diff
@@ -31,12 +40,31 @@ public boolean hasQuestionDefinition() {
...
+
+  /** Return a program question definition with the {@link QuestionDefinition} set. */
+  public ProgramQuestionDefinition setQuestionDefinition(QuestionDefinition questionDefinition) {
     return new AutoValue_ProgramQuestionDefinition(
-        questionDefinition.getId(), Optional.of(questionDefinition));
+        questionDefinition.getId(), required(), Optional.of(questionDefinition));
+  }
+
+  /** Return a program question definition with a new required setting. */
+  public ProgramQuestionDefinition setRequired(boolean required) {
```
carolinedanzi, line 67: question (non-blocking): Should we use a builder here instead?

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -243,6 +243,23 @@ ProgramDefinition setBlockPredicate(
   ProgramDefinition deleteBlock(long programId, long blockDefinitionId)
       throws ProgramNotFoundException, ProgramNeedsABlockException;
 
+  /**
+   * Set a program question definition to required or optional. If the question definition ID is not
+   * present in the program's block, then nothing is changed.
```
carolinedanzi, line 248: question (non-blocking): Why not throw if the question isn't present?

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -411,6 +411,26 @@ public ProgramDefinition setBlockPredicate(
...
+  @Transactional
+  public ProgramDefinition setProgramQuestionDefinitionRequired(
+      long programId, long blockDefinitionId, long questionDefinitionId, boolean required)
+      throws ProgramNotFoundException, ProgramBlockDefinitionNotFoundException {
+    ProgramDefinition programDefinition = getProgramDefinition(programId);
+    BlockDefinition blockDefinition = programDefinition.getBlockDefinition(blockDefinitionId);
+
+    ImmutableList<ProgramQuestionDefinition> programQuestionDefinitions =
+        blockDefinition.programQuestionDefinitions().stream()
+            .map(pqd -> pqd.id() == questionDefinitionId ? pqd.setRequired(required) : pqd)
```
carolinedanzi, line 424: praise: clever use of `map` :)

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -321,63 +321,128 @@ private ContainerTag blockEditPanel(
...
+                Styles.FLEX,
+                Styles.GAP_1,
+                Styles.ITEMS_CENTER,
+                isRequired ? Styles.TEXT_BLACK : Styles.TEXT_GRAY_400,
+                Styles.FONT_MEDIUM,
+                Styles.BG_TRANSPARENT,
+                Styles.ROUNDED_FULL,
+                StyleUtils.hover(Styles.BG_GRAY_400, Styles.TEXT_GRAY_300))
+            .withType("submit")
+            .with(p("required").withClasses("hover-group:text-white"))
```
carolinedanzi, line 385: question (non-blocking): I think in the mocks this read "optional", does it matter?

carolinedanzi (approved):
This looks fantastic! One question I have is whether it's possible to test loading existing questions - do they parse correctly? Is it possible to add a unit test for that or something?

universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
kendrickt: I'd rather not. It isn't so complex that it needs a builder yet.

universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockEditView.java
kendrickt: it doesn't. we can flip it on the front end once we do more styling. 

my two cents: it seems like `required` should be what is shown, if that is the special case. If questions are default optional, then the toggle should be to turn on required.

universal-application-tool-0.0.1/app/services/program/ProgramQuestionDefinition.java
```diff
@@ -18,6 +18,15 @@
   @JsonProperty("id")
   public abstract long id();
 
+  /**
+   * True if this program question definition is required. Otherwise it is optional.
+   *
+   * <p>This field was added in June. Program question definitions created before this field will
+   * default to optional (false).
```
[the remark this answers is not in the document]
kendrickt: Just checked with Neel. Sounds like we want `required` to be default. I'll make the change!

AWBaumann (approved):

PR #1370, opened 2021-06-16
Reordering block with predicate gives warning for illegal move

carolinedanzi:
### Description
Moving a block fails and gives an error in the following cases:
  1. I try to move a block with a predicate before its predicate question(s)
  1. I try to move block A, which has predicate question(s), after block B that has the predicate which depends on block A's questions

Details
1. Adds `getQuestions` to predicate definition, so we can get all the questions used for a particular block's predicate
2. Shows an error message if a block move fails

### Checklist

### Issue(s)
#322 
Screen Shot 2021-06-16 at 3 44 11 PM

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
@@ -122,11 +122,11 @@ public class AdminProgramBlocksController extends CiviFormController {
     Direction direction = Direction.valueOf(requestData.get("direction"));
     try {
       programService.moveBlock(programId, blockId, direction);
-    } catch (ProgramNotFoundException e) {
-      return notFound(e.toString());
     } catch (IllegalBlockMoveException e) {
       return redirect(routes.AdminProgramBlocksController.edit(programId, blockId))
           .flashing("error", e.getLocalizedMessage());
+    } catch (ProgramNotFoundException e) {
+      return notFound(e.toString());
     }
     return redirect(routes.AdminProgramBlocksController.edit(programId, blockId));
   }
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
@@ -16,6 +16,7 @@ import play.mvc.Result;
 import services.CiviFormError;
 import services.ErrorAnd;
 import services.program.BlockDefinition;
+import services.program.IllegalBlockMoveException;
 import services.program.ProgramBlockDefinitionNotFoundException;
 import services.program.ProgramDefinition;
 import services.program.ProgramDefinition.Direction;
@@ -123,6 +124,9 @@ public class AdminProgramBlocksController extends CiviFormController {
       programService.moveBlock(programId, blockId, direction);
     } catch (ProgramNotFoundException e) {
       return notFound(e.toString());
+    } catch (IllegalBlockMoveException e) {
+      return redirect(routes.AdminProgramBlocksController.edit(programId, blockId))
+          .flashing("error", e.getLocalizedMessage());
     }
     return redirect(routes.AdminProgramBlocksController.edit(programId, blockId));
   }
--- a/universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
+++ b/universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
@@ -191,6 +191,12 @@ public abstract class ProgramDefinition {
         blockSlice.startsBefore(otherBlockSlice) ? otherBlockSlice : blockSlice;
     ImmutableList.Builder<BlockDefinition> blocksBuilder = ImmutableList.builder();
 
+    ImmutableSet<Long> predicateQuestions =
+        blockDefinitions().subList(blockSlice.startIndex(), blockSlice.endIndex()).stream()
+            .filter(b -> b.visibilityPredicate().isPresent())
+            .flatMap(b -> b.visibilityPredicate().get().getQuestions().stream())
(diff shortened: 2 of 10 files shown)


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlocksController.java
```diff
@@ -123,6 +124,9 @@ public Result move(Request request, long programId, long blockId) {
       programService.moveBlock(programId, blockId, direction);
     } catch (ProgramNotFoundException e) {
       return notFound(e.toString());
+    } catch (IllegalBlockMoveException e) {
```
kendrickt, line 127: (super nit) can you move this up since its an exception we expect to see often, as opposed to the other ones which should never really happen?

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -189,23 +189,38 @@ public ProgramDefinition moveBlock(long blockId, Direction direction)
         blockSlice.startsBefore(otherBlockSlice) ? blockSlice : otherBlockSlice;
     BlockSlice latterSlice =
         blockSlice.startsBefore(otherBlockSlice) ? otherBlockSlice : blockSlice;
+
+    // An illegal move is when a predicate in a later block depends on an earlier block.
+    // Get the questions used in any predicate in the latter block.
+    ImmutableSet<Long> predicateQuestions =
```
kendrickt, line 195: (non-blocking) would it be helpful if we have a method that checks for valid predicate ordering on a program e.g. `ProgramDefinition#hasValidPredicateOrdering()`, and then check after each move if it is still valid. I'm assuming that method would be helpful in a few other places. But if not, then forget about it!

If you did this, the check would have to move right before the return statement.

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -189,23 +189,38 @@ public ProgramDefinition moveBlock(long blockId, Direction direction)
...
     // Swap the two slices, assuming these slices are consecutive slices
-    blockDefinitions().stream()
-        .limit(earlierSlice.startIndex())
-        .forEach(blockDefinition -> blocksBuilder.add(blockDefinition));
+    blockDefinitions().stream().limit(earlierSlice.startIndex()).forEach(blocksBuilder::add);
     blockDefinitions().stream()
         .skip(latterSlice.startIndex())
         .limit(latterSlice.endIndex() - latterSlice.startIndex())
-        .forEach(blockDefinition -> blocksBuilder.add(blockDefinition));
+        .forEach(blocksBuilder::add);
```
kendrickt, line 218: praise: nice!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -189,23 +189,38 @@ public ProgramDefinition moveBlock(long blockId, Direction direction)
         blockSlice.startsBefore(otherBlockSlice) ? blockSlice : otherBlockSlice;
     BlockSlice latterSlice =
         blockSlice.startsBefore(otherBlockSlice) ? otherBlockSlice : blockSlice;
+
+    // An illegal move is when a predicate in a later block depends on an earlier block.
+    // Get the questions used in any predicate in the latter block.
+    ImmutableSet<Long> predicateQuestions =
+        blockDefinitions().subList(latterSlice.startIndex(), latterSlice.endIndex()).stream()
+            .filter(b -> b.visibilityPredicate().isPresent())
```
kendrickt, line 197: (super nit) you could use `Optional.map` instead of `Stream.filter`. Not actually recommending it because it might be _less_ readable...

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -189,23 +189,38 @@ public ProgramDefinition moveBlock(long blockId, Direction direction)
...
+            .flatMap(b -> b.visibilityPredicate().get().getQuestions().stream())
+            .collect(toImmutableSet());
+
+    if (!predicateQuestions.isEmpty()) {
+      if (blockDefinitions().subList(earlierSlice.startIndex(), earlierSlice.endIndex()).stream()
+          .flatMap(b -> b.programQuestionDefinitions().stream().map(ProgramQuestionDefinition::id))
+          .anyMatch(predicateQuestions::contains)) {
```
kendrickt, line 204: (super nit) this would call `Set#contains` for each id. Instead you could turn these ids into a set and check for an intersect. This is def premature optimization.

kendrickt (approved):

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: I just clicked the IntelliJ idea button, I take no credit :)

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: I like your idea! Added `hasValidPredicateOrdering` and used it in `moveBlock`

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: I always forget about `Optional.map`, thanks for the reminder! This is no longer relevant with the new method

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
carolinedanzi: Implemented the `hasValidPredicateOrdering` method, which renders this code obsolete

PR #1368, opened 2021-06-16
Load inactive program definitions properly

hmhsu:
### Description
Create `ReadOnlyVersionedQuestionServiceImpl.java` to load all questions in a particular version. This enables loading inactive programs and questions.

### Checklist
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)
Unblock #1334

PR #1373, opened 2021-06-16
Dedupe active programs and programs with draft applications

kendrickt:
### Description

If an applicant started a program but did not submit it, it would appear twice in the applicant's program index view. This is because it would appear as a program that the applicant had a draft application open for, and then again as an active program. This PR dedupes the list of programs an applicant sees, prioritizing programs that they have draft applications open for.

### UI changes:
* center the main content
* distinguish between applications the applicant can continue, or new programs they can apply for

image

### OUT OF SCOPE
This PR does not fix the bug that's being fixed by #1368. If an applicant has a draft program open, and then a new version is published with updated questions, the applicant is currently unable to access their draft application because some questions in that program are now obsolete and cannot be found in the read only question service.

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/UserRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/UserRepository.java
@@ -16,7 +16,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Provider;
 import models.Account;
@@ -58,15 +57,13 @@ public class UserRepository {
 
   /**
    * Returns all programs that are appropriate to serve to an applicant - which is any program
-   * program where they have an application in the draft stage, and any active program. Programs
-   * with an application in the draft stage and active programs are deduped, with the ones with an
-   * application in the draft stage taking priority.
+   * program where they have an application in the draft stage, and any active program.
    */
   public CompletionStage<ImmutableMap<LifecycleStage, ImmutableList<ProgramDefinition>>>
       programsForApplicant(long applicantId) {
     return supplyAsync(
         () -> {
-          List<Program> inProgressPrograms =
+          ImmutableList<ProgramDefinition> inProgressPrograms =
               ebeanServer
                   .find(Program.class)
                   .alias("p")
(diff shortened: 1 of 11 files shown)


universal-application-tool-0.0.1/app/repository/UserRepository.java
```diff
@@ -55,40 +57,54 @@ public UserRepository(
...
+                          .raw("program.id = p.id")
+                          .query())
+                  .endOr()
+                  .findList();
+          Set<String> draftProgramAdminNames =
+              inProgressPrograms.stream()
+                  .map(program -> program.getProgramDefinition().adminName())
+                  .collect(Collectors.toSet());
+
+          // Get active programs, but dedupe the ones with a draft application.
```
hmhsu, line 89: Suggestion (non-blocking): Since you are returned a map, you may delay deduping to the consumers of this function.

hmhsu (approved):
Looks good. Didn't look closely at the UI code though.

universal-application-tool-0.0.1/app/repository/UserRepository.java
kendrickt: that's a great idea!

kendrickt: did the change!

universal-application-tool-0.0.1/conf/messages
```diff
@@ -72,6 +72,12 @@ link.externalLink=External site
 # The title of the page for the list of programs.
 title.programs=Programs & services
 
+# Subtitle for the list of programs with draft applications
+titles.inProgressPrograms=Continue
+
+# Subtitle for the list of active programs
```
natsid, line 78: nit: "active programs" is kind of confusing because in the context of applications, "active" makes me think of "in progress. Suggest "Subtitle for the list of programs for which the applicant has no draft applications."

universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
```diff
@@ -110,26 +115,53 @@ private ContainerTag topContent(String titleText, String infoTextLine1, String i
...
             div()
-                .withClasses(ApplicantStyles.PROGRAM_CARDS_CONTAINER)
+                .withId("main-content")
+                .withClasses(Styles.MX_AUTO, Styles.MY_4, StyleUtils.responsiveSmall(Styles.M_10))
+                .with(
+                    h2().withText(messages.at(MessageKey.TITLE_PROGRAMS.getKeyName()))
+                        .withClasses(
+                            Styles.BLOCK, Styles.MB_4, Styles.TEXT_XL, Styles.FONT_SEMIBOLD))
+                .with(
+                    h2().withText(messages.at(MessageKey.TITLE_PROGRAMS_IN_PROGRESS.getKeyName()))
```
natsid, line 133: question (blocking): Does the applicant still see both of these `h2`s for in-progress and new programs even if there are no programs beneath them? If so, we should probably not do that and only show these subtitles when there are programs under both... Or at least only show the subtitle of the one that has any, if that makes sense? But I think it would be kind of weird on your first visit if you saw:

# Programs and services
## New programs
<program cards>

Likewise this is awkward:

#Programs and services
## In-progress programs
<no content>
## New programs
<all program cards>

natsid (approved):

natsid:
Just looked at the UI changes. I have one comment about the subtitles.

universal-application-tool-0.0.1/app/views/applicant/ProgramIndexView.java
kendrickt: good catch!!! fixing now.

PR #1382, opened 2021-06-17
Add service method and routing endpoint for removing a block visibility predicate

natsid:
### Description
This is part one of allowing admins to remove predicates. The next and final part is to add a button to the UI that calls the new `destroy` endpoint.

### Checklist

### Issue(s)
Works toward #322

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
@@ -106,4 +106,19 @@ public class AdminProgramBlockPredicatesController extends CiviFormController {
               "success", String.format("Saved visibility condition: %s", predicateDefinition));
     }
   }
+
+  @Secure(authorizers = Authorizers.Labels.UAT_ADMIN)
+  public Result destroy(long programId, long blockDefinitionId) {
+    try {
+      programService.removeBlockPredicate(programId, blockDefinitionId);
+    } catch (ProgramNotFoundException e) {
+      return notFound(String.format("Program ID %d not found.", programId));
+    } catch (ProgramBlockDefinitionNotFoundException e) {
+      return notFound(
+          String.format("Block ID %d not found for Program %d", blockDefinitionId, programId));
+    }
+
+    return redirect(routes.AdminProgramBlockPredicatesController.edit(programId, blockDefinitionId))
+        .flashing("success", "Removed the visibility condition for this block.");
+  }
 }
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -106,4 +106,19 @@ public Result update(Request request, long programId, long blockDefinitionId) {
               "success", String.format("Saved visibility condition: %s", predicateDefinition));
     }
   }
+
+  @Secure(authorizers = Authorizers.Labels.UAT_ADMIN)
+  public Result destroy(long programId, long blockDefinitionId) {
```
carolinedanzi, line 111: question (non-blocking): should we plan for removing only specific predicate nodes, instead of the entire predicate? Maybe not, I'm not sure how that would work since they don't have IDs. Maybe we can make an issue or todo instead

universal-application-tool-0.0.1/app/services/program/ProgramService.java
```diff
@@ -222,7 +222,7 @@ ProgramDefinition removeQuestionsFromBlock(
           QuestionNotFoundException;
 
   /**
-   * Set the hide {@link PredicateDefinition} for a block. This predicate describes under what
+   * Set the visibility {@link PredicateDefinition} for a block. This predicate describes under what
```
carolinedanzi, line 225: praise: thanks for keeping this up to date!

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -414,12 +414,19 @@ public ProgramDefinition setBlockPredicate(
```
carolinedanzi, line 414: suggestion: We should add the `@Nullable` annotation to indicate `predicate` can be null, like this:

```
setBlockPredicate(long programId, long blockDefinitionId, @Nullable PredicateDefinition predicate)
```

(I can't make a comment on that line in Github)

carolinedanzi (approved):
Nice work!

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
natsid: Ohh nice! Will do.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
natsid: Yep, I think let's leave this endpoint as being "totally clear it out" (especially because I think this will be useful even once we do have more complex predicates). And then we can make other methods that do smaller, more specific removals.

PR #1388, opened 2021-06-18
Store ProgramQuestionDefinition in Block

kendrickt:
### Description
`Block` has `ProgramQuestionDefinition`s instead of just `QuestionDefinition`s, so that the blocks can eventually use the `ProgramQuestionDefinition#optiona()` field.

Question definition samples need to be created with a fake question definition ID so they can be turned into `ProgramQuestionDefinition`s for `ApplicantQuestion`. Similarly, tests had to be updated so `ApplicantQuestion`s were created with `QuestionDefinition`s with IDs so they could be turned into real `ProgramQuestionDefinition`s.

### Issue(s)
Progress towards #1307


carolinedanzi (approved):
Nice change!

PR #1397, opened 2021-06-21
Predicate form: Change the input type of the "Value" field, depending on the scalar type

natsid:
### Description
This makes it so that the "compared value" that the admin enters on the right hand side will match the "scalar" or "field" type on the left hand side.

The type will change on the fly. Meaning that if the "Field" dropdown has more than one different type of scalar listed, the "Value" input type will change as you change the "Field" dropdown.

Random other change: Use single quote (`'`) instead of double quote (`"`) throughout `main.ts` to match the [Google TypeScript Style Guide](https://google.github.io/styleguide/tsguide.html). I can't find exactly where it says that, but if you look at all the code snippets, they use single quotes.

Screen Shot 2021-06-21 at 12 06 29 PM
Screen Shot 2021-06-21 at 12 06 55 PM

### Checklist

### Issue(s)
Works toward #322

### DIFF
--- a/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
+++ b/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
@@ -1,5 +1,5 @@
 /**
- * We're trying to keep the JS pretty mimimal for CiviForm, so we're only using it
+ * We're trying to keep the JS pretty minimal for CiviForm, so we're only using it
  * where it's necessary to improve the user experience.
  *
  * Appropriate uses include:
@@ -9,22 +9,22 @@
  */
 
 function attachDropdown(elementId: string) {
-  const dropdownId = elementId + "-dropdown";
+  const dropdownId = elementId + '-dropdown';
   const element = document.getElementById(elementId);
   const dropdown = document.getElementById(dropdownId);
   if (dropdown && element) {
     // Attach onclick event to element to toggle dropdown visibility.
-    element.addEventListener("click", () => toggleElementVisibility(dropdownId));
+    element.addEventListener('click', () => toggleElementVisibility(dropdownId));
 
     // Attach onblur event to page to hide dropdown if it wasn't the clicked element.
-    document.addEventListener("click", (e) => maybeHideElement(e, dropdownId, elementId));
+    document.addEventListener('click', (e) => maybeHideElement(e, dropdownId, elementId));
   }
 }
 
 function toggleElementVisibility(id: string) {
   const element = document.getElementById(id);
   if (element) {
-    element.classList.toggle("hidden");
+    element.classList.toggle('hidden');
   }
 }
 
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -266,9 +266,8 @@ private ContainerTag createActionDropdown(String blockName) {
 
   private Tag createHiddenQuestionDefinitionInput(QuestionDefinition questionDefinition) {
     return input()
-        .isHidden()
         .withName("questionId")
-        .withType("number")
+        .withType("hidden")
```
carolinedanzi, line 270: question (non-blocking): I'm not familiar with this, what does a `hidden` type mean? Is it the same as a `hidden` class?

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * We're trying to keep the JS pretty mimimal for CiviForm, so we're only using it
+ * We're trying to keep the JS pretty minimal for CiviForm, so we're only using it
```
carolinedanzi, line 2: praise: thanks for standardizing the quotes and fixing typos!

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -151,86 +151,119 @@ function removeExistingEnumeratorField(event: Event) {
...
+    option.classList.remove('hidden');
     // If this operator is not for the currently selected type, hide it.
     if (!(selectedScalarType in option.dataset)) {
-      option.classList.add("hidden");
+      option.classList.add('hidden');
     }
   });
 }
 
+function configurePredicateValueInput(
```
carolinedanzi, line 193: question (blocking): How does this work in the case where the value depends on the operator selected? For example, if I want to say `NUMBER is one of "12, 14, 16"`, the value input should be `text`, not `number`

carolinedanzi (commented):
Love the clean solution! Have one question, but otherwise looks great

universal-application-tool-0.0.1/app/assets/javascripts/main.ts
natsid: Oohh, great catch! Will implement.

natsid: Discussed offline:

We actually don't have `is one of` and `is none of` currently configured as operators for scalar type LONG. That was kind of an accident, but it also seems like we can skip adding that feature right now while we try to prioritize the things we know we need and get them done by EOD Weds.

carolinedanzi: Thanks Natalie! Added to #322 so others can work on it

carolinedanzi (approved):

PR #1395, opened 2021-06-21
Pretty print existing predicates

carolinedanzi:
### Description
1. Adds `toDisplayString` methods for predicate nodes that are human-readable
2. Uses question definitions to convert values to readable representations
3. Shows pretty-printed phrase in block predicate edit view

### Checklist
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)
#322 
Screen Shot 2021-06-21 at 2 13 48 PM
Screen Shot 2021-06-21 at 3 42 54 PM

PR #1411, opened 2021-06-21
Fix: add a default value for new AutoValue field

carolinedanzi:
### Description
1. I added a `type` field in `PredicateValue` in #1395, but existing predicates don't have this field. It broke staging
2. Adds a default type value

### Checklist

### Issue(s)
#322


carolinedanzi:
Fixed in rollback PR

Issue #1391, opened 2021-06-18
Crash when entering ".00" in number input


elisekalstad:
**Describe the bug**
Received a bug report that when entering income, if you put $5000.00 (with decimals,) you get an error, but entering $5000 without decimals was fine

**To Reproduce**
1. Start an application
2. Enter income with a decimal (i.e. $5000.000)
3. See error

**Expected behavior**
No error message when using decimanls

**Screenshots**
![Error_decimals](https://user-images.githubusercontent.com/86082659/122620526-0977c900-d048-11eb-95cb-acb9ccf9263c.jpg)

neel-a-mehta:
Reproducing / comments:

1. Creating a new program that has a number question
![image](https://user-images.githubusercontent.com/77128518/122829836-58b63780-d29c-11eb-93b4-45559c1d9c3e.png)

2. Try `500` => no problem
3. Try `500.` => no problem, saved as 500
4. Try `500.1` => form can't be submitted, error message shown saying "Please enter a valid value -- nearest are 500 and 501"
5. Try `500.0` => get error 
![image](https://user-images.githubusercontent.com/77128518/122829978-8ac79980-d29c-11eb-97c8-5ac52bafa7fd.png)
6. Try `500.00` => same thing

So we handle every floating-point value besides `.0`, `.00`, `.000`, etc. just fine. We only run into trouble with someone enters these zeroes.  My guess is that our frontend validation code sees that the inputted value is exactly equal to an integer and therefore allows it, but our backend tries parsing these things and fails because it sees that it's formatted as a float.

hmhsu:
If anyone is working on this, please let me know. Planning to take a look today.

neel-a-mehta:
Testing:

1. Create new program with a number input
2. Publish program
3. Log in as guest applicant
4. Try entering some numbers and seeing what happens

* `32`: ok
* `32.5`: anything after the decimal point is auto-removed from the input field
* `32.`: ok (saved as plain `32`)
* `-32`: negative number input not allowed (that should be OK, nobody needs to put in a negative number)
* `1,234`: comma auto-removed, which is ok
* `3e2`: the `e` auto-deletes everything in the field!
* `53-`: the `-` auto-deletes everything in the field!

This bug is fixed but I'll raise another one for the `e` and `-` wipings. That is probably due to a quirk of `parseInt()`.

PR #1413, opened 2021-06-22
Parse number into integer on client-side

hmhsu:
### Description
User can only input integer in number input fields.

### Checklist

### Issue(s)
#1391

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
+++ b/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
@@ -196,7 +196,7 @@ public class FieldWithLabel {
         fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
       }
       // We only allow integer input.
-      fieldTag.attr("oninput", "this.value=(parseInt(this.value)||0)");
+      fieldTag.attr("oninput", "n=parseInt(this.value);this.value=Number.isNaN(n)?'':n;");
     } else {
       fieldTag.withValue(this.fieldValue);
     }
--- a/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
+++ b/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
@@ -195,6 +195,8 @@ public class FieldWithLabel {
       if (this.fieldValueNumber.isPresent()) {
         fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
       }
+      // We only allow integer input.
+      fieldTag.attr("oninput", "this.value=(parseInt(this.value)||0)");
     } else {
       fieldTag.withValue(this.fieldValue);
     }
--- a/universal-application-tool-0.0.1/test/views/components/FieldWithLabelTest.java
+++ b/universal-application-tool-0.0.1/test/views/components/FieldWithLabelTest.java
@@ -38,7 +38,7 @@ public class FieldWithLabelTest {
   @Test
   public void number_setsNoValueByDefault() {
     FieldWithLabel fieldWithLabel = FieldWithLabel.number();
-    assertThat(fieldWithLabel.getContainer().render()).doesNotContain("value");
+    assertThat(fieldWithLabel.getContainer().render()).doesNotContain(" value");
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


universal-application-tool-0.0.1/test/views/components/FieldWithLabelTest.java
```diff
@@ -38,7 +38,7 @@ public void createEmail_rendersEmailInput() {
   @Test
   public void number_setsNoValueByDefault() {
     FieldWithLabel fieldWithLabel = FieldWithLabel.number();
-    assertThat(fieldWithLabel.getContainer().render()).doesNotContain("value");
+    assertThat(fieldWithLabel.getContainer().render()).doesNotContain(" value");
```
natsid, line 41: question: Why add a space here?

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -195,6 +195,8 @@ public ContainerTag getContainer() {
       if (this.fieldValueNumber.isPresent()) {
         fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
       }
+      // We only allow integer input.
+      fieldTag.attr("oninput", "this.value=(parseInt(this.value)||0)");
```
natsid, line 199: question: Does this mean that if parseInt doesn't work then we fill it with 0 instead? I'm slightly concerned with this approach if so, but I might be reading it wrong. :)

hmhsu: You are right. parseInt parses until an illegal character (e.g. non-digit) and truncates the rest of the string: so `12a34` is parsed as `12`. Here, the result is that the user will not be able to input `a` character in the first place, which seems a reasonable experience to me.

The only case where parseInt returns a `NaN` (it doesn't fail) is when there is an empty string. I think it is reasonable to set the value to 0 in this situation. The result is that once the user tries to input something, they can never set it back to empty string, but that also feels like a reasonable experience to me.

universal-application-tool-0.0.1/test/views/components/FieldWithLabelTest.java
hmhsu: The intent here is to check there is no value set for the input element and we check by a substring match. This matches with the new `this.value` in `onintput` attribute. Adding a space is a workaround for that. Agreed it is not obvious, I can add some comments to explain.

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
natsid: Ahh, the first paragraph sounds great to me. Just wanted to make sure that e.g. "50.0" would return 50 rather than 0.

As to empty string --> 0... sounds a little less ideal, especially that you can never "clear out" an answer with this. Can we special case empty string? We need the ability for number fields to be able to be optional anyway.

universal-application-tool-0.0.1/test/views/components/FieldWithLabelTest.java
natsid: Ohh like no `value="<value>"` set?

hmhsu: Yes.

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
natsid: What do you think? If it makes it really difficult to deal with then don't worry about it. But I think allowing empty inputs is probably desirable.

natsid (approved):

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
hmhsu: Yes, "50.0" would be 50.

Yeah, I also don't like losing the ability to have empty string. I think I have something working.

AWBaumann:
I don't think that this would account for negative numbers?

hmhsu:
> I don't think that this would account for negative numbers? Perhaps we just want to ignore the '.' character oninput?

Users can "step down" with the dial along side the input field and then key in the digit part of the negative numbers. Does that address the negative numbers part of your concern?

PR #1419, opened 2021-06-22
Revert "Revert "Minify prod css (#1386)" (#1392)"

ndmckinley:
This reverts commit a2d9bfe95a788daef5255a6ee5d05972fc574e6c.

### Description
This solution is only okay.  Because we have copied every class - including very many that we do not use - into `Styles.java`, and because we programmatically create utility classes, we have to keep an awful lot of CSS that we don't actually need.  Still, this brings us from 4.6 MB to 1.6 MB, which is something - and eliminates the prod/dev distinction, so we can see if we're about to screw up the styles in prod.

It also adds a tool to refresh styles without restarting `bin/run-dev`, which is a little helpful for development here.

We can get down to below a megabyte by removing all the styles from `Styles.java` that we do not use.  What we cannot do is analyze that file at minification-time to see which variables are in use.  I'll leave that choice - whether to delete all the classes we don't need - to someone else.

### Checklist

### Issue(s)

### DIFF
--- a/bin/refresh-styles
+++ b/bin/refresh-styles
@@ -1,0 +1,7 @@
+#! /bin/bash
+pushd $(git rev-parse --show-toplevel)
+
+# allocate a tty for better test output even though not strictly needed.
+docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd)/universal-application-tool-0.0.1:/usr/src/universal-application-tool-0.0.1 --entrypoint npx civiform tailwindcss build -i ./app/assets/stylesheets/styles.css -o ./public/stylesheets/tailwind.css
+
+popd
--- a/bin/refresh-styles
+++ b/bin/refresh-styles
@@ -1,7 +1,10 @@
 #! /bin/bash
 pushd $(git rev-parse --show-toplevel)
 
-# allocate a tty for better test output even though not strictly needed.
-docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd)/universal-application-tool-0.0.1:/usr/src/universal-application-tool-0.0.1 --entrypoint npx civiform tailwindcss build -i ./app/assets/stylesheets/styles.css -o ./public/stylesheets/tailwind.css
+docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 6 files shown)


bin/refresh-styles
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+pushd $(git rev-parse --show-toplevel)
+
+# allocate a tty for better test output even though not strictly needed.
+docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd)/universal-application-tool-0.0.1:/usr/src/universal-application-tool-0.0.1 --entrypoint npx civiform tailwindcss build -i ./app/assets/stylesheets/styles.css -o ./public/stylesheets/tailwind.css
```
bion, line 5: suggestion (non-blocking): break onto multiple lines for easier readability

universal-application-tool-0.0.1/app/views/style/BaseStyles.java
```diff
@@ -1,5 +1,12 @@
 package views.style;
 
+/**
+ * Constant class containing the names of styles that we have added to tailwind.
+ *
+ * <p>This file is special - strings in this file, within double quotes, are *not* stripped from the
+ * tailwind CSS during production optimization. If you add a string here, run bin/refresh-styles or
+ * restart bin/run-dev.
+ */
```
bion, line 9: 👍🏻

bin/refresh-styles
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+pushd $(git rev-parse --show-toplevel)
+
+# allocate a tty for better test output even though not strictly needed.
```
hmhsu, line 4: nit: remove `test output`

universal-application-tool-0.0.1/app/views/style/BaseStyles.java
```diff
@@ -1,5 +1,12 @@
 package views.style;
 
+/**
+ * Constant class containing the names of styles that we have added to tailwind.
```
hmhsu, line 4: praise: thanks for adding the docstring

bion (approved):

hmhsu (approved):

PR #1425, opened 2021-06-22
Properly parse `DATE` and `LONG` values in block visibility predicate form

natsid:
Couldn't find a great way to unit test this, unfortunately...

### Checklist
- [x] Manually tested the predicate flow for dates and numbers (both admin and applicant sides)

### Issue(s)
Works toward #322


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -131,4 +138,19 @@ public Result destroy(long programId, long blockDefinitionId) {
     return redirect(routes.AdminProgramBlockPredicatesController.edit(programId, blockDefinitionId))
         .flashing("success", "Removed the visibility condition for this block.");
   }
+
+  private PredicateValue parsePredicateValue(Scalar scalar, String value) {
+    switch (scalar.toScalarType()) {
```
carolinedanzi, line 143: suggestion: I think this needs to be the right hand side of the operator, since some operators have a left hand side that uses a different type from the right (ex: `is one of` has a single string on the left and a list of strings on the right)

natsid: ohh... great catch!!

natsid: I'm wondering if this is actually blocked on the discussion about SELECTION vs SELECTIONS.

natsid: Nvm - I think I get how to do this now

natsid: I _think_ it's right now. I manually tested the parts of this that are currently supported. :)

natsid: Need to add tests for `PredicateValue.java`

universal-application-tool-0.0.1/app/services/program/predicate/OperatorRightHandType.java
```diff
@@ -0,0 +1,21 @@
+package services.program.predicate;
+
+/**
+ * An enum defining the types supported on the "right hand side" of an {@link Operator} in a {@link
```
carolinedanzi, line 4: praise: love this docstring!

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
```diff
@@ -46,16 +45,21 @@ public static PredicateValue of(ImmutableList<String> value) {
             .map(PredicateValue::surroundWithQuotes)
             .collect(toImmutableList())
             .toString(),
-        ScalarType.LIST_OF_STRINGS);
+        OperatorRightHandType.LIST_OF_STRINGS);
+  }
+
+  public static PredicateValue ofListOfLongs(ImmutableList<Long> value) {
```
carolinedanzi, line 51: suggestion (non-blocking): how do you feel about naming this `listOfLongs` and the string list one `listOfStrings`?

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
```diff
@@ -46,16 +45,21 @@ public static PredicateValue of(ImmutableList<String> value) {
...
+    return create(
+        value.stream().collect(toImmutableList()).toString(), OperatorRightHandType.LIST_OF_LONGS);
```
carolinedanzi, line 53: Does this work?

```suggestion
        value.toString(), OperatorRightHandType.LIST_OF_LONGS);
```

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
```diff
@@ -46,16 +45,21 @@ public static PredicateValue of(ImmutableList<String> value) {
...
   }
 
   @JsonCreator
   private static PredicateValue create(
-      @JsonProperty("value") String value, @JsonProperty("type") ScalarType type) {
+      @JsonProperty("value") String value, @JsonProperty("type") OperatorRightHandType type) {
```
carolinedanzi, line 58: question: is this backwards compatible? I think it _might_ be, since the enum values are the same and could be parsed. If not, we'll want a custom `@JsonCreator` to handle the legacy types

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
```diff
@@ -26,18 +25,18 @@
 public abstract class PredicateValue {
 
   public static PredicateValue of(long value) {
-    return create(String.valueOf(value), ScalarType.LONG);
+    return create(String.valueOf(value), OperatorRightHandType.LONG);
```
carolinedanzi, line 28: praise: thanks for updating this to `OperatorRightHandType`!

universal-application-tool-0.0.1/app/services/program/predicate/OperatorRightHandType.java
```diff
@@ -0,0 +1,21 @@
...
+ * services.question.types.ScalarType} {@link Operator} {@link OperatorRightHandType}.
+ *
+ * <ul>
+ *   For example:
+ *   <li>STRING IS_EQUAL_TO STRING
+ *   <li>DATE IS_AFTER DATE
+ *   <li>LONG NOT_IN LIST_OF_LONGS
+ * </ul>
+ */
+public enum OperatorRightHandType {
```
carolinedanzi, line 15: question: now that we have this, does this clean up anything in the js for updating the value input type?

carolinedanzi (approved):
Nice work!

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
natsid: That sounds good to me!

universal-application-tool-0.0.1/app/services/program/predicate/OperatorRightHandType.java
natsid: That is a great question! I will check that out.

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
natsid: Ohh great question. I was slightly worried about this as well. Let me see if I can figure out its backward compatibility.

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
natsid: Oh! I guess maybe I misunderstood what needed to be donee here. Seems like this should work. :)

universal-application-tool-0.0.1/app/services/program/predicate/OperatorRightHandType.java
natsid: It's not immediately obvious to me that this would help us clean up anything in the JS.

universal-application-tool-0.0.1/app/services/program/predicate/PredicateValue.java
natsid: OK, after looking in the database and seeing that type is stored like `"type": "LIST_OF_STRINGS"`, etc, I am fairly confident this is backwards compatible for the reason you mentioned. All the existing `type`s should be able to be parsed as `OperatorRightHandType`s since that enum includes all the same names as `ScalarType` but just adds one more.

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
natsid: done

PR #1422, opened 2021-06-22
Fix operator/scalar mismatch 

carolinedanzi:
### Description
Single-select questions are stored as STRINGs, but we had `LIST_OF_STRINGS` as the type for all `selection`s in `Scalar`. This prevented us from being able to submit the predicate form for single-select questions

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)
#322

PR #1429, opened 2021-06-22
Clean up `Scalar` enum to use its new built-in scalar types

natsid:
This is an overdue cleanup from when I updated `Scalar` in a previous PR to take its `ScalarType` in the constructor.

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
@@ -4,7 +4,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -117,15 +116,14 @@ public class ApplicantQuestion {
    */
   public ImmutableMap<Path, ScalarType> getContextualizedScalars() {
     try {
-      return ImmutableMap.<Scalar, ScalarType>builder()
-          .putAll(Scalar.getScalars(getType()))
-          .putAll(Scalar.getMetadataScalars())
+      return ImmutableSet.<Scalar>builder()
+          .addAll(Scalar.getScalars(getType()))
+          .addAll(Scalar.getMetadataScalars())
           .build()
-          .entrySet()
           .stream()
           .collect(
               ImmutableMap.toImmutableMap(
-                  entry -> getContextualizedPath().join(entry.getKey()), Map.Entry::getValue));
+                  scalar -> getContextualizedPath().join(scalar), scalar -> scalar.toScalarType()));
     } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
       throw new RuntimeException(e);
     }
--- a/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/question/Scalar.java
@@ -1,6 +1,5 @@
 package services.applicant.question;
 
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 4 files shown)


carolinedanzi (commented):
This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?

carolinedanzi:
> Looks great! I wonder how this affects #1422 though.

Thanks for thinking of this! I need to re-think #1422, so I'm ok with this breaking it

kendrickt:
> This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?

Oo good catch. Would it make sense to split it up into `selection` and `selections`?

carolinedanzi:
> > This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?
> 
> Oo good catch. Would it make sense to split it up into `selection` and `selections`?

I'm testing a refactor to see how feasible this is - theoretically the worst that would happen is that applicants would need to re-answer checkbox questions, since we wouldn't see an answer for `selections`

kendrickt:
> > > This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?
> > 
> > 
> > Oo good catch. Would it make sense to split it up into `selection` and `selections`?
> 
> I'm testing a refactor to see how feasible this is - theoretically the worst that would happen is that applicants would need to re-answer checkbox questions, since we wouldn't see an answer for `selections`

I think this is a bug that should be addressed early. I think it really should be two separate scalars and scalar types.

carolinedanzi (approved):

carolinedanzi:
> > > > This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?
> > > 
> > > 
> > > Oo good catch. Would it make sense to split it up into `selection` and `selections`?
> > 
> > 
> > I'm testing a refactor to see how feasible this is - theoretically the worst that would happen is that applicants would need to re-answer checkbox questions, since we wouldn't see an answer for `selections`
> 
> I think this is a bug that should be addressed early. I think it really should be two separate scalars and scalar types.

Sounds good - approved this PR and will work on splitting `selection` into the two it always should have been

natsid:
> > > > > This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?
> > > > 
> > > > 
> > > > Oo good catch. Would it make sense to split it up into `selection` and `selections`?
> > > 
> > > 
> > > I'm testing a refactor to see how feasible this is - theoretically the worst that would happen is that applicants would need to re-answer checkbox questions, since we wouldn't see an answer for `selections`
> > 
> > 
> > I think this is a bug that should be addressed early. I think it really should be two separate scalars and scalar types.
> 
> Sounds good - approved this PR and will work on splitting `selection` into the two it always should have been

Thanks for thinking thru the details of this, both! Sorry I missed the convo over lunch. 😃  So to clarify, this is good to merge for now?

carolinedanzi:
> > > > > > This looks good, except that the type for multi vs. single select theoretically should be different. In practice, it is treated the same in `ApplicantServiceImpl`. Thoughts?
> > > > > 
> > > > > 
> > > > > Oo good catch. Would it make sense to split it up into `selection` and `selections`?
> > > > 
> > > > 
> > > > I'm testing a refactor to see how feasible this is - theoretically the worst that would happen is that applicants would need to re-answer checkbox questions, since we wouldn't see an answer for `selections`
> > > 
> > > 
> > > I think this is a bug that should be addressed early. I think it really should be two separate scalars and scalar types.
> > 
> > 
> > Sounds good - approved this PR and will work on splitting `selection` into the two it always should have been
> 
> Thanks for thinking thru the details of this, both! Sorry I missed the convo over lunch. 😃 So to clarify, this is good to merge for now?

Yes, feel free to merge!

PR #1430, opened 2021-06-22
Split `Scalar.SELECTION` into two different scalars

carolinedanzi:
### Description
1. Changes `MultiSelectQuestion` to use a scalar `selections`, and `SingleSelectQuestion` uses scalar `selection`
2. This helps us distinguish between the type of data stored - `selection` stores a single choice, whereas `selections` stores a list of choices

**Note: This will essentially  "erase" any existing answers to checkbox questions in prod**

### Checklist

### Issue(s)
#322


hmhsu (approved):
Question: is it correct that the only effect is that users lose their existing answers to questions whose scalars are changed?

hmhsu (approved):
Still LGTM

natsid (approved):
Thank you!!

carolinedanzi:
> Question: is it correct that the only effect is that users lose their existing answers to questions whose scalars are changed?

That's the only major drawback I can think of, and it only affects checkbox questions (radio and dropdown are unaffected).  Additionally, existing checkbox answers will still be in applicant JSON, but we'll ignore the old `selection` field

PR #1451, opened 2021-06-23
Stop storing blank values in applicant data.

### LINKED ISSUE #1417: All question types have the same notion of "answered" and are required

We have five states for each applicant question:
1. unanswered
1. visited but unanswered in the past
1. visited but unanswered in the current program
1. answered in the past
1. answered in the current program

Right now, "is answered" means either:
1. it was seen by the applicant and left blank
2. a real answer was provided.

We need to be clearer and more intentional with "is answered", vs "is visited", vs "is complete", with respect to required vs optional questions.

We need consistent behavior for "required" questions. And all questions should be by default required.

"is answered" should just be that a real answer was provided, and null values shouldn't be stored.

**Additional context**
`[1, 2]` should be "incomplete" from a block's perspective.

`[3, 4, 5]` should be "complete" from a block's perspective.

**Done when**
All question types behave the same:
* everything is required
* no null or empty values are stored in applicant data
* "is answered" for all questions returns true if an actual answer is provided by applicant, not if it was skipped, which shouldn't even be possible.

kendrickt:
## Description

Before, updates with empty values ("") would be stored in applicant data json. Now we don't store empty values, and updates with empty values delete the value that was stored.

Metadata is updated for every question, which we can use to mark whether a question has been visited even though it is unanswered. 

Arrays (enumerator and multi select questions) with empty updates get deleted, but the metadata is still updated.

Questions are "answered" if any of their scalars have a value in applicant data. There might be weird behavior with old applicant data that is storing blank values, but blank values shouldn't exist in the future.

Scalar specific errors for multi-field questions (address and name) for each of their fields are present if the question is answered, not just if the field is answered.

All questions are required. This includes:
* enumerator questions need at least one repeated entity
* multi select questions must have at least one selection

### Validation
Questions that are required, are not answered, and have been visited in the current program will prevent application progress the same way validation errors on questions do. 

Adds `ApplicantQuestion#isRequiredButWasUnansweredInCurrentProgram`, and `ApplicantQuestionRenderer` uses that method to show an "question is required" error message.

A similar method is added to `Block`, which is used by `ApplicantProgramBlocksController` to hold up application progression the same way `Block#hasErrors()` blocks progress.

Had to implement `empty updates -> deleting json data` to make things work well. Also had to implement `delete enumerator metadata if its the only thing at that path` for similar reasons.

## Issue(s)
Fixes #1417

### DIFF
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -236,11 +236,15 @@ public class ApplicantData {
     }
   }
 
-  /** Clears an array in preparation of updates. */
+  /**
+   * Clears an array in preparation of updates, regardless of whether there are anything values
+   * present.
+   */
   public void maybeClearArray(Path path) {
+    checkLocked();
     if (path.isArrayElement()) {
       putParentIfMissing(path);
-      putAt(path.withoutArrayReference(), new ArrayList<>());
+      jsonData.delete(path.withoutArrayReference().toString());
     }
   }
 
@@ -364,6 +368,14 @@ public class ApplicantData {
       index++;
     }
     return listBuilder.build();
+  }
+
+  /** If there are no repeated entities at the path, remove the array entirely. */
+  public void maybeClearRepeatedEntities(Path path) {
+    checkLocked();
+    if (readRepeatedEntities(path).isEmpty()) {
+      jsonData.delete(path.withoutArrayReference().toString());
+    }
   }
 
   /**
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -248,10 +248,14 @@ public class ApplicantData {
     }
   }
 
-  /** Delete whatever is there. */
-  public void delete(Path path) {
+  /** Delete whatever is there if it exists. Returns whether a delete actually happened. */
+  public boolean maybeDelete(Path path) {
     checkLocked();
-    jsonData.delete(path.toString());
+    if (hasPath(path)) {
+      jsonData.delete(path.toString());
+      return true;
+    }
+    return false;
   }
 
   private void putAt(Path path, Object value) {
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -249,7 +249,7 @@ public class ApplicantData {
   }
 
   /** Delete whatever is there. */
-  public boolean delete(Path path) {
+  public void delete(Path path) {
     checkLocked();
     jsonData.delete(path.toString());
   }
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -370,12 +370,17 @@ public class ApplicantData {
     return listBuilder.build();
   }
 
-  /** If there are no repeated entities at the path, remove the array entirely. */
-  public void maybeClearRepeatedEntities(Path path) {
+  /**
+   * If there are no repeated entities at the path, remove the array entirely. Returns true if it
+   * was deleted.
+   */
+  public boolean maybeClearRepeatedEntities(Path path) {
     checkLocked();
     if (readRepeatedEntities(path).isEmpty()) {
       jsonData.delete(path.withoutArrayReference().toString());
+      return true;
     }
+    return false;
   }
 
   /**
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -237,8 +237,8 @@ public class ApplicantData {
   }
 
   /**
-   * Clears an array in preparation of updates, regardless of whether there are anything values
-   * present.
+   * Clears an array in preparation of updates if the path is pointing to an array element,
+   * regardless of whether there are any values present.
    */
   public void maybeClearArray(Path path) {
     checkLocked();
@@ -246,6 +246,12 @@ public class ApplicantData {
       putParentIfMissing(path);
       jsonData.delete(path.withoutArrayReference().toString());
     }
+  }
+
+  /** Delete whatever is there. */
+  public boolean delete(Path path) {
+    checkLocked();
+    jsonData.delete(path.toString());
   }
 
   private void putAt(Path path, Object value) {
--- a/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
+++ b/universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
@@ -244,18 +244,16 @@ public class ApplicantData {
     checkLocked();
     if (path.isArrayElement()) {
       putParentIfMissing(path);
-      jsonData.delete(path.withoutArrayReference().toString());
+      maybeDelete(path.withoutArrayReference());
     }
   }
 
   /** Delete whatever is there if it exists. Returns whether a delete actually happened. */
-  public boolean maybeDelete(Path path) {
+  public void maybeDelete(Path path) {
     checkLocked();
     if (hasPath(path)) {
       jsonData.delete(path.toString());
-      return true;
     }
-    return false;
   }
 
   private void putAt(Path path, Object value) {
@@ -381,8 +379,8 @@ public class ApplicantData {
   }
 
   /**
-   * If there are no repeated entities at the path, remove the array entirely. Returns true if it
-   * was deleted.
+   * If there are no repeated entities at the path, remove the array entirely. Returns true if there
+   * are no repeated entities anymore.
    *
    * <p>This method needs to check that there are no repeated entity data stored before deleting
    * because we do not want to delete repeated entity data via this method. To delete data for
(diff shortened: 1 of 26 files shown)


hmhsu:
I think this change will break the optional fileupload question. Should I hide the `skip` button? Is there a configuration for deciding whether optional feature is on?

kendrickt:
(blocking) empty updates should delete path

kendrickt:
(blocking) when adding a repeated element after leaving an enumerator question unanswered, if there's an object there, delete it.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -366,6 +370,14 @@ private void putParentArray(Path path) {
     return listBuilder.build();
   }
 
+  /** If there are no repeated entities at the path, remove the array entirely. */
+  public void maybeClearRepeatedEntities(Path path) {
```
kendrickt, line 374: (blocking) return boolean true if it is deleted. false otherwise.

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -410,6 +410,13 @@ private void stageEnumeratorUpdates(
             .map(Integer::valueOf)
             .collect(ImmutableList.toImmutableList());
     applicantData.deleteRepeatedEntities(enumeratorPath, deleteIndices);
+
+    // If there are no repeated entities at this point, we still need to save metadata for this
+    // question.
+    if (applicantData.readRepeatedEntities(enumeratorPath).isEmpty()) {
+      applicantData.maybeClearRepeatedEntities(enumeratorPath);
```
kendrickt, line 417: (blocking) use the returned boolean to determine whether to write metadata or not here

kendrickt: done

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -366,6 +370,19 @@ private void putParentArray(Path path) {
     return listBuilder.build();
   }
 
+  /**
+   * If there are no repeated entities at the path, remove the array entirely. Returns true if it
+   * was deleted.
+   */
+  public boolean maybeClearRepeatedEntities(Path path) {
+    checkLocked();
+    if (readRepeatedEntities(path).isEmpty()) {
```
kendrickt, line 379: (blocking) add comment to say why we check it is empty before deleting it

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -397,6 +392,11 @@ private void stageEnumeratorUpdates(
       throw new PathNotInBlockException(block.getId(), unknownUpdates.iterator().next().path());
     }
 
+    // Before adding anything, if only metadata is being stored then delete it
```
carolinedanzi, line 395: suggestion (blocking): add something like "We cannot put an array of entities at the question path if a JSON object with metadata already exists" for clarity, since this is a weird case that isn't very obvious

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -479,25 +484,30 @@ private void stageNormalUpdates(
...
-        case LIST_OF_STRINGS:
-        case STRING:
-          applicantData.putString(currentPath, update.value());
-          break;
-        case LONG:
-          applicantData.putLong(currentPath, update.value());
-          break;
-        default:
-          throw new UnsupportedScalarTypeException(type);
+      // An empty update means the applicant doesn't want to store anything.
```
carolinedanzi, line 487:
```suggestion
      // An empty update means the applicant doesn't want to store anything. We already cleared the multi-select array above in preparation for updates, so do not remove the path.
```

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
```diff
@@ -76,6 +76,20 @@ public boolean isOptional() {
     return programQuestionDefinition.optional();
   }
 
+  /**
+   * Return true this question is required but was left unanswered while filling out the current
```
carolinedanzi, line 80:
```suggestion
   * Return true if this question is required but was left unanswered while filling out the current
```

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
```diff
@@ -76,6 +76,20 @@ public boolean isOptional() {
...
+   * program.
+   */
+  public boolean isRequiredButWasUnansweredInCurrentProgram() {
+    return !isOptional() && !errorsPresenter().isAnswered() && wasRecentlyUpdated();
+  }
+
+  /** Returns true if this question was most recently updated in this program. */
+  private boolean wasRecentlyUpdated() {
```
carolinedanzi, line 88: super nit (non-blocking): Could we call this `wasUpdatedInThisProgram`, since that's really what it's checking? We're not checking timestamps here so "recently" is a bit confusing

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
```diff
@@ -139,28 +153,32 @@ public boolean hasErrors() {
...
 
+  private Optional<Long> getMetadata(Scalar metadataScalar) {
+    Path contextualizedMetadataPath = getMetadataPath(metadataScalar);
     return applicantData.readLong(contextualizedMetadataPath);
   }
 
-  public Optional<Long> getLastUpdatedTimeMetadata() {
-    Path contextualizedMetadataPath = getContextualizedPath().join(Scalar.UPDATED_AT);
+  private Path getMetadataPath(Scalar metadataScalar) {
+    Path contextualizedMetadataPath = getContextualizedPath().join(metadataScalar);
```
carolinedanzi, line 169: super nit (non-blocking): could this just be an if/else? We don't use the `contextualizedMetadataPath` in the if statement

universal-application-tool-0.0.1/app/services/applicant/question/NameQuestion.java
```diff
@@ -53,7 +53,7 @@ public boolean hasTypeSpecificErrors() {
   }
 
   public ImmutableSet<ValidationErrorMessage> getFirstNameErrors() {
-    if (isFirstNameAnswered() && getFirstNameValue().isEmpty()) {
```
carolinedanzi, line 56: question (blocking): can we remove the `isFirstNameAnswered()` etc. methods from name and address questions now? Or are they used elsewhere?

universal-application-tool-0.0.1/test/support/ProgramBuilder.java
```diff
@@ -202,13 +202,22 @@ public BlockBuilder withPredicate(PredicateDefinition predicate) {
...
 
+    /** Add a required question to the block. */
     public BlockBuilder withQuestion(Question question) {
       blockDefBuilder.addQuestion(
           ProgramQuestionDefinition.create(
               question.getQuestionDefinition(), Optional.of(programBuilder.programDefinitionId)));
       return this;
     }
 
+    public BlockBuilder withQuestion(Question question, boolean optional) {
```
carolinedanzi, line 213: suggestion: might be clearer to have `withRequiredQuestion` and `withOptionalQuestion` methods - seems a little more readable

carolinedanzi (approved):
Really nice work! Thank you for adding tests, it helps me feel more confident this change won't break everything

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -236,14 +236,28 @@ private void putArrayIfMissing(Path path) {
...
+  /**
+   * Clears an array in preparation of updates if the path is pointing to an array element,
+   * regardless of whether there are any values present.
+   */
   public void maybeClearArray(Path path) {
+    checkLocked();
     if (path.isArrayElement()) {
       putParentIfMissing(path);
-      putAt(path.withoutArrayReference(), new ArrayList<>());
+      jsonData.delete(path.withoutArrayReference().toString());
```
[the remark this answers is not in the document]
kendrickt: nice call. fixed!

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
kendrickt: fixed!

universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java
kendrickt: great suggestion!

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
kendrickt: fixed

universal-application-tool-0.0.1/app/services/applicant/question/ApplicantQuestion.java
kendrickt: duh. fixed!

universal-application-tool-0.0.1/app/services/applicant/question/NameQuestion.java
kendrickt: Used in `isAnswered` methods

universal-application-tool-0.0.1/test/support/ProgramBuilder.java
kendrickt: I'll do a follow up PR for this

PR #1459, opened 2021-06-23
Add program id to program question definition

kendrickt:
Larger than I had hoped... but this gets a `program id` reference into a `ProgramQuestionDefinition`. The main change is `ProgramQuestionDefinition#setQuestionDefinition` turned into `ProgramQuestionDefinition#loadCompletely`.

Most other changes change how objects are constructed in tests.

### Issue(s)
Progress towards #1417


natsid (approved):
so much bigger than I'd have expected, but makes sense b/c of the tests.

PR #1479, opened 2021-06-24
Change review page so that applicants can only navigate to the first unanswered question.

AWBaumann:
### Description
Edit/Continue now works the same way in preview and review. Users can only navigate to previously answered questions and the first unanswered question.


browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -152,7 +152,7 @@ describe('End to end enumerator test', () => {
     expect(await page.innerText("#application-summary")).not.toContain("12");
 
     // Go back and add an enumerator answer.
-    await page.click('.cf-applicant-summary-row:has(div:has-text("Household members")) a:has-text("Edit")');
+    await page.click('.cf-applicant-summary-row:has(div:has-text("Household members")) a:has-text("Continue")');
```
hmhsu, line 155: Thoughts (non-blocking): shouldn't this be answered so should be `edit`?

hmhsu (approved):

PR #1486, opened 2021-06-24
Remove admin access for global admins to programs which don't have admins

### LINKED ISSUE #1461: CiviForm Admins should not be able to access program applications
Currently, CiviForm Admins can access the program applications for programs that have no program admin. This is not the behavior desired by the city. CiviForm Admins should not able be able to access program applications under any circumstances.

ndmckinley:
### Description
Admins should in fact not have access to programs, even if there are no program admins.

### Checklist

### Issue(s)
Fixes #1461.


hmhsu (approved):
Looks good. Do we need to change the logic whom to notify on application submission?

ndmckinley:
Ideally we should not - someone should still get notified, and global admins are the right people to do something to add new admins, so probably it's right as-is.

PR #1488, opened 2021-06-24
Allow a list of values for multi-option question predicates

natsid:
### Description
Now predicates based on multi-options questions can actually have lists as the `values` field on the right hand side! Works for both single select (radios, dropdowns) and multi select questions.

* Use a set of checkbox inputs for the `value` field when the left hand side of the predicate operator is a selection or selections from a multi-option question.
* Parse these checkbox values differently than if we were getting one string value.
* Removed  "is equal to" and "is not equal to" operator options from single-select predicates. This is because then we'd need to limit the number of checkboxes selected in that case. We could have added extra validation to enforce selecting only a single option in that case. But, since "is one of" with just one value is equivalent to "is equal to" and likewise for "is none of" and "is not equal to", we think this strategy actually _helps_ the user by not getting in their way. If they select multiple options on the right side, they probably mean "is one of" anyway.  A validation error when a user uses "is equal to" and multiple options on the right hand side seems not very helpful.
* Note that `string` scalars can already have "lists" on the right-hand side prior to this PR, but the value is just entered as a comma-separated list in that case.

Screen Shot 2021-06-24 at 2 01 35 PM

### Checklist
- [x] Manually tested this flow with both single select and multi select question predicates

### Issue(s)
Works toward #322

### DIFF
--- a/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
+++ b/universal-application-tool-0.0.1/app/assets/javascripts/main.ts
@@ -165,14 +165,17 @@ function configurePredicateForm(event: Event) {
   const selectedScalarType = scalarDropdown.options[scalarDropdown.options.selectedIndex].dataset.type;
   const selectedScalarValue = scalarDropdown.options[scalarDropdown.options.selectedIndex].value;
 
-  filterOperators(scalarDropdown, selectedScalarType);
+  filterOperators(scalarDropdown, selectedScalarType, selectedScalarValue);
   configurePredicateValueInput(scalarDropdown, selectedScalarType, selectedScalarValue);
 }
 
 /**
  * Filter the operators available for each scalar type based on the current scalar selected.
  */
-function filterOperators(scalarDropdown: HTMLSelectElement, selectedScalarType: string) {
+function filterOperators(
+  scalarDropdown: HTMLSelectElement,
+  selectedScalarType: string,
+  selectedScalarValue: string) {
   // Filter the operators available for the given selected scalar type.
   const operatorDropdown =
     scalarDropdown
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/assets/javascripts/main.ts
```diff
@@ -165,31 +165,55 @@ function configurePredicateForm(event: Event) {
...
 
+function shouldHideOperator(
+  selectedScalarType: string,
+  selectedScalarValue: string,
+  operatorOption: HTMLOptionElement
+): boolean {
+  // If this operator is not for the currently selected type, hide it.
+  return (
+    !(selectedScalarType in operatorOption.dataset) ||
+    // Special case for SELECTION scalars (which are of type STRING):
```
carolinedanzi, line 204: praise: thank you for the detailed docstring!

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -163,13 +178,11 @@ private PredicateValue parsePredicateValue(Scalar scalar, Operator operator, Str
             return PredicateValue.of(Long.parseLong(value));
         }
 
-      default: // STRING, LIST_OF_STRINGS
+      default: // STRING, should not include LIST_OF_STRINGS since that should be covered at the top
+        // of the method with `values`.
         switch (operator) {
           case ANY_OF:
           case NONE_OF:
-          case IN:
```
carolinedanzi, line 170: question: I'm not sure I understand why these operators disappear. I thought all of these have a list of strings on the right?

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramBlockPredicatesControllerTest.java
```diff
@@ -127,6 +128,69 @@ public void update_withValidFormData_savesNewPredicate() {
         .doesNotContain("This block is always shown.");
   }
 
+  @Test
+  public void update_withSingleSelectQuestion() {
```
carolinedanzi, line 132: praise: thanks for adding tests!

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramBlockPredicatesControllerTest.java
```diff
@@ -127,6 +128,69 @@ public void update_withValidFormData_savesNewPredicate() {
...
+            .build();
+
+    Result result = controller.update(request, programWithThreeBlocks.id, 3L);
+
+    assertThat(result.status()).isEqualTo(SEE_OTHER);
+    assertThat(result.redirectLocation())
+        .hasValue(
+            routes.AdminProgramBlockPredicatesController.edit(programWithThreeBlocks.id, 3L).url());
+    assertThat(result.flash().get("error")).isEmpty();
+    assertThat(result.flash().get("success").get()).contains("Saved visibility condition blah");
```
carolinedanzi, line 156: question: blah?

carolinedanzi (approved):
Thanks for doing this! I think it makes sense - so many edge cases!

universal-application-tool-0.0.1/test/controllers/admin/AdminProgramBlockPredicatesControllerTest.java
natsid: LOL oh no. Thanks for the catch

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
natsid: They do all have list of strings on the right, that is correct! Let me see if I can explain this effectively in writing.

Once we get to this point, we know the scalar on the left is of type STRING, and that it is *not* of value SELECTION. That's because I added that `if` statement at the top that parses the `values` list rather than the `value` string. I updated the logic in that if statement so it's hopefully a little more clear what's going on.

Additionally, any `STRING` scalar type only has `ANY_OF`, `NONE_ON`, `EQUAL_TO`, `NOT_EQUAL_TO` as operator options. The other ones (`IN`, etc) are not available unless there is a `LIST_OF_STRINGS` on the left hand side.

The outer switch statement is based on the scalar on the *left hand side*. I know this is kind of collaged together and awkward, but hopefully it works. :)

PR #1491, opened 2021-06-24
Add delete file button and continue button

hmhsu:
### Description
Add a delete button if an uploaded file is present, also a continue button allowing user not to re-upload again.

#### Optional FileUpload Question
No file uploaded yet
Screen Shot 2021-06-24 at 4 37 08 PM
An uploaded file present
Screen Shot 2021-06-24 at 4 36 59 PM
#### Required FileUpload Question
No file uploaded yet
Screen Shot 2021-06-25 at 11 20 21 AM
An uploaded file present
Screen Shot 2021-06-25 at 11 20 40 AM

### Checklist


universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -199,34 +254,67 @@ private Tag renderReviewButton(Params params) {
...
   }
 
+  private boolean hasUploadedFile(Params params) {
+    return params.block().getQuestions().stream()
+        .map(ApplicantQuestion::createFileUploadQuestion)
+        .map(FileUploadQuestion::getFileKeyValue)
+        .anyMatch(maybeValue -> maybeValue.isPresent());
+  }
+
+  private boolean hasRequiredQuestion(Params params) {
```
kendrickt, line 314: (non-blocking) I get it, but this function is confusing because its really only useful for file upload screens. Maybe if it was renamed "hasAtLeastOneRequriedQuestion" it would be a bit clearer when its used. I would probably opt for not making this a function if it is only used in one place.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -199,34 +254,67 @@ private Tag renderReviewButton(Params params) {
...
+            .attr(FORM, FILEUPLOAD_CONTINUE_FORM_ID)
+            .withClasses(ApplicantStyles.BUTTON_BLOCK_NEXT)
+            .withId("fileupload-continue-button");
+    return Optional.of(button);
+  }
+
+  private Optional<Tag> renderDeleteButton(Params params) {
+    String buttonText;
+    String buttonId;
+    if (hasUploadedFile(params)) {
```
kendrickt, line 278: (blocking) I think if a question is required, we don't want to provide a delete button. So there could be an early return if the question is required to return `Optional.empty()`.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -199,34 +254,67 @@ private Tag renderReviewButton(Params params) {
...
+    }
+    Tag button =
+        submitButton(params.messages().at(MessageKey.BUTTON_KEEP_FILE.getKeyName()))
+            .attr(FORM, FILEUPLOAD_CONTINUE_FORM_ID)
+            .withClasses(ApplicantStyles.BUTTON_BLOCK_NEXT)
+            .withId("fileupload-continue-button");
+    return Optional.of(button);
+  }
+
+  private Optional<Tag> renderDeleteButton(Params params) {
```
kendrickt, line 275: (blocking) This method returns a delete button or a skip button... a comment with a more accurate name please!

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -150,22 +156,66 @@ private Tag renderFileUploadBlockSubmitForm(Params params) {
...
+            .with(
+                each(
+                    params.block().getQuestions(),
+                    question -> renderQuestion(question, rendererParams)));
+    Tag skipForms = renderSkipFileUploadForms(params);
+    Tag buttons = renderFileUploadBottomNavButtons(params);
+    return div(uploadForm, skipForms, buttons);
+  }
+
+  private Tag renderSkipFileUploadForms(Params params) {
```
kendrickt, line 174: (blocking) this renders skip and delete forms. please update name, and maybe add a comment if you think that's helpful.

kendrickt (changes requested):
A couple changes.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
hmhsu: Done. PTAL

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
hmhsu: Done. PTAL

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
hmhsu: Good call. Done.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
hmhsu: Renamed. I am a big fan of using short functions as a way to abstract lower-level details so the code is more concise and readable at a higher-level of abstraction, even when the functions are only used in one place. The added benefit is that it is usually very easy to look at a short function and visually verify if it is functionally correct because it should be doing one and only one simple thing.

kendrickt (approved):

Issue #1456, opened 2021-06-23
Give all question types non-empty `label`s for screen readers' benefit


neel-a-mehta:
Screen readers need to see HTML `<label>`s for all `<input>`s so they can read them out. At present, all our question types have `<label>`s, but (as far as I can tell) they are empty.

![image](https://user-images.githubusercontent.com/77128518/123153448-8628ef80-d41a-11eb-8ad5-a7451261afaa.png)

We should [follow this guide](https://www.w3.org/WAI/tutorials/forms/labels/) for putting non-empty, useful labels next to each input. You can use [Tailwind's `sr-only` class](https://tailwindcss.com/docs/screen-readers) to hide the label from non-screen-reading users.

Consider putting the question's title and/or description in the `<label>` next to the `<input>`.

This applies to all question types that use `<input>` tags, which I think is everything.

## Original tester comments

> Actual results:
At step 6 above, observe the 'date question text updated new version updated' dropdown list is unlabeled
> Expected results:
All controls should hae a proper label associated and screenreader should be able to annouce the label , state, controltype & value as seen applicable

neel-a-mehta:
Just tested out "Program for Question Lifecycle" with a screen reader. Looks like it can handle every question type properly, giving a reasonable voiceover description thanks to our `<label>`s. Great!

PR #1492, opened 2021-06-25
Add labels to every field so that screen readers can interact with them.

ndmckinley:
### Description
Multi-part questions are rendered with their labels adjacent and content-ful - but for the single-part questions where we don't, this PR adds the content.  It also associates all the labels with the inputs.  for the ones where the associated label would be duplicative, apply `sr-only` class.

Tested by adding an assert - browser tests all pass - removed the assert because a ton of unit tests would have needed to be updated to no real benefit.

### Checklist

### Issue(s)
Addresses #1456.

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
@@ -210,6 +210,7 @@ public final class QuestionEditView extends BaseHtmlView {
           FieldWithLabel.number()
               .setFieldName("nextAvailableId")
               .setValue(((MultiOptionQuestionForm) questionForm).getNextAvailableId())
+              .setScreenReaderText("next id, hidden")
               .getContainer()
               .withClasses(Styles.HIDDEN));
     }
--- a/universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
@@ -210,7 +210,6 @@ public final class QuestionEditView extends BaseHtmlView {
           FieldWithLabel.number()
               .setFieldName("nextAvailableId")
               .setValue(((MultiOptionQuestionForm) questionForm).getNextAvailableId())
-              .setScreenReaderText("next id, hidden")
               .getContainer()
               .withClasses(Styles.HIDDEN));
     }
--- a/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
+++ b/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
@@ -6,6 +6,7 @@ import static j2html.TagCreator.each;
 import static j2html.TagCreator.label;
 import static j2html.TagCreator.textarea;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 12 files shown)


universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
```diff
@@ -210,6 +210,7 @@ private ContainerTag multiOptionQuestionField(QuestionForm questionForm) {
           FieldWithLabel.number()
               .setFieldName("nextAvailableId")
               .setValue(((MultiOptionQuestionForm) questionForm).getNextAvailableId())
+              .setScreenReaderText("next id, hidden")
```
kendrickt, line 213: (question): should hidden fields have aria lables?

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -218,9 +230,12 @@ public ContainerTag getContainer() {
...
     ContainerTag labelTag =
         label()
-            .condAttr(!this.id.isEmpty(), Attr.FOR, this.id)
-            .withClasses(labelText.isEmpty() ? "" : BaseStyles.INPUT_LABEL)
-            .withText(labelText);
+            .attr(Attr.FOR, this.id)
+            .withCondClass(!labelText.isEmpty(), BaseStyles.INPUT_LABEL)
+            // If the text is screen-reader text, then we want the label to be screen-reader
+            // only.
+            .withCondClass(labelText.isEmpty(), Styles.SR_ONLY)
```
kendrickt, line 237: (blocking) it is clearer to do `withClass(labelText.isEmpty() ? Styles.SR_ONLY : BaseStyles.INPUT_LABEL)`.

universal-application-tool-0.0.1/app/views/questiontypes/DropdownQuestionRenderer.java
```diff
@@ -40,6 +40,7 @@ public Tag render(ApplicantQuestionRendererParams params) {
                         toImmutableMap(
                             LocalizedQuestionOption::optionText,
                             option -> String.valueOf(option.id()))));
+    select.setScreenReaderText(question.getQuestionText());
```
kendrickt, line 43: (question) is setting the screen reader text to the question text for a field the right thing to do?

kendrickt (changes requested):
Looks like name and address question are missing screen reader labels.

universal-application-tool-0.0.1/app/views/questiontypes/DropdownQuestionRenderer.java
kendrickt: > Looks like name and address question are missing screen reader labels.

Ahh I see. The name and address question fields use labels, so we don't need the screen reader label for them. It would be helpful to comment that in at least the PR description that the multi-field questions have labels attached to the fields, and the single-field questions don't.

kendrickt:
> Looks like name and address question are missing screen reader labels.

> Looks like name and address question are missing screen reader labels.

Ahh I see. The name and address question fields use labels, so we don't need the screen reader label for them. It would be helpful to comment that in at least the PR description that the multi-field questions have labels attached to the fields, and the single-field questions don't.

kendrickt (approved):

universal-application-tool-0.0.1/app/views/admin/questions/QuestionEditView.java
ndmckinley: eh, probably not valuable.  I added it so that the asserts I was testing with would pass.  Removed.

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
ndmckinley: Sure.

universal-application-tool-0.0.1/app/views/questiontypes/DropdownQuestionRenderer.java
ndmckinley: I _think_ so, yeah.

PR #1503, opened 2021-06-25
rename block to screen for user-facing strings

kendrickt:
Renames "block" to "screen" for ALL user-facing strings. That is:
* strings in admin views
* default name is "screen 1" for example
* error messages

Not refactoring any class or method names in this PR. That's overkill for now.


universal-application-tool-0.0.1/app/services/applicant/Block.java
```diff
@@ -219,6 +219,6 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return "Block [id: " + this.id + "]";
+    return "Screen [id: " + this.id + "]";
```
hmhsu, line 222: Suggestion (blocking): This could potentially be confusing to a developer. Could you add a comment in this class saying `Screen` is the user-facing equivalent of `Block`?

universal-application-tool-0.0.1/app/repository/VersionRepository.java
```diff
@@ -194,7 +194,7 @@ public void updateQuestionVersions(Program draftProgram) {
     ProgramDefinition.Builder updatedDefinition =
         draftProgram.getProgramDefinition().toBuilder().setBlockDefinitions(ImmutableList.of());
     for (BlockDefinition block : draftProgram.getProgramDefinition().blockDefinitions()) {
-      LOG.trace("Updating block {}.", block.id());
+      LOG.trace("Updating screen {}.", block.id());
```
hmhsu, line 197: Nit (non-blocking): technically log is not user-facing, I wonder whether block or screen is clearer for a person reading the log. What do you think about `screen (block)`?

hmhsu (commented):
Still have way through, I think it might be more helpful to keep exception messages and logs to use block since they are not technically user facing.

universal-application-tool-0.0.1/app/repository/VersionRepository.java
kendrickt: 👍

universal-application-tool-0.0.1/app/services/applicant/Block.java
kendrickt: 👍

hmhsu (approved):

hmhsu:
Please consider updating the message files if there are any references to block.

hmhsu (approved):

PR #1514, opened 2021-06-25
Enable confirmation for link elements, and add confirmation to the bu…

kendrickt:
…tton that publishes all programs.

### Description
The "publish all programs" button has too many consequences that it should have a confirmation dialog to protect it. This PR adds a confirmation dialog, and updates the browser tests to click OK on that dialog.

image

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/LinkElement.java
+++ b/universal-application-tool-0.0.1/app/views/components/LinkElement.java
@@ -53,6 +53,7 @@ public class LinkElement {
   private String text = "";
   private String href = "";
   private String styles = "";
+  private String confirmationMessage = "";
 
   public LinkElement setId(String id) {
     this.id = id;
@@ -71,6 +72,16 @@ public class LinkElement {
 
   public LinkElement setStyles(String... styles) {
     this.styles = StyleUtils.joinStyles(styles);
+    return this;
+  }
+
+  public LinkElement enableConfirmation(String confirmationMessage) {
+    this.confirmationMessage = confirmationMessage;
+    return this;
+  }
+
+  public LinkElement disableConfirmation() {
+    this.confirmationMessage = "";
     return this;
   }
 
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/views/components/LinkElement.java
```diff
@@ -128,11 +143,22 @@ public ContainerTag asHiddenFormLink(Http.Request request) {
...
             .withCondId(!Strings.isNullOrEmpty(id), id);
     return form;
   }
+
+  private String confirmationScript(String message) {
+    return String.format(
+        "if(confirm('%s')){ return true; } else { var e = arguments[0] || window.event;"
+            + " e.stopImmediatePropagation(); return false; }",
+        message);
+  }
```
AWBaumann, line 163: I left a similar comment on pr #1517 but I'll say the same here...

I don't love that we're writing JS in Java. It's not compile-time checked and it looks really suspect. I've been following the pattern of attaching listeners via typescript when we load the relevant classes.

I understand that for this and the other change you're trying to avoid having to ship down the translated message, but you still are. It's just currently wrapped in javascript.

My suggestion would be to send down a hidden div with an id for the specific message, then attach a `data-confirmation-text` attribute to the button.

**Option 1**
```
<div id="confirmation-text-id" class="hidden">Show confirmation dialog with this text</div>
...
<div id="some-confirmation-button" data-confirmation-id="confirmation-text-id">Click here for confirmation>
```

**Option 2**
```
<div id="some-confirmation-button" class="some-button" data-confirmation-text="Show confirmation dialog with this text">Click here for confirmation>
```

**Typescript**
```
onLoad() {
    const buttons = Array.from(document.querySelectorAll('.some-button'));
    buttons.forEach(button => {
        const buttonText = "";
        button.addEventListener('click', showConfirmation(buttonText));
    });
}

showConfirmation(text: String) : boolean {
    if (confirm(text)) {
        return true;
    } else { 
      return false;
    }
}
```

kendrickt: I'm just going to use a modal since its built for this, kind of.

PR #1517, opened 2021-06-28
Add confirmation dialog to enumerator deletion.

### LINKED ISSUE #1117: Add a dialog for confirming deletion of enumerator entities

From @kendrickt on #1110:

> Problem: Applicant cannot un-delete an entity.
> Solution: add a warning to verify with applicant that they want to delete it. This can include a message about needing to click save and continue to actually persist these changes.

**Done when**
When an applicant clicks the remove entity button, a dialog pops up to warn them that they are about to delete the entity

WaterLily:
### Description
I added a browser confirmation that shows when a user deletes an enumerator entity.  If the user clicks "cancel", the click event is not propagated to the normal entity-deleting listener.

The styling of the dialog isn't very pretty, because it's the default browser confirm dialog that we don't have control over.
enumerator_confirmation

Tested locally.

### Checklist

### Issue(s)
Fixes #1117

### DIFF
--- a/universal-application-tool-0.0.1/app/views/questiontypes/EnumeratorQuestionRenderer.java
+++ b/universal-application-tool-0.0.1/app/views/questiontypes/EnumeratorQuestionRenderer.java
@@ -118,10 +118,18 @@ public class EnumeratorQuestionRenderer extends ApplicantQuestionRenderer {
                     localizedEntityType))
             .addReferenceClass(ReferenceClasses.ENTITY_NAME_INPUT)
             .getContainer();
+    String confirmationMessage =
+        messages.at(MessageKey.ENUMERATOR_DIALOG_CONFIRM_DELETE.getKeyName(), localizedEntityType);
     Tag removeEntityButton =
         TagCreator.button()
             .withType("button")
             .withCondId(existingIndex.isPresent(), existingIndex.map(String::valueOf).orElse(""))
+            .attr(
+                "onclick",
+                String.format(
+                    "if(confirm('%s')){ return true; } else { var e = arguments[0] ||"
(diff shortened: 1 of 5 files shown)


universal-application-tool-0.0.1/app/views/questiontypes/EnumeratorQuestionRenderer.java
```diff
@@ -118,10 +118,18 @@ private static Tag enumeratorField(
...
     Tag removeEntityButton =
         TagCreator.button()
             .withType("button")
             .withCondId(existingIndex.isPresent(), existingIndex.map(String::valueOf).orElse(""))
+            .attr(
+                "onclick",
+                String.format(
+                    "if(confirm('%s')){ return true; } else { var e = arguments[0] ||"
+                        + " window.event; e.stopImmediatePropagation(); return false; }",
+                    confirmationMessage))
```
AWBaumann, line 132: Nit. I've been trying to avoid writing JS within the Java layer, but the only other way I can think of doing it would be to add a hidden div just to ship the translation down to the client...

AWBaumann (approved):
Approved, but I'd like to see what others think about the JS in Java section.

universal-application-tool-0.0.1/app/views/questiontypes/EnumeratorQuestionRenderer.java
natsid: This is not the most helpful second opinion (re your top-level comment), but I don't feel strongly one way or the other. I would say our JS in general scares me a bit, but having this one line of JS here doesn't bother me any extra.

PR #1518, opened 2021-06-28
Predicate browser test for all scalar types and right-hand types

carolinedanzi:
### Description
1. Adds a test that checks every type of right-hand value for the predicate expression, and every type of scalar.
2. Fixes a bug where IN and NOT_IN operator values were not parsed correctly

### Checklist

### Issue(s)
#322

### DIFF
--- a/browser-test/src/admin_predicates.test.ts
+++ b/browser-test/src/admin_predicates.test.ts
@@ -93,5 +93,4 @@ describe('create and edit predicates', () => {
     // We should now be on the summary page
     await applicant.submitFromReviewPage(programName);
     await endSession(browser);
-  });
 })
--- a/browser-test/src/admin_predicates.test.ts
+++ b/browser-test/src/admin_predicates.test.ts
@@ -1,0 +1,26 @@
+import { AdminPredicates, AdminPrograms, AdminQuestions, endSession, loginAsAdmin, logout, startSession } from './support'
+
+describe('create and edit predicates', () => {
+  it('add a predicate', async () => {
+    const { browser, page } = await startSession();
+
+    await loginAsAdmin(page);
+    const adminQuestions = new AdminQuestions(page);
+    const adminPrograms = new AdminPrograms(page);
+
+    // Add a program with two blocks
+    await adminQuestions.addTextQuestion('predicate-q');
+    await adminQuestions.addTextQuestion('other-q');
+
+    const programName = 'create predicate';
+    await adminPrograms.addProgram(programName);
+    await adminPrograms.editProgramBlock(programName, 'first block', ['predicate-q']);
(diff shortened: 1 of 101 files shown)


universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
```diff
@@ -183,7 +183,9 @@ private PredicateValue parsePredicateValue(
         // of the method with `values`.
         switch (operator) {
           case ANY_OF:
+          case IN:
```
natsid, line 186: suggestion (blocking): I think we should either add *all* the right-hand-side list operators here, or else only keep `IN` and `NOT_IN`. The reason I had only kept `ANY_OF` and `NONE_OF` here in the first place instead of all of them was because I was only trying to include right-hand-side list operators that were relevant to left-hand-side string operators, if that makes sense. Unfortunately I clearly chose the wrong two operators. But anyway, it might be safer in some ways to include all of the list operators. Even though in theory, by this point in the code, `IN` and `NOT_IN` should be the only ones of concern.

PS THANK YOU FOR FIXING THIS

browser-test/src/admin_predicates.test.ts
```diff
@@ -125,5 +125,78 @@ describe('create and edit predicates', () => {
...
+    await adminPrograms.goToEditBlockPredicatePage(programName, 'Screen 5');
+    await adminPredicates.addPredicate('date', 'shown if', 'date', 'is earlier than', '2021-01-01');
+
+    // Lists of strings on both sides (multi-option question checkbox)
+    await adminPrograms.goToEditBlockPredicatePage(programName, 'Screen 6');
+    await adminPredicates.addPredicate('both sides are lists', 'shown if', 'selections', 'contains any of', 'dog,cat');
+
+    await adminPrograms.publishProgram(programName);
+
+    // Switch to applicant view - if they answer each question according to the predicate,
```
natsid, line 177: praise: Great test - and clever solution here to try to compact the test as much as possible while still thoroughly testing behavior.

natsid (approved):

universal-application-tool-0.0.1/app/controllers/admin/AdminProgramBlockPredicatesController.java
carolinedanzi: Ah that makes sense, thank you for clarifying! I added all the operators and updated the comment to explain that only IN and NOT_IN are expected here

browser-test/src/admin_predicates.test.ts
carolinedanzi: Thanks! I tried :)

PR #1520, opened 2021-06-28
Admin predicate UI: Add help text for entering "comma-separated values" in some cases

natsid:
Also fixed a small bug in another part of the JavaScript where we should have checked for "SELECTIONS" _in addition to "SELECTION". Before this PR there was a null error being printed in the console in some cases.

Screen Shot 2021-06-28 at 3 00 11 PM
Screen Shot 2021-06-28 at 3 00 18 PM
Screen Shot 2021-06-28 at 3 09 35 PM

### Issue(s)
Works toward #322


carolinedanzi (approved):
Thank you for adding the help text, it really makes things clearer!

PR #1529, opened 2021-06-29
Add an aria-label to the Edit link in application review page

### LINKED ISSUE #1464: Add unique labels to "Edit" buttons in application review pages
Like in #1455, we reuse the same "Edit" text for every link on the application review pages. Also like #1455, this makes things hard for screen readers.

image

We should ensure each "Edit" link has some unique markings, such as an `aria-label` or using Tailwind's [`.sr-only`](https://tailwindcss.com/docs/screen-readers) class (I would recommend the latter).

carolinedanzi:
### Description
The "Edit" link does not provide enough context to know what you are editing. This adds an `aria-label` so screen readers will read out the question you are editing 

Verified with a Chrome screen reader

### Checklist

### Issue(s)
Fixes #1464 
Screen Shot 2021-06-29 at 12 01 47 PM

### DIFF
--- a/universal-application-tool-0.0.1/conf/messages
+++ b/universal-application-tool-0.0.1/conf/messages
@@ -152,7 +152,7 @@ link.edit=Edit
 
 # An aria-label for screen readers that helps provide context for the associated edit link. The edit link is for a specific
 # question, so this might say "Edit What is your name?". The {0} variable is the question text.
-aria-label.edit=Edit {0}
+ariaLabel.edit=Edit {0}
 
 # Program Preview Title
 title.programPreview=Program application preview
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/conf/messages
```diff
@@ -150,10 +150,14 @@ link.begin=Start here
 # The text that appears next to an answered question that the applicant can click on to modify the response.
 link.edit=Edit
 
-# Program Prewiew Title
+# An aria-label for screen readers that helps provide context for the associated edit link. The edit link is for a specific
+# question, so this might say "Edit What is your name?". The {0} variable is the question text.
+aria-label.edit=Edit {0}
```
kendrickt, line 155: (non-blocking) I think `ariaLabel.edit` would be more consistent.

carolinedanzi: good point! will update

hmhsu (approved):

PR #1531, opened 2021-06-29
Add an aria-label for the language select in applicant header

### LINKED ISSUE #1471: Add `<label>` for language dropdown in top-right corner
The language dropdown in the top right corner shows the list of available languages and lets the user choose one. The problem is that we never say what that dropdown does, and so a screen reader would just read the active language without explaining what the dropdown is for:

5gKzSxBVrZftJaz

Instead, we should add a `<label>` for this `<select>`, [like in this guide](https://www.w3schools.com/tags/att_option_label.asp). We can use Tailwind's [`.sr-only` class](https://tailwindcss.com/docs/screen-readers) to hide it from people not using screen readers, though

This way, when the screen reader goes to the dropdown, it will read something like "Choose your language:" before reading off the list of languages. This will be much easier to understand.

## Original tester's report

> ACTUAL RESULT: Language popup button is labeled with default selected list item 'English' button.
> EXPECTED RESULT: Predefined label should be associated with the Language popup button Instead announcing the selected language as a button.

carolinedanzi:
### Description
Adds an aria-label to the language select in applicant header

The screen reader reads "Tagalog, Choose a language, menu pop up collapsed, button"

### Checklist

### Issue(s)
Fixes #1471 
Screen Shot 2021-06-29 at 2 22 07 PM

### DIFF
--- a/universal-application-tool-0.0.1/app/services/MessageKey.java
+++ b/universal-application-tool-0.0.1/app/services/MessageKey.java
@@ -53,7 +53,7 @@ public enum MessageKey {
   EXTERNAL_LINK("link.externalLink"),
   FILEUPLOAD_VALIDATION_FILE_REQUIRED("validation.fileRequired"),
   FOOTER_SUPPORT_LINK_DESCRIPTION("footer.supportLinkDescription"),
-  LANGUAGE_LABEL("label.language"),
+  LANGUAGE_LABEL_SR("label.languageSr"),
   LINK_ADMIN_LOGIN("link.adminLogin"),
   LINK_ALL_DONE("link.allDone"),
   LINK_APPLY_TO_ANOTHER_PROGRAM("link.applyToAnotherProgram"),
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/services/MessageKey.java
```diff
@@ -53,6 +53,7 @@
   EXTERNAL_LINK("link.externalLink"),
   FILEUPLOAD_VALIDATION_FILE_REQUIRED("validation.fileRequired"),
   FOOTER_SUPPORT_LINK_DESCRIPTION("footer.supportLinkDescription"),
+  LANGUAGE_LABEL("label.language"),
```
hmhsu, line 56: Suggestion (non-blocking): I've seen the usage of suffix `SR` to denote the string is used specifically for screen reader. I don't feel strongly that convention needs to be followed here though.

hmhsu (approved):
> The screen reader reads "Tagalog, Choose a language, menu pop up collapsed, button"

Question (non-blocking): I can look this up myself but thought it doesn't hurt to ask - what does "tagalog" mean?

universal-application-tool-0.0.1/app/services/MessageKey.java
carolinedanzi: Oh good idea! I now see Sr used in other places - added!

PR #1536, opened 2021-06-29
Fix modal header spacing between title and close button

kendrickt:
### Description
Use `flex-grow` and `gap-4` to make sure there is space between the modal header and the "x" that closes the modal.

OLD
image

NEW
image


natsid (approved):
Thanks for fixing this! Looks way better.

PR #1540, opened 2021-06-30
Add option to set modal width

kendrickt:
### Description
Allow `Modal` to have its width set, and then set the block name/desc modal to w-1/3.

old
image

new
image

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/Modal.java
+++ b/universal-application-tool-0.0.1/app/views/components/Modal.java
@@ -18,6 +18,7 @@ public class Modal {
   private String triggerButtonText;
   private Optional<Tag> triggerButtonContent;
   private String buttonStyles;
+  private Optional<String> width;
 
   private Modal(ModalBuilder builder) {
     this.modalId = builder.modalId;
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/views/components/Modal.java
```diff
@@ -102,6 +105,11 @@ public ModalBuilder setTriggerButtonStyles(String buttonStyles) {
       return this;
     }
 
+    public ModalBuilder setWidth(String width) {
```
natsid, line 108: I guess the kind of funny thing about this is that it's just a style string, so really, we could just make this a general `addStyles` or `setStyles` method. Thoughts? If we made it more general, it should probably take `String... styles`

natsid: If you don't make it more general, I think you should at least add a docstring explaining the format this method expects for the `width` argument. I.e., that it should probably be a tailwind class. Or that it will basically just be taken and applied as a class.

kendrickt: I'm gonna make it an enum for modal widths. I don't want to change the styling for the general modal container too much.

universal-application-tool-0.0.1/app/views/components/Modal.java
```diff
@@ -112,4 +121,20 @@ private void setOptionalFields() {
       triggerButtonText = Optional.ofNullable(triggerButtonText).orElse(modalTitle);
     }
   }
+
+  public enum Width {
```
natsid, line 125: suggestion: Given this, should we use one of these as the "default" width for modals?  And/or add AUTO to this?

natsid (approved):

universal-application-tool-0.0.1/app/views/components/Modal.java
kendrickt: done!

Issue #1309, opened 2021-06-07
Browser test flake - failed program/question creation


hmhsu:
**Describe the bug**
The latest attempt to deflake browser test #1305  didn't fully fix our prod promotion. We have issue when we create a new question or program, somehow the browser test still stays on the create form page.

**Error message**
Program creation:
```
FAIL browser: chromium src/admin_program_translation.test.ts (48.774 s)
  ● Admin can manage translations › create a question and add translations

    expect(received).toEqual(expected) // deep equality

    Expected: "All programs"
    Received: "New program information"

      15 |
      16 |   async expectAdminProgramsPage() {
    > 17 |     expect(await this.page.innerText('h1')).toEqual('All programs');
         |                                             ^
      18 |   }
      19 |
      20 |   async expectProgramExist(programName: string, description: string) {

      at AdminPrograms.<anonymous> (src/support/admin_programs.ts:17:45)
      at step (src/support/admin_programs.ts:33:23)
      at Object.next (src/support/admin_programs.ts:14:53)
      at fulfilled (src/support/admin_programs.ts:5:58)
```
Question creation:
```
FAIL browser: chromium src/admin_program_creation.test.ts
  ● Create program with enumerator and repeated questions › create program with enumerator and repeated questions

    expect(received).toEqual(expected) // deep equality

    Expected: "All Questions"
    Received: "New address question"

      16 |
      17 |   async expectAdminQuestionsPage() {
    > 18 |     expect(await this.page.innerText('h1')).toEqual('All Questions');
         |                                             ^
      19 |   }
      20 |
      21 |   async fillInQuestionBasics(questionName: string,

      at AdminQuestions.<anonymous> (src/support/admin_questions.ts:18:45)
      at step (src/support/admin_questions.ts:33:23)
      at Object.next (src/support/admin_questions.ts:14:53)
      at fulfilled (src/support/admin_questions.ts:5:58)
```

See full error message here https://github.com/seattle-uat/civiform/runs/2765182536?check_suite_focus=true

**Expected behavior**
The tests pass without errors, i.e. the creation succeeds and returns back to program/question list page.

**Additional context**
We recently found out that sometimes page may not be fully loaded (especially javascript) before the browser tests start interacting with the page. This *may* be also caused by such timing issue and adding additional wait guard *may* solve it. More investigation needed though.

hmhsu:
This is still relevant. The issue still happens from time to time today.

leonwong:
[...]

PR #1556, opened 2021-07-30
Wait for navigation after click to make next page checks reliable (#1309)

leonwong:
An attempt to improve browser test reliability. Many flaky test failures seem to involve checks for content on pages that get navigated to after some action, but the failures suggest that the navigation hasn't happened yet. Unfortunately, a reliable local repro has been elusive so I'm proposing we submit this test only change to evaluate impact on our prober runs which are much flakier. 

### Description
Made admin question creation/update tests wait for navigation back to the main questions page after clicking the submit button before processing checks for next page content. If this works, there are other tests that have similar failures, so this is not a complete fix for #1309.

### Checklist

### Issue(s)
#1309

### DIFF
--- a/browser-test/src/support/admin_questions.ts
+++ b/browser-test/src/support/admin_questions.ts
@@ -14,6 +14,13 @@ export class AdminQuestions {
     await this.expectAdminQuestionsPage();
   }
 
+  async clickSubmitButtonAndNavigate(buttonText: string) {
+    await Promise.all([
+      this.page.waitForNavigation(),
+      this.page.click('button:has-text("' + buttonText + '")'),
+    ]);
+  }
+
   async expectAdminQuestionsPage() {
     expect(await this.page.innerText('h1')).toEqual('All Questions');
   }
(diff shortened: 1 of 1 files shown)


dkatzz (approved):

browser-test/src/support/admin_questions.ts
```diff
@@ -14,6 +14,13 @@ export class AdminQuestions {
     await this.expectAdminQuestionsPage();
   }
 
+  async clickSubmitButtonAndNavigate(buttonText: string) {
+    await Promise.all([
+      this.page.waitForNavigation(),
```
ettengererin, line 19: Curiosity question...

I see that this.page.waitForNavigation is called above the button click. My understanding is that we want to wait for navigation **after** the button click. How do we know that the waitForNavigation will happen after the button click?  I assume it's something to do with this being an async function and the Promise type?

ettengererin (approved):

browser-test/src/support/admin_questions.ts
leonwong: Good question. This seems to be a fairly common design pattern e.g., https://playwright.dev/docs/events/ where you asynchronously start listening for the result, then asynchronously trigger the event, and then use await Promise.all to wait for both async events to complete. The reason the waitForNavigation comes first is that you don't want there to be even a tiny gap between where you kick off the event and before you listen for the result. Listening first ensures you'll catch the result and since it's async it does a non-blocking wait while you trigger the click event.

Perhaps a better explanation here: https://github.com/microsoft/playwright/issues/5470

What's confusing is that Playwright functions do a lot of magical waiting themselves so it can be a bit hard to tell where this is essential to do.

PR #1557, opened 2021-08-04
Success toast should appear when action is completed successfully

### LINKED ISSUE #1356: Success toasts should not be red
For example, when I edit a question as a CiviForm admin, I get a toast saying "{Question} updated." That's good, but the toast is red, which makes me think there was an error, even though there wasn't. We should make this toast (and other success toasts we can find throughout the app) another color, perhaps green.

Screenshot 2021-06-14 11 53 39 AM

dkatzz:
### Description
Previously, error toasts were displaying on redirect to the questions page. When the render function on the QuestionListView file is called with the optional "maybeFlash" set, it always should be a success message (since we only render to this page with that set on "Create" or "Update"). I also considered having a "maybeToastType" optional param that can be passed along with the type (success / error / etc.) that the toast should be, but this seems like over-engineering considering we don't know if there would be an error case where we navigate to the questions list (all error cases right now we stay on the question page).  

### Checklist
- [x] Extended the README / documentation, if necessary - N/A

### Issue(s)
Fixes #1356

### DIFF
--- a/browser-test/src/support/admin_questions.ts
+++ b/browser-test/src/support/admin_questions.ts
@@ -18,6 +18,21 @@ export class AdminQuestions {
     expect(await this.page.innerText('h1')).toEqual('All Questions');
   }
 
+  async expectAdminQuestionsPageWithSuccessToast(successText: string) {
+    const toastContainer = await this.page.innerHTML('#toast-container');
+    expect(toastContainer).toContain('bg-green-200');
+    expect(toastContainer).toContain(successText);
+    await this.expectAdminQuestionsPage();
+  }
+
+  async expectAdminQuestionsPageWithUpdateSuccessToast() {
+    await this.expectAdminQuestionsPageWithSuccessToast('updated');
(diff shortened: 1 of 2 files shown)


browser-test/src/support/admin_questions.ts
```diff
@@ -25,6 +25,21 @@ export class AdminQuestions {
     expect(await this.page.innerText('h1')).toEqual('All Questions');
   }
 
+  async expectAdminQuestionsPageWithSuccessToast(successText: string) {
+    const toastContainer = await this.page.innerHTML('#toast-container');
```
ettengererin, line 29: nit: Google typescript testing style guidelines recommend leaving an empty space between setting constants and the expect statements in a test. In other words, add an empty line between line 29 and 30.

dkatzz: Oh nice, thanks! Updated here and in a couple of other spots in the file :)

PR #1559, opened 2021-08-04
Remove hard-coded references to seattle.gov domains

bion:
Removes hard-coded references to seattle-specific domains in application code and moves them into the application.conf file

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
+++ b/universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
@@ -11,6 +11,7 @@ import auth.ProfileUtils;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import com.typesafe.config.Config;
 import forms.AddApplicantToTrustedIntermediaryGroupForm;
 import java.util.Optional;
 import javax.inject.Inject;
@@ -39,6 +40,7 @@ public class TrustedIntermediaryController {
   private final UserRepository userRepository;
   private final MessagesApi messagesApi;
   private final FormFactory formFactory;
+  private final String baseUrl;
 
   @Inject
   public TrustedIntermediaryController(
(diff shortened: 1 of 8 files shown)


universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -115,10 +119,13 @@ public Result addApplicant(Long id, Http.Request request) {
       return redirect(
           routes.TrustedIntermediaryController.dashboard(Optional.empty(), Optional.empty()));
     } catch (EmailAddressExistsException e) {
+      String tiUrl = baseUrl + "/trustedIntermediaries";
```
ettengererin, line 122: readability nit: Use descriptive names to be clear to a new reader. I would change "tiURL" to "trustedIntermediaryUrl"

universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
```diff
@@ -66,7 +68,7 @@ public HtmlBundle getBundle(HtmlBundle bundle) {
         meta().attr("name", "viewport").attr("content", "width=device-width, initial-scale=1"));
 
     // Add the warning toast, only for staging
-    if (STAGING_HOST_NAME.equals(hostName)) {
+    if (isStaging) {
```
ettengererin, line 71: Optional: Why add this isStaging field if it is only used once? 

You could leave it as if(hostName.equals(stagingHostname)). I understand you might be going for clarity with isStaging field.

universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -28,11 +28,13 @@
 
   private final BaseHtmlLayout layout;
   private final Config config;
+  private final FakeAdminClient fakeAdminClient;
 
   @Inject
-  public LoginForm(BaseHtmlLayout layout, Config config) {
+  public LoginForm(BaseHtmlLayout layout, Config config, FakeAdminClient fakeAdminClient) {
```
ettengererin, line 34: Out of curiosity, can you explain why this change is necessary?

ettengererin (approved):

universal-application-tool-0.0.1/app/controllers/ti/TrustedIntermediaryController.java
bion: Done

universal-application-tool-0.0.1/app/views/BaseHtmlLayout.java
bion: `isStaging` seemed like it takes slightly less effort for the reader's eyes to parse when they read the method

universal-application-tool-0.0.1/app/views/LoginForm.java
bion: Prior to this change, `canEnable` was a static method that relied on a hard-coded value that is now provided via start-time configuration. Static methods can't access start-time configuration, so `canEnable` needed to become an instance method, requiring an instance of `fakeAdminClient` be injected here.

As an aside, I generally try to encourage devs working in OO languages to prefer instance methods for public interfaces for this exact reason.

ettengererin: Great, thanks so much for this explanation.

PR #1564, opened 2021-08-09
Fix browser tests

dkatzz:
### Description
Browser tests seemed to be failing since the question bank clears out after a file upload question and sometimes 'second-static-q' was getting chosen instead of 'static-q' because of similarities in the name. Updated and it seems to pass now

### Checklist

### Issue(s)

### DIFF
--- a/browser-test/src/application_review.test.ts
+++ b/browser-test/src/application_review.test.ts
@@ -21,15 +21,15 @@ describe('normal application flow', () => {
     await adminQuestions.addNumberQuestion('number-q');
     await adminQuestions.addTextQuestion('text-q');
     await adminQuestions.addRadioButtonQuestion('radio-q', ['one', 'two', 'three']);
-    await adminQuestions.addStaticQuestion('static-q');
+    await adminQuestions.addStaticQuestion('q-static');
     await adminQuestions.addStaticQuestion('second-static-q');
 
     const programName = 'a shiny new program';
     await adminPrograms.addProgram(programName);
     await adminPrograms.editProgramBlock(programName, 'block description', ['date-q', 'address-q', 'name-q', 'radio-q', 'email-q']);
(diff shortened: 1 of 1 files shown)


browser-test/src/application_review.test.ts
```diff
@@ -21,15 +21,15 @@ describe('normal application flow', () => {
     await adminQuestions.addNumberQuestion('number-q');
     await adminQuestions.addTextQuestion('text-q');
     await adminQuestions.addRadioButtonQuestion('radio-q', ['one', 'two', 'three']);
-    await adminQuestions.addStaticQuestion('static-q');
+    await adminQuestions.addStaticQuestion('q-static');
```
leonwong, line 24: Just out of curiosity, was this name change significant to getting the test to pass? Super minor nit, but it departs from the naming convention of all the other questions.

leonwong (approved):

browser-test/src/application_review.test.ts
ettengererin: Seems like we need the two names for test static questions to be different enough so that expectActiveQuestionExist recognizes both. Seems like the issue before was that we had 'static-q' and 'second-static-q', so sometimes the test would look for 'static-q' and find 'second-static-q' and think it passed, but that is a faulty pass.

Let's name questions something more descriptive, or add a comment to explain why they're named like this. We want a new reader of the code to be able to easily why we did something -- it could help them solve bugs and write tests in the future. 

I think 'first-static-q' and 'second-static-q' should be enough to resolve flakiness here.  What do you think?

ettengererin (commented):
Nice debugging, Danielle! Holding off on approval until we get the naming fixed.

browser-test/src/application_review.test.ts
dkatzz: Yeah that describes it perfectly!! Agreed 'first-static-q' and 'second-static-q' are good names to solve the issue and stick to the convention. Updated, thanks!

PR #1580, opened 2021-08-12
Fix text wrapping on program summary and preview view.

### LINKED ISSUE #1567: ORCA static question text is truncated
**Describe the bug**
Code of conduct text and link are truncated

**To Reproduce**
Click on apply to ORCA Middle School Opportunity Program
Scroll to bottom of ORCA application 

**Expected behavior**
Code of conduct text should be legible, but they are truncated on the right.

**Screenshots**
IMG_4011

 - Device: iPhone 8+
 - OS: iOS 14.7.1
 - Browser: safari
 - Version: latest

**Additional context**
N/A

ettengererin:
### Description
Bug addressed: Text does not wrap for long words, like copy/pasted URLs, in the application summary rows. We added custom CSS to break long words in order to wrap text. 

There is a similar tailwind.css property - https://tailwindcss.com/docs/word-break#break-words. However, the tailwind css does not actually behave as expected and does not solve the issue (looks like this property is not maintained by tailwind anymore).

Adding Bion as a reviewer here to confirm that adding custom css in this way is safe and to recommend any alternatives.

Before: 
image

After:
image

### Checklist

### Issue(s)
Fixes #1567

### DIFF
--- a/universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
+++ b/universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
@@ -223,7 +223,8 @@ public final class ApplicantProgramSummaryView extends BaseHtmlView {
             Styles.P_2,
             Styles.PT_4,
             Styles.BORDER_B,
-            Styles.BORDER_GRAY_300);
+            Styles.BORDER_GRAY_300)
+        .attr("style", "word-break:break-word"); 
   }
 
   private ContainerTag renderRepeatedEntitySection(
--- a/universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
+++ b/universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
@@ -224,7 +224,7 @@ public final class ApplicantProgramSummaryView extends BaseHtmlView {
             Styles.PT_4,
             Styles.BORDER_B,
             Styles.BORDER_GRAY_300)
-        .attr("style", "word-break:break-word"); 
+        .attr("style", "word-break:break-word");
   }
 
   private ContainerTag renderRepeatedEntitySection(
--- a/universal-application-tool-0.0.1/app/views/style/BaseStyles.java
+++ b/universal-application-tool-0.0.1/app/views/style/BaseStyles.java
@@ -46,7 +46,6 @@ public final class BaseStyles {
           Styles.BLOCK,
           Styles.OUTLINE_NONE,
           Styles.BOX_BORDER,
-          Styles.H_12,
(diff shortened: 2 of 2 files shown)


bion:
Looks like an important improvement!

Please use the tailwind style system whenever possible. In this case I believe you can add this constant to the class list to get the same result https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/views/style/Styles.java#L2526

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -223,7 +223,8 @@ private ContainerTag renderQuestionSummary(
             Styles.P_2,
             Styles.PT_4,
             Styles.BORDER_B,
-            Styles.BORDER_GRAY_300);
+            Styles.BORDER_GRAY_300)
+        .attr("style", "word-break:break-word");
```
bion, line 227: Looks like an important improvement!

Please use the tailwind style system whenever possible. In this case I believe you can add this constant to the class list to get the same result https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/views/style/Styles.java#L2526

bion (changes requested):

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
bion: tailwind docs: https://tailwindcss.com/docs/word-break

ettengererin: We tried this tailwind property already, noted it in the PR description above:

Tailwind's break-words property does not actually behave as expected and does not solve the issue (looks like this property is not maintained by tailwind anymore). It does not properly wrap the long word text -- The URL link still overflows off the container. Seems like others experienced this problem: https://github.com/tailwindlabs/tailwindcss/issues/835

Break-words from tailwind adds this css:
{overflow-wrap: break-word;}

The custom css we use adds
{word-break: break-word}

Seems like older versions of tailwind worked properly for this use case, but the newer versions are causing this issue for folks.

bion: Ah! So sorry for not reading the description properly.

bion (approved):

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
ettengererin: No worries!

PR #1581, opened 2021-08-13
Don't include trailing punctuation in parsed URLs.

### LINKED ISSUE #1579: ORCA code of conduct link is broken
**Describe the bug**
Looks like the destination URL includes the period at the end of the sentence, so it ends up being broken for end users even though the display text of the URL is correct.

**To Reproduce**
Fill out ORCA form and navigate to the final screen with code of conduct info. 
Try to click on the King County code of conduct link in the following block of text:

"City of Seattle ORCA Opportunity Program Rules: Program participants are expected to follow the Metro Code of Conduct, http://kingcounty.gov/depts/transportation/metro/about/safety-security/code-of-conduct.aspx."

**Expected behavior**
Expect to navigate to http://kingcounty.gov/depts/transportation/metro/about/safety-security/code-of-conduct.aspx

... but it actually navigates to http://kingcounty.gov/depts/transportation/metro/about/safety-security/code-of-conduct.aspx. (notice the period at the end) which ends up hitting a 404 not found page.

leonwong:
### Description
The UrlDetector class we're using includes some types of trailing punctuation characters in detected URLs. While these are legal URL path characters, in the context of Civiform free text fields, these are more likely to be part of the surrounding text instead of the URL. This PR strips typical trailing punctuation from detected URLs.

### Checklist

### Issue(s)
Fixes #1579

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/TextFormatter.java
+++ b/universal-application-tool-0.0.1/app/views/components/TextFormatter.java
@@ -15,9 +15,11 @@ import com.linkedin.urls.detection.UrlDetector;
 import com.linkedin.urls.detection.UrlDetectorOptions;
 import j2html.tags.ContainerTag;
 import j2html.tags.DomContent;
+import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import views.style.Styles;
@@ -43,9 +45,22 @@ public class TextFormatter {
   private static final String BULLETED_ITEM = "* ";
 
   public static ImmutableList<DomContent> createLinksAndEscapeText(String content) {
-    List<Url> urls = new UrlDetector(content, UrlDetectorOptions.Default).detect();
+    // JAVASCRIPT option avoids including surrounding quotes or brackets in the URL.
+    List<Url> urls = new UrlDetector(content, UrlDetectorOptions.JAVASCRIPT).detect();
+
     ImmutableList.Builder<DomContent> contentBuilder = ImmutableList.builder();
     for (Url url : urls) {
+      try {
+        // While technically they could be part of the URL, trailing punctuation
+        // is more likely to be part of the surrounding text, so we strip.
+        url = Url.create(StringUtils.stripEnd(url.getOriginalUrl(), ".?!,:;"));
+      } catch (MalformedURLException e) {
+        LOG.error(
+            String.format(
+                "Failed to parse URL %s after stripping trailing punctuation",
+                url.getOriginalUrl()));
+      }
+
       int index = content.indexOf(url.getOriginalUrl());
       // Find where this URL is in the text.
       if (index == -1) {
--- a/universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
+++ b/universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
@@ -12,17 +12,22 @@ public class TextFormatterTest {
   @Test
   public void urlsRenderCorrectly() {
     ImmutableList<DomContent> content =
-        TextFormatter.createLinksAndEscapeText("hello google.com http://internet.website");
+        TextFormatter.createLinksAndEscapeText(
+            "Hello google.com (http://internet.website), http://mysite.com...!");
 
-    assertThat(content).hasSize(4);
-    assertThat(content.get(0).render()).isEqualTo(new Text("hello ").render());
+    assertThat(content).hasSize(7);
+    assertThat(content.get(0).render()).isEqualTo(new Text("Hello ").render());
     assertThat(content.get(1).render())
         .isEqualTo("<a href=\"http://google.com/\" class=\"opacity-75\">google.com</a>");
-    assertThat(content.get(2).render()).isEqualTo(new Text(" ").render());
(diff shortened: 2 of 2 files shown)


leonwong:
Reviewers - please also include feedback on whether you think this is safe to merge today for the ORCA launch. Thanks!

universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
```diff
@@ -12,17 +12,22 @@
   @Test
   public void urlsRenderCorrectly() {
     ImmutableList<DomContent> content =
-        TextFormatter.createLinksAndEscapeText("hello google.com http://internet.website");
+        TextFormatter.createLinksAndEscapeText(
+            "Hello google.com (http://internet.website), http://mysite.com...!");
```
ettengererin, line 16: Optional: I would be inclined to keep the existing urlsRenderCorrectly test as is, and add another test called urlsRenderCorrectly_stripsTrailingPunctuation that tests this URL.

This has two pros: (1) We still test for a url without trailing punctuation. (2) It's clearer to the user that we are testing a simple case and then a complex case. Having a urlsRenderCorrectly_stripsTrailingPunctuation test would make it clearer to a reader why we are testing such a complex string ("Hello google.com (http://internet.website), http://mysite.com...!"). When adding a new feature I am generally inclined to add more test coverage, rather than replace existing tests.

Either way, please add bak a test case for a URL without trailing punctuation, however you choose to do it.

universal-application-tool-0.0.1/app/views/components/TextFormatter.java
```diff
@@ -43,9 +45,22 @@
...
-    List<Url> urls = new UrlDetector(content, UrlDetectorOptions.Default).detect();
+    // JAVASCRIPT option avoids including surrounding quotes or brackets in the URL.
+    List<Url> urls = new UrlDetector(content, UrlDetectorOptions.JAVASCRIPT).detect();
+
     ImmutableList.Builder<DomContent> contentBuilder = ImmutableList.builder();
     for (Url url : urls) {
+      try {
+        // While technically they could be part of the URL, trailing punctuation
+        // is more likely to be part of the surrounding text, so we strip.
+        url = Url.create(StringUtils.stripEnd(url.getOriginalUrl(), ".?!,:;"));
```
ettengererin, line 56: Should we include any of the characters below here?

` < >  [ ] { } 

You may have avoided some of these because they can be in URLs and/or are unlikely to be used as trailing punctuation. Looks like UrlDetectorOptions.JAVASCRIPT detects some of these, so just wanna make sure we're not missing anything else.

universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
```diff
@@ -12,17 +12,22 @@
...
+    assertThat(content.get(0).render()).isEqualTo(new Text("Hello ").render());
     assertThat(content.get(1).render())
         .isEqualTo("<a href=\"http://google.com/\" class=\"opacity-75\">google.com</a>");
-    assertThat(content.get(2).render()).isEqualTo(new Text(" ").render());
+    assertThat(content.get(2).render()).isEqualTo(new Text(" (").render());
     assertThat(content.get(3).render())
         .isEqualTo(
             "<a href=\"http://internet.website/\""
                 + " class=\"opacity-75\">http://internet.website</a>");
+    assertThat(content.get(4).render()).isEqualTo(new Text("), ").render());
```
ettengererin, line 27: I am surprised this is accurately parsing the closed parenthesis ), since you don't include parenthesis in TextFormatter line 56. Do you know why this is passing? Perhaps I am missing something.

ettengererin: Oh, I think this passes because UrlDetectorOptions.JAVASCRIPT catches the parenthesis and doesn't include them. Let me know if I'm wrong, or feel free to ignore if I'm right :D

ettengererin (approved):
This seems safe to fix pre launch to me. We're only modifying URLs here, and if we see there is an issue with a URL then we can easily fix/rollback. If we were modifying all program text, I'd be more concerned :) 

As long as we have thoroughly manually tested, I approve.

universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
leonwong: Yes - that's right. the JAVASCRIPT option looks for matched pairs of quotes, and various kinds of brackets and avoids including in the URL if it does find a matching pair.

universal-application-tool-0.0.1/app/views/components/TextFormatter.java
leonwong: I avoided characters that should be already be covered by the JAVASCRIPT option <>[]{} plus single and double quotes. Overall seemed safest to just go after characters that are likely to be part of the surrounding text. The single opening quote you mentioned is an interesting case -- seems unlikely to be part of an URL, or to follow an URL in surrounding text. I don't have a strong feeling, but my inclination would be to err on the side of keeping this fixup hack more constrained.

ettengererin: That SGTM! Thanks for the explanation.

universal-application-tool-0.0.1/test/views/components/TextFormatterTest.java
leonwong: Sounds reasonable. Updated PR with commit to leave existing test and add an additional covering punctuation stripping.

PR #1606, opened 2021-08-30
Only include submitted applications in the program admin view and download

### LINKED ISSUE #1603: Draft applications show up in admin application view/csv
**Describe the bug**
Tanya Test from Civiform UX reported the following: She created a draft of an application and did NOT submit it. Rosa reported that she was able to see Tanya's application in her admin view, and that fields of her application were incomplete.

Rosa did **not** receive an email confirming Tanya's application submission. So, it seems like when an applicant creates a draft application, Civiform knows that their application is incomplete, but it still puts their draft app in the the application/csv view.

This is confusing for an admin because they think this applicant made mistakes in their applications. This is problematic for the applicant because they may not want the city to see her application if she has not submitted it.

**Expected behavior**
Draft applications should not appear in admin application.

**Additional context**

lindsayzad:
### Description
Updated the program getApplications to be getSubmittedApplications, filtering by lifecycleStage.Active.  This makes it so that draft applications, as well as deleted or obsolete applications, will not be shown.  They will not show up on the program admin view page or in the downloaded csv

### Checklist
- May need a follow up cl with a browser test replicating this.  The ProgramServiceImplTest file only makes changes from the program admin perspective; there is no mocking the program to return apps or actually filling out apps in that file.  Adding tests would require a larger test system refactor here, or adding something such as browser test that can cover both the admin roles and applicant roles in one file.  Changes to other files are all trivial; renaming so it is clear what the function provides

### Issue(s)
Fixes #1603

### DIFF
--- a/universal-application-tool-0.0.1/app/services/export/ExporterService.java
+++ b/universal-application-tool-0.0.1/app/services/export/ExporterService.java
@@ -73,7 +73,7 @@ public class ExporterService {
    * @throws ProgramNotFoundException If the program ID refers to a program that does not exist.
    */
   public String getProgramCsv(long programId) throws ProgramNotFoundException {
-    ImmutableList<Application> applications = programService.getProgramApplications(programId);
+    ImmutableList<Application> applications = programService.getSubmittedProgramApplications(programId);
     ProgramDefinition program = programService.getProgramDefinition(programId);
     CsvExporter csvExporter;
     if (program.exportDefinitions().stream()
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/services/export/ExporterService.java
```diff
@@ -115,7 +116,7 @@ public String exportCsv(CsvExporter csvExporter, ImmutableList<Application> appl
   CsvExportConfig generateDefaultCsvConfig(long programId) {
     ImmutableList<Application> applications;
     try {
-      applications = programService.getProgramApplications(programId);
+      applications = programService.getSubmittedProgramApplications(programId);
```
leonwong, line 119: More a comment about the previous implementation, but now that we're doing some more work inside getSubmittedProgramApplications, would it make more sense to pass in the submitted applications from getProgramCsv rather than looking them up here again by programId?

leonwong (approved):

universal-application-tool-0.0.1/app/services/export/ExporterService.java
lindsayzad: I did a small refactor on getProgramCsv so that getSubmittedProgramApplications will not be called multiple times unnecessarily

PR #1609, opened 2021-09-02
Fix Nav bar error

lindsayzad:
### Description
Nav bar not displaying correctly for civiform admins.
Run formatting script on browser tests

Previously: Nav bar is not present for civiform admin.  No way to select 'Questions', 'Versions', etc
Screen Shot 2021-09-02 at 12 16 25 PM

Now: nav bar for civiform admins is restored
Screen Shot 2021-09-02 at 12 16 34 PM

One Open Q; Why didn't browser tests catch this broken nav bar? Locally and in past PRs browser tests all passed, only the prober successfully caught it

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
@@ -35,7 +35,7 @@ public final class ProgramApplicationListView extends BaseHtmlView {
 
   @Inject
   public ProgramApplicationListView(AdminLayout layout) {
-    this.layout = checkNotNull(layout).setProgramAdminType();
+    this.layout = checkNotNull(layout).setOnlyProgramAdminType();
   }
 
   public Content render(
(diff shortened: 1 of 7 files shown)


universal-application-tool-0.0.1/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -35,7 +35,7 @@
 
   @Inject
   public ProgramApplicationListView(AdminLayout layout) {
-    this.layout = checkNotNull(layout).setProgramAdminType();
+    this.layout = checkNotNull(layout).setOnlyProgramAdminType();
```
leonwong, line 38: I think this logic made more sense when civiform admins couldn't also be program admins, since this page would only be visible to program admins so you could make this assumption. Was thinking that ideally we'd want to check what admin type this is and only call this if the user is only a program admin.

Happy to defer this to a future PR, since the ideal implementation might involve a bit of a refactor to consolidate the logic across ProgramApplicationListView, ProgramApplicationView, and perhaps others.

leonwong (approved):
Approved. Just had one comment that can wait for a future PR.

PR #1610, opened 2021-09-03
Whitelabel civic entity name and logos

bion:
Allows specifying a civic entity and logos other than Seattle via config file.

There is still hard-coded codepaths for serving the local asset files for the Seattle logos. We could avoid this by asking Seattle to make the logos available somewhere else and providing URLs for them. I'm inclined to not bother with them that though. If someone feels differently we can change this.

### DIFF
--- a/universal-application-tool-0.0.1/app/views/LoginForm.java
+++ b/universal-application-tool-0.0.1/app/views/LoginForm.java
@@ -69,7 +69,7 @@ public class LoginForm extends BaseHtmlView {
       content.with(
           img()
               .withSrc(maybeLogoUrl.get())
-              .withAlt(civicEntityFullName + "logo")
+              .withAlt(civicEntityFullName + "Logo")
               .attr("aria-hidden", "true")
               .withClasses(Styles.W_1_4, Styles.PT_4));
     } else {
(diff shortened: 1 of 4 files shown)


universal-application-tool-0.0.1/app/views/LoginForm.java
```diff
@@ -54,17 +65,27 @@ public Content render(Http.Request request, Messages messages, Optional<String>
...
-            .viewUtils
-            .makeLocalImageTag("ChiefSeattle_Blue")
-            .withAlt("City of Seattle Logo")
-            .attr("aria-hidden", "true")
-            .withClasses(Styles.W_1_4, Styles.PT_4));
+    if (maybeLogoUrl.isPresent()) {
+      content.with(
+          img()
+              .withSrc(maybeLogoUrl.get())
+              .withAlt(civicEntityFullName + "logo")
```
dkatzz, line 72: Nit: logo isn't capitalized here but is capitalized below.

dkatzz (approved):

universal-application-tool-0.0.1/app/views/LoginForm.java
bion: Done

PR #1614, opened 2021-09-13
squashed changes

### LINKED ISSUE #1604: Allow the ability to publish programs to a limited audience only
When the civiform admin creates a new program, they may want to create a draft of the program and collect feedback/test it out before publishing it more broadly to the general public.  Currently, there is no good way to do this; staging gets wiped regularly, and publishing on prod makes it so that the program is visible to everyone and anyone who wants to can start applying.

Allow the ability for a civiform admin to manage who has access to a program when they publish it.  They could publish it broadly to everyone so it goes on the civiform hompage, or they should be able to enter only a few select users, and only those users when logged in should have access to the program.  This makes it so that the program can be tested/reviewed by other team members before going public

**Additional context*
This is a critical feature to making sure a program is ready before it launches on civiform, and will be used more as we onboard larger programs to the platform.  It makes civiform more polished, and establishes trust for the government partners new to it

**Done when**
A draft/limited access program can be published for review without being globally published to all

lindsayzad:
### Description
Allow the civiform admin to hide a program that they don't want to appear on the civiform applicant landing page.  This program can still be accessed through its deeplink, and will still act normally once accessed.

This change adds a new boolean hideFromView field to the programs DB that is then set by the civiform admin for a program and is accessed when showing programs on the applicant home screen.  The boolean defaults to false when it is not set; programs default to not being hidden.

Civiform admin checks the 'Hide from view' box in the program info page, and publishes the new version of the program with this update.

Screen Shot 2021-09-13 at 2 57 17 PM

Depending on the hide from view status, the program is or is not visible in the applicant home screen.
Screen Shot 2021-09-09 at 3 55 02 PM
Screen Shot 2021-09-09 at 4 41 33 PM

### Checklist
- [ X ] Created tests which fail without the change (if possible). This change includes a browser test that verifies a program can be hidden, unhidden, and seen or not as intended by the applicant

### Issue(s)
Fixes #1604

### DIFF
--- a/browser-test/src/admin_program_hidden.test.ts
+++ b/browser-test/src/admin_program_hidden.test.ts
@@ -1,0 +1,44 @@
+import { startSession, loginAsProgramAdmin, loginAsAdmin, AdminQuestions, AdminPrograms, endSession, logout, loginAsTestUser, selectApplicantLanguage, ApplicantQuestions, userDisplayName } from './support'
+
+describe('Hide a program that should not be public yet', () => {
+  it('Create a new hidden program, verify applicants cannot see it on the home page, unhide the program, verify applicant access', async () => {
+    const { browser, page } = await startSession()
+    page.setDefaultTimeout(5000);
+
+    await loginAsAdmin(page);
+    const adminQuestions = new AdminQuestions(page);
+    const adminPrograms = new AdminPrograms(page);
+
+    // Create a hidden program
+    const programName = 'Hidden Program';
+    const programDescription = 'Description';
+    await adminPrograms.addProgram(programName, programDescription, "", true); // Will this work? lol
+    await adminPrograms.publishAllPrograms();
+
+    // Login as applicant
+    await logout(page);
+    await loginAsTestUser(page);
+    await selectApplicantLanguage(page, 'English');
+    const applicantQuestions = new ApplicantQuestions(page);
+    await applicantQuestions.validateHeader('en-US');
+
+    // Verify the program cannot be seen
+    await applicantQuestions.expectProgramNotExist(programName);
+
+    await logout(page);
+    await loginAsAdmin(page);
+
+    // Unhide the program and publish
+    await adminPrograms.createUnhiddenVersion(programName);
+    await adminPrograms.publishAllPrograms();
+
+    await logout(page);
+
+    // Verify applicants can now see the program
+    await loginAsTestUser(page);
+    await selectApplicantLanguage(page, 'English');
+    await applicantQuestions.expectProgramExist(programName, programDescription);
+
+    await endSession(browser);
+  })
+})
--- a/browser-test/src/support/admin_programs.ts
+++ b/browser-test/src/support/admin_programs.ts
@@ -25,7 +25,7 @@ export class AdminPrograms {
     expect(tableInnerText).toContain(description);
   }
 
-  async addProgram(programName: string, description = 'program description', externalLink = '') {
+  async addProgram(programName: string, description = 'program description', externalLink = '', hidden = false) {
     await this.gotoAdminProgramsPage();
     await this.page.click('#new-program-button');
 
@@ -34,6 +34,10 @@ export class AdminPrograms {
     await this.page.fill('#program-display-name-input', programName);
     await this.page.fill('#program-display-description-textarea', description);
     await this.page.fill('#program-external-link-input', externalLink);
+
+    if (hidden) {
+      await this.page.check(`label:has-text("Hide this program")`);
+    }
 
     await this.page.click('#program-update-button');
 
@@ -199,6 +203,16 @@ export class AdminPrograms {
     await this.page.click('#program-update-button');
     await this.expectDraftProgram(programName);
   }
+
+  async createUnhiddenVersion(programName: string) {
+    await this.gotoAdminProgramsPage();
+    await this.expectActiveProgram(programName);
+    await this.page.click(this.selectWithinProgramCard(programName, 'ACTIVE', ':text("New Version")'));
+    await this.page.uncheck(`label:has-text("Hide this program")`);
+    //await this.page.pause();
+    await this.page.click('#program-update-button');
+    await this.expectDraftProgram(programName);
+    }
 
   async viewApplications(programName: string) {
     await this.page.click(this.selectWithinProgramCard(programName, 'ACTIVE', 'a:text("Applications")'));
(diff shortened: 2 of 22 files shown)


universal-application-tool-0.0.1/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -63,6 +66,13 @@ private static ContainerTag buildProgramForm(
             .setLabelText("Describe this program for administrative use")
             .setValue(adminDescription)
             .getContainer(),
+        FieldWithLabel.checkbox()
+            .setId("program-hide-from-view-checkbox")
+            .setFieldName("hideFromView")
+            .setLabelText("Hide this program")
```
bion, line 72: suggestion (blocking): "hide" here could be a bit ambiguous since the program is still accessible via URL. recommend changing the label text to `hide this program on public index page`

bion: i tried to think of other things that could go here but they're all kinda wordy, unfortunately. on the plus side, the admin will definitely be on the laptop and there is plenty of lateral space here in the UI

lindsayzad: Good point, done

Screen Shot 2021-09-13 at 3 27 37 PM

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -50,6 +50,9 @@ public static Builder builder() {
   /** An external link to a page containing more details for this program. */
   public abstract String externalLink();
 
+  /** Whether the program should be hidden from the Applicant's program overview page. */
+  public abstract boolean hideFromView();
```
bion, line 54: suggestion (blocking): this could would be a bit more open to extension if instead of using a boolean `hideFromView` field it used an enum `displayMode` value. For now it would just have two possible values: `PUBLIC` and `HIDDEN_IN_INDEX`, but could be extended with other options they wanted for example to have another mode that also changes the behavior of the deep link.

browser-test/src/admin_program_hidden.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    page.setDefaultTimeout(5000);
+
+    await loginAsAdmin(page);
+    const adminQuestions = new AdminQuestions(page);
+    const adminPrograms = new AdminPrograms(page);
+
+    // Create a hidden program
+    const programName = 'Hidden Program';
+    const programDescription = 'Description';
+    await adminPrograms.addProgram(programName, programDescription, "", true); // Will this work? lol
```
dkatzz, line 15: Remove the comment? :)

browser-test/src/support/applicant_questions.ts
```diff
@@ -72,6 +72,19 @@ export class ApplicantQuestions {
     await this.page.click(`#continue-application-button`);
   }
 
+  async expectProgramExist(programName: string, description: string) {
```
dkatzz, line 75: Nit: maybe "expectProgramVisible" and below "expectProgramNotVisible" or something instead?

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
lindsayzad: Done

browser-test/src/admin_program_hidden.test.ts
lindsayzad: Thanks!

browser-test/src/support/applicant_questions.ts
lindsayzad: Done

browser-test/src/admin_program_hidden.test.ts
```diff
@@ -0,0 +1,44 @@
+import { startSession, loginAsProgramAdmin, loginAsAdmin, AdminQuestions, AdminPrograms, endSession, logout, loginAsTestUser, selectApplicantLanguage, ApplicantQuestions, userDisplayName } from './support'
+
+describe('Hide a program that should not be public yet', () => {
```
dkatzz, line 3: Is there any flakiness here? Maybe we should add `await waitForPageJsLoad(page);` in certain places to ensure page loads are complete (following what @leonwong did in https://github.com/seattle-uat/civiform/pull/1617)?

dkatzz, line 3: Is there any flakiness here? Maybe we should add await waitForPageJsLoad(page); in certain places to ensure page loads are complete (following what @leonwong did in #1617)?

lindsayzad: Good idea.  I merged that code in and added that in the appropriate places (after new page.click uses)

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -50,6 +50,9 @@ public static Builder builder() {
   /** An external link to a page containing more details for this program. */
   public abstract String externalLink();
 
+  /** The program's display mode. */
+  public abstract String displayMode();
```
bion, line 54: question (blocking): why `string` here rather than `DisplayMode`? it looks like most of the code ends up dealing directly with strings rather than enums for this value which sidesteps helpful type checking.

browser-test/src/support/admin_programs.ts
```diff
@@ -216,6 +222,17 @@ export class AdminPrograms {
     await this.expectDraftProgram(programName);
   }
 
+  async createPublicVersion(programName: string) {
```
ettengererin, line 225: nit: It's nice to separate expects and test setup with empty lines -- we used this as style best practices on my last team. This way, we can see clearly what part of this function is general setup (clicking buttons, navigating to pages) and what part is actually expecting something.

So for example, I would add some empty lines to this function to make it look like:
async createPublicVersion(){
    wait this.gotoAdminProgramsPage();

    await this.expectActiveProgram(programName);

    await this.page.click(this.selectWithinProgramCard(programName, 'ACTIVE', ':text("New Version")'));
    await waitForPageJsLoad(this.page);
    await this.page.check(`label:has-text("Public")`);
    await this.page.click('#program-update-button');
    await waitForPageJsLoad(this.page);

    await this.expectDraftProgram(programName);
}

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
lindsayzad: Done!

browser-test/src/support/admin_programs.ts
lindsayzad: Thanks, made that change

bion (approved):

ettengererin:
This is an awesome feature, Lindsay! Seattle will be so excited about this.

PR #1627, opened 2021-09-29
Multi option front end validation

### LINKED ISSUE #1587: When creating a dropdown question, all progress is wiped if you leave a field blank and click "submit".
**Describe the bug**
When creating a dropdown question, I will create a question with a few dropdown answer options. If I leave an option field blank and click "create", the creation will fail with an error "Multi option questions cannot have a blank option" (this is good) BUT it will also wipe out all the option fields I did fill out. This is frustrating as an admin because I wish my populated fields would still be populated. However, now I need to go fill out all my answer options again. 

This is extra annoying for long dropdowns -- for example, a dropdown of all the public schools in Seattle. If I accidentally leave an option field blank and submit, then I have to re-enter 100+ options.

**To Reproduce**
1. Go to Admins Questions tabs
2. Create a new Dropdown question
3. Fill out all fields. 
4. Add an answer option and fill it out.
5. Add another answer option and leave it blank.
4. Hit submit
5. All answer options will be empty fields will be empty :(

**Expected behavior**
Same error message should be shown, and previously populated answer options to still be populated. User should just get a chance to fill out or delete the empty answer option.

**Screenshots**
image

image

dkatzz:
### Description
Add front end validation for empty options in the admin multi-option question create / edit flow. Previously, the options would be cleared when create / edit was clicked and there was an empty option, but now they will be persisted. We'll plan to move other admin question validation to the front end too as a follow up to prevent other error cases from being cleared. 

This is similar to what happens in applicant validation (https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/assets/javascripts/validation.ts), but the admin validation is only loaded in admin flows. 

We'll still need to translate the text to other languages (assuming we offer translations for admins?)

Worked on with @ettengererin 

### Checklist

### Issue(s)
Fixes #1587

Screen Shot 2021-09-29 at 4 42 48 PM
Screenshots of errors shown:


universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+  static readonly MULTI_OPTION_QUESTION_FIELD_NAME_EDIT = '#question-settings input[name="options[]"]';
+
+  isMultiOptionCreateValid = true;
+  isMultiOptionEditValid = true;
+
+  constructor() {
+    // Attach listener to admin program edit form.
+    const adminProgramEditForm = document.getElementById("question-form");
+    if (adminProgramEditForm) {
+      adminProgramEditForm.addEventListener("submit", (event) => { return this.attemptSubmit(event); });
```
bion, line 13:
```suggestion
      adminProgramEditForm.addEventListener("submit", (event) => this.attemptSubmit(event));
```

bion: or alternatively

```suggestion
      adminProgramEditForm.addEventListener("submit", this.attemptSubmit.bind(this));
```

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  private attemptSubmit(event: Event): boolean {
+    this.checkFields();
+    if (!this.isValid()) {
+      event.preventDefault();
+      return false;
+    }
+    return true;
```
bion, line 23: suggestion (blocking): afaik the return value for the event listener is ignored, so no need for return statements in `attemptSubmit`

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
+/** The validation controller provides basic client-side validation of admin form fields. */
+class AdminValidationController {
+  static readonly MULTI_OPTION_QUESTION_FIELD_NAME_CREATE = '#question-settings input[name="newOptions[]"]';
+  static readonly MULTI_OPTION_QUESTION_FIELD_NAME_EDIT = '#question-settings input[name="options[]"]';
+
+  isMultiOptionCreateValid = true;
+  isMultiOptionEditValid = true;
```
bion, line 7: question (blocking): should these be private?

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+  private attemptSubmit(event: Event): boolean {
+    this.checkFields();
+    if (!this.isValid()) {
+      event.preventDefault();
+      return false;
+    }
+    return true;
+  }
+
+  private checkFields() {
```
bion, line 26: suggestion (non-blocking): rename `checkFields` to `validate` for consistency with the methods it delegates to

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.isMultiOptionCreateValid = this.validateMultiOptionQuestionCreate();
+    this.isMultiOptionEditValid = this.validateMultiOptionQuestionEdit();
+  }
+
+  updateFieldErrorState(element: Element, fieldName: string, isValid: boolean) {
```
bion, line 31: suggestion (blocking): it looks like this and the methods below are all for class-internal use only. please make them private as well for principle of least access

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Attach listener to admin program edit form.
+    const adminProgramEditForm = document.getElementById("question-form");
+    if (adminProgramEditForm) {
+      adminProgramEditForm.addEventListener("submit", (event) => { return this.attemptSubmit(event); });
+    }
+  }
+
+  private attemptSubmit(event: Event): boolean {
+    this.checkFields();
+    if (!this.isValid()) {
```
bion, line 19: suggestion (non-blocking): recommend removing `isValid` and having `checkFields` (a.k.a. `validateAll`) return a boolean. `isValid` scares me a little bit because it relies on another method getting called first for it its return value to be correct.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+  isValid() {
+    return this.isMultiOptionCreateValid && this.isMultiOptionEditValid;
+  }
+
+  /** Validates that there are no empty options. Returns true if all fields are valid.  */
+  validateMultiOptionQuestionOptions(options: HTMLInputElement[]) {
+    let multiOptionIsValid = true;
+    for (const option of options) {
+      const inputIsValid = option.value !== '';
+      this.updateFieldErrorState(option, ".cf-multi-option-input", inputIsValid);
```
bion, line 53: suggestion (non-blocking): extract `".cf-multi-option-input"` to constant

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,74 @@
...
+    return true;
+  }
+
+  private checkFields() {
+    this.isMultiOptionCreateValid = this.validateMultiOptionQuestionCreate();
+    this.isMultiOptionEditValid = this.validateMultiOptionQuestionEdit();
+  }
+
+  updateFieldErrorState(element: Element, fieldName: string, isValid: boolean) {
+    const errorDiv = element.parentElement ? element.parentElement.querySelector(fieldName + '-error') : element.querySelector(fieldName + '-error');
```
bion, line 32: question (blocking): when does the element not have a parent?

browser-test/src/support/admin_questions.ts
```diff
@@ -423,6 +455,19 @@ export class AdminQuestions {
...
+
+    await this.expectDraftQuestionExist(questionName, questionText);
+  }
+
+  async createRadioButtonQuestion(questionName: string,
+    options: Array<string>,
+    description = 'radio button description',
+    questionText = 'radio button question text',
+    helpText = 'radio button question help text',
+    enumeratorName = AdminQuestions.DOES_NOT_REPEAT_OPTION){
```
bion, line 470: suggestion (non-blocking): this is a lot of params, it may be easier to read at the call site if they were bundled in an object

```suggestion
type CreateRadioButtonQuestionParams = {
  questionName: string;
  options: Array<string>;
  description?: string;
  questionText?: string;
  helpText?: string;
  enumeratorName?: string;
}

  async createRadioButtonQuestion({
    questionName,
    options,
    description = 'radio button description',
    questionText = 'radio button question text',
    helpText = 'radio button question help text',
    enumeratorName = AdminQuestions.DOES_NOT_REPEAT_OPTION}: CreateRadioButtonQuestionParams) {
```

browser-test/src/support/admin_questions.ts
```diff
@@ -308,28 +322,46 @@ export class AdminQuestions {
...
+  }
 
-    await this.clickSubmitButtonAndNavigate('Create');
+  /** Changes the input field of a multi option answer. */
+  async changeMultiOptionAnswer(index: number, text: string) {
+    await this.page.fill(`:nth-match(#question-settings div.flex-row, ${index}) input`, text);
+  }
 
+  /** Fills out the form for a dropdown question, clicks submit, and verifies the new question exists.  */
+  async addDropdownQuestion(questionName: string,
```
bion, line 356: suggestion (non-blocking): same suggestion as with `createRadioButtonQuestion`

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Yes, nice catch. Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Good idea. Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: That's cleaner, good suggestion. Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Done. Extracted as a static readonly constant...let me know if it should be something else.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
ettengererin: Good point -- I think we were trying to catch a case that actually doesn't exist. I deleted this -- but...

@dkatzz please take a look at this because you wrote this line of code and I don't want to delete logic that you think is important.

dkatzz: Yeah, the compiler complained about it possibly being null when doing `element.parentElement.querySelector...`, but the exclamation point is a good way to make this clearer - thanks!

browser-test/src/support/admin_questions.ts
dkatzz: Updated - let me know what you think! Can also change the add__Question functions to follow this pattern as a follow up.

browser-test/src/support/admin_questions.ts
dkatzz: Fixed.

bion: It's still using positional args

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,65 @@
...
+    // Attach listener to admin program edit form.
+    const adminProgramEditForm = document.getElementById("question-form");
+    if (adminProgramEditForm) {
+      adminProgramEditForm.addEventListener("submit", (event) => this.attemptSubmit(event));
+    }
+  }
+
+  private attemptSubmit(event: Event) {
+    this.validateForm();
+    if (!this.validateForm()) {
```
bion, line 17: suggestion (blocking): no need to call `validateForm` twice

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
```diff
@@ -0,0 +1,65 @@
...
+      field.classList.toggle('border-red-600', !isValid);
+    }
+  }
+
+  /** Validates that there are no empty options. Returns true if all fields are valid.  */
+  private validateMultiOptionQuestionOptions(options: HTMLInputElement[]) {
+    let multiOptionIsValid = true;
+    for (const option of options) {
+      const inputIsValid = option.value !== '';
+      this.updateFieldErrorState(option, AdminValidationController.MULTI_OPTION_QUESTION_OPTION_CLASS, inputIsValid);
```
bion, line 44: suggestion (blocking): the only usage of `MULTI_OPTION_QUESTION_OPTION_CLASS` concatenates it with `"-error"`, recommend adding that bit to the constant and removing the concatenation

browser-test/src/support/admin_questions.ts
```diff
@@ -308,28 +331,46 @@ export class AdminQuestions {
...
+        await this.page.fill(`:nth-match(#question-settings div.flex-row, ${matchIndex}) input`, options[index]);
+      }
+      
+      await this.clickSubmitButtonAndNavigate('Create');
+  }
 
-    await this.clickSubmitButtonAndNavigate('Create');
+  /** Changes the input field of a multi option answer. */
+  async changeMultiOptionAnswer(index: number, text: string) {
+    await this.page.fill(`:nth-match(#question-settings div.flex-row, ${index}) input`, text);
```
bion, line 361: suggestion (blocking): here and L353 uses the same selector logic which could be abstracted. more importantly: using presentational CSS classes in the selector, in this case `flex-row` is brittle to superficial changes in the HTML. if you need more specificity please add class(es) with semantic name(s).

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
bion: That works if you're super confident that `element` will always be the element you're expecting. Otherwise could add some assertions that throw descriptive errors if it's not.

browser-test/src/support/admin_questions.ts
dkatzz: Good catch! Updated.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
dkatzz: Fixed.

browser-test/src/support/admin_questions.ts
dkatzz: Fixed.

universal-application-tool-0.0.1/app/assets/javascripts/admin_validation.ts
dkatzz: Fixed.

browser-test/src/support/admin_questions.ts
dkatzz: Oops, updated "createDropdownQuestion" instead. This function already existed and there are a lot similar to it, so I refactored the other functions as well, but put it in a separate PR because of how many of these function types there are - https://github.com/seattle-uat/civiform/pull/1632.  We can merge that into this branch or merge it into main after these changes go in.

browser-test/src/support/admin_questions.ts
```diff
@@ -41,6 +50,20 @@ export class AdminQuestions {
     await this.expectAdminQuestionsPageWithSuccessToast('created');
   }
 
+  async expectMultiOptionBlankOptionError(options: String[]) {  
+    const questionSettings = await this.page.$("#question-settings");
+    const errors = await questionSettings.$$('.cf-multi-option-input-error');
+    // Checks that the error is not hidden when it's corresponding option is empty. The order of the options array corresponds to the order of the errors array.
+    for(var i in errors){
```
bion, line 57:
```suggestion
    for(let i in errors) {
```

browser-test/src/support/admin_questions.ts
```diff
@@ -41,6 +50,20 @@ export class AdminQuestions {
...
 
+  async expectMultiOptionBlankOptionError(options: String[]) {  
+    const questionSettings = await this.page.$("#question-settings");
+    const errors = await questionSettings.$$('.cf-multi-option-input-error');
+    // Checks that the error is not hidden when it's corresponding option is empty. The order of the options array corresponds to the order of the errors array.
+    for(var i in errors){
+      if(options[i] === ""){
+        expect(await errors[i].isHidden()).toEqual(false);
+      }
+      else{
```
bion, line 61:
```suggestion
      if (options[i] === "") {
        expect(await errors[i].isHidden()).toEqual(false);
      } else {

```

browser-test/src/support/admin_questions.ts
```diff
@@ -308,28 +331,46 @@ export class AdminQuestions {
...
-      var matchIndex = Number(index) + 1;
-      await this.page.fill(`:nth-match(#question-settings div.flex-row, ${matchIndex}) input`, options[index]);
-    }
+      waitForPageJsLoad(this.page);
+  
+      await this.fillInQuestionBasics(questionName, description, questionText, helpText, enumeratorName);
+  
+      for (var index in options) {
+        await this.page.click('#add-new-option');
+        var matchIndex = Number(index) + 1;
```
bion, line 352:
```suggestion
      for (let index in options) {
        await this.page.click('#add-new-option');
        let matchIndex = Number(index) + 1;
```

bion (approved):

PR #1634, opened 2021-10-05
WIP: Fix deep links for guests

### LINKED ISSUE #846: Support deep-linking directly into a specific program application

bion:
Fixes #846

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantInformationController.java
+++ b/universal-application-tool-0.0.1/app/controllers/applicant/ApplicantInformationController.java
@@ -7,6 +7,7 @@ import com.google.common.collect.ImmutableSet;
 import controllers.CiviFormController;
 import forms.ApplicantInformationForm;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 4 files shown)


browser-test/src/support/admin_questions.ts
```diff
@@ -1,563 +1,563 @@
-import { Page } from 'playwright'
```
bion, line 1: This file contained a mix of different line ending characters, suggesting that it probably received edits from someone on Windows with a text editor that didn't respect Unix text encoding. I re-encoded the file in UTF-8 with Unix-style line endings to make it consistent. Result is a big diff though.

browser-test/src/support/index.ts
```diff
@@ -1,105 +1,107 @@
-import { Browser, chromium, Page } from 'playwright'
```
bion, line 1: This file contained a mix of different line ending characters, suggesting that it probably received edits from someone on Windows with a text editor that didn't respect Unix text encoding. I re-encoded the file in UTF-8 with Unix-style line endings to make it consistent. Result is a big diff though.

browser-test/src/support/admin_questions.ts
bion: I probably should have done this in a separate PR since there is a small code diff in here 😞 .

browser-test/src/support/index.ts
bion: I probably should have done this in a separate PR since there is a small code diff in here 😞 .

PR #1642, opened 2021-10-11
Include applications from all program versions in CSV

### LINKED ISSUE #1586: Let a program admin download all applications regardless of program version
Currently, when downloading applications, a program admin must download the applications data separately for each version of the program.  If an edit was made to the program partway through, this makes it hard for the admin to consolidate all the info, and it leads to the error prone behavior of missing some past or present applications

Allow the ability for a program admin to download all applications for a given program, not just the ones for the specified version.  Alternately, the program admin could select which all version's data to include when they are doing the download.  This will involve changes on the BE as well as the FE.

**Additional context**
This is a real issue and source of confusion for users/partners. I.e. "Where are all the other applications? I'm not seeing any"
It is a usability add that will add value to our partners and streamline their processes.

bion:
fixes #1586

For program admins, the "Download all (CSV)" button when looking at the index page for a given program's applications would previously only download the applications for the version of the program the admin was currently looking at.

This PR changes the text of the button to "Download all versions (CSV)" and the behavior of the download to include all applications from all versions of the program.

image

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
+++ b/universal-application-tool-0.0.1/app/controllers/admin/AdminApplicationController.java
@@ -81,10 +81,9 @@ public class AdminApplicationController extends CiviFormController {
   }
 
   /**
-   * Download a CSV file containing all applications to the specified program version.
-   * This was the original behavior for the program admin CSV download but is currently unused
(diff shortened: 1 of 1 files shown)


bion:
@lindsayzad done, see updated description. Thanks for the suggestion!

Issue #1623, opened 2021-09-22
Entering a decimal in a number question type erases the next number


elisekalstad:
**Describe the bug**
On the UDP form, we ask residents to enter in their monthly income. If the resident types a decimal point, the decimal point will be erased when the next digit is entered. It can lead to a resident accidentally submitting a form with monthly income of 10000 instead of 1000.

**To Reproduce**
1. Go to TEST UDP program application
2. Enter monthly income
3. Use a decimal point
4. Next number typed will not appear
5. Decimal point will be erased

**Expected behavior**
Decimal points should not erase numbers nor disappear.'

**Screenshots**
Attached
![Decimal erasing](https://user-images.githubusercontent.com/86082659/134432510-d9c755a8-cec0-4321-8b56-847451eb74d7.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

lindsayzad:
This is related to #1573 and #1433 ; see those issues for further context

shanemc-goog:
Currently the number field is required to be an Integer and is massaged to be or it's erased.

I'm not sure why the '.' is retained, but when the first decimal is typed the double is massaged into an int and replaces the field content.

JS validation:
https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java#L212

elisekalstad:
@shanemc-goog -- Hi Shane--the UDP team is curious about the lift on this one. Is this something that could be fixed soon? Just wondering what I can report back to them. Thank you!!!

shanemc-goog:
We'll address this issue by implementing the Currency type (go/uat-question-requirements) as tracked in https://github.com/seattle-uat/civiform/issues/450

shanemc-goog:
We shouldn't separately figure out why you can type a decimal-point into the existing Number field as its extraneous and I believe the current validation code is designed to not allow it.

PR #1647, opened 2021-10-19
fix(1433): prevent default behavior for `e`, `+`, and `-`  in number input

### LINKED ISSUE #1433: In number input, entering `e`, `-`, or `+` wipes out the input field
In #1413, we fixed #1391, which caused an error when the user entered a value that ends in `.0` or `.00` (etc.) in a numeric input field.

I was just testing out the new numeric input fields. The desired behavior, it appears, is that any character besides a digit will be "banned" from input. 

If I type | Then I see
--- | ---
`123` | `123`
`abc123` | `123`
`123abc` | `123`
`12!@#$%^&*()3` | `123`
`12[]3` | `123`

These all behave as expected. But certain math-y symbols cause the entire input field to be wiped out!

If I type | Then I see
--- | ---
`12d3` | `123`
`12e3` | `3`
`12+3` | `3`
`12-3` | `3`

In the case of `e`, `+`, or `-`, I would expect those inputs to be silently banned and the numbers in the field to stick around... but instead the numbers I've already entered vanish as soon as I press the offending keys!

My guess is that it has something to do with the way we parse numbers: `-` is used to label negative numbers, `+` is used to coerce  values into JavaScript, `e` is used to describe powers of ten (`5e3 === 5000`). We may need to have a more sophisticated parsing method.

shubha-rajan:
### Description
Added an `onkeydown` event handler for number inputs that prevents typing characters like `e`, `+` and `-`, which are allowed in number inputs, but cause the built-in [value sanitization algorithm](https://html.spec.whatwg.org/multipage/input.html#number-state-(type=number)) to get NaN when parsing a string like "1e", and then set the value to an empty string. Also added a regex filter to block any input that isn't a number or decimal point, so that if a user were to copy and paste a number containing those characters, it would still be filtered out.

### Checklist

### Issue(s)
Fixes #1433


shanemc-goog:
FWIW There's other issues with number inputs, such that I think the better solution is a pre-submit check rather than managing every key stroke.  I'm guessing key stroke limiting is not great for accessibility either. As far as I know this is the only situation we manage the key strokes.

https://github.com/seattle-uat/civiform/issues/1623

shubha-rajan:
> FWIW There's other issues with number inputs, such that I think the better solution is a pre-submit check rather than managing every key stroke. I'm guessing key stroke limiting is not great for accessibility either. As far as I know this is the only situation we manage the key strokes.
> 
> #1623

I can move this code back to `validation.ts` and turn it into a pre-submit check if that's the better option. My goal here was to try to keep the existing behavior, which blocked non-numeric characters from the input.

bion (approved):

Issue #1665, opened 2021-11-01
Program Admin unable to get access to completed applications


elisekalstad:
**Describe the bug**
Thor is a program admin for the UDP program. He is signed up to receive email notifications when there is a new application. When he clicks on the link to open the application though, he gets a 403 Forbidden Error.

He is also unable to view the UDP program by logging in as an admin. 

I tried to remove his email from the list of program admins and re-add him. No luck.

He IS able to view everything on the TEST_Program. The only difference there is that his email is Thor.Peterson@seattle.gov vs. no capitalization thor.peterson@seattle.gov. I tried his email with capitalization and it saved without capitalization.

**To Reproduce**
1. Thor receives an email notification about a new applicant 
2. Click on the link to the application
3. See 403 Forbidden Error

4. Log in as an "Admin"
5. Unable to see the UDP program in the program list

6. CiviForm Admin adds Thor's email address (Thor.Peterson@seattle.gov)
7. Select Save
8. See that email reverts to thor.peterson@seattle.gov

**Expected behavior**
Thor should be able to view applications from the emailed link.
Thor should be able to view the UDP program when logging in as an "Admin".
![Program Admin - Error](https://user-images.githubusercontent.com/86082659/139720150-85bfe22f-80ac-46cb-ad80-5c0e5ecd4567.png)

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

ettengererin:
This bug is being tracked in https://github.com/seattle-uat/civiform/issues/1612

Issue #1612, opened 2021-09-13
Program Admins unable to view program/applications


elisekalstad:
**Describe the bug**
There are three program admins who are unable to view anything when logging in. They are able to login to CiviForm as program admins, but there aren't any programs for them to view. It's only happening to a small number of users--everyone else seems to be able to access their forms just fine. They are logged into VPN, tried it out on several browsers (Chrome and Edge) with the same result. I tried removing them and re-adding them--still the same result.
![Error--viewing programs and applications](https://user-images.githubusercontent.com/86082659/133146551-ba6cf425-5a5d-4c71-824f-f3f407a3824c.png)

Users: thor.peterson@seattle.gov; jen.chawlibog@seattle.gov; maria.ponce@seattle.gov

**To Reproduce**
1. Go to "Admin Login" (after logging into VPN_
2. See error -- i.e. no programs listed

**Expected behavior**
Program admins should be able to view their program and all associated applications

**Screenshots**
Attached

**Desktop (please complete the following information):**
-Tried on Chrome and Edge.

bion:
@elisekalstad could you please take a screenshot of the CiviForm Admin page that shows these three users have been configured as admins of at least one program?

bion:
@elisekalstad Also, if you add these same three users as admins to a program on staging, are they able to login and see it?

elisekalstad:
Hi @bion -- Thanks for taking a look! Here is the screenshot of the admin page, attached:
![Program Admins - Test Program](https://user-images.githubusercontent.com/86082659/133167670-b19be3a0-4ec1-4f7c-99ef-f147c430c2f7.JPG)

elisekalstad:
> @elisekalstad Also, if you add these same three users as admins to a program on staging, are they able to login and see it?

Ohh! Let me try! Hang on!

elisekalstad:
@bion I had Thor try it out (feel free to reach out, if it's easier to talk to him directly: thor.peterson@seattle.gov).

It worked great in staging! He went to the staging site >> selected "Admin Login" in the bottom right. And was able to view the program I assigned him to as well as the applications and CSV download. (pic attached)
![Program Admin - staging Success](https://user-images.githubusercontent.com/86082659/133169004-5f48b079-5f39-4c6a-ac01-cd61f962ad75.JPG)
[...]

PR #1674, opened 2021-11-04
Make email addresses case sensitive when promoting accounts to Program Admin Role

ettengererin:
### Description
This change reverts the change in https://github.com/seattle-uat/civiform/pull/1641. 

This will allow Seattle city admins to be added as program admins and they will be case sensitive. This means that Seattle city admins will have to be added with title case.

This fix is a step in the right direction of figuring out how Civiform will handle case sensitivity for accounts. We still need to make a larger design decision about how Civiform will handle case sensitivity for accounts. Until then, this fix will do!

### Checklist
- [X] Created tests which fail without the change (if possible) -- removed tests

### Issue(s)
Addresses #1612, #1665

### DIFF
--- a/universal-application-tool-0.0.1/test/services/role/RoleServiceTest.java
+++ b/universal-application-tool-0.0.1/test/services/role/RoleServiceTest.java
@@ -53,6 +53,38 @@ public class RoleServiceTest extends WithPostgresContainer {
     assertThat(account2.getAdministeredProgramNames()).containsOnly(programName);
   }
 
+    @Test
+    public void makeProgramAdmins_emailsAreCaseSensitive() throws ProgramNotFoundException {
+    String emailUpperCase = "Fake.Person@email.com";
+    String emailLowerCase = "fake.person@email.com";
+    Account account = new Account();
+    account.setEmailAddress(emailUpperCase);
+    account.save();
+
+    String programName = "test program";
+    Program program = ProgramBuilder.newDraftProgram(programName).build();
+
+    // Make the lower case email a program admin.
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/test/services/role/RoleServiceTest.java
```diff
@@ -53,49 +53,6 @@ public void makeProgramAdmins_allPromoted() throws ProgramNotFoundException {
     assertThat(account2.getAdministeredProgramNames()).containsOnly(programName);
   }
 
-  @Test
```
lindsayzad, line 56: Can we add a test that verifies that both of the emails are added as program admins?

dkatzz (approved):

universal-application-tool-0.0.1/test/services/role/RoleServiceTest.java
ettengererin: Not sure I understand what case that is covering. Can you explain what the purpose of that test would be?

If we add both emails as program admins and then look them up separately, we're not really proving anything -- that test would probably also pass for case-insensitive logic. Are you suggesting that I add a test to check case sensitivity? If so, we should probably add a test that saves an account with an upper case email, and then looks up the account with the lower case email, and expect the found account to not have any programs. 

What do you think?

bion (approved):

universal-application-tool-0.0.1/test/services/role/RoleServiceTest.java
ettengererin: Added a test to check the lower case and upper case email admin permissions!

PR #1677, opened 2021-11-08
Add previous button to programs

### LINKED ISSUE #1650: Add previous button to application pages
Currently, each page of the application has a "next" and "review" button, but no previous button. This means, there is no way to navigate back to a static question, since those won't appear on the review page. 

We should add a "previous" button, so people can navigate to each page of the application, including static questions. See button spec here: https://www.figma.com/file/LeVGh2ONW3UJg8xZea4vUA/CiviForm_UDP_UXbugs?node-id=0%3A1

dkatzz:
### Description
Add previous button, so applicants can navigate to the previous page without going through the review page. 

### Checklist

### Issue(s)
Fixes #1650

Screen Shot 2021-11-08 at 4 46 07 PM

### DIFF
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -88,4 +88,6 @@ describe('applicant navigation flow', () => {
     })
   });
 
+  // TODO: Add tests for "next" navigation
+
 })
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -1,0 +1,88 @@
+import {AdminPrograms, AdminQuestions, ApplicantQuestions, loginAsAdmin, loginAsGuest, logout, selectApplicantLanguage, startSession, resetSession} from './support'
+
+describe('applicant navigation flow', () => {
+  let pageObject;
+
+  beforeAll(async () => {
+    const {page} = await startSession()
+    pageObject = page;
+  });
+
+  afterEach(async () => {
+    await resetSession(pageObject);
+  });
+
+  describe('navigation with four blocks', () => {
+    let applicantQuestions;
+    const programName = 'test program for navigation flows';
+
+    beforeAll(async () => {
+      await loginAsAdmin(pageObject);
+      const adminQuestions = new AdminQuestions(pageObject);
+      const adminPrograms = new AdminPrograms(pageObject);
+      applicantQuestions = new ApplicantQuestions(pageObject);
+
+      await adminQuestions.addDateQuestion({questionName: 'nav-date-q'});
+      await adminQuestions.addEmailQuestion({questionName: 'nav-email-q'});
+      await adminQuestions.addAddressQuestion({questionName: 'nav-address-q'});
+      await adminQuestions.addRadioButtonQuestion({questionName: 'nav-radio-q', options: ['one', 'two', 'three']});
+      await adminQuestions.addStaticQuestion({questionName: 'nav-static-q'});
+
+      await adminPrograms.addProgram(programName);
+      await adminPrograms.editProgramBlock(programName, 'first description', ['nav-date-q', 'nav-email-q']);
+      await adminPrograms.addProgramBlock(programName, 'second description', ['nav-static-q']);
+      await adminPrograms.addProgramBlock(programName, 'third description', ['nav-address-q']);
+      await adminPrograms.addProgramBlock(programName, 'fourth description', ['nav-radio-q']);
+
+      await adminPrograms.gotoAdminProgramsPage();
+      await adminPrograms.publishProgram(programName);
+
+      await logout(pageObject);
+    });
+
+    it('clicking previous on first block goes to summary page', async () => {
+      await loginAsGuest(pageObject);
+      await selectApplicantLanguage(pageObject, 'English');
+
+      await applicantQuestions.applyProgram(programName);
+      await applicantQuestions.clickPrevious();
+
+      // Assert that we're on the preview page.
+      expect(await pageObject.innerText('h1')).toContain('Program application review');
+    });
+
+    it('clicking previous on later blocks goes to previous blocks', async () => {
+      await loginAsGuest(pageObject);
+      await selectApplicantLanguage(pageObject, 'English');
+
+      await applicantQuestions.applyProgram(programName);
+
+      // Fill out the first block and click next
+      await applicantQuestions.answerDateQuestion('2021-11-01');
+      await applicantQuestions.answerEmailQuestion('test1@gmail.com');
+      await applicantQuestions.clickNext();
+
+      // Nothing to fill in since this is the static question block
+      await applicantQuestions.clickNext();
+
+      // Fill out address question and click next
+      await applicantQuestions.answerAddressQuestion('1234 St', 'Unit B', 'Sim', 'Ames', '54321');
+      await applicantQuestions.clickNext();
+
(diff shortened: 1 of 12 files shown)


ettengererin:
UX question/suggestion - Would it look better if the "Previous" button was on the left side of the screen? Just an idea, might make the buttons look less cluttered. Did Tanya suggest this current layout?

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramBlocksControllerTest.java
```diff
@@ -117,6 +117,22 @@ public void edit_withMessages_returnsCorrectButtonText() {
     assertThat(contentAsString(result)).contains("Próximo");
   }
 
+  @Test
+  public void previous_toAnExistingBlock_rendersTheBlock() {
+
+    Request request =
```
ettengererin, line 123: nit: remove empty line above

dkatzz:
> UX question/suggestion - Would it look better if the "Previous" button was on the left side of the screen? Just an idea, might make the buttons look less cluttered. Did Tanya suggest this current layout?

dkatzz:
@ettengererin [Here](https://www.figma.com/file/LeVGh2ONW3UJg8xZea4vUA/CiviForm_UDP_UXbugs?node-id=0%3A1) are the mocks I followed, but I can see what you're saying. We can see what Tanya thinks.

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -159,6 +167,58 @@ public ApplicantProgramBlocksController(
             });
   }
 
+  @Secure
+  public CompletionStage<Result> goToPreviousPage(
```
bion, line 171: Question (blocking): why have both this method and `previous`?

browser-test/src/application_navigation.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      // Click previous and see static question page
+      await applicantQuestions.clickPrevious();
+      await applicantQuestions.seeStaticQuestion('static question text');
+
+      // Click previous and see date and name questions
+      await applicantQuestions.clickPrevious();
+      await applicantQuestions.checkDateQuestionValue('2021-11-01');
+      await applicantQuestions.checkEmailQuestionValue('test1@gmail.com');
+
```
ettengererin, line 84: Is there a point where we don't see the "Previous" button anymore? If so, we should add a check to expect that we don't see it (whenever that is, i'd assume on the first block).

browser-test/src/application_navigation.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  beforeAll(async () => {
+    const {page} = await startSession()
+    pageObject = page;
+  });
+
+  afterEach(async () => {
+    await resetSession(pageObject);
+  });
+
+  describe('navigation with four blocks', () => {
```
ettengererin, line 15: Since this is now the test file that tests navigating through an application, we should also add tests to ensure that "Next" button works. This would be a good check to add anyway since we're introducing the "Previous" logic, we should ensure that "next" still works. 

You can add a separate test for this, or just check it within your existing tests (i.e. add checks after each .clickNext()). If you choose the latter, remember to modify the test names.

ettengererin:
@dkatzz re: mocks -- those looks good to me, no need to change it since this was Tanya's suggestion.

browser-test/src/application_navigation.test.ts
ettengererin: Or...can we check that the final "Previous" button brings you back to the initial Preview page?

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -274,6 +275,27 @@ private Tag renderReviewButton(Params params) {
...
+    int previousBlockIndex = params.blockIndex() - 1;
+    String redirectUrl;
+
+    if (previousBlockIndex >= 0) {
+      redirectUrl =
+          routes.ApplicantProgramBlocksController.previous(
+                  params.applicantId(), params.programId(), previousBlockIndex, params.inReview())
+              .url();
+    } else {
+      redirectUrl =
```
ettengererin, line 288: My understanding is that this should be directing to the Preview page. Can you help me understand how this brings the applicant to the preview page rather than the final review page?

ettengererin (commented):
Super nice work on this, Danielle! This is a hefty change, and it's looking great.

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
dkatzz: Good catch - meant to clean this up! Consolidated into `previous`.

universal-application-tool-0.0.1/test/controllers/applicant/ApplicantProgramBlocksControllerTest.java
dkatzz: Done.

browser-test/src/application_navigation.test.ts
dkatzz: The first test in this file actually tests that the first previous brings you to the review page. Added here too though.

browser-test/src/application_navigation.test.ts
dkatzz: Yeah, I'll add this as a follow-up if that's ok?

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java
dkatzz: Ah yes - it should have been "preview", thanks for calling this out! Fixed.

browser-test/src/application_navigation.test.ts
ettengererin: Sure! Perhaps add a TODO to this file to keep track.

ettengererin (approved):

browser-test/src/application_navigation.test.ts
dkatzz: Good call, added!

dkatzz:
@elisekalstad how should we handle translations for the "Previous" button? Is there a process we can get kicked off to get the translations for the other languages? Once we have those, I can add those in before submitting.

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -159,6 +167,58 @@ public ApplicantProgramBlocksController(
...
+    CompletionStage<String> applicantStage = this.applicantService.getName(applicantId);
+
+    return applicantStage
+        .thenComposeAsync(
+            v -> checkApplicantAuthorization(profileUtils, request, applicantId),
+            httpExecutionContext.current())
+        .thenComposeAsync(
+            v -> applicantService.getReadOnlyApplicantProgramService(applicantId, programId),
+            httpExecutionContext.current())
+        .thenApplyAsync(
```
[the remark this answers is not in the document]
dkatzz: Updated!

bion (approved):

PR #1680, opened 2021-11-10
Fix localstack failure

dkatzz:
### Description
Localstack used to say services were "running", but they have updated to use "available" instead. We debugged by adding  logs to the healthy() check in the wait script and saw that all services were appearing as "available".  Tested and this fixes the browser test failures. 

Paired with @bion 

### Checklist


bion (approved):

shanemc-goog:
FWIW My local setup needs "running" still.  I'll look more FYI if others are seeing an issue.

PR #1683, opened 2021-11-11
Push to docker hub

bion:
Logging in to docker hub isn't working, not sure why yet. This fixes CI for the time being.


bin/build-dev
```diff
@@ -18,8 +18,8 @@ docker tag civiform-dev:latest public.ecr.aws/t1q6b4h2/civiform-dev:latest
 docker push public.ecr.aws/t1q6b4h2/civiform-dev:latest
 
 # log the docker CLI into the Docker Hub container registry
```
shanemc-goog, line 20: Add a header comment that this is a Work In Progress and maybe why so, such that random eyes have some understanding :)

shanemc-goog (approved):

bin/build-dev
bion: done

PR #1703, opened 2021-11-15
Add validation when managing program admins as a global admin

### LINKED ISSUE #1612: Program Admins unable to view program/applications

### LINKED ISSUE #1665: Program Admin unable to get access to completed applications

ettengererin:
### Description
The goal of this change is to make it more obvious when an admin adds an unintended/incorrect email address as Program admin for a given program. Specifically, this addresses case sensitivity issues experienced by Seattle.

When a global admin wants to add or remove Program admins for a given program, they do so by adding email addresses on the Manage Admins page.

### Behavior before this change: 
- If the admin added an email that does not have an account, Civiform would create an account for that email. 

### With this change:
- Civiform no longer creates an account for an email that does not exist.
- If a global admin adds an email that does not exist, the email will not be saved, and an error toast appears on the page saying that "xyz@email.com does not have an account and could not be added as a Program Admin".

Error toast:
image

### Checklist

### Issue(s)
Fixes #1612; Fixes #1665


universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
```diff
@@ -86,8 +87,24 @@ public Result update(Http.Request request, long programId) {
...
+          ImmutableList<String> programAdmins =
+              programRepository.getProgramAdministrators(programId).stream()
+                  .map(Account::getEmailAddress)
+                  .collect(toImmutableList());
+          return ok(
+              manageAdminsView.render(
+                  request,
+                  program.get().getProgramDefinition(),
+                  programAdmins,
+                  maybeError.get().message()));
```
ettengererin, line 106: This return statement is changing the behavior of the page after there's an error.

If there's an error saving a program admin (the account doesn't exist or the account is already a global admin), the page will stay on the Manage Admins page and show the error. Previously, it would redirect to the "All programs" page and show the error. I think it makes more sense to show the error and **stay** on the manage admins page -- this makes it clearer to the global admin that there was an issue/something wasn't saved that they intended to save.

@Reviewers feel free to let me know if you have differing opinions on this.

bion: Nice work, definitely an improvement in behavior

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
```diff
@@ -86,8 +87,24 @@ public Result update(Http.Request request, long programId) {
           roleService.makeProgramAdmins(
               programId, ImmutableSet.copyOf(manageAdminForm.getAdminEmails()));
       Result result = redirect(routes.AdminProgramController.index());
-      if (maybeError.isPresent()) {
-        return result.flashing("error", maybeError.get().message());
+
+      if (maybeError.isPresent()){
```
bion, line 91: nit: change to `if (!maybeError.isPresent())` with an early return so that the bulk of the new code isn't at an additional level of indention. also avoids nested `if`s (even though cyclomatic complexity is identical)

universal-application-tool-0.0.1/app/views/admin/programs/ManageProgramAdminsView.java
```diff
@@ -47,7 +48,10 @@ public ManageProgramAdminsView(AdminLayout layout) {
 
   /** Display a form with a list of inputs for adding and removing admins. */
   public Content render(
-      Http.Request request, ProgramDefinition program, ImmutableList<String> existingAdminEmails) {
+      Http.Request request,
+      ProgramDefinition program,
+      ImmutableList<String> existingAdminEmails,
+      String message) {
```
bion, line 54: suggestion (blocking): IMO the intention of calling render would be more clear from the call site if you overloaded the method, leaving the original call site (ProgramAdminManagementController.java at L63) unchanged and changed the signature for this version of the method to use `Optional<String> message`, making it more obvious that the message is not always expected to be present.

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
```diff
@@ -86,8 +87,24 @@ public Result update(Http.Request request, long programId) {
...
+        Optional<Program> program =
+            programRepository.lookupProgram(programId).toCompletableFuture().join();
+        if (program.isEmpty()) {
+          return notFound(String.format("Program with ID %s was not found", programId));
+        } else {
+          ImmutableList<String> programAdmins =
+              programRepository.getProgramAdministrators(programId).stream()
+                  .map(Account::getEmailAddress)
+                  .collect(toImmutableList());
+          return ok(
```
bion, line 101: nit: empty line before `return` statement and rest of method body

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
```diff
@@ -86,8 +87,24 @@ public Result update(Http.Request request, long programId) {
           roleService.makeProgramAdmins(
               programId, ImmutableSet.copyOf(manageAdminForm.getAdminEmails()));
       Result result = redirect(routes.AdminProgramController.index());
-      if (maybeError.isPresent()) {
-        return result.flashing("error", maybeError.get().message());
+
+      if (maybeError.isPresent()){   
+        Optional<Program> program =
+            programRepository.lookupProgram(programId).toCompletableFuture().join();
+        if (program.isEmpty()) {
```
bion, line 94: nit: empty line separating the multi-line `if` from its peer method contents

universal-application-tool-0.0.1/app/services/role/RoleService.java
```diff
@@ -63,14 +63,17 @@ public RoleService(ProgramService programRepository, UserRepository userReposito
...
-        });
+    for (String email: accountEmails){
+      if (sysAdminEmails.contains(email)) {
+        invalidEmailBuilder.add(email);
+      } else {
+        Optional<CiviFormError> maybeError = userRepository.addAdministeredProgram(email, program);
+        // If there was an error adding the administered program, return the error. 
+        if (maybeError.isPresent()){
+          return maybeError;
+        }
```
bion, line 74: suggestion (blocking): The behavior here is a little unpredictable -- if the list of emails has valid emails A and B and invalid email I and the list is in the order AIB then A will get updated correctly but B won't due to the early return triggering for "I". Suggest creating a new error that combines messages as needed if there is more than one error and looping through the entire list regardless of validity.

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
ettengererin: Done.

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
ettengererin: Done.

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
ettengererin: Done (i think? let me know if you meant something else)

universal-application-tool-0.0.1/app/views/admin/programs/ManageProgramAdminsView.java
ettengererin: Great point. Implemented the overloading -- let me know if theres a better way to do it.

Question: What is the point of having `Optional<String> message` in this version of the method? If it's an overload method, why does the signature have to be Optional?

bion: It makes it clear that `message` is optional and the behavior of the method will change based on whether it's there or not. That's not readily apparent with a plain `String` and a `if` in the method body to check if it's empty.

universal-application-tool-0.0.1/app/services/role/RoleService.java
ettengererin: Really good point. Made a fix, let me know if this works.

universal-application-tool-0.0.1/app/views/admin/programs/ManageProgramAdminsView.java
ettengererin: That makes sense to me. I'm still unclear on why we'd do this for an overloaded function. My sense is having two versions of the function makes it clear that the String message is not always needed. I would think that the Optional<String> signature would be helpful if we only have one version of the function. I haven't worked with overloaded functions much, so I could be off course here. 

Here's a simple example of how I'm thinking about it...

function printMe(string a) {
    print(a)
}

function printMe(string a, string b) { /// why make string b optional?
   print(a + b);
}

Thanks for your explanations here!

ettengererin: Also let me know if theres a better way to do it (i.e. is it better to have a base function for the shared logic that each render() function calls?)

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
```diff
@@ -86,10 +87,29 @@ public Result update(Http.Request request, long programId) {
...
+                .map(Account::getEmailAddress)
+                .collect(toImmutableList());
+
+        return ok(
+            manageAdminsView.render(
+                request,
+                program.get().getProgramDefinition(),
+                programAdmins,
+                Optional.of(maybeError.get().message())));
+      }
```
bion, line 112: suggestion (non-blocking): could maybe factor this out to reduce duplicated code with `edit(`

bion (approved):

universal-application-tool-0.0.1/app/views/admin/programs/ManageProgramAdminsView.java
bion: Ah, yeah, it's a little awkward with both overloading and the optional. Feel to pick one or the other. Sorry for the churn!

ettengererin: Sounds good! And no worries, Thanks for the suggestions.

universal-application-tool-0.0.1/app/controllers/admin/ProgramAdminManagementController.java
ettengererin: Good idea, done!

PR #1705, opened 2021-11-16
Closes #1698

### LINKED ISSUE #1698: New Program Visibility setting is required but optional in the UI
**Describe the bug**
When making a new Program you can select between having it be Hidden of Public, but neither is selected by default and the UI doesn't require one be selected

If you submit without selecting one, a generic BE error is issued to the FE

**To Reproduce**
1. Create a new Program without selecting either Hidden or Public and submit

**Screenshots**
displaymode

zaralouis-sf:
Closes #1698

### Description
Makes display mode a required field for the front end when an admin creates a new program.

### Change in user behavior
In the past, when an admin tried to add a program without setting the display mode (public or hidden in index), they would be able to hit create, then they would see a generic backend error message. Now, while usesr are still on the program page, they will get an error message "display mode must be set".

Screenshot: 
Civiforrm program fix

### Issue(s)
Fixes #1698

### DIFF
--- a/universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
+++ b/universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
@@ -119,6 +119,7 @@ public class ProgramServiceImpl implements ProgramService {
     validateProgramText(errorsBuilder, "admin description", adminDescription);
     validateProgramText(errorsBuilder, "display name", defaultDisplayName);
     validateProgramText(errorsBuilder, "display description", defaultDisplayDescription);
+    validateProgramText(errorsBuilder, "display mode", displayMode);
     ImmutableSet<CiviFormError> errors = errorsBuilder.build();
     if (!errors.isEmpty()) {
       return ErrorAnd.error(errors);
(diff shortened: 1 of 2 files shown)


bion:
> Closes #1698
Please write a descriptive title for each pull request

bion:
@zaralouis-sf What is the change in behavior from the user's point of view? Please include a screenshot if convenient to do so

universal-application-tool-0.0.1/app/services/program/ProgramServiceImpl.java
```diff
@@ -119,6 +119,7 @@ public ActiveAndDraftPrograms getActiveAndDraftPrograms() {
     validateProgramText(errorsBuilder, "admin description", adminDescription);
     validateProgramText(errorsBuilder, "display name", defaultDisplayName);
     validateProgramText(errorsBuilder, "display description", defaultDisplayDescription);
+    validateProgramText(errorsBuilder, "display mode", displayMode);
```
bion, line 122: suggestion (blocking): please write a unit test that covers this change in behavior. see https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/test/services/program/ProgramServiceImplTest.java

bion (approved):

PR #1706, opened 2021-11-17
fix(1467, 1623): improve number inputs

### LINKED ISSUE #1467: Use `inputmode=decimal` for Number inputs
Apparently, iOS doesn't show a very good number picker for `<input type="number">` fields. To get the number pad we're used to, you have to use:

```html
<input type="number" inputmode="decimal">
```

So let's use that wherever we ask for numeric inputs. [See this article](https://css-tricks.com/better-form-inputs-for-better-mobile-user-experiences/#why-should-i-care-about-inputmode) for more.

### LINKED ISSUE #1623: Entering a decimal in a number question type erases the next number

shubha-rajan:
### Description
A continuation of work in #1647 to improve user experience for number inputs by
- Adding `input=decimal` to use a more accessible keyboard on iOS (#1467 )
- Using built-in html validation of number type to enforce that numbers are integers, and convert any float inputs (if they somehow get past client-side validation) to `Long` on the back end (#1623)
- Validate min and max client-side using built-in html validation before submitting

### Screenshots:
On iOS 12 before adding decimal input:
screenshot-without-decimal-input
On iOS 12 after adding decimal input:
screenshot-with-decimal-input
Validation for integer input: 

Screen Shot 2021-11-12 at 3 59 24 PM

Validation for min/max:
image

### Checklist

### Issue(s)
Fixes #1467; Fixes #1623;


universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
```diff
@@ -376,7 +376,13 @@ private void putParentArray(Path path) {
     try {
       return this.read(path, Long.class);
     } catch (JsonPathTypeMismatchException e) {
-      return Optional.empty();
+      try {
+        // check if value is a double/float and cast to long if needed
```
shanemc-goog, line 380: The Question type definition is to only allow integers, and the FE tries to enforce that, so I think we should have consistency with the backend checking too.

universal-application-tool-0.0.1/app/assets/javascripts/number.ts
```diff
@@ -8,8 +8,9 @@ class NumberController {
       <NodeListOf<HTMLInputElement>>document.querySelectorAll('input[type=number]'));
     numberQuestions.forEach(numberQuestion => {
       numberQuestion.addEventListener('input', () => {
-        // Only allow integer input.
-        const n = parseInt(numberQuestion.value.replace(/[^.\d]/g, ''));
+        // Block non-numeric characters like like 'e', 'E', '+', '-'
+        // which will cause problems when parsing to a number
+        const n = parseFloat(numberQuestion.value.replace(/[^.\d]/g, ''));
```
shanemc-goog, line 13: I think we probably want a validation.ts method to show an error message rather than manipulating user key strokes. This is similar to currency and Address Zip code validation where we validate upon Submit.

shanemc-goog: app/assets/javascripts/validation.ts

shanemc-goog:
I left some high level comments, but in general the Number type is defined as an Integer as it, so we should keep it that way without a broader discussion to redefine it, or differentiate Int vs float.

go/uat-question-requirements -> 
https://docs.google.com/document/d/1HYVHBno6dk7fmS5w6V6Y3jxERHGfDoQKTtzBZr27Y6o/edit#

shubha-rajan:
Updatied this to use the same onSubmit validation method as the other question types in `validation.ts`. Invalid input now looks like this (feel free to suggest a better error message): 
image

This PR still relies on the built-in HTML validation for the number input type to enforce min and max instead of rendering those errors in the error div. Would it be better to write our own client-side validation?

browser-test/src/number_input.test.ts
```diff
@@ -1,9 +1,27 @@
 import {
-  startSession, loginAsAdmin, loginAsTestUser, logout, selectApplicantLanguage, AdminQuestions, AdminPrograms, ApplicantQuestions, endSession
+  startSession, loginAsAdmin, loginAsTestUser, logout, selectApplicantLanguage, AdminQuestions, AdminPrograms, ApplicantQuestions, waitForPageJsLoad, endSession
 } from './support'
 
+describe('input type', () => {
+  it('sets the inputmode attribute to decimal', async () => {
```
shanemc-goog, line 6: I don't think we need this test as it is more of a "change detector" test (Did the attributes change?) where we should be testing functionality as you have below.

browser-test/src/number_input.test.ts
```diff
@@ -27,16 +45,14 @@ describe('input validation for number questions', () => {
     await applicant.applyProgram(programName);
     await applicant.validateHeader('en-US');
 
-    const testValues = [
-      '123', 'abc123', '123abc', '12!@#$%^&*()3', '12[]3', '12d3', '12e3', '12E3', '12+3', '12-3'
-    ]
-    const expectedValue = '123';
+    const testValues = ['12e3', '12E3', '-123']
```
shanemc-goog, line 48: include a decimal value

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -153,6 +156,25 @@ public FieldWithLabel setAttribute(String attribute) {
     return this;
   }
 
+  public FieldWithLabel setMin(OptionalLong value) {
+    if (!this.fieldType.equals("number")) {
+      throw new RuntimeException(
+          "setting an OptionalLong min value is only available on fields of type `number`");
```
shanemc-goog, line 162: Change the tics around number to single quotes, here and below.

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
```diff
@@ -241,6 +263,17 @@ public ContainerTag getContainer() {
       ContainerTag textAreaTag = textarea().withType("text").withText(this.fieldValue);
       fieldTag = textAreaTag;
     } else if (this.fieldType.equals("number")) {
+      // Setting inputmode to decimal gives iOS users a more accessible keyboard
+      fieldTag.attr("inputmode", "decimal");
+
+      // Set min and max values for client-side validation
+      if (this.minValue.isPresent()) {
+        fieldTag.attr("min", minValue.getAsLong());
```
shanemc-goog, line 271: May want to add a simple test that these are set in the output in test/views/components/FieldWithLabelTest.java

shanemc-goog:
> Updated this to use the same onSubmit validation method as the other question types in `validation.ts`. Invalid input now looks like this (feel free to suggest a better error message): image
> 
> This PR still relies on the built-in HTML validation for the number input type to enforce min and max instead of rendering those errors in the error div. Would it be better to write our own client-side validation?

I don't really have an opinion, so whatever looks better to you.

shanemc-goog:
Looking good!

universal-application-tool-0.0.1/conf/messages.en-US
```diff
@@ -71,6 +71,7 @@ validation.textTooShort=Must contain at least {0} characters.
 validation.textTooLong=Must contain at most {0} characters.
 validation.numberTooSmall=Must be at least {0}.
 validation.numberTooBig=Must be at most {0}.
+validation.numberNonInteger=Number must be a positive whole number and can only contain numeric characters 0-9.
```
shanemc-goog, line 74: This needs to be duplicated in conf/messages

browser-test/src/number_input.test.ts
shubha-rajan: removed this and added a test to `FieldWithLabelTest.java` to test for this

browser-test/src/number_input.test.ts
```diff
@@ -27,16 +45,14 @@ describe('input validation for number questions', () => {
...
     const numberInput = 'div.cf-question-number input'
+    const numberInputError = 'div.cf-question-number-error'
 
     for (const testValue of testValues) {
       await page.type(numberInput, testValue);
-      expect([testValue, await page.inputValue(numberInput)])
-        .toStrictEqual([testValue, expectedValue]);
+      await applicant.clickNext();
+      expect([testValue, await page.isHidden(numberInputError)]).toStrictEqual([testValue, false]);
```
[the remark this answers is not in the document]
shubha-rajan: exactly!

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
shubha-rajan: done

universal-application-tool-0.0.1/app/views/components/FieldWithLabel.java
shubha-rajan: done

universal-application-tool-0.0.1/conf/messages.en-US
shubha-rajan: done

universal-application-tool-0.0.1/app/assets/javascripts/number.ts
shanemc-goog: Do we still need this?  All the validation should be a submit not typing time.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
shanemc-goog: This still needs reverting correct?

universal-application-tool-0.0.1/app/assets/javascripts/number.ts
shanemc-goog: I think we still don't need this BTW.

universal-application-tool-0.0.1/app/services/applicant/ApplicantData.java
shanemc-goog: ?

shanemc-goog:
Just wanted to make sure you saw the comments in number.ts and ApplicantData.java

shubha-rajan:
> Just wanted to make sure you saw the comments in number.ts and ApplicantData.java

I reverted the changes in ApplicantData.java and removed number.ts (moved the validation logic to validation.ts instead) so I can't see the comments anymore.

shanemc-goog (approved):

Issue #1711, opened 2021-11-22
Program Admin - Data export shows residents' most recent application


elisekalstad:
**Describe the bug**
When a program admin exports the CSV of all applications, it includes only a resident's most recent application. For example, if a resident applied to a program twice, once on 11/1/2021 and a second time on 11/10/2021, only the 11/10/2021 application will show. The expected behavior is to show all applications that a resident submits. It could be that they needed to "re-apply" but it's helpful to show all applications. The program admin can then filter out older applications on their end, once they've exported the CSV.

**To Reproduce**
1. Login as program admin
2. Export data via CSV
3. Review list of applications
4. See that previous applications are no longer included, if the resident re-submitted a form

**Expected behavior**
CSV should include all applications, not just the most recent one.

**Screenshots**
Attached -- include the data export from program admin's view and what the civiform admin sees.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

elisekalstad:
[Data Export Issues with CiviForm.docx](https://github.com/seattle-uat/civiform/files/7585049/Data.Export.Issues.with.CiviForm.docx)
[Error- Time stamp.xlsx](https://github.com/seattle-uat/civiform/files/7585050/Error-.Time.stamp.xlsx)

elisekalstad:
Upon closer review there are two issues:
1) The "submit date" updates for the applicant's most recent submittal (for CiviForm Admin export AND Program Admin view/export). Would expect it to show the actual submit date for the specific application.
2) Program Admins can only see the most recent application. If a resident re-submits their application, they don't see the original application. Would expect the export to include ALL applications for the resident.

I think the attachments show this pretty well.

dkatzz:
@elisekalstad - program admins should now be able to see the previous submitted applications by an applicant. The "submit date" part is still in progress. Thanks!

PR #1715, opened 2021-11-23
Show previously submitted applications by an applicant

dkatzz:
### Description
Currently, only the most recent application submission is showed to program admins, but they'd like to see all submissions by an applicant. "OBSOLETE" applications are older versions of an application submitted. We will show these too in order to show all submitted applications. There is another issue with submission time getting updated when an application is marked as obsolete, but this will be handled in a separate PR.

### Checklist

### Issue(s)
 #1711


bion (approved):
looks great!

PR #1748, opened 2021-12-10
Cache program definition database lookups to reduce export time.

### LINKED ISSUE #1747: CSV export makes expensive DB call for each row
exportCsv loops over applications to generate their CSV data.

As part of processing each row, the Program definition is looked up which means that every row in the CSV incurs a DB read, and for data that is identical on subsequent reads.  This is costly and cumulatively is exceeding our web server timeouts and a reasonable amount of processing time.

#1730 is an example of this impacting Seattle.

A proposed solution would be to bubble up the Program lookup in [ApplicantServiceImpl.getReadOnlyApplicantProgramService](https://github.com/seattle-uat/civiform/blob/eec281490e869720c262885a81240aff51f41d00/universal-application-tool-0.0.1/app/services/applicant/ApplicantServiceImpl.java#L117)

```
programService.getProgramDefinition(application.getProgram().id),
```

To it's caller in [exportCsv](https://github.com/seattle-uat/civiform/blob/eec281490e869720c262885a81240aff51f41d00/universal-application-tool-0.0.1/app/services/export/ExporterService.java#L154), cache the value and pass it down to getReadOnlyApplicantProgramService

shanemc-goog:
Cache program definition database lookups to reduce export time.

This saves most of the per-row processing time.  Locally 4300 rows went from ~60s to 10s to export.

### Description
exportCsv method maintains a cache of Program data so that it's not looked up for every row that pertains to the Program.

### Checklist

### Issue(s)
Fixes #1747

### DIFF
--- a/universal-application-tool-0.0.1/app/services/export/ExporterService.java
+++ b/universal-application-tool-0.0.1/app/services/export/ExporterService.java
@@ -150,12 +150,20 @@ public class ExporterService {
     try {
       OutputStream inMemoryBytes = new ByteArrayOutputStream();
       Writer writer = new OutputStreamWriter(inMemoryBytes, StandardCharsets.UTF_8);
+      HashMap<Long, ProgramDefinition> programDefinitions = new HashMap<>();
       for (Application application : applications) {
+        Long programId = application.getProgram().id;
+        if (!programDefinitions.containsKey(programId)) {
+          try {
+            programDefinitions.put(programId, programService.getProgramDefinition(programId));
+          } catch (ProgramNotFoundException e) {
+            throw new RuntimeException("Cannot find a program that has applications for it.", e);
+          }
+        }
+        ProgramDefinition programDefinition = programDefinitions.get(programId);
+
         ReadOnlyApplicantProgramService roApplicantService =
-            applicantService
-                .getReadOnlyApplicantProgramService(application)
-                .toCompletableFuture()
-                .join();
+                applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
         csvExporter.export(application, roApplicantService, writer);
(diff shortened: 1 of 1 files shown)


universal-application-tool-0.0.1/app/services/export/ExporterService.java
```diff
@@ -150,12 +150,20 @@ public String exportCsv(CsvExporter csvExporter, ImmutableList<Application> appl
...
+      HashMap<Long, ProgramDefinition> programDefinitions = new HashMap<>();
       for (Application application : applications) {
+        Long programId = application.getProgram().id;
+        if (!programDefinitions.containsKey(programId)) {
+          try {
+            programDefinitions.put(programId, programService.getProgramDefinition(programId));
+          } catch (ProgramNotFoundException e) {
+            throw new RuntimeException("Cannot find a program that has applications for it.", e);
+          }
+        }
```
bion, line 162: suggestion (non-blocking): load all programs in a single query rather than one query per program or leave a TODO and file an issue for doing that in the future.

Loading all of the programs needed for the export with a single DB query would be an improvement over loading them individually. That being said that query doesn't exist yet on `ProgramRepository` (though very similar queries do) and this is time sensitive.

bion (approved):

universal-application-tool-0.0.1/app/services/export/ExporterService.java
shanemc-goog: I'll do a TODO for that.

shanemc-goog: TODOs for single lookup and Javacache

Issue #1686, opened 2021-11-11
File uploading supports Azure backend


bion:
Uploading, viewing, and generating expiring access links for user-provided files currently relies on hard-coded calls to the AWS S3 SDK. Those calls should be abstracted to switch at runtime based on a configuration value so Azure file storage can be used instead.

Done when: CiviForm can use Azure as a file storage backend

PR #1752, opened 2021-12-15
Azure storage dev upload functionality

### LINKED ISSUE #1763: Update Azure dev codebase to use NPM instead of a 3rd party CDN
Blocking #1686 
**Describe the bug**
Our dev Azure file upload site is relying on a third party CDN for importing Azure blob storage. We want to remove this dependency and import Azure blob storage the package manager instead, but this will require updating how we're importing assets into the play framework.

zaralouis-sf:
### Description
Begin work on file upload for Azure storage. 

This PR
- creates an Azure storage client for prod, dev (Azurite) and unit testing. 
-  abstracts away the logic of file uploads from hardcoded AWS to an interface that can use AWS or Azure, depending on application.conf setting
    - `universal-application-tool-0.0.1/app/modules/CloudStorageModule.java` contains the logic for determining which storage client to use
- enables the dev file upload view to use Azure by adding a Typescript file to handle the file upload logic. 
    - To test, set `STORAGE_SERVICE_NAME` environment variable to `azure-blob`, then run ```bin/run-dev --azure```. 
    
TODOs (not implemented in this PR)
- integrate azure blob storage for ApplicantProgramBlockEditView and ApplicantProgramBlockEditController
- Make CORS rules more secure
- Use the user-set file names instead of UUIDs (we're setting the filename on the backend because the filename is needed to generate a blob for a single blob, and giving container-level access is a bad idea). Maybe look into [Directory Scoped Access](https://docs.microsoft.com/en-us/rest/api/storageservices/create-user-delegation-sas#user-delegation-sas-support-for-directory-scoped-access)
- Configure Azurite emulator to use HTTPS

image

### Checklist

### Issue(s)
Towards #1686
Fixes #1763


.gitignore
```diff
@@ -14,6 +14,7 @@ node_modules
 bundle.js
 tailwind.css
 *.pem
+package-lock.json
```
bion, line 17: question (blocking): what's the intention with ignoring this? there's no top-level package.json file so i'm surprised a package lock file shows up here.

bin/azurite/wait
```diff
@@ -0,0 +1,32 @@
...
+pushd $(git rev-parse --show-toplevel) > /dev/null
+
+set -e
+
+azurite_endpoint=http://localhost:10000
+
+if [ "$1" ]; then
+    azurite_endpoint=$1
+fi
+
```
bion, line 14:
```suggestion

echo "Using azurite endpoint $azurite_endpoint"

```

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,53 @@
...
+      const redirectUrl = new URL(successActionRedirect);
+
+      const blockBlobURL = azblob.BlockBlobURL.fromBlobURL(
+        new azblob.BlobURL(
+          `${blobUrl}?${sasToken}`,
+          azblob.StorageURL.newPipeline(new azblob.AnonymousCredential)
+        ));
+      azblob.uploadBrowserDataToBlockBlob(azblob.Aborter.none, file, blockBlobURL).then((resp, err) => {
+        if (err) {
+          console.log(err);
```
bion, line 39:
```suggestion
          console.error(err);
```

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  getValueFromInputLabel(label: string): string {
+    return (<HTMLInputElement>document.getElementsByName(label)[0]).value;
+  }
+
+  attemptUpload(event: Event, uploadContainer: HTMLElement | null) {
+    event.preventDefault();
+    const azblob = window["azblob"];
+    if (uploadContainer != null) {
```
bion, line 19: suggestion (blocking): invert the conditional and log an error for the null argument

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,53 @@
...
+    uploadContainer.addEventListener("submit", (event) => this.attemptUpload(event, uploadContainer));
+  }
+
+  getValueFromInputLabel(label: string): string {
+    return (<HTMLInputElement>document.getElementsByName(label)[0]).value;
+  }
+
+  attemptUpload(event: Event, uploadContainer: HTMLElement | null) {
+    event.preventDefault();
+    const azblob = window["azblob"];
```
bion, line 18: question (blocking): what sets this value? why not `window.azblob`?

.gitignore
shubha-rajan: I think that's supposed to be `*/package-lock.json`. Zara and I kept accidentally committing the package.json file so we decided to add it.

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
shubha-rajan: So this is set by the Azure javascript client that we're importing by setting a script tag that links to a CDN. And I tried `window.azblob` which worked with regular Javascript, but the typescript compiler didn't like it.

.gitignore
bion: Ah, yeah that's a bit of an annoyance but we do need to have a lock file to ensure dependencies don't change out from under us

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
bion: Gotcha. Please add a comment explaining that

.gitignore
```diff
@@ -14,6 +14,7 @@ node_modules
 bundle.js
 tailwind.css
 *.pem
+*/package-lock.json
```
bion, line 17: suggestion (blocking): please revert. the lock files are important for keeping our dependencies consistent.

.gitignore
shubha-rajan: got it. I can revert it then. in that case, when do we check in the lockfile? are we periodically updating it?

.gitignore
shubha-rajan: done

.gitignore
bion: When we explicitly update a dependency version or add/remove a dependency

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
```diff
@@ -22,43 +22,63 @@
...
-                        div()
-                            .withClasses(Styles.GRID, Styles.GRID_COLS_2)
-                            .with(div().with(h2("Current Files:")).with(pre(renderFiles(files))))));
+    ContainerTag fileUploadForm;
+
+    HtmlBundle bundle = layout.getBundle();
+
+    if (signedRequest.serviceName().equals(StorageServiceName.AZURE_BLOB.getString())) {
+      fileUploadForm = azureBlobFileUploadForm((BlobStorageUploadRequest) signedRequest);
+      bundle.addFooterScripts(viewUtils.makeCdnJsTag(AZURE_STORAGE_BLOB_CDN));
```
bion, line 65: suggestion (blocking): please use the package manager for managing JavaScript dependencies rather than a CDN: https://github.com/Azure/azure-storage-js/blob/master/blob/README.md#npm

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  attemptUpload(event: Event, uploadContainer: HTMLElement | null) {
+    event.preventDefault();
+    // This is set by an imported JavaScript Azure blob client.
+    // window["azblob"] was used because the TS compiler complained about using
+    // window.azblob.
+    const azblob = window["azblob"];
+    if (uploadContainer == null) {
+      console.error("Attempted to upload to null container");
+    } else {
```
bion, line 24: suggestion (blocking): use an early return instead of an `else` or throw an exception instead of printing to the console. either way saves a level of indentation for the bulk of the method body

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,58 @@
...
+      const sasToken = this.getValueFromInputLabel("sasToken");
+      let blobUrl = this.getValueFromInputLabel("blobUrl");
+      if (blobUrl.includes("azurite")) { // TODO: replace this with a regex to detect azurite URLs
+        blobUrl = blobUrl.replace("azurite", "localhost");
+      }
+      const successActionRedirect = this.getValueFromInputLabel("successActionRedirect");
+      const containerName = this.getValueFromInputLabel("containerName");
+      const file = (<HTMLInputElement>uploadContainer.querySelector('input[type=file]')).files[0];
+      let fileName = this.getValueFromInputLabel("fileName");
```
bion, line 33: suggestion (blocking): define a type called `AzureUploadProps` or somesuch and move all logic for collecting/formatting these values into a private method that returns an object of that type

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,58 @@
...
+      const redirectUrl = new URL(successActionRedirect);
+
+      const blockBlobURL = azblob.BlockBlobURL.fromBlobURL(
+          new azblob.BlobURL(
+              `${blobUrl}?${sasToken}`,
+              azblob.StorageURL.newPipeline(new azblob.AnonymousCredential)
+          ));
+      azblob.uploadBrowserDataToBlockBlob(azblob.Aborter.none, file, blockBlobURL).then((resp, err) => {
+        if (err) {
+          console.error(err);
```
bion, line 44: suggestion (blocking): having thought about it for a third time i think i like throwing errors as the right approach here and above

universal-application-tool-0.0.1/app/controllers/dev/FileUploadController.java
```diff
@@ -57,17 +59,30 @@ public Result create(Request request) {
     if (!isDevEnvironment()) {
       return notFound();
     }
-    Optional<String> bucket = request.queryString("bucket");
-    Optional<String> key = request.queryString("key");
     Optional<String> etag = request.queryString("etag");
-    if (!bucket.isPresent() || !key.isPresent()) {
-      return redirect(routes.FileUploadController.index().url());
+    String successMessage;
+    if (storageClient.getStorageServiceName() == StorageServiceName.AWS_S3) {
```
bion, line 64: suggestion (blocking): It looks like the `else` case here is implicitly Azure.

To clean this code up I recommend factoring out a [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) based on the file storage backend with a separate class for each. This will extend nicely when we add GCP.

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.inject.AbstractModule;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigException;
+import play.Environment;
+import services.cloud.StorageClient;
+import services.cloud.StorageServiceName;
+
+public class CloudStorageModule extends AbstractModule {
```
bion, line 12: suggestion (blocking): please add javadoc for the class

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private static final String AZURE_STORAGE_CLASS_NAME = "services.cloud.azure.BlobStorage";
+  private static final String AWS_STORAGE_CLASS_NAME = "services.cloud.aws.SimpleStorage";
+
+  private final Environment environment;
+  private final Config config;
+
+  public CloudStorageModule(Environment environment, Config config) {
+    this.environment = environment;
+    this.config = config;
```
bion, line 22: suggestion (blocking): add `checkNotNull` for both args

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  protected void configure() {
+    // cloud.storage = "azure-blob"
+    // cloud.storage = "s3"
```
bion, line 28: question (blocking): what's the intention of this comment?

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+    String className = AWS_STORAGE_CLASS_NAME;
+    try {
+      final String storageProvider = checkNotNull(config).getString("cloud.storage");
```
bion, line 31:
```suggestion
      String storageProvider = checkNotNull(config).getString("cloud.storage");
```

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+      if (storageProvider.equals(StorageServiceName.AWS_S3.getString())) {
+        className = AWS_STORAGE_CLASS_NAME;
+      } else if (storageProvider.equals(StorageServiceName.AZURE_BLOB.getString())) {
+        className = AZURE_STORAGE_CLASS_NAME;
+      }
+    } catch (ConfigException ex) {
+      // Ignore missing config and default to S3 for now
+    }
```
bion, line 39: suggestion (blocking): refactor to use a `switch (storageProvider)` in a private method called `getStorageProviderClassName`

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      Class<? extends StorageClient> boundClass =
+          environment.classLoader().loadClass(className).asSubclass(StorageClient.class);
+      bind(StorageClient.class).to(boundClass);
```
bion, line 43: praise: great work dealing with a tricky dependency problem!

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
```diff
@@ -0,0 +1,12 @@
+package services.cloud;
+
+import java.net.URL;
+
+public interface StorageClient {
```
bion, line 5: suggestion (blocking): add javadoc for the interface and each method in it

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
```diff
@@ -0,0 +1,12 @@
+package services.cloud;
+
+import java.net.URL;
```
bion, line 3: something to be aware of: `java.net.URL` has some pretty severe gotchas to look out for. In general, never put one in a hash table or use its equality comparison. Either operation triggers a blocking network request to perform DNS resolution. More info here: https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/net/URL.html#equals(java.lang.Object)

universal-application-tool-0.0.1/app/services/cloud/StorageServiceName.java
```diff
@@ -0,0 +1,16 @@
+package services.cloud;
+
+public enum StorageServiceName {
```
bion, line 3: suggestion (blocking): add javadoc for this enum

universal-application-tool-0.0.1/app/services/cloud/StorageUploadRequest.java
```diff
@@ -0,0 +1,6 @@
+package services.cloud;
+
+public interface StorageUploadRequest {
```
bion, line 3: suggestion (blocking): add javadoc for this interface and its method

universal-application-tool-0.0.1/app/services/cloud/aws/SignedS3UploadRequest.java
```diff
@@ -30,7 +32,7 @@
  * SimpleStorage#getSignedUploadRequest}.
  */
 @AutoValue
-public abstract class SignedS3UploadRequest {
+public abstract class SignedS3UploadRequest implements StorageUploadRequest {
```
bion, line 35: praise: nice work driving toward a common interface for these storage backends!

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import play.Environment;
+import services.cloud.StorageClient;
+import services.cloud.StorageServiceName;
+
+/**
+ * SimpleStorage provides methods to create federated links for users of CiviForm to upload and
+ * download files directly to and from AWS Simple Storage Service (S3).
+ */
```
bion, line 31: suggestion (blocking): fix the javadoc

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+@Singleton
+public class BlobStorage implements StorageClient {
+
+  public static final String AZURE_STORAGE_ACCT_CONF_PATH = "azure.blob.account";
+  public static final String AZURE_CONTAINER_CONF_PATH = "azure.blob.container";
+  public static final String AZURE_REGION_CONF_PATH = "java.time.zoneid";
+  public static final Duration AZURE_SAS_TOKEN_DURATION = Duration.ofMinutes(10);
+  public static final Duration AZURE_USER_DELEGATION_KEY_DURATION = Duration.ofMinutes(60);
```
bion, line 39: question (blocking): what does the user delegation key duration do?

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+
+  public Client getClient() {
+    return client;
+  }
+
+  @Override
+  public BlobStorageUploadRequest getSignedUploadRequest(
+      String fileName, String successActionRedirect) {
+    // Azure blob must know the name of a file to generate a SAS for it, so we'll use a UUID
+    // TODO: figure out how to map this to user-uploaded filename
```
bion, line 86: thought (non-blocking): I *think* so long as we store the original filename in the database associated with the uploaded file it's ok if Azure's version of it is named with a UUID

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
bion: also please document all public methods in this class

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+            .setSuccessActionRedirect(successActionRedirect);
+    return builder.build();
+  }
+
+  @Override
+  public StorageServiceName getStorageServiceName() {
+    return StorageServiceName.AZURE_BLOB;
+  }
+
+  interface Client {
```
bion, line 104: suggestion (blocking): add javadoc for interface and its methods

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+      return "sasToken";
+    }
+
+    @Override
+    public String getBlobUrl(String fileName) {
+      return "http://www.blobUrl.com";
+    }
+  }
+
+  class AzureBlobClient implements Client {
```
bion, line 127: suggestion (blocking): add javadoc for class

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Override
+    public String getBlobUrl(String fileName) {
+      BlobClient blobClient =
+          blobServiceClient.getBlobContainerClient(container).getBlobClient(fileName);
+      return blobClient.getBlobUrl();
+    }
+  }
+
+  class AzuriteClient implements Client {
```
bion, line 179: suggestion (blocking): add javadoc for class

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+          new BlobServiceProperties()
+              .setCors(
+                  List.of(
+                      new BlobCorsRule()
+                          // TODO: Make these more specific
+                          .setAllowedOrigins("*")
+                          .setAllowedHeaders("*")
+                          .setExposedHeaders("*")
+                          .setAllowedMethods("GET,PUT,OPTIONS")
+                          .setMaxAgeInSeconds(500)));
```
bion, line 213: question (blocking): will this only be used in dev? if so the cors policy isn't a big deal

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Override
+    public String getSasToken(String fileName) {
+      BlobClient blobClient = blobContainerClient.getBlobClient(fileName);
+
+      BlobSasPermission blobSasPermission =
+          new BlobSasPermission()
+              .setReadPermission(true)
+              .setWritePermission(true)
+              .setListPermission(true);
```
bion, line 225: question (blocking): do we want to grant list permission? seems like a security vulnerability

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,241 @@
...
+
+      BlobServiceSasSignatureValues signatureValues =
+          new BlobServiceSasSignatureValues(
+                  OffsetDateTime.now(zoneId).plus(AZURE_SAS_TOKEN_DURATION), blobSasPermission)
+              .setProtocol(SasProtocol.HTTPS_HTTP); // TODO: Get this to work with HTTPS_ONLY
```
bion, line 230: thought (non-blocking): we can't use HTTPS in dev so this may not be a possible TODO

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
```diff
@@ -0,0 +1,89 @@
+package services.cloud.azure;
+
+import com.google.auto.value.AutoValue;
+import services.cloud.StorageServiceName;
+import services.cloud.StorageUploadRequest;
+
+/** This holds the information needed to upload a blob storage request. */
```
bion, line 7: nit: is it for uploading a request or making a request to upload a file to blob storage?

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    /**
+     * Get file name for the blob. This is used to build the canonicalized resource part of the
+     * signature string.
+     */
+    abstract String fileName();
```
bion, line 38: question (blocking): why are getters on the builder?

universal-application-tool-0.0.1/app/services/cloud/azure/Credentials.java
```diff
@@ -0,0 +1,32 @@
...
+import play.Environment;
+
+/**
+ * This class retrieves Azure credentials through default provider.
+ *
+ * <p>See
+ * https://docs.microsoft.com/en-us/java/api/overview/azure/identity-readme?view=azure-java-stable
+ * for more details.
+ */
+public class Credentials {
```
bion, line 16: question (blocking): is this meant to be a singleton?

universal-application-tool-0.0.1/app/views/ViewUtils.java
```diff
@@ -28,6 +28,11 @@ public Tag makeLocalJsTag(String filename) {
         .withType("text/javascript");
   }
 
+  /** Generates an CDN script tag for loading the javascript file found at that CDN URL. */
+  public Tag makeCdnJsTag(String cdnUrl) {
+    return script().withSrc(cdnUrl).withType("text/javascript");
+  }
```
bion, line 34: suggestion (blocking): please do not use a 3rd party CDN for retrieving assets. It complicates the app's CORS policy and for some civic entities may even be banned.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
```diff
@@ -22,43 +22,63 @@
...
-                    .with(div().with(fileUploadForm(signedRequest)))
-                    .with(
-                        div()
-                            .withClasses(Styles.GRID, Styles.GRID_COLS_2)
-                            .with(div().with(h2("Current Files:")).with(pre(renderFiles(files))))));
+    ContainerTag fileUploadForm;
+
+    HtmlBundle bundle = layout.getBundle();
+
+    if (signedRequest.serviceName().equals(StorageServiceName.AZURE_BLOB.getString())) {
```
bion, line 63: suggestion (blocking): similar to the code in the controller, I suggest factoring out a strategy pattern with a class for handling each strategy (AWS, Azure, etc) to clean this up.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
```diff
@@ -75,10 +95,14 @@ private ContainerTag renderFiles(ImmutableList<StoredFile> files) {
   }
 
   private String getPresignedURL(StoredFile file) {
-    return s3Client.getPresignedUrl(file.getName()).toString();
+    String url = storageClient.getPresignedUrl(file.getName()).toString();
+    if (url.contains("azurite")) {
+      return url.replace("azurite", "localhost");
+    }
```
bion, line 101: suggestion (blocking): this looks like an Azure-specific implementation detail leaking into code that should be generic. Can this logic be pushed back into the storage client?

universal-application-tool-0.0.1/build.sbt
```diff
@@ -12,15 +12,20 @@ lazy val root = (project in file("."))
       javaJdbc,
       // JSON libraries
       "com.jayway.jsonpath" % "json-path" % "2.5.0",
-      "com.fasterxml.jackson.datatype" % "jackson-datatype-guava" % "2.10.3",
-      "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8" % "2.10.3",
+      "com.fasterxml.jackson.datatype" % "jackson-datatype-guava" % "2.12.5",
+      "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8" % "2.12.5",
+      "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.12.5",
```
bion, line 17: question (non-blocking): were these dependency version changes necessary for the azure SDK?

universal-application-tool-0.0.1/conf/application.conf
```diff
@@ -433,6 +434,21 @@ aws.s3.bucket=civiform-local-s3
...
+## Azure
+azure.blob.account=devstoreaccount1
+azure.blob.account=${?AZURE_STORAGE_ACCOUNT_NAME}
+azure.blob.container=civiform-local-blob
+azure.blob.container=${?AZURE_STORAGE_ACCOUNT_CONTAINER}
+azure.local.connection="DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite:10000/devstoreaccount1;"
+azure.local.connection=${?AZURE_LOCAL_CONNECTION_STRING}
+
+## Java
+java.time.zoneid="America/Los_Angeles"
```
bion, line 446: question (blocking): what is this needed for?

universal-application-tool-0.0.1/project/plugins.sbt
```diff
@@ -2,3 +2,4 @@
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.8")
 addSbtPlugin("com.typesafe.sbt" % "sbt-play-ebean" % "6.0.0")
 addSbtPlugin("name.de-vries" % "sbt-typescript" % "2.6.2")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
bion, line 5: thought (non-blocking): would generally prefer this kind of change be done in its own PR

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
shubha-rajan: Good catch! We fixed that for the prod client but missed the dev client.

universal-application-tool-0.0.1/project/plugins.sbt
shubha-rajan: didn't mean to commit this file! will revert the change.

bion:
Great work! A few additional overall thoughts:

1. Please add a browser test that exercises the Azure upload logic with Azurite
1. The PR description mentions "Make CORS rules more secure", where is that done?
1. Some additional whitespace throughout would make the code a bit easier for my eyes at least. In particular empty lines separating multi-line statements (e.g. conditional blocks, loops, `try/catch` blocks etc) from code at the same level of indentation
1. For future reference, please break the work into smaller PRs. Examples of things that on their own should be a single PR:
    - adding a new dependency (Java or JS)
    - adding a new docker container to the dev environment compose
    - backend logic for a feature
    - front-end logic for a feature

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
zaralouis-sf: This was all @shubha-rajan Thank you Shubha for figuring that out!!

universal-application-tool-0.0.1/build.sbt
shubha-rajan: Yes, we were getting dependency conflict warnings and NoClassDefFoundErrors because the azure SDK requires at least v2.12.5

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
shubha-rajan: Stack Overflow and the Guice wiki on GitHub are the real heroes here. ;)

universal-application-tool-0.0.1/build.sbt
bion: Got it, thanks!

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
shubha-rajan: The user delegation key is used to sign SAS tokens instead of using an account key and it uses "user" credentials (the term user is misleading here because it's actually using the application's Managed Identity credentials) to authenticate. The key needs to be rotated and this duration is the interval at which it's rotated. We weren't sure how long to make the interval. More info here: https://docs.microsoft.com/en-us/rest/api/storageservices/create-user-delegation-sas

bion: Got it, thank you. Please document this in the code.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
shubha-rajan: moved this logic to BlobStorage.java in the AzuriteClient class

universal-application-tool-0.0.1/app/services/cloud/azure/Credentials.java
shubha-rajan: Yup. fixed in latest commit.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
shubha-rajan: done

universal-application-tool-0.0.1/conf/application.conf
zaralouis-sf: ZoneID is needed for generating the SAS. SAS takes in an OffSetDateTime that's the expiration date/time for the SAS. In order to get the currnet time, we have to specify the time zone we're in.

bion: Ah, will this be an issue when the app is hosted and accessed on the east coast?

shubha-rajan: it shouldn't be, as long as we set this config option to whatever timezone the application is hosted in.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
shubha-rajan: because we used the other AWS builder as a template originally, and then changed the implementation without removing the getters. they're removed now.

universal-application-tool-0.0.1/conf/application.conf
bion: Great. Let's make this a configurable setting then. Different instances will be deployed in different time zones.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
bion: Looks like they're still here?

shubha-rajan: For some reason, sometimes github doesn't show the latest version by default. you'll have to go to this menu on the top left and choose "show all changes"
image

bion:
Aha, thanks!

universal-application-tool-0.0.1/app/views/ViewUtils.java
zaralouis-sf: Discussed offline, keeping this as a CDN for now. Creating a bug to follow up on this. Currently, the CDN is only being used in a dev Azure environment.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
zaralouis-sf: issue #1763. I've updated the codebase to only inject this CDN in an Azure dev environment, PTAL.

zaralouis-sf:
Wrt: https://github.com/seattle-uat/civiform/pull/1752#issuecomment-995326391 
Instead of writing a new browser test, I updated the browser test environment so that it can run either in the aws or azure dev environment. Currently, the only browser test that passes when run with Azurite is the dev_file_upload.test.ts test (as expected). You can run this by starting the browser test env with ```bin/run-browser-test-env --azure``` and then in a separate terminal running ```bin/run-browser-tests dev_file_upload.test.ts```. Let me know if this is different from what you had in mind.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
zaralouis-sf: Yep, this is just for dev so I removed that TODO.

universal-application-tool-0.0.1/app/views/ViewUtils.java
zaralouis-sf: I am going to play around some more with trying to get webjars working with this, if possible.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
shubha-rajan: I can open up a PR to do this as soon as this PR is merged. In the meantime I opened https://github.com/seattle-uat/civiform/issues/1768

bin/run-browser-test-env
```diff
@@ -12,11 +12,37 @@ if [ -f ".secrets" ]; then
...
+
+_setArgs "$@"
+
+# default to AWS for now
+if [ -z "$cloudProvider" ]; then
+  cloudProvider="aws" && emulator="localstack" && devServer="http://localhost:6645"
+fi
+
+# start emulator
+docker-compose --profile $cloudProvider -f browser-test/browser-test-compose.yml -f browser-test/browser-test-compose.dev.yml up -d $emulator
```
bion, line 42:
```suggestion
docker-compose --profile $cloudProvider \
  -f browser-test/browser-test-compose.yml \
  -f browser-test/browser-test-compose.dev.yml \
  up -d $emulator
```

bin/run-browser-test-env
```diff
@@ -12,11 +12,37 @@ if [ -f ".secrets" ]; then
...
 
-# wait till localstack running
-bin/localstack/wait http://localhost:6645
+# wait till emulator running
```
bion, line 44:
```suggestion
# wait until emulator running
```

bion: Easier for non-native english speakers

browser-test/package.json
```diff
@@ -11,7 +11,7 @@
     "@types/node": "^14.14.31",
     "jest": "^26.6.3",
     "jest-playwright-preset": "^1.5.1",
-    "playwright": "^1.9.2",
+    "playwright": "^1.18.0-alpha-dec-23-2021",
```
bion, line 14: question (blocking): did anything specific prompt the upgrade? why use an alpha release instead of latest stable?

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+class AzureUploadController {
+  static UPLOAD_CONTAINER_ID = 'azure-upload-form-component'
+
+  constructor() {
+    const uploadContainer = document.getElementById(
+      AzureUploadController.UPLOAD_CONTAINER_ID
+    )
+    uploadContainer.addEventListener('submit', (event) =>
+      this.attemptUpload(event, uploadContainer)
+    )
```
bion, line 14: issue (blocking): with these DOM queries in the constructor this code is a little brittle to its load/initialization pattern i.e. if it were loaded asynchronously or injected into the page before the DOM elements it's looking for it would fail.

possible solution: initialize the class after the `load` event https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  attemptUpload(event: Event, uploadContainer: HTMLElement | null) {
+    event.preventDefault()
+    // This is set by an imported JavaScript Azure blob client.
+    // window["azblob"] was used because the TS compiler complained about using
+    // window.azblob.
+    const azblob = window['azblob']
+    if (uploadContainer == null) {
+      throw 'Attempted to upload to null container'
```
bion, line 28:
```suggestion
      throw new Error('Attempted to upload to null container')
```

bion: Wrapping this in an error object will provide a stack trace and more typical exception semantics

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    const azureUploadProps = this.getAzureUploadProps(uploadContainer)
+
+    const redirectUrl = new URL(azureUploadProps.successActionRedirect)
+
+    const blockBlobURL = azblob.BlockBlobURL.fromBlobURL(
+      new azblob.BlobURL(
+        `${azureUploadProps.blobUrl}?${azureUploadProps.sasToken}`,
```
bion, line 36: suggestion (blocking): could `azureUploadProps.blobUrl` as constructed on the server include the `sasToken` instead of doing string formatting here?

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+    azblob
+      .uploadBrowserDataToBlockBlob(
+        azblob.Aborter.none,
+        azureUploadProps.file,
+        blockBlobURL
+      )
+      .then((resp, err) => {
+        if (err) {
+          throw err
+        } else {
```
bion, line 50: nit: no need for the `else {` here, can save a level of indentation for the rest of this function body

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+    // window.azblob.
+    const azblob = window['azblob']
+    if (uploadContainer == null) {
+      throw 'Attempted to upload to null container'
+    }
+    const azureUploadProps = this.getAzureUploadProps(uploadContainer)
+
+    const redirectUrl = new URL(azureUploadProps.successActionRedirect)
+
+    const blockBlobURL = azblob.BlockBlobURL.fromBlobURL(
```
bion, line 34: nit: google styleguide would be `blockBlobUrl`

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
```diff
@@ -0,0 +1,83 @@
...
+            'userFileName',
+            azureUploadProps.file.name
+          )
+          redirectUrl.searchParams.set('etag', resp.eTag)
+          redirectUrl.searchParams.set('fileName', azureUploadProps.fileName)
+          redirectUrl.searchParams.set(
+            'container',
+            azureUploadProps.containerName
+          )
+          window.location.replace(redirectUrl.toString())
```
bion, line 62: suggestion (non-blocking): it'd be nice to wrap this code in a descriptive function so the overall flow read a bit easier. What is happening here?

bin/run-browser-test-env
```diff
@@ -12,11 +12,37 @@ if [ -f ".secrets" ]; then
...
+       cloudProvider="aws"
+       emulator="localstack"
+       devServer="http://localhost:6645"
+       ;;
+    esac
+    shift
+  done
+}
+
+_setArgs "$@"
```
bion, line 34: praise: nice work accommodating both emulators!

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageStrategy.java
```diff
@@ -0,0 +1,19 @@
...
+import play.mvc.Http.Request;
+import play.mvc.Result;
+import repository.StoredFileRepository;
+
+/**
+ * Interface for interacting with the different cloud storage providers. Each cloud storage provider
+ * (currently we support Azure and AWS) implements the interface methods, which can then be called
+ * by FileUploadController.
+ */
+public interface CloudStorageStrategy {
```
bion, line 12: suggestion (blocking): these strategies are in `app/controllers/dev/` but aren't controllers and are not specific to a development task. There's not an immediately obvious place for them. One option could be to consolidate file upload code under `app/services/files`

universal-application-tool-0.0.1/app/controllers/dev/AwsStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+public class AwsStrategy implements CloudStorageStrategy {
+
+  @Override
+  public Result create(StoredFileRepository storedFileRepository, Request request) {
+    Optional<String> etag = request.queryString("etag");
+    Optional<String> bucket = request.queryString("bucket");
+    Optional<String> key = request.queryString("key");
+    if (!bucket.isPresent() || !key.isPresent()) {
+      return redirect(routes.FileUploadController.index().url());
+    }
```
bion, line 21: question (blocking): is hitting this conditional an expected part of user flow or is it guarding an error case? if the latter please log an error before redirecting.

universal-application-tool-0.0.1/app/controllers/dev/AwsStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+    updateFileRecord(storedFileRepository, key.get());
+    String successMessage =
+        String.format(
+            "File successfully uploaded to S3: bucket: %s, key: %s, etag: %s.",
+            bucket.get(), key.get(), etag.orElse(""));
```
bion, line 26: suggestion (blocking): this looks like a temporary message for use in development. Please add a TODO to remove it before the feature is available to users.

universal-application-tool-0.0.1/app/controllers/dev/AzureStrategy.java
```diff
@@ -0,0 +1,34 @@
...
+  public Result create(StoredFileRepository storedFileRepository, Request request) {
+    Optional<String> etag = request.queryString("etag");
+    Optional<String> container = request.queryString("container");
+    Optional<String> fileName = request.queryString("fileName");
+    Optional<String> userFileName = request.queryString("userFileName");
+    updateFileRecord(storedFileRepository, fileName.get());
+    String successMessage =
+        String.format(
+            "File successfully uploaded to Azure: container: %s, file name: %s, etag: %s, user"
+                + " file name: %s.",
```
bion, line 24: suggestion (blocking): this looks like a temporary message for use in development. Please add a TODO to remove it before the feature is available to users.

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageStrategy.java
bion: Ah actually `app/services/cloud` looks like the right place for this.

universal-application-tool-0.0.1/app/controllers/dev/FileUploadController.java
```diff
@@ -6,44 +6,51 @@
...
-import services.aws.SimpleStorage;
+import services.cloud.StorageClient;
+import services.cloud.StorageUploadRequest;
 import views.dev.FileUploadView;
 
-/** Controller for interacting with S3 directly in dev mode. */
+/**
+ * Controller for interacting with S3 and blob storage directly in dev mode. The logic for uploading
+ * files to cloud storage has been extracted out into {@link CloudStorageStrategy}
+ */
```
bion, line 22: issue (non-blocking): I missed that this was originally implemented in `controllers/dev`. We should move it out of this directory and clean up the naming for the class it extends.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
bion: Sounds good!

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
```diff
@@ -0,0 +1,76 @@
...
+        .setServiceName(StorageServiceName.AZURE_BLOB.getString());
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    /**
+     * Get file name for the blob. This is used to build the canonicalized resource part of the
+     * signature string.
+     */
```
bion, line 37: suggestion (blocking): move all doc comments into the outer main class rather than having them in the builder

universal-application-tool-0.0.1/app/views/ViewUtils.java
```diff
@@ -28,6 +28,14 @@ public Tag makeLocalJsTag(String filename) {
         .withType("text/javascript");
   }
 
+  /**
+   * Generates a script tag for loading a javascript asset that is provided by a web JAR and found
+   * at the given asset route.
+   */
+  public Tag makeWebJarsTag(String assetsRoute) {
+    return script().withSrc(assetsFinder.path(assetsRoute));
+  }
```
bion, line 37: praise: nice! i had tried to get this figured out but was missing this piece i think

universal-application-tool-0.0.1/app/views/dev/AwsViewStorageStrategy.java
```diff
@@ -0,0 +1,58 @@
+package views.dev;
```
bion, line 1: suggestion (blocking): move out of `dev` package. `dev` should only be for classes that are used in the development environment.

universal-application-tool-0.0.1/app/views/dev/AwsViewStorageStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+
+/** Strategy class for creating a file upload form for AWS. */
+public class AwsViewStorageStrategy implements FileUploadViewStorageStrategy {
+
+  @Override
+  public ContainerTag getFileUploadForm(
+      ViewUtils viewUtils, StorageUploadRequest storageUploadRequest, HtmlBundle bundle) {
+    if (!(storageUploadRequest instanceof SignedS3UploadRequest)) {
+      return null;
+    }
```
bion, line 23: suggestion (blocking): avoid returning `null` or assigning `null` to variables. If a value may not be present change the return type to `Optional` and return an `Optiona.empty()`

question (blocking): Is hitting this conditional an expected part of user flow or is it guarding an error case? if the latter please log an error before redirecting. if the former please add a comment.

universal-application-tool-0.0.1/app/views/dev/AwsViewStorageStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+                input()
+                    .withType("hidden")
+                    .withName("success_action_redirect")
+                    .withValue(request.successActionRedirect()))
+            .with(
+                input()
+                    .withType("text")
+                    .withName("X-Amz-Credential")
+                    .withValue(request.credential()));
+    if (!request.securityToken().isEmpty()) {
```
bion, line 40: question (blocking): when would the security token be empty?

universal-application-tool-0.0.1/app/views/dev/AzureViewStorageStrategy.java
```diff
@@ -0,0 +1,48 @@
+package views.dev;
```
bion, line 1: suggestion (blocking): move out of dev package. dev should only be for classes that are used in the development environment.

universal-application-tool-0.0.1/app/views/dev/AzureViewStorageStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private static final String AZURE_STORAGE_BLOB_WEB_JAR =
+      "lib/azure__storage-blob/browser/azure-storage-blob.min.js";
+
+  @Override
+  public ContainerTag getFileUploadForm(
+      ViewUtils viewUtils, StorageUploadRequest storageUploadRequest, HtmlBundle bundle) {
+    if (!(storageUploadRequest instanceof BlobStorageUploadRequest)) {
+      return null;
+    }
```
bion, line 25: suggestion (blocking): avoid returning null or assigning null to variables. If a value may not be present change the return type to Optional and return an Optional.empty()

question (blocking): Is hitting this conditional an expected part of user flow or is it guarding an error case? if the latter please log an error before returning or throw an exception. if the former please add a comment.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
```diff
@@ -22,43 +19,58 @@
...
-                    .with(h1(title))
-                    .with(div().with(fileUploadForm(signedRequest)))
-                    .with(
-                        div()
-                            .withClasses(Styles.GRID, Styles.GRID_COLS_2)
-                            .with(div().with(h2("Current Files:")).with(pre(renderFiles(files))))));
+    ContainerTag fileUploadForm;
+
+    HtmlBundle bundle = layout.getBundle();
+    fileUploadForm = storageStrategy.getFileUploadForm(viewUtils, signedRequest, bundle);
```
bion, line 62: suggestion (blocking): seeing that this is where the `null` would end up from the strategy classes, I suggest throwing an exception instead of returning `null` or `Optional`

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
```diff
@@ -22,43 +19,58 @@
...
+        .setTitle(title)
+        .addMainContent(
+            div()
+                .with(div(maybeFlash.orElse("")))
+                .with(h1(title))
+                .with(div().with(fileUploadForm))
+                .with(
+                    div()
+                        .withClasses(Styles.GRID, Styles.GRID_COLS_2)
+                        .with(div().with(h2("Current Files:")).with(pre(renderFiles(files))))));
```
bion, line 73: suggestion (non-blocking): please move the "Current Files:" string into the english i18n messages or file a bug and leave a  TODO

universal-application-tool-0.0.1/app/views/dev/FileUploadViewStorageStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+import j2html.tags.ContainerTag;
+import services.cloud.StorageUploadRequest;
+import services.cloud.aws.SignedS3UploadRequest;
+import services.cloud.azure.BlobStorageUploadRequest;
+import views.HtmlBundle;
+import views.ViewUtils;
+
+/**
+ * Interface for interacting with different cloud storage providers for the file upload view. We
+ * currently support Azure and AWS.
```
bion, line 12:
```suggestion
 * Interface for interacting with different cloud storage providers for the file upload view.
```
So this doesn't need to be updated when we add GCP

universal-application-tool-0.0.1/app/views/dev/FileUploadViewStorageStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public interface FileUploadViewStorageStrategy {
+
+  /**
+   * Method for getting the file upload form for different cloud providers. This currently supports
+   * AWS, which takes in a {@link SignedS3UploadRequest} and Azure, which takes in a {@link
+   * BlobStorageUploadRequest}.
```
bion, line 19:
```suggestion
   * Method for getting the file upload form for different cloud providers.
```

universal-application-tool-0.0.1/build.sbt
```diff
@@ -100,12 +105,14 @@ JsEngineKeys.engineType := JsEngineKeys.EngineType.Node
 resolvers += Resolver.bintrayRepo("webjars","maven")
 libraryDependencies ++= Seq(
     "org.webjars.npm" % "react" % "15.4.0",
-    "org.webjars.npm" % "types__react" % "15.0.34"
+    "org.webjars.npm" % "types__react" % "15.0.34",
+    "org.webjars.npm" % "azure__storage-blob" % "10.5.0",
+    "org.webjars" % "requirejs" % "2.2.0",
```
bion, line 110: question (blocking): is requirejs needed for this PR?

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageStrategy.java
bion: Didn't realize these strategies were for Azurite and Localstack. Please clarify that in comments and ignore my previous suggestion!

universal-application-tool-0.0.1/build.sbt
zaralouis-sf: Nope sorry, added it when testing and forgot to remove.

browser-test/package.json
zaralouis-sf: This was an accident, I have reverted.

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
zaralouis-sf: I've updated so that the strategy patterns throw RuntimeExceptions, can you take a look and let me know if the implementation looks okay?

universal-application-tool-0.0.1/app/views/dev/AzureViewStorageStrategy.java
zaralouis-sf: I believe we discussed offline and this should stay as is.

universal-application-tool-0.0.1/app/views/dev/AwsViewStorageStrategy.java
zaralouis-sf: It is an error case so I have updated to throw an exception. It should technically never happen though

universal-application-tool-0.0.1/app/controllers/dev/AwsStrategy.java
zaralouis-sf: This was intentional, since this file is just for dev file upload. What do you think?

universal-application-tool-0.0.1/app/views/dev/AwsStorageDevViewStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+/** Strategy class for creating a file upload form for AWS. */
+public class AwsStorageDevViewStrategy implements CloudStorageDevViewStrategy {
+
+  @Override
+  public ContainerTag getFileUploadForm(
+      ViewUtils viewUtils, StorageUploadRequest storageUploadRequest, HtmlBundle bundle)
+      throws RuntimeException {
+    if (!(storageUploadRequest instanceof SignedS3UploadRequest)) {
+      throw new RuntimeException(
+          "Trying to upload a file to AWS dev file storage using incorrect upload request type.");
```
bion, line 24: suggestion (blocking): include "localstack" in the error message to clarify e.g. "...file to localstack (AWS emulator)..."

universal-application-tool-0.0.1/app/views/dev/AzureStorageDevViewStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final String AZURE_STORAGE_BLOB_WEB_JAR =
+      "lib/azure__storage-blob/browser/azure-storage-blob.min.js";
+
+  @Override
+  public ContainerTag getFileUploadForm(
+      ViewUtils viewUtils, StorageUploadRequest storageUploadRequest, HtmlBundle bundle)
+      throws RuntimeException {
+    if (!(storageUploadRequest instanceof BlobStorageUploadRequest)) {
+      throw new RuntimeException(
+          "Trying to upload file to dev Azure blob storage using incorrect upload request type.");
```
bion, line 26: suggestion (blocking): include "azurite" in the error message to clarify e.g. "...file to azurite (Azure emulator)..."

universal-application-tool-0.0.1/app/controllers/dev/AwsStrategy.java
bion: Yep, I was confused about this. Fine as-is, thank you!

universal-application-tool-0.0.1/app/views/dev/AzureViewStorageStrategy.java
bion: Yes, thanks!

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
zaralouis-sf: Filed a bug since this PR is so big: #1792

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: We can do the string formatting in the BlobStorage class, but we have to do string formatting I believe because you have to get the blob URL and SAS token separately from the server. I could be wrong but based on my understanding.

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageDevControllerStrategy.java
```diff
@@ -0,0 +1,18 @@
+package controllers.dev;
```
goldblatt, line 1: just a comment for this file/location-- a little confused why this is being stored in the dev folder since it does seem to be used in non- dev instances?

universal-application-tool-0.0.1/app/controllers/dev/AwsStorageDevControllerStrategy.java
```diff
@@ -0,0 +1,35 @@
+package controllers.dev;
```
goldblatt, line 1: like should this controller class be in /dev/ is it only used for dev stuff?

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: Let me know how the updated code looks to you

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageDevControllerStrategy.java
zaralouis-sf: AFAIK this file is only used in dev/FileUploadController. What non-dev instances are you finding concerning? This class is supposed to be dev-mode specific.

universal-application-tool-0.0.1/app/controllers/dev/AwsStorageDevControllerStrategy.java
zaralouis-sf: This class is only used for the AWS (localhost is the AWS emulator) dev file upload view

goldblatt: oo i think im just confused like don't we use it https://github.com/seattle-uat/civiform/pull/1752/files#diff-dd152546f04a2023052ca2e60280bd5fad6875511262887f7c8e1259b35a8458R59 to specify that we are using aws/azure apis

zaralouis-sf: Specifying whether to use azure or AWS happens in the application.conf file. CloudStorageModule is for dependency injection. We're checking the conf file and then binding the appropriate dev storage view/controller class. We're not doing that for any prod code because we only currently only support AWS in prod, we haven't gotten around to the Azure implementations yet.

zaralouis-sf: Does that make sense? Am I answering your question or not really?

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageDevControllerStrategy.java
bion: The name is a little misleading since it's not actually using cloud storage, it's using local storage that simulates cloud storage. Also that it's used in a controller doesn't need to be part of the name. Maybe rename to `CloudEmulatorFileStorageStrategy.java`?

universal-application-tool-0.0.1/app/controllers/dev/AwsStorageDevControllerStrategy.java
bion: Similarly to the interface name, maybe rename this to `AwsLocalstackFileStorageStrategy`?

zaralouis-sf: Done, and changed the Azure one to ```AzuriteFileStorageStrategy``` does that sound okay?

universal-application-tool-0.0.1/app/controllers/dev/CloudStorageDevControllerStrategy.java
zaralouis-sf: Done

universal-application-tool-0.0.1/app/views/dev/AwsViewStorageStrategy.java
zaralouis-sf: This is the way the existing code was written, we didn't change this from the existing AWS dev file upload implementation.

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: Let me know how this looks.

universal-application-tool-0.0.1/app/controllers/dev/AzureStrategy.java
zaralouis-sf: This is a dev screen only so we want to keep this.

universal-application-tool-0.0.1/app/controllers/dev/AzuriteFileStorageStrategy.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static play.mvc.Results.redirect;
+
+import java.util.Optional;
+import models.StoredFile;
+import play.mvc.Http.Request;
+import play.mvc.Result;
+import repository.StoredFileRepository;
+
+/** Implements file uploading for Azure cloud storage. Redirects users on success. */
```
bion, line 11: suggestion (blocking): update comment to specify that it uploads to local azurite, not Azure cloud

universal-application-tool-0.0.1/app/controllers/dev/CloudEmulatorFileStorageStrategy.java
```diff
@@ -0,0 +1,18 @@
+package controllers.dev;
+
+import play.mvc.Http.Request;
+import play.mvc.Result;
+import repository.StoredFileRepository;
+
+/**
+ * Interface for interacting with the different cloud storage providers in dev. Each cloud storage
```
bion, line 8:
```suggestion
 * Interface for interacting with the different cloud storage provider emulators in dev. Each cloud storage
```

universal-application-tool-0.0.1/app/controllers/dev/AwsLocalstackFileStorageStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static play.mvc.Results.redirect;
+
+import java.util.Optional;
+import models.StoredFile;
+import play.mvc.Http.Request;
+import play.mvc.Result;
+import repository.StoredFileRepository;
+
+/** Implements file uploading for AWS cloud storage. Redirects users on success. */
```
bion, line 11:
```suggestion
/** Implements file uploading for localstack AWS cloud storage emulator. Redirects users on success. */
```

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: So, in the past, we were returning the full SAS URL from ```BlobStorage.java``` class, but made the decision to break that up because in all use cases except this one, we are using the blob URL separately from the SAS token. We can revert to the state where the BlobStorageClient is returning the full URL if you would prefer that, I'm just not sure it makes the most sense.

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
zaralouis-sf: Discussed offline and fixed!

universal-application-tool-0.0.1/app/assets/javascripts/azure_upload.ts
bion: Gotcha, fine to leave as-is then

universal-application-tool-0.0.1/app/views/dev/AzureViewStorageStrategy.java
zaralouis-sf: It is not expected, so I've updated the code to throw an exception

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
zaralouis-sf: Ack thanks

universal-application-tool-0.0.1/app/views/dev/FileUploadView.java
zaralouis-sf: Updated to using a web JAR now

bin/run-dev
```diff
@@ -20,10 +20,12 @@ _setArgs(){
       "--azure")
        cloudProvider="azure"
        emulator="azurite"
+       export STORAGE_SERVICE_NAME="azure-blob"
        ;;
       "--aws")
        cloudProvider="aws"
        emulator="localstack"
+       export STORAGE_SERVICE_NAME="azure-blob"
```
shubha-rajan, line 28: should this be "s3"

zaralouis-sf: Yes! Thank you

bin/run-browser-test-env
```diff
@@ -12,11 +12,43 @@ if [ -f ".secrets" ]; then
   . .secrets
 fi
 
-# start localstack
-docker-compose -f browser-test/browser-test-compose.yml -f browser-test/browser-test-compose.dev.yml up -d localstack
+# Get cloud provider (currently either AWS or Azure)
```
bion, line 15:
```suggestion
# Get cloud provider
```
comments like this often get out of date when the code changes

bin/run-dev
```diff
@@ -12,12 +12,40 @@ if [ -f ".secrets" ]; then
...
+  export STORAGE_SERVICE_NAME="s3"
+fi
+
+# start emulator
+docker-compose --profile $cloudProvider -f docker-compose.yml -f docker-compose.dev.yml up -d $emulator
 
 # wait till localstack running
 # otherwise other services crash
-bin/localstack/wait
+bin/$emulator/wait
```
bion, line 48:
```suggestion
bin/${emulator}/wait
```

as per https://google.github.io/styleguide/shellguide.html#variable-expansion

bin/run-dev
```diff
@@ -12,12 +12,40 @@ if [ -f ".secrets" ]; then
...
+_setArgs $*
+
+# default to AWS for now
+if [ -z "$cloudProvider" ]; then 
+  cloudProvider="aws" && emulator="localstack"
+  export STORAGE_SERVICE_NAME="s3"
+fi
+
+# start emulator
+docker-compose --profile $cloudProvider -f docker-compose.yml -f docker-compose.dev.yml up -d $emulator
```
bion, line 44:
```suggestion
docker-compose --profile $cloudProvider \
  -f docker-compose.yml \
  -f docker-compose.dev.yml \
  up \
  -d $emulator
```

universal-application-tool-0.0.1/app/modules/CloudStorageModule.java
```diff
@@ -0,0 +1,80 @@
...
+import controllers.dev.AzuriteFileStorageStrategy;
+import controllers.dev.CloudEmulatorFileStorageStrategy;
+import play.Environment;
+import services.cloud.StorageClient;
+import services.cloud.StorageServiceName;
+import views.dev.AwsStorageDevViewStrategy;
+import views.dev.AzureStorageDevViewStrategy;
+import views.dev.CloudStorageDevViewStrategy;
+
+/** CloudStorageModule configures and initializes the AWS and Azure file storage classes. */
```
bion, line 18:
```suggestion
/** Configures and initializes the classes for interacting with file storage backends. */
```
So that the comment remains accurate if the class is renamed or additional backends are added

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
```diff
@@ -0,0 +1,26 @@
...
+
+/** Interface for working with Cloud storage. We currently support AWS S3 and Azure Blob storage. */
```
bion, line 5:
```suggestion
/** Interface for working with cloud file storage backends. */
```

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
```diff
@@ -0,0 +1,26 @@
...
+   */
+  URL getPresignedUrl(String fileName);
+
+  /**
+   * Creates and returns a request to upload a file to cloud storage.
+   *
+   * @param fileName The file to upload to cloud storage
+   * @param successRedirectActionLink Where a user should be redirected upon successful file upload.
+   */
+  StorageUploadRequest getSignedUploadRequest(String fileName, String successRedirectActionLink);
```
bion, line 22: The comments state that both of these methods are for uploading, but isn't `getPresignedUrl` for downloading?

universal-application-tool-0.0.1/app/services/cloud/StorageClient.java
bion: suggestion (blocking)

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    try {
+      return new URL(signedUrl);
+    } catch (java.net.MalformedURLException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public Client getClient() {
+    return client;
+  }
```
bion, line 86: suggestion (blocking): this is only used in test, please remove `public` access modifier to keep its access level as restricted as possible

bion: It may seem pedantic, but we want this class to encapsulate interaction with the library code (`client`) and that's less likely to remain the case if this method is public

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorageUploadRequest.java
```diff
@@ -0,0 +1,71 @@
...
+    public abstract Builder setContainerName(String containerName);
+
+    public abstract Builder setBlobUrl(String blobUrl);
+
+    public abstract Builder setSasToken(String sasToken);
+
+    public abstract Builder setSuccessActionRedirect(String successActionRedirect);
+
+    /** Get the service name (this is always set to "azure-blob" so the setter is private */
+    abstract Builder setServiceName(String serviceName);
```
bion, line 62: suggestion (blocking): move comment to getter on outer class

bion (approved):

PR #1754, opened 2021-12-15
Use updated dependency syntax

bion:
The `in` keyword is deprecated in sbt in favor of `/` syntax. This updates our build files to the newer syntax.


dkatzz (approved):
Thanks for doing this! Adding [this link](https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash) as more context on the change.

PR #1758, opened 2021-12-21
Use docker hub container registry for everything except deployment

bion:
Stage 2 of 3 for moving away from AWS ECR. Stage 1 was uploading logging in to docker hub and uploading our main image. 

This PR switches to using docker hub for all day-to-day docker image management while continuing to upload images to AWS ECR in case there's a problem with our docker hub approach.

The final stage will be removing ECR from the regular development cycle altogether.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Build test container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -t civiform --cache-from public.ecr.aws/t1q6b4h2/civiform-dev:latest ./
+        run: docker build -t civiform --cache-from docker.io/civiform/civiform:latest ./
       - name: Run tests
         run: docker run -v /var/run/docker.sock:/var/run/docker.sock civiform test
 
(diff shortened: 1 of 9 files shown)


shanemc-goog:
Is there documentation of our nomenclature in this group of script? If not it might help to have that as a follow along, and snap these to those if they are out of sync.

EG. 

are bin/build-dev and bin/run-dev the same "dev"? 

I _think_ the second dev is more "local" or the former is more .... "docker-dev" "public-dev" etc.

shanemc-goog (approved):

bion:
> Is there documentation of our nomenclature in this group of script? If not it might help to have that as a follow along, and snap these to those if they are out of sync.

@shanemc-goog I'll write a wiki page describing the different docker images

> are bin/build-dev and bin/run-dev the same "dev"?
> I think the second dev is more "local" or the former is more .... "docker-dev" "public-dev" etc.

Sort of, both of them reference the same image tag. It's possible to override the shared civiform-dev image with a local one when you `bin/run-dev`, but by default it uses the same one that was built and uploaded using `bin/build-dev`.

shanemc-goog:
> Sort of, both of them reference the same image tag. It's possible to override the shared civiform-dev image with a local one when you `bin/run-dev`, but by default it uses the same one that was built and uploaded using `bin/build-dev`.

ok, so I guess it's non intuitive to me; we may need some new names for cleaner lines.  I think it's unintuitive that build-dev will affect not my local system, when run-dev is how I see my local CiviForm changes.

So the former is also a little error-prone too; though possibly ACLs will prevent accidents, however it then adds confusion around run-dev for local development.

PR #1766, opened 2021-12-30
Only show errors when a question is required or partially filled in

### LINKED ISSUE #1652: Error validation shows on optional question if empty and required question has error
**Describe the bug**
On a program block where there is an optional and required question, if the user makes a mistake on the required question and clicks next, error validation will show on both questions (even though the optional one is blank). It is kind of confusing and makes it seem like the optional question is required. https://github.com/seattle-uat/civiform/issues/1423 would be helpful in making this more clear, but maybe error validation should only show for optional questions if they've started filling them.

**To Reproduce**
1. Create a program with a block that has an optional and required question
2. When filling out the page as an applicant, click next without filling out the questions
3. See error validation on both the required and optional question

**Expected behavior**
I think error validation should only show for optional questions if they've started filling them.

**Screenshots**
Uploaded

dkatzz:
### Description
Error validation shows on optional question if it is empty and the required question has error. This change ensures errors only show when a question is required or (if it's optional), the errors will only show if that question is partially filled in.

### Checklist

### Issue(s)
Fixes #1652

### DIFF
--- a/browser-test/src/address.test.ts
+++ b/browser-test/src/address.test.ts
@@ -315,13 +315,6 @@ describe('address applicant flow', () => {
       })
 
       it('optional has no errors', async () => {
-        await loginAsGuest(pageObject)
-        await selectApplicantLanguage(pageObject, 'English')
-
-        await applicantQuestions.applyProgram(programName)
-        await applicantQuestions.answerAddressQuestion('', '', '', '', '', 1)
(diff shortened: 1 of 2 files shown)


bion (approved):

shanemc-goog:
LGTM

Issue #1764, opened 2021-12-24
Multiple users tied to temporary IDCS account


leonwong:
**Describe the bug**
From Thor Peterson in Seattle: 

> We’re currently using CiviForm for the Utility Discount Program. A customer is reporting that when he logged into his account, the system displayed another customer’s name, and the Seattle City Light account number of yet another person.
> 
> Obviously, this is really problematic! Can you assist in troubleshooting?
>

dkatzz:
Members on the team and I weren't able to reproduce. Thor is going to ask the client for repro steps. With the AWS access I have, I'm unfortunately unable to query the database in staging or prod to find issues with the entries.

I'd like to query the database to see if there are any cases where an account has multiple applicants (which shouldn't be the case based on [this comment](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/models/Account.java#L134) or applicants with the [same email](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/auth/CiviFormProfileAdapter.java#L89), which also shouldn't be possible. I also wonder if there is any weird case with [merging](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/auth/CiviFormProfileAdapter.java#L96) profile data, but I wouldn't expect the applicant to see a name they are unfamiliar with even in that case.

dkatzz:
Update from Elise:
- John Yeung is looking into the issue on the IDCS Oracle side of things.
- We have two known examples of this happening (one on 11/22, the other on 12/23).
- IF a resident is using a shared computer, logins could be stored within the same web browser session. That may not be the case in these examples though.

dkatzz:
John searched the IDCS profiles and they are linked correctly there. I'm trying to get the AWS permissions to query the database to get more information about what may be stored incorrectly and causing this.

rasmi:
While others investigate -- in case this is an issue related to caching of user session credentials behind some shared proxy, I created a [PR that adds the `no-store` header](https://github.com/seattle-uat/civiform/pull/1767) for completeness on disabling caches. I don't know if this is the root cause, but it can't hurt.

dkatzz:
I was able to query the database by following [this](https://github.com/seattle-uat/civiform/wiki/Breaking-glass-on-database). I only found one of the users' emails that was impacted, but this user seemed to be linked correctly across the tables. Ran the following queries:

[...]

PR #1767, opened 2021-12-30
Update caching headers for completeness

rasmi:
### Description
Adding the [`no-store`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) flag to caching headers for completeness. We already disable browser and proxy caching through existing flags, but various online references indicate `no-store` should also be present.

### Issue(s)
Possibly related to #1764.


dkatzz (approved):

rasmi:
(I am not sure how this may or may not affect guest applications.)

bion:
@rasmi sorry for being so slow on this, should have merged it ages ago. If you update your branch with the latest from main the new tests will pass and I'll get this in.

PR #1780, opened 2022-01-05
Fix image push in building browser tests

syntapy:
### Description
Change so that pushing the browser test images happens only in the [push workflow](https://github.com/seattle-uat/civiform/blob/main/.github/workflows/push.yaml#L35-L44) and not the other two workflows that run the `bin/build-browser-tests` script

Without fix this I was getting `Error: Cannot perform an interactive login from a non TTY device` from the [Build browser testing container](https://github.com/seattle-uat/civiform/blob/main/.github/workflows/pr.yaml#L33-L38) workflow

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -36,7 +36,7 @@ jobs:
         id: build_browser_tests
         env:
           DOCKER_BUILDKIT: 1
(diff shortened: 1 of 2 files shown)


bion (approved):

bion:
Thank you, @syntapy !

PR #1786, opened 2022-01-06
Adding in variables and an example tfvars file

### LINKED ISSUE #1718: Terraform: Parameterize hard coded names
Create an isolated source for hard coded names that can be updated independent of the generalized Terraform configuration.

EG

```
variable "location" {
  type        = string
  description = "Azure Resource Region Location"
}

```

.tfvars file:
```
location    = "eastus"
```

goldblatt:
### Description
Adds in a bunch of variables for the terraform config. That way cities can just configure their own .tfvars file according to their specific needs. We include a single example.tfvars that can be used as an example vars file to show how this paradigm works. 

### Checklist
- [X] Extended the README / documentation, if necessary

### Issue(s)
Fixes #1718


cloud/azure/main.tf
```diff
@@ -153,23 +153,23 @@ resource "azurerm_postgresql_server" "civiform" {
...
+  administrator_login          = var.postgres_admin_login
+  administrator_login_password = var.postgres_admin_password
 
   // fqdn civiform-db.postgres.database.azure.com
 
-  sku_name   = "GP_Gen5_4"
-  version    = "11"
-  storage_mb = 5120
+  sku_name   = var.postgres_sku_name
+  version    = var.postgres_version
```
shubha-rajan: suggestion(blocking): Upgrading/downgrading postgres version might cause breaking changes in the application so we'd probably want to hard-code this instead of defaulting it in `variables.tf`

cloud/azure/main.tf
```diff
@@ -153,23 +153,23 @@ resource "azurerm_postgresql_server" "civiform" {
...
-  backup_retention_days        = 7
+  backup_retention_days        = var.postgres_backup_retention_days
   geo_redundant_backup_enabled = false
   auto_grow_enabled            = true
 
   public_network_access_enabled = false
 
   ssl_enforcement_enabled          = true
-  ssl_minimal_tls_version_enforced = "TLS1_2"
+  ssl_minimal_tls_version_enforced = var.postgres_ssl_minimal_tls_version_enforced
```
shubha-rajan, line 172: suggestion(blocking): We probably want to hard code this to "TLS1_2" since we're storing sensitive data and TLS 1.1 and 1.0 are insecure and deprecated. See https://www.digicert.com/blog/depreciating-tls-1-0-and-1-1

cloud/azure/variables.tf
```diff
@@ -18,3 +18,99 @@ variable "image_tag_name" {
...
+  default = [
+    "10.0.0.0/16"
+  ]
+}
+
+variable "subnet_address_prefixes" {
+  type        = list(string)
+  description = "List of the apps subnet address prefixes"
+  default = [
+    "10.0.2.0/24"
```
shubha-rajan, line 40: suggestion(non-blocking): do you think it would be useful to add a comment here to make sure that this range doesn't overlap with the range for the postgres subnet address prefixes?

cloud/azure/variables.tf
```diff
@@ -18,3 +18,99 @@ variable "image_tag_name" {
...
+  ]
+}
+
+variable "app_secret_key" {
+  type        = string
+  description = "Secret Key For the app"
+}
+
+variable "app_sku" {
+  type        = map(string)
```
shubha-rajan, line 50: praise: this map is so much cleaner than the hard coded strings we had before!

cloud/azure/variables.tf
```diff
@@ -18,3 +18,99 @@ variable "image_tag_name" {
...
+  type        = string
+  description = "The SKU for the sever logs"
+  default     = "PerGB2018"
+}
+
+variable "log_retention" {
+  type        = number
+  description = "The number of days the logs will be retained for"
+  default     = 30
+}
```
shubha-rajan, line 109: nit: add a newline here

shubha-rajan (commented):
Great work! I left a few comments. As we discussed offline, I think a good way to think about what to hardcode vs. make a variable is whether it's something we want or expect them to change. Another thought is whether we think it's something that'll vary between deployments.

cloud/azure/example.auto.tfvars
```diff
@@ -0,0 +1,4 @@
+app_secret_key = "insecure-secret-key"
+
+postgres_admin_login    = "psqladmin"
+postgres_admin_password = "H@Sh1CoR3!"
```
shubha-rajan, line 4: nit: add a newline

shubha-rajan (approved):

PR #1801, opened 2022-01-12
Add scripts for pulling program data from production into local dev

bion:
- [x] adds bin/dev-restore-db for truncating the local dev db and importing a dump
- [x] adds bin/pull-prod-admin-config for automating program and question config export from prod


bin/dev-restore-db
```diff
@@ -0,0 +1,53 @@
...
+TRUNCATE TABLE versions_programs;
+TRUNCATE TABLE versions_questions;
+EOF
+)
+
+# remove newlines
+TRUNCATE_SQL=${TRUNCATE_SQL//$'\n'/}
+
+# location of dump file in docker container
+DUMP_FILE="project/$1"
```
shanemc-goog, line 36: test if this exists

if [[ ! -f "$DUMP_FILE" ]]; then
    echo "$1 doesn't exist in the base directory."
    popd;
    exit 1;
fi

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
...
+set -e
+
+export AWS_DEFAULT_REGION=us-west-2
+TIMESTAMP=$(date +%s)
+ENVIRONMENT=staging
+STAGING_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:staging-databasepassword-rh1wfF
+PROD_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:prod-databasepassword-OM7x4q
+DB_PASSWORD_ID=$STAGING_DB_PASSWORD_ID
+
+if [ "$1" = "prod" ]; then
```
shanemc-goog, line 14: for reasons, guide prefers [[ over [

https://google.github.io/styleguide/shellguide.html#test----and---

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
+#! /bin/bash
+
+# DOC: Dump program and question configuration from production or staging PostgreSQL database to a local archive file using pg_dump.
+
+set -e
+
+export AWS_DEFAULT_REGION=us-west-2
+TIMESTAMP=$(date +%s)
+ENVIRONMENT=staging
+STAGING_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:staging-databasepassword-rh1wfF
```
shanemc-goog, line 10: Per the guide "constants" should be marked readonly

https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
+#! /bin/bash
+
+# DOC: Dump program and question configuration from production or staging PostgreSQL database to a local archive file using pg_dump.
+
+set -e
+
+export AWS_DEFAULT_REGION=us-west-2
+TIMESTAMP=$(date +%s)
+ENVIRONMENT=staging
```
shanemc-goog, line 9: the guide says to prefer quoting bare strings.
-------

Prefer quoting strings that are “words” (as opposed to command options or path names).


\# "prefer quoting words", not compulsory
readonly USE_INTEGER='true'



https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
...
+STAGING_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:staging-databasepassword-rh1wfF
+PROD_DB_PASSWORD_ID=arn:aws:secretsmanager:us-west-2:405662711265:secret:prod-databasepassword-OM7x4q
+DB_PASSWORD_ID=$STAGING_DB_PASSWORD_ID
+
+if [ "$1" = "prod" ]; then
+    ENVIRONMENT="prod"
+    DB_PASSWORD_ID=$PROD_DB_PASSWORD_ID
+fi
+echo "Connecting to $ENVIRONMENT DB..."
```
shanemc-goog, line 18: The guide suggests marking the above two as readonly here now that they're set

https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
+#! /bin/bash
```
shanemc-goog, line 1: Is this largely a copy of breakglass?  Can we modularize that at all so that we're not duplicating complicated scripts?

shanemc-goog (changes requested):
Before I realized the new file is largely a copy of the breakglass one I left some style comments per the google shell guide.

Given we haven't formally standardized on a style, you can take or leave them.

bin/dev-restore-db
bion: already tested on L10. the `DUMP_FILE` path has `project` prepended to match the volume path in the docker container.

bin/dev-restore-db
```diff
@@ -0,0 +1,53 @@
+#! /usr/bin/env bash
+
+# DOC: Load a postgres dump into the dev database (pg_restore), replacing whatever state is currently in it. Takes a file name argument, file must be in current directory.
```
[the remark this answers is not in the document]
bion: done

bin/dev-restore-db
```diff
@@ -0,0 +1,53 @@
...
+
+set -e
+set +x
+
+if [[ ! -f "$1" ]]; then
+  echo "Usage: bin/dev-restore-db DUMP_FILE_NAME"
+  exit 1
+fi
+
+# find repo root name (currently civiform), also support any names for backward and forward compatibility
```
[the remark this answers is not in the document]
bion: done

bin/pull-prod-admin-config
bion: I'm open to thoughts on this. Yes, it feels gross to duplicate a bunch of code like this instead of factoring it out. OTOH the pull code here is very specific to a single deployment and I'm hesitant to invest the time in refactoring it. I also generally err to the side of reducing duplication once it gets to three rather than two instances.

shanemc-goog: wrt duplication: it seems like we have no good way to manage that "3rd times the charm" part though....  and seemingly we're already modifying one of the forks with the style changes here, so it's already kinda impossible to dedup them when a 3rd need comes along.

the "specific to a single deployment" part is also kinda messy right since that's not clear either.. but is implicit in the nature of the code base.

What's a good way to not get bogged down but leave it better than we found it?  Make a bin/seattle dir with this there so at least it's more isolated from the larger ideally agnostic code base?

bin/dev-restore-db
shanemc-goog: ah missed that

bin/pull-prod-admin-config
bion: > What's a good way to not get bogged down but leave it better than we found it? Make a bin/seattle dir with this there so at least it's more isolated from the larger ideally agnostic code base?

That sounds good to me. I would like our shell scripts to meet our quality bar, but I don't think refactoring this stuff is the best use of time right now. Admittedly it would probably not be such a time concern if I were better at working with shell scripts.

Another issue is that I can't test these scripts without running them in cloud shell which would make refactoring even slower.

bin/seattle/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
...
+        -t versions \
+        -t versions_programs \
+        -t versions_questions \
+        postgres > program_data.dump"
+
+echo "Copying file to local computer using scp"
+scp -o UserKnownHostsFile=/dev/null \
+  -o StrictHostKeyChecking=no \
+  -o IdentitiesOnly=yes \
+  -i $KEY_FILE ubuntu@${INSTANCE_IPADDRESS}:/home/ubuntu/program_data.dump .
```
shanemc-goog, line 125: put all the args on their own line for consistency and readability.

bin/seattle/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
...
+echo "Revoke access to DB security group"
+aws ec2 revoke-security-group-ingress \
+    --group-id $DBSG_ID \
+    --protocol tcp \
+    --port 0-65535 \
+    --source-group $SG_ID
+
+# Remove the emergency security group
+echo "Removing the emergency security group"
+until aws ec2 wait instance-terminated --instance-ids $INSTANCE_ID; do :; done
```
shanemc-goog, line 146: Is the : in "do :" correct?  What's it doing?

should there be a sleep in here?

shanemc-goog (approved):

bin/seattle/pull-prod-admin-config
bion: uhh good question

bion: i have no idea why it would be done this way vs

```
aws ec2 wait instance-terminated --instance-ids $INSTANCE_ID
```

bin/pull-prod-admin-config
bion: done

bin/pull-prod-admin-config
bion: done

bin/pull-prod-admin-config
bion: done

bin/pull-prod-admin-config
bion: done

bin/pull-prod-admin-config
```diff
@@ -0,0 +1,147 @@
...
+DB_PASSWORD_ID=$STAGING_DB_PASSWORD_ID
+
+if [ "$1" = "prod" ]; then
+    ENVIRONMENT="prod"
+    DB_PASSWORD_ID=$PROD_DB_PASSWORD_ID
+fi
+echo "Connecting to $ENVIRONMENT DB..."
+
+VPC_ID=$(aws ec2 describe-vpcs \
+    --filters Name=tag:Name,Values=$ENVIRONMENT.civiform.vpc \
```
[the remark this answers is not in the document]
bion: done

bin/seattle/pull-prod-admin-config
```diff
@@ -0,0 +1,156 @@
...
+
+readonly ENVIRONMENT
+readonly DB_PASSWORD_ID
+
+echo "Connecting to ${ENVIRONMENT} DB..."
+
+readonly VPC_ID=$(aws ec2 describe-vpcs \
+    --filters "Name=tag:Name,Values=${ENVIRONMENT}.civiform.vpc" \
+    --query 'Vpcs[0].VpcId' \
+    --output text) && echo "${ENVIRONMENT} VPC ID: $VPC_ID"
```
shanemc-goog, line 29: need curlies on VPC_ID

shanemc-goog (approved):

bin/seattle/pull-prod-admin-config
bion: done

PR #1805, opened 2022-01-12
Deploy staging on all merges into main

bion:
It used to be we deployed to staging as part of the build-and-push task. I removed that code since it seemed a poor place to do it given the name and responsibility of the task and didn't notice that the deploy-staging task only run if there were changes in `infra/*`. This PR makes that step run on every merge into main.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -49,7 +49,6 @@ jobs:
         run: bin/build-and-push
 
       - name: Run deploy-staging.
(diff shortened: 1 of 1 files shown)


goldblatt (approved):

goldblatt:
I've also update the secrets in the staging github to use a new service principal. we needed to add 
AppRoleAssignment.ReadWrite to the app registration associated with the service principal. we also needed to add the service principal to the subscription IAM. i think the prior made service principal didn't have the correct access due to moving tenants

PR #1828, opened 2022-01-20
Run Azurite browser tests in CI

### LINKED ISSUE #1794: Add Azure browser tests to GitHub testing CI
We have Azure browser tests that can be run by running: ```bin/run-browser-test-env --azure``` then, in a separate terminal, run: ```bin/run-browser-tests dev_file_upload.test.ts```

This needs to be added to the GitHub CI tests, which are run with this command: ```bin/run-browser-tests-ci```

bion:
- [x] run the azurite browser tests in a separate check on each pull request (just one test file, for now)
- [x] rename `run_tests` check to `run_junit_tests`
- [x] fix port mapping for azurite in browser test compose file
- [x] encapsulate logic for running the CI browser test environment in shell script (trying to reduce the amount of logic in the GH actions yaml files)
- [x] added a simple library system for shared shell code in bin/lib, use it with `source bin/lib.sh` in your other scripts
- [x] added `bin/lib/truth.sh` with functions for handling boolean variables in scripts
- [x] added `bin/lib/out.sh` with a function for printing to stdout

closes #1794


entrypoint.sh
```diff
@@ -1,3 +1,4 @@
 #! /bin/bash
+
```
bion, line 2: this just looked wrong without a newline after the shebang

browser-test/browser-test-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
     profiles: ["azure"]
     container_name: azurite-test
     ports:
-    - 10000
+      - 10000:10000
```
bion, line 20: previously this just mapped the port in the docker compose network but we need it mapped to the bridged (localhost) network as well

bin/run-browser-tests-ci
```diff
@@ -12,6 +12,7 @@ docker run \
   -v "$(pwd)/browser-test/bin:/usr/src/civiform-browser-tests/bin" \
   --network browser-test_default \
   civiform-browser-test:latest \
-  /usr/src/civiform-browser-tests/bin/wait_for_server_start_and_run_tests.sh
+  /usr/src/civiform-browser-tests/bin/wait_for_server_start_and_run_tests.sh \
+  $@
```
bion, line 16: allows passing a specific test filename through since we don't want to run the whole suite for azure tests, just the ones that exercise azure integration

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
...
+  readonly SERVER_URL="http://localhost:6645"
+
   export STORAGE_SERVICE_NAME="s3"
 fi
 
+compose_command="docker-compose --profile ${CLOUD_PROVIDER}"
+compose_command="${compose_command} -f browser-test/browser-test-compose.yml"
+
+if [[ "${ci}" = "no" ]]; then
+  compose_command="${compose_command} -f browser-test/browser-test-compose.dev.yml"
```
bion, line 59: don't attempt to use the local sbt cache directories in CI

bin/azurite/wait
```diff
@@ -6,29 +6,29 @@ pushd $(git rev-parse --show-toplevel) > /dev/null
 
 set -e
 
-azurite_endpoint=http://localhost:10000
+azurite_endpoint="http://localhost:10000"
```
bion, line 9: just formatting changes in this file, i didn't do the whole thing but can if desired. it's late.

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
   . .secrets
 fi
 
+ci="no"
```
shanemc-goog, line 15: doc the flags at the top of the file, in particular so abbreviations like ci are understandable and that azure/aws are mutually exclusive.

shanemc-goog, line 15: style, optional: Google general style eschews abbreviations especially if they're not core to the system, as it tends to result in problems for the benefit of typing less.

I'd recommend spelling this out, especially internal to the file.

shanemc-goog, line 15: How about using shell versions of bools rather than magic strings?

run_ci = 0

if [[ ${run_ci} -eq 1 ]]
then
  ....

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
...
     shift
   done
 }
 
 _setArgs "$@"
 
 # default to AWS for now
-if [ -z "$cloudProvider" ]; then
-  cloudProvider="aws" && emulator="localstack" && devServer="http://localhost:6645"
+if [ -z "${CLOUD_PROVIDER}" ]; then
```
shanemc-goog, line 47: We can remove this duplication by setting these above the case on 20 as the defaults and have it explicitly at the start and remove the aws block.

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
...
+        ci="yes"
+        ;;
+
       "--azure")
-       cloudProvider="azure"
-       emulator="azurite"
-       devServer="http://localhost:10000"
-       export STORAGE_SERVICE_NAME="azure-blob"
-       ;;
+        readonly CLOUD_PROVIDER="azure"
```
shanemc-goog, line 26: optional: Would be nice to check if azure and aws are both set and issue an error.

bin/run-browser-test-env
shanemc-goog: Put these informative comments in the file :)

It's also unclear to me how this file accomplishes that goal, maybe comment that too.

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
...
+echo "${START_EMULATOR_COMMAND}"
+
 # start emulator
-docker-compose --profile $cloudProvider \
-  -f browser-test/browser-test-compose.yml \
-  -f browser-test/browser-test-compose.dev.yml \
-  up -d $emulator
+eval "${START_EMULATOR_COMMAND}"
+
+# wait until the emulator is running
```
shanemc-goog, line 71: Init caps and trialing periods is Google style.  L17 uses init cap, so do something for consistency at least :)

bin/run-browser-test-env
```diff
@@ -12,45 +12,73 @@ if [ -f ".secrets" ]; then
...
+if [[ "${ci}" = "yes" ]]; then
+  compose_command="${compose_command} > .dockerlogs 2>&1 &"
+fi
 
-# wait until emulator running
-bin/$emulator/wait $devServer
+echo "Running docker compose to start the CiviForm environment:"
+echo "${compose_command}"
 
 # start everything else
```
shanemc-goog, line 81: Move this to L74?

browser-test/browser-test-compose.yml
shanemc-goog: Optional: the port:port syntax isn't intuitive to me, but I don't have yml readability either :D Consider if an inline comment would help.

shanemc-goog (commented):
Nitty/style things

bin/run-browser-test-env
bion: done

bin/run-browser-test-env
bion: done

bin/run-browser-test-env
bion: done

bin/run-browser-test-env
bion: The comment?

browser-test/browser-test-compose.yml
bion: Here's doc for that: https://docs.docker.com/compose/compose-file/compose-file-v3/#ports

I'm hesitant to leave a comment since there is a lot in this file that isn't particularly intuitive without reading the docs. Also the "ports" config option is used in a bunch of places and commenting each of them seems cumbersome.

bin/run-browser-test-env
bion: done

bin/run-browser-test-env
bion: done

bin/run-browser-test-env
shanemc-goog: yeah, I read it as a section comment and the section starts further above now. optional.

.github/workflows/pr.yaml
```diff
@@ -36,18 +36,39 @@ jobs:
...
+        run: docker build -t civiform --cache-from docker.io/civiform/civiform:latest ./
+      - name: Build browser testing container
+        env:
+          DOCKER_BUILDKIT: 1
+          DOCKER_HUB_ACCESS_TOKEN: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
+        run: bin/build-browser-tests
+      - name: Bring up test env with Azurite
+        run: bin/run-browser-test-env --azure --ci
+      - name: Run browser tests with Azurite
```
zaralouis-sf, line 66: Suggestion (non-blocking): I'm wondering if, for now, we should say something like "run dev file upload browser test with Azurite" just for clarity, until we have all browser tests working, then we can clean this up to. Would that make sense?

bin/run-browser-test-env
bion: done

.github/workflows/pr.yaml
bion: good call!

bin/run-browser-test-env
bion: Ah, leaving out for now since it feels more like it's specific to the eval statement. LMK if you'd prefer I move it or add a comment for the stuff above

zaralouis-sf (approved):

bin/run-browser-test-env
```diff
@@ -1,84 +1,139 @@
...
+
+--azure Use the Azure emulator (Azurite) for simluating cloud services,
+        cannot be used with --aws flag.
+EOF
+        exit 0
+        ;;
+
       "--ci")
-        ci="yes"
+        continuous_integration=0
```
shanemc-goog, line 74: It looks like you're using 1 to mean don't do CI and 0 to mean to do it which is counter intuitive and backwards of my expectation that 1 is yes/on/do-CI

bin/run-browser-test-env
```diff
@@ -1,84 +1,139 @@
...
+# Ensure that only one cloud provider flag was passed tot he script.
+# Exit with error message and status code 1 if more than
+# Globals:
+#   cloud_provider
+#   already_set_cloud_provider
+# Arguments:
+#   1: the cloud provider indicated by the flag currently being processed
+#######################################
+already_set_cloud_provider=1
+ensure_only_one_cloud_provider_flag() {
```
shanemc-goog, line 28: Fancy!!!!!

shanemc-goog (changes requested):

bin/run-browser-test-env
bion: just got fancier

bin/run-browser-test-env
bion: added bin/lib/truth.sh for encapsulating this kinda thing

bin/lib/truth.sh
```diff
@@ -0,0 +1,122 @@
...
+#######################################
+# Check if variable is both a boolean and false.
+# Exits with error message and status code 2 if variable is not a boolean.
+# Arguments:
+#   1: the variable name
+#######################################
+function truth::is_disabled() {
+  truth::assert_is_boolean "${1}"
+
+  [[ "${!1}" != "true" ]]
```
shanemc-goog, line 24: optional: do ! is_enabled  ?

bin/lib/truth.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+#######################################
+# Set boolean variable to true
```
shanemc-goog, line 28: trailing periods like above

shanemc-goog (changes requested):
just the nit. looks great

bin/run-browser-test-env
```diff
@@ -1,56 +1,139 @@
...
+function set_args() {
   while [ "${1:-}" != "" ]; do
     case "$1" in
+      "--help")
+        cat <<-EOF
+Usage: bin/run-browser-test-env [options]
+
+Run a local civiform environment for the browser tests to exercise.
+
+Options:
```
zaralouis-sf, line 61: I know you put it above, but I think it could be helpful here again to reiterate what the defaults are if these flags aren't set.

bion: done

bin/lib/truth.sh
bion: done

bin/lib/truth.sh
bion: done

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
```diff
@@ -260,9 +260,7 @@ public String getSasToken(String fileName, Optional<String> prefixedOriginalFile
 
     @Override
     public String getBlobUrl(String fileName) {
-      BlobClient blobClient = blobContainerClient.getBlobClient(fileName);
-      String signedUrl = blobClient.getBlobUrl().replace("azurite", "localhost");
-      return signedUrl;
+      return blobContainerClient.getBlobClient(fileName).getBlobUrl();
```
shanemc-goog, line 263: fyi this is showing up in the PR

shanemc-goog (approved):
I think your PR got clouded by some other changes.  Maybe you need to sync your checkout?

universal-application-tool-0.0.1/app/services/cloud/azure/BlobStorage.java
bion: This is intentional.

The one azure test we're running right now is broken on main (and this branch) because of the replace here. The replace was accidentally merged into main in a different PR, this fixes the test.

PR #1829, opened 2022-01-20
Adding in the SES module

goldblatt:
### Description
Set up SES via terraform and pass the sender email to the app. 

### Checklist

### Issue(s)


cloud/aws/providers.tf
```diff
@@ -0,0 +1,12 @@
...
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      version = "3.72.0"
+    }
+  }
+} 
+
+provider "aws" {
+  region = "us-east-1"
```
bion, line 11: suggestion (non-blocking, if-minor): use a variable

cloud/deploys/staging/providers.tf
```diff
@@ -1,3 +1,7 @@
 provider "azurerm" {
   features {}
 }
+
+provider "aws" {
+  region = "us-east-1"
```
bion, line 6: suggestion (non-blocking, if-minor): use a variable

cloud/deploys/staging/main.tf
```diff
@@ -21,4 +25,10 @@ module "app" {
 
   application_name = var.application_name
   app_secret_key   = var.app_secret_key
+  ses_sender_email = var.sender_email_address
+}
+
+module "email_service" {
+  source               = "../../aws/modules/ses"
+  sender_email_address = var.sender_email_address
```
bion, line 33: suggestion (non-blocking, if-minor): could we make using SES with Azure optional on a config value? very much an only-if-minor suggestion

bion (approved):

cloud/deploys/staging/main.tf
goldblatt: i would think that this file would change if we were going to the deploy differently if that makes sense not passing in the config values here but fully just not having this exists in the deploys folder

bion:
aaaah that makes a lot of sense

PR #1831, opened 2022-01-20
Terraform Support For Blob Storage

### LINKED ISSUE #1772: Add provisioning for Azure Blob Storage in Terraform configs
Provision an Azure Blob Storage account and container, make sure it can only be reached from within the same VNET as the application (using a private link), and set the appropriate environment variables in the application config to allow the application to connect to the storage instance.

goldblatt:
### Description
Adds terraform support for private blob storage and then passes that into the server. 

### Checklist

### Issue(s)
Fixes #1772


cloud/azure/modules/app/main.tf
```diff
@@ -12,6 +18,47 @@ resource "azurerm_virtual_network" "civiform_vnet" {
...
+  account_replication_type = "LRS"
+}
+
+data "http" "myip" {
+  url = "https://ipv4.icanhazip.com"
+}
+
+# adding the bypass/ip_rules is a workaround for azure's firewall
+# if we don't add this we can't add files to this container
+resource "azurerm_storage_account_network_rules" "example" {
```
bion, line 45: nit (blocking): change label from "example" to something else

cloud/azure/modules/app/main.tf
```diff
@@ -12,6 +18,47 @@ resource "azurerm_virtual_network" "civiform_vnet" {
...
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+}
+
+data "http" "myip" {
+  url = "https://ipv4.icanhazip.com"
+}
+
+# adding the bypass/ip_rules is a workaround for azure's firewall
+# if we don't add this we can't add files to this container
```
bion, line 44: > can't add files to this container
what does it mean to add files to the container?

bion (approved):

cloud/azure/modules/app/main.tf
goldblatt: yeah good point let me do a lil renaming here. i couldn't add the container (which i think is like a bucket in s3) to the storage account

PR #1860, opened 2022-02-01
Upgrade deps: sbt and play-ebean

bion:
Upgrades:

- [x] sbt (the project's build tool) upgraded 1.3.13 -> 1.6.2
- [x] play-ebean (the plugin that integrates ebean into Play) upgraded 6.0.0 -> 6.2.0-RC4
- [x] ebean (the project's ORM) upgraded 11.45.1 -> 12.8.1 (implicitly via [play-ebean](https://github.com/playframework/play-ebean#releases))

SBT versions higher than 1.3.13 (current project version) are not compatible with play-ebean 6.0.0 (current project version) so these need to be upgraded together.

outstanding issues:

- [x] fix the ebean deprecation warnings
- [x] fix compile loop problem

closes https://github.com/seattle-uat/civiform/issues/1942

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -15,9 +15,7 @@ jobs:
       - name: Build test container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -t civiform --cache-from docker.io/civiform/civiform:latest ./
-      - name: sbt clean
-        run: docker run -v /var/run/docker.sock:/var/run/docker.sock civiform clean
+        run: docker build -t civiform ./
       - name: Run tests
         run: docker run -v /var/run/docker.sock:/var/run/docker.sock civiform test
 
@@ -31,7 +29,7 @@ jobs:
       - name: Build test app container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -t civiform --cache-from docker.io/civiform/civiform:latest ./
+        run: docker build -t civiform ./
       - name: Build browser testing container
         env:
           DOCKER_BUILDKIT: 1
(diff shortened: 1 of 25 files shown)


shanemc-goog:
LGTM but may want to hold off pending the prod deployment this week

shanemc-goog (approved):

universal-application-tool-0.0.1/project/plugins.sbt
```diff
@@ -1,6 +1,6 @@
 // The Play plugin
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.8")
-addSbtPlugin("com.typesafe.sbt" % "sbt-play-ebean" % "6.0.0")
+addSbtPlugin("com.typesafe.play" % "sbt-play-ebean" % "6.2.0-RC4")
```
dcsobral, line 3: Is this change required? Both the sbt version change and this are the ones that seem most likely to be associated with the build problem, so doing only one of them would be good for diagnosis.

bion: Unfortunately they need to be made together. The newer `play-ebean` versions don't work with sbt 1.3.13 and require at least 1.4y. That being said it may be worth trying a smaller jump for sbt than 1.6.1.

.github/workflows/pr.yaml
```diff
@@ -10,12 +10,10 @@ jobs:
     steps:
       - name: check out pr branch
         uses: actions/checkout@v2
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
       - name: Build test container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -t civiform --cache-from docker.io/civiform/civiform:latest ./
+        run: DOCKER_BUILDKIT=1 docker build -t civiform ./
```
shanemc-goog, line 16: this would appear to be redundant with the env right above... Is this WAI and just unfamiliar to me or odd enough we should comment it?

shanemc-goog (approved):

.github/workflows/pr.yaml
bion: done

PR #1863, opened 2022-02-01
Enable manually triggering CI workflow

bion:
Allows manually triggering the CI workflow from the GitHub web UI. See https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -3,6 +3,9 @@ name: ci
 on:
   push:
     branches: main
+
+  # Setting this enables manually triggering workflow in the GitHub UI
+  # see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yaml
```diff
@@ -3,6 +3,7 @@ name: ci
 on:
   push:
     branches: main
+  workflow_dispatch: {}
```
shanemc-goog, line 6: look at that magic... Can you comment why this unreadable blob is here which will also help have it not removed in the future? :D

shanemc-goog (changes requested):

shanemc-goog (approved):

.github/workflows/push.yaml
bion: done

shanemc-goog (approved):

PR #1874, opened 2022-02-03
Allow for a Dev Terraform

goldblatt:
### Description
- Refactors the custom hostname into it's own module 
- Creates a new deploy folder for dev, which does not use shared state (so can individually deploy to your own dev resource group)

### Checklist

### Issue(s)

### DIFF
--- a/cloud/deploys/dev_azure/README.md
+++ b/cloud/deploys/dev_azure/README.md
@@ -7,7 +7,7 @@ The process to deploy is similar
 to the staging process but we do not have to do the shared state set up. It 
 should be `terraform init` and then `terraform apply`
 
-# Local Docker Build to Terrafrom
+# Local Docker Build to Remote Azure Deploy
 If you want to do local onto terraform we build/tag/deploy the docker image 
 and then update the azure app service to point to the local image. 
 
@@ -20,7 +20,7 @@ docker tag <DOCKER_TAG> <DOCKER_USERNAME>/<DOCKER_REPO_NAME>:<DOCKER_TAG>
 docker push <DOCKER_USERNAME>/<DOCKER_REPO_NAME>:<DOCKER_TAG>
 ```
 
-## 2. Update the docker repo for your staging
+## 2. Update the image name/tag for your remote azure deploy
 You can do this one of two ways. Terraform deploy or update via the azure 
 portal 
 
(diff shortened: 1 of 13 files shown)


goldblatt:
confirmed this deploys the staging and can deploy the local dev

zaralouis-sf (approved):
Praise: This is going to be super helpful for testing, thank you! Looks good on the staging website

cloud/deploys/staging_azure/main.tf
```diff
@@ -19,15 +19,23 @@ module "app" {
 
   postgres_admin_login    = var.postgres_admin_login
   postgres_admin_password = var.postgres_admin_password
+  # note that we must use GP tier
+  postgres_sku_name       = "GP_Gen5_2"
```
bion, line 23: question (blocking): should this use the variable/its default rather than be harded code?

cloud/deploys/staging_azure/main.tf
```diff
@@ -19,15 +19,23 @@ module "app" {
...
 
   application_name = var.application_name
   app_secret_key   = var.app_secret_key
   ses_sender_email = var.sender_email_address
+  custom_hostname  = var.custom_hostname
+}
 
-  custom_hostname = var.custom_hostname
+module "custom_hostname" {
+  source              = "../../azure/modules/custom_hostname"
```
bion, line 35: question (non-blockng): is it possible to use an absolute path from the project base directory? nbd if not

bion (changes requested):
Two minor comments, please also add instructions somewhere for deploying a local image to azure staging

cloud/deploys/staging_azure/main.tf
goldblatt: ahh it is one but for staging we def want to use this tier

cloud/deploys/staging_azure/main.tf
goldblatt: as far as i understand it's probably easiest to be local path. if it doesn't start with './ or ../' it assumes it's pulling from a remote repo and we could make a private registry but i think that might be more owrk than it's work https://www.terraform.io/language/modules/sources#modules-in-package-sub-directories

bion: ah, no worries then

cloud/deploys/dev_azure/README.md
```diff
@@ -0,0 +1,40 @@
+# Dev Azure
```
bion, line 1: 🤩

bion (approved):

PR #1875, opened 2022-02-03
Validate deployment variable definitions

bion:
Adds a CI check that runs for each PR and `main` merge to ensure that each deployment variable definition:

1. uses a unique name
2. has the required fields for its type

### DIFF
--- a/cloud/shared/validate_variable_definitions.py
+++ b/cloud/shared/validate_variable_definitions.py
@@ -42,6 +42,10 @@ class ValidateVariableDefinitions():
         if not isinstance(variable_definition.get("secret", None), bool):
             errors.append("Missing 'secret' field.")
 
+        if not isinstance(variable_definition.get("type", None), str):
+            errors.append("Missing 'type' field.")
+            return errors
+
         type_specific_validators = {
             "float": self.validate_float_definition_type,
             "integer": self.validate_integer_definition_type,
@@ -49,7 +53,7 @@ class ValidateVariableDefinitions():
             "enum": self.validate_enum_definition_type,
         }
 
-        validator = type_specific_validators[variable_definition["type"]]
+        validator = type_specific_validators.get(variable_definition["type"], None)
 
         if validator:
             validator(variable_definition, errors)
--- a/cloud/shared/validate_variable_definitions.py
+++ b/cloud/shared/validate_variable_definitions.py
@@ -1,7 +1,18 @@
 import os
-import glob
 import json
 
+# Loads all configuration variable definition files and validates each
+# definition for correctness. Exercised by the accompanying test file
+# which is run for every pull request.
+#
+# Requires that:
+#     - Each variable definition file is referenced in
+#       def load_repo_variable_definitions_files():
+#
+#     - All variables have, at minimum, 'type', 'required', and 'secret' fields
+#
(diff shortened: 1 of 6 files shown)


cloud/shared/validate_variable_definitions.py
```diff
@@ -0,0 +1,76 @@
...
+        return all_errors
+
+    def validate(self, variable_definition):
+        errors = []
+
+        if not isinstance(variable_definition.get("required", None), bool):
+            errors.append("Missing 'required' field.")
+
+        if not isinstance(variable_definition.get("secret", None), bool):
+            errors.append("Missing 'secret' field.")
```
shubha-rajan, line 43: Do you think it's better to explicitly set these fields for every variable or can we default them to False?

cloud/shared/validate_variable_definitions.py
```diff
@@ -0,0 +1,76 @@
...
+
+class ValidateVariableDefinitions():
+    def __init__(self, variable_definitions = {}):
+        self.variable_definitions = variable_definitions
+
+    def load_repo_variable_definitions_files(self):
+        self.variable_definitions = {}
+        cwd = os.getcwd()
+
+        definition_file_paths = glob.glob(cwd + '/cloud/**/variable_definitions.json', recursive=True)
```
shubha-rajan, line 13: Do you expect that there will be more than one `variable_definitions.json` file in for a single cloud provider, or will we define one for each cloud we're deploying to (plus a shared)? Staging and Prod environments should require the same variables, just with different values, right?

cloud/shared/validate_variable_definitions.py
```diff
@@ -0,0 +1,76 @@
...
+        return errors
+
+    def validate_float_definition_type(self, variable_definition, errors):
+        pass
+
+    def validate_integer_definition_type(self, variable_definition, errors):
+        pass
+
+    def validate_string_definition_type(self, variable_definition, errors):
+        pass
```
shubha-rajan, line 72: is there an issue filed for implementing these?

shubha-rajan (approved):
LGTM, just a few non-blocking questions!

cloud/shared/validate_variable_definitions.py
bion: I'd prefer they were explicitly set, having seen a few production incidents that were the result of misspelled key in the configuration getting ignored and the default value causing a problem.

cloud/shared/validate_variable_definitions.py
bion: My guess is we'll have one shared and then one for each provider. The glob here might be a little silly and we should just load them each explicitly as we add them.

> Staging and Prod environments should require the same variables, just with different values, right?

yep!

cloud/shared/validate_variable_definitions.py
bion: No, mostly because I don't think we know quite what we'll want yet.  My hope is that as we add configuration values we'll notice "oh it'd be nice if we could guarantee this integer is between 0 and 10" or somesuch and then come back and add the ability to specify that and validate its config here.

cloud/shared/validate_variable_definitions.py
shubha-rajan: makes sense

PR #1882, opened 2022-02-04
Fix dev dump and restore scripts

bion:
The dump script appeared to work but actually generates a corrupted binary archive, I think because docker might be doing something to stdout output. This change fixes that by mounting the project directory into the docker image and writing the archive file in there instead of from the output from docker.

The line continuation in the restore script was wrong.

### DIFF
--- a/bin/dev-restore-db
+++ b/bin/dev-restore-db
@@ -33,8 +33,9 @@ readonly DUMP_FILE="project/${1}"
 readonly TRUNCATE_DB_CMD="/usr/bin/psql -h db -U postgres -c '${TRUNCATE_SQL}' postgres"
 
 # command for invoking pg_restore with the dev database
-readonly PG_RESTORE_CMD="/usr/bin/pg_restore -h db -U postgres -d postgres"\
-  " < ${DUMP_FILE}"
+
+readonly PG_RESTORE_CMD="/usr/bin/pg_restore -h db -U \
+  postgres -d postgres < ${DUMP_FILE}"
 
 readonly CMD="${TRUNCATE_DB_CMD} && ${PG_RESTORE_CMD}"
 
--- a/bin/dump-dev-program-config
+++ b/bin/dump-dev-program-config
@@ -21,4 +21,4 @@ docker run --rm -it \
     -t versions \
     -t versions_programs \
     -t versions_questions \
-    postgres > /project/dev_db.dump"
+    postgres > /project/dev_programs.dump"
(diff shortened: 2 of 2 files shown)


bin/dev-restore-db
```diff
@@ -33,8 +33,9 @@ readonly DUMP_FILE="project/${1}"
 readonly TRUNCATE_DB_CMD="/usr/bin/psql -h db -U postgres -c '${TRUNCATE_SQL}' postgres"
 
 # command for invoking pg_restore with the dev database
-readonly PG_RESTORE_CMD="/usr/bin/pg_restore -h db -U postgres -d postgres"\
-  " < ${DUMP_FILE}"
+
+readonly PG_RESTORE_CMD="/usr/bin/pg_restore -h db -U \
+  postgres -d postgres < ${DUMP_FILE}"
```
zaralouis-sf, line 38: These file changes are just formatting right? Want to make sure I'm not missing anything

bion: This actually changes behavior somehow. I'm still fuzzy on how line continuations work in strings with bash but I tested this script with the output of the other one and it gives a convincing appearance of working.

bin/dump-dev-program-config
```diff
@@ -7,17 +7,18 @@ source bin/lib.sh
...
+    -w \
+    -Fc \
+    -h db \
+    -U postgres \
+    -t programs \
+    -t questions \
+    -t versions \
+    -t versions_programs \
+    -t versions_questions \
+    postgres > /project/dev_db.dump"
```
zaralouis-sf, line 24: Nit: I think you need to update this name in the docs at the top of this file

bion: aha, reverted to the original name. thanks!

bin/dump-dev-program-config
```diff
@@ -7,17 +7,18 @@ source bin/lib.sh
...
-  -Fc \
-  -h db \
-  -U postgres \
-  -t programs \
-  -t questions \
-  -t versions \
-  -t versions_programs \
-  -t versions_questions \
-  postgres > dev_programs.dump
+  /bin/bash -c "pg_dump \
```
zaralouis-sf, line 14: Why was this changed from /usr/bin? And why does the other file still use usr/bin ?

bion: tbh mostly habit and tradition, at least for bash.

system binaries are mostly in `/bin` whereas user binaries often go in `/usr/bin`. it's common to invoke the system version of bash and afaik on most linux distros `/bin/bash` and `/usr/bin/bash` are actually the same inode.

in the case of `pg_dump`, it's userland program rather than a system utility, so it's unlikely to ever be in `/bin`

zaralouis-sf (approved):

PR #1883, opened 2022-02-04
Makes a bastion host & Script to access the bastion host

### LINKED ISSUE #1896: Create a bastion instance and script to allow users to connect to our staging subnet
- helpful for a way to access the postgres database directly
- somewhat needed to allow for user testing

goldblatt:
### Description
Will need to add something that can push more than on ssh key onto the instance. Butttt was able to ssh onto the machine and grab the stuff correctly. Need to confirm security is set up right. 

### Checklist

### Issue(s)
Fixes #1896


cloud/azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  source_image_reference {
+    publisher = "Canonical"
+    offer     = "UbuntuServer"
+    sku       = "16.04-LTS"
+    version   = "latest"
+  }
+
+  admin_ssh_key {
+    username = "adminuser"
+    # ssh-keygen -t rsa -b 4096 -C "sgoldblatt@google.com" -f $HOME/.ssh/bastion
```
bion, line 97: question (non-blocking): your work email intentionally included?

bion (changes requested):

cloud/azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,100 @@
...
+
+  os_disk {
+    caching              = "ReadWrite"
+    storage_account_type = "Standard_LRS"
+  }
+
+  source_image_reference {
+    publisher = "Canonical"
+    offer     = "UbuntuServer"
+    sku       = "16.04-LTS"
```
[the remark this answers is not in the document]
goldblatt: i think the latest one 18.04 (did a az vm image list --output table to list out the available images)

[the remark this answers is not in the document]
bion: ah, let's do whatever latest LTS is available

cloud/azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,100 @@
+# create a public IP for so we can ssh to the bastion
+resource "azurerm_public_ip" "bastion_pip" {
```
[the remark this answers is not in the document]
goldblatt: yes it was!

cloud/azure/modules/bastion/main.tf
goldblatt: noo no was just trying to just keep that command around it should be out now

cloud/azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  security_rule {
+    name                       = "allow-ssh-all"
+    priority                   = 100
+    direction                  = "Inbound"
+    access                     = "Allow"
+    protocol                   = "Tcp"
+    source_port_range          = "*"
+    destination_port_range     = "22"
+    source_address_prefix      = "*"
+    destination_address_prefix = "*"
```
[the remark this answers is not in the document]
goldblatt: i don't think think this is possible as we assign the public IP to this bastion instance. i could set aside a prefix and then only allow those but im not sure if that's what you mean

cloud/deploys/azure/bin/db-connection
```diff
@@ -0,0 +1,27 @@
+# example: ./db-connection -g sgdev -d civiform-artistic-halibut -k bastion
```
bion, line 1: suggestion (blocking): for both scripts

- quote all strings (see https://google.github.io/styleguide/shellguide.html)
- line length limit 80 chars
- add shebang at the top
- add `# DOC: ...` comments along with `...../bin/help` script, can copy from root repo `bin/help`
- validate flag values, at minimum for presence

cloud/deploys/azure/bin/db-connection
```diff
@@ -0,0 +1,27 @@
...
+while getopts g:d:k flag
+do
+    case "${flag}" in
+        g) resource_group=${OPTARG};;
+        d) database_name=${OPTARG};;
+        k) key_name=${OPTARG};;
+    esac
+done
+ssh-keygen -t rsa -b 4096 -f $HOME/.ssh/${key_name}
```
bion, line 10: question (blocking): i wonder if it would be better to store the key in the project root directory and instruct the user to destroy it after they're done? thoughts?

bion (changes requested):

cloud/deploys/azure/bin/db-connection
goldblatt: yeah i think maybe the other option would be to check if it exists then if it does already just skip the generation step. not sure if it really matters if the key gets deleted

cloud/deploys/azure/bin/db-connection
goldblatt: hmmm should these live in that repo since the help script imports some stuff into it

goldblatt: okay i think i should have grabbed them all

cloud/deploys/azure/bin/db-connection
goldblatt: alright i did the like skip if it already exists option

cloud/deploys/azure/bin/db-connection
goldblatt: a lil duplicated code here, but i think since the goal is for this to be in a fully different repo it's probably okay-ish

cloud/azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,91 @@
+# Make sure the bastion is within the vnet of the database
+resource "azurerm_subnet" "bastion_subnet" {
+  name                 = "bastion_subnet"
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.vnet_name
+  address_prefixes = [
+    "10.0.6.0/24"
```
shubha-rajan, line 7: question (blocking) should this be a variable? I think we made the other subnet prefixes variables, so let's keep it consistent.

cloud/deploys/bin/azure/db-connection
```diff
@@ -0,0 +1,52 @@
...
+do
+    case "${flag}" in
+        g) resource_group=${OPTARG};;
+        d) database_name=${OPTARG};;
+        k) key_name=${OPTARG};;
+    esac
+done
+
+if [[ ! -z "${resource_group}" ]] || [[ ! -z "${database_name}" ]]; then
+  echo "arguments -g and -d must be provided"
```
bion, line 16: suggestion (blocking): print to stderr for error messages

cloud/deploys/bin/azure/get-bastion-commands
```diff
@@ -0,0 +1,37 @@
...
+do
+    case "${flag}" in
+        g) resource_group=${OPTARG};;
+        d) database_name=${OPTARG};;
+        k) key_name=${OPTARG};;
+    esac
+done
+
+if [[ ! -z "${resource_group}" ]] || [[ ! -z "${database_name}" ]]; then
+  echo "arguments -g and -d must be provided"
```
bion, line 16: suggestion (blocking): print to stderr for error messages

cloud/deploys/azure/bin/db-connection
bion: I don't think this code will ever be in a fully different repo since it's for our permanent staging server. I'd recommend adding `source bin/lib.sh` at the top, will be less need for duplication then.

bion: Even if we do want to switch to dogfooding the deployment system for our staging servers, we want as much as possible to keep shared code in the mainline repo and delegate to it from the forked deployment code. Scripts like this are a good example of the things we want to keep consistent.

This is making me wonder if these bin scripts should be somewhere other than `cloud/deploy/*` since they shouldn't be specific to a single deployment.

cloud/deploys/bin/azure/get-bastion-commands
bion: can use `out::error` from the lib code

cloud/deploys/bin/azure/get-bastion-commands
```diff
@@ -0,0 +1,37 @@
...
+  exit 1
+fi
+
+echo "Getting the bastion vm's ip address"
+vm_ip_address=$(az network public-ip show \
```
bion, line 21: suggestion (blocking): constants should be `readonly VM_IP_ADDRESS`

cloud/deploys/bin/azure/db-connection
```diff
@@ -0,0 +1,52 @@
...
+    -g "${resource_group}" \
+    -n "${database_name}" \
+    --query "fullyQualifiedDomainName" | tr -d '"')
+
+# echo "Adding the public key to the bastion vm"
+az vm user update \
+    -u adminuser \
+    -g "${resource_group}" \
+    -n "${resource_group}-bstn-vm" \
+    --ssh-key-value "$(< ~/.ssh/${key_name}.pub)"
```
bion, line 43: question (blocking): how can we clean these up after access is finished so we don't have permanent ingress to the DB?

cloud/deploys/bin/azure/get-bastion-commands
```diff
@@ -0,0 +1,37 @@
...
+    -g "${resource_group}" \
+    -n "${database_name}" \
+    --query "fullyQualifiedDomainName" | tr -d '"')
+
+echo "ssh command: ssh -i $HOME/.ssh/${key_name} adminuser@${vm_ip_address}"
+echo "psql command: \
+    export DEBIAN_FRONTEND='noninteractive'; \
+    yes | sudo apt-get update > /dev/null; \
+    yes | sudo apt-get install postgresql-client > /dev/null; \
+    psql -h ${postgres_host} -d postgres -U psqladmin@${postgres_host}"
```
bion, line 37: suggestion (non-blocking): could reduce duplication between this and `db-connection` by having `db-connection` use `eval` to execute these commands

bion (changes requested):

cloud/deploys/azure/bin/db-connection
goldblatt: i mean this one is kind of specific since it includes azure commands?

cloud/deploys/bin/azure/db-connection
goldblatt: well i guess is this something we need to do?

cloud/deploys/azure/bin/db-connection
bion: i'd expect they're generic to any azure deployment though?

cloud/deploys/bin/azure/db-connection
bion: yes

zaralouis-sf (commented):
I don't have any feedback beyond what others have said so I'll defer to those with more expertise but thank you for documenting everything so well! It's super helpful

cloud/deploys/azure/bin/db-connection
goldblatt: yup! but doesn't that mean they should live in this cloud directory not the bin?

cloud/azure/modules/bastion/main.tf
goldblatt: done!

cloud/deploys/bin/azure/db-connection
goldblatt: done!

cloud/deploys/bin/azure/get-bastion-commands
goldblatt: cool re arranged a lot of the files

cloud/deploys/bin/azure/db-connection
goldblatt: cool okay i remove it from the users machine

cloud/deploys/azure/bin/db-connection
goldblatt: and i also destory the key afterwards so i will resolve this

cloud/deploys/bin/azure/get-bastion-commands
goldblatt: done

cloud/deploys/bin/azure/bastion
```diff
@@ -0,0 +1,74 @@
...
+# Arguments:
+#   1: the resource group name
+#   2: the key name to use to connect to
+#######################################
+function bastion::update_bastion_ssh_keys() {
+  echo "az vm user update \
+  -u adminuser \
+  -g "${1}" \
+  -n "${1}-bstn-vm" \
+  --ssh-key-value \"$(< ~/.ssh/${2}.pub)\""
```
bion, line 61: nit: indent continuation

cloud/deploys/bin/azure/bastion
```diff
@@ -0,0 +1,74 @@
...
+#######################################
+# Get the command to connect to postgres
+# Arguments:
+#   1: the postgres host to connect to 
+#######################################
+function bastion::get_connect_to_postgres_command() {
+  echo "export DEBIAN_FRONTEND='noninteractive'; \
+  yes | sudo apt-get update > /dev/null; \
+  yes | sudo apt-get install postgresql-client > /dev/null; \
+  psql -h ${1} -d postgres -U psqladmin@${1}"
```
bion, line 73: nit: indent

cloud/deploys/bin/azure/bastion
```diff
@@ -0,0 +1,74 @@
+#! /usr/bin/env bash
```
bion, line 1: nit: to follow the style guide, library shell files should have `.sh` file extension

cloud/deploys/bin/azure/bastion
```diff
@@ -0,0 +1,74 @@
...
+# Get the ip for the bastion VM 
+# Arguments:
+#   1: the resource group name
+#######################################
+function bastion::get_vm_ip() {
+  readonly VM_IP_ADDRESS=$(az network public-ip show \
+    -g "${1}" \
+    -n "${1}-ip" \
+    --query "ipAddress" | tr -d '"')
+  echo "${VM_IP_ADDRESS}"
```
bion, line 13: suggestion (non-blocking):
```suggestion
  echo $(az network public-ip show \
    -g "${1}" \
    -n "${1}-ip" \
    --query "ipAddress" | tr -d '"')
```

cloud/deploys/bin/azure/bastion.sh
```diff
@@ -20,11 +19,10 @@ function bastion::get_vm_ip() {
...
 function bastion::get_postgres_host() {
-  readonly POSTGRES_HOST=$(az postgres server show \
-  -g "${1}" \
-  -n "${2}" \
-  --query "fullyQualifiedDomainName" | tr -d '"')
-  echo "${POSTGRES_HOST}"
+  echo $(az postgres server show \
+    -g "${1}" \
+    -n "${2}" \
+    --query "fullyQualifiedDomainName" | tr -d '"')
```
bion, line 25: question (blocking): is the echo required or could this be
```suggestion
  az postgres server show \
    -g "${1}" \
    -n "${2}" \
    --query "fullyQualifiedDomainName" \
    | tr -d '"'
```

cloud/deploys/azure/bin/db-connection
bion: yes, you're right it should. i think the thing i'm trying to sort out is how the repo can share shell lib code between the development and deploy scripts so we don't have multiple copies of functions that do basic things like formatting output.

we don't need to figure that out in this PR though.

cloud/deploys/bin/azure/bastion.sh
```diff
@@ -0,0 +1,72 @@
...
+#######################################
+# Get the ip for the bastion VM 
+# Arguments:
+#   1: the resource group name
+#######################################
+function bastion::get_vm_ip() {
+  echo $(az network public-ip show \
+    -g "${1}" \
+    -n "${1}-ip" \
+    --query "ipAddress" | tr -d '"')
```
bion, line 12: suggestion (blocking): the echo isn't necessary here i don't think

cloud/deploys/bin/azure/bastion.sh
```diff
@@ -0,0 +1,72 @@
...
+#   1: the resource group name
+#   2: the key name to use to connect to
+#######################################
+function bastion::update_bastion_ssh_keys() {
+  echo "az vm user update \
+    -u adminuser \
+    -g "${1}" \
+    -n "${1}-bstn-vm" \
+    --ssh-key-value \"$(< ${2}.pub)\""
+}
```
bion, line 60: question (blocking): in a previous iteration there were separate scripts for printing the commands one could run and actually running the commands, which is why I suggested echoing strings like this and using `eval`. If we don't need the former though I recommend not using `echo` / `eval` since it makes the code harder to read and reason about.

Is it still the intention to provide the commands in both ways?

bion (changes requested):

cloud/deploys/bin/azure/bastion.sh
goldblatt: ahh yeah the other ones are echo'd then eval'd but they could just be eval'd let me try that

cloud/deploys/azure/bin/db-connection
goldblatt: ahh gotcha gotcha

goldblatt: made an issue https://github.com/seattle-uat/civiform/issues/1897

cloud/deploys/bin/azure/bastion.sh
goldblatt: yeah i changed them so that just the get ssh command and the get postgres command are print/eval and the other ones are just executed!

goldblatt: if i don't do the ssh commands in the echo eval it doesn't forward the std in/std out (which I think means something about the command is a lil off)

cloud/azure/modules/bastion/main.tf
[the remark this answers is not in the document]
goldblatt: switched this to always deny and then the bash script sets it to be allow for a specific IP address and then resets it to be deny again after the script is done!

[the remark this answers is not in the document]
bion: Great!

cloud/deploys/bin/azure/db-connection
```diff
@@ -0,0 +1,52 @@
...
+
+if [[ ! "${resource_group}" ]] || [[ ! "${database_name}" ]]; then
+  out::error "arguments -g and -d must be provided"
+  exit 1
+fi
+
+echo "add my ip address to the security group for ssh"
+bastion::allow_ip_security_group "${resource_group}"
+
+readonly KEY_FILE="$HOME/.ssh/${key_name}"
```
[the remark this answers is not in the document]
goldblatt: Had to do this on my local, pushed a fix.

cloud/deploys/bin/azure/db-connection
```diff
@@ -0,0 +1,52 @@
...
+    k) key_name=${OPTARG};;
+  esac
+done
+
+if [[ ! "${resource_group}" ]] || [[ ! "${database_name}" ]]; then
+  out::error "arguments -g and -d must be provided"
+  exit 1
+fi
+
+echo "add my ip address to the security group for ssh"
```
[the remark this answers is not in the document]
goldblatt: Had to do this on my local, pushed a fix.

cloud/deploys/bin/azure/bastion.sh
```diff
@@ -0,0 +1,114 @@
...
+# Get the command to connect to postgres
+# Arguments:
+#   1: the postgres host to connect to 
+#######################################
+function bastion::get_connect_to_postgres_command() {
+  db_password=$(bastion::get_pg_password)
+  echo "export DEBIAN_FRONTEND='noninteractive'; \
+    yes | sudo apt-get update > /dev/null; \
+    yes | sudo apt-get install postgresql-client > /dev/null; \
+    PGPASSWORD='${db_password}' psql -h ${1} -d postgres -U psqladmin@${1}"
```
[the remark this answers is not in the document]
goldblatt: Had to do this on my local, pushed a fix.

cloud/deploys/bin/azure/bastion.sh
```diff
@@ -0,0 +1,114 @@
...
+}
+
+#######################################
+# Connect to the bastion
+# Arguments:
+#   1: the ip of the vm you will connect to
+#   2: the key name to use to connect to
+#   3: command to run after ssh'ing
+#######################################
+function bastion::bastion_ssh_connect() {
```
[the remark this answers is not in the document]
goldblatt: done

cloud/deploys/bin/azure/bastion.sh
goldblatt: okay altered this a decent amount, resolving as i think it should be fixed

cloud/deploys/bin/azure/db-connection
```diff
@@ -2,27 +2,31 @@
 source lib/out.sh
 source azure/bastion.sh
 
-# DOC: Connect to the database via the bastion
-# example: azure/db-connection -g sgdev -d civiform-artistic-halibut
-key_name="bastion"
-while getopts g:d:k flag
+# DOC: Connect to the database via the bastion. 
+# DOC: Required to pass in the resource group (-g), the database name (-d) and the keyvault name (-v). 
+# DOC: example: azure/db-connection -g sgdev -d civiform-artistic-halibut -v sgdev-civiform-kv
```
bion, line 7: suggestion (blocking): i think the help script only works if there's a single line that starts with `DOC: ` which is a little embarrassing 😅, please collapse these to a single line or fix the script

bion (approved):
approved with one last small thing

cloud/deploys/bin/azure/db-connection
goldblatt: i tried this and this lil hack does work!

PR #1893, opened 2022-02-07
Don't include screens without questions in total block count

bion:
To determine if an application is ok to submit or requires more answers from the applicant, we count the number of active blocks and compare that with the number of active and completed blocks. This caused a bug where blocks that have no questions would count towards the total of active but incomplete blocks, making it impossible to complete the application.

This change alters the logic for tallying completed blocks by not including blocks that have no questions in the tally.


bion:
closing in favor of an alternative approach: https://github.com/seattle-uat/civiform/pull/1894

PR #1907, opened 2022-02-11
Add aria attributes for applicant form UI

bion:
- [ ] form inputs use `aria-describedby` to reference questions help text
- [ ] required inputs are marked `aria-required`
- [ ] form inputs with invalid contents or missing non-optional contents are marked as `aria-invalid`
- [ ] form inputs with invalid contents or missing non-optional contents reference error messages with `aria-errormessage`


bion:
likely taking a different approach to improving a11y here

PR #1908, opened 2022-02-11
Update permissions and expand README for keyvault

shubha-rajan:
Fixes permissions and expands README for setting up keyvault integration. The biggest change was creating the key vault in a separate resource group, since in a real deployment, it will be set up before the terraform-managed resources.

### DIFF
--- a/cloud/azure/main.tf
+++ b/cloud/azure/main.tf
@@ -7,7 +7,7 @@ terraform {
     }
     random = {}
   }
-  backend "azurerm" {}
+  # backend "azurerm" {}
   required_version = ">= 0.14.9"
 }
 
--- a/cloud/azure/main.tf
+++ b/cloud/azure/main.tf
@@ -7,7 +7,7 @@ terraform {
     }
     random = {}
   }
-  # backend "azurerm" {}
+  backend "azurerm" {}
   required_version = ">= 0.14.9"
 }
 
--- a/cloud/azure/modules/app/main.tf
+++ b/cloud/azure/modules/app/main.tf
@@ -57,7 +57,7 @@ resource "azurerm_storage_account_network_rules" "files_storage_rules" {
 
 data "azurerm_key_vault" "civiform_key_vault" {
   name                = var.key_vault_name
-  resource_group_name = azurerm_resource_group.rg.name
+  resource_group_name = var.key_vault_resource_group
 }
 
 data "azurerm_key_vault_secret" "postgres_password" {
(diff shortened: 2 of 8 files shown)


cloud/azure/main.tf
```diff
@@ -7,10 +7,10 @@ terraform {
     }
     random = {}
   }
-  backend "azurerm" {}
+  # backend "azurerm" {}
```
goldblatt, line 10: can we uncomment this out

cloud/azure/modules/app/main.tf
```diff
@@ -57,7 +57,7 @@ resource "azurerm_storage_account_network_rules" "files_storage_rules" {
 
 data "azurerm_key_vault" "civiform_key_vault" {
   name                = var.key_vault_name
-  resource_group_name = azurerm_resource_group.rg.name
+  resource_group_name = var.key_vault_resource_group
```
goldblatt, line 60: why did we add a specific key vault resource group?

cloud/deploys/staging_azure/README.md
```diff
@@ -94,11 +94,18 @@ portal the resource group name is 'tfstate' and the storage_account_name is
 Before applying the Terraform configuration, you'll need to make sure that Azure Key Vault is
 properly configured to store the `postgres-password` and `app-secret-key` secrets. To do this, run the command `az keyvault list` to list all the key vaults in your project. Find the key vault that stores the secrets for Civiform. Then run `az keyvault secret list --vault-name=[your vault name]`. The `postgres-password` and `app-secret-key` secrets should be listed.
 
-If the secrets are not listed, you'll need to [create a key vault](https://docs.microsoft.com/en-us/azure/key-vault/general/quick-create-cli) if you haven't already. Then set the secrets:
+If the secrets are not listed, you'll need to [create a key vault](https://docs.microsoft.com/en-us/azure/key-vault/general/quick-create-cli) if you haven't already. Make a note of the resource group that is used. Next, run
```
goldblatt, line 97: would add a comment about making these rbac when it asks for the security specifications

goldblatt (changes requested):

cloud/azure/modules/app/main.tf
shubha-rajan: This is because of a weird circular dependency where we have to create a resource group before we create key vault, but we have to create the key vault before running terraform, which creates a resource group.

cloud/deploys/staging_azure/README.md
shubha-rajan: The RBAC specification is covered in the next step where you run 
```
az keyvault update --name=[your keyvault name] --enable-rbac-authorization
```

goldblatt (approved):

zaralouis-sf (approved):

PR #1959, opened 2022-02-17
Scaffolding for deployment command delegation

bion:
Puts the pieces in place to delegate deployment commands from the deployment repo to executables in the main repo and ensures the delegated command has access to shell script library code. See also https://github.com/civiform/civiform-deploy/pull/2

The order of operations for a command run from the deployment repo is:

1. load the deployment repo's shell library code
1. load the deployment repo's civiform configuratoin
1. resolve the git revision for the main repo the command should run in
1. exec the `cloud/shared` version of the command in the main repo
1. resolve the cloud provider (report an error if the cloud provider is invalid or missing)
1. exec the provider-specific version of the command in the main repo
1. load the shell library code in bin/lib.sh
1. load the cloud provider-specific shell library code 
1. perform the command

### DIFF
--- a/cloud/azure/bin/deploy
+++ b/cloud/azure/bin/deploy
@@ -1,0 +1,5 @@
+#! /usr/bin/env bash
+
+KEEP_ORIGINAL_PWD=true source "${PROJECT_BASE_DIR}/bin/lib.sh"
+source "${PROJECT_BASE_DIR}/cloud/shared/bin/lib.sh"
+source "${PROJECT_BASE_DIR}/cloud/azure/bin/lib.sh"
--- a/cloud/azure/bin/deploy
+++ b/cloud/azure/bin/deploy
@@ -3,3 +3,5 @@
 KEEP_ORIGINAL_PWD=true source "${PROJECT_BASE_DIR}/bin/lib.sh"
 source "${PROJECT_BASE_DIR}/cloud/shared/bin/lib.sh"
 source "${PROJECT_BASE_DIR}/cloud/azure/bin/lib.sh"
+
+# TODO: implement actual Azure deployment logic
--- a/cloud/shared/bin/validate_cloud_provider
+++ b/cloud/shared/bin/validate_cloud_provider
@@ -1,0 +1,13 @@
+#! /usr/bin/env bash
+
+readonly VALID_CLOUD_PROVIDERS=(azure)
+
+if [[ "${CLOUD_PROVIDER}" = "" ]]; then
+  >&2 echo "Must specify CLOUD_PROVIDER configuration variable."
+  exit 1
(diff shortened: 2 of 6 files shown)


cloud/azure/bin/deploy
```diff
@@ -3,3 +3,5 @@
 KEEP_ORIGINAL_PWD=true source "${PROJECT_BASE_DIR}/bin/lib.sh"
 source "${PROJECT_BASE_DIR}/cloud/shared/bin/lib.sh"
 source "${PROJECT_BASE_DIR}/cloud/azure/bin/lib.sh"
+
+# TODO: implement actual Azure deployment logic
```
shubha-rajan, line 7: question: is this where we would validate all the azure-specific variables?

cloud/shared/bin/validate_cloud_provider
```diff
@@ -0,0 +1,13 @@
+#! /usr/bin/env bash
+
+readonly VALID_CLOUD_PROVIDERS=(azure)
```
shubha-rajan, line 3: question (blocking): is this file going to remain the source of truth for what `CLOUD_PROVIDER` values are valid? I thought we were using an enum type variable definition here to enforce the accepted values:

https://github.com/seattle-uat/civiform/blob/715234ed2b9930f90edc84e3af855815903e244e/cloud/shared/variable_definitions.json#L6

shubha-rajan (changes requested):

cloud/azure/bin/deploy
bion: yeah, this is where the "figure out which code to execute" dance ends and we actually get down to business, probably starting with config validation

cloud/shared/bin/validate_cloud_provider
bion: I thought so too, but I was writing/thinking about this it just seemed easier to have this one config var validated here and then load everything else after we've figured out which cloud provider we're working with. Otherwise we'll end up loading a bunch of the same code once before we've resolved the cloud provider and then a second time after.

If you'd rather take it the other direction I'm totally up for seeing what that could look like though.

shubha-rajan: That makes sense. A thought I had was that we could validate all of the shared variables at the same time we validate the cloud provider, and then validate the provider-specific ones in the provider-specific deploy script.

bion: > A thought I had was that we could validate all of the shared variables at the same time we validate the cloud provider, and then validate the provider-specific ones in the provider-specific deploy script.

When you get into config validation, if that approach ends up looking better please feel free to restructure this a bit!

PR #1962, opened 2022-02-17
Deployment log shell lib

bion:
Adds shell library code for working with the deployment log. So far only supports:

- [x] initializing the log file
- [x] recording deployment success and failure
- [x] initialize function does not wipe out the previous log file if one already exists


cloud/azure/bin/lib/azure_log.sh
```diff
@@ -0,0 +1,123 @@
+#! /usr/bin/env bash
+
+readonly AZURE_LOG_STORAGE_ACCOUNT_NAME="civiformdeploymentlog"
```
shubha-rajan, line 3: question (blocking): should we have this be read in from config, since account names have to be globally unique?

cloud/azure/bin/lib/azure_log.sh
```diff
@@ -0,0 +1,123 @@
...
+  az storage blob exists \
+    --auth-mode login \
+    --account-name "${AZURE_LOG_STORAGE_ACCOUNT_NAME}" \
+    --container-name "${AZURE_LOG_CONTAINER_NAME}" \
+    --name "${AZURE_LOG_FILE_NAME}" \
+    | grep -q true
+}
+
+#######################################
+# Initialize the deplooy log in Azure blob storage.
```
shubha-rajan, line 65: nit:
```suggestion
# Initialize the deploy log in Azure blob storage.
```

cloud/shared/bin/lib/log.sh
```diff
@@ -0,0 +1,97 @@
...
+#######################################
+function log::initialized() {
+  log::ensure_log_file_fetched
+  log::validate_token "${1}"
+
+  echo "$(log::timestamp) INITIALIZED SUCCESS ${1}" >> "${LOG_TEMPFILE}"
+}
+
+#######################################
+# Check that at least two arguments exist and do not contain spaces.
```
shubha-rajan, line 48: question (blocking): should we check arguments after the 2nd one (if there are any) to make sure they don't contain spaces? Right now it looks like it's only checking the first two.

cloud/shared/bin/lib/log.sh
```diff
@@ -0,0 +1,97 @@
...
+# Writes an error message to stderr and exits with status 1 upon failure.
+# Args:
+#   1..2: Strings suitable as tokens in the deployment log.
+#######################################
+function log::check_deploy_args() {
+  for i in {1..2}; do
+    if [[ -z "${!i}" || -z "${!i}" ]]; then
```
shubha-rajan, line 55: question: what does this condition check for? It looks like it's checking the same thing on both sides of the `||`

shubha-rajan (changes requested):
Just a few clarifying questions

cloud/azure/bin/lib/azure_log.sh
bion: Ah dang, yeah I didn't realize the storage account name had to be globally unique

cloud/shared/bin/lib/log.sh
bion: woops, yep that's a mistake, it only needs to check one.

it's checking to ensure each arg has a value

cloud/shared/bin/lib/log.sh
bion: Done

cloud/azure/bin/lib/azure_log.sh
bion: removed

cloud/shared/bin/lib/log.sh
```diff
@@ -0,0 +1,102 @@
...
+#######################################
+# Record a deployment success event in the log.
+# Args:
+#   1: The tag of the deployed server binary.
+#   2: Identifier for the user who initiated the command.
+# Globals read:
+#   LOG_TEMPFILE
+#######################################
+function log::deploy_succeeded() {
+  log::ensure_log_file_fetched
```
shubha-rajan, line 12: question (blocking): How do we make sure that the existing log file is up to date? For example, lets say we fetched a log file and while we were appending to it, but before we upload, someone else logged some stuff and uploaded it. and then we'd end up overwriting anything they logged. To minimize that risk, should we make sure to fetch an up-to date log file immediately before any log operation?

bion: Valid concern, moreover if two entities are running commands in overlapping timeframes we'll likely have other problems such as inconsistent cloud state.

IMO the canonical solution here is to introduce some manner of expiring mutex lock on the log and therefore the overall cloud state so that if a command is one progress, another can't be run and interfere with the outcome.

There are a few ways we could approach this, but none of them are very simple. I'm inclined to add a section on this to our design doc and revisit it once initial deployments are working.

shubha-rajan (approved):

PR #1973, opened 2022-02-22
Remove firewall for Azure Storage

shubha-rajan:
We recently discovered that this firewall prevents SAS URLs from working from any IP address that isn't explicitly allowlisted (like an end user's IP address) which is not what we want. We should secure our storage account using user permissions instead.


bion (approved):

goldblatt (approved):
cool cool did we verify this like doesn't allow just access but requires the presigned stuff to access!

shubha-rajan:
> cool cool did we verify this like doesn't allow just access but requires the presigned stuff to access!

Yup! verified this by trying to access the bucket while logged into another azure account

PR #1975, opened 2022-02-23
Add no-store header

bion:
The no-store response directive indicates that any caches of any kind (private or shared) should not store this response.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-store

before merging:
- [ ] check to see if this prevents caching of the stylesheet


bion:
this prevents the stylesheet from caching

PR #1986, opened 2022-02-23
Fix file download and enable Azurite browser tests

### LINKED ISSUE #1686: File uploading supports Azure backend

### LINKED ISSUE #1796: All browser tests pass in both the AWS and Azure environments

zaralouis-sf:
### Description
Fixes file downloads and enables browser tests for Azurite

This ended up being a larger change than expected because I added the field "original file name" to answer data. When we upload a file to Azure blob storge, we generate a UUID to set as the file key, because file name isn't immediately available when we needed. This means that during applicant review, users would see a generated file key instead of the name of their file. To fix this, I added a field to the answer data, which is then shown to the user. 

Another major change is in the browser tests. Civiform uses SES to send emails for both Azure and AWS deployments. Unfortunately, this means I had to start both localstack and Azurite for all the browser tests to work. If you think there is a better way to have all the browser tests functional please let me know but this is the only solution I could find so far.

### Issue(s)
Fixes #1796; Fixes #1686


universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -152,9 +158,14 @@ private ContainerTag renderQuestionSummary(
...
     if (data.fileKey().isPresent()) {
-      String fileLink =
-          controllers.routes.FileController.show(applicantId, data.fileKey().get()).url();
-      answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      try {
+        String encodedUrl = URLEncoder.encode(data.fileKey().get(), StandardCharsets.UTF_8);
+        String fileLink = controllers.routes.FileController.show(applicantId, encodedUrl).url();
+        answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      } catch (Exception e) {
+        LOG.error("Attempted to download file with invalid download URL");
```
bion, line 166: suggestion (blocking): please log more info about the invalid download URL e.g. the applicant ID and the encoded URL

bion (changes requested):

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -152,9 +158,14 @@ private ContainerTag renderQuestionSummary(
 
     ContainerTag answerContent;
     if (data.fileKey().isPresent()) {
-      String fileLink =
-          controllers.routes.FileController.show(applicantId, data.fileKey().get()).url();
-      answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      try {
+        String encodedUrl = URLEncoder.encode(data.fileKey().get(), StandardCharsets.UTF_8);
+        String fileLink = controllers.routes.FileController.show(applicantId, encodedUrl).url();
```
shubha-rajan, line 163: question (blocking): Would this encoded string need to be decoded in the FileController.show method?

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -152,9 +158,14 @@ private ContainerTag renderQuestionSummary(
 
     ContainerTag answerContent;
     if (data.fileKey().isPresent()) {
-      String fileLink =
-          controllers.routes.FileController.show(applicantId, data.fileKey().get()).url();
-      answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      try {
+        String encodedUrl = URLEncoder.encode(data.fileKey().get(), StandardCharsets.UTF_8);
```
shubha-rajan, line 162: nit: rename this variable to `encodedFileKey` since it's not the full URL

shubha-rajan (changes requested):

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
zaralouis-sf: Done!

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
zaralouis-sf: Done.

goldblatt (approved):
not the best at reviewing the java code but this eems good to me!

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -258,25 +259,40 @@ public ApplicantProgramBlocksController(
...
+              String applicantFileUploadQuestionOriginalFileNamePath =
+                  fileUploadQuestion.getOriginalFileNamePath().toString();
+
+              ImmutableMap.Builder<String, String> builder =
+                  new ImmutableMap.Builder<String, String>();
+              builder.put(applicantFileUploadQuestionKeyPath, key.get());
+              if (originalFileName.isPresent()) {
+                builder.put(
+                    applicantFileUploadQuestionOriginalFileNamePath, originalFileName.get());
+              }
```
shubha-rajan, line 291: praise: great work figuring out how to handle optionally adding originalFileNames to the formData!

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -39,6 +43,8 @@
 
   private final ApplicantLayout layout;
 
+  private static Logger LOG = LoggerFactory.getLogger(ApplicantProgramSummaryView.class);
```
shubha-rajan, line 46: suggestion (blocking): All-caps is only for things that are final. Logger names should be lowercase. Rename to `logger`

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -152,9 +158,18 @@ private ContainerTag renderQuestionSummary(
...
     ContainerTag answerContent;
     if (data.fileKey().isPresent()) {
-      String fileLink =
-          controllers.routes.FileController.show(applicantId, data.fileKey().get()).url();
-      answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      try {
+        String encodedFileKey = URLEncoder.encode(data.fileKey().get(), StandardCharsets.UTF_8);
+        String fileLink = controllers.routes.FileController.show(applicantId, encodedFileKey).url();
+        answerContent = a().withHref(fileLink).withClasses(Styles.W_2_3);
+      } catch (Exception e) {
```
shubha-rajan, line 165: Can we catch a more specific exception here?

shubha-rajan (changes requested):

bion:
> Unfortunately, this means I had to start both localstack and Azurite for all the browser tests to work. If you think there is a better way to have all the browser tests functional please let me know but this is the only solution I could find so far.

IMO this isn't unfortunate, it's just accurately emulating how CiviForm runs in this environment

bion (changes requested):
Great work on this!

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
zaralouis-sf: My IDE kept complaining that there's no exception thrown, so I took another look at the documentation: https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html. The exception is only thrown from the function if the encoding is not supported, but we're using a standard encoding, UTF 8, which is "guaranteed to be available on every implementation of the Java platform" (https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html) so I don't think that is ever going to be a possibility. I am going to remove the try catch block.

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -258,25 +259,40 @@ public ApplicantProgramBlocksController(
...
-              ImmutableMap<String, String> formData =
-                  ImmutableMap.of(applicantFileUploadQuestionKeyPath, key.get());
+                      .createFileUploadQuestion();
+
+              String applicantFileUploadQuestionKeyPath =
+                  fileUploadQuestion.getFileKeyPath().toString();
+              String applicantFileUploadQuestionOriginalFileNamePath =
+                  fileUploadQuestion.getOriginalFileNamePath().toString();
+
+              ImmutableMap.Builder<String, String> builder =
```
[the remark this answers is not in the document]
zaralouis-sf: Done, let me know if that name is okay.

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -275,17 +282,18 @@ private boolean showBlock(Block block) {
...
-      case FILEUPLOAD:
-        FileUploadQuestion fileUploadQuestion = question.createFileUploadQuestion();
-        if (!fileUploadQuestion.isAnswered()) {
-          return Optional.empty();
-        }
-        return fileUploadQuestion.getFileKeyValue();
-      default:
-        return Optional.empty();
+  private Optional<String> getFileKey(Optional<FileUploadQuestion> question) {
+    if (question.isEmpty() || !question.get().isAnswered()) {
```
shanemc-goog, line 286: Optional: you can make this method body the following.


return question.filter(FileUploadQuestion::isAnswered)
    .map(FileUploadQuestion::getFileKeyValue)
    .orElseGet(() -> Optional.empty());

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -18,6 +18,7 @@
 
   private final ApplicantQuestion applicantQuestion;
   private Optional<String> fileKeyValue;
+  private Optional<String> originalFileNameValue;
```
shanemc-goog, line 21: Given the usage these should be `Optional<Optional<String>>`.  and initialized in the constructor.  Add a comment to indicate that this is serving double duty as a singleton load of the value.

I don't think we have a standard on using nulls or not, but I don't think we _need_ to use nulls here.

bin/run-browser-test-env
```diff
@@ -39,25 +44,26 @@ Options:
 --aws   Use the AWS emulator (Localstack) for simulating cloud services,
         cannot be used with --azure flag. This is the default.
 
---azure Use the Azure emulator (Azurite) for simluating cloud services,
-        cannot be used with --aws flag.
+--azure Use the Azure emulator (Azurite) for simluating cloud services.
```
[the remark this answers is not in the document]
zaralouis-sf: Done

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -102,6 +116,8 @@ public String getAnswerString() {
     if (getFilename().isEmpty()) {
       return "-- NO FILE SELECTED --";
     }
-    return String.format("-- %s UPLOADED (click to download) --", getFilename().get());
+    return String.format(
+        "-- %s UPLOADED (click to download) --",
+        getOriginalFileName().isPresent() ? getOriginalFileName().get() : getFilename().get());
```
shanemc-goog, line 121: pull the ?: out into a var for readability.

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -152,8 +154,8 @@ private ContainerTag renderQuestionSummary(
 
     ContainerTag answerContent;
```
shanemc-goog, line 155: Unrelated, optional: You can put a final here to ensure all parts of the if/else set this before the first read and only once.  A neat use of final.

shanemc-goog (changes requested):

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
shanemc-goog: optional.

universal-application-tool-0.0.1/app/controllers/FileController.java
```diff
@@ -46,7 +48,9 @@ public FileController(
               if (!fileKey.contains(String.format("applicant-%d", applicantId))) {
                 return notFound();
               }
-              return redirect(storageClient.getPresignedUrlString(fileKey));
+              return redirect(
+                  storageClient.getPresignedUrlString(
+                      URLDecoder.decode(fileKey, StandardCharsets.UTF_8)));
```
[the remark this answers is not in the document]
zaralouis-sf: The encoding/decoding is only changing "/" -> "2f%" from what I can tell. We had to add this because the URL was wasn't being recognized because everything that was supposed to be "/" was instead "2f%"

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -258,25 +259,40 @@ public ApplicantProgramBlocksController(
 
               Optional<String> bucket = request.queryString("bucket");
               Optional<String> key = request.queryString("key");
+
+              // Original file name is only set for Azure, where we have to generate a UUID when
```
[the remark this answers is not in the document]
zaralouis-sf: Trial and error honestly. This has to be done because of the way the Azure Blob Storage SDK works, compared to how S3 does. Future contributors will have to make a decision based on how GCP file upload works, I believe

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
zaralouis-sf: I've updated, do you mind taking a look if I implemented it the way you had in mind? I got a little lost in all the optionals haha

universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramSummaryView.java
zaralouis-sf: I didn't know you could do that! Super cool

bin/run-browser-test-env
```diff
@@ -39,25 +44,26 @@ Options:
...
+      truth::enable continuous_integration
+      ;;
+
+    "--azure")
+      emulators::set_azurite_emulator_vars
+      truth::enable both_emulators
+      ;;
+
+    "--aws")
+      # We don't need this check for Azurite because we want both emulators!
```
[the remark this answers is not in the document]
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/controllers/FileController.java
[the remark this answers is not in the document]
bion: Even though it works both ways consider decoding it before doing anything else so that all current and future logic here is dealing with the decoded version

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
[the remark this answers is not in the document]
shubha-rajan: For the record, GCS supports POST uploads so we won't have to do this weird workaround. Azure is the odd one out here. https://cloud.google.com/storage/docs/xml-api/post-object-forms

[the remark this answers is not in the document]
zaralouis-sf: Would that be worth adding as a comment for future contributors or is that over kill?

universal-application-tool-0.0.1/app/controllers/FileController.java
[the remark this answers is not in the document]
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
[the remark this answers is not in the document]
bion: > Would that be worth adding as a comment for future contributors or is that over kill?

the more context/info in the comments the better, honestly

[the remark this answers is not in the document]
zaralouis-sf: Added! Thanks for that info and link @shubha-rajan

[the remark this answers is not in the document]
shanemc-goog: "Future contributors will have to make a decision based on how GCP file upload works"

How would they know they need to make that decision, or the decision to make?  

IE: Should we have a wikipage on FileUpload and include this?

You're in effect designing a general way that this works now, so the more we can set future folks up for success the better

universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
shanemc-goog: I learned through review like this :D but this is effectively the "Streams" feature added in Java8.  

Collections can all become streams with .stream(), and Optional added some of the filter/map concepts too.

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html

universal-application-tool-0.0.1/app/controllers/applicant/ApplicantProgramBlocksController.java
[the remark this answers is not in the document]
zaralouis-sf: Yes!! This sounds like an excellent subject for docs day I'll add it to the list

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -17,11 +17,16 @@
...
 
   private final ApplicantQuestion applicantQuestion;
-  private Optional<String> fileKeyValue;
-  private Optional<String> originalFileNameValue;
+  private Optional<Optional<String>> fileKeyValue;
+  private Optional<Optional<String>> originalFileNameValue;
 
   public FileUploadQuestion(ApplicantQuestion applicantQuestion) {
     this.applicantQuestion = applicantQuestion;
+    // This value is serving double duty as a singleton load of the value.
```
shanemc-goog, line 25: Suggestion:

I'd put this on the class members bove since they have the semantics instead of this one location that sets them.

Also, in looking at other code I was cleaning them up by also adding Cache to the end of the name to help convey their multi purpose use.

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -63,22 +68,24 @@ public ValidationErrorMessage fileRequiredMessage() {
   }
 
   public Optional<String> getFileKeyValue() {
-    if (fileKeyValue != null) {
-      return fileKeyValue;
+    if (fileKeyValue.get().isPresent()) {
```
shanemc-goog, line 71: remove the get here as you're seeing if the cached value has been set yet. The code here is assuming it is (which should [throw an exception](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#get()) on the first call also, I'd expect some test to fail)

Adding Cache to this name I think will help too

```
if (fileKeyValueCache.isPresent()) {
  return fileKeyValueCache.get();
}
```

shanemc-goog (changes requested):

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
zaralouis-sf: In order to get this working, looks like I'll have to change at least 6 test files. Should I make these changes in a separate PR or is that fine with you? I'm just a little worried about this PR becoming too big but happy to make the fixes if you'd like

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
shanemc-goog: yep! looks good.

universal-application-tool-0.0.1/app/services/applicant/question/FileUploadQuestion.java
shanemc-goog: What you have here is what I expected.

If you thought I meant for you to add it in the other unrelated uses, I didn't.  I was conveying that I had been adding Cache to those names to add clarity.

shanemc-goog (approved):

bion (approved):

shubha-rajan (approved):

PR #1988, opened 2022-02-24
Pin NPM version and use latest package versions

bion:
- [x] pins NPM version to 8.5.1 which will stop the annoying package-lock.json diffs from constantly getting generated
- [x] upgrades the docker image packages at build time. this improves the security of the image
- [x] adds `bin/dev-bash` script for convenience of getting a bash shell running in a dev container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,10 @@ ENV SBT_URL "https://github.com/sbt/sbt/releases/download/v${SBT_VERSION}/sbt-${
 
 RUN set -o pipefail && \
     apk update && \
+    apk add --upgrade apk-tools && \
+    apk upgrade --available && \
     apk add --no-cache --update bash wget npm git openssh && \
+    npm install -g npm@8.5.1 && \
     mkdir -p "${SBT_HOME}" && \
     wget -qO - "${SBT_URL}" | tar xz -C "${INSTALL_DIR}" && \
     echo -ne "- with sbt ${SBT_VERSION}\n" >> /root/.built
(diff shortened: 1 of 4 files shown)


shanemc-goog (approved):

bin/dev-bash
```diff
@@ -0,0 +1,16 @@
+#! /usr/bin/env bash
+
+# DOC: Open a bash shell in the dev environment container.
+
+source bin/lib.sh
+
+bin/pull-image
+
+docker run -it --rm \
```
goldblatt, line 9: i guess a lil nit if we are doing all the inputs on new lines maybe these should be new lined too?

Dockerfile
```diff
@@ -17,7 +17,10 @@ ENV SBT_URL "https://github.com/sbt/sbt/releases/download/v${SBT_VERSION}/sbt-${
 
 RUN set -o pipefail && \
     apk update && \
+    apk add --upgrade apk-tools && \
+    apk upgrade --available && \
     apk add --no-cache --update bash wget npm git openssh && \
+    npm install -g npm@8.5.1 && \
```
goldblatt, line 23: nit? why not pin to @latest ?

bion: on the theory that explicit versioning is better than implicit? i could be convinced otherwise

bin/dev-bash
bion: done

bion: done

Dockerfile
goldblatt: nah this is all good just curious

goldblatt (approved):

PR #2020, opened 2022-03-04
Add missing Downs for adding account.authority_id

shanemc-goog:
Add missing Downs for adding account.authority_id

Can't rollback through this evolution due to missing downs.


bion (approved):

shanemc-goog:
I think we can't do this since it's already made it to staging.  locally my server sees this as an error and won't start

shanemc-goog:
I'll try adding it as a new evolution.

Issue #2001, opened 2022-03-01
Add SAML emulator container to docker-compose and add tests


shubha-rajan:
Continue work from #1994 
We need both unit and browser tests

goldblatt:
Shane is doing similar work for OIDC but we need tests for SAML

shanemc-goog:
The work for OIDC is a bit less than what we need for SAML.

The key thing for both is having a provider we can use in tests.  Someone else [found a public one, and wrote a shim layer](https://github.com/seattle-uat/civiform/tree/main/test-support) which is used in browser and unit tests.   The work I'm doing is just to make it work for local dev.

goldblatt:
we have unit tests from #2025 will need just the browser tests

PR #2025, opened 2022-03-04
Add unit tests for SAML profile adapter

shubha-rajan:
Towards #2001

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
+++ b/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
@@ -165,11 +165,10 @@ public class SamlCiviFormProfileAdapter extends AuthenticatorProfileCreator {
 
   private String extractAttributeFromArrayList(SAML2Profile profile, String attr) {
     ArrayList attributeArray = profile.getAttribute(attr, ArrayList.class);
+    if (attributeArray != null) { return ""; };
     StringJoiner sj = new StringJoiner(" ");
-    if (attributeArray != null) {
-      for (Object s : attributeArray) {
-        sj.add((String) s);
-      }
+    for (Object s : attributeArray) {
+      sj.add((String) s);
     }
     return sj.toString();
   }
--- a/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
+++ b/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
@@ -44,8 +44,8 @@ public class SamlCiviFormProfileAdapter extends AuthenticatorProfileCreator {
     super();
     this.profileFactory = Preconditions.checkNotNull(profileFactory);
     this.applicantRepositoryProvider = Preconditions.checkNotNull(applicantRepositoryProvider);
-    this.saml2Client = Preconditions.checkNotNull(client);
-    this.saml2Configuration = Preconditions.checkNotNull(configuration);
+    this.saml2Client = client;
+    this.saml2Configuration = configuration;
   }
 
   @Override
--- a/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
+++ b/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
@@ -165,7 +165,9 @@ public class SamlCiviFormProfileAdapter extends AuthenticatorProfileCreator {
 
   private String extractAttributeFromArrayList(SAML2Profile profile, String attr) {
     ArrayList attributeArray = profile.getAttribute(attr, ArrayList.class);
-    if (attributeArray != null) { return ""; };
+    if (attributeArray == null) {
+      return "";
+    }
     StringJoiner sj = new StringJoiner(" ");
     for (Object s : attributeArray) {
       sj.add((String) s);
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
```diff
@@ -166,8 +166,10 @@ public CiviFormProfileData mergeCiviFormProfile(
   private String extractAttributeFromArrayList(SAML2Profile profile, String attr) {
     ArrayList attributeArray = profile.getAttribute(attr, ArrayList.class);
     StringJoiner sj = new StringJoiner(" ");
-    for (Object s : attributeArray) {
-      sj.add((String) s);
+    if (attributeArray != null) {
```
shanemc-goog, line 169: Do the null check right after you get this on line 1 of the method; then `return "";` early if it's null.

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
```diff
@@ -166,8 +166,10 @@ public CiviFormProfileData mergeCiviFormProfile(
   private String extractAttributeFromArrayList(SAML2Profile profile, String attr) {
     ArrayList attributeArray = profile.getAttribute(attr, ArrayList.class);
     StringJoiner sj = new StringJoiner(" ");
```
shanemc-goog, line 168: optional: Missed it on the original CL but this can also be

```
return attributeArray.stream()
     .map(s -> (String) s)
     .collect(Collectors.joining(" "));
```

universal-application-tool-0.0.1/test/auth/ProfileMergeTest.java
```diff
@@ -32,14 +36,26 @@ public UserRepository get() {
                 return repository;
               }
             });
+    samlProfileAdapter =
+        new SamlCiviFormProfileAdapter(
+            null,
```
shanemc-goog, line 41: Can you inline comment these nulls per the style guide which says to doc ambiguous params at the call site.

/* paramName */ null

shanemc-goog (changes requested):

zaralouis-sf:
Praise: Thank you for doing this and increasing our test coverage! I think other than Shane's comments this looks good to me!

zaralouis-sf (approved):

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
shubha-rajan: I tried this, but the only way I could get it to work was to do:
```
    return (String) attributeArray.stream()
        .map(s -> (String) s)
        .collect(Collectors.joining(" "));
```
Idk how I feel about casting to String twice, and to me the StringJoiner looks cleaner. Thoughts?

universal-application-tool-0.0.1/test/auth/ProfileMergeTest.java
shubha-rajan: done. I fixed the idcsProfileAdapter too.

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
shubha-rajan: done

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
shanemc-goog: Don't worry about it! thanks for trying.

I'm guessing something needs coercing, maybe Collectors<String> etc, but just guessing.

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
```diff
@@ -165,11 +165,12 @@ public CiviFormProfileData mergeCiviFormProfile(
 
   private String extractAttributeFromArrayList(SAML2Profile profile, String attr) {
     ArrayList attributeArray = profile.getAttribute(attr, ArrayList.class);
+    if (attributeArray == null) {
```
shanemc-goog, line 168: I think we should prefer Optional.ofNullable(.....) and isPresent checks in general.

universal-application-tool-0.0.1/test/auth/ProfileMergeTest.java
```diff
@@ -27,8 +27,8 @@ public void setupFactory() {
     UserRepository repository = instanceOf(UserRepository.class);
     idcsProfileAdapter =
         new IdcsProfileAdapter(
-            null,
-            null,
+            /* configuration */ null,
```
shanemc-goog, line 30: Please add a = after each name.    We'll be warned if the name is wrong (EG we're not setting what we think we are.)

I just found out that this is enforced by the complier, I thought it was a special tool internally.

shanemc-goog (approved):

shanemc-goog:
Please see the above non blocking comments

universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java
shanemc-goog: Update: Blocking: please use optional.

I discussed with Bion and going forward our code should prefer Optional over null .

shanemc-goog (changes requested):

shanemc-goog (approved):

PR #2026, opened 2022-03-07
Fix java logger naming and keywords.

### LINKED ISSUE #2012: Logger LOG constants should be lowercased
Per the [style guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) loggers should be lowercased but a typical "bug" is to make them CONST_CASE.

We have both [LOG](https://github.com/seattle-uat/civiform/search?q=%22Logger+LOG%22) and [logger](https://github.com/seattle-uat/civiform/search?q=%22Logger+logger%22) style and should fix them so that the bad form doesn't keep getting copied forward.

shanemc-goog:
Make all java loggers be private static final and use lowercase naming as they are not immutable objects per the style guide.

Fixes #2012


bion (approved):
thanks for doing this! this PR made me realize we should probably do more logging than we are right now 😬

PR #2027, opened 2022-03-07
Refactor authentication code

zaralouis-sf:
### Description
See [this doc](https://docs.google.com/document/d/104ZGwQmj33XIHAemWKiFuPJGq5LR4hpA3aNGW9hv7ns/edit#) for more context. Refactor civiform authentication code to make it more extensible for the future. This refactor replaces the various auth client annotations with an AdminAuthClient and ApplicantAuthClient. Those clients implement [IndirectClient](https://www.pac4j.org/apidocs/pac4j/3.1.0/org/pac4j/core/client/IndirectClient.html). We bind those clients to their corresponding client providers in SecurityModule. Currently, the AdminAuthClient is always bound to the AdOidcClientProvider, but in the future this can be easily modified as we onboard more authentication IDPs. The ApplicantAuthClient is bound based on the auth.admin_idp field found in the config.


bion:
Before this merges, it'd be good to build and push it to Seattle staging to test that IDCS and ADFS integration are still working. @gwendolyngoetz would you be up for trying that?

Steps are:

1. checkout this branch
1. build a production image with the staging tag:
```
docker build -f prod.Dockerfile \
  -t public.ecr.aws/t1q6b4h2/universal-application-tool:latest \
  --cache-from civiform/civiform:latest \
  --build-arg BUILDKIT_INLINE_CACHE=1 \
  .
```
3. tell ecs to deploy:
```
REGION=us-west-2

ECSSERVICE=$(aws cloudformation describe-stacks \
    --region ${REGION} \
    --stack-name civiform-staging |
      jq -r '.Stacks[0].Outputs[] | select(.OutputKey == "ECSService") | .OutputValue')

ECSCLUSTER=$(aws cloudformation describe-stacks \
    --region ${REGION} \
    --stack-name civiform-staging |
      jq -r '.Stacks[0].Outputs[] | select(.OutputKey == "ECSCluster") | .OutputValue')

aws ecs update-service \
  --region=${REGION} \
  --cluster "$ECSCLUSTER" \
  --service "$ECSSERVICE" \
  --force-new-deployment
```

This would go a long way toward ensuring this change doesn't cause auth issues for Seattle that would require reverting it.

gwendolyngoetz:
@bion Don't I need to do a docker push on the locally built container?

bion:
@gwendolyngoetz ah yep, after building it `docker push public.ecr.aws/t1q6b4h2/universal-application-tool:latest`

gwendolyngoetz:
I follow the steps, adding the docker push between steps 2 and 3. I am able to log in via both ADFS and IDCS still. Nice work team!

zaralouis-sf:
Thank you so much for your help @gwendolyngoetz!!  This is a much needed sanity check for me

universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -101,175 +95,72 @@ protected void configure() {
...
         new PlayCookieSessionStore(new ShiroAesDataEncrypter(aesKey));
     bind(SessionStore.class).toInstance(sessionStore);
-  }
 
-  @Provides
-  @Singleton
-  protected GuestClient guestClient(ProfileFactory profileFactory) {
-    return new GuestClient(profileFactory);
+    // Default to these options
+    String applicantAuthClient = "adfs";
```
shubha-rajan, line 100: question (blocking): shouldn't the applicant auth client default to IDCS?

universal-application-tool-0.0.1/conf/routes
```diff
@@ -139,11 +139,9 @@ POST    /callback/:client_name       controllers.CallbackController.callback(req
 
 # Log into application
 GET     /loginForm                   controllers.HomeController.loginForm(request: Request, message: java.util.Optional[String])
-GET     /idcsLogin                   controllers.LoginController.idcsLoginWithRedirect(request: Request, redirectTo: java.util.Optional[String])
+GET     /adminLogin                  controllers.LoginController.adminLogin(request: Request)
+GET     /applicantLogin              controllers.LoginController.applicantLogin(request: Request, redirectTo: java.util.Optional[String])
```
shubha-rajan, line 143: praise: these routes look so much cleaner, I love it!!!

universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -4,50 +4,44 @@
...
-import org.pac4j.oidc.client.OidcClient;
-import org.pac4j.oidc.config.OidcConfiguration;
 import org.pac4j.play.CallbackController;
 import org.pac4j.play.LogoutController;
 import org.pac4j.play.http.PlayHttpActionAdapter;
 import org.pac4j.play.store.PlayCookieSessionStore;
 import org.pac4j.play.store.ShiroAesDataEncrypter;
-import org.pac4j.saml.client.SAML2Client;
-import org.pac4j.saml.config.SAML2Configuration;
+import org.springframework.lang.Nullable;
```
shubha-rajan, line 43: question (blocking): did you mean to switch from using `javax.annotation.Nullable` to `org.springframework.lang.Nullable`?

universal-application-tool-0.0.1/app/auth/oidc/IdcsOidcProvider.java
```diff
@@ -0,0 +1,59 @@
...
+  private final ProfileFactory profileFactory;
+  private final Provider<UserRepository> applicantRepositoryProvider;
+
+  @Inject
+  public IdcsOidcProvider(
+      com.typesafe.config.Config configuration,
+      ProfileFactory profileFactory,
+      Provider<UserRepository> applicantRepositoryProvider) {
+    this.configuration = checkNotNull(configuration);
+    this.baseUrl = configuration.getString("base_url");
```
shubha-rajan, line 26: question (blocking): are we validating that `base_url` is set anywhere?

universal-application-tool-0.0.1/app/auth/saml/LoginRadiusSamlProvider.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class LoginRadiusSamlProvider implements Provider<SAML2Client> {
+
+  private final com.typesafe.config.Config configuration;
+  private final ProfileFactory profileFactory;
+  private final Provider<UserRepository> applicantRepositoryProvider;
+  private final String baseUrl;
+
+  @Inject
+  public LoginRadiusSamlProvider(
```
shubha-rajan, line 21: praise: great work writing these provider classes! they look so clean and it's nice to have the logic specific to each IDP separated out into its own file.

shubha-rajan (changes requested):

universal-application-tool-0.0.1/app/auth/oidc/IdcsOidcProvider.java
zaralouis-sf: I realized we actually don't, anywhere in the code base. How do you think I should handle it-- should we throw an exception in these provider classes if it's not set? Would it make more sense to throw an exception from somewhere else? I actually think we have a default setting for it in application.conf so maybe we don't need to check it? Curious your thoughts

universal-application-tool-0.0.1/app/modules/SecurityModule.java
zaralouis-sf: No I definitely did not thank you!

universal-application-tool-0.0.1/app/modules/SecurityModule.java
zaralouis-sf: Ah yes thank you for catching that!

universal-application-tool-0.0.1/app/auth/oidc/IdcsOidcProvider.java
shubha-rajan: It's defaulted to `localhost:9000` right? my worry here is that someone deploying to a staging or production instance might forget to set this value, and then the authentication won't work because the callback is wrong. @bion @shanemc-goog what are your thoughts on this?

universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -101,175 +95,72 @@ protected void configure() {
...
+    AuthIdentityProviderName idpName = AuthIdentityProviderName.forString(applicantIdpName).get();
+
+    switch (idpName) {
+      case LOGIN_RADIUS_APPLICANT:
+        bind(IndirectClient.class)
+            .annotatedWith(ApplicantAuthClient.class)
+            .toProvider(LoginRadiusSamlProvider.class);
+        break;
+      case IDCS_APPLICANT:
+      default:
```
shanemc-goog, line 130: Similar to the other comment, I don't know that we should have a default, all options should be equal I think, regardless of the existing state.

Are you concerned that some of these values won't be set in the existing Seattle deployment?

universal-application-tool-0.0.1/conf/routes
```diff
@@ -139,11 +139,9 @@ POST    /callback/:client_name       controllers.CallbackController.callback(req
 
 # Log into application
 GET     /loginForm                   controllers.HomeController.loginForm(request: Request, message: java.util.Optional[String])
-GET     /idcsLogin                   controllers.LoginController.idcsLoginWithRedirect(request: Request, redirectTo: java.util.Optional[String])
```
shanemc-goog, line 142: We need to be careful with removing existing routes, often times things like this need to be maintained for a release cycle as through the release different versions of the server may be available, as well if a route is effectively loaded into a users browser before a release, they may try to access it after.   That seems like a concern here correct?

This isn't something we have a good plan on afaik.

universal-application-tool-0.0.1/app/controllers/LoginController.java
```diff
@@ -33,88 +30,70 @@
...
+      return login(request, applicantClient);
     }
-    return login(request, loginRadiusClient)
+    return login(request, applicantClient)
         .addingToSession(request, REDIRECT_TO_SESSION_KEY, redirectTo.get());
   }
 
   public Result register(Http.Request request) {
-    String registerUrl = null;
+    String idpProvider;
```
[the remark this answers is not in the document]
zaralouis-sf: IDP is identity provider, but this is the identity provider provider haha. It's the like the SAML client provider, if that makes sense?

universal-application-tool-0.0.1/app/modules/SecurityModule.java
zaralouis-sf: Yes, I feel a lot better if the code defaults to the existing prod options because I'd like to remove the possibility of breaking prod as much as possible. This is what we did for file upload as well, and it felt a lot safer to me.

universal-application-tool-0.0.1/conf/routes
zaralouis-sf: Yes, that is definitely a concern. Should I leave the former routes in this file for now and then have a TODO to clean them up in the future?

universal-application-tool-0.0.1/app/auth/saml/LoginRadiusSamlProvider.java
```diff
@@ -0,0 +1,61 @@
...
+    if (!this.configuration.hasPath("login_radius.keystore_password")
+        || !this.configuration.hasPath("login_radius.private_key_password")
+        || !this.configuration.hasPath("login_radius.api_key")) {
+      return null;
+    }
+
+    String metadataResourceUrl =
+        String.format(
+            "%s?apikey=%s&appName=%s",
+            this.configuration.getString("login_radius.metadata_uri"),
```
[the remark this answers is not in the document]
zaralouis-sf: Done.

universal-application-tool-0.0.1/app/controllers/LoginController.java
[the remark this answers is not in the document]
zaralouis-sf: Ahh actually I was looking at the wrong place! you're right, it is redundant.

universal-application-tool-0.0.1/app/controllers/LoginController.java
```diff
@@ -33,88 +30,70 @@
...
-      registerUrl = config.getString("idcs.register_uri");
+      idpProvider = config.getString("auth.applicant_idp");
     } catch (ConfigException.Missing e) {
-      // leave it as null / empty.
+      // Default to IDCS
+      idpProvider = AuthIdentityProviderName.IDCS_APPLICANT.toString();
     }
-    if (Strings.isNullOrEmpty(registerUrl)) {
-      return badRequest("Registration is not enabled.");
+    // This register behavior is specific to IDCS. Because this is only being called when we know
```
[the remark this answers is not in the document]
zaralouis-sf: I'm not super clear on why we only need it for IDCS, I just know that we don't for LoginRadius. I think it's just how different IDPs work

universal-application-tool-0.0.1/app/controllers/LoginController.java
```diff
@@ -33,88 +30,70 @@
...
 
   public Result register(Http.Request request) {
-    String registerUrl = null;
+    String idpProvider;
     try {
-      registerUrl = config.getString("idcs.register_uri");
+      idpProvider = config.getString("auth.applicant_idp");
     } catch (ConfigException.Missing e) {
-      // leave it as null / empty.
+      // Default to IDCS
```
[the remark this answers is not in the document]
zaralouis-sf: As discussed before, to reduce the chances of breaking prod

universal-application-tool-0.0.1/app/modules/SecurityModule.java
```diff
@@ -101,175 +95,71 @@ protected void configure() {
...
-  @Provides
-  @Singleton
-  protected GuestClient guestClient(ProfileFactory profileFactory) {
-    return new GuestClient(profileFactory);
+    String applicantAuthClient = "idcs";
+
+    try {
+      applicantAuthClient = configuration.getString("auth.applicant_idp");
+    } catch (ConfigurationException ignore) {
+      // Default to IDCS.
```
zaralouis-sf, line 104: Just fyi, I had to bring this comment back because ErrorProne was complaining that "this exception should not be ignored" when I had a completely empty catch block

universal-application-tool-0.0.1/app/auth/saml/LoginRadiusSamlProvider.java
```diff
@@ -58,4 +67,19 @@ public SAML2Client get() {
     client.init();
     return client;
   }
+
+  private Optional<String> formatMetadataResourceUrl() {
+    String metadataResourceUrl;
```
shanemc-goog, line 72: put this in the try since you're returning from there.

universal-application-tool-0.0.1/app/modules/SecurityModule.java
shanemc-goog: Ah, looks like this is in internal/old exception that the public tool doesn't use.

https://github.com/google/error-prone/issues/1654

universal-application-tool-0.0.1/app/controllers/LoginController.java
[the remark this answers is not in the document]
shanemc-goog: Suggestion:

restructure/rephrase this similar to the following to return early in the common case, and draw out the exception more.

boolean isIDCS = idp.equals(AuthIdentityProviderName.IDCS_APPLICANT.toString());

if(!isIDCS) {
  return login(request, applicantClient);
}

// IDCS requires a special registration flow.
.......


-----------
This might also indicate that the IDCS code blog should be a method that is called here to compartmentalize things.

shanemc-goog (changes requested):

universal-application-tool-0.0.1/conf/routes
shanemc-goog: Yeah, if that works with your PR leaving them and cleaning up later is safest.

universal-application-tool-0.0.1/app/auth/AdminAuthClient.java
```diff
@@ -6,7 +6,7 @@
 
 /**
  * AdminAuthClient is the annotation for the auth client responsible for admin authentication. This
- * client will implement IndirectClient.
+ * client must implement IndirectClient -> {@link org.pac4j.core.client.IndirectClient}.
```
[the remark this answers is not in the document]
zaralouis-sf: Sorry I am a javadoc novice I guess! How does this look?

universal-application-tool-0.0.1/conf/routes
zaralouis-sf: So, looking at it now, I am realizing that in order to leave the routes, I'll also have to leave the deprecated code in LoginController. Is that something I should do? Wondering if we should discuss this specific code in our next daily sync.

zaralouis-sf: I would have to leave all the methods in LoginController that are referenced by the old routes for the code not to break. Sorry, not sure if my above comment made total sense.

universal-application-tool-0.0.1/app/auth/saml/LoginRadiusSamlProvider.java
```diff
@@ -58,4 +67,19 @@ public SAML2Client get() {
...
+    try {
+      metadataResourceUrl =
+          String.format(
+              "%s?apikey=%s&appName=%s",
+              configuration.getString("login_radius.metadata_uri"),
+              configuration.getString("login_radius.api_key"),
+              configuration.getString("login_radius.saml_app_name"));
+      return Optional.of(metadataResourceUrl);
+    } catch (IllegalFormatException | NullPointerException e) {
```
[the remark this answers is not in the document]
zaralouis-sf: Yes good catch! Should be fixed.

universal-application-tool-0.0.1/app/controllers/LoginController.java
```diff
@@ -33,68 +30,65 @@
...
+    // Because this is only being called when we know IDCS is available, this route should
+    // technically
+    // never happen.
+    if (!isIDCS) {
+      return login(request, applicantClient);
     }
-    return login(request, loginRadiusClient)
-        .addingToSession(request, REDIRECT_TO_SESSION_KEY, redirectTo.get());
+
+    return idcsRegister(request);
```
zaralouis-sf, line 86: @shanemc-goog do these changes look like what you had in mind?

universal-application-tool-0.0.1/app/controllers/LoginController.java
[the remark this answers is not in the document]
zaralouis-sf: I restructured and tagged you in the comment, let me know what you think

universal-application-tool-0.0.1/app/controllers/LoginController.java
shanemc-goog: Per the team discussion I believe you'll have the existing url routes internally use your new controller.

zaralouis-sf: Should be fixed, let me know what you think

universal-application-tool-0.0.1/app/controllers/LegacyRoutesController.java
```diff
@@ -11,14 +13,14 @@
 public class LegacyRoutesController {
 
   public Result idcsLoginWithRedirect(Http.Request request, Optional<String> redirectTo) {
```
shanemc-goog, line 15: Should the request be used in the body? I see it's dropped here.  Is the intent to always redirect to the new handler rather than handle the old route through new refactored flow?

shubha-rajan: If I understand correctly, I think the `redirect` method used automatically passes in the request context. Here's the docs on it: https://www.playframework.com/documentation/2.8.x/JavaRouting#Reverse-routing

shubha-rajan (approved):

universal-application-tool-0.0.1/app/controllers/LegacyRoutesController.java
zaralouis-sf: Yes Shubha that was my understanding as well! Thanks for the docs! I can add them as a comment if you think it would be helpful Shane?

shanemc-goog: Great! thanks yall!

shanemc-goog (approved):

PR #2031, opened 2022-03-08
Move azure TF main into template

bion:
- [x] update README
- [x] refactor staging_azure
- [x] backend_vars to ensure ARM access key are available for every TF command
- [x] ensure calls to terraform have correct `-chdir` (all), `-state` (apply), `-backend-config` (init)
- [x] don't default `civiform_applicant_auth_protocol` app/main.tf
- [x] script for running TF commands to set up the new env


cloud/deploys/staging_azure/staging.tfvars
```diff
@@ -0,0 +1,18 @@
+postgres_admin_login="psqladmin"
```
goldblatt, line 1: this is the wrong quote

goldblatt (approved):

shubha-rajan (approved):

PR #2041, opened 2022-03-10
Correctly set next ID for MultiOptionQuestionForm

### LINKED ISSUE #2013: New options to questions are not saved
After a CiviForm admin edits a question to have more options, whether that be adding options to a dropdown or a radio button, that selection is able to be chosen by the applicant initially but is not actually saved when the applicant goes to the review page or the admin exports the user data.

**Here's an example:**

1. The admin initially created a question "What is <$this> relationship to you?" with the following options: 
- Self
- Parent
- Child

2. They then added more options so the full option list is:
- Self
- Parent
- Child
- Grandparent
- Other

3. An applicant sees the question "What is <$this> relationship to you?" with the following options:
- Self
- Parent
- Child
- Grandparent
- Other

But if they select Grandparent, in the Review page it says that they selected Child. When the Admin exports the program responses, they also see Child as that applicant's response. The same goes for if they select Other, in the Review page and in the CSV export it shows them selecting Self.

**Workaround:**
Seattle is able to workaround this issue by instead of editing a question, deleting it altogether.

zaralouis-sf:
### Description
Issue #2013: After a CiviForm admin edits a question to have more options, whether that be adding options to a dropdown or a radio button, that selection is able to be chosen by the applicant initially but is not actually saved when the applicant goes to the review page or the admin exports the user data.

In MultiOptionQuestionForm, there's a field called "nextAvailableId". This field is set with "setNextAvailableId", which is never actually called. To fix this, I set the nextAvailableId every time the question options are updated. 

To test, I recreated the question linked in that bug. 

Here, I am picking Grandparent, which was added later on as the 4th option: 
image

I do the same with this edited radio button question:
image

As you can see here, Grandparent now shows up in both places instead of "Child", which was happening previously
image

### Issue(s)
Fixes #2013

### DIFF
--- a/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
@@ -143,27 +143,31 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
       predicateBuilder.setMaxChoicesAllowed(getMaxChoicesAllowed());
     }
 
-    ImmutableList.Builder<QuestionOption> questionOptions = ImmutableList.builder();
+    ImmutableList.Builder<QuestionOption> questionOptionsBuilder = ImmutableList.builder();
     Preconditions.checkState(
         this.optionIds.size() == this.options.size(),
         "Option ids and options are not the same size.");
 
     // Note: the question edit form only sets or updates the default locale.
     for (int i = 0; i < options.size(); i++) {
-      questionOptions.add(
+      questionOptionsBuilder.add(
           QuestionOption.create(
               optionIds.get(i), i, LocalizedStrings.withDefaultValue(options.get(i))));
     }
+    setNextAvailableId(options.size());
     for (int i = 0; i < newOptions.size(); i++) {
-      questionOptions.add(
+      questionOptionsBuilder.add(
           QuestionOption.create(
               nextAvailableId.orElse(0L) + i,
               options.size() + i,
               LocalizedStrings.withDefaultValue(newOptions.get(i))));
     }
+    ImmutableList<QuestionOption> questionOptions = questionOptionsBuilder.build();
+
+    setNextAvailableId(questionOptions.size());
 
     return super.getBuilder()
-        .setQuestionOptions(questionOptions.build())
+        .setQuestionOptions(questionOptions)
         .setValidationPredicates(predicateBuilder.build());
   }
 
--- a/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
@@ -22,7 +22,10 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
   private List<String> options;
   // Options added to the list during the edit.
   private List<String> newOptions;
+  // The IDs of each option are not expected to be in any particular order.
   private List<Long> optionIds;
+  // This value is the max existing ID + 1. The max ID will not necessarily be the last one in the
+  // optionIds list,we do not store options by order of their IDs.
   private OptionalLong nextAvailableId;
   private OptionalInt minChoicesRequired;
   private OptionalInt maxChoicesAllowed;
@@ -148,7 +151,7 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
         this.optionIds.size() == this.options.size(),
         "Option ids and options are not the same size.");
 
-    if (this.optionIds.size() > 0) {
+    if (!optionIds.isEmpty()) {
       setNextAvailableId(this.optionIds.get(0) + 1);
     }
 
@@ -157,6 +160,8 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
       questionOptionsBuilder.add(
           QuestionOption.create(
               optionIds.get(i), i, LocalizedStrings.withDefaultValue(options.get(i))));
+      // The IDs are not guaranteed to be in any type of order, so doing this ensures that we find
+      // the largest ID in the list and accurately set the next largest.
       if (optionIds.get(i) > nextAvailableId.orElse(0L)) {
         setNextAvailableId(optionIds.get(i) + 1);
       }
--- a/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
@@ -162,7 +162,7 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
               optionIds.get(i), i, LocalizedStrings.withDefaultValue(options.get(i))));
       // The IDs are not guaranteed to be in any type of order, so doing this ensures that we find
       // the largest ID in the list and accurately set the next largest.
-      if (optionIds.get(i) > nextAvailableId.orElse(0L)) {
+      if (optionIds.get(i) == nextAvailableId.orElse(0L)) {
         setNextAvailableId(optionIds.get(i) + 1);
       }
     }
--- a/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
+++ b/universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
@@ -14,7 +14,9 @@ import services.question.QuestionOption;
 import services.question.types.MultiOptionQuestionDefinition;
 import services.question.types.QuestionDefinitionBuilder;
 
-/** Superclass for all forms for updating a multi-option question. */
+/**
+ * Superclass for all forms for updating a multi-option question.
+ */
 public abstract class MultiOptionQuestionForm extends QuestionForm {
 
   // Caution: This must be a mutable list type, or else Play's form binding cannot add elements to
@@ -64,9 +66,9 @@ public abstract class MultiOptionQuestionForm extends QuestionForm {
         this.nextAvailableId =
             OptionalLong.of(
                 qd.getOptionsForLocale(LocalizedStrings.DEFAULT_LOCALE).stream()
-                        .mapToLong(LocalizedQuestionOption::id)
-                        .max()
-                        .getAsLong()
+                    .mapToLong(LocalizedQuestionOption::id)
+                    .max()
+                    .getAsLong()
                     + 1);
       }
     } catch (TranslationNotFoundException e) {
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/test/controllers/admin/AdminQuestionControllerTest.java
```diff
@@ -245,6 +248,41 @@ public void update_redirectsOnSuccess() {
...
+    // We can only update draft questions, so save this in the DRAFT version.
+    testQuestionBank.maybeSave(definition, LifecycleStage.DRAFT);
+
+    ArrayList<String> newOptions = new ArrayList<>();
+    newOptions.add("cookie");
+    newOptions.add("mint");
+    newOptions.add("pistachio");
+
+    DropdownQuestionForm questionForm = new DropdownQuestionForm(definition);
+    questionForm.setNewOptions(newOptions);
```
shanemc-goog, line 278: suggestion: You can inline and remove the one-off array with ImmutableList.of("cookie", "mint"...)'

shanemc-goog (changes requested):
Can you try to add a unit-test to MultiOptionQuestionFormTest.java too please.

Focused testing will help clean up a pattern doing most of the testing at a very high level.

shanemc-goog:
Suggestion: Please update the PR title with the change so it's more informative when viewed amongst all PRs

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
```diff
@@ -143,27 +143,40 @@ public QuestionDefinitionBuilder getBuilder() {
...
     }
 
-    ImmutableList.Builder<QuestionOption> questionOptions = ImmutableList.builder();
+    ImmutableList.Builder<QuestionOption> questionOptionsBuilder = ImmutableList.builder();
     Preconditions.checkState(
         this.optionIds.size() == this.options.size(),
         "Option ids and options are not the same size.");
 
+    if (this.optionIds.size() > 0) {
+      setNextAvailableId(this.optionIds.get(0) + 1);
```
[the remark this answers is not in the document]
swatkat1: May be it will be easier to use :  setNextAvailableId(this.optionIds.size()) because the Ids start from 0.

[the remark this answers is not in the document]
zaralouis-sf: Yes that was the original reasoning, but now that I am really thinking about it I'm wondering if it was completely unncessary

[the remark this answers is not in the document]
zaralouis-sf: I'm not sure setting the next available as the size will always work because IDs aren't guaranteed to start at 0. I actually am thinking based on these comments that maybe I should just delete this line. It is starting to feel completely unncessary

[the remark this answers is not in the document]
zaralouis-sf: Okay, went ahead and removed those lines. Let me know what y'all think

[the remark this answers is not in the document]
zaralouis-sf: False alarm, I actually brought those lines back. I realized that if there are any IDs set I want that to be the next available ID before going into the for loop, so that I'm not constantly comparing againt 0L. Let me know if that makes sense. I can comment in the code also if that would help.

universal-application-tool-0.0.1/test/forms/MultiOptionQuestionFormTest.java
```diff
@@ -92,4 +93,39 @@ public void getBuilder_emptyStringMinMax_noPredicateSet() throws Exception {
...
+  @Test
+  public void getBuilder_addNewOptions_setsNextIdCorrectly() throws Exception {
+    MultiOptionQuestionForm form = new DropdownQuestionForm();
+    form.setQuestionName("name");
+    form.setQuestionDescription("description");
+    form.setQuestionText("What is the question text?");
+    form.setQuestionHelpText("help text");
+    form.setMinChoicesRequired("");
+    form.setMaxChoicesAllowed("");
+    form.setOptions(ImmutableList.of("one", "two"));
```
shanemc-goog, line 123: Suggestion: add some comments as the next line isn't very "readable" without understanding exactly what setOptionsIds is doing, and also that you're skipping a few IDs for some reason.

EG:  // Add two existing options with IDs 4 and 1.

I'd recommend not adding that complexity here since this test is about adding ids, not setting the original next ID correctly, which is already tested in the previous test.

shanemc-goog (changes requested):

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
```diff
@@ -143,27 +146,42 @@ public QuestionDefinitionBuilder getBuilder() {
...
+
     // Note: the question edit form only sets or updates the default locale.
     for (int i = 0; i < options.size(); i++) {
-      questionOptions.add(
+      questionOptionsBuilder.add(
           QuestionOption.create(
               optionIds.get(i), i, LocalizedStrings.withDefaultValue(options.get(i))));
+      // The IDs are not guaranteed to be in any type of order, so doing this ensures that we find
+      // the largest ID in the list and accurately set the next largest.
+      if (optionIds.get(i) == nextAvailableId.orElse(0L)) {
```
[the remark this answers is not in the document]
zaralouis-sf: Wow using stream() is waaay cleaner, thank you! I need to start incorpating that in my code more.

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
```diff
@@ -143,27 +146,42 @@ public QuestionDefinitionBuilder getBuilder() {
...
+        setNextAvailableId(optionIds.get(i) + 1);
+      }
     }
+
     for (int i = 0; i < newOptions.size(); i++) {
-      questionOptions.add(
+      questionOptionsBuilder.add(
           QuestionOption.create(
               nextAvailableId.orElse(0L) + i,
```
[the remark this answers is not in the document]
zaralouis-sf: Done.

universal-application-tool-0.0.1/test/forms/MultiOptionQuestionFormTest.java
zaralouis-sf: I changed the test case so that the IDs are straightforward (1, 2, 3, 4), let me know if that wasn't what you had in mind.

universal-application-tool-0.0.1/test/controllers/admin/AdminQuestionControllerTest.java
```diff
@@ -245,6 +247,36 @@ public void update_redirectsOnSuccess() {
     assertThat(result.flash().get("message").get()).contains("updated");
   }
 
+  @Test
```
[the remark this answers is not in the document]
zaralouis-sf: Added!

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
```diff
@@ -143,27 +146,42 @@ public QuestionDefinitionBuilder getBuilder() {
       predicateBuilder.setMaxChoicesAllowed(getMaxChoicesAllowed());
     }
 
-    ImmutableList.Builder<QuestionOption> questionOptions = ImmutableList.builder();
+    ImmutableList.Builder<QuestionOption> questionOptionsBuilder = ImmutableList.builder();
     Preconditions.checkState(
         this.optionIds.size() == this.options.size(),
         "Option ids and options are not the same size.");
 
+    if (this.optionIds.size() > 0) {
```
shanemc-goog, line 154: This presetting to "something" isn't needed anymore right?

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
```diff
@@ -143,27 +146,42 @@ public QuestionDefinitionBuilder getBuilder() {
...
+    // The IDs are not guaranteed to be in any type of order, so doing this ensures that we find
+    // the largest ID in the list and accurately set the next largest.
+    Long maxId = optionIds.stream().max(Long::compareTo).orElse(0L);
+    setNextAvailableId(maxId + 1);
+
     for (int i = 0; i < newOptions.size(); i++) {
-      questionOptions.add(
+      questionOptionsBuilder.add(
           QuestionOption.create(
               nextAvailableId.orElse(0L) + i,
```
shanemc-goog, line 173: This orElse isn't necessary anymore, and if it's actually used it'd indicate a coding error since we just set it above, so we'd prefer the exception rather than the incorrect data.

shanemc-goog (changes requested):

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
zaralouis-sf: Done!

universal-application-tool-0.0.1/app/forms/MultiOptionQuestionForm.java
zaralouis-sf: Removed

universal-application-tool-0.0.1/test/controllers/admin/AdminQuestionControllerTest.java
```diff
@@ -245,6 +246,36 @@ public void update_redirectsOnSuccess() {
...
+                    2L, LocalizedStrings.of(Locale.US, "strawberry", Locale.FRENCH, "fraise")),
+                QuestionOption.create(
+                    3L, LocalizedStrings.of(Locale.US, "vanilla", Locale.FRENCH, "vanille")),
+                QuestionOption.create(
+                    4L, LocalizedStrings.of(Locale.US, "coffee", Locale.FRENCH, "café"))));
+    // We can only update draft questions, so save this in the DRAFT version.
+    testQuestionBank.maybeSave(definition, LifecycleStage.DRAFT);
+
+    DropdownQuestionForm questionForm = new DropdownQuestionForm(definition);
+    questionForm.setNewOptions(ImmutableList.of("cookie", "mint", "pistachio"));
```
shanemc-goog, line 271: It'd be good to check that these get the correct IDs too, and not just that another field in the data is updated.

zaralouis-sf: Added, let me know what you think!

zaralouis-sf: Okay, this made me realize we need to update the options IDs list when adding new options. Just wanted to call attention to that line in MultiOptionQuestionForm -- let me know if it doesn't look right to you

shanemc-goog (approved):

shanemc-goog:
Woohoo!

universal-application-tool-0.0.1/test/controllers/admin/AdminQuestionControllerTest.java
zaralouis-sf: Please ignore that previous comment, I was wrong. Tests should all be passing and code should be working as expected!

PR #2044, opened 2022-03-10
Update terraform files to support deployment slots

shubha-rajan:
### Description
Updated the terraform files to support using deployment slots. Also split main.tf out into multiple files to make it more readable. Some noteworthy changes that might break existing deployments:
- the BASE_URL environment variable is no longer set via terraform. This app setting needs to be bound to a specific slot, and the terraform provider doesn't support that, which means we'll have to set the value for both slots via the CLI during setup.
-The app container is no longer set by terraform. if we want to do a two-step deploy where we wait for the new version to be ready before swapping it into the primary slot, we have to set the containers for the primary and secondary slot manually before running terraform.

### Checklist


cloud/azure/modules/app/outputs.tf
```diff
@@ -2,8 +2,8 @@ output "app_service_default_hostname" {
   value = "https://${azurerm_app_service.civiform_app.default_site_hostname}"
 }
 
-output "pg_server_privatelink_fqdn" {
-  value = local.postgres_private_link
+output "app_service_canary_hostname" {
+  value = "https://${azurerm_app_service_slot.canary.default_site_hostname}"
 }
```
goldblatt, line 8: do we need to add the hostname for the non canary?

cloud/azure/modules/app/locals.tf
```diff
@@ -3,11 +3,61 @@ locals {
...
   postgres_private_link = azurerm_private_endpoint.endpoint.private_dns_zone_configs[0].record_sets[0].fqdn
-  generated_hostname    = "${var.application_name}-${random_pet.server.id}.azurewebsites.net"
 
   postgres_password_keyvault_id = "postgres-password"
   app_secret_key_keyvault_id    = "app-secret-key"
   adfs_secret_keyvault_id       = "adfs-secret"
   aws_secret_access_token       = "aws-secret-access-token"
   aws_access_key_id             = "aws-access-key-id"
+
+  app_settings = {
```
goldblatt, line 13: this stayed the same minus moving to locals correct?

goldblatt: also why locals over just living in the main?

cloud/deploys/staging_azure/staging.tfvars
```diff
@@ -1,8 +1,6 @@
 postgres_admin_login = "psqladmin"
 
 application_name       = "sgtest"
-docker_repository_name = "civiform"
```
goldblatt, line 4: i'd slightly prefer that we do this in your next PR cuz won't this break the deploy currently (or am i missing something)

goldblatt (changes requested):

cloud/azure/modules/app/locals.tf
shubha-rajan: I had to remove `BASE_URL` but other than that, it's the same.

shubha-rajan: this block was originally in main bc it was used by the app service resource, but now that it's being used by the main deployment and the canary slot I moved it to locals

cloud/deploys/staging_azure/staging.tfvars
shubha-rajan: yeah this will break the current deploy. i recommend merging #2045 first

cloud/azure/modules/app/outputs.tf
shubha-rajan: I think that's already in there!

shubha-rajan: I took out the postgres link because it's useless outside the vnet and I didn't want to clutter the output

bion (approved):

goldblatt (approved):
spoke over chat and understand the changes!

PR #2049, opened 2022-03-14
WIP: Save OIDC iss/sub as Accounts.authority_id

shanemc-goog:
### Description
Stores an Accounts.authority_id for OIDC accounts.

### Checklist

### Issue(s)

### DIFF
--- a/universal-application-tool-0.0.1/app/auth/oidc/OidcCiviFormProfileAdapter.java
+++ b/universal-application-tool-0.0.1/app/auth/oidc/OidcCiviFormProfileAdapter.java
@@ -32,10 +32,9 @@ import repository.UserRepository;
  * implementations of the two abstract methods.
  */
 public abstract class OidcCiviFormProfileAdapter extends OidcProfileCreator {
+  private static Logger LOG = LoggerFactory.getLogger(OidcCiviFormProfileAdapter.class);
   protected final ProfileFactory profileFactory;
   protected final Provider<UserRepository> applicantRepositoryProvider;
-
-  private static Logger LOG = LoggerFactory.getLogger(OidcCiviFormProfileAdapter.class);
 
   public OidcCiviFormProfileAdapter(
       OidcConfiguration configuration,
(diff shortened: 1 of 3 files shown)


universal-application-tool-0.0.1/app/auth/oidc/OidcCiviFormProfileAdapter.java
```diff
@@ -53,11 +52,37 @@ public OidcCiviFormProfileAdapter(
...
+    // Applicants.
+    String issuer = oidcProfile.getAttribute("iss", String.class);
+    // Subject identifies the specific user in the issuer.
+    // Pac4j treats the subject as special, and you can't simply ask for the "sub" claim.
+    String subject = oidcProfile.getId();
+    if (issuer == null || subject == null) {
+      return Optional.empty();
+    }
+    // This string format can never change. It is the unique ID for OIDC based account.
+    return Optional.of(String.format("iss: %s sub: %s", issuer, subject));
```
bion, line 70: thought (non-blocking): if we ever have need to search for a particular ID by sub or issuer in the DB, it'd be slightly easier if the hard-coded portions of the string weren't in here.

universal-application-tool-0.0.1/test/auth/ProfileMergeTest.java
```diff
@@ -77,13 +121,96 @@ public void testSuccessfulProfileMerge() {
...
+                idcsProfileAdapter.mergeCiviFormProfile(
+                    profileFactory.wrapProfileData(profileData), conflictingProfile))
+        .hasCauseInstanceOf(ProfileMergeConflictException.class);
+  }
+
+  @Test
+  public void testProfileMerge_oidc_fails_differentAuthorityIssuer() {
+    OidcProfile oidcProfile = createOidcProfile("foo@example.com", "issuer", "subject");
+    OidcProfile conflictingProfile =
+        createOidcProfile("foo@example.com", "a different issuer", "subject");
```
bion, line 213: praise: THANK YOU FOR THE THOROUGH AND CAREFUL TESTING

bion (approved):

universal-application-tool-0.0.1/app/auth/oidc/OidcCiviFormProfileAdapter.java
shanemc-goog: I think that scenario is rough eitherway.

Since we're constructing a "meta" value your scenario would only care about part of the stored value, so we'd need partial search ability

without the field names in the value it then is a tad harder to interpret.

Perhaps this should be a json string... but not sure I want to rely on deterministic stringification of a json object.

Open to more discussion here since there is a point of no return.

shanemc-goog: Likewise, SAML will have different constituent parts.. so a general search would seemingly have to be either a full substring search index or tokenized index and full token matches.

shubha-rajan (approved):
LGTM!

PR #2065, opened 2022-03-16
Pass in docker tag as arg to deploy script

shubha-rajan:


cloud/azure/bin/deploy
```diff
@@ -4,6 +4,14 @@ set -e
 source "cloud/shared/bin/lib/health.sh"
 source "cloud/azure/bin/lib.sh"
 
+while getopts t: flag
+do
+  case "${flag}" in
+    t) docker_tag=${OPTARG};;
```
bion, line 10: suggestion (blocking): flag name should be `--tag` to match https://github.com/seattle-uat/civiform-tf-deploy/blob/document-variables/bin/deploy#L3

bion (changes requested):

bion (approved):

PR #2078, opened 2022-03-17
Add storage account contributor role before initializing log file

shubha-rajan:


cloud/azure/bin/init-azure-log
```diff
@@ -4,6 +4,8 @@ set -e
 
 source "cloud/azure/bin/lib.sh"
 
+storage::assign_storage_account_contributor_role_to_user "${AZURE_RESOURCE_GROUP}"
```
goldblatt, line 7: maybe this should live inthe azure_log_file instead of here since we do an assugment there as well

bion (approved):
+1 to Susan's comment, otherwise lgtm

PR #2085, opened 2022-03-18
Overwrite SAML keystore if it already exists

shubha-rajan:


kmcnellis (approved):

cloud/azure/bin/lib/storage.sh
```diff
@@ -64,5 +64,6 @@ function storage::upload_blob {
   az storage blob upload \
     --account-name "${1}" \
     --container-name "${2}" \
-    --file "${3}"
+    --file "${3}" \
+    --overwrite "true"
```
rasmi, line 68: No need to act on this, but is it possible to instead pass `--overwrite` as an extra flag explicitly in the situation it's needed (in this case, in `setup-saml-keystore`), rather than set it as the default for the function globally?

shubha-rajan: in pretty much every case that we'd use this function (the other use case is in azure_log.sh though we're not using this function there yet), we'd want to overwrite, and previous versions of the azure cli used to default to overwriting if a file already exists. The last released version of the cli made a breaking change to allow overwrites only if we use the flag. I think a better approach might be to pass in an optional no-overwrite flag to this function. I can create an issue for that.

shubha-rajan: filed https://github.com/seattle-uat/civiform/issues/2087

PR #2091, opened 2022-03-19
Allow configuring civic entity names and URLs with terraform

shubha-rajan:


bion (changes requested):
I think last we discussed this we decided to switch to serving the logo directly from CiviForm and using an env var to specify which logo to serve. Doing it that way avoids potential CORS issues and also reduces that likelihood that a change in a completely different website or service will break the CiviForm deployment.

shubha-rajan:
> I think last we discussed this we decided to switch to serving the logo directly from CiviForm and using an env var to specify which logo to serve. Doing it that way avoids potential CORS issues and also reduces that likelihood that a change in a completely different website or service will break the CiviForm deployment.

We can change the terraform code once the changes for swapping out the image file are implemented. This is mostly a temporary solution to get Rochester's current deployment working when we wrap up and test later today.

goldblatt (approved):

bion (approved):

PR #2100, opened 2022-03-22
Implement rollback command

bion:


cloud/azure/bin/deploy
```diff
@@ -1,56 +1,37 @@
...
+# loads and validates config variables and generates tfvars file
 python3 "cloud/shared/bin/lib/tf_apply_setup.py"
-readonly TF_VAR_FILE="setup.auto.tfvars"
-readonly BACKEND_VARS_FILENAME="backend_vars"
-
-cloud/azure/bin/setup_tf_shared_state "${TERRAFORM_TEMPLATE_DIR}/${BACKEND_VARS_FILENAME}"
-
-readonly USER_ID="$(azure::get_current_user_id)"
 
+azure::set_common_vars
```
goldblatt, line 10: ugggh i realllly wish we could use the config loader class in python for this

cloud/azure/bin/deploy
```diff
@@ -1,56 +1,37 @@
...
-        -chdir="${TERRAFORM_TEMPLATE_DIR}" \
-        apply \
-        -var-file="${TF_VAR_FILE}"
-)
-readonly TF_STATUS=$?
-set -e
-
-if (( $TF_STATUS != 0 )); then
-    log::deploy_failed "${IMAGE_TAG}" "${USER_ID}"
+if ! terraform::perform_apply; then
```
goldblatt, line 13: slightly prefer to have this live alone and then store as a var and then check the result of it

cloud/shared/bin/lib.sh
```diff
@@ -1,9 +1,18 @@
 #!/usr/bin/env bash
 
+set -e
+set -o pipefail
+
+export TF_VAR_FILE="setup.auto.tfvars"
+export BACKEND_VARS_FILENAME="backend_vars"
```
goldblatt, line 7: can we update the config loader to get these form the env as they are stored in the setup class currently

cloud/shared/bin/lib/terraform.sh
```diff
@@ -0,0 +1,24 @@
...
+
+#######################################
+# Generates terraform variable files and runs terraform init and apply.
+# Also initializes the storage bucket for tfstate if it's not setup yet.
+# Globals:
+#   TERRAFORM_TEMPLATE_DIR
+#   BACKEND_VARS_FILENAME
+#   TF_VAR_FILE
+#######################################
+function terraform::perform_apply() {
```
goldblatt, line 11: i think we can use this in the setup script as well

goldblatt (changes requested):
mmm maybe we should chat a lil about if we want to consolidate to use python here too or if that is more worth than it's worth?

cloud/azure/bin/deploy
bion: like this?
```suggestion
set +e
terraform::perform_apply
readonly TF_STATUS=$?
set -e

if (( $TF_STATUS != 0 )); then
...
```

cloud/azure/bin/deploy
bion: yeah, that would be preferable huh. do you see a good way to do that?

cloud/shared/bin/lib/terraform.sh
bion: `cloud/azure/bin/setup_tf_shared_state` relies on a bunch of bash functions being loaded first, any thoughts on how that would work with calling this from py?

cloud/shared/bin/lib.sh
bion: done, PTAL i *think* i did it correctly

cloud/azure/bin/deploy
goldblatt: ugh no the way you had it was much better

cloud/deploys/staging_azure/setup
```diff
@@ -53,4 +53,4 @@ terraform \
 terraform \
   -chdir="${TEMPLATE_DIR}" \
   apply \
-  -var-file="${TF_VAR_FILE}"
+  -var-file="${TF_VAR_FILENAME}"
```
goldblatt, line 56: i think this is wrong and also like we probably don't need this file. since the chdir changes to the directory the file has to be colocated

cloud/shared/bin/lib/config_loader.py
```diff
@@ -12,16 +12,12 @@
...
 """
 class ConfigLoader:
-    TF_VARS_FILENAME = "setup.auto.tfvars"
-    BACKEND_CONFIG_FILENAME = "backend_vars"
-    
-    @property
+
     def tfvars_filename(self):
-        return self.TF_VARS_FILENAME
+        return os.environ.get('TF_VAR_FILENAME', None)
```
goldblatt, line 17: we probably would want to fail here it can't be null?

cloud/shared/bin/lib/terraform.sh
goldblatt: wait sorry does it not source the things it needs? also we ALSo probably need a way to handle if the user doesn't want to use backend vars. i guess like can't we do the like source xyz && call the script right?

goldblatt (changes requested):

cloud/shared/bin/lib/config_loader.py
bion: done

cloud/shared/bin/lib/terraform.sh
bion: > wait sorry does it not source the things it needs?

it does now

bion: > also we ALSo probably need a way to handle if the user doesn't want to use backend vars

what var do we switch on for that?

cloud/deploys/staging_azure/setup
bion: we shouldn't pass `-var-file=`?

shubha-rajan: no I think we don't need this setup bash file, since all the setup is being done with the python file. Also TF_VAR_FILENAME should just be a filename without the directory path before it, since chdir changes into the directory where the file is located

bion: removed

cloud/shared/bin/lib/terraform.sh
bion: done

goldblatt (approved):

shubha-rajan (approved):

PR #2191, opened 2022-03-28
Need to Be properties to run config_loader

goldblatt:
### Description
the setup expects there to be properties not functions!

### Checklist

### Issue(s)


bion (approved):
why is this?

goldblatt:
these get used without function calls so it caused the code to fail. I could switch this to a getter/setter https://google.github.io/styleguide/pyguide.html#getters-and-setters; doesn't seem like the style guide cares especially but i figured since they are ostensibly variables we are trying to access the property was cooler.

PR #2194, opened 2022-03-29
Add upgrade flag to terraform init

goldblatt:
if the version changes have to include upgrade to get the latest version or else command will fail.

### Description

### Checklist

### Issue(s)


bion (approved):

goldblatt:
I think this has unintended side effects (it always looks for the latest available version of the module to use). I couldn't run on my local without upgrade when we updated the aws version. if we pin the version i think we might avoid the unitended side effect though

goldblatt:
https://discuss.hashicorp.com/t/what-does-terraform-init-upgrade-do/32638/4

PR #2204, opened 2022-03-29
Remove zip file from prod container to reduce size

bion:
closes https://github.com/seattle-uat/civiform/issues/2202


prod.Dockerfile
```diff
@@ -17,23 +17,33 @@ RUN set -o pipefail && \
...
 FROM adoptopenjdk/openjdk11:jdk-11.0.10_9-alpine-slim AS stage2
-COPY --from=stage1 /usr/src/universal-application-tool-0.0.1/target/universal/universal-application-tool-0.0.1.zip /civiform.zip
+COPY --from=stage1 /universal-application-tool-0.0.1 /universal-application-tool-0.0.1
+
+# Upgrade packages for stage2 to include latest versions.
+RUN set -o pipefail && \
+    apk update && \
+    apk add --upgrade apk-tools && \
+    apk upgrade --available && \
+    apk add --no-cache --update bash openssh
```
gwendolyngoetz, line 42: Is this updating the image OS or our application?

bion: It's updating the image

gwendolyngoetz (approved):
It looks about right to me

bion:
@gwendolyngoetz unfortunately the image is still 1.19GB after this change so this will need some more investigation. I'm not going to have time for it the rest of this week and I'm out next week so feel free to pick it up if you'd like, otherwise I'll get back to it mid-April.

clouser2:
@bion When I patch this in and run `docker image ls`, I see the image size go down from 1.18GB to 762MB. I *believe* this works.

clouser2 (approved):

PR #2212, opened 2022-03-29
add checks for formatting and validating terraform

### LINKED ISSUE #1807: Add CI check to format and validate terraform files

shubha-rajan:
Fixes #1807


.github/workflows/pr.yaml
```diff
@@ -97,6 +97,21 @@ jobs:
...
+    steps:
+      - name: check out pr branch
+        uses: actions/checkout@v3
+      - name: generate bastion key
+        run:  ssh-keygen -q -t rsa -b 4096 -N '' -f "${HOME}/.ssh/bastion"
+      - name: setup terraform
+        uses: hashicorp/setup-terraform@v1
+      # as we add more templates, we will add checks to validate them. for now, we only have one
+      - name: terraform init azure_saml_ses 
+        run: terraform -chdir=cloud/azure/templates/azure_saml_ses init -backend=false
```
goldblatt, line 111: ooo does this work like this with backend=false

goldblatt: lol i should do this instead of the overrides

shubha-rajan:
> I think ideally we'd have something that confirms like we generate the setup.auto.tfvars correctly as well but i this is dope!

Yeah I considered adding a plan step there too, but then we have to deal with credentials and authentication for azure and I didn't want to do that. 

you can see that the validation failed before I added the bastion key, and the formatting failed before I formatted the GCP files. So what's already in the PR is working.

.github/workflows/pr.yaml
shubha-rajan: https://www.terraform.io/cli/commands/validate
"Validation requires an initialized working directory with any referenced plugins and modules installed. To initialize a working directory for validation without accessing any configured remote backend, use:
```
terraform init -backend=false
```
To verify configuration in the context of a particular run (a particular target workspace, input variable values, etc), use the terraform plan command instead, which includes an implied validation check."

PR #2223, opened 2022-03-30
Remove unused webjar dependencies

shubha-rajan:
We're not using React, so let's remove the React dependencies.


zaralouis-sf (approved):
Thanks for doing this!

bion (approved):

PR #2248, opened 2022-04-02
ProgramRepository Formatting/comment updates.

shanemc-goog:
non functional cleanup

### DIFF
--- a/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
@@ -27,6 +27,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import services.PaginationResult;
 import services.PaginationSpec;
+import services.program.ProgramDefinition;
 import services.program.ProgramNotFoundException;
 
 /**
@@ -110,12 +111,9 @@ public class ProgramRepository {
       // Ensure we didn't add a duplicate with other code running at the same time.
       Preconditions.checkState(
           draftVersion.getPrograms().stream()
-                  .filter(
-                      program ->
-                          program
-                              .getProgramDefinition()
-                              .adminName()
-                              .equals(existingProgram.getProgramDefinition().adminName()))
+                  .map(Program::getProgramDefinition)
+                  .map(ProgramDefinition::adminName)
+                  .filter(name -> name.equals(existingProgram.getProgramDefinition().adminName()))
                   .count()
               == 1,
           "Must be exactly one program with this name in the draft.");
--- a/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
+++ b/universal-application-tool-0.0.1/app/repository/ProgramRepository.java
@@ -109,11 +109,12 @@ public class ProgramRepository {
           draftVersion.getLifecycleStage().equals(LifecycleStage.DRAFT),
           "Draft version must remain a draft throughout this transaction.");
       // Ensure we didn't add a duplicate with other code running at the same time.
+      String programName = existingProgram.getProgramDefinition().adminName();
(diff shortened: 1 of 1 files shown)


universal-application-tool-0.0.1/app/repository/ProgramRepository.java
```diff
@@ -68,61 +72,66 @@ public Program updateProgramSync(Program program) {
...
-        newDraft.addVersion(draftVersion).save();
-        draftVersion.refresh();
-        Preconditions.checkState(
-            draftVersion.getPrograms().contains(newDraft),
-            "Must have successfully added draft version.");
-        Preconditions.checkState(
-            draftVersion.getLifecycleStage().equals(LifecycleStage.DRAFT),
-            "Draft version must remain a draft throughout this transaction.");
-        Preconditions.checkState(
-            draftVersion.getPrograms().stream()
```
shanemc-goog, line 99: Made this check cleaner

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
```diff
@@ -68,61 +72,67 @@ public Program updateProgramSync(Program program) {
...
+      versionRepository.get().updateQuestionVersions(newDraft);
+      transaction.commit();
+      return newDraft;
+    } catch (IllegalStateException e) {
+      transaction.rollback();
+      // We must end the transaction here since we are going to recurse and try again.
+      // We cannot have this transaction on the thread-local transaction stack when that
+      // happens.
+      transaction.end();
+      return createOrUpdateDraft(existingProgram);
```
goldblatt, line 130: would this forever loop in some situations?

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
```diff
@@ -68,61 +72,67 @@ public Program updateProgramSync(Program program) {
...
+      Preconditions.checkState(
+          draftVersion.getLifecycleStage().equals(LifecycleStage.DRAFT),
+          "Draft version must remain a draft throughout this transaction.");
+      // Ensure we didn't add a duplicate with other code running at the same time.
+      String programName = existingProgram.getProgramDefinition().adminName();
+      Preconditions.checkState(
+          draftVersion.getPrograms().stream()
+                  .map(Program::getProgramDefinition)
+                  .map(ProgramDefinition::adminName)
+                  .filter(programName::equals)
```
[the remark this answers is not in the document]
shanemc-goog: No, stream methods tend to take a method signature that accepts 1 param (using generic typing) and returns a contextually relevant value. 

Here It's a Predicate (which returns a boolean) and the method signature of .equals on the programName variable fits that through duck typing, so in effect programName.equals(X) gets called on the current streamed value.

It's equivalent to .filter(draftName -> programName.equals(draftName))

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#filter-java.util.function.Predicate-


For the map() above I believe it's calling the class method provided on the relevant object through inheritance and then the return type becomes the input type for the next stream operation.

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
shanemc-goog: seemingly so.... yeah that looked like it could be a problem.  I'll review this when I dig into how transactions work in our code.

goldblatt (approved):
approving with the like goal of looking into the forever loop scenario later

universal-application-tool-0.0.1/app/repository/ProgramRepository.java
shanemc-goog: Just to be clear, this is a pre-existing issue, nothing I did changes it either way.

PR #2253, opened 2022-04-04
Open external links in a new tab

### LINKED ISSUE #2128: Web links open in the same browser tab
**Describe the bug**
When an applicant clicks on a link within a form, it opens in the same browser tab. It's an issue because it opens over the draft form the applicant was working on. To navigate back to the form, the applicant would need to select "back" on their web browser. The preferred experience would be to force any link within a form to open in a new tab. 

**To Reproduce**
1. Go to https://civiform.seattle.gov/programs/testing_genericprogram
2. Complete the form until you see a link (like in the attachment)
3. Click the link
4. See that it opens in the same browsers tab as the draft form

**Expected behavior**
Links should open in a new tab

**Screenshots**
[Links open in same browser tab.pdf](https://github.com/seattle-uat/civiform/files/8344933/Links.open.in.same.browser.tab.pdf)
Attached

**Additional context**

gwendolyngoetz:
### Description
Set the target to blank where urls are being created.

This would will open all links made this way in a new tab. So I've swapped my problem. I want to keep links to pages within the application still opening in the same tab.

There are two ways I'm thinking of doing this

1) See if the url starts with a forward slash. That looks like that would be a good indicator of a relative link. I could also see if the url starts with http|https as a cue for an external link, it's probably fine just looking at those. It's probably very unlikely that ftp or another scheme would be used. I'd prefer this route, but I'm not sure if I can guarantee all internal app links would have relative paths.

2) Use the hostname to determine if a link is internal or external. If the link starts with the hostname it's internal otherwise it's external. I don't know if getting the hostname is accessible to the method already.

### Checklist

### Issue(s)
Fixes #2128

### DIFF
--- a/universal-application-tool-0.0.1/app/views/components/TextFormatter.java
+++ b/universal-application-tool-0.0.1/app/views/components/TextFormatter.java
@@ -78,7 +78,9 @@ public class TextFormatter {
       contentBuilder.add(
           a().withText(url.getOriginalUrl())
               .withHref(url.getFullUrl())
-              .withClasses(Styles.OPACITY_75));
+              .withClasses(Styles.OPACITY_75)
+              .withTarget("_blank"));
+
       content = content.substring(index + url.getOriginalUrl().length());
     }
     // If there's content leftover, add it.


goldblatt:
i think here would prefer to add a argument to the function that allow things to specify what they would want; not sure opening in a new tab is the long term outcome we want for all links we create

PR #2254, opened 2022-04-04
Adding in robots.txt. Hard-coded in controller action for now.

### LINKED ISSUE #2252: Staging site is being indexed by search engines
Need to add robots.txt

gwendolyngoetz:
### Description
I want control indexing by search engines, but maybe I'm overthinking the problem. 

Currently I have it set to have the robots.txt deny indexing of all pages, except if the site is prod then to allow indexing / and the loginForm page. The values are hard-coded in the controller action since I'm not sure where best to put them.

I figure this will block indexing for any staging sites, but still allow some indexing for prod for SEO purposes. I don't think we want to have bots indexing past the login page. Especially in the event they manage to successfully continue as a guest.

If we want to just block all indexing no matter the environment I'll just switch this to a static asset file and drop the controller action.

### Checklist

### Issue(s)
Fixes #2252

### DIFF
--- a/universal-application-tool-0.0.1/app/controllers/HomeController.java
+++ b/universal-application-tool-0.0.1/app/controllers/HomeController.java
@@ -10,6 +10,7 @@ import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.play.java.Secure;
+import play.Environment;
 import play.i18n.MessagesApi;
 import play.libs.concurrent.HttpExecutionContext;
 import play.mvc.Controller;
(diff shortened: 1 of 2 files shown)


goldblatt (commented):
do we need to use https://developers.google.com/search/docs/advanced/crawling/block-indexing to block indexing?

universal-application-tool-0.0.1/app/controllers/HomeController.java
```diff
@@ -82,6 +86,18 @@ public Result loginForm(Http.Request request, Optional<String> message)
     return ok(loginForm.render(request, messagesApi.preferred(request), message));
   }
 
+  public Result robotsTxt() {
+    if (environment.isProd()) {
+      return ok("User-agent: *\n" +
+              "Allow: /$\n" +
+              "Allow: /loginForm$\n" +
```
goldblatt, line 93: do we need to end with $ here?

gwendolyngoetz: > do we need to use https://developers.google.com/search/docs/advanced/crawling/block-indexing to block indexing?

After doing to looking it sounds like using meta robots in tandem with robots.txt is often recommended these days. I'll look into adding that as well.

> do we need to end with $ here?

I was looking to specifically target the site index and the loginForm page. The $ will make sure to match to the end of the string. My understanding is that without the $ it could end up matching loginForm, loginForm2, loginForm/nested/path.

goldblatt: cool feel free to add this and do the meta stuff in another PR

PR #2268, opened 2022-04-07
Add local dev testing OIDC provider.

shanemc-goog:
### Description
Runs a dev OIDC service and points civiform at it.

The Civiform service and the developers' browser need to be able to load the same hostname:port.  Civiform does it through Docker Networking, the user must do it by adding the container name as a localhost alias in their /etc/hosts file.

### Checklist

### Issue(s)
Fixes #https://github.com/seattle-uat/civiform/issues/1938


shanemc-goog:
This setup bit rotted I think with the change to abstract the auth providers, Perhaps they need to be singletons (or not if they are....)

CompilationException: Compilation error[method getExistingApplicant(org.pac4j.oidc.profile.OidcProfile) is already defined in class auth.oidc.OidcCiviFormProfileAdapter]

bion (approved):

bion:
reminder to update dev docs

PR #2269, opened 2022-04-07
Fix behavior of discardDraft(Question) to remove silent creation of invalid data.

### LINKED ISSUE #2249: discarding a question draft doesn't update draft questions that refer to it.
**Describe the bug**

Enumerators are referenced by other question.  When they are updated drafts are made for them and the questions the refer to them.  When an enumerator is discarded those questions are not updated.  The code updates programs, but ignores questions.

https://github.com/seattle-uat/civiform/blob/a6d85511f1143c0c7de385032e792e56a08b8b11/universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java#L129

That method needs to update all questions similar to the programs it does update.

This is a possible cause for a silent application flow error in #2040 (note though that in that bug the question doesn't exist at all which is unexpected) and part of the effort in #2047

**To Reproduce**
1. Publish an Enumerator Question (EnumQ)  and a question (UsageQ) that uses it.
Notice: the DB looks as expected UsageQ active rev has EnumQ's active ID in it's question.enumerator_id field.

2. Create a draft of EnumQ

Notice: both questions have a new Draft revision, with the draft UsageQ enumerator_id set to the new daft revision ID of EnumQ.

3. Discard draft on EnumQ.

Notice: The draft UsageQ enumerator_id is still set to the now abandoned previous-draft revision ID of EnumQ.

4. Publish add drafts

The new Active UsageQ revision now references an enumerator Question that is not Active; the system requires the reference be active.

shanemc-goog:
### Description
1. Fixes discardDraft to also update Questions that reference the discarded draft, and prevents discarding the very first draft.

The code already handled programs but didn't account for enumeration references which leaves those in an invalid state.

2. The behavior of non-existent -> DRAFT -> discard was allowable on all drafts and largely undefined.  This could create invalid state if the draft was referenced before discard.  EG programs or enumerated questions if one referenced the discard question.

The code now disallows discarding the very first draft (through exception) rather than allowing invalid state to be silently created. We can allow it in the future but it'd be more complicate code, as we would have to detect tif he situation is or is not permissible.  

### Checklist

### Issue(s)
Fixes #2249

### DIFF
--- a/universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
+++ b/universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
@@ -2,6 +2,7 @@ package services.question;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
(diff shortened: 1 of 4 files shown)


kmcnellis:
Should we have a warning that deleting the question is going to change these other ones?  I can see a situation in which someone wanted to just move it to another page, and ends up not realizing they have to re-do a bunch of work elsewhere

shanemc-goog:
> Should we have a warning that deleting the question is going to change these other ones?

All of this is a part of the versioning design, and largely intended to be invisible to the user (even though it does reveal itself a little bit). While it "changes" the other questions (and in the existing code the programs) it's all book keeping to adhere to the modeling design.


An example of this interaction is the data modeling cascade effects outlines here.
https://github.com/seattle-uat/civiform/wiki/Data-versioning-model/#update-a-question-referenced-by-a-program


let me know if you have more concerns/questions.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -137,6 +137,15 @@ public void discardDraft(Long id) throws InvalidUpdateException {
       throw new InvalidUpdateException("Did not find question in draft version.");
     }
     question.get().save();
+
+    // Update any repeated questions that may have referenced the discarded question.
+    Version activeVersion = versionRepositoryProvider.get().getActiveVersion();
+    Optional<Question> activeQuestion =
+        activeVersion.getQuestionByName(question.get().getQuestionDefinition().getName());
+    questionRepository.updateAllRepeatedQuestions(
```
goldblatt, line 145: should we check if they are equal first before updating them?

shanemc-goog: I waffled but went ahead an added it out of caution as it implies data corruption; so that at least we're clearly aware of it.

The same revision ID for a question can't be ACTIVE and DRAFT, and it was checked as being a valid draft ID so this state would indicate data corruption if it happened.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
```diff
@@ -126,18 +127,40 @@ public void archiveQuestion(Long id) throws InvalidUpdateException {
...
+            .join()
+            .orElseThrow(() -> new InvalidUpdateException("Did not find question."));
+
+    // Find the Active version.
+    Version activeVersion = versionRepositoryProvider.get().getActiveVersion();
+    Long activeId =
+        activeVersion
+            .getQuestionByName(question.getQuestionDefinition().getName())
+            // TODO: If nothing depends on this question then it could be removed.
+            .orElseThrow(() -> new InvalidUpdateException("Deleting the first version of a question is not supported."))
```
shanemc-goog, line 144: FYI: I added this too. In the current code if you delete the first revision of a question and a draft program uses it, then the program becomes corrupt as there is not active revision of the program to use.

We could detect and not allow for it in the UI, but not a concern for this PR, and the Program won't be corrupted at least.

shanemc-goog (commented):
I need to update the .ts tests I realize now.

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
goldblatt: ooo okay gotcha, i mean if we ant to handle it in the deletion section and not here that's also fine by me was just curious

goldblatt (approved):

shanemc-goog:
@kmcnellis FYI on the response above, not sure why I couldn't reply directly to your comment

universal-application-tool-0.0.1/app/services/question/QuestionServiceImpl.java
shanemc-goog: Not sure what you mean, but it's best to assert everything is as expected before making partial changes.

PR #2271, opened 2022-04-08
Setting the TF to have a storage policy for files

goldblatt:
the log retention is a misnomer setting it to zero and making it less noisy

### DIFF
--- a/cloud/azure/modules/app/logs.tf
+++ b/cloud/azure/modules/app/logs.tf
@@ -15,6 +15,7 @@ resource "azurerm_monitor_diagnostic_setting" "app_service_log_analytics" {
     category = "AppServiceAppLogs"
 
     retention_policy {
+      days    = 0
       enabled = false
     }
   }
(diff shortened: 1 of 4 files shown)


cloud/azure/modules/app/logs.tf
```diff
@@ -15,6 +15,7 @@ resource "azurerm_monitor_diagnostic_setting" "app_service_log_analytics" {
     category = "AppServiceAppLogs"
 
     retention_policy {
+      days    = 0
```
bion, line 18: question (blocking): why default these all to zero?

bion (changes requested):

cloud/azure/modules/app/logs.tf
goldblatt: ahh this is a bug in terraform azure these aren't enabled so it shouldn't matter but we have to default them to false with days=0 https://github.com/hashicorp/terraform-provider-azurerm/issues/7235 otherwise it shows that there are terraform changes when you try to apply when there were no terraform changes

bion (approved):

PR #2276, opened 2022-04-08
Validate all name inputs

### LINKED ISSUE #1693: Name validation only uses the status of the last question.
https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/assets/javascripts/validation.ts#L383

The returned status is overwritten each loop.

tjbanghart:
### Description
Causes `validateNameQuestion() to fail if any of the name inputs are invalid. Previously, only the last name input determined whether the question was valid. See `with first invalid does not submit` test as an example.

### Checklist

### Issue(s)
Fixes #1693

### DIFF
--- a/browser-test/src/name.test.ts
+++ b/browser-test/src/name.test.ts
@@ -1,0 +1,270 @@
+import { Page } from 'playwright'
+import {
+  AdminPrograms,
+  AdminQuestions,
+  ApplicantQuestions,
+  loginAsAdmin,
+  loginAsGuest,
+  logout,
+  resetSession,
+  selectApplicantLanguage,
+  startSession,
+} from './support'
+
+const NAME_FIRST = ".cf-name-first";
+const NAME_LAST = ".cf-name-last";
+
+describe('name applicant flow', () => {
+  let pageObject: Page
+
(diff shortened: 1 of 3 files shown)


browser-test/src/name.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      applicantQuestions = new ApplicantQuestions(pageObject)
+
+      await adminQuestions.addNameQuestion({
+        questionName: 'name-test-q',
+      })
+      await adminPrograms.addAndPublishProgramWithQuestions(
+          ['name-test-q'],
+          programName
+      )
+      debugger;
```
shanemc-goog, line 46: what's this I'm not familiar with it?

Likely should remove it though?

tjbanghart: Yikes, definitely shouldn't have been left in there! 

Maybe I should cut a ticket for lint checks on these test files?

shanemc-goog: Eh, I don't think this is common, I still don't know what it does :D and it didn't seem to affect the test runs.

shanemc-goog (approved):

PR #2279, opened 2022-04-12
Create a terraform deploy script to remove all remote resources

kmcnellis:
### Description
- Add a dev deploy script to bring down your azure resources (needs to have a fake bastion ssl key to work)

### Checklist

### DIFF
--- a/cloud/deploys/dev_azure/bin/destroy
+++ b/cloud/deploys/dev_azure/bin/destroy
@@ -1,0 +1,10 @@
+#! /usr/bin/env bash
+
+source cloud/deploys/dev_azure/bin/lib.sh
+source cloud/azure/bin/lib/bastion.sh
+
+bastion::ssh_keygen "$HOME/.ssh/bastion"
+
+cloud/shared/bin/destroy
+
+bastion::remove_ssh_key "$HOME/.ssh/bastion"
--- a/cloud/shared/bin/destroy
+++ b/cloud/shared/bin/destroy
@@ -7,7 +7,6 @@ source cloud/shared/bin/lib.sh
 # This setups the environment of civiform
 # it will execute a bunch of scripts that help setup terraform, and the
 # various cloud providers.
-
-export TF_VAR_image_tag="${IMAGE_TAG}"
+export TF_VAR_image_tag="${IMAGE_TAG:=latest}"
 
 terraform -chdir="$TERRAFORM_TEMPLATE_DIR" destroy
--- a/cloud/shared/bin/destroy
+++ b/cloud/shared/bin/destroy
@@ -1,0 +1,13 @@
+#! /usr/bin/env bash
+
+set -e
+
+source cloud/shared/bin/lib.sh
+
+# This setups the environment of civiform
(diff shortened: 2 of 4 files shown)


cloud/deploys/dev_azure/bin/destroy
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env bash
+
+source cloud/deploys/dev_azure/bin/lib.sh
+source cloud/azure/bin/lib/bastion.sh
+
+bastion::ssh_keygen "$HOME/.ssh/bastion"
+
+cloud/shared/bin/destroy
```
goldblatt, line 8: yeah the image tag might get a lil lost here, i think you either need to source the config again or to have the user pass in the image_tag; it's unclear if it really matters like if you pass it in

goldblatt (changes requested):

cloud/shared/bin/destroy
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+
+source cloud/shared/bin/lib.sh
+
+# This setups the environment of civiform
+# it will execute a bunch of scripts that help setup terraform, and the
+# various cloud providers.
+
+export TF_VAR_image_tag="${IMAGE_TAG}"
```
goldblatt, line 11: do we need this?

goldblatt (changes requested):

cloud/deploys/dev_azure/bin/destroy
kmcnellis: Hm, it seems to be running for me without needing to add a tag flag.  Is there something that you are seeing?  I don't think we really need a tag, right?

cloud/shared/bin/destroy
kmcnellis: When I don't include this, the terraform template prompts on the commandline for a tag

cloud/deploys/dev_azure/bin/destroy
kmcnellis: (it runs for me without that flag because of the export TF_VAR_image_tag="${IMAGE_TAG}" in the other file)

cloud/shared/bin/destroy
goldblatt: yeah that's cuz it's relying on the exported one existing in the config.sh but we actually moved away from including it there. i think in this case it just doesn't matter what the tag is so we should export like 'latest' or some string so we don't fail if you don't have the IMAGE_TAG in the env variables

goldblatt (changes requested):
This is almost right there's just some confusion around the image_tag env variables for me!

cloud/shared/bin/destroy
kmcnellis: done

cloud/deploys/dev_azure/bin/destroy
kmcnellis: This one doesn't really need the image tag

goldblatt (approved):

kmcnellis:
https://github.com/seattle-uat/civiform/pull/2384 makes this not needed

PR #2281, opened 2022-04-12
Bump J2html to 1.5.0

### LINKED ISSUE #2228: Update j2html
CiviForm uses the [j2html](https://j2html.com) library for rendering HTML. There is a new major version available with improved type safety and APIs.

This task is to update the library and our use of it: https://j2html.com/download.html#upgrade

This is a good first issue for someone interested in doing deeper work in the CiviForm UI.

syntapy:
### Description
Update j2html from 1.4.0 to 1.5.0

### Checklist
- [x] Pass tests
- [x] Manual UI testing
- [x] Lighthouse reports for accessibility testing

### Issue(s)

Fixes #2228


syntapy:
Currently on `admin/programs/ProgramBlockEditView.java` and `components/Modal.java`

syntapy:
Some question on how to handle return types in `admin/programs/ProgramIndexView.java`. As I've handled it there are some otherwise uselessly nested `div` elements holding `form` elements. But thats also the case where theres empty div inside a div. So that could be probably refactored a bit, but is a different issue than this one

syntapy:
Might have a problem compiling `components/SelectWithLabel.java` since I may not have passed the generic type correctly

syntapy:
So I had been converting general html tag type into specific tags (`Tag` and `ContainerTag` to types such as `DivTag` or `FormTag` or `ATag` etc...) like the [doc](https://j2html.com/download.html] says

However, I encountered [one method](https://github.com/syntapy/universal-application-tool/blob/j2html_1p5/server/app/views/HtmlBundle.java#L83-L86) which is called from 27 different files, some of which are multiple times. So that meant going through each and making sure they are all returning the same html type (probably will end up being a `DivTag`)

This made me wonder if I was on the right track, and that maybe I should have been using general `Tag` types instead. 

When I first set up a minimal example to see what I could convert, I thought I was able to set things as `Tag` types

However, after revisiting, the code snippet below gives `cannot find symbol: symbol:   method with(j2html.tags.Tag)`
error

I am able cast though, like the snippet below. But that seems kind of ugly
casting_works

So I guess I will stick to more specific types

syntapy:
After talking to bion I have better direction. Mainly I need to remember to *try* using general tag types for method input parameters, and now its `ContainerTag` that can use `.with(..)` instead of `Tag`

Here is [one place](https://github.com/syntapy/universal-application-tool/blob/j2html_1p5/server/app/views/admin/ti/TrustedIntermediaryGroupListView.java#L54-L58) where I need to figure out. `BaseHtmlView.java` [fields](https://github.com/syntapy/universal-application-tool/blob/j2html_1p5/server/app/views/HtmlBundle.java#L34-L46) might need to be general types

syntapy:
So apparently the [footer scripts field](https://github.com/syntapy/universal-application-tool/blob/j2html_1p5/server/app/views/HtmlBundle.java#L36) needs to be `ContainerTag` since the google analytics portion has a `.wtih(..)` to include raw html

syntapy:
Refactoring `FieldWithLabel.java`. The way I'm doing it might make it difficult to create lists of multiple types of field tags, as done in `app/views/admin/TranslationFormView.java::renderTranslationForm`

That requires it being all `InputTag` types. But what if we want a list that also has a `TextareaTag` element?

This is why we should probably pass in the `ImmutableList<FieldWithLabel>` instead of e.g. a `ImmutableList<InputTag>`, so that the final element can be created when being added to the DOM tags

So for that I need to just make a method in `FieldWithLabel` that returns the final tag, regardless if its a `InputTag` or `TextareaTag`. So maybe wrapping it in a `DivTag`? Oh right, I'm already returning `DivTag`s for every one of them.

I can have both a `maybeInputTag` and a `maybeTextareaTag` fields. However, the issue there is I need to call some of the tag attribute setters like `nonCheckboxRadioFinalBuild` on whichever field is being used. 

Now if every one of those methods called in `nonCheckboxRadioFinalBuild` is in `Tag`, then I could pass in the field being used to that

syntapy:
Just had to change `" new version"` to `"new version"` over [here](https://github.com/syntapy/universal-application-tool/blob/4ed410e13653026f28b132bf4b088de320af8305/browser-test/src/support/admin_questions.ts#L265) to make a [browser test](https://github.com/syntapy/universal-application-tool/blob/4ed410e13653026f28b132bf4b088de320af8305/browser-test/src/end_to_end_enumerators.test.ts) pass

syntapy:
One browser test failure is on a point where it expects a green bg but the context obviously needs a red bg. So maybe need to add different tests for each conext. Having trouble figuring out how to track that down with playwright, so I'll need to [read up](https://playwright.dev/docs/trace-viewer) more and all

syntapy:
@bion can you run my CI tests?

syntapy:
I need to get long stack traces working in this, otherwise seems like debugging the browser tests will be a ***nightmare***

options include
- [adding a `--async-stack-trace`](https://thecodebarbarian.com/async-stack-traces-in-node-js-12) if possible
- [async hooks](https://nodejs.medium.com/what-use-cases-exist-for-async-hooks-6eab74d7eefd)

syntapy:
@bion can you run CI tests again please?

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,100 +262,158 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
     return this;
   }
 
-  public ContainerTag getContainer() {
+  protected void genRandIdIfEmpty() {
```
syntapy, line 265: Maybe reorder methods so that diff shows that `getContainer` was not deleted or made `protected`/`private`

syntapy (commented):
Some things I've noted to probably change. But I only looked at half of it so far. Need to continue off from `FieldWithLabel.java`

syntapy:
@clouser2 I did a self review and found some points of improvement as well as where it could be made easier to review. How about I do those edits and then if you all want to review afterwards? Its a long diff and I think it took me about an hr to get halfway through lol

browser-test/yarn.lock
```diff
@@ -493,13 +493,27 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@colors/colors@1.5.0":
```
clouser2, line 496: Why were there changes to this file? As I understand, the J2HTML change should be purely a Java change.

browser-test/package.json
```diff
@@ -3,8 +3,8 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "scripts": {
-    "test": "jest --testTimeout=180000 --forceExit --maxConcurrency=1 --runInBand",
-    "probe": "jest --testTimeout=180000 --forceExit --maxConcurrency=1 --runInBand"
+    "test": "jest --testTimeout=18000000 --forceExit --maxConcurrency=1 --runInBand",
+    "probe": "jest --testTimeout=18000000 --forceExit --maxConcurrency=1 --runInBand"
```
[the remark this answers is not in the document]
clouser2: Blocking: Agreed.

server/app/views/HtmlBundle.java
```diff
@@ -26,30 +35,30 @@ public class HtmlBundle {
...
+  private ArrayList<LinkTag> stylesheets = new ArrayList<>();
   private ArrayList<ToastMessage> toastMessages = new ArrayList<>();
 
   public HtmlBundle addBodyStyles(String... styles) {
     bodyStyles.addAll(Arrays.asList(styles));
     return this;
   }
 
-  public HtmlBundle addFooterContent(Tag... tags) {
+  public HtmlBundle addFooterContent(DivTag... tags) {
```
clouser2, line 56: Is there any specific reason that we need to specify DivTags here? This feels like a case where we want to use a more general type rather than something specific.

server/app/views/HtmlBundle.java
```diff
@@ -59,12 +68,12 @@ public HtmlBundle addFooterStyles(String... styles) {
...
   }
 
-  public HtmlBundle addHeadScripts(Tag... sources) {
+  public HtmlBundle addHeadScripts(ScriptTag... sources) {
     headScripts.addAll(Arrays.asList(sources));
     return this;
   }
 
-  public HtmlBundle addHeaderContent(Tag... tags) {
+  public HtmlBundle addHeaderContent(NavTag... tags) {
```
clouser2, line 76: Same here. Can we use a more general Tag type?

clouser2 (changes requested):
Thanks for helping us get this upgraded! Overall, I really really really like having stronger types available for our HTML generation.

Took an initial look. I can do a more in-depth review once we've settled on a few of the higher-level questions I had.

browser-test/yarn.lock
syntapy: Isn't yarn.lock similar to package-lock.json? I didn't modify any of the packages in `browser-test/package.json` so these kinds of changes in yarn.lock would probably happen every time someone sets up their browser tests no?

server/app/views/FileUploadViewStrategy.java
```diff
@@ -188,20 +192,20 @@ private Tag renderDeleteAndContinueFileUploadForms(Params params) {
...
             .with(
                 each(
                     params.block().getQuestions(),
                     question -> renderFileKeyField(question, rendererParams)));
-    Tag deleteForm =
+    FormTag deleteForm =
         form()
             .withId(FILEUPLOAD_DELETE_FORM_ID)
-            .withAction(formAction)
+            .attr("action", formAction)
```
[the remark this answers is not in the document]
syntapy: Oh ok, apparently looking at the code again I guess `FormTag` does have that through the `IAction` interface

server/app/views/HtmlBundle.java
syntapy: Ok, but what about the `footerScripts` and `headScripts` fields? Should those be left as `ScriptTag`?

clouser2: I'd leave those as `ScriptTags`, yes. I think the general issue I was pointing out was that the method specifies content, which doesn't imply a specific tag type to me. In those cases, I think a generalized type (e.g. `Tag`) makes sense. In cases where callers would assume/need a specific type, it makes sense to return a more specific type.

There's no hard or fast rule here. One potential indicator that a more general type is needed is if we end up modifying the actual DOM structure returned just to satisfy the Java type system.

server/app/views/components/LinkElement.java
```diff
@@ -106,79 +108,96 @@ public LinkElement opensInNewTab() {
...
+      atag.withId(id).withClasses(DEFAULT_LINK_STYLES, styles);
+      if (doesOpenInNewTab) {
+        atag.attr("target", "_blank");
+      }
+      atag.attr("href", href);
+
+      tag = div().with(atag);
+
+      return tag;
+    }
```
syntapy, line 132: Not sure I know how to deal with this, but I am sure I didn't do it correctly as of now. The problem is that it either returns a `ATag` or a `DivTag` depending on whether there is an `href` defined. As of now it wraps the `ATag` with a `DivTag` and that causes problems for the `withTarget(..)` method

Maybe I can refactor it to split into two kinds of methods, one returning an `ATag` if there is an `href`, and the other returning a `DivTag` if there isnt. The caller should be able to tell which to use

clouser2: This might be a case where the method could return a more general type (e.g. `ContainerTag`). Then there'd be no need to wrap in a div. This is a case where I'd prefer we not have to restructure the DOM in order to return a consistent type. Downstream callers of `asAnchorText` hopefully won't need to care about the details of what the tag type actually is.

syntapy: I'm kind of snagged on an error [here](https://github.com/seattle-uat/civiform/blob/122815a732b8f5962894e4efb0c45c8161815332/server/app/views/applicant/ApplicantProgramSummaryView.java#L210)

error_3_trace

syntapy: I tried doing similarly with a minimal example and it worked:

error_3_works_minexample

syntapy: Oh nevermind. I reproduced the problem in the minimal example

error_3_minexample_reproduced_trace


error_3_minexample_reproduced

syntapy: @clouser2 what do you think of this approach? Instead of returning either an `a` or a `div` from the [asAnchorText](https://github.com/seattle-uat/civiform/blob/122815a732b8f5962894e4efb0c45c8161815332/server/app/views/components/LinkElement.java#L113-L119), or places like [here](https://github.com/seattle-uat/civiform/blob/122815a732b8f5962894e4efb0c45c8161815332/server/app/views/admin/programs/ProgramAdministratorProgramListView.java#L169-L185), we return an `Optional<ATag>` instead of either return an `ATag` or a `DivTag`?

I guess I'm not sure how an anchor tag can be valid if it doesn't have an `href`. As such, why potentially return a `div` from the `asAnchorText` method?

It might take a bit of work to try this out and see if the browser tests still pass

syntapy: Honestly, I think this should probably be future work. The browser tests had passed at some point, so maybe this should be a future PR

clouser2: It looks like you're running into a situation where our `LinkElement` abstraction definitely has some usability issues. `asAnchorText` previously returned either 1) an `<a>` tag if an HREF attribute was provided or 2) a `<div>` containing the link text, styled to look like a link (but not actually clickable).

This is definitely a strange behavior and not worth trying to change as part of this upgrade. Due to this, we don't want downstream callers to assume that it's always an `<a>` tag. I think this is a case for having the following method signature in `LinkElement`:

`ContainerTag<?> asAnchorText()`

Downstream callers could then use `ContainerTag<?>` as the return type. Then only methods available for all container elements would be available (namely `.with`). Any attempts to call `.setHref` / `.setTarget` would actually indicate a leakage of abstraction. Furthermore, those could call the corresponding `LinkElement.setHref` / `LinkElement.opensInNewTab` methods rather than calling them on the returned element.

I still think that we want to avoid restructuring the DOM due to the way that J2HTML works. In my opinion, any instances where this is occurring are likely code smells where the previous application code might not have actually been working as we intended (e.g. calling `.withHref` on a `<div>` tag would have had no practical effect.

syntapy: Ok, thats helpfull. Yeah adding the `<?>` to `ContainerTag` definitely fixes the compile error. Thank you

clouser2 (changes requested):
This is definitely looking really good! I've reviewed all files besides FieldWithLabel (which I'll do shortly).

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
-        fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
-      }
-    } else {
-      fieldTag.withValue(this.fieldValue);
+  }
+
+  protected LabelTag genLabelTag() {
+
+    return label()
+        .attr(Attr.FOR, this.id)
```
clouser2, line 310: This can replaced with `withFor`

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
-      // For number types, only set the value if it's present since there is no empty string
-      // equivalent for numbers.
-      if (this.fieldValueNumber.isPresent()) {
-        fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
-      }
-    } else {
-      fieldTag.withValue(this.fieldValue);
+  }
+
+  protected LabelTag genLabelTag() {
```
clouser2, line 307: Does this need to be protected? Can we instead make it private? Same for any other newly added protected methods.

Generally, I have a preference towards making things private to start with and opening them as needed to avoid exposing too many of the implementation details.

server/app/views/components/PathTag.java
```diff
@@ -0,0 +1,9 @@
+package views.components;
+
+import j2html.tags.ContainerTag;
+
+public final class PathTag extends ContainerTag<PathTag> {
+  public PathTag() {
+    super("svg");
```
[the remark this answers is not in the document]
syntapy: Oh, I think I forgot to change to "path" when I copied SvgTag.java

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
+
+    return label()
+        .attr(Attr.FOR, this.id)
+        // If the text is screen-reader text, then we want the label to be screen-reader
+        // only.
+        .withClass(labelText.isEmpty() ? Styles.SR_ONLY : BaseStyles.INPUT_LABEL)
+        .withText(labelText.isEmpty() ? screenReaderText : labelText);
+  }
+
+  protected DivTag wrapInDivTag(Tag fieldTag, Tag labelTag, String fieldErrorsId) {
```
clouser2, line 317: I'd give this method a name more specific to what it's doing. Is this building the base container or something else? `wrapInDivTag` implies that it's a utility method, but it appears that the logic here is quite specific to the use.

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
+    return div(
+            labelTag,
+            div(fieldTag, buildFieldErrorsTag(fieldErrorsId))
+                .withClasses(Styles.FLEX, Styles.FLEX_COL))
+        .withClasses(
+            StyleUtils.joinStyles(referenceClassesBuilder.build().toArray(new String[0])),
+            BaseStyles.FORM_FIELD_MARGIN_BOTTOM);
+  }
+
+  protected boolean getHasFieldErrors() {
```
clouser2, line 327: Consider naming this `hasFieldErrors` instead.

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
+    return !fieldErrors.isEmpty() && showFieldErrors;
+  }
+
+  protected void numberTagApplyAttrs(Tag fieldTag) {
```
clouser2, line 331: Can this argument be specifies as an `InputTag`?

If so, we can call `.setStep` and `setValue` below rather than calling that attr setters directly.

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
-      fieldTag.attr("aria-invalid", "true");
-      fieldTag.attr("aria-describedBy", fieldErrorsId);
+    // For number types, only set the value if it's present since there is no empty string
+    // equivalent for numbers.
+    if (this.fieldValueNumber.isPresent()) {
+      fieldTag.attr("value", String.valueOf(this.fieldValueNumber.getAsLong()));
     }
+  }
 
+  protected void generalApplyAttrsClassesToTag(Tag fieldTag, boolean hasFieldErrors) {
```
clouser2, line 354: Can this take a more specific type? If we're setting the "name" and "placeholder" attributes, we should ensure that the specified tag passed in supports them.

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
-    return label()
-        .withClasses(
-            StyleUtils.joinStyles(referenceClassesBuilder.build().toArray(new String[0])),
-            BaseStyles.CHECKBOX_LABEL,
-            BaseStyles.FORM_FIELD_MARGIN_BOTTOM,
-            labelText.isEmpty() ? Styles.W_MIN : "")
-        .condAttr(!this.id.isEmpty(), Attr.FOR, this.id)
-        .with(fieldTag.withClasses(BaseStyles.CHECKBOX))
-        .withText(this.labelText);
+    return div()
```
clouser2, line 395: Can we avoid wrapping this in a div?

clouser2 (changes requested):

server/app/views/components/QuestionBank.java
```diff
@@ -58,8 +61,8 @@ public QuestionBank setQuestionAction(String actionUrl) {
     return this;
   }
 
-  public QuestionBank setCsrfTag(Tag csrfTag) {
-    this.csrfTag = csrfTag;
+  public QuestionBank setCsrfTag(InputTag csrfTag) {
+    this.csrfTag.with(csrfTag);
```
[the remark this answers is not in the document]
syntapy: Good question. Main reason I did it was cuz the `csrfTag` field was initialized to a `div()` in the class def. And then redefining it as a `InputTag` later wouldn't work

I gotta work out how to do this. Maybe we don't need to initialize the `csrfTag` in the calass def

server/app/views/components/FieldWithLabel.java
syntapy: Otherwise done

server/app/views/components/FieldWithLabel.java
syntapy: Ok, I changed to `buildBaseContainer`. Does that work?

server/app/views/admin/programs/ProgramAdministratorProgramListView.java
```diff
@@ -162,7 +165,8 @@ private String extractProgramStatusText(
     throw new IllegalArgumentException("Program neither active nor draft.");
   }
 
-  Tag maybeRenderViewApplicationsLink(String text, Optional<ProgramDefinition> activeProgram) {
+  ContainerTag<?> maybeRenderViewApplicationsLink(
```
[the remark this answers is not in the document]
clouser2: I was mostly curious as to whether returning `Tag` rather than `ContainerTag` would work. This one's really minor, so if you'd prefer to keep as `ContainerTag` and it doesn't require any downstream wrapping, works for me.

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -144,28 +146,34 @@ private ContainerTag getLanguageForm(
...
                 .attr("aria-label", messages.at(MessageKey.LANGUAGE_LABEL_SR.getKeyName()));
         languageForm =
-            form()
-                .withAction(updateLanguageAction)
-                .withMethod(Http.HttpVerbs.POST)
-                .with(csrfInput)
-                .with(redirectInput)
-                .with(languageDropdown)
-                .with(TagCreator.button().withId("cf-update-lang").withType("submit").isHidden());
+            languageForm.with(
```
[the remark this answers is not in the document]
clouser2: Marking unresolved since this still appears to be creating a subform.

server/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -124,26 +124,28 @@ private Tag renderBlockWithSubmitForm(Params params) {
...
-        .withId(BLOCK_FORM_ID)
-        .withAction(formAction)
-        .withMethod(HttpVerbs.POST)
-        .with(makeCsrfTokenInputTag(params.request()))
-        .with(
-            each(
-                params.block().getQuestions(),
-                question -> renderQuestion(question, rendererParams)))
-        .with(renderBottomNavButtons(params));
+    return div(
```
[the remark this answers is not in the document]
clouser2: Unresolving since this still appears to be the case.

server/app/views/applicant/ApplicantLayout.java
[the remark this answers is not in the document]
syntapy: Oh I havn'et pushed yet, sorry!

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +243,183 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
+  protected FieldWithLabel setTagTypeTextarea() {
+    tagType = "textarea";
+    return this;
+  }
+
+  protected boolean isTagTypeInput() {
+    return tagType.equals("input");
+  }
+
+  protected boolean isTagTypeTextarea() {
```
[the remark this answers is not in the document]
syntapy: For whatever reason when I tried marking as private it resulted in `Warnings found and -Werror specified` or something. Aaand `SelectWithLabel` isn't using them so I'm confused

server/app/views/applicant/ApplicantLayout.java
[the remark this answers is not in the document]
syntapy: From now on I will not mark as resolved untill **after** I push. Sorry about the confusion!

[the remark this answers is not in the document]
clouser2: No worries. After I re-opened a few, I thought that might have been the case and thought I'd just wait to look again until the next set of pushes.

server/app/views/admin/programs/ProgramAdministratorProgramListView.java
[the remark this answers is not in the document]
syntapy: Ok, I changed a few to `Tag` and it compiled. Would you be able to run my CI tests before you leave? Since I'm a community contributor I need one of you all to run my tests

bion (changes requested):

syntapy:
Need to
- [ ] Make sure there are no general calls to `.attr(<attribute>, <value>)` when there is a more type-specific method available. This may involve refactoring such as [here](https://github.com/seattle-uat/civiform/blob/53dddfc21d9f8b609fcb8eafec2e74c1246727ba/server/app/views/components/FieldWithLabel.java#L335-L346)
- [ ] Refactor `FieldWithlabel` to not use any general types like `ContainerTag`. Since it can return either `LabelTag`, `InputTag`, or `TextareaTag`, I will probably need a different `getContainer` or whatnot for each of those
- [ ] Make sure there are *no* DOM changes whatsoever

server/app/views/components/FieldWithLabel.java
syntapy: As is, this method takes in an `InputTag`, a `TextareaTag`, and in the subclass a `SelectTag`

The `.withClasses`, `.withId`, and `.condAttr` are from `Tag` while the `.withName` is from the `IName` interface

I think I found a way to make it work with all three types being passed to it

What do you think? Although its not entirely 'specific' but it is a bit more I guess?

server/app/views/BaseHtmlLayout.java
```diff
@@ -105,23 +105,23 @@ public Content render(HtmlBundle bundle) {
...
 
   /** Creates Google Analytics scripts for the site. */
-  private ImmutableList<Tag> getAnalyticsScripts(String trackingTag) {
-    Tag scriptImport =
+  private ImmutableList<ScriptTag> getAnalyticsScripts(String trackingTag) {
+    ScriptTag scriptImport =
         script()
             .withSrc("https://www.googletagmanager.com/gtag/js?id=" + trackingTag)
             .attr("async", "true")
-            .withType("text/javascript");
```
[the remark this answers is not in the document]
syntapy: I think I done it!

server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -254,15 +256,19 @@ private ContainerTag blockMoveButtons(
...
             .with(
                 form()
                     .withAction(moveUpFormAction)
                     .withMethod(HttpVerbs.POST)
                     .with(makeCsrfTokenInputTag(request))
-                    .with(input().isHidden().withName("direction").withValue(Direction.UP.name()))
+                    .with(
+                        input()
+                            .isHidden()
+                            .attr("name", "direction")
```
clouser2, line 270: Mind switching to `withName` / `withValue`

Same for other `.attr` references in this and other files.

server/app/views/components/FieldWithLabel.java
```diff
@@ -246,107 +245,166 @@ public FieldWithLabel showFieldErrors(boolean showFieldErrors) {
...
-      fieldTag.attr("aria-describedBy", fieldErrorsId);
+    // For number types, only set the value if it's present since there is no empty string
+    // equivalent for numbers.
+    if (this.fieldValueNumber.isPresent()) {
+      fieldTag.withValue(String.valueOf(this.fieldValueNumber.getAsLong()));
     }
+  }
 
+  private <T extends Tag<T> & IName<T> & IDisabled<T>> void generalApplyAttrsClassesToTag(
+      T fieldTag, boolean hasFieldErrors) {
```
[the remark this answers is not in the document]
clouser2: Given that this is a private method that knows the details on how the relevant FieldWithLabels are constructed, I'd be ok losing a bit of type safety here in order to avoid having long generic type declarations.

Using a `.attr` or `.condAttr` call here with a comment as to why we're not using a more type-safe version seems like it'd be an ok tradeoff. This would reduce the number of potentially moving parts in this CL.

RE separate subclasses per field type: Maybe. I'd keep it separate from this CL to avoid adding extra complexity.

clouser2 (changes requested):

server/app/views/admin/programs/ProgramBlockEditView.java
syntapy: Oh crap I thought I had gotten all those. Maybe those are from the merge?

syntapy:
Is [this](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/applicant/TrustedIntermediaryDashboardView.java#L174) supposed to be `FieldWithLabel.email()` instead of `FieldWithLabel.input()`?

syntapy:
Ok, so I realized another issue

We have lists of `FieldWithLabel` objects. E.g. over [here](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/admin/programs/ProgramTranslationView.java#L81-L96), which is later iterated through with the `::getContainer(..)` method [here](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/admin/TranslationFormView.java#L90)

Now, all the lists of type `<FieldWithLabel>` all have `input()` type fields (i.e. not `checkbox()` or `textarea()` etc.., so if we have a get method for each container type it should still 'work'

But... its kinda ugly. It might be good to have those all just return `ContainerTag`, which makes the type safety enhancement leave the internal methods and possibly pollute the rest of the code

Splitting each field type into a separate class would have the problem of a list with multiple types or using a superlace type and trying to call a child class's `getContainer` method on each obj

But maybe I can use [reflection](https://stackoverflow.com/questions/37628/what-is-reflection-and-why-is-it-useful)?

clouser2:
> Is [this](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/applicant/TrustedIntermediaryDashboardView.java#L174) supposed to be `FieldWithLabel.email()` instead of `FieldWithLabel.input()`?

Sure looks like it. Rather than fixing here, I'd suggest filing an issue for cleanup and leaving a TODO here. Even though the effort would likely be small to fix, I prefer trying to keep refactorings like this as scoped as possible.

clouser2:
> Ok, so I realized an issue
> 
> We have lists of `FieldWithLabel` objects. E.g. over [here](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/admin/programs/ProgramTranslationView.java#L81-L96), which is later iterated through with the `::getContainer(..)` method [here](https://github.com/seattle-uat/civiform/blob/8604ba9960811194e52809a7fe2d369cdeaf5bda/server/app/views/admin/TranslationFormView.java#L90)
> 
> Now, all the lists of type `<FieldWithLabel>` all have `input()` type fields (i.e. not `checkbox()` or `textarea()` etc.., so if we have a get method for each container type it should still 'work'
> 
> But... its kinda ugly. It might be good to have those all just return `ContainerTag`, which makes the type safety enhancement leave the internal methods and possibly pollute the rest of the code
> 
> Splitting each field type into a separate class would have the problem of a list with multiple types or using a superlace type and trying to call a child class's `getContainer` method on each obj
> 
> But maybe I can use [reflection](https://stackoverflow.com/questions/37628/what-is-reflection-and-why-is-it-useful)?

I'd recommend against reflection as the APIs are a bit complicted and there's a runtime penalty. My general view is that if code is expecting a list of `FieldWithLabels`, it inherently shouldn't assume anything about the structure of the rendered element. In my opinion, it'd make sense to return a `ContainerTag`, semi-documenting that downstream callers shouldn't care about the return type. Anything that assumed specifics about the rendered element (e.g. it's an `input`) would be a code smell to be fixed independently.

If possible, I'd suggest avoiding a refactoring of this type in the current PR so we can keep it scoped only to the J2HTML upgrade.

syntapy:
current failing browser tests in `run-browser-tests-ci`
- `question_lifecycle.test.ts` (didn't find string 'Applicant Name')
- `application_download.test.ts` (some mismatch between really long strings)
- `currency.test.ts` (timeout error)
- `file.test.ts` (timeout error)

failing tests from `run-browser-tests-local`
- `application_review.test.ts`
- `question_lifecycle.test.ts`
- `application_download.test.ts`
- `file.test.ts`

server/app/views/components/Modal.java
```diff
@@ -48,36 +49,36 @@ public Tag getButton() {
     }
   }
 
-  private Tag getContent() {
+  private Tag<?> getContent() {
```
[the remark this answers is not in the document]
syntapy: yeah, no idea wtf I was thinking ther

server/app/views/components/PathTag.java
```diff
@@ -0,0 +1,9 @@
+package views.components;
+
+import j2html.tags.ContainerTag;
+
+public final class PathTag extends ContainerTag<PathTag> {
+  public PathTag() {
+    super("path");
+  }
+}
```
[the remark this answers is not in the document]
syntapy: Would it be nicer if I refactor to this?

```java
package views.components;

import j2html.tags.ContainerTag;

public final class NiceTag extends ContainerTag<NiceTag> {
  public NiceTag() {
    super("path");
  }
}
```

server/app/views/admin/TranslationFormView.java
```diff
@@ -75,17 +77,17 @@ private ContainerTag renderLanguageLink(
...
       String formAction,
       ImmutableList<FieldWithLabel> formFields) {
-    ContainerTag form =
+    FormTag form =
         form()
             .withMethod("POST")
             .with(makeCsrfTokenInputTag(request))
             .withAction(formAction)
-            .with(each(formFields, FieldWithLabel::getContainer))
+            .with(each(formFields, FieldWithLabel::getInputTag)) // TODO fix this ugliness
```
clouser2, line 90: Prefer TODOs using a specific issue number (e.g. the J2HTML issue) so we don't lose context on why the TODO is needed.

Additionally, when we eventually go to close the issue, we can check for any outstanding TODOs prior to considering the work "done".

server/app/views/admin/programs/ProgramAdministratorProgramListView.java
```diff
@@ -116,32 +118,33 @@ public Tag renderProgramListItem(
...
             .withClasses(Styles.FLEX, Styles.TEXT_SM, Styles.W_FULL);
 
-    Tag programDeepLink =
+    LabelTag programDeepLink =
         label("Deep link, use this URL to link to this program from outside of CiviForm:")
             .withClasses(Styles.W_FULL)
             .with(
                 input()
-                    .withValue(
+                    .attr(
```
clouser2, line 137: I believe this is incorrect. The new version of the code sets an attribute of the URL rather than setting the "value" attribute to the URL.

I think we should switch this back to `withValue`.

server/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -310,22 +314,27 @@ private ContainerTag createScalarDropdown(QuestionDefinition questionDefinition)
...
                   operator
                       .getOperableTypes()
-                      .forEach(type -> option.attr("data-" + type.name().toLowerCase()));
+                      .forEach(
+                          type ->
+                              option.attr(
+                                  "data-",
+                                  type.name()
+                                      .toLowerCase())); // TODO probably a bug since a value needs
+                  // to be passed?
```
clouser2, line 337: Can we add a TODO with a specific issue number?

I'm not sure I follow on what the bug is here. Mind expanding the TODO to capture it?

server/app/views/admin/ti/EditTrustedIntermediaryGroupView.java
```diff
@@ -66,8 +69,9 @@ public Content render(TrustedIntermediaryGroup tiGroup, Http.Request request) {
     return layout.renderCentered(htmlBundle);
   }
 
-  private Tag renderAddNewButton(TrustedIntermediaryGroup tiGroup, Http.Request request) {
-    ContainerTag formTag =
+  // TODO emailField should call `.email()` instead of `.input()`
```
clouser2, line 72: Please use a TODO against a specific issue number.

server/build.sbt
```diff
@@ -202,3 +202,6 @@ resolveFromWebjarsNodeModulesDir := true
 playRunHooks += TailwindBuilder(baseDirectory.value)
 // Reload when the build.sbt file changes.
 Global / onChangedBuildSource := ReloadOnSourceChanges
+// uncomment to show debug logging.
+logLevel := Level.Debug
```
clouser2, line 206: Blocking: Comment this out prior to merging.

server/app/views/components/QuestionBank.java
```diff
@@ -68,32 +70,32 @@ public QuestionBank setQuestions(ImmutableList<QuestionDefinition> questionDefin
...
+  public FormTag getContainer() {
     return questionBankPanel();
   }
 
-  private ContainerTag questionBankPanel() {
-    ContainerTag questionForm =
-        form(this.csrfTag).withMethod(HttpVerbs.POST).withAction(questionAction);
+  private FormTag questionBankPanel() throws RuntimeException {
+    if (this.maybeCsrfTag.isEmpty()) {
+      throw new RuntimeException("trying to access empty csrf tag for rendering");
```
clouser2, line 79: I'd consider not throwing an exception here and instead just letting the user request fail downstream.

clouser2 (commented):
Reviewed everything except for FieldWithLabel / SelectWithLabel-related changes. I'm going to defer to bion@ for that part of the review.

server/app/views/admin/ti/EditTrustedIntermediaryGroupView.java
syntapy: https://github.com/seattle-uat/civiform/issues/2762

server/app/views/admin/programs/ProgramBlockPredicatesEditView.java
syntapy: I found the with `withData` method this time

server/app/views/admin/TranslationFormView.java
syntapy: https://github.com/seattle-uat/civiform/issues/2764

server/app/views/components/QuestionBank.java
syntapy: should I do the same [here](https://github.com/syntapy/universal-application-tool/blob/31e95063d292f4e0bbdf024eb8d5ce3ff4424b4e/server/app/views/components/LinkElement.java#L123-L129) too?

server/test/views/questiontypes/RadioButtonQuestionRendererTest.java
```diff
@@ -75,12 +75,12 @@ public void render_generatesIds_formatsWhitespaceAsUnderscore() {
...
         applicantData, question.getContextualizedPath(), 2L);
-    Tag result = renderer.render(params);
+    DivTag result = renderer.render(params);
 
     assertThat(result.render())
         .contains(
             "<input id=\"peanut_butter\" type=\"radio\""
                 + " name=\"applicant.favorite_ice_cream.selection\""
-                + " value=\"2\" checked=\"\"");
+                + " value=\"2\" checked");
```
syntapy, line 84: Updated cuz as a result of using `.isChecked` over [here](https://github.com/seattle-uat/civiform/blob/3c2e72e928c195c653dedf0efb82af36e34373ca/server/app/views/components/FieldWithLabel.java#L470) insead of the old `.attr(..)` over [here](https://github.com/seattle-uat/civiform/blob/1f74106e27bfb7f777dbb91b98f411a101dad933/server/app/views/components/FieldWithLabel.java#L328)

syntapy:
I might need to add another `dropTables` in the `beforeAll` for `file.test.ts` to get the failling batch azurite browser test working. Currently double checking on `main` before adding a separate PR for this

Otherwise as is I can get `file.test.ts` to pass on the azurite browser-test-env passing after a couple tries when ran alone

server/app/views/admin/programs/ProgramBlockPredicatesEditView.java
```diff
@@ -310,22 +314,22 @@ private ContainerTag createScalarDropdown(QuestionDefinition questionDefinition)
...
                   // whether to show or hide each operator based on the current type of scalar
                   // selected.
-                  ContainerTag option =
-                      option(operator.toDisplayString()).withValue(operator.name());
+                  OptionTag option = option(operator.toDisplayString()).withValue(operator.name());
                   operator
                       .getOperableTypes()
-                      .forEach(type -> option.attr("data-" + type.name().toLowerCase()));
+                      .forEach(type -> option.withData(type.name().toLowerCase(), ""));
+                  // to be passed?
```
clouser2, line 332: Can this now be removed?

server/app/views/components/LinkElement.java
```diff
@@ -106,79 +112,98 @@ public LinkElement opensInNewTab() {
...
 
-  public ContainerTag asButton() {
-    ContainerTag tag =
-        Strings.isNullOrEmpty(href)
-            ? div(text)
-            : a(text).withHref(href).withCondTarget(doesOpenInNewTab, "_blank");
-    return tag.withCondId(!Strings.isNullOrEmpty(id), id)
-        .withClasses(DEFAULT_LINK_BUTTON_STYLES, styles);
+  private <T extends ContainerTag<T> & IHref<T> & ITarget<T>> void setTargetMaybeHref(T tag)
+      throws RuntimeException {
```
clouser2, line 124: Not blocking: There's no need to explicitly declare this as it's an unchecked exeception.

clouser2 (approved):
Nice work and thanks for sticking with this :)

I looked at all changes besides FieldWithLabel.

I'd defer to bion@'s review on that before merging.

Additionally, I patched the change in locally and did some basic testing of the UI and didn't notice any obvious differences in rendering!

server/app/views/admin/programs/ProgramBlockPredicatesEditView.java
syntapy: you mean just delete the `.forEach(..)` call?

syntapy: oh, the comment (facepalm lol)

server/app/views/components/LinkElement.java
syntapy: so you mean get rid of the `throws RuntimeException`?

gwendolyngoetz:
Think look to be rendering correctly. Even the static text bulleted lists and accordions.

server/app/views/components/LinkElement.java
clouser2: Yes, sorry about that.

bion (approved):

PR #2284, opened 2022-04-12
Validate all file upload inputs

### LINKED ISSUE #1695: FileUpload validation only uses the status of the last question
https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/assets/javascripts/validation.ts#L343

Multiple upload questions can be in a block but only one/first is validated.

### LINKED ISSUE #1696: Add focused testing for FileUpload Question
There is currently no focused test for the FileUpload question.

https://github.com/seattle-uat/civiform/blob/main/browser-test/src/address.test.ts

can be used as a template to test.

clouser2:
### Description
On-page file upload validation now validates ALL file upload questions in a block rather than assuming there is just one. Also fixes issue where Azure uploads threw an uncaught exception if a required file hadn't been uploaded (rather than showing a validation error to the user).

Note: Only one file upload is allowed per block by the admin controller. That said, there's not a strong need to bake that in at the validation level.

While here, I also:
  1) Renamed `validateFooQuestion` method names to `validateFooQuestions` to indicate they would validate all questions rather than just a single question.
  2) Added focused tests for file upload questions (which exposed a bug in the Azure file upload path, also fixed here).

### Checklist

### Issue(s)
Fixes #1695; Fixes #1696

### DIFF
--- a/browser-test/src/file.test.ts
+++ b/browser-test/src/file.test.ts
@@ -1,15 +1,15 @@
 import {
-    AdminPrograms,
-    AdminQuestions,
-    ApplicantQuestions,
-    loginAsAdmin,
-    loginAsGuest,
-    logout,
-    resetSession,
-    selectApplicantLanguage,
-    startSession,
-  } from './support'
-  
+  AdminPrograms,
+  AdminQuestions,
+  ApplicantQuestions,
+  loginAsAdmin,
+  loginAsGuest,
+  logout,
+  resetSession,
+  selectApplicantLanguage,
+  startSession,
+} from './support'
+
 describe('file upload applicant flow', () => {
   let pageObject
 
@@ -115,7 +115,7 @@ describe('file upload applicant flow', () => {
       const error = await pageObject.$('.cf-fileupload-error')
       expect(await error.isHidden()).toEqual(false)
       await applicantQuestions.clickSkip()
-      
+
       await applicantQuestions.submitFromReviewPage(programName)
     })
 
--- a/browser-test/src/file.test.ts
+++ b/browser-test/src/file.test.ts
@@ -1,0 +1,132 @@
+import {
+    AdminPrograms,
+    AdminQuestions,
+    ApplicantQuestions,
+    loginAsAdmin,
+    loginAsGuest,
+    logout,
+    resetSession,
+    selectApplicantLanguage,
+    startSession,
+  } from './support'
+  
+  describe('file upload applicant flow', () => {
+    let pageObject
+  
+    beforeAll(async () => {
+      const { page } = await startSession()
+      pageObject = page
+    })
+  
+    afterEach(async () => {
+      await resetSession(pageObject)
+    })
+  
+    describe('single file upload question', () => {
+      let applicantQuestions
+      const programName = 'test program for single file upload'
+  
+      beforeAll(async () => {
+        await loginAsAdmin(pageObject)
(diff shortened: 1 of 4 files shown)


browser-test/src/file.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        await loginAsGuest(pageObject)
+        await selectApplicantLanguage(pageObject, 'English')
+
+        await applicantQuestions.applyProgram(programName)
+        await applicantQuestions.clickSkip()
+        await applicantQuestions.submitFromReviewPage(programName)
+      })
+    })
+  })
+
```
shanemc-goog, line 132: need a trailing newline it looks like?

shanemc-goog (changes requested):
Looks good to me, any idea what's up with the test?

browser-test/src/file.test.ts
clouser2: Fixed.

clouser2:
> Looks good to me, any idea what's up with the test?

The failing test actually exposed a bug in the Azure file upload logic where `azure_upload.ts` was triggering prior to the code in `validation.ts`. This then led to an attempt to assume the file had always been uploaded, leading to an uncaught exception when that wasn't the case.

To fix this, I reworked how Azure uploads worked so that the logic in `validation.ts` is triggered first, which now correctly shows an error when attempting to click "Upload" without first having selected a file.

These changes are definitely worth taking another look. All tests are now passing and I've taken the pull request out of draft mode.

browser-test/src/file.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        [],
+        'file-upload-test-optional-q'
+      )
+      await adminPrograms.gotoAdminProgramsPage()
+      await adminPrograms.publishAllPrograms()
+
+      await logout(pageObject)
+    })
+
+    it('with missing file can be skipped', async () => {
```
goldblatt, line 106: can it though?

shanemc-goog:
Thanks, I can look in the morning

shanemc-goog:
How can I see the existing incorrect behavior that you're fixing in this PR?

shanemc-goog (approved):

browser-test/src/file.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  ApplicantQuestions,
+  loginAsAdmin,
+  loginAsGuest,
+  logout,
+  resetSession,
+  selectApplicantLanguage,
+  startSession,
+} from './support'
+
+describe('file upload applicant flow', () => {
```
shanemc-goog, line 13: I don't see a case for 2 upload questions.  Was that not actually doable after all?

The name question tests should have examples for you if that's helpful.

shanemc-goog (changes requested):

browser-test/src/file.test.ts
clouser2: The UI actually appears to disallow adding more than one file upload question to a single screen. I originally wrote a test which attempted this and ran into that failure.

shanemc-goog: Ah sorry, for some reason I thought Name was similar in only allowing one.

shanemc-goog (approved):

browser-test/src/file.test.ts
clouser2: No worries at all. Resolving this.

PR #2285, opened 2022-04-13
Line break question action items to be clearer.

shanemc-goog:
### Description
Add a `<BR>` after each question action for clearer alignment and usability.

Before:
before

After:
after


goldblatt (approved):
everything about this is better

PR #2295, opened 2022-04-14
Admin UI tweaks

bion:
- Increase the max width of the `main` element for the admin so Seattle's question content doesn't require a horizontal scrollbar
- Reverse the list of versions so most recent is at the top

### DIFF
--- a/universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
@@ -53,8 +53,9 @@ public class VersionListView extends BaseHtmlView {
     ImmutableList<Version> olderVersions =
         allVersions.stream()
             .filter(version -> version.getLifecycleStage().equals(LifecycleStage.OBSOLETE))
-            .collect(ImmutableList.toImmutableList())
-            .reverse();
+            // Sort from newest to oldest. IDs are DB-generated and increment monotonically.
+            .sorted((a, b) -> (int) (b.id - a.id))
+            .collect(ImmutableList.toImmutableList());
 
     String title = "Program Versions";
     HtmlBundle htmlBundle =
--- a/universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
+++ b/universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
@@ -53,7 +53,7 @@ public class VersionListView extends BaseHtmlView {
     ImmutableList<Version> olderVersions =
         allVersions.stream()
             .filter(version -> version.getLifecycleStage().equals(LifecycleStage.OBSOLETE))
-            .collect(ImmutableList.toImmutableList());
+            .collect(ImmutableList.toImmutableList()).reverse();
 
     String title = "Program Versions";
     HtmlBundle htmlBundle =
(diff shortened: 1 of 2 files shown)


universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
```diff
@@ -53,7 +53,8 @@ public Content render(List<Version> allVersions, Http.Request request) {
     ImmutableList<Version> olderVersions =
         allVersions.stream()
             .filter(version -> version.getLifecycleStage().equals(LifecycleStage.OBSOLETE))
-            .collect(ImmutableList.toImmutableList());
+            .collect(ImmutableList.toImmutableList())
+            .reverse();
```
shanemc-goog, line 57: suggestion: We likely need to explicitly sort here too.  The method does not require they be passed in sorted or in any particular order, and it's likely a matter of happenstance that they are ordered and in a particular way that we want to arbitrarily reverse.  So I think we need to sort here too.

universal-application-tool-0.0.1/app/views/style/AdminStyles.java
```diff
@@ -72,7 +72,7 @@ public class AdminStyles {
           Styles.FLEX);
 
   public static final String MAIN_CENTERED =
-      StyleUtils.joinStyles(Styles.PX_2, Styles.MAX_W_SCREEN_XL, Styles.MX_AUTO);
+      StyleUtils.joinStyles(Styles.PX_2, Styles.MAX_W_SCREEN_2XL, Styles.MX_AUTO);
```
shanemc-goog, line 75: woohoo!!!!

universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
bion: added explicit sorting

universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
```diff
@@ -53,6 +53,8 @@ public Content render(List<Version> allVersions, Http.Request request) {
     ImmutableList<Version> olderVersions =
         allVersions.stream()
             .filter(version -> version.getLifecycleStage().equals(LifecycleStage.OBSOLETE))
+            // Sort from newest to oldest. IDs are DB-generated and increment monotonically.
+            .sorted((a, b) -> (int) (b.id - a.id))
```
shanemc-goog, line 57: optional suggestion for readability and remove the casting.

.sorted((a, b) -> a.id.compareTo(b.id))
.reversed



should do the tricky here.

shanemc-goog (approved):
Do we care about this enough to add a change detection test?

universal-application-tool-0.0.1/app/views/admin/versions/VersionListView.java
bion: nice, thank you!

bion:
@shanemc-goog 
> Do we care about this enough to add a change detection test?

you mean for the ordering? i'm inclined not to since it'd be a pretty ugly unit test (we generally don't unit test views) and imo it's not worth the browser test overhead.

PR #2297, opened 2022-04-15
More compose

kmcnellis:
### Description
- Allow for docker compose to be used in more commands, removing extra hardcoded arguments.  
- Remove docker-compose container_names, allowing for multiple compose projects to be running at once.
- Simplify the browser test docker-compose file to mostly inherit from the base docker-compose file
- Allow dev, unit test, and browser tests to run at the same time locally
- Rename bin/dump-dev-program-config -> bin/dev-dump-program-config and allow specifying file name for dumping 

### Checklist
- [x] Ran the affected scripts


bin/run-test-ci
```diff
@@ -3,16 +3,19 @@
 # DOC: Run the unit tests in CI mode.
 
 export COMPOSE_PROJECT_NAME=test-support
-
 source bin/lib.sh
 
-docker::ensure_unit_test_env
```
shanemc-goog, line 9: What's the general thought behind removing this utility and inline it?

civiform.code-workspace
```diff
@@ -12,5 +12,8 @@
 			"**/target": true
 		},
 		"java.configuration.updateBuildConfiguration": "automatic"
+	},
+	"extensions": {
+		"recommendations": []
```
shanemc-goog, line 17: what is this file and these lines?  Is it related to compose?

docker-compose.dev.yml
```diff
@@ -33,10 +32,27 @@ services:
...
   db:
     image: postgres:12.5
+    stdin_open: true
+    tty: true
     volumes:
       # isolate and persist dev database.
+      - db-data:/var/lib/postgresql/data
       - project-data:/project
+
+  pgadmin:
```
shanemc-goog, line 42: with the additions here and in the unit test yml files are you removing the composition with docker-compose.yml or is something else going on?

browser-test/browser-test-compose.yml
```diff
@@ -1,67 +1,37 @@
-# Use postgres/example user/passwordcredentials
-version: '3.1'
+# Builds on docker-compose.yml
+version: "3.4"
 
 services:
-
   localstack:
-    image: localstack/localstack
```
shanemc-goog, line 7: Where are the images being specified now?  Were these redundant with the "base" yml file?

shanemc-goog (commented):
As much as I understand docker configs this looks good, had a few questions.

bin/attach-dev
```diff
@@ -20,4 +20,4 @@ echo "Civiform starting locally at http://localhost:9000/loginForm"
 
 echo "Run bin/stop-dev to stop the server"
 
-docker attach civiform
+docker attach "${COMPOSE_PROJECT_NAME}-civiform-1"
```
goldblatt, line 23: i only see this getting set for the tests is it supposed to get set somewhere for just running these commands too?

goldblatt (changes requested):
Looks good just a question around the env variables. Otherwise makes sense to me!

bin/dev-pgadmin
```diff
@@ -24,17 +24,12 @@ tee /tmp/servers.json > /dev/null <<EOT
...
-  -e 'PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED=False' \
-  -v /tmp/servers.json:/pgadmin4/servers.json \
-  -d dpage/pgadmin4 
+docker compose \
+    -f docker-compose.yml \
+    -f docker-compose.dev.yml \
+    --profile pgadmin \
+    up pgadmin \
+    --wait \
+    -d
```
bion, line 33: this isn't covered by a test, so please confirm the script still works

bin/dev-psql
```diff
@@ -6,10 +6,18 @@ source bin/lib.sh
...
+${COMPOSE_CMD} \
+  up db \
+  --no-deps \
+  --wait \
+  -d
+${COMPOSE_CMD} exec \
   -e "PGPASSWORD=example" \
-  --network "${DOCKER_NETWORK_NAME}" \
-  postgres:12.5 \
+  db \
```
bion, line 20: this isn't covered by a test, so please confirm the script still works with manual testing

bin/dev-restore-db
```diff
@@ -29,21 +29,27 @@ TRUNCATE_SQL="${TRUNCATE_SQL//$'\n'/}"
...
-  -v "$(pwd):/project" \
-  -e "PGPASSWORD=example" \
-  --network "${DOCKER_NETWORK_NAME}" \
-  postgres:12.5 \
-  /bin/bash -c "${CMD}"
+# command for invoking pg_restore with the dev database
+${COMPOSE_CMD} \
+  exec -it -e "PGPASSWORD=example" db \
+  /usr/bin/pg_restore -h db -U postgres \
+   -f "${DUMP_FILE}"
```
bion, line 55: this isn't covered by a test, so please confirm the script still works with manual testing

bin/dump-dev-program-config
```diff
@@ -21,4 +16,17 @@ docker run --rm -it \
...
+  -f docker-compose.dev.yml"
+
+${COMPOSE_CMD} \
+  up db \
+  --no-deps \
+  --wait \
+  -d
+
+${COMPOSE_CMD} exec -it -e "PGPASSWORD=example" db ${CMD}
+${COMPOSE_CMD} cp db:/project/dev_programs.dump dev_programs.dump
```
bion, line 32: this isn't covered by a test, so please confirm the script still works with manual testing

bin/attach-dev
kmcnellis: If you checkout civiform in a different folder, docker uses that folder as the prefix.  COMPOSE_PROJECT_NAME captures that and generates the container name accordingly.

kmcnellis: It's mostly just unfortunate that there's no docker-compose command for this

bin/dev-restore-db
kmcnellis: confirmed

bion (dismissed):

docker-compose.dev.yml
```diff
@@ -21,23 +21,52 @@ services:
...
-    container_name: civiform-shell
     image: civiform-dev
     entrypoint: /bin/bash
     profiles: ["shell"]
     volumes: *sbt-volumes
     stdin_open: true # docker run -i
     tty: true        # docker run -t
 
+  localstack:
+    ports:
```
goldblatt, line 38: nit: can this be expose instead of the port mapping

goldblatt (approved):

shanemc-goog (approved):

gwendolyngoetz (dismissed):

docker-compose.dev.yml
kmcnellis: Unfortunately, no.  We use the ports externally in our startup scripts to poll for the emulators being up.

docker-compose.dev.yml
kmcnellis: I really like it the pgadmin!  Its just in here to make things easier to map together with the db volume.

kmcnellis:
Sorry, need to re-get reviews, since I rebased to master!

bin/dev-dump-program-config
```diff
@@ -0,0 +1,31 @@
...
+    -h db \
+    -U postgres \
+    -t programs \
+    -t questions \
+    -t versions \
+    -t versions_programs \
+    -t versions_questions \
+    postgres -f /project/${DUMPFILE}"
+
+${COMPOSE_CMD} \
```
goldblatt, line 22: just a note that this i think is supposed to be $() in a subshell and all in quotes according to the style guide? as far as i can tell but it's not super clear on putting commands in variables

goldblatt (approved):

bion (approved):

bin/dev-dump-program-config
kmcnellis: Not in this case - We're saving a command and some of it's arguments to a variable, so quoting it here will prevent them from splitting into different values and correctly getting run.

Issue #1938, opened 2022-02-14
Unable to exercise auth code in dev


cassandralfernandes:
We use a containerized [node app](https://github.com/panva/node-oidc-provider) to simulate OIDC auth services in dev and test, but it only actually works for test mode. This means that it's not possible to exercise the real auth code in the local dev environment.

We should

1. fix the integration with the fake auth service in dev mode
2. create a means of connecting to a real auth backend in dev mode to exercise the real integration

shanemc-goog:
I have a local working config for OIDC.
Need to find a solution for SAML.

shanemc-goog:
Closing as the OIDC need is resolved.

PR #2314, opened 2022-04-20
Add ability to specify the port the test OIDC provider runs on

shanemc-goog:
### Description
Conditionally uses a supplied port instead of the default 3380.  This is necessary to address #1938 so that the local server can specify a different port than the unit tests.

### DIFF
--- a/test-support/test_oidc_provider.js
+++ b/test-support/test_oidc_provider.js
@@ -43,8 +43,5 @@ process.on('SIGINT', () => {
 });
 
 const server = oidc.listen(oidcPort, () => {
-  console.log(
-    'oidc-provider listening on port %d, check http://localhost:%d/.well-known/openid-configuration',
-    oidcPort,
-    oidcPort);
+  console.log(`oidc-provider listening on port ${oidcPort}, check http://localhost:${oidcPort}/.well-known/openid-configuration`);
 });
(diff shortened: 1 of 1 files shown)


test-support/test_oidc_provider.js
```diff
@@ -34,14 +34,17 @@ const configuration = {
   }
 };
 
-const oidc = new Provider('http://localhost:3380', configuration);
+const oidcPort = process.env.OIDC_PORT ? process.env.OIDC_PORT : 3380;
```
bion, line 37:
```suggestion
const oidcPort = process.env.OIDC_PORT || 3380;
```

test-support/test_oidc_provider.js
```diff
@@ -34,14 +34,17 @@ const configuration = {
...
   process.exit(0)
 });
 
-const server = oidc.listen(3380, () => {
-  console.log('oidc-provider listening on port 3380, check http://localhost:3380/.well-known/openid-configuration');
+const server = oidc.listen(oidcPort, () => {
+  console.log(
+    'oidc-provider listening on port %d, check http://localhost:%d/.well-known/openid-configuration',
+    oidcPort,
+    oidcPort);
```
bion, line 49:
```suggestion
  console.log(`oidc-provider listening on port ${oidcPort}, check http://localhost:${oidcPort}/.well-known/openid-configuration`);
```

bion (approved):
couple JS style things but looks good!

Issue #2315, opened 2022-04-20
Java CVEs related to authentication


rasmi:
Recent Java versions have a cryptography vulnerability allowing attackers to forge certain auth credentials/tokens, including OIDC and SAML. See details [here](https://neilmadden.blog/2022/04/19/psychic-signatures-in-java/) and [here](https://www.oracle.com/security-alerts/cpuapr2022.html). 

CiviForm [relies on OpenJDK 11](https://github.com/seattle-uat/civiform/blob/06d88b0b29b86f9c47ca469f25d7369136eb0f36/Dockerfile#L7), which is [not affected](https://openjdk.java.net/groups/vulnerability/advisories/2022-04-19) by this specific auth issue, but is affected by other CVEs. We should upgrade OpenJDK to a version >11.0.14 to resolve those issues, and notify users to upgrade their CiviForm installations accordingly.

shanemc-goog:
@gwendolyngoetz is oncall and addressing this afaik

PR #2316, opened 2022-04-20
Update OpenJDK to 11.0.14.1

rasmi:
See Issue #2315 for context. This only updates  the x86/amd64 version, and does not bump the arm64 version, since it is [not yet available](https://hub.docker.com/r/bellsoft/liberica-openjdk-alpine/tags). arm64 is not urgent (to my knowledge) since it is only used for local Mac M1 development.

Fixes (in part) issue #2315.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG PLATFORM="amd64"
 # workaround uses an aarch64 (arm64) image instead when an optional platform argument is set to arm64.
 # Docker's BuildKit skips unused stages so the image for the platform that isn't used will not be built.
 
-FROM adoptopenjdk/openjdk11:jdk-11.0.11_9-alpine-slim as amd64
+FROM adoptopenjdk/openjdk11:jdk-11.0.14.1_1-alpine-slim as amd64
 FROM bellsoft/liberica-openjdk-alpine:11.0.11-9-aarch64 as arm64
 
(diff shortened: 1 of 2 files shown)


bion:
Upgrading to 11.0.14.x hits a compiler error for Java. It looks like it's probably related to a [bugfix in OpenJDK](https://bugs.openjdk.java.net/browse/JDK-8274727) that detects "." in names in zip files and throws an exception in order to avoid an infinite loop. Unfortunately this change appears to break our build.

It doesn't appear this is broadly an issue with Play Framework or SBT -- there are no github issues related to this error in either project, and OpenJDK 11.0.14 has been out since January.

I tried building the [hello world Play Framework demo](https://github.com/playframework/play-samples/tree/2.8.x/play-java-hello-world-tutorial) with our Dockerfile and it worked both with our current version of OpenJDK as well as 11.0.14.1, which further suggests there's something particular to how we're building civiform specifically that is producing this problem.

bion (changes requested):
We need to fix the compiler error before merging

bion:
The issue appears to be our use of webjars, removing these lines fixes the problem: https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/build.sbt#L142-L144

Will need to figure out a different way to load the azure storage JS library.

Issue #2237, opened 2022-03-31
Add creation/modification timestamps to Questions and Programs


shanemc-goog:
https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/models/Program.java
https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/models/Question.java

Neither have a timestamp associated with them, which means we have no record of when they were made or last updated for when they're in a draft revision state.

This is generally useful for debugging, creation times can be set by the Database automatically too.

Additionally, it's likely that we'll need some form of data to detect when a client is working with an out dated representation of  the question or program and a "last modified" timestamp seems as good as anything.

goldblatt:
is this completed? this isn't super urgent, but if it's useful and easy go for it?

azizova-leyla:
It is not completed for Questions. Because we have many types of questions and use inheritance for them it's a bit more work than it was for programs.

shanemc-goog:
@clouser2 you did Questions right?

PR #2317, opened 2022-04-20
Adding create and last modified times to Program

### LINKED ISSUE #1657: Program 'last updated' times are statically hardcoded
**Describe the bug**
The "Last updated" time in the program list shows the same value for all programs in staging.  As well it shows a possibly incorrect time.  I just made a new program but the time is "2 hours ago"
Screen Shot 2021-10-27 at 1 12 36 PM

azizova-leyla:
### Description
Add create and last modified timestamps for programs.

Screen Shot 2022-04-21 at 2 45 59 PM

### Checklist

### Issue(s)
Fixes #1657; Does #2237 for Programs part.


shanemc-goog:
looking!

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -69,6 +71,30 @@ public static Builder builder() {
   /** The list of {@link ExportDefinition}s that make up the program. */
   public abstract ImmutableList<ExportDefinition> exportDefinitions();
 
+  /** When was this program created. Could be null for older programs. */
+  public abstract @Nullable Instant lastModifiedTime();
```
shanemc-goog, line 75: We want to prefer Optional over null as it makes the need to check for "is missing" clear and obvious.

We can't easily do that in the models files, or in data we serialize, but the sooner in the stack that we can start using Optional the better; like here.

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
```diff
@@ -69,6 +71,30 @@ public static Builder builder() {
...
+
+  /** When was this program last modified. Could be null for older programs. */
+  public abstract @Nullable Instant currentVersionCreateTime();
+
+  static final Instant DEFAULT_TIME = Instant.parse("2020-01-01T00:00:00.00Z");
```
shanemc-goog, line 80: Move this to the start of the class since it's a field and more so static.

universal-application-tool-0.0.1/test/services/program/ProgramServiceImplTest.java
```diff
@@ -213,7 +213,8 @@ public void updateProgram_updatesProgram() throws Exception {
     ProgramDefinition found = ps.getProgramDefinition(updatedProgram.id());
 
     assertThat(ps.getActiveAndDraftPrograms().getDraftSize()).isEqualTo(1);
-    assertThat(found).isEqualTo(updatedProgram);
+    assertThat(found.adminName()).isEqualTo(updatedProgram.adminName());
+    assertThat(found.lastModifiedTime()).isNotEqualTo(originalProgram.lastModifiedTime());
```
shanemc-goog, line 217: suggestion: check that the new time is after the previous one for good measure.

shanemc-goog (changes requested):

universal-application-tool-0.0.1/app/services/program/ProgramDefinition.java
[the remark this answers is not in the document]
azizova-leyla: removed this altogether in favor of optional and checking for presence.

shanemc-goog (approved):

Issue #1944, opened 2022-02-14
Validation error messages for applicants are mixed between server and client-side rendering


cassandralfernandes:
For some question types, validation error messages are rendered on the server, for others they are rendered on the client-side using JavaScript. This causes a bug where two questions on the same [screen/block](https://screen.corp.google.com/block) with validations handled in different places can cause:

1. applicant attempts to submit question answers
2. applicant sees error messages for client-side rendered questions
3. applicant corrects their answers as prompted
4. applicant attempts to submit again
5. applicant sees a new set of error messages they haven't see yet
6. The solution is to move all error message presentation into the server.

sussman:
Consensus from Bion/Susan/Ben:  the best solution for now is to go through every question-type, remove any client-side validation discovered, and then reimplement it in the server.  (Someday there may be an effort to rewrite the whole client-side architecture, but that's not happening soon.)

clouser2:
I did some investigation and took a few notes.

**Background / Motivation**

Most of the client-side validation occurs in [validation.ts](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/assets/javascripts/validation.ts), which is a script that is [included in applicant forms](https://github.com/seattle-uat/civiform/blob/4e3f04ed954eaf0638b89a1eb8b14b2388748e74/universal-application-tool-0.0.1/app/views/applicant/ApplicantProgramBlockEditView.java#L92). This replicates a fair amount of the logic already implemented server-side.

Example validation for "Name" question types:
*   [Client-side](https://github.com/seattle-uat/civiform/blob/40a7ad3b942a58cbd477a7e9646160ff3ec970e1/universal-application-tool-0.0.1/app/assets/javascripts/validation.ts#L495-L506)
*   [Server-side](https://github.com/seattle-uat/civiform/blob/40a7ad3b942a58cbd477a7e9646160ff3ec970e1/universal-application-tool-0.0.1/app/services/applicant/question/NameQuestion.java#L55-L56)

[...]

PR #2318, opened 2022-04-20
ApplicationQuestionRenderer refactoring to make subclasses responsible for providing a method to render form content

clouser2:
### Description
Inverting ApplicantQuestionRenderer pattern from having implementations call renderInternal, which creates the wrapping div to instead have them implement renderTag.

Why? Simpler to ensure all input-providing question types are wrapped in the same div.

ApplicantQuestionRenderer was converted to an interface since static questions don't render with any input values.

For consistency, also switched error validation for a few question types from `getQuestionErrors` to `getAllTypeSpecificErrors`. The former errors are shown above any specific field, while the latter are shown for specific fields.

Noticed this as part of #1944, since server-side rendering of errors will likely be updated for various question types.

### Checklist

### DIFF
--- a/universal-application-tool-0.0.1/app/views/questiontypes/ApplicantQuestionRendererImpl.java
+++ b/universal-application-tool-0.0.1/app/views/questiontypes/ApplicantQuestionRendererImpl.java
@@ -3,10 +3,12 @@ package views.questiontypes;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static j2html.TagCreator.div;
 
+import com.google.common.collect.ImmutableSet;
 import j2html.tags.ContainerTag;
 import j2html.tags.Tag;
 import play.i18n.Messages;
 import services.MessageKey;
+import services.applicant.ValidationErrorMessage;
 import services.applicant.question.ApplicantQuestion;
 import views.BaseHtmlView;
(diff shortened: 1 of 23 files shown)


universal-application-tool-0.0.1/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import views.components.TextFormatter;
+import views.style.ApplicantStyles;
+import views.style.ReferenceClasses;
+import views.style.Styles;
+
+/**
+ * Superclass for all applicant question renderers with input field(s) for the applicant to answer
+ * the question.
+ */
+public abstract class ApplicantQuestionRendererImpl implements ApplicantQuestionRenderer {
```
bion, line 23: suggestion (non-blocking): looks like this and the c'tor could be package-private rather than `public`?

bion (approved):

universal-application-tool-0.0.1/app/views/questiontypes/ApplicantQuestionRendererImpl.java
clouser2: Done!

PR #2348, opened 2022-04-25
Use CDN to load azure blob browser lib

bion:
We've been using https://webjars.org to load javascript libraries (just one so far) but the azure blob storage library specifically causes a Java compilation error on the latest patch version of Java 11.

This change is to unblock upgrading to Java 11.0.14 [while we figure out how to resolve the webjar problem](https://github.com/webjars/webjars/issues/1967).

Unblocks https://github.com/seattle-uat/civiform/pull/2316, https://github.com/seattle-uat/civiform/issues/2315

### DIFF
--- a/universal-application-tool-0.0.1/app/views/AzureFileUploadViewStrategy.java
+++ b/universal-application-tool-0.0.1/app/views/AzureFileUploadViewStrategy.java
@@ -105,7 +105,7 @@ public class AzureFileUploadViewStrategy extends FileUploadViewStrategy {
     return div(formTag, skipForms, buttons)
         .withId("azure-upload-form-component")
         .with(
-            footer(viewUtils.makeWebJarsTag(WebJarJsPaths.AZURE_STORAGE_BLOB)),
+            footer(viewUtils.makeAzureBlobStoreScriptTag()),
             footer(viewUtils.makeLocalJsTag("azure_upload")));
   }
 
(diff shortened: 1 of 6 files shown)


universal-application-tool-0.0.1/app/views/AzureFileUploadViewStrategy.java
```diff
@@ -105,7 +105,7 @@ public Tag renderFileUploadBlockSubmitForms(
     return div(formTag, skipForms, buttons)
         .withId("azure-upload-form-component")
         .with(
-            footer(viewUtils.makeWebJarsTag(WebJarJsPaths.AZURE_STORAGE_BLOB)),
```
clouser2, line 108: Looks good. Is it worth creating a separate issue tracking reverting to this path once the Azure issue is fixed upstream? We could then add a TODO against it here.

clouser2 (approved):
Nice!

universal-application-tool-0.0.1/app/views/AzureFileUploadViewStrategy.java
bion: Done.

bion:
oof, the browser test that's failing in CI is passing when i run it on my local machine. will look into it tomorrow.

clouser2 (approved):

PR #2365, opened 2022-04-26
Switch from distinct getQuestionErrors/getAllTypeSpecificErrors to a single getValidationErrors

clouser2:
### Description
Why? Hopefully more of a generalized pattern for field validation that avoids having two distinct validation methods with overlapping concerns. Instead, a map of errors associated with each question field is returned.

This should be purely mechanical and not involve any behavioral changes. In an upcoming PR for issue #1944, we'll remove client-side validation and the calls to `showFieldErrors` in the respective question renderers. While here, I left several TODOs that will be cleaned up as part of removing client-side validation.

### Checklist

### Issue(s)
#1944


server/app/services/applicant/question/AddressQuestion.java
```diff
@@ -37,14 +38,24 @@ protected ImmutableSet<QuestionType> validQuestionTypes() {
   }
 
   @Override
-  public ImmutableSet<ValidationErrorMessage> getQuestionErrors() {
-    if (!isAnswered()) {
-      return ImmutableSet.of();
-    }
+  protected ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> getValidationErrorsInternal() {
+    return ImmutableMap.<Path, ImmutableSet<ValidationErrorMessage>>builder()
+        .put(applicantQuestion.getContextualizedPath(), getQuestionErrors())
```
clouser2, line 43: Note: Here we unconditionally put the result of getQuestionErrors, which may be empty. This has the effect of returning a Map that has several keys and no entries. There are presently only two callers of getValidationErrors which account for this. Additionally, in QuestionImpl.java, the implementation of getValidationErrors filters out keys without any validation errors.

server/app/services/applicant/question/AddressQuestion.java
```diff
@@ -59,24 +70,13 @@ public ImmutableSet<ValidationErrorMessage> getQuestionErrors() {
...
-  }
-
-  public ImmutableSet<ValidationErrorMessage> getAddressErrors() {
+  private ImmutableSet<ValidationErrorMessage> validateAddress() {
     // TODO: Implement address validation.
     return ImmutableSet.of();
   }
 
-  public ImmutableSet<ValidationErrorMessage> getStreetErrors() {
-    if (isAnswered() && getStreetValue().isEmpty()) {
```
clouser2, line 79: Here and in other QuestionImpl subclasses, I removed the call to `isAnswered` for each validator since we now call that in the base class prior to calling `getValidationErrorsInternal`.

server/app/services/applicant/question/AddressQuestion.java
bion: suggestion (blocking): move this note into a comment in QuestionImpl.java

bion (approved):

server/app/services/applicant/question/AddressQuestion.java
clouser2: Done!

bion (approved):

jeanniefu (commented):
I like this new map based approach!

clouser2 (commented):
Thanks much for the reviews. @jeanniefu , please take a look and let me know your thoughts.

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -39,7 +46,10 @@ protected Tag renderTag(ApplicantQuestionRendererParams params) {
                         messages.at(MessageKey.ADDRESS_PLACEHOLDER_STREET.getKeyName()))
                     .setValue(addressQuestion.getStreetValue().orElse(""))
                     .setFieldErrors(messages, addressQuestion.getStreetErrorMessage())
```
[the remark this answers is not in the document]
jeanniefu: Ah gotcha, that makes sense. Thanks for the explanation!

server/app/services/applicant/question/AddressQuestion.java
jeanniefu: Thanks for the detailed thoughts! Your approach sounds good to me, since ultimately the public function will return correct data either way. And I also think that this approach is less error-prone, since the base class handles all the filtering.

Removing QuestionImpl also sounds good to me!

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -49,6 +59,10 @@ protected Tag renderTag(ApplicantQuestionRendererParams params) {
                     .setPlaceholderText(
                         messages.at(MessageKey.ADDRESS_PLACEHOLDER_LINE_2.getKeyName()))
                     .setValue(addressQuestion.getLine2Value().orElse(""))
+                    .setFieldErrors(
```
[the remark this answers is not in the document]
jeanniefu: Sounds good, thanks for the details!

server/app/services/applicant/question/QuestionImpl.java
```diff
@@ -49,7 +49,11 @@ public final ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> getValidat
         .build();
   }
 
-  /** Question-type specific implementation of {@link Question.getValidationErrors}. */
+  /**
+   * Question-type specific implementation of {@link Question.getValidationErrors}. Note that keys
+   * with an empty set of errors will be filtered out by {@link Question.getValidationErrors} so
+   * that calls to isEmpty on the result are sufficient to indicate if there any errors.
```
jeanniefu, line 55: nit: replace "the result" with "getValidationErrors result" to be super clear

server/app/services/applicant/question/QuestionImpl.java
```diff
@@ -30,6 +34,25 @@ public QuestionImpl(ApplicantQuestion applicantQuestion) {
...
+    // For ease of implementation, subclasses may build the error list by putting a field key
+    // in the map along with a call to a validator method that may return an empty set of errors.
+    // We remove keys with an empty set of errors here to help defend against downstream consumers
+    // assumes that calling isEmpty on the map means that there are no errors.
+    return ImmutableMap.<Path, ImmutableSet<ValidationErrorMessage>>builder()
+        .putAll(Maps.filterEntries(getValidationErrorsInternal(), e -> !e.getValue().isEmpty()))
+        .build();
+  }
+
+  /** Question-type specific implementation of {@link Question.getValidationErrors}. */
```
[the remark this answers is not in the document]
jeanniefu: Sounds good, thanks for updating the comment!

jeanniefu (approved):

jeanniefu (approved):

bion (approved):

PR #2370, opened 2022-04-27
Updates to make VSCode work by default with Metals.

clouser2:
### Description
**bin/vscode-setup**
*   Updated the Maven properties indicating that this is Java 11. Fixes a host of problems reported by VSCode around our use of lambdas and other language features. If unspecified, this defaults to Java 5.

**civiform.code-workspace** 
*   Inverted the order of project definitions. Prior to this, Metals didn't detect a SBT and didn't automatically compile.
*   Added .bloop / target files to list of exemptions for searching. This prevents these files from showing in the quick file open candidate list.

**server/build.sbt**
*   Marked auto-value-parent as [pomOnly](https://stackoverflow.com/q/45608473) since it doesn't have a corresponding JAR. Without this, the call to makePom from vscode-setup adds this as a dependency and VSCode fails to resolve it, preventing any further indexing from occurring.

With these changes, the remaining problems reported by VSCode are largely around AutoValue and auto-generated Scala routes. We'll hopefully address these separately.

### Tested
*   Created a fresh clone of the repository and cleared all home directory Maven / Ivy caches. 
*   Ran `bin/vscode-setup`
*   Opened `civiform.code-workspace` from VSCode and waited for build / index to complete
*   Observed auto-complete and jump to definition working as intended  
*   Ran `bin/run-dev` and confirmed I could make requests
*   Confirmed `bin/build-and-push` successfully built an image

### DIFF
--- a/bin/vscode-setup
+++ b/bin/vscode-setup
@@ -25,7 +25,7 @@ rm -rf server/temp
 
 BUILD=$(
          cat <<END
-<properties>
+    <properties>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
     </properties>
@@ -47,4 +47,4 @@ BUILD=$(
 END
 )
 
-perl -i -pe "s|<modelVersion>|${BUILD}|g" server/pom.xml
+perl -i -pe "s|\s+<modelVersion>|${BUILD}|g" server/pom.xml
--- a/bin/vscode-setup
+++ b/bin/vscode-setup
@@ -25,7 +25,11 @@ rm -rf server/temp
 
 BUILD=$(
          cat <<END
-<build>
+<properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+    <build>
         <sourceDirectory>app</sourceDirectory>
         <testSourceDirectory>test</testSourceDirectory>
         <resources>
--- a/server/build.sbt
+++ b/server/build.sbt
@@ -67,7 +67,7 @@ lazy val root = (project in file("."))
       // Autovalue
       "com.google.auto.value" % "auto-value-annotations" % "1.9",
       "com.google.auto.value" % "auto-value" % "1.9",
-      "com.google.auto.value" % "auto-value-parent" % "1.9",
+      "com.google.auto.value" % "auto-value-parent" % "1.9" pomOnly(),
 
       // Errorprone
       "com.google.errorprone" % "error_prone_core" % "2.13.1",
--- a/server/build.sbt
+++ b/server/build.sbt
@@ -67,7 +67,7 @@ lazy val root = (project in file("."))
       // Autovalue
       "com.google.auto.value" % "auto-value-annotations" % "1.9",
       "com.google.auto.value" % "auto-value" % "1.9",
-      "com.google.auto.value" % "auto-value-parent" % "1.9",
+      //"com.google.auto.value" % "auto-value-parent" % "1.9" pomOnly(),
 
       // Errorprone
       "com.google.errorprone" % "error_prone_core" % "2.13.1",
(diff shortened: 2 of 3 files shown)


server/build.sbt
```diff
@@ -67,7 +67,7 @@ lazy val root = (project in file("."))
       // Autovalue
       "com.google.auto.value" % "auto-value-annotations" % "1.9",
       "com.google.auto.value" % "auto-value" % "1.9",
-      "com.google.auto.value" % "auto-value-parent" % "1.9",
+      "com.google.auto.value" % "auto-value-parent" % "1.9" pomOnly(),
```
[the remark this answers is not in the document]
kmcnellis: And alternate approach would be to use the script to modify this line in the generated pom

bin/vscode-setup
```diff
@@ -25,7 +25,11 @@ rm -rf server/temp
 
 BUILD=$(
          cat <<END
-<build>
+<properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+    <build>
```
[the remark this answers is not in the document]
clouser2: Done. This was originally dedented since the perl replacement below didn't account for leading whitespace. I've updated that replacement so that it also captures leading whitespace for the line.

server/build.sbt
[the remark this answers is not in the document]
clouser2: Agreed that we could modify the generated pom. However, I'd prefer to do as little hand-editing as we can. The `auto-value-parent` package doesn't ever have a JAR, so this change should be a no-op. That said, good call on manually testing.

I confirmed that `bin/run-dev` continues to load. Additionally, I ran the docker build from the [build and push script](https://github.com/seattle-uat/civiform/blob/cc9628a5f398c5d19720f8acdb06bdadf90ab036/bin/build-and-push#L5-L17). I didn't think it necessary to push an artifact to DockerHub since the build itself is what we're verifying didn't change. Finally, passing browser / unit test should hopefully bolster confidence. Updated the "Tested" section of the PR to reflect this.

.github/workflows/scorecards.yml
```diff
@@ -22,12 +22,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # v3.0.0
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
```
kmcnellis, line 25: I don't think this is related

README.md
```diff
@@ -1,4 +1,5 @@
 ![CI status](https://github.com/seattle-uat/civiform/actions/workflows/push.yaml/badge.svg)
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/6008/badge)](https://bestpractices.coreinfrastructure.org/projects/6008)
```
kmcnellis, line 2: This is probably a separate change

browser-test/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "devDependencies": {
     "@types/jest": "27.4.1",
-    "@types/node": "16.11.27",
+    "@types/node": "16.11.31",
```
kmcnellis, line 11: Unrelated

server/build.sbt
```diff
@@ -22,7 +22,7 @@ lazy val root = (project in file("."))
       "com.j2html" % "j2html" % "1.4.0",
 
       // Amazon AWS SDK
-      "software.amazon.awssdk" % "aws-sdk-java" % "2.17.161",
+      "software.amazon.awssdk" % "aws-sdk-java" % "2.17.179",
```
kmcnellis, line 25: Unrelated

server/build.sbt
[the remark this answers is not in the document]
kmcnellis: Awesome!

server/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "license": "CC0-1.0",
   "devDependencies": {
-    "@tailwindcss/line-clamp": "0.3.1",
+    "@tailwindcss/line-clamp": "0.4.0",
```
kmcnellis, line 13: Unrelated

kmcnellis (approved):

kmcnellis (approved):

gwendolyngoetz (approved):

PR #2372, opened 2022-04-27
Persistence layer for API keys

bion:
### Description
- [x] create a new DB table for api keys
- [x] add an ebean model for interacting with the table
- [x] add a repository class for async DB queries with api keys
- [x] implement ApiKeyGrants class, serialize to jsonb and deserialize on load

### Checklist

### DIFF
--- a/server/app/auth/ApiKeyGrants.java
+++ b/server/app/auth/ApiKeyGrants.java
@@ -1,0 +1,67 @@
+package auth;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+
+/**
+ * Stores permissions for {@link models.ApiKey}s.
+ *
+ * <p>Permissions are represented as a resource identifier paired with an ability. E.g.
+ * "utility-discount" program with "read" ability. Multiple permissions may be stored for the same
+ * resource. E.g. having both the "read" and "write" permission for a given resource.
+ */
+public class ApiKeyGrants {
+
+  /** Enumerates the abilities an ApiKey may have with respect to a resource */
+  public enum Permission {
+    WRITE,
+    READ;
+  }
+
+  /** Create a new instance with no grants. */
+  public ApiKeyGrants() {
+    this.programGrants = HashMultimap.create();
+  }
+
+  /**
+   * Used by EBean to deserialize an instance of ApiKeyGrants stored in the database.
+   */
+  @JsonCreator
+  public ApiKeyGrants(@JsonProperty("programGrants") Multimap<String, Permission> programGrants) {
+    this.programGrants = checkNotNull(programGrants);
+  }
+
+  @JsonProperty("programGrants")
+  private Multimap<String, Permission> programGrants;
+
+  /**
+   * Grant the ability to do {@code permission} for the program identified by {@code programSlug}.
+   */
+  public void grantProgramPermission(String programSlug, Permission permission) {
+    programGrants.put(programSlug, permission);
+  }
+
+  /**
+   * Revoke the ability to do {@code permission} for the program identified by {@code programSlug}.
+   */
+  public void revokeProgramPermission(String programSlug, Permission permission) {
+    programGrants.remove(programSlug, permission);
+  }
+
+  /**
+   * Check for the ability to do {@code permission} for the program identified by {@code
+   * programSLug}.
+   */
+  public boolean hasProgramPermission(String programSlug, Permission permission) {
+    return programGrants.get(programSlug).contains(permission);
+  }
+
+  /** Revoke all permissions for all programs. */
+  public void revokeAllProgramPermissions() {
+    this.programGrants = HashMultimap.create();
+  }
+}
--- a/server/app/auth/ApiKeyGrants.java
+++ b/server/app/auth/ApiKeyGrants.java
@@ -22,6 +22,9 @@ public class ApiKeyGrants {
     READ;
   }
 
+  @JsonProperty("programGrants")
+  private Multimap<String, Permission> programGrants;
+
   /** Create a new instance with no grants. */
   public ApiKeyGrants() {
     this.programGrants = HashMultimap.create();
@@ -32,9 +35,6 @@ public class ApiKeyGrants {
   public ApiKeyGrants(@JsonProperty("programGrants") Multimap<String, Permission> programGrants) {
     this.programGrants = checkNotNull(programGrants);
   }
-
-  @JsonProperty("programGrants")
-  private Multimap<String, Permission> programGrants;
 
   /**
    * Grant the ability to do {@code permission} for the program identified by {@code programSlug}.
--- a/server/app/auth/ApiKeyGrants.java
+++ b/server/app/auth/ApiKeyGrants.java
@@ -27,9 +27,7 @@ public class ApiKeyGrants {
     this.programGrants = HashMultimap.create();
   }
 
-  /**
-   * Used by EBean to deserialize an instance of ApiKeyGrants stored in the database.
-   */
+  /** Used by EBean to deserialize an instance of ApiKeyGrants stored in the database. */
   @JsonCreator
   public ApiKeyGrants(@JsonProperty("programGrants") Multimap<String, Permission> programGrants) {
     this.programGrants = checkNotNull(programGrants);
--- a/server/app/auth/ApiKeyGrants.java
+++ b/server/app/auth/ApiKeyGrants.java
@@ -16,7 +16,7 @@ import com.google.common.collect.Multimap;
  */
 public class ApiKeyGrants {
 
-  /** Enumerates the abilities an ApiKey may have with respect to a resource */
+  /** Enumerates the abilities an ApiKey may have with respect to a resource. */
   public enum Permission {
     WRITE,
     READ;
--- a/server/app/models/ApiKey.java
+++ b/server/app/models/ApiKey.java
@@ -5,6 +5,7 @@ import io.ebean.annotation.DbJsonB;
 import io.ebean.annotation.WhenCreated;
 import io.ebean.annotation.WhenModified;
 import java.time.Instant;
+import java.util.Optional;
 import javax.persistence.Entity;
 import javax.persistence.Table;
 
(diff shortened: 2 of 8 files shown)


server/app/models/ApiKey.java
```diff
@@ -0,0 +1,98 @@
...
+  public ApiKeyGrants getGrants() {
+    return grants;
+  }
+
+  public ApiKey setGrants(ApiKeyGrants grants) {
+    this.grants = grants;
+    return this;
+  }
+
+  @DbJsonB private ApiKeyGrants grants;
```
shanemc-goog, line 35: move up with the other members.

server/app/repository/ApiKeyRepository.java
```diff
@@ -0,0 +1,54 @@
...
+  private final Database database;
+  private final DatabaseExecutionContext executionContext;
+
+  @Inject
+  public ApiKeyRepository(DatabaseExecutionContext executionContext) {
+    this.database = DB.getDefault();
+    this.executionContext = checkNotNull(executionContext);
+  }
+
+  /** Insert a new ApiKey record using a thread from {@link DatabaseExecutionContext}. */
```
shanemc-goog, line 24: Suggestion: Replace DatabaseExecutionContext with more general "async" language.  The public API here doesn't really discuss the threading, if anything add DatabaseExecutionContext to the class comment as it underlies the general operation.

server/test/auth/ApiKeyGrantsTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class ApiKeyGrantsTest {
+
+  private static final String PROGRAM_1_SLUG = "program-1";
+  private static final String PROGRAM_2_SLUG = "program-2";
+  private ApiKeyGrants subject;
```
shanemc-goog, line 12: question: where is the use of "subject" in our tests from? I haven't seen it as a generic identifier in this context and it was somewhat unreadable to me in other test code.  I've only seen cut (Code under Test).

server/test/auth/ApiKeyGrantsTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Before
+  public void setUp() {
+    subject = new ApiKeyGrants();
+  }
+
+  @Test
```
shanemc-goog, line 19: Add the following tests
- No setup, permissions aren't allowed (IE an unknown Slug isn't allowed)
- Multiple SLUGS (One read, one write perhaps) have appropriate permissions.

shanemc-goog (changes requested):

server/app/auth/ApiKeyGrants.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>Permissions are represented as a resource identifier paired with an ability. E.g.
+ * "utility-discount" program with "read" ability. Multiple permissions may be stored for the same
+ * resource. E.g. having both the "read" and "write" permission for a given resource.
+ */
+public class ApiKeyGrants {
+
+  /** Enumerates the abilities an ApiKey may have with respect to a resource */
+  public enum Permission {
+    WRITE,
```
gwendolyngoetz, line 21: Should we remove WRITE for now since we aren't supporting it in v1?

server/conf/evolutions/default/38.sql
```diff
@@ -0,0 +1,27 @@
+ --- Create the api_keys table.
+
+# --- !Ups
+
+CREATE TABLE IF NOT EXISTS api_keys (
+  id BIGSERIAL PRIMARY KEY NOT NULL,
+  name VARCHAR NOT NULL,
+  create_time TIMESTAMP NOT NULL,
+  update_time TIMESTAMP NOT NULL,
+  created_by VARCHAR NOT NULL,
```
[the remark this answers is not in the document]
bion: I probably should have addressed this in the design doc.

I was planning to store authority ID. Storing email seemed like an anti-pattern given our auth protocol says not to use it for identity. Storing the Account.id gives us almost nothing to go on if the Account gets removed for some reason.

Thoughts?

server/conf/evolutions/default/38.sql
```diff
@@ -0,0 +1,27 @@
...
+  retired_time TIMESTAMP,
+  retired_by VARCHAR,
+  key_id VARCHAR UNIQUE NOT NULL,
+  salted_key_secret VARCHAR UNIQUE NOT NULL,
+  subnet VARCHAR NOT NULL,
```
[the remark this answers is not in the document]
bion: Yes, after reading a little more I realized "subnet mask" specifically refers to the notation that doesn't include the `/N` e.g. `/32` suffix. The latter format is called CIDR notation, and is the one actually specified by the design.

I'll document details about this column in the model file.

server/app/models/ApiKey.java
bion: Done

server/app/auth/ApiKeyGrants.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /** Used by EBean to deserialize an instance of ApiKeyGrants stored in the database. */
+  @JsonCreator
+  public ApiKeyGrants(@JsonProperty("programGrants") Multimap<String, Permission> programGrants) {
+    this.programGrants = checkNotNull(programGrants);
+  }
+
+  @JsonProperty("programGrants")
+  private Multimap<String, Permission> programGrants;
```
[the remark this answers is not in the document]
bion: Done

server/test/auth/ApiKeyGrantsTest.java
bion: changed to `apiKeyGrants`.

unit tests where the object under test is referred to as `subject` is an old habit, I agree it's not a very clear naming convention.

server/test/repository/ApiKeyRepositoryTest.java
```diff
@@ -0,0 +1,67 @@
...
+  @Test
+  public void insert_persistsANewKey() {
+    ApiKey foundKey;
+    ApiKey apiKey = new ApiKey();
+    apiKey.setName("key name");
+    apiKey.setKeyId("key-id");
+    apiKey.setCreatedBy("test@example.com");
+    apiKey.setSaltedKeySecret("secret");
+    apiKey.setSubnet("0.0.0.0/32");
+    apiKey.setExpiration(Instant.ofEpochSecond(100));
```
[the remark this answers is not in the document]
bion: done

server/app/auth/ApiKeyGrants.java
bion: I wanted to have more than one permission type so that the unit test can assert correct behavior for handling it.

server/app/models/ApiKey.java
```diff
@@ -0,0 +1,98 @@
...
+    this.saltedKeySecret = saltedKeySecret;
+    return this;
+  }
+
+  public String getSubnet() {
+    return subnet;
+  }
+
+  public ApiKey setSubnet(String subnet) {
+    this.subnet = subnet;
```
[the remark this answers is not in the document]
bion: > Suggestion: Add validation to make sure the subnet string is in the correct CIDR format and valid IP.

Generally speaking, validation should not be done in the model, and should be treated as a separate concern from persistence, which is the intended scope of this PR (though I did go beyond that with doing the full ApiKeyGrants implementation). CiviForm follows this guideline for the most part. One could argue that database invariants (e.g. `NOT NULL`) are a form of validation, which is true, but those are also not a concern of the model and conforming with them should be handled along with other validations in validation-specific code.

> Are we going to handle IPv4 and IPv6?

The current plan is to support IPv4, as per the design. Extending it to support v6 is doable without the need for rewriting much code I hope, but it's not currently in scope unless we have programs looking to onboard that will be identifying their API clients via IPv6.

server/test/auth/ApiKeyGrantsTest.java
bion: > No setup, permissions aren't allowed (IE an unknown Slug isn't allowed)

done

> Multiple SLUGS (One read, one write perhaps) have appropriate permissions. (IE we don't mix up permissions.)

ah sharp eye, done.

server/app/models/ApiKey.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public ApiKey setGrants(ApiKeyGrants grants) {
+    this.grants = grants;
+    return this;
+  }
+
+  @DbJsonB private ApiKeyGrants grants;
+
+  public Instant getCreateTime() {
```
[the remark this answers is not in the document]
bion: done

server/app/models/ApiKey.java
```diff
@@ -0,0 +1,98 @@
...
+  public String getSaltedKeySecret() {
+    return saltedKeySecret;
+  }
+
+  public ApiKey setSaltedKeySecret(String saltedKeySecret) {
+    this.saltedKeySecret = saltedKeySecret;
+    return this;
+  }
+
+  public String getSubnet() {
```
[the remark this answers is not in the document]
bion: done

server/app/repository/ApiKeyRepository.java
bion: done

gwendolyngoetz (approved):

server/conf/evolutions/default/38.sql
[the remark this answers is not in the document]
shanemc-goog: authority_id for a bare token is good.

Curious: when would we remove an Account especially an Admin one that we'd store here.   I was wondering if this should be relational to Account, no real opinions though just a curiosity.

server/app/models/ApiKey.java
```diff
@@ -32,67 +36,107 @@ public ApiKey setGrants(ApiKeyGrants grants) {
...
   public ApiKey setSubnet(String subnet) {
     this.subnet = subnet;
     return this;
   }
+
+  /**
+   * The client IPv4 address of the last request to successfully auth with the ApiKey. Empty if the
+   * ApiKey has never been used.
+   */
+  public Optional<String> getLastCallIpAddress() {
```
shanemc-goog, line 133: Is this meant to be material to the API or more like light logging for dev testing/curiosity?

shanemc-goog (approved):

server/app/models/ApiKey.java
bion: The latter, also useful for light security / auditing purposes.

server/conf/evolutions/default/38.sql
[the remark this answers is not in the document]
bion: At the moment we don't have any code that would remove an account, but I could see implementing account deletion in the future for compliance purposes -- even admin accounts.

I don't think we need the ORM-level association to the account yet if ever.

gwendolyngoetz (approved):

PR #2373, opened 2022-04-28
Don't run PR checks when they're not relevant

bion:
Attempts to limit the PR checks to when they're relevant to the changes in the PR. E.g. changing cloud deployment code doesn't need the browser tests to run.


shanemc-goog (approved):
woohoo!!!

shanemc-goog:
Looks like there's a syntax error, from the test run logs

The workflow is not valid. .github/workflows/pr.yaml (Line: 12, Col: 9): Unrecognized named-value: 'steps'. Located at position 17 within expression: contains(toJSON(steps.file_changes.outputs.files), 'server/') .github/workflows/pr.yaml (Line: 27, Col: 9): Unrecognized named-value: 'steps'. Located at position 17 within expression: contains(toJSON(steps.file_changes.outputs.files), 'browser-test/') || contains(toJSON(steps.file_changes.outputs.files), 'server/')

azizova-leyla (approved):

.github/workflows/pr.yaml
```diff
@@ -20,8 +21,10 @@ jobs:
         run: docker build -t civiform-dev --cache-from docker.io/civiform/civiform-dev:latest ./
       - name: Run tests
         run: USE_LOCAL_CIVIFORM=true bin/run-test-ci
+
   run_browser_tests_aws:
     runs-on: ubuntu-latest
+    if: contains(toJSON(steps.file_changes.outputs.files), 'browser-test/') || contains(toJSON(steps.file_changes.outputs.files), 'server/')
```
kmcnellis, line 27: These should also run when the bin scripts that they run change, or when the docker files change

kmcnellis (changes requested):

.github/workflows/pr.yaml
bion: Done

azizova-leyla (dismissed):

shanemc-goog (approved):

.github/workflows/pr_infra.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - 'infra/**'
+
+permissions: read-all
+
+jobs:
+  cloudformation:
+    runs-on: ubuntu-18.04 # for older version of awscli.
+    steps:
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.4
```
kmcnellis, line 17: This one already doesn't run if you don't change something in infra/ .  (And I don't think it works at all if you do change something in infra/)

.github/workflows/pr_server.yaml
```diff
@@ -0,0 +1,100 @@
+name: pr_server
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - 'server/**'
```
kmcnellis, line 8: And this directory (.github/workflows) so they run if you just change the test

kmcnellis (changes requested):
image
Not sure why there are so many tests listed here

.github/workflows/pr_server.yaml
bion: done

.github/workflows/pr_cloud.yaml
```diff
@@ -1,16 +1,15 @@
-name: pr_test
+name: pr_cloud
 
 on:
   pull_request:
-    branches: main
+    branches:
+      - 'main'
+    paths:
+      - 'cloud/**'
```
kmcnellis, line 8: These should run when tests change (or at least, when pr_cloud.yaml changes)

kmcnellis (approved):

kmcnellis (commented):
If we aren't running all the tests on each PR, maybe we can run them all on commits to main?  That way it's easier to track down where a test started failing if we do have an issue.

.github/workflows/pr_cloud.yaml
bion: done

bion:
@kmcnellis yep we're running tests on main https://github.com/seattle-uat/civiform/blob/main/.github/workflows/push.yaml#L65-L67

azizova-leyla (approved):

PR #2380, opened 2022-04-28
Persist selected export state upon question creation.

### LINKED ISSUE #1889: Export Value Defaulting to "No Export" in New Version of Question
From Brenda: When creating a "new version" of a question, it should use the previously selected export value. Right now it just defaults back to "No export" no matter what was selected.

From Gloria: I see this as a bug and the solution being to carry over the previously selected export value into the "new version," but open to discussion if others see defaulting to "no export" as a good thing.

clouser2:
### Description
Previously, we were only persisting the selected export option when updating a question. We now do this as part of creation as well.

While here, I also updated the UI to preselect the "No Export" option since that's what happens in practice.

Finally, added a browser test to catch regressions in this area.

### Checklist

### Issue(s)
Fixes #1889

### DIFF
--- a/browser-test/src/support/admin_questions.ts
+++ b/browser-test/src/support/admin_questions.ts
@@ -109,7 +109,7 @@ export class AdminQuestions {
     if (!exportOption) {
       throw new Error('A non-empty export option must be provided')
     }
-    await this.page.check(`label:has-text("${exportOption}") input`)
+    await this.page.check(this.selectorForExportOption(exportOption))
   }
 
   async updateQuestionText(updateText: string) {
--- a/browser-test/src/support/admin_questions.ts
+++ b/browser-test/src/support/admin_questions.ts
@@ -8,12 +8,17 @@ type QuestionParams = {
   questionText?: string
   helpText?: string
   enumeratorName?: string
+  exportOption?: string
 }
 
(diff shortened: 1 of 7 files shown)


browser-test/src/support/admin_questions.ts
```diff
@@ -46,6 +51,10 @@ export class AdminQuestions {
     // expect(await this.page.isDisabled(`text=${questionName}`)).toEqual(true)
   }
 
+  async expectExportState(exportOption: string) {
+    expect(await this.page.isChecked(`label:has-text("${exportOption}") input`)).toEqual(true)
```
clouser2, line 55: Not quite sure I understand what's going on here. CI appears to be failing, even though this test passes when I run `bin/run-browser-tests` or `bin/run-browser-tests-ci`. Perhaps it could be something different about the underlying browser used by CI versus my local machine.

clouser2: Moving this back to draft mode while I try and figure out why CI is failing.

clouser2: This is now ready for review. Sorry about that!

bion (approved):
looks great!

azizova-leyla (approved):

PR #2384, opened 2022-04-28
Using Terraform Resource to generate ssh key

### LINKED ISSUE #2383: Terraform Deplouy: Failing b/c it cannot find the bastion.pub file
Try to run bin/setup and see it fails with a missing ssh/bastion.pub error

goldblatt:
### Description
Terraform deploys are failing due to not being able to find a file, there's a terraform resource that lets us create the initial key so we can use that here. note this is purely to make the azure terraform code happy as we prevent all ingres to the machine and then to access the bastion you have to manually add your own key to it via the scripts

### Checklist

### Issue(s)
Fixes #2383

### DIFF
--- a/cloud/azure/bin/lib/azure.sh
+++ b/cloud/azure/bin/lib/azure.sh
@@ -175,6 +175,7 @@ function azure::get_current_user_id() {
 #   3. scope name
 #######################################
 function azure::ensure_role_assignment() {
+  echo "creating role assignment for ${2}"
   local USER_TYPE="$(az account show --query user.type -o tsv)"
   local object_id=""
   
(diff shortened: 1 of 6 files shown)


cloud/azure/bin/lib/azure.sh
```diff
@@ -175,6 +175,7 @@ function azure::get_current_user_id() {
 #   3. scope name
 #######################################
 function azure::ensure_role_assignment() {
+  echo "creating role assignment for ${2}"
```
shanemc-goog, line 178: Suggestion: capture ${2} into a named var for readability.

shanemc-goog (approved):

PR #2392, opened 2022-04-29
Export pdf Functionality

### LINKED ISSUE #2084: Export applications to a PDF
Program Administrators would like to have the option to export and save a PDF version of a program application.

It would be helpful if there was an "Export to PDF" button when Program Admins view an individual application.

**Additional context**
This is needed for auditing purposes. Program staff also find it easier to view individual applications from the "View Applications" screen than reviewing in the excel export.

Mock-up attached ("Export PDF of Application")
Export PDF of Application

swatkat1:
### Description
This changes is to enable the downlaod an application as a PDF functionality.

Added Unit test.

### Issue(s)
Fixes #2084 

Screenshots:
image

Final PDF:
image


server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -151,13 +155,39 @@ public Result downloadDemographics() {
...
+    String applicantNameWithApplicationId =
+            String.format("%s (%d)", application.getApplicantData().getApplicantName(), application.id);
+
+    String filename = String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    byte[] pdf = null;
+    try{
+      pdf = pdfExporter.export(application);
+    }
+    catch(Exception e)
+    {
```
bion, line 185: nit: need to run `bin/fmt` to fix formatting

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -151,13 +155,39 @@ public Result downloadDemographics() {
...
+    Application application = applicationMaybe.get();
+    String applicantNameWithApplicationId =
+            String.format("%s (%d)", application.getApplicantData().getApplicantName(), application.id);
+
+    String filename = String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    byte[] pdf = null;
+    try{
+      pdf = pdfExporter.export(application);
+    }
+    catch(Exception e)
```
bion, line 184: issue (blocking): Catching all exceptions with `catch (Exception e)` is pretty risky -- there are a number of reasons the action could fail and with this code in place they would all result in a 404, which would mask server errors in monitoring and alerting tools.

Please change this to only catching errors that require explicit handling. If you are unsure how to handle something going wrong on the server, most likely the best thing to do is to catch the exception and rethrow it wrapped in a `RuntimeException`, which are [unchecked](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html).

server/test/services/export/PdfExporterTest.java
```diff
@@ -2,98 +2,98 @@
...
+        assertThat(linesFromPDF.get(1)).isEqualTo("Program Name : " + programName);
+
+        List<String> linesFromStaticString  = Splitter.on("\n").splitToList(applicationOneString);
+
+        for(int i=3;i<linesFromPDF.size();i++) {
+            assertThat(linesFromPDF.get(i)).isEqualTo(linesFromStaticString.get(i));
+        }
+
+    }
+    public static String applicationOneString =
```
bion, line 52: suggestion (blocking): this should be a constant
```suggestion
    public static final String APPLICATION_ONE_STRING =
```

server/test/services/export/PdfExporterTest.java
```diff
@@ -2,98 +2,98 @@
...
-    assertThat(outputField).isNotNull();
-    assertThat(outputField.getValueAsString()).isEqualTo(APPLICANT_VALUE);
-  }
+
+
+
+public class PdfExporterTest extends AbstractExporterTest {
+
+    @Test
+    public void validatePDFExport() throws IOException, DocumentException {
```
bion, line 19: nice work testing this!

bion (changes requested):
Great work on this, it's a major piece of functionality and much needed feature! I've requested a few changes here and there but overall good stuff!

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -148,16 +154,42 @@ public Result downloadDemographics() {
...
+        String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    byte[] pdf = null;
+    try {
+      pdf = pdfExporter.export(application, applicantNameWithApplicationId);
+    } catch (DocumentException | IOException e) {
+      throw new RuntimeException(
+          "Document or IOException exception detected while processing the PDF export!");
+    }
+    return ok(pdf)
+        .as("application/pdf")
```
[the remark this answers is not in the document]
swatkat1: Just not for application/pdf ... there is one for binary, json

bin/run-browser-tests
```diff
@@ -5,7 +5,7 @@
 source bin/lib.sh
 docker::set_project_name_browser_tests
 
-docker run --ipc=host --rm -it \
+docker run --rm -it \
```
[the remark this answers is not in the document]
swatkat1: I am trying to implement a browser test for download, Once I get it running locally, will remove this tag.

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -148,16 +154,41 @@ public Result downloadDemographics() {
...
+      return notFound(String.format("Application %d does not exist.", applicationId));
+    }
+
+    Application application = applicationMaybe.get();
+    String applicantNameWithApplicationId =
+        String.format("%s (%d)", application.getApplicantData().getApplicantName(), application.id);
+
+    String filename =
+        String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    byte[] pdf = null;
```
bion, line 182: nit:

```suggestion
    byte[] pdf;
```

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -148,16 +154,41 @@ public Result downloadDemographics() {
...
+    try {
+      pdf = pdfExporter.export(application, applicantNameWithApplicationId);
+    } catch (DocumentException | IOException e) {
+      throw new RuntimeException(
+          "Document or IOException exception detected while processing the PDF export!");
```
bion, line 187: suggestion (blocking): include the original exception as the cause
```suggestion
      throw new RuntimeException(e);
```

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -1,12 +1,7 @@
 package views.admin.programs;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static j2html.TagCreator.a;
-import static j2html.TagCreator.div;
-import static j2html.TagCreator.each;
-import static j2html.TagCreator.h1;
-import static j2html.TagCreator.h2;
-import static j2html.TagCreator.p;
+import static j2html.TagCreator.*;
```
bion, line 4: issue (blocking): no wildcard import, looks like this was maybe done automatically by your editor?

bion (changes requested):

server/app/views/admin/programs/ProgramApplicationView.java
swatkat1: IDE is adding this. I have changed it back.

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -148,16 +154,40 @@ public Result downloadDemographics() {
...
+      return notFound(String.format("Application %d does not exist.", applicationId));
+    }
+
+    Application application = applicationMaybe.get();
+    String applicantNameWithApplicationId =
+        String.format("%s (%d)", application.getApplicantData().getApplicantName(), application.id);
+
+    String filename =
+        String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    byte[] pdf;
```
bion, line 182: suggestion (blocking): formatting the applicant's name and the PDF filename are not HTTP concerns so probably would be best encapsulated somewhere other than the controller. Since they're fairly specific to PDF export, the `PdfExporter` may be a good place for them.

A way to do cleaner separation of concerns here could be to have `PdfExporter.export` return an instance of a new class `InMemoryPdf` that has getters for `byte [] getBytes()`, `String getFilename()`. A convenient place to define the new class could be as an inner class of the exporter:
```java
class PdfExporter {
...
  public static class InMemoryPdf { ... }
```

This would keep the details of preparing the PDF file in the exporter and keep the controller focused on HTTP concerns.

server/app/services/export/PdfExporter.java
```diff
@@ -1,46 +1,87 @@
...
+      document.add(answer);
+      document.add(time);
     }
-    ByteArrayOutputStream inMemoryFile = new ByteArrayOutputStream();
-    baseDocument.save(inMemoryFile);
-    inMemoryFile.close();
-    writer.write(inMemoryFile.toString(StandardCharsets.UTF_8));
+    document.close();
+    writer.close();
+    byteArrayOutputStream.close();
```
bion, line 84: suggestion (blocking): if generated the PDF produces a IOException or DocumentException these two open file handles won't necessarily close. To ensure that they do you could wrap the code that can throw in a `try {` and use `finally {` to ensure these close. https://docs.oracle.com/javase/tutorial/essential/exceptions/finally.html

bion (changes requested):

server/app/services/export/PdfExporter.java
swatkat1: Good catch! Will do

server/app/services/export/PdfExporter.java
```diff
@@ -1,46 +1,123 @@
...
+        String.format("%s (%d)", application.getApplicantData().getApplicantName(), application.id);
+    String filename =
+        String.format("%s-%s.pdf", applicantNameWithApplicationId, clock.instant().toString());
+    inMemoryPdf.byteArray =
+        buildPDF(
+            answers,
+            applicantNameWithApplicationId,
+            application.getProgram().getProgramDefinition().adminName());
+    inMemoryPdf.fileName = filename;
+    return inMemoryPdf;
```
bion, line 58: Suggestion (blocking): rather than directly setting the members of `inMemoryPdf`, use a public constructor:
```java
return new InMemoryPdf(fileName, byteArray);
```

In general avoid directly assigning instance variables from outside a class.

server/app/services/export/PdfExporter.java
```diff
@@ -1,46 +1,123 @@
...
+                .toLocalDate();
+        Paragraph time =
+            new Paragraph("Answered on : " + date, FontFactory.getFont(FontFactory.HELVETICA, 10));
+        time.setAlignment(Paragraph.ALIGN_RIGHT);
+        document.add(question);
+        document.add(answer);
+        document.add(time);
       }
+    } catch (DocumentException e) {
+      throw e;
```
bion, line 102: suggestion (blocking): this is a no-op and can be removed
```suggestion
```

bion (changes requested):
Nearly there, a couple last changes

bion (approved):

goldblatt (approved):

PR #2395, opened 2022-04-29
Add auto approve for Service Prinicipals in TF 

goldblatt:
We need to auto approve `terraform apply` when using a service principal and not use that if it's a user input

### DIFF
--- a/cloud/shared/bin/lib/terraform.sh
+++ b/cloud/shared/bin/lib/terraform.sh
@@ -1,5 +1,13 @@
 #! /usr/bin/env bash
 
+readonly TERRAFORM_BASE_COMMAND="terraform \
+        -chdir=${TERRAFORM_TEMPLATE_DIR}"
+
+readonly TERRAFORM_APPLY="${TERRAFORM_BASE_COMMAND} \
+        apply \
+        -input=false \
+        -var-file=${TF_VAR_FILENAME}"
+
 #######################################
 # Generates terraform variable files and runs terraform init and apply.
 # Also initializes the storage bucket for tfstate if it's not setup yet.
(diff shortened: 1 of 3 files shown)


goldblatt:
info here: https://learn.hashicorp.com/tutorials/terraform/automate-terraform

cloud/shared/bin/lib/terraform.sh
```diff
@@ -18,14 +19,25 @@ function terraform::perform_apply() {
...
       -backend-config="${BACKEND_VARS_FILENAME}"
   fi
   
-  terraform \
-    -chdir="${TERRAFORM_TEMPLATE_DIR}" \
-    apply \
-    -var-file="${TF_VAR_FILENAME}"
+  if azure::is_service_principal; then
+    terraform \
+        -chdir="${TERRAFORM_TEMPLATE_DIR}" \
```
kmcnellis, line 29: [nit] Might be worth putting this into a variable, so we know to edit both commands together (and just add the extra flag on the end)

kmcnellis (approved):

kmcnellis:
Not sure this PR name is accurate?

cloud/shared/bin/lib/terraform.sh
```diff
@@ -7,6 +7,7 @@
 #   TERRAFORM_TEMPLATE_DIR
 #   BACKEND_VARS_FILENAME
 #   TF_VAR_FILENAME
+#
```
bion, line 10: nit: 
```suggestion
```

bion (approved):

bion (dismissed):

cloud/shared/bin/lib/terraform.sh
```diff
@@ -10,22 +18,23 @@
 #######################################
 function terraform::perform_apply() {
   if [[ "${CIVIFORM_MODE}" == "dev" ]]; then
-    terraform -chdir="${TERRAFORM_TEMPLATE_DIR}" init -upgrade
+    ${TERRAFORM_BASE_COMMAND} init -upgrade
```
shanemc-goog, line 21: Need quotes around all the new ${} commands similar to the below command you're keeping.

shanemc-goog (changes requested):

cloud/shared/bin/lib/terraform.sh
goldblatt: ahh i thought it would make a string

goldblatt: ya this doesn't work as expected-- i can switch it

goldblatt: okay i think this is now what the style guide reccomends

bion (approved):

cloud/shared/bin/lib/terraform.sh
shanemc-goog: #sorry.....  So many characters

shanemc-goog (approved):

PR #2406, opened 2022-05-02
Refactored ProfileAdapter 3 way merge logic into separate class

### LINKED ISSUE #2229: Remove duplication in ProfileAdapters
[SamlCiviFormProfileAdapter.java](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/auth/saml/SamlCiviFormProfileAdapter.java) is a fork of [IdcsProfileAdapter.java](https://github.com/seattle-uat/civiform/blob/main/universal-application-tool-0.0.1/app/auth/oidc/IdcsProfileAdapter.java) which contains a lot profile merging and setup logic.

A common base class can likely be extracted to reduce future errors and copying for the next auth provider.

th174:
### Description
SamlCiviFormProfileAdapter and OidcCiviFormProfileAdapter both conduct a 3 way merge between the database, session context, and authentication provider. This PR extracts the 3 way merge logic into a new class, CiviFormProfileAdapterHelper

### Checklist

### Issue(s)
Fixes #2229

### DIFF
--- a/server/app/auth/CiviFormProfileAdapterHelper.java
+++ b/server/app/auth/CiviFormProfileAdapterHelper.java
@@ -10,6 +10,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import repository.UserRepository;
 
+/**
+ * This class serves to abstract out common application logic in ProfileAdapters into shared helper
+ * methods.
+ */
 public class CiviFormProfileAdapterHelper {
 
   private static final Logger logger = LoggerFactory.getLogger(CiviFormProfileAdapterHelper.class);
@@ -17,23 +21,25 @@ public class CiviFormProfileAdapterHelper {
   private final ProfileFactory profileFactory;
   private final Provider<UserRepository> applicantRepositoryProvider;
 
-  public CiviFormProfileAdapterHelper(ProfileFactory profileFactory,
-      Provider<UserRepository> applicantRepositoryProvider) {
+  public CiviFormProfileAdapterHelper(
+      ProfileFactory profileFactory, Provider<UserRepository> applicantRepositoryProvider) {
     this.profileFactory = profileFactory;
     this.applicantRepositoryProvider = applicantRepositoryProvider;
   }
 
   /**
-   * Now we have a three-way merge situation.
+   * This method handles a three way merge between an existing applicant, an existing guest profile,
+   * and an external profile
    *
    * @param existingApplicant a potentially existing applicant in the database
    * @param existingProfile a guest profile in the browser cookie
    * @param externalProfile an OIDC account in the callback from the OIDC server
-   * <p>We will merge 1 and 2, if present, into `existingProfile`, then merge in `externalProfile`.
    * @param mergeFunction a function that merges an external profile into a Civiform profile
    */
-  public <T> Optional<UserProfile> threeWayMerge(Optional<Applicant> existingApplicant,
-      Optional<CiviFormProfile> existingProfile, T externalProfile,
+  public <T> Optional<UserProfile> threeWayMerge(
+      Optional<Applicant> existingApplicant,
+      Optional<CiviFormProfile> existingProfile,
+      T externalProfile,
       BiFunction<CiviFormProfile, T, UserProfile> mergeFunction) {
 
     if (existingApplicant.isPresent()) {
@@ -48,19 +54,27 @@ public class CiviFormProfileAdapterHelper {
     }
 
     // Now merge in the information sent to us by the OIDC server.
-    return Optional.of(mergeFunction.apply(existingProfile.orElseGet(() -> {
-      logger.debug("Found no existing profile in session cookie.");
-      return profileFactory.wrapProfileData(profileFactory.createNewApplicant());
-    }), externalProfile));
+    return Optional.of(
+        mergeFunction.apply(
+            existingProfile.orElseGet(
+                () -> {
+                  logger.debug("Found no existing profile in session cookie.");
+                  return profileFactory.wrapProfileData(profileFactory.createNewApplicant());
+                }),
+            externalProfile));
   }
 
-  private Optional<CiviFormProfile> twoWayMerge(CiviFormProfile existingProfile, Applicant existingApplicant) {
+  private Optional<CiviFormProfile> twoWayMerge(
+      CiviFormProfile existingProfile, Applicant existingApplicant) {
     // For account, use the existing account and ignore the guest account.
     Applicant guestApplicant = existingProfile.getApplicant().join();
     Account existingAccount = existingApplicant.getAccount();
-    Applicant mergedApplicant = applicantRepositoryProvider.get()
-        .mergeApplicants(guestApplicant, existingApplicant, existingAccount)
-        .toCompletableFuture().join();
+    Applicant mergedApplicant =
+        applicantRepositoryProvider
+            .get()
+            .mergeApplicants(guestApplicant, existingApplicant, existingAccount)
+            .toCompletableFuture()
+            .join();
     return Optional.of(profileFactory.wrap(mergedApplicant));
   }
 }
--- a/server/app/auth/CiviFormProfileAdapterHelper.java
+++ b/server/app/auth/CiviFormProfileAdapterHelper.java
@@ -1,0 +1,66 @@
+package auth;
+
+import java.util.Optional;
+import java.util.function.BiFunction;
+import javax.inject.Provider;
+import models.Account;
+import models.Applicant;
+import org.pac4j.core.profile.UserProfile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import repository.UserRepository;
+
+public class CiviFormProfileAdapterHelper {
+
+  private static final Logger logger = LoggerFactory.getLogger(CiviFormProfileAdapterHelper.class);
+
+  private final ProfileFactory profileFactory;
+  private final Provider<UserRepository> applicantRepositoryProvider;
+
+  public CiviFormProfileAdapterHelper(ProfileFactory profileFactory,
+      Provider<UserRepository> applicantRepositoryProvider) {
+    this.profileFactory = profileFactory;
+    this.applicantRepositoryProvider = applicantRepositoryProvider;
+  }
+
+  /**
+   * Now we have a three-way merge situation.
+   *
+   * @param existingApplicant a potentially existing applicant in the database
+   * @param existingProfile a guest profile in the browser cookie
+   * @param externalProfile an OIDC account in the callback from the OIDC server
+   * <p>We will merge 1 and 2, if present, into `existingProfile`, then merge in `externalProfile`.
+   * @param mergeFunction a function that merges an external profile into a Civiform profile
+   */
+  public <T> Optional<UserProfile> threeWayMerge(Optional<Applicant> existingApplicant,
+      Optional<CiviFormProfile> existingProfile, T externalProfile,
+      BiFunction<CiviFormProfile, T, UserProfile> mergeFunction) {
+
+    if (existingApplicant.isPresent()) {
+      if (existingProfile.isEmpty()) {
+        // Easy merge case - we have an existing applicant, but no guest profile.
+        // This will be the most common.
+        existingProfile = Optional.of(profileFactory.wrap(existingApplicant.get()));
+      } else {
+        // Merge the two applicants and prefer the newer one.
+        existingProfile = twoWayMerge(existingProfile.get(), existingApplicant.get());
+      }
+    }
+
+    // Now merge in the information sent to us by the OIDC server.
+    return Optional.of(mergeFunction.apply(existingProfile.orElseGet(() -> {
+      logger.debug("Found no existing profile in session cookie.");
+      return profileFactory.wrapProfileData(profileFactory.createNewApplicant());
+    }), externalProfile));
+  }
+
+  private Optional<CiviFormProfile> twoWayMerge(CiviFormProfile existingProfile, Applicant existingApplicant) {
+    // For account, use the existing account and ignore the guest account.
+    Applicant guestApplicant = existingProfile.getApplicant().join();
+    Account existingAccount = existingApplicant.getAccount();
+    Applicant mergedApplicant = applicantRepositoryProvider.get()
+        .mergeApplicants(guestApplicant, existingApplicant, existingAccount)
+        .toCompletableFuture().join();
+    return Optional.of(profileFactory.wrap(mergedApplicant));
+  }
+}
--- a/server/app/auth/CiviFormProfileMerger.java
+++ b/server/app/auth/CiviFormProfileMerger.java
@@ -14,14 +14,14 @@ import repository.UserRepository;
  * This class serves to abstract out common application logic in ProfileAdapters into shared helper
  * methods.
  */
-public class CiviFormProfileAdapterHelper {
+public class CiviFormProfileMerger {
 
-  private static final Logger logger = LoggerFactory.getLogger(CiviFormProfileAdapterHelper.class);
+  private static final Logger logger = LoggerFactory.getLogger(CiviFormProfileMerger.class);
 
   private final ProfileFactory profileFactory;
   private final Provider<UserRepository> applicantRepositoryProvider;
 
-  public CiviFormProfileAdapterHelper(
+  public CiviFormProfileMerger(
       ProfileFactory profileFactory, Provider<UserRepository> applicantRepositoryProvider) {
     this.profileFactory = profileFactory;
     this.applicantRepositoryProvider = applicantRepositoryProvider;
--- a/server/app/auth/CiviFormProfileMerger.java
+++ b/server/app/auth/CiviFormProfileMerger.java
@@ -8,7 +8,7 @@ import models.Applicant;
 import org.pac4j.core.profile.UserProfile;
 import repository.UserRepository;
 
-/** This class serves to abstract out common user profile merging logic into a shared helper. */
+/** Helper class for common {@code UserProfile} merging logic. */
 public class CiviFormProfileMerger {
 
   private final ProfileFactory profileFactory;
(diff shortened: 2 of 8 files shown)


server/app/auth/CiviFormProfileAdapterHelper.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.function.BiFunction;
+import javax.inject.Provider;
+import models.Account;
+import models.Applicant;
+import org.pac4j.core.profile.UserProfile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import repository.UserRepository;
+
+public class CiviFormProfileAdapterHelper {
```
bion, line 13: suggestion (blocking): please add documentation for this new class

bion (commented):
Nice improvement! Ready to improve once there's a little more documentation in the code comments

server/app/auth/CiviFormProfileAdapterHelper.java
bion: Got it, `CiviFormProfileAdapterHelper` is pretty generic. I'd prefer to keep it scoped to just what it's implementing in `threeWayMerge`. Maybe `CiviFormProfileMerger`?

bion (changes requested):

server/app/auth/CiviFormProfileAdapterHelper.java
th174: Sure sounds good.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,80 @@
...
+import models.Account;
+import models.Applicant;
+import org.pac4j.core.profile.UserProfile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import repository.UserRepository;
+
+/**
+ * This class serves to abstract out common application logic in ProfileAdapters into shared helper
+ * methods.
```
bion, line 15: nit: update docstring

bion (changes requested):

bion (approved):

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,77 @@
...
+        // Merge the two applicants and prefer the newer one.
+        existingProfile = twoWayMerge(existingProfile.get(), existingApplicant.get());
+      }
+    }
+
+    // Now merge in the information sent to us by the OIDC server.
+    return Optional.of(
+        mergeFunction.apply(
+            existingProfile.orElseGet(
+                () -> {
```
shanemc-goog, line 57: I think this part is skipping creating the profile from the auth data as in


civiformProfileFromOidcProfile
civiformProfileFromSamlProfile

in the two existing flows.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,77 @@
...
+        existingProfile = Optional.of(profileFactory.wrap(existingApplicant.get()));
+      } else {
+        // Merge the two applicants and prefer the newer one.
+        existingProfile = twoWayMerge(existingProfile.get(), existingApplicant.get());
+      }
+    }
+
+    // Now merge in the information sent to us by the OIDC server.
+    return Optional.of(
+        mergeFunction.apply(
```
shanemc-goog, line 55: I'm not following what this is doing, it seems like a no-op.

If there's an existingProfile, won't this merge it with itself?  And if there isn't it'll merge it with an empty one.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,77 @@
...
+    this.applicantRepositoryProvider = applicantRepositoryProvider;
+  }
+
+  /**
+   * This method handles a three way merge between an existing applicant, an existing guest profile,
+   * and an external profile
+   *
+   * @param existingApplicant a potentially existing applicant in the database
+   * @param existingProfile a guest profile in the browser cookie
+   * @param externalProfile an OIDC account in the callback from the OIDC server
```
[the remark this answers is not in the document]
th174: You're right. I copy pasted overzealously

server/app/auth/CiviFormProfileMerger.java
th174: If there is an existingProfile, it gets merged with the external profile. If there isn't an existing profile, then external profile is merged with an empty one.

server/app/auth/CiviFormProfileMerger.java
th174: Oh good catch. I'll add that.

bion:
JFYI @th174 I saw that you added nat-henderson as a reviewer. Nat contributed a bunch to this project but is no longer on it (no longer at Google, in fact) so while they are more than welcome to contribute if they'd like, we shouldn't expect their time, so probably most appropriate to remove them as a reviewer.

server/app/auth/CiviFormProfileMerger.java
shanemc-goog: oh existingProfile vs externalProfile is hard to read, I'd suggest renaming externalProfile.  authenticatorProfile, etc

shanemc-goog: Also, Where are you doing the civiformProfileFromOidcProfile and civiformProfileFromSamlProfile functionality?  

You're removing the only callers of them in this PR

nat-henderson:
> JFYI @th174 I saw that you added nat-henderson as a reviewer. Nat contributed a bunch to this project but is no longer on it (no longer at Google, in fact) so while they are more than welcome to contribute if they'd like, we shouldn't expect their time, so probably most appropriate to remove them as a reviewer.

Thanks - I'd like to, and if I'd gotten the email during my unemployed window I'd have been happy to... but my new employer would need to go through an approval process for me to be able to chime in, and the lawyers haven't gotten back to me yet. Sorry!

server/app/auth/CiviFormProfileMerger.java
shanemc-goog: You can also dev test OIDC with these steps

https://docs.civiform.us/contributor-guide/developer-guide/authentication-providers#testing

and you should see Accounts.authority_id get set correctly with an iss: and sub: data in the string.

th174: > Also, Where are you doing the civiformProfileFromOidcProfile and civiformProfileFromSamlProfile functionality?

The previous logic was this for `civiformProfileFromOidcProfile`

1. Creating an empty profile
2. Merging the authenticator profile into the empty profile.

Since step two of that logic is duplicated, I instead replaced those methods with `createEmptyCiviFormProfile` in the new commit of this PR.

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
```diff
@@ -97,7 +107,7 @@ public CiviFormProfileData mergeCiviFormProfile(
   }
 
   /** Create a totally new CiviForm profile from the provided OidcProfile. */
```
shanemc-goog, line 109: Need to update this doc as the new profile isn't based on the OidcProfile anymore.  This kinda "smells" though which likely means we need more refactoring after this PR.

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
```diff
@@ -73,7 +77,13 @@ protected Optional<String> getAuthorityId(OidcProfile oidcProfile) {
 
   /** Merge the two provided profiles into a new CiviFormProfileData. */
   public CiviFormProfileData mergeCiviFormProfile(
-      CiviFormProfile civiformProfile, OidcProfile oidcProfile) {
+      Optional<CiviFormProfile> maybeCiviformProfile, OidcProfile oidcProfile) {
+    var civiformProfile =
+        maybeCiviformProfile.orElseGet(
+            () -> {
+              logger.debug("Found no existing profile in session cookie.");
+              return this.createEmptyCiviFormProfile(oidcProfile);
```
shanemc-goog, line 85: nit: don't need "this."

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
```diff
@@ -73,7 +77,13 @@ protected Optional<String> getAuthorityId(OidcProfile oidcProfile) {
 
   /** Merge the two provided profiles into a new CiviFormProfileData. */
   public CiviFormProfileData mergeCiviFormProfile(
```
shanemc-goog, line 79: I don't believe this parent method is ever called with an empty optional correct?

 so another "smell" that likely implies we need more refactoring subsequently.   This may need to be a protected method with a different name for instance.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param applicantInDatabase a potentially existing applicant in the database
+   * @param sessionProfile a guest profile from the browser session
+   * @param authProviderProfile profile data from an external auth provider, such as OIDC
+   * @param mergeFunction a function that merges an external profile into a Civiform profile, or
+   *     provides one if it doesn't exist
+   */
+  public <T> Optional<UserProfile> threeWayMerge(
+      Optional<Applicant> applicantInDatabase,
+      Optional<CiviFormProfile> sessionProfile,
```
shanemc-goog, line 35: nit: sessionProfile -> sessionGuestProfile or just guestProfile

Non guests also have session data though not a profile I believe., so just trying to be as clear as possible.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+        // Easy merge case - we have an existing applicant, but no guest profile.
+        // This will be the most common.
+        sessionProfile = Optional.of(profileFactory.wrap(applicantInDatabase.get()));
+      } else {
+        // Merge the two applicants and prefer the newer one.
+        sessionProfile = twoWayMerge(sessionProfile.get(), applicantInDatabase.get());
+      }
+    }
+
+    // Merge externalProfile into existingProfile
```
shanemc-goog, line 50:
```suggestion
    // Merge externalProfile into existingProfile.
```

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+    // Merge function will create a new CiviFormProfile if it doesn't exist,
+    // or otherwise handle it in some way
```
shanemc-goog, line 52: "in some way"  can you clarify this better? it doesn't add much as is.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public CiviFormProfileMerger(
+      ProfileFactory profileFactory, Provider<UserRepository> applicantRepositoryProvider) {
+    this.profileFactory = profileFactory;
+    this.applicantRepositoryProvider = applicantRepositoryProvider;
+  }
+
+  /**
+   * This method handles a three way merge between an existing applicant, an existing guest profile,
+   * and an external profile
```
shanemc-goog, line 25: We follow the google java and javadoc styleguides https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

So this summary line should be more like this. a fragment with ending punctuation.

\* Merges multiple profile sources into a {@code UserProfile}.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+
+    // Merge externalProfile into existingProfile
+    // Merge function will create a new CiviFormProfile if it doesn't exist,
+    // or otherwise handle it in some way
+    return Optional.of(mergeFunction.apply(sessionProfile, authProviderProfile));
+  }
+
+  private Optional<CiviFormProfile> twoWayMerge(
```
shanemc-goog, line 56: Worth a javadoc too, and maybe a different name.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Optional;
+import java.util.function.BiFunction;
+import javax.inject.Provider;
+import models.Account;
+import models.Applicant;
+import org.pac4j.core.profile.UserProfile;
+import repository.UserRepository;
+
+/** This class serves to abstract out common user profile merging logic into a shared helper. */
```
shanemc-goog, line 11: Similar to the below comment, use a sentence fragment.


 \* Common {@code UserProfile} merging logic.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+
+    // Merge externalProfile into existingProfile
+    // Merge function will create a new CiviFormProfile if it doesn't exist,
+    // or otherwise handle it in some way
+    return Optional.of(mergeFunction.apply(sessionProfile, authProviderProfile));
+  }
+
+  private Optional<CiviFormProfile> twoWayMerge(
+      CiviFormProfile existingProfile, Applicant existingApplicant) {
+    // For account, use the existing account and ignore the guest account.
```
shanemc-goog, line 58: this comment doesn't make a lot of sense out of context, maybe it'd be better on the caller of this?

server/test/auth/CiviFormProfileMergerTest.java
```diff
@@ -0,0 +1,134 @@
...
+    applicant = spy(new Applicant());
+    applicant.setAccount(account);
+    account.setApplicants(Collections.singletonList(applicant));
+  }
+
+  @Test
+  public void threeWayMerge_succeeds_noExistingApplicantAndNoExistingProfile() {
+    var merged =
+        civiFormProfileMerger.threeWayMerge(
+            Optional.empty(),
```
shanemc-goog, line 65: inline doc all "ambiguous" parameters for readability as well if it ever changes we'll get a lint error which may imply that the functionality has also changed.

EG

 /* applicantInDabase= */

server/test/auth/ProfileMergeTest.java
```diff
@@ -82,7 +83,7 @@ public void testProfileCreation() throws ExecutionException, InterruptedExceptio
     OidcProfile oidcProfile = createOidcProfile("foo@example.com", "issuer", "subject");
 
     CiviFormProfileData profileData =
-        idcsProfileAdapter.civiformProfileFromOidcProfile(oidcProfile);
+        idcsProfileAdapter.mergeCiviFormProfile(Optional.empty(), oidcProfile);
```
shanemc-goog, line 86: inline doc the Optional.empty() added in this file similar to the other comment.

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -0,0 +1,69 @@
...
+   * This method handles a three way merge between an existing applicant, an existing guest profile,
+   * and an external profile
+   *
+   * @param applicantInDatabase a potentially existing applicant in the database
+   * @param sessionProfile a guest profile from the browser session
+   * @param authProviderProfile profile data from an external auth provider, such as OIDC
+   * @param mergeFunction a function that merges an external profile into a Civiform profile, or
+   *     provides one if it doesn't exist
+   */
+  public <T> Optional<UserProfile> threeWayMerge(
```
shanemc-goog, line 33: I would be nice to have clearer names here.  "threeWay" doesn't add much beyond looking at the parameters.

Just merge() is good.

Also if you can provide any extra method docs to help the reader understand the interaction among the 4 parameters, that'd be great!

shanemc-goog (changes requested):
Thank you!  Mostly comment/naming things.

server/app/auth/CiviFormProfileMerger.java
th174: This is kept vague intentionally. The merge function is passed in, so the caller can decide how to handle a non-existent CiviFormProfile. The current callers handle it by creating either a new Admin profile or a new Applicant profile, but other potential options include throwing an exception, providing a different existing profile in some way, etc.

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
th174: The parent method is called with an empty optional if there's no existing profile. This replaces the existing logic in `createCiviFormProfileFromOidcProfile`, combining the two methods with similar logic into one.

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
th174: The new profile is potentially dependent on the OidcProfile, because the AdfsProfileAdapter uses it to decide whether to create a global admin or program admin.

shanemc-goog: Let's leave it as is in this PR, but it's an oddity that usually is a hint/smell that the abstractions aren't correct.

Update the comment though to break that the new profile is "from" the Oidc one since that's not necessarily the case anymore.  Maybe "informed by the OidcProfile"

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
shanemc-goog: To rephrase, the only callers of this OIDC Adapter mergeCiviFormProfile, are the children's [overrides](https://github.com/seattle-uat/civiform/search?q=mergeCiviFormProfile) as far as I see.

And they all ensure a non empty Optional is passed in here, which means that handling that case is not necessary here. 

So the fact that we don't need it but must have it due to inheritance likely indicates an abstraction issue.  Given it's not actually called outside of its children, indicates that this functionality doesn't need to be inherited and is more of a utility to the children.

A slightly better definition might be

```
/**
 * Some doc indicating that if the optional is empty one will be made.
 */
abstract public CiviFormProfileData mergeCiviFormProfile(Optional<CiviFormProfile> maybeCiviformProfile, OidcProfile oidcProfile);

/** Merge the two provided profiles into a new CiviFormProfileData. */
protected CiviFormProfileData mergeCiviFormProfile(CiviFormProfile civiformProfile, OidcProfile oidcProfile) {
...... existing code
}
```

th174: The parent's `mergeCiviFormProfile` is passed into the `CiviFormProfileMerger` on line 137. It then gets called inside CiviFormProfileMerger. This behavior is not overridden by the children.

bion (approved):

server/app/auth/oidc/OidcCiviFormProfileAdapter.java
shanemc-goog: This seems good. The parent takes the Optional and resolves it rather than all the children.

Can the non-optional method on L94 be private now?

th174: It can't be private because IdcsProfileAdapter overrides it.

shanemc-goog: ah ok, thanks.

Yeah add the VisibleForTesting.

server/test/auth/CiviFormProfileMergerTest.java
```diff
@@ -0,0 +1,134 @@
...
+import repository.ResetPostgres;
+import repository.UserRepository;
+
+public class CiviFormProfileMergerTest extends ResetPostgres {
+
+  private static final String EMAIL_ATTR = "user_emailid";
+  private static final String EMAIL1 = "foo@bar.com";
+  private static final String EMAIL2 = "bar@foo.com";
+  private static final Long ACCOUNT_ID = 12345L;
+  private static final UserProfile USER_PROFILE = new CommonProfile();
```
shanemc-goog, line 28: Classes tend to not be deeply immutable, so they shouldn't use const case per the style guide.  EG: they can effectively change and CONSTS should never change.

We likely want to make a new one for each test just in case too.

server/test/auth/CiviFormProfileMergerTest.java
```diff
@@ -0,0 +1,134 @@
...
+  private OidcProfile oidcProfile;
+  private CiviFormProfileData civiFormProfileData;
+  private Applicant applicant;
+  private Account account;
+
+  private CiviFormProfileMerger civiFormProfileMerger;
+
+  @Before
+  public void setup() {
+    profileFactory = spy(instanceOf(ProfileFactory.class));
```
shanemc-goog, line 42: We have full access to the database in these ResetPostgres tests.  Did you otherwise need spies rather than using the database directly?

In the other test you'll see things like 

profileFactory = instanceOf(ProfileFactory.class);
UserRepository repository = instanceOf(UserRepository.class);

th174: You're right, I started out trying to use the database, but then switched to using mocks because the database logic was already being tested in other unit test classes. I'll switched over to using Mockito fully.

shanemc-goog (approved):
THANK YOU!

bion:
Great work!

PR #2409, opened 2022-05-02
Fix hard-coded references to `America/Los_Angeles` timezone

### LINKED ISSUE #2341: Time rendering is fixed to Los Angeles timezone
**Describe the bug**
Our [visual date rendering](https://github.com/seattle-uat/civiform/blob/4ca24782255d2b1f09daf58c43408c46621b0f43/universal-application-tool-0.0.1/app/views/BaseHtmlView.java#L84) fixes times to LA.

We need to make this work correctly for any timezone.

clouser2:
### Description
We allow overriding the timezone value through the `java.time.zoneid` [configuration value](https://github.com/seattle-uat/civiform/blob/26a548ff2ae4135590a7cb22c2e466e689163383/server/conf/application.conf#L476). This wasn't consistently applied in the codebase, including where we provide a `Clock` instance to downstream callers.

This change has the Guice module consult the configuration value when building a `Clock` instance and converts callsites that read the config value directly to instead inject the provided Clock instance. 

### Checklist

### Issue(s)
Fixes #2341


shanemc-goog:
unit tests aren't passing FYI

swatkat1:
I believe the browser tests also look for the exported files.. may need update there..

server/app/modules/MainModule.java
```diff
@@ -22,10 +26,15 @@ public class MainModule extends AbstractModule {
...
-    // Use the system clock as the default implementation of Clock
-    bind(Clock.class).toInstance(Clock.system(ZoneId.of("America/Los_Angeles")));
     bind(ProgramService.class).to(ProgramServiceImpl.class);
     bind(QuestionService.class).to(QuestionServiceImpl.class);
     bind(ApplicantService.class).to(ApplicantServiceImpl.class);
   }
+
+  @Provides
+  public Clock provideClock(Config config) {
+    ZoneId zoneId = ZoneId.of(checkNotNull(config).getString("java.time.zoneid"));
```
shanemc-goog, line 36: Do you know how this java.time.zoneid param works? I'm surprised you have to explicitly read it versus java handling it for us.

clouser2:
This is now ready for review. Sorry for all of the noise as I figured out why the test was failing.

It turned out that my changes to MainModuleTest didn't tear down the created Play app originally. This led to a leaked Postgres connection, which in turn cause subsequent tests to fail since we have a connection pool limit on the number of Postgres connections. I'll brainstorm if there's a way to try and prevent this (a separate PR for sure) and update the Troubleshooting docs on the wiki.

server/app/modules/MainModule.java
clouser2: This actually isn't something that Java handles by default, though it's name definitely implies it. I believe the original naming was due to the fact that it's setting the string value for a `java.time.ZoneId` instance :/.

clouser2:
> I believe the browser tests also look for the exported files.. may need update there..

Currently, I don't believe browser tests are looking for the exported file.

server/app/modules/MainModule.java
shanemc-goog: I'd suggest we remove that confusion if it's in scope for this PR.  Does anything use it currently?

swatkat1 (commented):
Please update the PdfExporter class which defines the name of the pdf file.

server/app/modules/MainModule.java
clouser2: As I understand it, no. To be extra-cautious, I'd suggest we rename the config value in this PR and keep the environment-variable override as `JAVA_TIME_ZONE_ID` until we've confirmed that no deployments set that. Once that's confirmed, I'll also rename the environment variable.

Latest commit incorporates your change.

swatkat1:
In application_download.test.ts, there is a GetJson() and getCsv() which loads the downloaded file into the tests and checks the file. But to your point, they check the downloaded file but not the name of the file per say.

server/app/modules/MainModule.java
bion: Thanks for being cautious but in this case I think you can make the more expedient change.

Seattle isn't setting it: https://github.com/seattle-uat/civiform-deploy/blob/main/infra/containers.yaml#L123-L161
Rochester isn't setting it either since it's not configurable via terraform at the moment.

bion (approved):
This is a huge improvement and something that's well overdue. Thank you!

The only suggestion I have is potentially binding more values so not everything is being accessed through injecting `Clock`. Totally optional, but may be a little more convenient/ergonomic for developers.

E.g. for `now`:

Add a new `BindingAnnotations.java` class:

```java
@Qualifier
@Target({ FIELD, PARAMETER, METHOD })
@Retention(RUNTIME)
public @interface Now {}
```

Then in `MainModule.java`:

```java
@Provides
@Now
public LocalDateTime provideNow(Clock clock) {
  return LocalDateTime.now(clock);
}

@Provides
public ZoneId provideLocalTimezone(Config config) {
  return ZoneId.of(checkNotNull(config).getString("civiform.time.zoneid"));
}

@Provides
public Clock provideClock(ZoneId zoneId) {
  return Clock.system(zoneId);
}
```

server/app/modules/MainModule.java
clouser2: Done.

clouser2:
@bion. Incorporated your suggestion for other bindings, good idea!

One caveat is that when injecting the `LocalDateTime` in various constructors, I made sure to inject a `@Now Provider<LocalDateTime>` so that the instant wouldn't be cached on construction. Other than that, the updates were purely mechanical.

shanemc-goog (approved):
Looks good! Just small suggestions

server/test/modules/MainModuleTest.java
```diff
@@ -0,0 +1,68 @@
...
+      Helpers.stop(app);
+    }
+  }
+
+  @Test
+  public void testTimeZone_configValueNotSet_DefaultsToPST() {
+    app = new GuiceApplicationBuilder().build();
+
+    Clock clock = app.injector().instanceOf(Clock.class);
+    assertThat(clock.getZone()).isEqualTo(ZoneId.of("America/Los_Angeles"));
```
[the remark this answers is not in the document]
clouser2: Done.

server/test/modules/MainModuleTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.runner.RunWith;
+import play.Application;
+import play.inject.BindingKey;
+import play.inject.guice.GuiceApplicationBuilder;
+import play.test.Helpers;
+
+@RunWith(JUnitParamsRunner.class)
+public class MainModuleTest {
+
+  private Application app;
```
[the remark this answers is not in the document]
clouser2: RE Optional: Converted.

RE @Before: Added a brief comment. We initialize the app in each test since we want a distinct configuration for each. The call to Helpers.stop() is necessary to prevent leaking a DB connection that would cause further tests to fail. Having a @Before isn't necessary since we initialize the app in each test method.

PR #2412, opened 2022-05-03
Service for creating API keys

bion:
Introduces a new class `services.apikeys.ApiKeyService` that is meant to encapsulate management of API keys. This PR implements `ApiKeyService.createApiKey` which takes form input for creating a key, validates the input, and creates a key.

I did a somewhat sloppy job of an interactive rebase where I split this with some other changes for the UI side in an effort to reduce PR size for reviewers so there's a couple items that aren't directly used yet but they're minor enough I'm hoping it's ok to include them here.


server/app/services/program/ProgramServiceImpl.java
```diff
@@ -105,6 +106,18 @@ private CompletionStage<ProgramDefinition> syncProgramAssociations(Program progr
     return CompletableFuture.completedStage(programDefinition.orderBlockDefinitions());
   }
 
+  @Override
+  public ImmutableSet<String> getAllProgramNames() {
+    return programRepository.getAllProgramNames();
+  }
```
bion, line 112: This method isn't used in this PR but will be in a follow on for implementing the UI.

server/build.sbt
```diff
@@ -80,6 +80,9 @@ lazy val root = (project in file("."))
       // Slugs for deeplinking.
       "com.github.slugify" % "slugify" % "2.5",
 
+      // Apache libraries for testing subnets
+      "commons-net" % "commons-net" % "3.8.0",
```
bion, line 84: Not used in this PR but will be in a follow-on.

bion, line 84: Provides `SubnetUtils`

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+
+    if (authorityId != null) {
+      return authorityId;
+    }
+
+    if (environment.isDev()) {
+      return "dev-admin";
+    }
+
+    throw new RuntimeException("ApiKey creator must have authority_id.");
```
kmcnellis, line 225: How does play handle Runtime Exceptions?  Is this something that should returned a permission_denied or similar error code?

server/test/services/apikey/ApiKeyServiceTest.java
```diff
@@ -0,0 +1,177 @@
...
+                "grant-program-read[test-program]", "true"));
+
+    assertThrows(
+        ProgramNotFoundException.class, () -> apiKeyService.createApiKey(form, adminProfile));
+  }
+
+  private DynamicForm buildForm(ImmutableMap<String, String> formContents) {
+    return formFactory.form().bindFromRequest(fakeRequest().bodyForm(formContents).build());
+  }
+}
```
kmcnellis, line 177: Add a test for "ApiKey creator must have authority_id"?

kmcnellis (changes requested):

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+   */
+  public ApiKeyCreationResult createApiKey(DynamicForm form, CiviFormProfile profile)
+      throws ProgramNotFoundException {
+    if (environment.isProd() && secretSalt.equals("changeme")) {
+      throw new RuntimeException("Must set api_secret_salt in production environment.");
+    }
+
+    ApiKey apiKey = new ApiKey();
+
+    form = resolveKeyName(form, apiKey);
```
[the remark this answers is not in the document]
bion: The circumstance is a bit unfortunate. `apiKey` is an ebean entity/model and is mutable. `form` is play form object and is immutable -- adding validation error messages using `form.withError` returns a new form object with the error message added.

> Maybe make them more functional? Only use the helper to extract the value needed, and the main function can apply it to the api key (and throw/catch the error).

I'm having a hard time with what that would like in practice, given the constraints of the underlying objects and desired behavior.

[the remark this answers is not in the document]
bion: What do you think of just adding an explanatory comment over this set of lines?

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    } else {
+      form = form.withError("keyName", "Key name cannot be blank.");
+    }
+
+    return form;
+  }
+
+  private DynamicForm resolveExpiration(DynamicForm form, ApiKey apiKey) {
+    Optional<String> maybeExpirationString =
+        Optional.ofNullable(form.rawData().getOrDefault("expiration", null));
```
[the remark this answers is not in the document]
bion: Ah nice one, I did it slightly differently but same idea

server/app/services/apikey/ApiKeyService.java
bion: Unhandled exceptions result in a 500 status code.

In the circumstance here, there should always be an authority_id. Only admins can call this method, and all admins should have an authority_id in prod. In dev and test we can use a dummy value. If this code somehow runs in prod and is passed a profile that doesn't have an authority_id, that's a violation of an invariant in the system and a server error we should look into.

server/test/services/apikey/ApiKeyServiceTest.java
bion: done

shanemc-goog (changes requested):
I don't have more time for a bit, so shooting these comments to you now

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+      this.credentials = credentials;
+      this.form = form;
+    }
+
+    /** Returns true for the key was created. */
+    public boolean successful() {
+      return credentials.isPresent();
+    }
+
+    /** Returns the API key if successful, throws {@code NullPointerException} if not. */
```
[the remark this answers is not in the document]
shanemc-goog: by declared I meant a non RuntimeException

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    } catch (IllegalArgumentException e) {
+      return form.withError("subnet", "Subnet must be in CIDR notation.");
+    }
+
+    apiKey.setSubnet(maybeSubnetString.get());
+
+    return form;
+  }
+
+  private static final Pattern GRANT_PROGRAM_READ_PATTERN =
```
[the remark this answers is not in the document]
bion: The [styleguide](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents) entry on ordering doesn't require consts at the top of the class. In cases where a const is only used in one place IMO it's nicer to have it near that place for ease of reference when reading the code. How strong is your preference for moving it?

Added an example comment.

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+  /** Holds state relevant to the result of attempting to create an {@link ApiKey}. */
+  public static class ApiKeyCreationResult {
+    private final Optional<ApiKey> apiKey;
+    private final Optional<DynamicForm> form;
+    private final Optional<String> credentials;
+
+    /** Constructs an instance in the case of success. */
+    public static ApiKeyCreationResult success(ApiKey apiKey, String credentials) {
+      return new ApiKeyCreationResult(
+          Optional.of(apiKey), Optional.of(credentials), Optional.empty());
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /** Constructs an instance in the case of failure. */
+    public static ApiKeyCreationResult failure(DynamicForm form) {
+      return new ApiKeyCreationResult(Optional.empty(), Optional.empty(), Optional.of(form));
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    private ApiKeyCreationResult(
+        Optional<ApiKey> apiKey, Optional<String> credentials, Optional<DynamicForm> form) {
+      this.apiKey = apiKey;
+      this.credentials = credentials;
+      this.form = form;
+    }
+
+    /** Returns true for the key was created. */
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    } catch (DateTimeParseException e) {
+      return form.withError("expiration", "Expiration must be in the form YYYY-MM-DD.");
+    }
+
+    return form;
+  }
+
+  private DynamicForm resolveSubnet(DynamicForm form, ApiKey apiKey) {
+    Optional<String> maybeSubnetString =
+        Optional.ofNullable(form.rawData().getOrDefault("subnet", null));
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+    apiKey.setCreatedBy(getAuthorityId(profile));
+
+    apiKey = repository.insert(apiKey).toCompletableFuture().join();
+
+    return ApiKeyCreationResult.success(apiKey, credentials);
+  }
+
+  private DynamicForm resolveKeyName(DynamicForm form, ApiKey apiKey) {
+    Optional<String> maybeKeyName =
+        Optional.ofNullable(form.rawData().getOrDefault("keyName", null));
```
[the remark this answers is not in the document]
bion: not seeing a method on Map called "maybeGet"? tried a different approach lmk what you think

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,291 @@
...
+  private final String secretSalt;
+  private final boolean banGlobalSubnet;
+
+  @Inject
+  public ApiKeyService(
+      ApiKeyRepository repository,
+      Environment environment,
+      Config config,
+      ProgramService programService) {
+    this.environment = checkNotNull(environment);
```
kmcnellis, line 66: [nit] it's nice to put these in the same order as the method params

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,291 @@
...
+    if (environment.isDev()) {
+      return "dev-admin";
+    }
+
+    throw new RuntimeException("ApiKey creator must have authority_id.");
+  }
+
+  /** Holds state relevant to the result of attempting to create an {@link ApiKey}. */
+  public static class ApiKeyCreationResult {
+    private final Optional<ApiKey> apiKey;
```
kmcnellis, line 242: [nit] it's nice to put these in the same order as the constructor params

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
kmcnellis: I'm fine with that.  And docstrings on the helper functions to mention that it's mutating the apiKey

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
kmcnellis: Gotcha

kmcnellis (approved):

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
shanemc-goog: For me it's largely about consistency and I don't think we inline in general..., at least I don't recall seeing it.

Def not strong enough to block, I guess if it's localized to the one usage then that is ok.

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
bion: Hmm how about instead of letting the optional nullpointerexeption throw we instead throw a runtimexception with a message like "cannot retrieve credentials/apiKey from a failed ApiKeyMutationResult"?

server/app/services/apikey/ApiKeyService.java
bion: done

server/app/services/apikey/ApiKeyService.java
bion: done

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
bion: done

server/app/models/ApiKey.java
```diff
@@ -28,6 +28,10 @@ public class ApiKey extends BaseModel {
   @DbJsonB private ApiKeyGrants grants;
 
   public ApiKeyGrants getGrants() {
+    if (grants == null) {
+      this.grants = new ApiKeyGrants();
```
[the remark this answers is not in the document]
bion: good call out, yeah the more established pattern is to pass the object to be serialized in as a constructor arg. updated ApiKey to do that instead

server/app/services/DateConverter.java
```diff
@@ -0,0 +1,20 @@
...
+public class DateConverter {
+
+  /**
+   * Parses a string containing a ISO-8601 date (i.e. "YYYY-MM-DD") and converts it to an {@link
+   * Instant} at the beginning of the day specified by the date in UTC.
+   */
+  public static Instant parseIso8601DateToStartOfDateInstant(String dateString) {
+    return LocalDate.parse(dateString, DateTimeFormatter.ISO_DATE)
+        .atStartOfDay()
+        .toInstant(ZoneOffset.UTC);
```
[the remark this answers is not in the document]
bion: switched to using the app clock's time zone.

> (btw When applicants enter a date is that LA or UTC? that's the closest analog here)

we store dates as strings in `ApplicantData`, so no need for time zones

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,278 @@
...
+   * @param form Stores data specified by the user for creation of the key, as well as validation
+   *     errors
+   * @param profile The user profile of the account creating this API key, used for recording who
+   *     created it
+   * @return a result object containing the credentials and created key if successful, form with
+   *     validation messages if not
+   */
+  public ApiKeyCreationResult createApiKey(DynamicForm form, CiviFormProfile profile)
+      throws ProgramNotFoundException {
+    if (environment.isProd() && secretSalt.equals("changeme")) {
```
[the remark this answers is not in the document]
bion: done

server/app/services/DateConverter.java
```diff
@@ -0,0 +1,30 @@
...
+    this.clock = checkNotNull(clock);
+  }
+
+  /**
+   * Parses a string containing a ISO-8601 date (i.e. "YYYY-MM-DD") and converts it to an {@link
+   * Instant} at the beginning of the day in local time zone.
+   */
+  public Instant parseIso8601DateToStartOfDateInstant(String dateString) {
+    return LocalDate.parse(dateString, DateTimeFormatter.ISO_DATE)
+        .atStartOfDay(clock.getZone())
```
shanemc-goog, line 27: Suggestion: Add a todo for Brett's Issue to update this too with the new injected Zone he's adding.  assuming you don't merge and update before.

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,302 @@
...
+    if (!keyName.isBlank()) {
+      apiKey.setName(keyName);
+    } else {
+      form = form.withError(FORM_FIELD_NAME_KEY_NAME, "Key name cannot be blank.");
+    }
+
+    return form;
+  }
+
+  // apiKey is mutable and modified here, form is immutable so a new instance is returned
```
shanemc-goog, line 137:
```suggestion
  // {@code apiKey} is mutable and modified here, {@code form} is immutable so a new instance is returned.
```

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,302 @@
...
+   * @param form Stores data specified by the user for creation of the key, as well as validation
+   *     errors
+   * @param profile The user profile of the account creating this API key, used for recording who
+   *     created it
+   * @return a result object containing the credentials and created key if successful, form with
+   *     validation messages if not
+   */
+  public ApiKeyCreationResult createApiKey(DynamicForm form, CiviFormProfile profile)
+      throws ProgramNotFoundException {
+    if (environment.isProd() && secretSalt.equals(DEFAULT_API_KEY_SECRET_SALT)) {
```
shanemc-goog, line 90: Shipping a in-code default that can't be used is a bit smelly.

Do we have dev configs that we could put that in instead, and then check that the salt is set at all here?

server/test/services/apikey/ApiKeyServiceTest.java
```diff
@@ -0,0 +1,213 @@
...
+import play.data.FormFactory;
+import repository.ApiKeyRepository;
+import repository.ResetPostgres;
+import services.DateConverter;
+import services.apikey.ApiKeyService.ApiKeyCreationResult;
+import services.program.ProgramNotFoundException;
+
+public class ApiKeyServiceTest extends ResetPostgres {
+
+  ApiKeyService apiKeyService;
```
shanemc-goog, line 29: make all these private

server/test/services/apikey/ApiKeyServiceTest.java
```diff
@@ -0,0 +1,213 @@
...
+  ApiKeyRepository apiKeyRepository;
+  ProfileFactory profileFactory;
+  CiviFormProfile adminProfile;
+  FormFactory formFactory;
+  DateConverter dateConverter;
+
+  @Before
+  public void setUp() throws Exception {
+    apiKeyService = instanceOf(ApiKeyService.class);
```
shanemc-goog, line 38: general question: can these instanceOfs be in a @beforeClass ?

If so that'd be good and then have the other setup in a @Before.

Google style suggests having different/multiple @Before s for different setup concerns; not saying we have to do that, but made me wonder if these allow for that separation.

shanemc-goog (changes requested):

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,302 @@
...
+
+    private ApiKeyCreationResult(
+        Optional<ApiKey> apiKey, Optional<String> credentials, Optional<DynamicForm> form) {
+      this.apiKey = apiKey;
+      this.credentials = credentials;
+      this.form = form;
+    }
+
+    /** Returns true if the key was created. */
+    public boolean successful() {
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,302 @@
...
+
+    if (environment.isDev()) {
+      return "dev-admin";
+    }
+
+    throw new RuntimeException("ApiKey creator must have authority_id.");
+  }
+
+  /** Holds state relevant to the result of attempting to create an {@link ApiKey}. */
+  public static class ApiKeyCreationResult {
```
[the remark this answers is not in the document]
bion: done

server/app/models/ApiKey.java
```diff
@@ -27,6 +27,10 @@ public class ApiKey extends BaseModel {
   /** Permissions granted to this ApiKey by the admin. */
   @DbJsonB private ApiKeyGrants grants;
 
+  public ApiKey(ApiKeyGrants grants) {
```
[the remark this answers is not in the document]
bion: done

server/app/services/apikey/ApiKeyService.java
bion: > Do we have dev configs that we could put that in instead, and then check that the salt is set at all here?

We could, but the result is that we've have to require the secret salt to be present for the server to start at all, rather than this behavior which will just block this one feature if it hasn't been set yet.

I think I prefer the option that doesn't risk downtime.

server/app/services/apikey/ApiKeyService.java
bion: done

server/app/services/apikey/ApiKeyService.java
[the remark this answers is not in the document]
bion: added to the class-level javadoc

server/test/services/apikey/ApiKeyServiceTest.java
bion: done

server/test/services/apikey/ApiKeyServiceTest.java
bion: `@BeforeClass` can only be used with static methods, so nothing that relies on dependency injection can use it

server/app/services/DateConverter.java
shanemc-goog: Bion updated post Brett's merge

server/app/services/apikey/ApiKeyService.java
```diff
@@ -0,0 +1,250 @@
...
+ *
+ * <p>The plaintext secret is revealed to the admin user once after creation, after which it is not
+ * recoverable from CiviForm.
+ */
+public class ApiKeyService {
+
+  public static final String FORM_FIELD_NAME_KEY_NAME = "keyName";
+  public static final String FORM_FIELD_NAME_EXPIRATION = "expiration";
+  public static final String FORM_FIELD_NAME_SUBNET = "subnet";
+  public static final String DEFAULT_API_KEY_SECRET_SALT = "changeme";
```
shanemc-goog, line 50: Can you add comments here and in application.conf to the effect of IfThisThenThat since we don't have it?

EG the two values must stay in sync.

shanemc-goog (changes requested):

server/app/services/apikey/ApiKeyService.java
bion: done

shanemc-goog (approved):

PR #2415, opened 2022-05-03
Prevent program name slug collisions

bion:
When a program is created, CiviForm checks to ensure there isn't already a program with the same name. This is critical because the program name is immutable, and used to link multiple program records as versions of the same program.

The program slug -- the name formatted for use in a URL -- must also be unique for the deep link feature to work. However it is possible to have two name strings that are different that slugify to the same value. This change adds a validation check to ensure new program names do not slugify to slugs that collide with existing program slugs.

### Checklist

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -5,11 +5,13 @@ import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import io.ebean.DB;
 import io.ebean.Database;
 import io.ebean.ExpressionList;
 import io.ebean.PagedList;
 import io.ebean.Query;
+import io.ebean.SqlRow;
 import io.ebean.Transaction;
 import io.ebean.TxScope;
 import java.util.List;
@@ -64,6 +66,17 @@ public class ProgramRepository {
   public Program updateProgramSync(Program program) {
     database.update(program);
     return program;
+  }
+
+  public ImmutableSet<String> getAllProgramNames() {
+    ImmutableSet.Builder<String> names = ImmutableSet.builder();
+    List<SqlRow> rows = database.sqlQuery("SELECT DISTINCT name FROM programs").findList();
+
+    for (SqlRow row : rows) {
+      names.add(row.getString("name"));
+    }
+
+    return names.build();
   }
 
   /**
--- a/server/app/services/program/ProgramServiceImpl.java
+++ b/server/app/services/program/ProgramServiceImpl.java
@@ -2,6 +2,7 @@ package services.program;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.github.slugify.Slugify;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -114,15 +115,19 @@ public class ProgramServiceImpl implements ProgramService {
       String displayMode) {
 
     ImmutableSet.Builder<CiviFormError> errorsBuilder = ImmutableSet.builder();
+
     if (hasProgramNameCollision(adminName)) {
       errorsBuilder.add(CiviFormError.of("a program named " + adminName + " already exists"));
     }
+
     validateProgramText(errorsBuilder, "admin name", adminName);
     validateProgramText(errorsBuilder, "admin description", adminDescription);
     validateProgramText(errorsBuilder, "display name", defaultDisplayName);
     validateProgramText(errorsBuilder, "display description", defaultDisplayDescription);
     validateProgramText(errorsBuilder, "display mode", displayMode);
+
     ImmutableSet<CiviFormError> errors = errorsBuilder.build();
+
     if (!errors.isEmpty()) {
       return ErrorAnd.error(errors);
     }
@@ -135,6 +140,7 @@ public class ProgramServiceImpl implements ProgramService {
             defaultDisplayDescription,
             externalLink,
             displayMode);
+
     program.addVersion(versionRepository.getDraftVersion());
     return ErrorAnd.of(programRepository.insertProgramSync(program).getProgramDefinition());
   }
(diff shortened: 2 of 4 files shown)


server/app/services/program/ProgramServiceImpl.java
```diff
@@ -210,7 +216,17 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
...
-    return getActiveAndDraftPrograms().getProgramNames().contains(programName);
+    ImmutableSet<String> programNames = getActiveAndDraftPrograms().getProgramNames();
+
+    if (programNames.contains(programName)) {
+      return true;
+    }
+
+    Slugify slugifier = new Slugify();
+    ImmutableSet<String> programSlugs =
+        programNames.stream().map(slugifier::slugify).collect(ImmutableSet.toImmutableSet());
```
th174, line 227: (nit) Is there a difference between collecting as a set, and then calling contains, vs just using `Stream::anyMatch`? 

If not I think using `Stream::anyMatch` might be more clear.

th174 (dismissed):

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -210,7 +216,17 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
   }
 
   private boolean hasProgramNameCollision(String programName) {
-    return getActiveAndDraftPrograms().getProgramNames().contains(programName);
+    ImmutableSet<String> programNames = getActiveAndDraftPrograms().getProgramNames();
```
shanemc-goog, line 219: While you're not changing this part; how about including obsolete ones (EG all programs) for good measure?  It's technically a valid state and we could allow for archiving programs in the future.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -210,7 +216,17 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
...
+
+    if (programNames.contains(programName)) {
+      return true;
+    }
+
+    Slugify slugifier = new Slugify();
+    ImmutableSet<String> programSlugs =
+        programNames.stream().map(slugifier::slugify).collect(ImmutableSet.toImmutableSet());
+
+    return programSlugs.contains(slugifier.slugify(programName));
```
shanemc-goog, line 229: I think we can reduce/simplify this method to only this check.  Any two program names that collide above will collide here too.

shanemc-goog (approved):

shanemc-goog (changes requested):
didn't mean to approve, oops

server/test/services/program/ProgramServiceImplTest.java
```diff
@@ -177,6 +177,34 @@ public void createProgram_protectsAgainstProgramNameCollisions() {
         .containsExactly(CiviFormError.of("a program named name already exists"));
   }
 
+  @Test
+  public void createProgram_protectsAgainstProgramSlugCollisions() {
+    ps.createProgramDefinition(
+        "name  one",
```
shanemc-goog, line 183: The comment on "name one" below is good, but it needs a similar one here (since you read it first) or just a pre-amble for the test that it's using different names that sluggify the same.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -210,7 +216,17 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
...
 
   private boolean hasProgramNameCollision(String programName) {
-    return getActiveAndDraftPrograms().getProgramNames().contains(programName);
+    ImmutableSet<String> programNames = getActiveAndDraftPrograms().getProgramNames();
+
+    if (programNames.contains(programName)) {
+      return true;
+    }
+
+    Slugify slugifier = new Slugify();
```
jeanniefu, line 225: nit, optional: Consider adding a comment with example to explain what this is to people like me (who just learned what slugging is :)).

e.g. Something like

Compare slugified version of programName, which is what is used in the url. e.g. "my program" and "my  program" will both be slugified to "my-program".

jeanniefu (approved):

server/app/services/program/ProgramServiceImpl.java
shanemc-goog: something like

```
String newProgramSlug = slugifier.slugify(programName);

return getActiveAndDraftPrograms()
.getProgramNames()
.stream
.map(slugifier::slugify)
.anyMatch(newProgramSlug::equals);
```

server/app/services/program/ProgramServiceImpl.java
bion: done

server/app/services/program/ProgramServiceImpl.java
bion: done

server/app/services/program/ProgramServiceImpl.java
bion: done

server/test/services/program/ProgramServiceImplTest.java
bion: done

server/app/services/program/ProgramServiceImpl.java
bion: done

bion: you might notice that there's a slug column, and you might wonder "why not query that column instead of the name column", and my answer is: because the slug column is not guaranteed to be populated like the name column is and changing that is more than i'd like to do in this PR.

it brings me no joy to mention this, as it brought me no joy to discover it.

server/app/repository/ProgramRepository.java
```diff
@@ -66,6 +68,17 @@ public Program updateProgramSync(Program program) {
     return program;
   }
 
+  public ImmutableSet<String> getAllProgramNames() {
+    ImmutableSet.Builder<String> names = ImmutableSet.builder();
+    List<SqlRow> rows = database.sqlQuery("SELECT DISTINCT name FROM programs").findList();
```
shanemc-goog, line 73: We don't use raw sql a lot that I've seen, can we stick with the DAO unless it can't meet our need here?

shanemc-goog (commented):
Looks good! just the one discussion.

server/app/services/program/ProgramServiceImpl.java
shanemc-goog: Seems like a reasonable thing to ask Seattle to shore up, it could be done with a little manual work too I would hope.  Shall we ask them about it?

server/app/repository/ProgramRepository.java
bion: I'm not sure how to do this otherwise, everything I try wants to return a list of `Program`s

server/app/services/program/ProgramServiceImpl.java
bion: Sounds good to me!

server/app/repository/ProgramRepository.java
shanemc-goog: ah ok, I'd thought it should be easy enough but we can figure it out later.

shanemc-goog (approved):

server/app/repository/ProgramRepository.java
bion: this returns a `Set<Program>`

PR #2420, opened 2022-05-04
Fix version comparison logic in cloud/shared/bin/doctor.

jeanniefu:
This PR:
- Renames version_greater_than() to version_at_least() to better represent actual behavior.
- Fixes an out of bounds indexing error. Previously, version_at_least("2.3.5", "2.3") would have errored. We now treat the shorter string as if it had trailing zeros (e.g. "2.3.0").
- Fixes a comparison error if leading digits are the same. Previously, version_at_least("2.34.1", "2.36.0") would have returned false (which was the issue I was running into).


cloud/shared/bin/lib/check_test.py
```diff
@@ -3,16 +3,17 @@
...
-        self.assertTrue(check.version_greater_than("1.1", "1.1.3"))
-        self.assertFalse(check.version_greater_than("1.2", "1.1.3"))
-        self.assertFalse(check.version_greater_than("1.0.4", "1.1.3"))
+        self.assertTrue(check.version_at_least("1", "2"))
+        self.assertFalse(check.version_at_least("2", "1"))
+        self.assertTrue(check.version_at_least("1.1", "1.1"))
+        self.assertTrue(check.version_at_least("1.1", "1.2"))
+        self.assertTrue(check.version_at_least("1.1", "1.1.3"))
+        self.assertFalse(check.version_at_least("1.2", "1.1.3"))
+        self.assertTrue(check.version_at_least("1.0.4", "1.1.3"))
```
jeanniefu, line 15: A bit hard to tell from the GitHub diffing, but this line I switched from "assertFalse" to "assertTrue".

bion (approved):

cloud/shared/bin/lib/check_test.py
```diff
@@ -3,16 +3,17 @@
 from check import Check
 
 class TestCheck(unittest.TestCase):
-    def test_version_greater_than(self):
+    def test_version_at_least(self):
```
goldblatt, line 6: can we add a test for like 1.0.0 == 1

goldblatt (changes requested):

cloud/shared/bin/lib/check_test.py
jeanniefu: Done!

cloud/shared/bin/lib/check.py
```diff
@@ -45,11 +45,15 @@ def _get_cloud_provider(self):
     def _semver_string_to_int_array(self, semver_string):
         return [int(x) for x in semver_string.split(".")]
 
-    def version_greater_than(self, min_version, test_version):
+    def version_at_least(self, min_version, test_version):
         min_version, test_version = map(self._semver_string_to_int_array, [min_version, test_version])
 
         for i, min_int in enumerate(min_version):
-            if min_int > test_version[i]:
+            if i >= len(test_version) or min_int > test_version[i]:
```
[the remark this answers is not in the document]
jeanniefu: Thank you, good catch! Used another variable. This does some unnecessary extra comparisons (e.g. we could terminate early for (1.0.0, 1) case), but I think it is clearer this way.

clouser2 (approved):

goldblatt (approved):

PR #2424, opened 2022-05-04
Add a script for sbt to run in the unit test environment

kmcnellis:
### Description
Adds a script for sbt to run in the unit test environment, and changes `bin/run-test` to use it.

You can run this using `bin/sbt-test` to get the command line, or pass arguments like `bin/sbt-test testOnly <test name>`.

It will start the docker containers needed for the tests in the background, and keep them running so you can do multiple commands without re-loading.  

Run `bin/remove-test-env` to shut down the docker containers.

### Checklist
- Ran scripts to test

### DIFF
--- a/bin/sbt-test
+++ b/bin/sbt-test
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
-# DOC: Run the Java test suite in Docker.
+# DOC: Open an sbt CLI in the Unit Test environment container.
 
 source bin/lib.sh
 docker::set_project_name_unit_tests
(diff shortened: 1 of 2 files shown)


bin/sbt-test
```diff
@@ -0,0 +1,13 @@
+#! /usr/bin/env bash
+
+# DOC: Run the Java test suite in Docker.
```
bion, line 3: suggestion (blocking): fix doc string

bion (approved):

gwendolyngoetz (approved):

bin/sbt-test
kmcnellis: Done

kmcnellis: Done

gwendolyngoetz (approved):

PR #2429, opened 2022-05-04
Removes client-side validation for number, address, date, and currency question types

### LINKED ISSUE #2125: Date Parsing Exception When Saving Application
Splitting this off from issue #2014 

--

From @elisekalstad 

3/16 ORCA Opportunity program: "Oops…..and error occurred. This exception has been logged with id 7n33879ep"

clouser2:
### Description
This removes client-side validation for the above question types in favor of having consistent server-side validation. This work surfaced several issues where we were relying on client-side validation to prevent server-side 500s for invalid inputs (for date / number / currency questions).

When staging updates to perform, we typically try to call the relevant `AppplicantData.put*` methods. If "abc" was passed for a `NumberQuestion`, this previously would have thrown a server-side 500. Instead, we now propagate failed update state through and pass it to the question when retrieving validation errors. This felt a bit clunky, since the failed updates were generated at the controller level. I added logic in the base question implementation to guard against a specific question type not checking for this type of invalid input.

The other question type changes were fairly mechanical and involved no longer unconditionally setting a validation error message for use client-side.

The only remaining client-side validation we have is for enumerator questions and file uploads. In an attempt to minimize the scope of the PR, I'll address each of these separately. Why not incrementally do each of the question types from this PR? A given screen can have a variety of question types available on it, so doing them one-by-one would further introduce inconsistency for applicants while it was in progress. Non-enumerator / file upload questions currently must be on their own separate screen.

### Checklist

### Issue(s)
#1944; Fixes #2125;


server/app/services/applicant/question/CurrencyQuestion.java
```diff
@@ -37,12 +38,18 @@ public ImmutableList<Path> getAllPaths() {
 
   @Override
   protected ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> getValidationErrorsInternal() {
-    // TODO(#1944): Validate that the provided currency is a valid number. Presently,
-    // this is only implemented in client-side validation.
+    // When staging updates, the attempt to update ApplicantData would have failed to
+    // convert to a currency and been noted as a failed update. We check for that here.
+    if (applicantQuestion.getApplicantData().getFailedUpdates().containsKey(getCurrencyPath())) {
```
clouser2, line 43: Passing this through feels a bit clunky to me. However, I believe we'd like to consolidate validation-related code to these classes. Presently, these are constructed after the controller has "staged" updates to the relevant ApplicantData.

It may be worth pursuing a separate type of refactoring where the candidate raw updates are passed here rather than staging them from the controller. I haven't thought that through enough yet to have confidence one way or another.

Assuming the current approach isn't too distasteful, I'd propose handling a refactoring such as that separately.

server/app/views/components/FieldWithLabel.java
```diff
@@ -216,16 +210,6 @@ public FieldWithLabel setValue(OptionalLong value) {
     return this;
   }
 
-  public FieldWithLabel setValue(Currency value) {
```
clouser2, line 219: This felt unnecessary as there's only one caller and it prevented us from round-tripping the invalid state (something that isn't convertable to a currency) that may have been provided by the applicant.

server/conf/messages
```diff
@@ -216,7 +219,6 @@ validation.streetRequired=Please enter valid street name and number.
 validation.cityRequired=Please enter city.
 validation.currencyMisformatted=Currency must be one of the following formats: 1000 1,000 1000.30 1,000.30
 validation.stateRequired=Please enter state.
-validation.zipcodeRequired=Please enter valid ZIP code.
```
clouser2, line 219: This was unnecessary since the current regexp for matching a zip-code was a more specific error message anyway.

server/app/services/applicant/ApplicantData.java
```diff
@@ -114,4 +118,22 @@ public void setUserName(
       putString(WellKnownPaths.APPLICANT_LAST_NAME, lastName);
     }
   }
+
+  @Override
+  public String asJsonString() {
+    if (!getFailedUpdates().isEmpty()) {
+      throw new IllegalStateException(
+          "data cannot be serialized since there " + "were failed updates");
```
bion, line 126: nit: 
```suggestion
          "data cannot be serialized since there were failed updates");
```

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -181,16 +182,18 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
               Block blockBeforeUpdate = maybeBlockBeforeUpdate.get();
 
               UpdateMetadata updateMetadata = UpdateMetadata.create(programId, clock.millis());
+              ImmutableMap<Path, String> failedUpdates = ImmutableMap.of();
```
bion, line 185: question (blocking): does this need to be initialized or only declared?

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -506,17 +513,32 @@ private void stageNormalUpdates(
       } else {
         switch (type) {
           case CURRENCY_CENTS:
-            applicantData.putCurrencyDollars(currentPath, update.value());
+            try {
+              applicantData.putCurrencyDollars(currentPath, update.value());
+            } catch (IllegalArgumentException e) {
+              applicantData.maybeDelete(currentPath);
```
bion, line 519: question (blocking): why is maybeDelete necessary here?

server/app/services/applicant/Currency.java
```diff
@@ -20,8 +20,7 @@
  */
 public class Currency {
 
-  // Currency validation regexs.  Note: there are frontend versions that need to stay in sync in
-  // app/assets/javascripts/validation.ts
+  // Currency validation regexs.
```
bion, line 23: good catch updating this comment

server/app/services/applicant/question/CurrencyQuestion.java
bion: That sounds good to me, I agree it's tricky to say which approach will be cleaner at moment.

server/app/services/applicant/question/NumberQuestion.java
```diff
@@ -35,17 +35,23 @@ public ImmutableList<Path> getAllPaths() {
...
-    // to the correct currency.
-    return ImmutableMap.of(applicantQuestion.getContextualizedPath(), validateNumber());
+    return ImmutableMap.of(getNumberPath(), validateNumber());
   }
 
   private ImmutableSet<ValidationErrorMessage> validateNumber() {
+    Optional<Long> numberValue = getNumberValue();
+    // When staging updates, the attempt to update ApplicantData would have failed to
+    // convert to a number and been noted as a failed update. We check for that here.
+    if (applicantQuestion.getApplicantData().getFailedUpdates().containsKey(getNumberPath())
```
bion, line 45: suggestion (blocking): may as well encapsulate this logic for better law of demeter adherence:
```java
class ApplicantData {
  public boolean updateDidFailAt(Path path) {
    return getFailedUpdates().containsKey(path);
  }
}
```

bion (changes requested):

server/app/services/applicant/ApplicantData.java
clouser2: Strange. I imagine this was some form of me doing line splitting and bin/fmt deciding I shouldn't have done it that way.

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: If the applicant were editing a previously answered question, the ApplicantData would have the previous value. When invalid data was provided, we store that input in the failedUpdates map. I thought it might be confusing to have both the previous value and the invalid value present on ApplicantData.

I don't feel strongly about that though, so could remove the call to maybeDelete. Do either of you have a preference either way?

server/app/services/applicant/question/NumberQuestion.java
clouser2: Done.

server/app/services/applicant/ApplicantServiceImpl.java
bion: It gets a little trickier to think about in other scenarios. E.g. if we introduce validations that rely on looking at multiple pieces of data, clearing the value results in different behavior than leaving the invalid value. This gets even trickier when we consider that the bad update may be attempting to change a value that is already set and previously valid -- deleting it creates yet a different validation state if we have those kinds of cross-cutting validations.

I'm not sure what the best approach is, I'm very slightly leaning towards leaving the invalid update in place since it seems simplest for now, but if you feel different I'm totally open to other thoughts.

bion (approved):
approved, up to you how to resolve the conversation on clearing invalid applicantdata values

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: Those are all reasonable concerns. I'll leave the old value in place for now. It's also worth calling out that I added logic to ApplicantData.toJsonString to ensure that we don't allow persisting invalid updates as a bit of defensive coding.

bion: Nice one, thanks for that!

server/app/services/applicant/ApplicantData.java
```diff
@@ -114,4 +118,25 @@ public void setUserName(
...
+
+  @Override
+  public String asJsonString() {
+    if (!getFailedUpdates().isEmpty()) {
+      throw new IllegalStateException("data cannot be serialized since there were failed updates");
+    }
+    return super.asJsonString();
+  }
+
+  public void setFailedUpdates(ImmutableMap<Path, String> updates) {
```
jeanniefu, line 130: nit: Add a comment? It may not be clear to others that the values in the map are the inputs that caused the "failed updates"

browser-test/src/currency.test.ts
```diff
@@ -60,15 +60,15 @@ describe('currency applicant flow', () => {
       await selectApplicantLanguage(pageObject, 'English')
 
       await applicantQuestions.applyProgram(programName)
-      const error = await pageObject.$('.cf-currency-value-error')
-      expect(await error.isHidden()).toEqual(true)
+      const currencyError = '.cf-currency-value-error'
+      expect(await pageObject.isHidden(currencyError)).toEqual(true)
```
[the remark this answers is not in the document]
clouser2: That's a great catch!

Apparently isHidden will return true if the selector doesn't match anything (https://playwright.dev/docs/api/class-page#page-is-hidden). 

I refactored to assert that the selector returns no elements instead here.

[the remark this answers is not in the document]
clouser2: Disregard my previous comment. We always add the error div and hide it based on whether it contains any validation errors (https://github.com/seattle-uat/civiform/blob/137b5980449567daf2444b345ef7691254eda21f/server/app/views/components/FieldWithLabel.java#L348). As such, we'd still want to assert it's hidden here (unless we refactored FieldWithLabel).

I've added a comment to this effect.

server/app/services/applicant/ApplicantData.java
clouser2: Done!

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -360,17 +363,19 @@ public ImmutableList<Application> getAllApplications() {
    *
    * @throws PathNotInBlockException if there are updates for questions that aren't in the block.
    * @throws UnsupportedScalarTypeException if there are updates for unsupported scalar types.
+   * @return any failures to update the applicant data, containing the desired {@link Path} as well
```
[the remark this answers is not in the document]
clouser2: Done!

server/app/services/applicant/question/QuestionImpl.java
```diff
@@ -36,17 +37,37 @@ public QuestionImpl(ApplicantQuestion applicantQuestion) {
...
     // assumes that calling isEmpty on the map means that there are no errors.
-    return ImmutableMap.<Path, ImmutableSet<ValidationErrorMessage>>builder()
-        .putAll(Maps.filterEntries(getValidationErrorsInternal(), e -> !e.getValue().isEmpty()))
-        .build();
+    ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> result =
+        ImmutableMap.<Path, ImmutableSet<ValidationErrorMessage>>builder()
+            .putAll(Maps.filterEntries(getValidationErrorsInternal(), e -> !e.getValue().isEmpty()))
+            .build();
+    // We shouldn't have an empty error map if we failed to convert some of the input. If there
+    // aren't
```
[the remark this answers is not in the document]
clouser2: Done. The auto-formatter got me again.

server/conf/messages
```diff
@@ -42,6 +42,9 @@ header.userName=Logged in as {0}
 # Error message to answer a required question
 validation.isRequired=This question is required.
 
+# Validation error that is shown when the user input couldn't be converted for any reason.
+validation.invalidInput=Please enter valid input.
```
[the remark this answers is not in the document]
clouser2: Good point. This is the fallback file, so en-US would fallback to this. It looks like we have the other messages duplicated in here, so I've added them as well. Thanks for flagging.

browser-test/src/currency.test.ts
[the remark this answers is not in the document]
jeanniefu: not blocking: Not for this PR, but for future work - what do you think about us refactoring FieldWithLabel, so that we don't create the element at all? It seems a bit weird if we create the element and never unhide it.

jeanniefu (approved):

clouser2:
RE refactoring of FieldLabel. That seems a reasonable direction to move. There are several other usages of `FieldWithLabel` outside of this, so we'd need to make sure we weren't breaking any expected functionality there.

PR #2430, opened 2022-05-04
UI for creating API keys

bion:
- [x] admin can fill out and submit a form to create a new API key
- [x] form shows validation errors when needed
- [x] credential viewing page after key is created
- [x] browser test for happy path
- [x] handle ProgramNotFoundException with new client-error handler

part of https://github.com/seattle-uat/civiform/issues/1744

Screenshot from 2022-05-10 09-18-21
Screenshot from 2022-05-10 10-06-15


server/app/views/admin/apikeys/ApiKeyCredentialsView.java
```diff
@@ -0,0 +1,50 @@
...
+    ContainerTag contentDiv =
+        div()
+            .withClasses(Styles.PX_20)
+            .with(
+                h1(title).withClasses(Styles.MY_4),
+                h2("Credentials"),
+                p(
+                    "Please copy your API key and it store it somewhere secure. This is your only"
+                        + " opportunity to copy the secret from CiviForm, if you "
+                        + "refresh the page or navigate away you will not be able to recover the"
```
shanemc-goog, line 41: Consistently have the continuation space character at the start or end of the different segments.

server/app/views/admin/apikeys/ApiKeyCredentialsView.java
```diff
@@ -0,0 +1,50 @@
...
+    String title = "Created API key: " + apiKey.getName();
+
+    ContainerTag contentDiv =
+        div()
+            .withClasses(Styles.PX_20)
+            .with(
+                h1(title).withClasses(Styles.MY_4),
+                h2("Credentials"),
+                p(
+                    "Please copy your API key and it store it somewhere secure. This is your only"
```
shanemc-goog, line 39: suggestion: It's probably cleaner to have this text as a class constant so that it doesn't need to wrap so much or divide up the page construction.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+                        FieldWithLabel.input()
+                            .setFieldName(ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                            .setId(ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                            .setLabelText("API key name"),
+                        dynamicForm,
+                        ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                    .getContainer(),
+                h2("Expiration date"),
+                p(
+                    "Specify a date when this API key will no longer be valid. The expiration date"
```
shanemc-goog, line 66: likewise to the other html page, making this a const in the file will likely aid readability all around.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+                        + " allows only the IP 8.8.8.8 to use the API key. All IP addresses can be"
+                        + " allowed, but this is strongly discouraged for security reasons because"
+                        + " it would allow any computer connected to the internet to use the key"
+                        + " if they obtain it. All IP addresses can be allowed with a value of"
+                        + " 0.0.0.0/0."),
+                setStateIfPresent(
+                        FieldWithLabel.input()
+                            .setFieldName(ApiKeyService.FORM_FIELD_NAME_SUBNET)
+                            .setId(ApiKeyService.FORM_FIELD_NAME_SUBNET)
+                            .setLabelText("API key name"),
```
shanemc-goog: I think "key name" needs updating for this field?  I get lost in the page construction so I might be off.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,84 @@
...
+    try {
+      result = apiKeyService.createApiKey(form, profile.get());
+    } catch (ProgramNotFoundException e) {
+      return notFound(e.toString());
+    }
+
+    if (!result.isSuccessful()) {
+      return badRequest(
+          newOneView.render(
+              request, programService.getAllProgramNames(), Optional.of(result.getForm())));
```
shanemc-goog, line 79: I don't believe getAllProgramsNames filters for program version, meaning it will show a Draft I think.  Probably best to explicitly get Active programs.

shanemc-goog (changes requested):

server/app/views/admin/apikeys/ApiKeyNewOneView.java
bion: done

server/app/views/admin/apikeys/ApiKeyCredentialsView.java
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
bion: done

server/app/views/admin/apikeys/ApiKeyCredentialsView.java
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+                    .getContainer(),
+                h2("Allowed IP addresses"),
+                p(
+                    "Specify a subnet using CIDR notation that is allowed to call the API. The"
+                        + " subnet may be edited after the API key is created. A single IP address"
+                        + " can be specified with a mask of /32. For example, \"8.8.8.8/32\""
+                        + " allows only the IP 8.8.8.8 to use the API key. All IP addresses can be"
+                        + " allowed, but this is strongly discouraged for security reasons because"
+                        + " it would allow any computer connected to the internet to use the key"
+                        + " if they obtain it. All IP addresses can be allowed with a value of"
```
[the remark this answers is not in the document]
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+                        FieldWithLabel.date()
+                            .setFieldName(ApiKeyService.FORM_FIELD_NAME_EXPIRATION)
+                            .setId(ApiKeyService.FORM_FIELD_NAME_EXPIRATION)
+                            .setLabelText("Expiration date"),
+                        dynamicForm,
+                        ApiKeyService.FORM_FIELD_NAME_EXPIRATION)
+                    .getContainer(),
+                h2("Allowed IP addresses"),
+                p(
+                    "Specify a subnet using CIDR notation that is allowed to call the API. The"
```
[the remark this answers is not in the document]
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+                            .setFieldName(ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                            .setId(ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                            .setLabelText("API key name"),
+                        dynamicForm,
+                        ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
+                    .getContainer(),
+                h2("Expiration date"),
+                p(
+                    "Specify a date when this API key will no longer be valid. The expiration date"
+                        + " may be edited after the API is created. It is strongly recommended to"
```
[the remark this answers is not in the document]
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,142 @@
...
+    String title = "Create a new API key";
+
+    ContainerTag formTag =
+        form()
+            .withMethod("POST")
+            .with(
+                makeCsrfTokenInputTag(request),
+                setStateIfPresent(
+                        FieldWithLabel.input()
+                            .setFieldName(ApiKeyService.FORM_FIELD_NAME_KEY_NAME)
```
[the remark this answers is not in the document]
bion: made it more consistent

server/app/views/admin/apikeys/ApiKeyNewOneView.java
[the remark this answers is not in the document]
shanemc-goog: don't see a change for this or the other ones you commented on, did you push the commit?

browser-test/src/support/index.ts
```diff
@@ -8,11 +9,13 @@ export { AdminTIGroups } from './admin_ti_groups'
 export { ApplicantQuestions } from './applicant_questions'
 export { clickAndWaitForModal, waitForPageJsLoad } from './wait'
 
-const {
-  BASE_URL = 'http://civiform:9000',
-  TEST_USER_LOGIN = '',
-  TEST_USER_PASSWORD = '',
-} = process.env
+import {
```
shanemc-goog, line 12: just curious: why this change?

server/app/annotations/BindingAnnotations.java
```diff
@@ -18,4 +19,12 @@ public final class BindingAnnotations {
...
   @Retention(RUNTIME)
   public @interface Now {}
+
+  /**
+   * Holds the {@link play.i18n.Messages} object associated with the "en-US" {@link play.i18n.Lang}.
+   */
+  @Qualifier
+  @Target({FIELD, PARAMETER, METHOD})
+  @Retention(RUNTIME)
+  public @interface EnUs {}
```
shanemc-goog, line 29: Suggestion: qualify this as a language.  EnUsLang maybe.

EnUs is on the edge of looking random if you just come across it, and if say we did portuguese instead @Pt would be incredibly unreadable on a method.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,84 @@
...
+    if (!profile.isPresent()) {
+      throw new RuntimeException("Unable to resolve profile.");
+    }
+
+    DynamicForm form = formFactory.form().bindFromRequest(request);
+
+    ApiKeyCreationResult result;
+    try {
+      result = apiKeyService.createApiKey(form, profile.get());
+    } catch (ProgramNotFoundException e) {
```
shanemc-goog, line 72: unrelated: but re the ErrorHandler discussion, this could be handled there and converted to a RuntimeException.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,84 @@
...
+      return notFound(e.toString());
+    }
+
+    if (!result.isSuccessful()) {
```
shanemc-goog, line 76: suggestion: Flip the conditions, positive assertions are generally more readable.

server/conf/routes
```diff
@@ -85,6 +85,11 @@ GET     /admin/programs/:programId/applications/:applicationId            contro
 GET     /admin/programs/:programId/applications/:applicationId/download   controllers.admin.AdminApplicationController.download(request: Request, programId: Long, applicationId: Long)
 GET     /admin/demographics                                               controllers.admin.AdminApplicationController.downloadDemographics()
 
+# Admin API key management
+GET     /admin/apiKeys                              controllers.admin.AdminApiKeysController.index()
+GET     /admin/apiKeys/new                          controllers.admin.AdminApiKeysController.newOne(request: Request)
```
shanemc-goog, line 90: question: newOne and create don't read too differently from me, is it normal in REST to not have the phrasing of them be more indicative of what you're getting? or does Get/Post largely make that distinction.  EG a GET wouldn't have side effects so newOne can be inferred to be a page that facilitates making a new one.

server/app/controllers/admin/AdminApiKeysController.java
bion: done

browser-test/src/support/index.ts
bion: so that BASE_URL could be use by other files in `src/support` without creating a circular dependency

server/app/controllers/admin/AdminApiKeysController.java
bion: done

server/app/annotations/BindingAnnotations.java
bion: done

server/app/controllers/admin/AdminApiKeysController.java
bion: yessss nice i'm gonna do it

bion: done

server/app/views/admin/apikeys/ApiKeyCredentialsView.java
```diff
@@ -21,6 +21,12 @@
 public final class ApiKeyCredentialsView extends BaseHtmlView {
   private final AdminLayout layout;
 
+  private static final String CREDENTIALS_DESCRIPTION =
+      "Please copy your API key and it store it somewhere secure. This is your only"
+          + " opportunity to copy the secret from CiviForm, if you "
+          + "refresh the page or navigate away you will not be able to recover the"
```
shanemc-goog, line 27: Move the trailing space before this to a leading space here.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -32,8 +35,33 @@ public final class ApiKeyNewOneView extends BaseHtmlView {
...
+          + " recommended. Once a year is acceptable in most situations.";
+  private static final DomContent[] SUBNET_DESCRIPTION = {
+    text("Specify a subnet using "),
+    new LinkElement()
+        .setText("CIDR notation")
+        .setHref("https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing")
+        .opensInNewTab()
+        .asAnchorText(),
+    text(
+        " that is allowed to call the API. The"
```
shanemc-goog, line 53: I think some of the text lines were deleted accidentally

server/app/services/apikey/ApiKeyService.java
```diff
@@ -206,7 +205,10 @@ private ApiKeyGrants resolveGrants(DynamicForm form) throws ProgramNotFoundExcep
       String programSlug = matcher.group(1);
 
       if (!programSlugs.contains(programSlug)) {
-        throw new ProgramNotFoundException(programSlug);
+        // Making the root exception cause a ProgramNotFoundException will cause
+        // controllers.ErrorHandler to handle it
+        // as a client-error, rather than server-error.
+        throw new RuntimeException(new ProgramNotFoundException(programSlug));
```
shanemc-goog, line 211: discussion: This is fine, though subsequently, I think it'd be good/fine to make ProgramNotFoundException a Runtime Exception.  If code really needs to know it can do a hasProgramId() kinda check, but I'm guessing most just catch and wrap this as notFound without doing any other recovery or reporting.

shanemc-goog (changes requested):

server/app/views/admin/apikeys/ApiKeyNewOneView.java
bion: looks right to me?

server/app/services/apikey/ApiKeyService.java
bion: Yeah, to make the change we'll need to take a look at all the places `ProgramNotFoundException` is being used and see if any of them are using it for control flow that's fundamentally different from what the default behavior with the new pattern does. If so we should probably leave it as a checked exception.

server/app/views/admin/apikeys/ApiKeyCredentialsView.java
bion: done

server/app/views/admin/apikeys/ApiKeyNewOneView.java
shanemc-goog: yeah, I'd deleted. I didn't see the construction just above.  But GH is duplicating my comments today.... :/

shanemc-goog (approved):

server/app/controllers/ErrorHandler.java
```diff
@@ -55,11 +60,12 @@ public CompletionStage<Result> onServerError(RequestHeader request, Throwable ex
...
    * our application exception. Anecdotally it's 2 levels down.
    */
-  static Optional<Throwable> findThrowableByType(Throwable exception, Class<?> search) {
+  static Optional<Throwable> findThrowableByTypes(
+      Throwable exception, ImmutableSet<Class<?>> search) {
     Optional<Throwable> root = Optional.of(exception);
     // Search a couple causes deep for the desired type.
     for (int i = 0; i < 5 && root.isPresent(); ++i) {
-      if (search.isInstance(root.get())) {
+      if (search.contains(root.get().getClass())) {
```
clouser2, line 68: Not-blocking: Previously, we'd match subclasses of the provided "search" parameter, whereas now it has to be an exact match.

I'd suggest documenting that and marking the two specific error types that we currently handle as "final" to prevent us trying to take any implicit dependency on that behavior.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,77 @@
...
+    this.apiKeyService = checkNotNull(apiKeyService);
+    this.newOneView = checkNotNull(newOneView);
+    this.apiKeyCredentialsView = checkNotNull(apiKeyCredentialsView);
+    this.programService = checkNotNull(programService);
+    this.formFactory = checkNotNull(formFactory);
+    this.profileUtils = checkNotNull(profileUtils);
+  }
+
+  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  public Result index() {
```
clouser2, line 49: Not-blocking: Consider removing this and reintroducing it when it's implemented. Alternatively, leave a TODO? Or ignore entirely, up to you.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,77 @@
...
+    return ok();
+  }
+
+  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
```
clouser2, line 53: Not-blocking: Out of curiosity, if all handlers in a controller require the same authorizer, is it possible to annotate at the class level instead? This would help prevent the case where a new handler is added to an existing controller and forgot to add an authorizer. It would potentially also reduce code duplication.

server/app/controllers/admin/AdminApiKeysController.java
```diff
@@ -0,0 +1,77 @@
...
+    Optional<CiviFormProfile> profile = profileUtils.currentUserProfile(request);
+
+    if (!profile.isPresent()) {
+      throw new RuntimeException("Unable to resolve profile.");
+    }
+
+    DynamicForm form = formFactory.form().bindFromRequest(request);
+    ApiKeyCreationResult result = apiKeyService.createApiKey(form, profile.get());
+
+    if (result.isSuccessful()) {
```
clouser2, line 69: Not-blocking: Complete style nit. Consider putting the error handling in the if-block and putting the "happy" case as the last statement. This is mostly a style preference that reading through the code ends in the case we expect. A perhaps weaker argument is that there's another indented block in this code which throws an exception, so it's perhaps slightly less readable to have the next indented block actually return the success case. My preference is pretty weak here, so feel free to discard if you'd like.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,158 @@
...
+            + " 0.0.0.0/0, though this is not recommended.")
+  };
+
+  @Inject
+  public ApiKeyNewOneView(AdminLayout layout, @EnUsLang Messages enUsMessages) {
+    this.layout = checkNotNull(layout);
+    this.enUsMessages = checkNotNull(enUsMessages);
+  }
+
+  public Content render(Request request, ImmutableSet<String> programNames) {
```
clouser2, line 69: Not-blocking: Since there's only one caller of this, consider removing the method overload and having the controller call render(request, programNames, Optional.empty()).

server/app/views/admin/apikeys/ApiKeyNewOneView.java
```diff
@@ -0,0 +1,158 @@
...
+
+    formTag.with(h2("Allowed programs"), p("Select the programs this key grants read access to."));
+
+    for (String name : programNames.stream().sorted().collect(ImmutableList.toImmutableList())) {
+      formTag.with(
+          FieldWithLabel.checkbox()
+              .setFieldName(programReadGrantFieldName(name))
+              .setLabelText(name)
+              .setId(slugifier.slugify(name))
+              .setValue("true")
```
clouser2, line 121: Does this imply that new keys grant access to all current programs by default? I'd consider not selecting anything by default and having a "select all" option. My rationale for this is that it aligns with the principle-of-least-privilege.

This point / distinction needn't block the larger review at all though.

server/app/views/components/FieldWithLabel.java
```diff
@@ -223,7 +224,20 @@ public FieldWithLabel setScreenReaderText(String screenReaderText) {
...
     this.messages = messages;
-    this.fieldErrors = errors;
+    this.fieldErrors =
+        errors.stream()
+            .map(
+                (ValidationErrorMessage vem) ->
+                    new ValidationError(vem.key().getKeyName(), vem.key().getKeyName(), vem.args()))
+            .collect(ImmutableSet.toImmutableSet());
+
+    return this;
```
clouser2, line 234: Not-blocking: Nit. Since this method overload will hopefully remain temporary, what about having it call the more preferred method with the converted errors.

e.g.

public FieldWithLabel setFieldErrors(Messages messages, ImmutableSet<ValidationErrorMessage> errors) {
  return setFieldErrors(messages, errors.stream().....);
}

clouser2 (approved):

server/app/controllers/admin/AdminApiKeysController.java
bion: The docs don't seem to suggest that you can: https://github.com/pac4j/play-pac4j/wiki/Apply-security

I tried moving the annotation to the class and while the app did run, I got some strange behavior when trying to access the secured routes as an authorized user.

server/app/controllers/admin/AdminApiKeysController.java
bion: hahah that was actually its original structure, @shanemc-goog requested a change to avoid having `if (!`. i don't have a strong opinion either.

server/app/controllers/admin/AdminApiKeysController.java
bion: it's already implemented in the next PR, i'll leave it as-is for now if that's alright

browser-test/src/admin_api_keys.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  it('Creating a new API key displays its credentials', async () => {
+    const { browser, page } = await startSession()
+
+    await loginAsAdmin(page)
+    const adminApiKeys = new AdminApiKeys(page)
+    const adminPrograms = new AdminPrograms(page)
+
+    const programName = 'API using program'
+    const programDescription = 'This program uses the API.'
+    await adminPrograms.addProgram(programName, programDescription, '', false)
```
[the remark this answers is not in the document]
bion: agree that's a good change, going to hold off as it's out of scope, as you mention

browser-test/src/admin_api_keys.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    await adminPrograms.publishAllPrograms()
+
+    const credentials = await adminApiKeys.createApiKey(
```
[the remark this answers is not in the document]
bion: done

browser-test/src/admin_api_keys.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      'Test API key',
+      '2022-01-31',
+      "8.8.8.8/32",
+      ['api-using-program']
+    )
+
+    expect(typeof credentials).toEqual("string")
```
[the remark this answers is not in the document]
bion: yeah, planning to do this once there's an API endpoint to call

server/app/controllers/ErrorHandler.java
```diff
@@ -40,7 +42,10 @@ public ErrorHandler(
   public CompletionStage<Result> onServerError(RequestHeader request, Throwable exception) {
     // Exceptions that reach here will generate 500s. Here we convert certain ones to different user
     // visible states. Note: there are methods on the parent that handle dev and prod separately.
-    Optional<Throwable> match = findThrowableByType(exception, NotChangeableException.class);
+    Optional<Throwable> match =
+        findThrowableByTypes(
+            exception,
+            ImmutableSet.of(NotChangeableException.class, ProgramNotFoundException.class));
```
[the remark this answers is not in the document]
bion: done

server/app/controllers/ErrorHandler.java
bion: good call! done

server/app/controllers/admin/AdminApiKeysController.java
shanemc-goog: hah lol yeah.  Happy either way, I can see brett's case too.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
bion: The opposite -- the browser only sends form data for checkbox inputs that are checked. So the default is to grant access to no programs.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
bion: There are two callers: the `newOne` method for the initial render, and the `create` method when a submission fails validation -- both `render` method versions get used.

server/app/views/components/FieldWithLabel.java
bion: Unfortunately `FieldWithLabel setFieldErrors(Messages messages, ImmutableSet<ValidationError> errors)` and `setFieldErrors(Messages messages, ImmutableSet<ValidationErrorMessage> errors)` have the same erasure so we can't do that overload. It is temporary though since we'll do the broader refactor after this PR goes in.

browser-test/src/admin_api_keys.test.ts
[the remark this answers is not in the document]
clouser2: Works for me.

server/app/controllers/admin/AdminApiKeysController.java
clouser2: No worries. Thanks for digging.

server/app/views/admin/apikeys/ApiKeyNewOneView.java
clouser2: Still not blocking, but I also might be misunderstanding.

I see that FieldWithLabel.checkbox() sets the input type to checkbox and only sets the "checked" attribute if `setChecked` is called (which it isn't here). The question I have is what setting the value to "true" does here? Can that be removed to avoid confusion?

bion: Ah, I've always set checkbox values to "true" out of habit. [Looks like the value is `on` by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) so maybe I've done that over the years because "true" fit better with the boolean concept in my head?

Anyways, thanks for asking! It looks like setting this unnecessary. I'll remove it in my next PR.

PR #2433, opened 2022-05-04
Make sure all docker containers get refreshed

kmcnellis:
### Description
The browser-test and formatter images don't get auto-refreshed.  Add them to the pull-image command and use the pull-image command in their respective scripts.

### DIFF
--- a/bin/refresh-styles
+++ b/bin/refresh-styles
@@ -4,11 +4,13 @@
 
 source bin/lib.sh
 
+bin/pull-image
+
 docker run -it --rm \
   -v /var/run/docker.sock:/var/run/docker.sock \
   -v $(pwd)/server:/usr/src/server \
   --entrypoint npx \
-  civiform/civiform-dev:latest \
+  civiform-dev \
   tailwindcss build \
   -i ./app/assets/stylesheets/styles.css \
   -o ./public/stylesheets/tailwind.css
--- a/browser-test/bin/fmt
+++ b/browser-test/bin/fmt
@@ -4,8 +4,10 @@ pushd $(git rev-parse --show-toplevel)/browser-test
 
 CMD="cd /code; npx prettier --write --config /.prettierrc.js --ignore-path /.prettierignore /code"
 
+bin/pull-image
+
 docker run --rm -it \
   -v "$(pwd)/src:/code" \
-  docker.io/civiform/formatter:latest sh -c "$CMD"
+  civiform/formatter:latest sh -c "$CMD"
 
 popd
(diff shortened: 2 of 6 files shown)


bion (approved):

bin/refresh-styles
```diff
@@ -4,11 +4,13 @@
...
 source bin/lib.sh
 
+bin/pull-image
+
 docker run -it --rm \
   -v /var/run/docker.sock:/var/run/docker.sock \
   -v $(pwd)/server:/usr/src/server \
   --entrypoint npx \
-  civiform/civiform-dev:latest \
+  civiform-dev \
```
shanemc-goog, line 13: Can you help me understand this better?

I don't quite understand what the left is using or what this is now using.

browser-test/bin/fmt
```diff
@@ -4,8 +4,10 @@ pushd $(git rev-parse --show-toplevel)/browser-test
 
 CMD="cd /code; npx prettier --write --config /.prettierrc.js --ignore-path /.prettierignore /code"
 
+bin/pull-image
+
 docker run --rm -it \
   -v "$(pwd)/src:/code" \
-  docker.io/civiform/formatter:latest sh -c "$CMD"
+  civiform/formatter:latest sh -c "$CMD"
```
shanemc-goog, line 11: likewise here. why is docker.io not good?

shanemc-goog (approved):

browser-test/bin/fmt
bion: It's the default, IMO it made more sense to include it when we had shell scripts that pushed to both docker hub and AWS ECR just to have extra clarity. Now that everything in the main repo is only interacting with docker hub it's inconsistent to have it a few places and less needed for disambiguation.

bin/refresh-styles
kmcnellis: These are both tags.   "civiform/civiform-dev:latest" is the default tag, that lets it know to go to docker-hub and download the image if it's not cached locally.   In bin/pull-image, we add an additional tag of "civiform-dev" to the  "civiform/civiform-dev:latest" image. 

This is just a short-cut to make this script align better with the others, as well as allow someone to re-build locally and tag the build image.

browser-test/bin/fmt
kmcnellis: This is just removing extra specificity that we don't need, and making it act like our other scripts.  The "docker.io" is a repository name,  so it just tells the command to go to docker.io to get the image if it's not local.  But since it's also the default repository, docker already knows to look there.

kmcnellis: Also, since we are adding it to the pull-image step, it's already downloaded

PR #2435, opened 2022-05-05
Update JSON exporter to match the design

bion:
Update JSON exporter to match the [agreed-upon design](https://docs.google.com/document/d/1g2XkMhDe6EaVGQabEes4KoDWpzL9F4S5pl2fUXoilSc/edit).

This PR changes the exported format of the currency and date question types to match the design.

The remaining piece is adding the `updated_at` field for each exported question answer. Getting that done will be a bit more complicated, so in the interest of getting this out sooner than later so that export consumers don't start relying on the old formats I'm planning to take care of that in a follow-up PR.


browser-test/src/application_download.test.ts
```diff
@@ -127,14 +127,28 @@ describe('normal application flow', () => {
...
+    ).toBeNull()
+    expect(
+      postEditJSonContent[0].application.namecsvdownload.last_name
+    ).toEqual('Guest')
+    expect(postEditJSonContent[0].application.csvdate.date).toEqual(
+      '1990-01-01'
+    )
+    expect(postEditJSonContent[0].application.numbercsvdownload.number).toEqual(
+      1600
+    )
```
bion, line 151: autoformatter change

server/app/services/export/JsonExporter.java
```diff
@@ -127,6 +136,20 @@ private CfJsonDocumentContext buildJsonApplication(
               jsonApplication.putNull(path);
             }
 
+            break;
+          }
+        case DATE:
```
jeanniefu, line 141: not blocking, for my own information: Does this mean we didn't export dates before?

jeanniefu (approved):

server/app/services/export/JsonExporter.java
bion: No, just that they were exported using the default logic which uses the same formatting as the application review page. The cases in this switch other than the default override that logic to do formatting specific to the JSON export.

browser-test/src/application_download.test.ts
kmcnellis: Maybe if the variable name was shorter, it wouldn't have to fold these into multiple lines? 🤷‍♀️

kmcnellis (approved):

PR #2441, opened 2022-05-05
Adding previous button to File Upload Page

### LINKED ISSUE #2282: Missing "Previous" Button on File Upload Pages
**Describe the bug**
There is no "previous" button on file upload pages

**To Reproduce**
1. Go to a form that requires documents to be uploaded
2. Navigate to the file upload page
3. See that there is no "previous" button

**Screenshots**
Attached

Missing back button

swatkat1:
### Description
Added a Previous Button to the File Upload page.

### Issue(s)
Fixes #2282 

Screenshot:
Adding Previous to FileUpload


clouser2 (approved):
Looks great! Is it worth it to add a unit test or a browser test for this?

swatkat1:
for unit tests - I do not see any specific tests for file uploads (but 'previous' button functionality is well tested in unit tests. This code also follows the same.)
For browser tests - File uploads is tested. Not sure that adding previous button tests actually adds value here as the functionality is same as 'previous' button in other pages.

clouser2:
> for unit tests - I do not see any specific tests for file uploads (but 'previous' button functionality is well tested in unit tests. This code also follows the same.) For browser tests - File uploads is tested. Not sure that adding previous button tests actually adds value here as the functionality is same as 'previous' button in other pages.

Your logic makes sense to me and agree that a browser test would be low value. Thanks for the explanation.

server/app/views/FileUploadViewStrategy.java
```diff
@@ -152,6 +152,26 @@ protected Tag renderUploadButton(Params params) {
...
+      redirectUrl =
+          routes.ApplicantProgramReviewController.preview(params.applicantId(), params.programId())
+              .url();
+    }
+    return a().attr(HREF, redirectUrl)
+        .withText(params.messages().at(MessageKey.BUTTON_PREVIOUS_SCREEN.getKeyName()))
+        .withClasses(ApplicantStyles.BUTTON_BLOCK_PREVIOUS)
+        .withId("cf-block-previous");
+  }
+
```
bion, line 174: suggestion (blocking): looks like this code was copied without modification from https://github.com/seattle-uat/civiform/blob/main/server/app/views/applicant/ApplicantProgramBlockEditView.java#L174-L193

Rather than introducing code duplication, please move the original method somewhere it can be shared. One option for this could be to create a class in this same package `class ApplicantBaseView extends BaseHtmlView` that implements `Tag renderPreviousButton(Params params)`, and then have other view classes in this package inherit from it.

For that to work you'll likely need to move the definition of views.applicant.ApplicantProgramBlockEditView.Params to that shared base class as well.

bion (changes requested):

bion (approved):

PR #2442, opened 2022-05-05
Fix cloud/shared/bin/doctor checks based on selected cloud provider.

jeanniefu:
- Check for Docker installation in cloud/shared/bin/doctor for both AWS and Azure.
- Fix should_skip check for AWS CLI. Previously, we were skipping this check for AWS and Azure. We should be checking this for both.

### DIFF
--- a/cloud/shared/bin/doctor_impl
+++ b/cloud/shared/bin/doctor_impl
@@ -77,7 +77,7 @@ class AwsCliCheck(Check):
 
     def should_skip(self):
         # AWS SES is used for both Azure and AWS deployments.
-        return self._get_cloud_provider() in ['aws', 'azure']
+        return self._get_cloud_provider() not in ['aws', 'azure']
 
     def is_ok(self):
         return self._check_command_succeeds(["aws", "help"])
--- a/cloud/shared/bin/doctor_impl
+++ b/cloud/shared/bin/doctor_impl
@@ -13,9 +13,6 @@ if not is_linux and not is_mac:
 class DockerCheck(Check):
     remedy = "Need to install Docker CLI: https://docs.docker.com/engine/install/"
 
-    def should_skip(self):
-        return self._get_cloud_provider() == 'azure'
-
     def name(self):
         return "Docker CLI"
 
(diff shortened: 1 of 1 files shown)


cloud/shared/bin/doctor_impl
```diff
@@ -13,9 +13,6 @@ if not is_linux and not is_mac:
 class DockerCheck(Check):
     remedy = "Need to install Docker CLI: https://docs.docker.com/engine/install/"
 
-    def should_skip(self):
```
jeanniefu, line 16: Btw, is this accurate? I believe we use Docker no matter what.

bion: oof, definitely not accurate. ty!

cloud/shared/bin/doctor_impl
```diff
@@ -79,7 +76,7 @@ class AwsCliCheck(Check):
         return "AWS CLI"
 
     def should_skip(self):
-        return self._get_cloud_provider() in ['aws', 'azure']
+        return self._get_cloud_provider() != 'aws'
```
bion, line 79: issue (blocking): the AWS CLI is needed for Azure deployments as well since our only azure deployment atm uses AWS SES

cloud/shared/bin/doctor_impl
```diff
@@ -133,7 +130,7 @@ class AwsLoginCheck(Check):
         return "AWS login"
 
     def should_skip(self):
-        return self._get_cloud_provider() not in ['aws', 'azure']
+        return self._get_cloud_provider() != 'aws'
```
bion, line 133: issue (blocking): the AWS CLI is needed for Azure deployments as well since our only azure deployment atm uses AWS SES

bion (changes requested):

cloud/shared/bin/doctor_impl
jeanniefu: Ah good to know! Reverted, and added a comment to clarify.

cloud/shared/bin/doctor_impl
jeanniefu: Ah good to know! Reverted, and added a comment to clarify.

bion (approved):

cloud/shared/bin/doctor_impl
jeanniefu: Actually the original check was missing a "not". Added that.

goldblatt (approved):

PR #2445, opened 2022-05-05
if there are multiple drafts for an application throw an error. This should not happen.

azizova-leyla:
### Description
If there are multiple drafts for an application - throw an error. This should not happen. 
Also refactored tests to make them clearer.

### Checklist

### Issue(s)

### DIFF
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -76,27 +76,25 @@ public class ApplicationRepository {
               .eq("applicant.id", applicant.id)
               .eq("program.name", program.getProgramDefinition().adminName())
               .findList();
-      Optional<Application> completedApplication = Optional.empty();
-      for (Application application : oldApplications) {
-        // Delete any in-progress drafts, and mark obsolete any old applications.
-        if (application.getLifecycleStage().equals(LifecycleStage.DRAFT)) {
-          application.setLifecycleStage(LifecycleStage.ACTIVE);
-          application.setSubmitTimeToNow();
-          completedApplication = Optional.of(application);
-        } else {
-          application.setLifecycleStage(LifecycleStage.OBSOLETE);
-        }
-        application.save();
-      }
+      Optional<Application> latestDraft = oldApplications.stream()
+              .filter(app -> app.getLifecycleStage().equals(LifecycleStage.DRAFT))
+              .sorted((app1, app2) -> app2.getCreateTime().compareTo(app1.getCreateTime())).findFirst();
       Application application =
-          completedApplication.orElse(new Application(applicant, program, LifecycleStage.ACTIVE));
+              latestDraft.orElse(new Application(applicant, program, LifecycleStage.ACTIVE));
+      application.setLifecycleStage(LifecycleStage.ACTIVE);
       application.setSubmitTimeToNow();
-
       if (submitterEmail.isPresent()) {
         application.setSubmitterEmail(submitterEmail.get());
       }
-
       application.save();
+
+      for (Application app : oldApplications) {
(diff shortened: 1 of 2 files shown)


clouser2 (approved):

shanemc-goog:
The title says application but the description says Program. Can you clarify those?

There should never be multiple drafts it's a violation of the System design, if there is that is a system crashing bug typically, and whatever is causing that state should be fixed.

Also what is the issue that this is trying to address?  I don't see a linked issue.

azizova-leyla:
> What is the issue that this is trying to address? I don't see a linked issue.
> 
> It'd be good to understand what puts the system in this incorrect state and address that up front. One issue could be if the "start" landing page is loaded in multiple tabs. We should likely address that case.

This is a hypothetical, I don't actually know how we would get to this state. Brett and I were doing a different change, and  saw this and it seemed that current handling could lead to errors if we ever have more than one draft for whatever reason.

azizova-leyla:
Suggestion from Bion: if this happens, throw 500. Don't try to gracefully recover.

clouser2 (approved):

server/app/repository/ApplicationRepository.java
```diff
@@ -76,27 +77,37 @@ private Application submitApplicationInternal(
...
-        application.save();
+      List<Application> latestDraft =
+          oldApplications.stream()
+              .filter(app -> app.getLifecycleStage().equals(LifecycleStage.DRAFT))
+              .collect(Collectors.toList());
+
+      if (latestDraft.size() > 1) {
+        throw new RuntimeException(
+            String.format(
+                "Found more than one DRAFT application " + "for applicant %d, program %d.",
```
bion, line 88: nit:
```suggestion
                "Found more than one DRAFT application for applicant %d, program %d.",
```

server/app/repository/ApplicationRepository.java
```diff
@@ -76,27 +77,37 @@ private Application submitApplicationInternal(
...
-          application.setSubmitTimeToNow();
-          completedApplication = Optional.of(application);
-        } else {
-          application.setLifecycleStage(LifecycleStage.OBSOLETE);
-        }
-        application.save();
+      List<Application> latestDraft =
+          oldApplications.stream()
+              .filter(app -> app.getLifecycleStage().equals(LifecycleStage.DRAFT))
+              .collect(Collectors.toList());
```
bion, line 83: suggestion (blocking): prefer immutable collections from guava as a general rule
```suggestion
      ImmutableList<Application> latestDraft =
          oldApplications.stream()
              .filter(app -> app.getLifecycleStage().equals(LifecycleStage.DRAFT))
              .collect(ImmutableList.toImmutableList());
```

server/app/repository/ApplicationRepository.java
```diff
@@ -76,27 +77,37 @@ private Application submitApplicationInternal(
...
+                "Found more than one DRAFT application " + "for applicant %d, program %d.",
+                applicant.id, program.id));
       }
+
       Application application =
-          completedApplication.orElse(new Application(applicant, program, LifecycleStage.ACTIVE));
-      application.setSubmitTimeToNow();
+          !latestDraft.isEmpty()
+              ? latestDraft.get(0)
+              : new Application(applicant, program, LifecycleStage.ACTIVE);
```
bion, line 95: suggestion (blocking): reads a bit easier without the inversion
```suggestion
          latestDraft.isEmpty()
              ? new Application(applicant, program, LifecycleStage.ACTIVE)
              : latestDraft.get(0);
```

bion (changes requested):

bion (approved):

PR #2446, opened 2022-05-05
Change the terraform command to use arrays so it evaluates properly

kmcnellis:
### Description
The quotes and command substitution `$()` weren't allowing the terraform command to execute correctly.  Instead, it was trying to run (as a command) the results of the terraform command (which was an error, due to a stray missing `\`).

Instead, lets use the recommended solution for storing simple commands, which is bash arrays, evaluated using `"${CMD[@]}"`

### Checklist
- Tested scripts using the github/civiform/civiform-deploy deploy script

### DIFF
--- a/cloud/shared/bin/lib/terraform.sh
+++ b/cloud/shared/bin/lib/terraform.sh
@@ -18,11 +18,11 @@ function terraform::perform_apply() {
     "cloud/${CIVIFORM_CLOUD_PROVIDER}/bin/setup_tf_shared_state" \
       "${TERRAFORM_TEMPLATE_DIR}/${BACKEND_VARS_FILENAME}"
 
-      terraform "${TERRAFORM_DIR}" \
-        init \
-        -input=false \
-        -upgrade \
-        -backend-config="${BACKEND_VARS_FILENAME}"
+    terraform "${TERRAFORM_DIR}" \
+      init \
+      -input=false \
+      -upgrade \
+      -backend-config="${BACKEND_VARS_FILENAME}"
   fi
 
   if azure::is_service_principal; then
--- a/cloud/shared/bin/lib/terraform.sh
+++ b/cloud/shared/bin/lib/terraform.sh
@@ -1,12 +1,7 @@
 #! /usr/bin/env bash
 
-readonly TERRAFORM_BASE_COMMAND="terraform \
-        -chdir=${TERRAFORM_TEMPLATE_DIR}"
-
-readonly TERRAFORM_APPLY="${TERRAFORM_BASE_COMMAND} \
-        apply \
-        -input=false \
-        -var-file=${TF_VAR_FILENAME}"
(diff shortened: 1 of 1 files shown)


cloud/shared/bin/lib/terraform.sh
```diff
@@ -18,24 +13,22 @@ readonly TERRAFORM_APPLY="${TERRAFORM_BASE_COMMAND} \
 #######################################
 function terraform::perform_apply() {
   if [[ "${CIVIFORM_MODE}" == "dev" ]]; then
-    "$(${TERRAFORM_BASE_COMMAND} init -upgrade)"
+    terraform "${TERRAFORM_DIR}" init -upgrade
```
goldblatt, line 16: this seems not quite right?

cloud/shared/bin/lib/terraform.sh
```diff
@@ -18,24 +13,22 @@ readonly TERRAFORM_APPLY="${TERRAFORM_BASE_COMMAND} \
...
   else
     "cloud/${CIVIFORM_CLOUD_PROVIDER}/bin/setup_tf_shared_state" \
       "${TERRAFORM_TEMPLATE_DIR}/${BACKEND_VARS_FILENAME}"
 
-    "$(
-      ${TERRAFORM_BASE_COMMAND}
+    terraform "${TERRAFORM_DIR}" \
```
goldblatt, line 21: same here

goldblatt (changes requested):

cloud/shared/bin/lib/terraform.sh
```diff
@@ -18,24 +13,22 @@ readonly TERRAFORM_APPLY="${TERRAFORM_BASE_COMMAND} \
...
-        -backend-config="${BACKEND_VARS_FILENAME}"
-    )"
+      -input=false \
+      -upgrade \
+      -backend-config="${BACKEND_VARS_FILENAME}"
   fi
 
   if azure::is_service_principal; then
-    "$(${TERRAFORM_APPLY} -auto-approve)"
+    "${TERRAFORM_APPLY[@]}" -auto-approve
```
azizova-leyla, line 29: just for my understanding: [@] here is for applying the arguments correctly?

cloud/shared/bin/lib/terraform.sh
kmcnellis: Thanks for catching this!  Yes, these are leftover from my first attempt

cloud/shared/bin/lib/terraform.sh
kmcnellis: same

cloud/shared/bin/lib/terraform.sh
kmcnellis: This is how you expand the elements of the array: https://www.baeldung.com/linux/store-command-in-variable

goldblatt (approved):

azizova-leyla (approved):

PR #2451, opened 2022-05-06
Removes the question interface as it adds unneeded indirection.

clouser2:
### Description
All question types extend a single abstract `QuestionImpl` class. There's not yet a concrete need for an interface. While here, I marked the subclasses as final and made their constructors package-private since we intend for the classes to be constructed using the factory method in `ApplicantQuestion`.

This is follow-on from PR requests for issue #1944.

All changes should be purely code motion.


server/app/services/applicant/question/Question.java
```diff
@@ -18,27 +22,71 @@
  *   <li>Present validation errors if any
  * </ul>
  */
-public interface Question {
```
clouser2, line 21: Note that all of the additions from this class are moving the implementation from QuestionImpl.java (now deleted) as-is.

clouser2:
This needs rework. I suspect my merge from main corrupted the state here. I'll send out a fresh PR once I've worked it out.

PR #2452, opened 2022-05-06
Removing client-side validation related to enumerators in favor of server-side validation

clouser2:
### Description
As part of this, I also added a browser assertion that blank inputs weren't accepted and fixed up a bug where the add button would remain disabled if a blank entity input were removed. Finally, moved some actual JS mostly unrelated to validation closer to the other enumerator behavior in main.ts

### Checklist

### Issue(s)
#1944


server/app/assets/javascripts/main.ts
```diff
@@ -162,6 +168,70 @@ function removeExistingEnumeratorField(event: Event) {
   enumeratorFieldDiv.appendChild(deletedEntityInput)
 }
 
+/** Add listeners to all enumerator inputs to update validation on changes. */
+function addEnumeratorListeners() {
+  // Assumption: There is only ever zero or one enumerators per block.
+  const enumeratorQuestion = document.querySelector('.cf-question-enumerator')
+  if (enumeratorQuestion) {
```
bion, line 175: suggestion (non-blocking): making this an early return would reduce the level of indentation for the rest of function
```suggestion
  if (!enumeratorQuestion) {
    return;
  }
```

bion (approved):

server/app/assets/javascripts/main.ts
clouser2: Done.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -157,14 +157,25 @@ describe('End to end enumerator test', () => {
     await applicantQuestions.answerNameQuestion('Daffy', 'Duck')
     await applicantQuestions.clickNext()
 
+    // Put an empty answer in the nested enumerator for enum two
```
shanemc-goog, line 160:
```suggestion
    // Put an empty answer in the nested enumerator for enum two
```
```suggestion
    // Put an empty answer in the nested enumerator for enum two.
```

server/app/assets/javascripts/main.ts
```diff
@@ -148,6 +148,12 @@ function removeExistingEnumeratorField(event: Event) {
   // submit the input to maintain entity ordering.
   const enumeratorFieldDiv = removeButton.parentElement
   enumeratorFieldDiv.classList.add('hidden')
+  // Why add the hidden class to the input as well?
```
shanemc-goog, line 151: I think this Q&A format is confusing, it reads more like you found something odd and don't understand it, rather than trying to explain a deliberate but non-obvious need.

In general comments are clearer if they're just direct about the situation:

EG:

// We must hide the child in addition to the parent because of .......

shanemc-goog (changes requested):
looks good!

browser-test/src/end_to_end_enumerators.test.ts
clouser2: Done.

server/app/assets/javascripts/main.ts
clouser2: Switched to a more direct comment. I've generally found the Q&A format to be helpful since it allows for describing multiple non-obvious behaviors. In this case, there's only a single behavior, so it's definitely less useful. I'm fine not using it, especially if it's not consistent with the rest of the codebase.

server/app/assets/javascripts/main.ts
```diff
@@ -162,6 +168,71 @@ function removeExistingEnumeratorField(event: Event) {
...
+  })
+
+  mutationObserver.observe(enumeratorQuestion, {
+    childList: true,
+    subtree: true,
+    characterDataOldValue: true,
+  })
+}
+
+/** if we have empty inputs then disable the add input button. (We don't need two blank inputs.) */
```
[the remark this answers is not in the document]
clouser2: Fixed. This was copied from the original implementation. Good catch though.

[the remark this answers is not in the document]
clouser2: Seems to be a dupe of the above.

server/app/assets/javascripts/main.ts
shanemc-goog: I can def see that. Maybe it just needs more pre-ample or something since it's not a norm that I've seen, or perhaps as you mentioned, in the singleton case it doesn't self-document the structure of the comment clearly.

server/app/assets/javascripts/main.ts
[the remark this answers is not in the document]
shanemc-goog: yeah, Github did a lot of duplication for some reason.

shanemc-goog (approved):

server/app/assets/javascripts/main.ts
clouser2: Yah. If I stumble upon a case where I think it's the strictly preferred format, I'll definitely call it out at that point. Thanks for the suggestion!

PR #2457, opened 2022-05-09
Deserialize dates using the UTC timezone instead of the system default.

### LINKED ISSUE #2342: Check timezones for date inputs for roc deployment
**Describe the bug**
The code base hard codes Los Angeles as the system timezone.  This may have undesired interactions such as when users enter a date, and we store that date as a UTC based time.

EG:  If they enter 2/1/22, and it's stored with an implication of PT, when it's later rendered it might be displayed as the previous/next day etc.  This might be OK, but we should verify and have it WAI for some set intention.

Just to work through an example:
Given LA is at or behind most states, the likely impact is that a date entered in Rochester (ET) will be stored as PT, and if it's rendered timezone aware it'd show as the same date but 3AM; and likely we'd not render the hours anyway.  If we had defaulted the date to ET, then it would likely cause the previous date to be shown for Seattle.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

clouser2:
### Description
If the result of calling `ZoneId.systemDefault()` resolved to a value behind UTC (e.g. PST), this meant that dates would be deserialized incorrectly (e.g. 2022-01-01 instead of 2022-02-02). This is due to the fact that we serialized the date [using UTC](https://github.com/seattle-uat/civiform/blob/main/server/app/services/CfJsonDocumentContext.java#L111-L118) and deserialized it using [`ZoneId.systemDefault()`](https://github.com/seattle-uat/civiform/blob/main/server/app/services/CfJsonDocumentContext.java#L312-L315).

Whether deployments are affected are dependent on what `ZoneId.systemDefault()` resolves to on the deployment. After this PR, the default TimeZone should have no effect on serialization (confirmed by the added test).

### Checklist

### Issue(s)
Fixes #2342;


server/test/services/CfJsonDocumentContextTest.java
```diff
@@ -5,12 +5,37 @@
...
+  private static TimeZone origTimeZone;
+
+  @BeforeClass
+  public static void setTimeZoneToNonAmerica() {
+    origTimeZone = TimeZone.getDefault();
+    // Set the default time zone to something behind UTC. We rely on this for
+    // the date serialization / deserialization test below, where we expect
+    // everything to occur in terms of UTC rather than the system default
+    // time zone.
+    TimeZone.setDefault(TimeZone.getTimeZone(BEHIND_UTC_ZONE));
```
clouser2, line 30: I wanted to ensure the test class reset the timezone to the original value after the tests were run. This should be safe assuming we never run tests in parallel. If we were to move towards that world, if other tests relied on a pre-set time zone, we could run into problems.

If that day comes, we could instead investigate setting the system default time zone for the test runner to this for ALL tests rather than doing it on a one-off basis.

Given that we're not in that world right now, I thought it made more sense to keep the logic closer to its use.

bion: > Given that we're not in that world right now, I thought it made more sense to keep the logic closer to its use.

That sounds good to me too

bion (approved):

shanemc-goog (approved):

PR #2458, opened 2022-05-09
Prevent non draft modification of Programs

### LINKED ISSUE #2402: Non Draft Programs can be modified after the fact
**Describe the bug**

You can load the editor for an Active/Obsolete program by:
1. Having a stale tab available after the program is changed elsewhere
2. Through the edit url by specifying the program ID for a previous revision.

You then can still edit the program which edits the configuration which is otherwise required to be imuttable.

Related to #2047

**To Reproduce**
1. Make a Draft program and leave the tab open in the editor.
2. In another tab publish the program

(Notice the config in the DB) 

4. In the first tab continue to edit the program

(notice the config in the DB is changing.

**Expected behavior**
If a non Draft is being edited a FE error is issued and require the user to reload.

**Screenshots**

**Additional context**

shanemc-goog:
### Description
Many of the program configuration handlers will arbitrarily modify the program ID sent by the FE.  The system design requires that only Draft programs can be modified; once a program is published it must be immuttable.

However the current behavior allows for editing Active programs.  The most likely way this situation would occur is if an Admin is working with a program in multiple tabs; where they publish in one but then continue to edit the previous-draft ID in the other tab thinking they're instead editing "the draft".

This PR blocks the data corruption aspect by throwing an exception, subsequently we may want to add better UX on helping the Admin get to the right place in the system they want, which would involve navigating to the current revision.

This PR also adds a new error handling mechanism in ErrorHandler so that we can throw RuntimeExceptions but have them surfaced to users as application error.

* Fixed some tests that were editing an Active program
* Replaced indirect NOT_FOUND responses with the new exceptions.  The previous were issued by utility calls and not explicitly declared by the handlers.  Given NOT_FOUND errors would come from errors in the UI or URL manipulation the new exception is fine in my opinion,

### Checklist

### Issue(s)
Fixes #2402


server/app/controllers/admin/NotChangeableException.java
```diff
@@ -0,0 +1,8 @@
+package controllers.admin;
+
+/** Exception when a request is sent to modify something that can not be changed. */
+class NotChangeableException extends RuntimeException {
+  public NotChangeableException(String message) {
+    super(message);
+  }
+}
```
bion, line 8: issue (blocking): since these aren't caught and are runtime exceptions, they'll result in a 500 status code which is incorrect -- the client is sending a bad request and should receive a 400.

It's annoying since there are so many places where this is checked. I wonder if an alternative could be to not extend RuntimeException and extend the [the default error handler](https://www.playframework.com/documentation/2.8.x/JavaErrorHandling) to handle  `NotChangeableException`

server/app/controllers/admin/RequestChecker.java
```diff
@@ -0,0 +1,25 @@
...
+import javax.inject.Inject;
+import repository.VersionRepository;
+
+/** Utility class for checking Request parameters. */
+class RequestChecker {
+  private VersionRepository versionRepository;
+
+  @Inject
+  public RequestChecker(VersionRepository versionRepository) {
+    this.versionRepository = versionRepository;
```
bion, line 12: nit: check not null

server/app/controllers/admin/RequestChecker.java
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  /**
+   * Verifies programId is associated with the Draft version and throws {@code
+   * NotChangeableException} if not.
+   */
+  void throwIfNotDraft(Long programId) {
+    if (versionRepository.isDraftProgram(programId)) {
```
bion, line 20: suggestion (blocking): since the method signature doesn't specify the resource type (just a `Long`) the method name should since there are other resources that can be in draft mode.
```suggestion
  void throwIfProgramNotDraft(Long programId) {
    if (versionRepository.isDraftProgram(programId)) {
```

bion (changes requested):

server/app/controllers/admin/NotChangeableException.java
shanemc-goog: My plan was to go down that route or piggy back on any changes Brett made subsequently.  Blocking these invalid edits seemed worth getting in.  Any pre-existing knowledge on the level of effort on doing the play handling for these?

server/app/controllers/admin/RequestChecker.java
shanemc-goog: done

server/app/controllers/admin/NotChangeableException.java
bion: > Any pre-existing knowledge on the level of effort on doing the play handling for these?

nope, my guess is it'd be fairly quick though. you might be able to get away with leaving `NotChangeableException` as a runtime exception, [adding a new singleton error handler that extends the default one](https://www.playframework.com/documentation/2.8.x/JavaErrorHandling#Extending-the-default-error-handler), and then overriding `onServerError` that checks if  the throwable is a `NotChangeableException` and doing `return Results.badRequest("form submitted with stale data");` if it is.

if that inflates the level of effort more than expected i'm ok punting though.

bion (changes requested):

server/app/controllers/ErrorHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import javax.inject.*;
+import play.*;
+import play.api.OptionalSourceMapper;
+import play.api.routing.Router;
+import play.http.DefaultHttpErrorHandler;
+import play.mvc.*;
+import play.mvc.Http.*;
```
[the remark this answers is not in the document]
shanemc-goog: done!  copy and paste.... :)

server/app/controllers/admin/NotChangeableException.java
shanemc-goog: Done!

bion (approved):

server/test/controllers/ErrorHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  public void findThrowableByType_findsSource() {
+    Throwable exception = new NotChangeableException("test exception");
+    assertThat(ErrorHandler.findThrowableByType(exception, RuntimeException.class))
+        .contains(exception);
+  }
+
+  @Test
+  public void findThrowableByType_findsNested() {
```
clouser2, line 37: Not blocking: Consider adding a test where the exception we want is nested >N levels deep and isn't found. The rationale being that it documents that there are limits to what the error handling can automatically detect.

server/test/controllers/admin/AdminProgramBlockPredicatesControllerTest.java
```diff
@@ -39,12 +40,11 @@ public void setup() {
   }
 
   @Test
-  public void edit_withInvalidProgram_notFound() {
-    Http.Request request = fakeRequest().build();
-
-    Result result = controller.edit(request, 1L, 1L);
-
-    assertThat(result.status()).isEqualTo(NOT_FOUND);
+  public void edit_withNonExistantProgram_notFound() {
```
clouser2, line 43: This test body is the same as edit_withActiveProgram_throws below. I think it needs a different body to be consistent with it's current name.

clouser2 (approved):

server/app/controllers/ErrorHandler.java
```diff
@@ -0,0 +1,69 @@
...
+   * Finds an exception of type {@code search} by looking through {@code exception}'s cause chain.
+   * Will also consider {@code exception}.
+   *
+   * <p>The framework provides wrapped exceptions to the methods in this class so we have to dig out
+   * our application exception. Anecdotally it's 2 levels down.
+   */
+  static Optional<Throwable> findThrowableByType(Throwable exception, Class<?> search) {
+    Optional<Throwable> root = Optional.of(exception);
+    // Search a couple causes deep for the desired type.
+    for (int i = 0; i < 5 && root.isPresent(); ++i) {
```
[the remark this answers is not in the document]
shanemc-goog: this is meant to account for framework level wrapping, which currently is only 2 levels deep anecdotally.  I did 5 to allow for that changing a little, but if the framework starts to wrap these a lot, we should adapt otherwise I think.

server/test/controllers/ErrorHandlerTest.java
shanemc-goog: I'm not inclined to document exclusively through tests perse, I added it to the protected method doc though.

I think having the limit isn't a feature as much as a trade off so having a test fail with 6 levels of nesting doesn't seem useful either.

server/test/controllers/admin/AdminProgramBlockPredicatesControllerTest.java
shanemc-goog: done!

server/app/controllers/ErrorHandler.java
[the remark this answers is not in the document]
clouser2: Works for me.

clouser2 (approved):

PR #2460, opened 2022-05-09
Reset the Optional for each test.

shanemc-goog:
### Description
TearDown may happen when unnecessary otherwise.

### Checklist

### Issue(s)

### DIFF
--- a/server/test/modules/MainModuleTest.java
+++ b/server/test/modules/MainModuleTest.java
@@ -34,7 +34,6 @@ public class MainModuleTest {
   public void setUpApp() {
     maybeApp = Optional.empty();
   }
-  
   @After
   public void tearDownApp() {
     if (maybeApp.isPresent()) {
--- a/server/test/modules/MainModuleTest.java
+++ b/server/test/modules/MainModuleTest.java
@@ -34,6 +34,7 @@ public class MainModuleTest {
   public void setUpApp() {
     maybeApp = Optional.empty();
   }
+
   @After
   public void tearDownApp() {
     if (maybeApp.isPresent()) {
--- a/server/test/modules/MainModuleTest.java
+++ b/server/test/modules/MainModuleTest.java
@@ -27,7 +27,6 @@ public class MainModuleTest {
 
   // Note: Each test is responsible for configuring an application object
   // since we're varying the configuration values of the application itself.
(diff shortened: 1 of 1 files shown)


clouser2 (approved):

server/test/modules/MainModuleTest.java
```diff
@@ -29,6 +30,11 @@ public class MainModuleTest {
   // This means that there's no need or a corresponding @Before method.
   private Optional<Application> maybeApp;
 
+  @Before
+  public void setUpApp() {
+    maybeApp = Optional.empty();
+  }
+
```
shanemc-goog, line 37:
```suggestion

```

server/test/modules/MainModuleTest.java
```diff
@@ -29,6 +30,10 @@ public class MainModuleTest {
   // This means that there's no need or a corresponding @Before method.
   private Optional<Application> maybeApp;
 
+  @Before
+  public void setUpApp() {
+    maybeApp = Optional.empty();
+  }
```
shanemc-goog, line 36:
```suggestion
  }

```

bion (approved):
Please add a code comment explaining why that code needs to be there

shanemc-goog:
> Please add a code comment explaining why that code needs to be there

I don't believe this strays from any standard test setup such that it needs calling out.  EG @Before configures state before tests and typically sets unset class members, as is happening here.

Are you feeling like the class member needs documentation more so?

server/test/modules/MainModuleTest.java
```diff
@@ -29,6 +30,11 @@ public class MainModuleTest {
   // This means that there's no need or a corresponding @Before method.
```
clouser2, line 30: This can be removed

clouser2 (approved):

bion:
> Are you feeling like the class member needs documentation more so?

ahhh maybe not, after looking at it more closely. feel free to ignore

server/test/modules/MainModuleTest.java
shanemc-goog, line 30:
```suggestion
```

shanemc-goog: done

PR #2463, opened 2022-05-09
Plumb CIVIFORM_TIME_ZONE_ID to Terraform.

clouser2:
### Description
This makes it possible for the `CIVIFORM_TIME_ZONE_ID` env variable to be set as part of configuring a deployment.

I could use some help manually testing this. Fixing this will unblock deployments that want to use a different timezone than `America/Los_Angeles`.

### Checklist
- [X] Manually confirmed that a default deployment  had times appear in PDT time.
- [X] Redeployed and set `export CIVIFORM_TIME_ZONE_ID="America/New_York"` in `civiform_config.sh`. Confirmed times appeared in EST time.

### Issue(s)
#2341


clouser2:
> @clouser2 failing TF validation check

Fixed. Thanks!

cloud/azure/templates/azure_saml_ses/variables.tf
```diff
@@ -3,6 +3,11 @@ variable "azure_resource_group" {
   description = "Name of the resource group where key vault is already created."
 }
 
+variable "civiform_time_zone_id" {
+  type        = string
+  description = "Time zone for Civiform server to use when displaying dates."
+}
```
goldblatt, line 9: i'd probably add a default here just so it doesn't break the deploy our vars will overwrite the default but while this value gets defaulted in the conf file it also needs to be defaulted in the terraform (otherwise terraform will complain)

clouser2: Done!

cloud/deploys/dev_azure/civiform_config.example.sh
```diff
@@ -9,6 +9,7 @@ export CIVIC_ENTITY_FULL_NAME="City of Sooschester"
 export CIVIC_ENTITY_SUPPORT_EMAIL_ADDRESS="civiform-azure-staging-email@googlegroups.com"
 export CIVIC_ENTITY_LOGO_WITH_NAME_URL="https://www.cityofrochester.gov/images/header-logo.png"
 export CIVIC_ENTITY_SMALL_LOGO_URL="https://www.cityofrochester.gov/assets/0/117/8589934986/c6791498-270e-40b1-9b81-aca8a510c1b5.jpg"
+export CIVIFORM_TIME_ZONE_ID="America/Los_Angeles"
```
kmcnellis, line 12: You'll need to make this change in the https://github.com/civiform/civiform-deploy repo, too

kmcnellis (approved):

azizova-leyla (approved):

cloud/deploys/dev_azure/civiform_config.example.sh
clouser2: Thanks for flagging, I'll send out a PR for that today. Since we marked this as "optional", nothing will break in the civiform-deploy repo without this change. However, without making it there, it won't be very discoverable at all :).

PR #2466, opened 2022-05-10
Updating browser tests to better handle the warning toast

gwendolyngoetz:
### Description

In trying to migrating the staging to deployment to seattle-uat/civiform-deploy I kept getting timeout failures of browser tests. The warning message toast may or may not exist in the DOM. When code tries to click the [x] to dismiss the toast and it isn't there the click timeout default is 30 seconds.

There are two options to address the clicking of element

**Option 1: Add a timeout to the click event**
```
await page.click('#warning-message-dismiss', { timeout: 500 })
```

**Option 2: Use a selector to find the element, and call the click method if the element was found**
```
var element = await page.$('#warning-message-dismiss')

if (element !== null){
  await element.click()
} 
```

I've tried both options and they work locally and against staging. I'm liking Option 2 better since it doesn't rely on hard coding a timeout value.

The changes in the admin_predicates.test.ts is an example of what I'm proposing I go do to all the test files.

### DIFF
--- a/browser-test/src/admin_predicates.test.ts
+++ b/browser-test/src/admin_predicates.test.ts
@@ -1,3 +1,4 @@
+import { Browser, Page } from 'playwright'
 import {
   AdminPredicates,
   AdminPrograms,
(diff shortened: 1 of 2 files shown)


goldblatt:
which tests are the issue? i guess i'm fine with ignoring the toast if the goal of the tests is for something else and if we have something that confirms the toast works and is dismissible

gwendolyngoetz:
It's a good number of files, most appear to be around admin pages. I can go through the log and pull out the full list. Because the fixed warning "Do not enter actual or personal data in this demo site" is in the way it needs to be dismissed in order to click on some links.

I'm thinking to have the tests automatically go and dismiss that toast right after the initial start of each test to get it out of the way. I've seen some tests perform an action and save and check for a success toast as part of their asserts, I'll have to check to see if there is a test specific to toasts.

PR #2477, opened 2022-05-11
Prevent double-clicking the submit button from submitting multiple applications

### LINKED ISSUE #2469: Double clicking on "Submit" should not create/submit 2 applications
If a user clicks on the "Submit" button X # of times during application submission, we create/submit X applications. We discovered this while practicing for the Bloomington UXR, when we clicked "Submit" 5 times and 5 applications were created / the Application ID # went up by 5. Some users are used to double clicking to take action, so this is something we should think about. Maybe we can disable the "Submit" button after the first click?

clouser2:
### Description
This change disables the submit button on the application review page for 5 seconds after it has been clicked.

See https://github.com/seattle-uat/civiform/issues/2469#issuecomment-1124077383 for rationale.

Note: A nice cleanup on the server side would also be to throw a 500 in these cases rather than creating duplicate applications. That said, we'd still want to prevent the bad data from coming in from the client-side so that the applicant doesn't see a 500.

### Issue(s)
Fixes #2469

### DIFF
--- a/server/app/assets/javascripts/main.ts
+++ b/server/app/assets/javascripts/main.ts
@@ -351,12 +351,9 @@ function attachFormDebouncers() {
       return
     }
     // Prevent double-clicks from submitting the form multiple times by
-    // disabling the button for a short period after the initial click.
+    // disabling the submit button after the initial click.
     el.addEventListener('submit', () => {
       submitEl.setAttribute('disabled', '')
-      setTimeout(() => {
-        submitEl.removeAttribute('disabled')
-      }, 5000)
     })
   })
 }
--- a/server/app/assets/javascripts/main.ts
+++ b/server/app/assets/javascripts/main.ts
@@ -350,8 +350,13 @@ function attachFormDebouncers() {
     if (!submitEl) {
       return
     }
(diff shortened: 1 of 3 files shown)


server/app/assets/javascripts/main.ts
```diff
@@ -344,6 +344,20 @@ function configurePredicateFormOnOperatorChange(event: Event) {
   )
 }
 
+function attachFormDebouncers() {
+  Array.from(document.querySelectorAll('.cf-debounced-form')).forEach((el) => {
+    const submitEl = el.querySelector('button[type="submit"]')
```
gwendolyngoetz, line 349: Suggestion: Rename el to formEl for added clarity.

gwendolyngoetz (approved):

bion (approved):

server/app/assets/javascripts/main.ts
clouser2: Done!

shanemc-goog (approved):

gwendolyngoetz (approved):

PR #2485, opened 2022-05-12
Move client-side validation for file uploads to a new file.

### LINKED ISSUE #1944: Validation error messages for applicants are mixed between server and client-side rendering

clouser2:
### Description
File upload questions follow a different pattern where some client-side validation is still required. AWS will reject empty file upload requests. Since CiviForm isn't the target of the form POST, our standard server-side error logic can't apply as easily. Updating file upload logic is a bit out of scope for removing client-side validation, so leaving this path intact.

For this, I renamed `validation.ts` to `file_upload.ts` and only included the script for file upload blocks.

### Issue(s)
Fixes #1944


server/app/assets/javascripts/main.ts
```diff
@@ -416,17 +416,19 @@ function configurePredicateFormOnOperatorChange(event: Event) {
 }
 
 function attachFormDebouncers() {
-  Array.from(document.querySelectorAll('.cf-debounced-form')).forEach((formEl) => {
```
clouser2, line 419: While technically unrelated, this was caught as part of running `bin/fmt`, which didn't catch this file before. As far as I can tell, the only changes here are formatting.

bion (approved):

server/app/assets/javascripts/file_upload.ts
```diff
@@ -0,0 +1,37 @@
+window.addEventListener('load', (event) => {
```
jeanniefu, line 1: Do we have test coverage for this?

jeanniefu (approved):

server/app/assets/javascripts/file_upload.ts
clouser2: There isn't currently any unit test coverage of our TS files :(. However, there is browser test coverage of the "requiredness" constraints that file is responsible for here: https://github.com/seattle-uat/civiform/blob/1c9ed6afa33336531fdf0eeca0729717007415fa/browser-test/src/file.test.ts#L75-L84

PR #2487, opened 2022-05-12
Don't show the skip button in Azure for required questions for file uploads

### LINKED ISSUE #2472: "Skip" doesn't let you skip the question if it's required for File Uploads
Another behavior we found surprising when playing with the UXR prototype. For document/file uploads, we saw that there was a "Skip" button, which helps with the case where the applicant cannot upload the file at the moment for whatever reason. But then the applicant actually can't skip the question if it's required: clicking on "Skip" shows a "This question is required." error. I don't think this makes sense and we should discuss intended behavior!

clouser2:
### Description
We could likely leverage some code sharing here to reduce the overall code duplication. I didn't pursue that here. We'll likely want to consider that as a separate workstream. Added a regression browser test.

### Checklist

### Issue(s)
Fixes #2472

### DIFF
--- a/server/app/views/AzureFileUploadViewStrategy.java
+++ b/server/app/views/AzureFileUploadViewStrategy.java
@@ -112,6 +112,10 @@ public class AzureFileUploadViewStrategy extends FileUploadViewStrategy {
 
   @Override
   protected Optional<ContainerTag> maybeRenderSkipOrDeleteButton(Params params) {
+    if (hasAtLeastOneRequiredQuestion(params)) {
+      // If the file question is required, skip or delete is not allowed.
+      return Optional.empty();
+    }
     String buttonText = params.messages().at(MessageKey.BUTTON_SKIP_FILEUPLOAD.getKeyName());
     String buttonId = FILEUPLOAD_SKIP_BUTTON_ID;
     Optional<ContainerTag> footer = Optional.empty();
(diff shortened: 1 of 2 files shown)


server/app/views/AzureFileUploadViewStrategy.java
```diff
@@ -112,6 +112,10 @@ public Tag renderFileUploadBlockSubmitForms(
 
   @Override
   protected Optional<ContainerTag> maybeRenderSkipOrDeleteButton(Params params) {
+    if (hasAtLeastOneRequiredQuestion(params)) {
+      // If the file question is required, skip or delete is not allowed.
```
swatkat1, line 116: If both the functions do the same, should we just move this to the FileUploadViewStrategy.java?

bion (approved):

server/app/views/AzureFileUploadViewStrategy.java
bion: there's definitely opportunity for deeper refactoring here, but this quick fix will resolve the bug in prod

swatkat1 (approved):

server/app/views/AzureFileUploadViewStrategy.java
clouser2: I definitely agree. There's a lot of UI building code here that would benefit from deduplication. For now, I'm going to proceed with the quick fix.

PR #2488, opened 2022-05-12
Add a deploy test for azure saml ses

goldblatt:
### Description
Adds a github action that will execute a deploy to staging which just 'plans' the terraform. this should do everthing outside of actually deploying. this should be reasonably fast and cost less than deploying. we can ideally run this on changes to the cloud directory to confirm if we are doing something to the deploy system.

### Checklist
- [ ] Add support for this to do setup script as well 
- [x] Setup this as part of the tests framework
- [x] confirm this fails when it fails: i.e if it fails to deploy we should error out, currently it looks like it passes
- [x] add the correct arm info 

### Issue(s)

### DIFF
--- a/.github/workflows/azure-saml-ses-deploy-test.yaml
+++ b/.github/workflows/azure-saml-ses-deploy-test.yaml
@@ -18,17 +18,19 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.SECRET_ACCESS_KEY }}
     steps:
-      - name: check out branch
+      - name: Check out branch
         uses: actions/checkout@v2
 
-      - name: login to Azure
+      - name: Login to Azure
         run: cloud/azure/bin/cli-login
 
-      - name: resolve latest snapshot tag
+      - name: Resolve latest snapshot tag
         run: source cloud/shared/bin/resolve-latest-snapshot-tag
 
-      - name: copy the staging file
+      - name: Copy the config from staging 
         run: |
+          #  Grab the config from the staging deploy repo to confirm
+          #  the changes work with actual data
           curl https://raw.githubusercontent.com/civiform/staging-azure-deploy/main/civiform_config.sh > civiform_config.sh
 
       - name: Source config, set to be test and deploy with latest snapshot tag
@@ -37,10 +39,4 @@ jobs:
           export CIVIFORM_MODE="test"
           echo "${LATEST_SNAPSHOT_TAG}"
           cloud/shared/bin/deploy "--tag=${LATEST_SNAPSHOT_TAG}"
-          if [[ $? == 1 ]]; then
-            echo "Terraform plan failed"
-            exit 1
-          else
-            echo "Terraform plan success"
-            exit 0
-          fi
+          exit $?
--- a/.github/workflows/azure-saml-ses-deploy-test.yaml
+++ b/.github/workflows/azure-saml-ses-deploy-test.yaml
@@ -34,10 +34,3 @@ jobs:
           source civiform_config.sh
           export CIVIFORM_MODE="test"
           cloud/azure/bin/deploy "--tag=${LATEST_SNAPSHOT_TAG}"
-          if $?; then
-            echo "exited successfully"
-            exit 0
-          else
-            echo "exited with failures"
-            exit 1
-          fi
--- a/.github/workflows/azure-saml-ses-deploy-test.yaml
+++ b/.github/workflows/azure-saml-ses-deploy-test.yaml
@@ -1,0 +1,35 @@
+name: azure_saml_ses deploy test
+on:
+  workflow_dispatch: {}
+  push:
+    paths:
+      - 'cloud/**'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    env:
+      ARM_TENANT_ID: ${{ secrets.AZURE_SERVICE_PRINCIPAL_TENANT }}
+      ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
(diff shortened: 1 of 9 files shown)


goldblatt:
note we might want to switch the secrets to do a new service principal here

cloud/shared/bin/lib/terraform.sh
```diff
@@ -2,7 +2,7 @@
 
 # https://www.baeldung.com/linux/store-command-in-variable
 readonly TERRAFORM_CMD=("terraform" "-chdir=${TERRAFORM_TEMPLATE_DIR}")
-readonly TERRAFORM_APPLY=(${TERRAFORM_CMD[@]} "apply" "-input=false" "-var-file=${TF_VAR_FILENAME}")
+readonly TERRAFORM_APPLY=(${TERRAFORM_CMD[@]} "apply" "-input=false" "-json")
```
bion, line 5: question (blocking): TERRAFORM_APPLY doesn't need the var file?

cloud/shared/bin/resolve-latest-snapshot-tag
```diff
@@ -0,0 +1,22 @@
...
+if [[ -z "${snapshot_tag}" ]]; then
+  echo "Latest snapshot tag not found." 2>&1
+  exit 1
+fi
+
+export LATEST_SNAPSHOT_TAG="${snapshot_tag}"
+
+if [[ -n "${GITHUB_ENV}" ]]; then
+  echo "LATEST_SNAPSHOT_TAG=${snapshot_tag}" >>"${GITHUB_ENV}"
+fi
```
bion, line 22: this looks familiar

bion (changes requested):

cloud/shared/bin/resolve-latest-snapshot-tag
goldblatt: will update the other repo to use this code once this code is checked in

cloud/shared/bin/lib/terraform.sh
goldblatt: the plan does now; i split it up to use plan which allows the test to work but also might allow for future automation and like approvals to deploy (which is what tf documentation about how to handle automation recommends)

.github/workflows/azure-saml-ses-deploy-test.yaml
```diff
@@ -33,3 +33,8 @@ jobs:
           source civiform_config.sh
           export CIVIFORM_MODE="test"
           cloud/azure/bin/deploy "--tag=${LATEST_SNAPSHOT_TAG}"
+          if $?; then
+            exit 0
+          else
+            exit 1
+          fi
```
bion, line 46: could this be `exit $?`?

bion (changes requested):

cloud/azure/bin/lib/azure_log.sh
```diff
@@ -17,6 +17,10 @@ readonly AZURE_LOG_FILE_NAME="civiform_deployment_log.txt"
 #   LOG_TEMPFILE: the path to a tempfile containing the deploy log.
 #######################################
 function azure_log::fetch_log_file() {
+  if [[ "${CIVIFORM_MODE}" == "test" ]]; then
```
bion, line 20: suggestion (blocking): extract a function
```suggestion
  if civiform_mode::is_test; then
```

.github/workflows/azure-saml-ses-deploy-test.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: login to Azure
+        run: cloud/azure/bin/cli-login
+
+      - name: resolve latest snapshot tag
+        run: source cloud/shared/bin/resolve-latest-snapshot-tag
+
+      - name: copy the staging file
+        run: |
+          curl https://raw.githubusercontent.com/civiform/staging-azure-deploy/main/civiform_config.sh > civiform_config.sh
```
bion, line 32: suggestion (blocking): please comment why we do this

bion (changes requested):

.github/workflows/azure-saml-ses-deploy-test.yaml
goldblatt: terraform will exit with 0 or 2 on successso had to change this out a lil. deets here https://www.terraform.io/cli/commands/plan#detailed-exitcode

goldblatt: ooo actually it might not do this if we don't pass that arg in

.github/workflows/azure-saml-ses-deploy-test.yaml
bion: @goldblatt one more here

bion (changes requested):

.github/workflows/azure-saml-ses-deploy-test.yaml
goldblatt: huh wait i added a comment might have gotten reverted

goldblatt: ahh yeah it should have it now, maybe github had a lil delay

bion (approved):

.github/workflows/azure-saml-ses-deploy-test.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Resolve latest snapshot tag
+        run: source cloud/shared/bin/resolve-latest-snapshot-tag
+
+      - name: Copy the config from staging 
+        run: |
+          #  Grab the config from the staging deploy repo to confirm
+          #  the changes work with actual data
+          curl https://raw.githubusercontent.com/civiform/staging-azure-deploy/main/civiform_config.sh > civiform_config.sh
+
+      - name: Source config, set to be test and deploy with latest snapshot tag
```
kmcnellis, line 36: [nit] "set to be test" -> "set to test mode"

kmcnellis (approved):

PR #2490, opened 2022-05-13
Auth API calls

bion:
Implements security logic for authenticating and authorizing API requests. Introduces a controller for the first API endpoint as well, but only authorizes the request --  will implement the controller action's business logic in the next PR.

- [x] make the cache API-key specific
- [x] configure the authenticator to match all `/api/` URL paths
- [x] wire up to the JSON exporter
- [x] test
- [x] document


shanemc-goog:
I reran the unit tests and they still fail btw.  Lots of random DB issues, no clear what might be the issue.

bion:
@shanemc-goog shucks, I thought I had them fixed! I'm pretty sure they're failing due to me adding a test that creates a fake application directly with guice and maybe doesn't clean up after itself properly. I'll look into when I get back from the dentist.

Otherwise this is ready for review.

bion:
@shanemc-goog unit tests are in good shape now

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,102 @@
...
+    // requests with the invalid key do not put pressure on the database.
+    Optional<ApiKey> maybeApiKey =
+        syncCacheApi.getOrElseUpdate(
+            keyId, () -> apiKeyService.get().findByKeyId(keyId), CACHE_EXPIRATION_TIME_SECONDS);
+
+    if (!maybeApiKey.isPresent()) {
+      throw new BadCredentialsException("API key does not exist");
+    }
+
+    ApiKey apiKey = maybeApiKey.get();
```
kmcnellis, line 77: [optional suggestion] You can simplify this into 
ApiKey apiKey = syncCacheApi.getOrElseUpdate(
            keyId, () -> apiKeyService.get().findByKeyId(keyId), CACHE_EXPIRATION_TIME_SECONDS)
           .orElseThrow(() => new BadCredentialsException("API key does not exist"));

server/app/controllers/api/ProgramApplicationsApiController.java
```diff
@@ -0,0 +1,25 @@
...
+
+/** API controller for admin access to a specific program's applications. */
+public final class ProgramApplicationsApiController extends CiviFormApiController {
+
+  @Inject
+  public ProgramApplicationsApiController(ProfileUtils profileUtils) {
+    super(profileUtils);
+  }
+
+  // Introduced to test API authentication logic. Implementation is next. Did not
```
kmcnellis, line 18: [nit] This would be a good place for a TODO w/ issue number

server/app/modules/SecurityModule.java
```diff
@@ -145,53 +152,134 @@ protected FakeAdminClient fakeAdminClient(ProfileFactory profileFactory) {
     return new FakeAdminClient(profileFactory, this.configuration);
   }
 
+  // CiviForm uses HTTP basic auth for authenticating API calls. The username in the basic auth
+  // credentials is the API
```
kmcnellis, line 156: [nit] this can be on 2 lines instead of 3

server/test/app/ApiAuthenticationTest.java
```diff
@@ -0,0 +1,84 @@
...
+                    controllers.api.routes.ProgramApplicationsApiController.list("mock-program")
+                        .url())
+                .remoteAddress("1.1.1.1")
+                .header("Authorization", "Basic " + creds));
+
+    assertThat(result.status()).isEqualTo(HttpConstants.OK);
+  }
+
+  @Test
+  public void apiRoutes_invalidKey() {
```
kmcnellis, line 64: Do we have an equivalent test for UNAUTHORIZED status for non-API paths?

server/test/auth/ApiAuthenticatorTest.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Test
+  public void validate_invalidSecret() {
+    var rawCredentials = keyId + ":" + "notthesecret";
+
+    assertBadCredentialsException(
+        buildFakeRequest(rawCredentials),
+        new UsernamePasswordCredentials(keyId, "notthesecret"),
+        "Invalid secret");
```
kmcnellis, line 148: I would think that we should just return an invalid key response to this case (and maybe some of the others) - no need to give extra information to someone who may be trying to break into the system

kmcnellis (approved):

server/app/auth/ApiAuthenticator.java
bion: done

server/app/modules/SecurityModule.java
bion: done

server/test/app/ApiAuthenticationTest.java
bion: yep SecurityBrowserTest

server/test/auth/ApiAuthenticatorTest.java
bion: The message in these exceptions isn't included in the response, just the server logs:

```
*   Trying 127.0.0.1:9000...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /api/v1/admin/programs/mock-program/applications HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.81.0
> Accept: */*
> Authorization: Basic TTVBK25FVlUyMlZzbjdPSFF2QWZIQT09OlVWSVAzd0ExZFBaMDhSNFZSTjY5djFOTExKUVFlTnBvQXQ5N0xjU3A4ZUE9
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Date: Tue, 17 May 2022 23:32:48 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```

server/app/controllers/api/ProgramApplicationsApiController.java
bion: done

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+ * the API key to perform authentication (this step).
+ *
+ * <p>Note that at this layer, the request is authenticated, not authorized. All API requests that
+ * reach a controller are already authenticated, but it is the controller's responsibility to check
+ * that the resource being accessed is authorized for the request's API key.
+ *
+ * <p>The API key ID and secret are provided as the basic auth username and password, respectively.
+ * To be valid, a request must reference a valid API key with its username credential, and further
+ * that key must:
+ * <li>Not be retired.
```
shanemc-goog, line 36: You need a \<ul\>\</ul\> wrapper, HTML in docs should be well formatted.

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+import play.cache.SyncCacheApi;
+import services.apikey.ApiKeyService;
+
+/**
+ * Authenticator for API requests based on HTTP basic auth and backed by the {@link ApiKey}
+ * resource. Background: https://en.wikipedia.org/wiki/Basic_access_authentication
+ *
+ * <p>When referenced by a request, {@code ApiKey}s are stored in the "api-keys" named cache, keyed
+ * by their key ID. This allows controller code to load the key without the need for a subsequent
+ * database query. This is necessary because pac4j creates profiles in the next step, but we need
```
shanemc-goog, line 26: "This is necessary because..."

I'm not following how this is a justification for the cache, especially since we read from the DB and write to the cache from this code which is seemingly ok despite the comment that profiles coming next is a problem.

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Override
+  public void validate(Credentials rawCredentials, WebContext context, SessionStore sessionStore) {
+    if (!(rawCredentials instanceof UsernamePasswordCredentials)) {
+      throw new RuntimeException("ApiAuthenticator must receive UsernamePasswordCredentials.");
+    }
+
+    UsernamePasswordCredentials credentials = (UsernamePasswordCredentials) rawCredentials;
+    String keyId = credentials.getUsername();
```
shanemc-goog, line 63: Are we overloading credentials in a confusing way by making a username and keyId synonymous?

Is there a sketch of this in your design doc?

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+ * that key must:
+ * <li>Not be retired.
+ * <li>Have an expiration date in the future.
+ * <li>Have a subnet that includes the IP address of the request.
+ * <li>Have a salted key secret that matches the salted password in the request's basic auth
+ *     credentials.
+ */
+public class ApiAuthenticator implements Authenticator {
+
+  private static final int CACHE_EXPIRATION_TIME_SECONDS = 60;
```
shanemc-goog, line 44: Comment why 60 seconds so future people have something to go on if/when it needs adjustment.

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+      throw new BadCredentialsException("API key is expired");
+    }
+
+    SubnetUtils allowedSubnet = new SubnetUtils(apiKey.getSubnet());
+    // Setting this to true includes the network and broadcast addresses.
+    // I.e. /31 and /32 will not be considered included in the subnetwork
+    // if this is false.
+    allowedSubnet.setInclusiveHostCount(true);
+
+    if (!allowedSubnet.getInfo().isInRange(context.getRemoteAddr())) {
```
shanemc-goog, line 88: Probably over engineering at this point, but want to toss something out.

These errors /1

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+    UsernamePasswordCredentials credentials = (UsernamePasswordCredentials) rawCredentials;
+    String keyId = credentials.getUsername();
+
+    // Cache the API key for quick lookup in the controller, also for subsequent requests.
+    // We intentionally cache the empty optional rather than throwing here so that subsequent
+    // requests with the invalid key do not put pressure on the database.
+    ApiKey apiKey =
+        syncCacheApi
+            .getOrElseUpdate(
+                keyId, () -> apiKeyService.get().findByKeyId(keyId), CACHE_EXPIRATION_TIME_SECONDS)
```
shanemc-goog, line 71: Suggestion: line break each param and the lambda make it hard to read if the expiry is part of it or the getOrElse

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,97 @@
...
+            .orElseThrow(() -> new BadCredentialsException("API key does not exist"));
+
+    if (apiKey.isRetired()) {
+      throw new BadCredentialsException("API key is retired");
```
shanemc-goog, line 75: If these only show up in the logs then I'd include the key so the sys Admin can do something actionable with them: eg identify keys that should work but aren't, or find bad actors request info.

server/app/auth/CiviFormHttpActionAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+import org.pac4j.core.context.WebContext;
+import org.pac4j.core.exception.http.HttpAction;
+import org.pac4j.play.PlayWebContext;
+import org.pac4j.play.http.PlayHttpActionAdapter;
+import play.mvc.Result;
+import play.mvc.Results;
+
+/**
+ * The action adapter allows pac4j-consuming code to intervene and change the default behavior of
+ * handling a given HTTP result. The default {@link PlayHttpActionAdapter} does not provide a way of
```
shanemc-goog, line 13: result -> request?

server/app/auth/CiviFormHttpActionAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+ * customizing behavior beyond mapping status codes to {@link Result} objects. In CiviForm's case,
+ * we want to handle unauthorized requests to the API differently, since there is not a UX
```
shanemc-goog, line 15: Suggestion: write the doc with the WHAT first and WHY second.

This reads as a justification for it's existence primarily without a lot of clear explanation what its actually doing.

shanemc-goog (commented):
some comments while I had time.

server/app/auth/CiviFormHttpActionAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+  public Result adapt(final HttpAction action, final WebContext context) {
+    if (isUnauthorizedApiRequest(action, context)) {
+      return ((PlayWebContext) context).supplementResponse(Results.unauthorized());
+    }
+
+    return super.adapt(action, context);
+  }
+
+  private static boolean isUnauthorizedApiRequest(HttpAction action, WebContext context) {
+    return action.getCode() == HttpConstants.UNAUTHORIZED
```
shanemc-goog, line 32: ?  Is code = UNAUTHORIZED coming from the ErrorHandler change?

server/test/auth/ApiAuthenticatorTest.java
kmcnellis: 👍

kmcnellis: Perhaps we can add that to the validate function documentation

server/app/controllers/api/CiviFormApiController.java
```diff
@@ -0,0 +1,31 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import auth.AccountNonexistentException;
+import auth.ApiKeyGrants;
+import auth.ProfileUtils;
+import auth.UnauthorizedApiRequestException;
+import controllers.CiviFormController;
+import play.mvc.Http;
+
+/** Base class for controllers that handle API requests */
```
shanemc-goog, line 12: This has no abstract methods so it's not really a base class, it's just a class that like all others can be instantiated and expanded.

Do you actually want an interface to reference implementations generically by; or something uninstantiable by making the constructors publicly inaccessible.

Given what I see so far I think you should make the default ctor and the ctor you have protected.

server/app/auth/ApiAuthenticator.java
bion: ah thanks, doc HTML confuses me with its lack of closing tags

server/app/auth/ApiAuthenticator.java
bion: update the last sentence of this paragraph to "This is necessary because pac4j creates profiles (which are what's made available to controller code through the request object) in the next step, but we need the API key to perform authentication (this step). The purpose of the cache is to avoid making multiple database calls to retrieve the API key throughout the cycle of the request."

server/app/auth/ApiAuthenticator.java
bion: The design specifies that we're using basic auth to authenticate API requests. "username" and "password" is the parlance basic auth uses for describing those values. Our internal consumption of them as "key ID" and "key secret" matches the CiviForm concepts, rather than the protocol ones.

server/app/auth/ApiAuthenticator.java
bion: done

server/app/auth/CiviFormHttpActionAdapter.java
bion: done

server/test/support/ResourceCreator.java
```diff
@@ -29,6 +32,21 @@ public ResourceCreator(Injector injector) {
     ProgramBuilder.setInjector(injector);
   }
 
+  public ApiKey createActiveApiKey(String name, String keyId, String keySecret) {
+    ApiKey apiKey =
+        new ApiKey()
+            .setName(name)
+            .setKeyId(keyId)
+            .setExpiration(Instant.now().plusSeconds(1))
```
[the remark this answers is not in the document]
bion: increased to 100

server/test/auth/ApiAuthenticatorTest.java
```diff
@@ -0,0 +1,173 @@
...
+  private static final String validRawCredentials = keyId + ":" + secret;
+  private static final SessionStore MOCK_SESSION_STORE = Mockito.mock(SessionStore.class);
+  private ApiKey apiKey;
+
+  @Before
+  public void setUp() {
+    // Most test classes extend ResetPostgres which accesses an injector via the
+    // application provided fakeApplication(). That injector is a play.inject.Injector
+    // which makes it much less easy to inject things that are annotated with annotations
+    // that take arguments, such as @NamedCache("api-cache'). Guice makes that fairly
```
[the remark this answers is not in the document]
bion: thanks, done

server/test/app/ApiAuthenticationTest.java
```diff
@@ -0,0 +1,84 @@
...
+    apiKey.save();
+  }
+
+  @Test
+  public void nonApiRoutes_doNotRequireApiAuth() {
+    Result result =
+        doGetRequest(
+            fakeRequest(
+                "GET", controllers.routes.HomeController.loginForm(Optional.empty()).url()));
+    System.out.println(result.headers());
```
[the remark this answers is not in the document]
bion: nope, thanks!

server/app/auth/ApiAuthenticator.java
shanemc-goog: Oh, like missing </p> ? Yeah, I'm not sure what's up with that.  I don't believe it's wrong to have them, but yeah doesn't seem required...

I know in general you don't need them for things like <li> because you can't compose <li>s so they can be implicitly understood to be a sibling of the previous and not a child.

server/test/support/ResourceCreator.java
```diff
@@ -37,7 +37,7 @@ public ApiKey createActiveApiKey(String name, String keyId, String keySecret) {
         new ApiKey()
             .setName(name)
             .setKeyId(keyId)
-            .setExpiration(Instant.now().plusSeconds(1))
+            .setExpiration(Instant.now().plusSeconds(100))
```
shanemc-goog, line 40: time based values like this in test setup can lead to inconsistent results, eg flakey tests if the real time is being used etc etc.

I'd suggest setting this to a "really long time"/infinite in the default case and set it shorter/adjust the test clock/etc in the tests that care about expiration.

Also comment that it effectively never expires.

server/app/auth/ApiAuthenticator.java
shanemc-goog: re: the slack convo, can you doc this here so it's clearer why this is happening and that it is ok despite looking odd.

server/app/auth/ApiAuthenticator.java
shanemc-goog: Looks the same, does the formatter not allow for it?

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,116 @@
...
+    // Cache the API key for quick lookup in the controller, also for subsequent requests.
+    // We intentionally cache the empty optional rather than throwing here so that subsequent
+    // requests with the invalid key do not put pressure on the database.
+    ApiKey apiKey =
+        syncCacheApi
+            .getOrElseUpdate(
+                keyId, () -> apiKeyService.get().findByKeyId(keyId), CACHE_EXPIRATION_TIME_SECONDS)
+            .orElseThrow(() -> new BadCredentialsException("API key does not exist"));
+
+    if (apiKey.isRetired()) {
```
shanemc-goog, line 91: ? Per the discussion you had with Kiana, would it be appropriate here to comment that the exception messages don't make it to the user?  

IE: just making it clear that this isn't informative to the user, or indicative of the particular failure to them.

server/app/auth/ProfileFactory.java
```diff
@@ -50,6 +57,13 @@ public CiviFormProfile wrapProfileData(CiviFormProfileData p) {
     return new CiviFormProfile(dbContext, httpContext, p);
   }
 
+  public Optional<ApiKey> retrieveApiKey(String keyId) {
```
shanemc-goog, line 60: Re the slack discussion, might be nice to add a comment in this file for why this makes sense, and also to dissuade adding profile unrelated code here (which this may appear to be)

server/app/auth/ProfileUtils.java
```diff
@@ -43,6 +45,22 @@ public Optional<CiviFormProfile> currentUserProfile(WebContext webContext) {
...
   }
 
+  public Optional<ApiKey> currentApiKey(Http.RequestHeader request) {
+    PlayWebContext webContext = new PlayWebContext(request);
+    return currentApiKey(webContext);
+  }
+
+  public Optional<ApiKey> currentApiKey(WebContext webContext) {
+    ProfileManager profileManager = new ProfileManager(webContext, sessionStore);
+    Optional<BasicUserProfile> p = profileManager.getProfile(BasicUserProfile.class);
```
shanemc-goog, line 55: Suggestion: p -> profile

The public guide only addresses this for [param names](https://google.github.io/styleguide/javaguide.html#s5.2.6-parameter-names), but I think in general avoiding arbitrary abbreviations is best for readability.

server/app/models/ApiKey.java
```diff
@@ -94,6 +94,11 @@ public Instant getExpiration() {
     return expiration;
   }
 
+  /** True if the supplied time is after the expiration date. */
```
shanemc-goog, line 97:
```suggestion
  /** True if the key is expired after {@code instant}. */
```

server/app/models/ApiKey.java
```diff
@@ -94,6 +94,11 @@ public Instant getExpiration() {
...
+  public boolean expiredAfter(Instant instant) {
```
shanemc-goog, line 98: ? Do we need to message anything about the key being UTC and this should be too?

server/build.sbt
```diff
@@ -9,6 +9,7 @@ lazy val root = (project in file("."))
     scalaVersion := "2.13.8",
     maintainer := "uat-public-contact@google.com",
     libraryDependencies ++= Seq(
+      caffeine,
```
shanemc-goog, line 12: Suggestion:  inline doc that this provides caching utilities.

server/conf/application.conf
```diff
@@ -274,10 +274,24 @@ play.ws {
...
-  #bindCaches = ["db-cache", "user-cache", "session-cache"]
+  # Specific caches can be injected using the @NamedCache annotation.
+  bindCaches = ["api-keys"]
 }
 
+## Security rules
+# https://github.com/pac4j/play-pac4j/wiki/Apply-security#2-protect-urls-via-the-securityfilter
+# ~~~~~
+# Specify authentication clients for request URL paths that match the regex.
+# If no regex matches, uses the default clients specified in SecurityModule.
```
shanemc-goog, line 285: Suggestion: maybe tweak the phrasing: "uses the default clients" as I find it a little unclear if the below 5 lines of config are doing this or if it's an implicit aspect of our overall setup.

There's no grammatical subject before "uses" so the reader has to mentally fill it in with Pac4j, CiviForm, the 5 lines below etc.

server/app/auth/ApiAuthenticator.java
```diff
@@ -0,0 +1,116 @@
...
+
+  // The cache expiration time is intended to be long enough reduce database queries from
+  // authenticating API calls while being short enough that if an admin retires a key or
+  // otherwise edits it, the edits will take effect within a reasonable amount of time.
+  // When tuning this value, consider the use-case of an API consumer rapidly paging through
+  // a list API, and also consider the admin retiring a key when they've discovered it has
+  // been compromised.
+  private static final int CACHE_EXPIRATION_TIME_SECONDS = 60;
+
+  private final SyncCacheApi syncCacheApi;
```
shanemc-goog, line 57: Include the Optional doc here as it's defining the "api" of this member, which is otherwise buried internally in 1 method.

server/test/auth/ApiAuthenticatorTest.java
```diff
@@ -0,0 +1,177 @@
...
+import play.mvc.Http;
+import play.test.Helpers;
+import support.ResourceCreator;
+
+public class ApiAuthenticatorTest {
+  @Inject Application application;
+
+  @Inject
+  @NamedCache("api-keys")
+  SyncCacheApi cacheApi;
```
shanemc-goog, line 41: Put the Optional.empty() doc here too as it defines the API.

---
Not a blocker, just a comment, but I wonder if we should wrap this in a thin class so that semantic can be in 1 authoritative place, an Inject that instead.

shanemc-goog (changes requested):
Looks good, mostly doc comments.

server/app/auth/ProfileFactory.java
```diff
@@ -18,15 +22,18 @@ public class ProfileFactory {
...
   private Provider<VersionRepository> versionRepositoryProvider;
+  private SyncCacheApi syncCacheApi;
 
   @Inject
   public ProfileFactory(
       DatabaseExecutionContext dbContext,
       HttpExecutionContext httpContext,
-      Provider<VersionRepository> versionRepositoryProvider) {
+      Provider<VersionRepository> versionRepositoryProvider,
+      @NamedCache("api-keys") SyncCacheApi syncCacheApi) {
```
shanemc-goog, line 32: Optional.empty() comment here too as it's part of the api we're adding to the built in cache.

shanemc-goog (changes requested):

server/app/auth/ApiAuthenticator.java
bion: must not then, i adjusted it and then ran the formatter before pushing

server/app/modules/SecurityModule.java
```diff
@@ -145,53 +152,133 @@ protected FakeAdminClient fakeAdminClient(ProfileFactory profileFactory) {
...
+    var actionAdapter = new CiviFormHttpActionAdapter();
+
+    actionAdapter
+        .getResults()
+        .putAll(
+            ImmutableMap.of(
+                // Redirect unauthorized requests to the login page. This behavior is bypassed
+                // for API requests in CiviFormHttpActionAdapter.
+                HttpConstants.UNAUTHORIZED,
+                redirect(routes.HomeController.loginForm(Optional.of("login"))),
```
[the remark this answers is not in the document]
jeanniefu: Oh ok gotcha, now I see the refactoring below.

jeanniefu (approved):

server/app/auth/ApiAuthenticator.java
bion: added a comment

server/test/support/ResourceCreator.java
bion: changed it to `Instant.MAX` and added a docstring for the method

server/app/auth/ApiAuthenticator.java
bion: added a paragraph to the end of the class doc

server/app/auth/ProfileFactory.java
bion: doc'd the method

server/app/auth/ProfileUtils.java
bion: done

server/app/models/ApiKey.java
bion: updated the method doc

server/build.sbt
bion: done

server/conf/application.conf
bion: updated

server/test/support/ResourceCreator.java
bion: turns out Instant.MAX is out of range for a postgres timestamp, switched to making a year in the future and used a constant.

server/test/auth/ApiAuthenticatorTest.java
bion: moved caching logic into the ApiKeyService with discussion of caching invalid key IDs

server/test/auth/ApiAuthenticatorTest.java
```diff
@@ -0,0 +1,177 @@
...
+    apiKey.retire("test-user");
+    apiKey.save();
+
+    assertBadCredentialsException(
+        buildFakeRequest(validRawCredentials), "API key is retired: " + keyId);
+  }
+
+  @Test
+  public void validate_isExpired() {
+    apiKey.setExpiration(Instant.now().minusSeconds(100));
```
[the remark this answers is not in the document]
bion: added comment

shanemc-goog (approved):

Issue #1238, opened 2021-05-21
When CiviForm admin creates a program, make it easier to find the program in the list


neel-a-mehta:
Right now, once you create a program, you go back to the page with a list of programs, but your new program is way at the bottom of the list -- so it's very hard to find the program. Can be confusing for admins.

Potential workarounds:

- Once you create a program, go straight to the page to edit it (or manage translations, whichever is a more natural landing point)
- Once you create a program, scroll down to the program's location in the list view
- Sort the list of programs by editing time, descending, so that the most recently added/edited programs are at the top of the list. (This is the most scalable solution IMO, and it'll also let us expand into other methods of sorting the programs list later.)

glorialiou:
Related issues:
https://github.com/seattle-uat/civiform/issues/942

Related issues (but for the question list / submitted applications list -- could be helpful to think about lists all together):
https://github.com/seattle-uat/civiform/issues/1912
https://github.com/seattle-uat/civiform/issues/1949
https://github.com/seattle-uat/civiform/issues/944

clouser2:
@rahaghas and I discussed this a bit today, and some immediate improvements we can make are to:

1) when creating a new program and displaying the list view, add a link to the newly created program in a toast
2) Consider auto-sorting the list by the most recently updated, so that the newly added program will be at the top

We'll also be looking into the redesign as well. The above will hopefully help make the editing experience smoother while we work on the redesign.

clouser2:
After investigating a bit, our current infrastructure for toast messages only supports showing a string, not a secondary link as well. Plumbing this through may take a bit of work.

clouser2:
Update: I've been focusing mainly on getting the UI mocks from https://github.com/seattle-uat/civiform/issues/1238#issuecomment-1106908001 in place. We'll likely start incrementally developing this (guarding the new UI with a query param until we've stabilized it). We're going to proceed somewhat incrementally:
  * Change layout of program cards generally for the new styling
  * Add sorting shortly after this

Currently, this page doesn't have pagination (nor does it really need to at this point, in my opinion). When implementing sort / filtering, rather than implementing all of this as server-side endpoints that require a page refresh, we can likely just do the sorting client-side in JS instead. Given that the number of programs likely won't be a scaling bottleneck, we can push this to the client and not have any need for sorting the DB entries server-side or implementing pagination.

[...]

PR #2499, opened 2022-05-17
Order programs by the last modification date in the program admin view

### LINKED ISSUE #2513: "Last updated" timestamps aren't updated for draft version when "publish all drafts"
**Describe the bug**
When clicking "publish all drafts" with a new program, the last updated timestamp doesn't update, even though it does for already published programs.

Technically, this is working as intended from how the system is currently implemented.

**To Reproduce**
1. Go to staging.seattle.civiform.com
2. Create a new program. Observe the "last updated" time.
3. Wait 1 minute since we render timestamps at minute granularity
4. Publish All Drafts
5. Observe all already-published programs have a new "Last Updated" timestamp, while the newly created version does not

**Expected behavior**
The "Last Updated" timestamp is also updated for the draft version OR the timestamps for already published versions aren't updated.

**Screenshots**
Screen Shot 2022-05-18 at 3 42 07 PM

**Additional context**
This is due to the fact that we create a new program entity when associating an existing program with a new version ([code](https://github.com/seattle-uat/civiform/blob/75bc407f70d51feb4e9e8d832aff8f197ca34cab/server/app/repository/VersionRepository.java#L82)).

A quick workaround is to also perform a save on the draft program versions as well.

clouser2:
### Description
Previously, the program list had no explicit ordering. In practice, this meant that programs were ordered based on their database ID. This meant that the most recently created programs appeared at the end of the list.

As part of #1238, we'll be updating the program list UI and potentially adding sorting options to it. This starts by applying a default sort order of last modified, hopefully helping to ensure that the most recently edited programs are at the top of the list.

Note: This sorting is done client-side per the rationale in https://github.com/seattle-uat/civiform/issues/1238#issuecomment-1127884289. The newly added admin_programs.ts file can grow to include additional sorting options as part of the work in #1238.

When introducing this sorting, two suprises occurred related to the "publish all drafts" functionality:
1. The last updated timestamps are based on when `VersionRepository.publishNewSynchronizedVersion` sequentially updates each version ([code](https://github.com/seattle-uat/civiform/blob/75bc407f70d51feb4e9e8d832aff8f197ca34cab/server/app/repository/VersionRepository.java#L55)). Due to this, programs have slightly different timestamps. I ordered the updates so that programs were updated from least recent to most recent, thus preserving the relative ordering before "publish all drafts" was invoked. Added a regression test for this
2. `VersionRepository.publishNewSynchronizedVersion` didn't update the timestamp for draft programs, but did create new program entities to associate already-published programs with the draft version (#2513). This led to already-published programs always appearing before the draft program. Here, the issue is fixed by performing a save on draft programs to ensure the timestamp is updated.

### Checklist

### Issue(s)
#1238; Fixes #2513


server/app/assets/javascripts/admin_programs.ts
```diff
@@ -0,0 +1,58 @@
...
+    )
+    if (!cardsParent) {
+      return
+    }
+    try {
+      this.sortCards(cardsParent as HTMLElement)
+    } finally {
+      // Make sure to always show the cards, even
+      // if there was an error while sorting.
+      cardsParent.classList.remove('invisible')
```
kmcnellis, line 19: nit: If this is invisible at first, it would be good to have a placeholder that says "loading" or something.  That way if the user doesn't have js enabled, it won't just look like there are 0 programs

kmcnellis (approved):

bion:
@clouser2 browser tests are failing

server/app/assets/javascripts/admin_programs.ts
clouser2: Added a placeholder since it doesn't hurt anything to have. RE JS enablement, I spoke with @bion about this a bit back, who indicated that we can assume that JS is required for CiviForm.

clouser2:
> @clouser2 browser tests are failing

Indeed. This was another issue that only reproed in CI. CI tests seem to have the database seeded with initial values. I've changed the assertions to only look at the top N programs, with the assumption being that the programs added by my tests will be at the top.

kmcnellis (approved):

clouser2:
> > @clouser2 browser tests are failing
> 
> Indeed. This was another issue that only reproed in CI. CI tests seem to have the database seeded with initial values. I've changed the assertions to only look at the top N programs, with the assumption being that the programs added by my tests will be at the top.

Actually, CI gave us a real issue that's a bit tough. When "Publish all drafts" is clicked, the each program version is updated. As such, the "Last modified" timestamp moves forward. However, since this isn't done transactionally, each program version has a distinct timestamp, which could result in a semi-random ordering. That's why we're seeing the issue above.

Some ideas for how to approach this:
* When "publishing all drafts", do this in a single transaction so that all last updated timestamps are the same and we could fall back on a secondary sort. Still non-ideal since the list order will shift from the previous state of last updated to whatever the secondary sort is (currently alphabetical)
* Don't update the "last updated" field for something like publish all drafts. Only update it when something has actually changed about the program (e.g. edits by the admin).

Until we've resolved the above, I'm moving this to draft. We'll also need to resolve this issue prior to adding the sorting described in #1238.

kmcnellis:
It's a convenient little workaround you found!   Though I'm probably more in favor of another field / only updating the timestamps when the program _actually_ changes, this works for me.

server/app/assets/javascripts/admin_programs.ts
bion: especially for the admin view we can make convenient assumptions: JS is enabled, they're on desktop not mobile, decent internet connection, etc

bion (approved):

clouser2:
> It's a convenient little workaround you found! Though I'm probably more in favor of another field / only updating the timestamps when the program _actually_ changes, this works for me.

Agreed that only updating timestamps when edits are actually made is definitely desirable. We'd discussed having a distinct timestamp that we manually update each time the admin performs an action. The tradeoff there is that we're responsible for making sure to update it for all paths manually (as opposed to it happening automatically with the current @WhenModified) behavior.

My opinion is that this work won't be high on the priority list unless we have more use cases in mind for a more accurate last modified timestamp.

clouser2:
This was reverted in #2520 since we were seeing unit and integration tests fail in CI. I couldn't reproduce this on my local machine, but @gwendolyngoetz was able to on an Ubuntu machine that was much speedier than mine.

When we printed out information about the relevant timestamps, we saw that the `last_modified_time` appeared to only be at millisecond granularity, even though we define the property as a Java `Instant`, which has nanosecond granularity ([code](https://github.com/seattle-uat/civiform/blob/d75b1b418703bbe7e31756922f38a727fe54f218/server/app/models/Program.java#L93)). This matters due to the fact that the timestamp for all programs is updated when "Publish all drafts" is invoked, since updating the M2M table triggers [EBean to trigger an update on the Program entity as well](https://github.com/seattle-uat/civiform/issues/2513#issuecomment-1133000483). Millisecond granularity isn't sufficient for these types of operations.

Why are we seeing millisecond granularity even though an `Instant` carries more granularity? When EBean is generating a new value for fields with @WhenModified present, it appears to retrieve the current time only at millisecond precision ([code](https://github.com/ebean-orm/ebean/blob/daf5220766c38c68fafd448039b98fb3645a6cdb/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java#L491)). EBean's strategy for generating the timestamps is entirely based on getting input as UTC millis ([code](https://github.com/ebean-orm/ebean/blob/539dd7c90c71e056e718368ccba90ceba3713dbe/ebean-core/src/main/java/io/ebeaninternal/server/deploy/generatedproperty/UpdateTimestampFactory.java#L28)). This isn't something we're going to be able to change.

As such, the strategy to have "publish all drafts" update timestamps in the same relative order is insufficient, since the current granularity level will cause many updates to be sorted in the same order. We could add a secondary sort, but that's not desirable since 1) it'll still look like a random order to the user and 2) it's not possible to write a non-flaky test.

As I see it, there are potentially three avenues we could pursue:
1. Use a distinct column for last edited timestamp that we manually control. Since we generate it, it should be serialized with nanosecond resolution. It has all of the challenges described [here](https://github.com/seattle-uat/civiform/pull/2499#issuecomment-1133060671).
2. See if we can disable the timestamp being updated when the M2M mapping is updated. This seems like an unnecessary and incorrect DB write. That said, it might be a bug in EBean that's not easily fixable.
3. Resubmit this change, accept that sometimes the ordering will be a bit strange, and remove the tests originally bundled with it since they can't be made non-flaky.

Option 3 is the least desirable in my opinion, but also may be sufficient. I'll spend a bit of time investigating option 2 since it seems like a real bug.

PR #2503, opened 2022-05-17
Adding api secret salt to cloud formation container configuration

gwendolyngoetz:
The test related to the creating a API key fails against staging when this is not set.

The new secret for staging, named _staging-api-secret-salt_, has been added in AWS.

My first time making changes to cloud formation so please double check I'm setting this correct for what is in the [application.conf](https://github.com/seattle-uat/civiform/blob/b7125166d20774d64f20d56dc720403923f2470d/server/conf/application.conf#L495) file.


bion (approved):
LGTM, but the cloudformation check failure is unclear

gwendolyngoetz:
That's the same error Kiana was having a few weeks ago. It's failing when it's trying to add a comment to the pull request. This started after the action permissions were changed to read-all. We can try giving the specific job `pull-requests: write` permissions otherwise I think we have to remove it.

PR #2507, opened 2022-05-18
fix browser test env debug port

syntapy:
### Description
Browser test env [outter](https://github.com/seattle-uat/civiform/blob/4ed410e13653026f28b132bf4b088de320af8305/browser-test/browser-test-compose.dev.yml#L21) and [inner](https://github.com/seattle-uat/civiform/blob/4ed410e13653026f28b132bf4b088de320af8305/browser-test/browser-test-compose.dev.yml#L22) ports don't match, so this fixes that

I had to do this to get IntelliJ debugger to communicate with the container


bion (approved):

syntapy:
I just ran `bin/run-browser-tests-ci` locally and that passed.

PR #2509, opened 2022-05-18
Add aria-describedBy and aria-invalid attributes for FieldWithLabel errors.

### LINKED ISSUE #1878: question validation error messages are not readable to a screen reader

clouser2:
### Description
This pattern is described here: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21

Before: When using VoiceOver on Mac, a field with validation errors wouldn't be announced.
After: A field with validation errors is announced as "invalid data".

### Checklist

### Issue(s)
Fixes #1878

### DIFF
--- a/server/app/views/components/FieldWithLabel.java
+++ b/server/app/views/components/FieldWithLabel.java
@@ -252,6 +252,7 @@ public class FieldWithLabel {
     if (this.id.isEmpty()) {
       this.id = RandomStringUtils.randomAlphabetic(8);
     }
+    String fieldErrorsId = String.format("%s-errors", this.id);
     if (fieldTag.getTagName().equals("textarea")) {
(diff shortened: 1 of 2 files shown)


bion (approved):

jeanniefu (approved):
nit: Can you add a brief before and after behavior explanation in the PR description?

If I understand correctly, I think the new behavior is that when the user is in the input field, the screen reader will (1) announce that the data is invalid (based on your [provided link](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21)), and (2) announce the error message (based on [this link](https://www.tpgi.com/using-aria-describedby-to-provide-helpful-form-hints/#:~:text=When%20a%20screen%20reader%20focuses,waiting%20for%20the%20rest.)).

PR #2515, opened 2022-05-19
Application list API

bion:


server/app/services/PaginationSpec.java
```diff
@@ -1,23 +1,30 @@
...
+
 /** Specifies pagination behavior for a paginated query. */
 public class PaginationSpec {
 
   public static PaginationSpec MAX_PAGE_SIZE_SPEC = new PaginationSpec(Integer.MAX_VALUE, 1);
 
-  private final int pageSize;
-  private final int currentPage;
+  protected final int pageSize;
+  protected final Optional<Integer> currentPage;
```
shanemc-goog, line 11: This is never used by the child so it doesn't need to be protected, and given it's never set by the child either this is likely another inheritance smell.

server/app/services/PaginationSpec.java
```diff
@@ -1,23 +1,30 @@
...
+    this.currentPage = Optional.of(currentPage);
   }
 
   public int getPageSize() {
     return this.pageSize;
   }
 
   public int getCurrentPage() {
-    return this.currentPage;
+    return this.currentPage.get();
```
shanemc-goog, line 28: This will throw an exception for the first ctor which is likely not desired or knowable by a caller.

server/app/repository/ProgramRepository.java
```diff
@@ -249,16 +265,29 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
...
 
-    PagedList<Application> pagedQuery =
-        query
-            .setFirstRow((paginationSpec.getCurrentPage() - 1) * paginationSpec.getPageSize())
-            .setMaxRows(paginationSpec.getPageSize())
-            .findPagedList();
+    PagedList<Application> pagedQuery;
+
+    if (paginationSpecEither.left.isPresent()) {
+      var paginationSpec = paginationSpecEither.left.get();
```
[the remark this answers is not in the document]
shanemc-goog: Also, given that left's type is a child of right, we should be able to accept the parent type and just use that instead of this pair construct.

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -254,14 +257,16 @@ public Result index(
...
+    PaginationResult<Application> applications;
     try {
-      PaginationResult<Application> applications =
+      applications =
           programService.getSubmittedProgramApplicationsAllVersions(
-              programId, new PaginationSpec(PAGE_SIZE, page.orElse(1)), search);
-
-      return ok(applicationListView.render(request, program, applications, search));
+              programId, F.Either.Right(paginationSpec), search);
     } catch (ProgramNotFoundException e) {
```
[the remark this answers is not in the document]
bion: yesss a good improvement. done.

server/app/repository/ProgramRepository.java
```diff
@@ -206,7 +209,8 @@ public ImmutableList<Program> getAllProgramVersions(long programId) {
    * Get all submitted applications for this program and all other previous and future versions of
    * it where the applicant's first name, last name, email, or application ID contains the search
    * query. Does not include drafts or deleted applications. Results returned in reverse order that
-   * the applications were created.
+   * the applications were created. Results are limited to applications submitted after and/or
```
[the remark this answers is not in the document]
bion: done

server/app/repository/ProgramRepository.java
```diff
@@ -227,6 +235,14 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
                 "lifecycle_stage",
                 ImmutableList.of(LifecycleStage.ACTIVE, LifecycleStage.OBSOLETE));
 
+    if (submitTimeFrom.isPresent()) {
+      query = query.where().gt("submit_time", submitTimeFrom.get());
```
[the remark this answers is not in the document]
bion: changed it for the from. most from - to semantics i've encountered are inclusive of from and exclusive of to.

> I'm unsure how this interact with pagination, if we ever expect a flow where someone takes the timestamp of the last result of one call, and passed that in as the start time for the next then we'll need to be a little more careful then.

since we can use ID as a proxy for submit time i don't think this will be an issue even if we allow more flexibility with sort order in the future.

server/app/repository/ProgramRepository.java
```diff
@@ -215,7 +219,11 @@ public ImmutableList<Program> getAllProgramVersions(long programId) {
    * application ID matching it.
    */
   public PaginationResult<Application> getApplicationsForAllProgramVersions(
-      long programId, PaginationSpec paginationSpec, Optional<String> searchNameFragment) {
+      long programId,
+      F.Either<OffsetBasedPaginationSpec<Long>, PaginationSpec> paginationSpecEither,
```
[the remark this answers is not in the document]
bion: done

server/app/repository/ProgramRepository.java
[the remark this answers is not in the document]
bion: removed the vars and the extends relationship

server/app/services/PaginationSpec.java
```diff
@@ -1,23 +1,30 @@
 package services;
 
+import java.util.Optional;
+
 /** Specifies pagination behavior for a paginated query. */
```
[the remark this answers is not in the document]
bion: updated the doc

server/app/services/OffsetBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
+package services;
+
+import java.util.Optional;
+
+/**
+ * Specifies pagination behavior for a query with offset identifier-based pagination. An offset
+ * identifier identifies the last item of the previous page of results using its sort order
+ * attribute.
+ */
+public class OffsetBasedPaginationSpec<T> extends PaginationSpec {
```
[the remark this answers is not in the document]
bion: done

server/app/services/OffsetBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
+package services;
+
+import java.util.Optional;
+
+/**
+ * Specifies pagination behavior for a query with offset identifier-based pagination. An offset
```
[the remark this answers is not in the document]
bion: done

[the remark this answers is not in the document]
bion: done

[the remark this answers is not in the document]
bion: done

server/app/services/PaginationSpec.java
bion: n/a

server/app/services/PaginationSpec.java
bion: n/a having removed the extends relationship

server/app/services/IdentifierBasedPaginationSpec.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private final int pageSize;
+  private final Optional<T> currentPageOffsetIdentifier;
+
+  public IdentifierBasedPaginationSpec(int pageSize) {
+    this.pageSize = pageSize;
+    this.currentPageOffsetIdentifier = Optional.empty();
+  }
+
+  public IdentifierBasedPaginationSpec(int pageSize, Optional<T> currentPageOffsetIdentifier) {
```
shanemc-goog, line 26: suggestion: Requiring an optional wrapper seems odd here given the other ctor exists.  I'd remove it unless there's a compelling use I don't see.

server/app/services/IdentifierBasedPaginationSpec.java
```diff
@@ -0,0 +1,38 @@
...
+    this.pageSize = pageSize;
+    this.currentPageOffsetIdentifier = checkNotNull(currentPageOffsetIdentifier);
+  }
+
+  public int getPageSize() {
+    return this.pageSize;
+  }
+
+  public Optional<T> getCurrentPageOffsetIdentifier() {
```
shanemc-goog, line 35: javadoc this, especially what an empty optional means.

server/app/services/PageNumberBasedPaginationSpec.java
```diff
@@ -0,0 +1,31 @@
+package services;
+
+import java.util.Optional;
+
+/** Specifies pagination behavior for a paginated query. */
```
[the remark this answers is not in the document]
bion: done

server/app/services/OffsetBasedPaginationSpec.java
```diff
@@ -0,0 +1,37 @@
+package services;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Optional;
+
+/**
+ * Specifies pagination behavior for a query with offset identifier-based pagination. An offset
```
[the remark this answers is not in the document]
bion: done

server/app/services/OffsetBasedPaginationSpec.java
```diff
@@ -0,0 +1,37 @@
...
+ * identifier identifies the last item of the previous page of results using its sort order
+ * attribute.
+ */
+public class OffsetBasedPaginationSpec<T> {
```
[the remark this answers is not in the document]
bion: any other type that could be used as a sort order. e.g. if we sorted accounts by authority_id we'd need a string identifier.

server/app/controllers/api/ApiPaginationTokenPayload.java
```diff
@@ -0,0 +1,70 @@
...
+    public PageSpec(
+        @JsonProperty("offsetIdentifier") String offsetIdentifier,
+        @JsonProperty("pageSize") int pageSize) {
+      this.offsetIdentifier = offsetIdentifier;
+      this.pageSize = pageSize;
+    }
+
+    /**
+     * The offset identifier identifies the last item of the previous page of results using its sort
+     * order attribute.
```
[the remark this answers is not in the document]
bion: it's referencing the attribute that's being used to sort the resource being paginated. in the case of this API, it's sorted by the application resource's ID attribute.

server/app/services/program/ProgramService.java
```diff
@@ -319,12 +332,32 @@ ImmutableList<Application> getSubmittedProgramApplications(long programId)
    * <p>If searchNameFragment is an unsigned integer, query will filter to applications with an
    * applicant ID matching it.
    *
-   * @param paginationSpec specification for paginating the results.
+   * @param paginationSpecEither either a OffsetBasedPagination or PaginationSpec spec.
```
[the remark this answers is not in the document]
bion: done

server/app/services/program/ProgramService.java
```diff
@@ -319,12 +332,32 @@ ImmutableList<Application> getSubmittedProgramApplications(long programId)
...
+      F.Either<IdentifierBasedPaginationSpec<Long>, PageNumberBasedPaginationSpec>
+          paginationSpecEither,
+      Optional<String> searchNameFragment)
+      throws ProgramNotFoundException;
+
+  /**
+   * Get all submitted applications for this program and all other previous and future versions of
+   * it where the application's submit time is in the specified range.
+   *
+   * @param paginationSpecEither either a OffsetBasedPagination or PaginationSpec spec.
```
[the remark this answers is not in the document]
bion: done

server/app/repository/ProgramRepository.java
```diff
@@ -249,16 +270,29 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
...
-            .findPagedList();
+    PagedList<Application> pagedQuery;
+
+    if (paginationSpecEither.left.isPresent()) {
+      IdentifierBasedPaginationSpec<Long> paginationSpec = paginationSpecEither.left.get();
+      pagedQuery =
+          query
+              .where()
+              .lt("id", paginationSpec.getCurrentPageOffsetIdentifier().orElse(Long.MAX_VALUE))
+              .setMaxRows(paginationSpec.getPageSize())
```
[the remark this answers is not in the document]
bion: > how do we grab the right entries? Does this guarantee order?

order is guaranteed on line 236 by `.orderBy("id desc")`

> If you set id < 10, and specify max rows as 5, will we get 5-9?

it depends on the specific value of `paginationSpec.getCurrentPageOffsetIdentifier()`

server/app/services/PageNumberBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
+package services;
+
+import java.util.Optional;
+
+/**
+ * Specifies pagination behavior for a query with page number-based pagination. With page-number
+ * based pagination, the current page of results is resolved by multiplying the page size by the
+ * current page - 1 to derive an offset for the the underlying query.
```
shanemc-goog:
```suggestion
 * current page - 1 to derive an offset for the underlying query.
```

server/app/services/PageNumberBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class PageNumberBasedPaginationSpec {
+
+  public static PageNumberBasedPaginationSpec MAX_PAGE_SIZE_SPEC =
```
shanemc-goog, line 13:
```suggestion
  public static final PageNumberBasedPaginationSpec MAX_PAGE_SIZE_SPEC =
```

server/app/services/PageNumberBasedPaginationSpec.java
shanemc-goog: Document in the class more that page is 1 based, not 0.

server/app/services/PageNumberBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
+package services;
+
+import java.util.Optional;
+
+/**
+ * Specifies pagination behavior for a query with page number-based pagination. With page-number
+ * based pagination, the current page of results is resolved by multiplying the page size by the
```
shanemc-goog, line 7: Suggestion: provide this calculation here, instead of listing the recipe for all users to then apply.   Is there any reason not to?

server/app/services/PageNumberBasedPaginationSpec.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public static PageNumberBasedPaginationSpec MAX_PAGE_SIZE_SPEC =
+      new PageNumberBasedPaginationSpec(Integer.MAX_VALUE, 1);
+
+  private final int pageSize;
+  private final Optional<Integer> currentPage;
+
+  public PageNumberBasedPaginationSpec(int pageSize) {
+    this.pageSize = pageSize;
+    this.currentPage = Optional.empty();
```
shanemc-goog, line 21: Is this constructor still necessary after removing the other class from being a child?

This will cause getCurrentPage() to throw an exception with no way to avoid it.

server/test/repository/ProgramRepositoryTest.java
```diff
@@ -202,7 +202,7 @@ public void getApplicationsForAllProgramVersions_withDateRange() {
     PaginationResult<Application> paginationResult =
         repo.getApplicationsForAllProgramVersions(
             program.id,
-            F.Either.Right(new PaginationSpec(10, 1)),
+            F.Either.Right(new PageNumberBasedPaginationSpec(10, 1)),
```
shanemc-goog, line 205: inline doc the params please since there's 2 and it's not readable how they relate.

Probably not necessary for the identifier one since it can read pretty easy with 1 value.

shanemc-goog (changes requested):
Reviewed it all, mostly just docs and nits

server/app/services/OffsetBasedPaginationSpec.java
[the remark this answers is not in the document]
jeanniefu: Ah I see. Optional, but maybe would be helpful to add a short comment with examples. I was assuming it would always be a number (but maybe that's just me :))

server/app/controllers/api/ApiPaginationTokenPayload.java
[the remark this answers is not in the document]
jeanniefu: Ah I see, thanks!

server/test/app/ApiAuthenticationTest.java
```diff
@@ -67,10 +62,7 @@ public void apiRoutes_invalidKey() {
 
     Result result =
         doGetRequest(
-            fakeRequest(
-                    "GET",
-                    controllers.api.routes.ProgramApplicationsApiController.list("mock-program")
-                        .url())
+            fakeRequest("GET", API_URL)
```
shanemc-goog, line 65: nit: API_URL -> AUTH_CHECK_URL

It doesn't read clearly in uses like there.

server/test/support/ResourceCreator.java
```diff
@@ -112,9 +112,17 @@ public Account insertAccount() {
   }
 
   public Applicant insertApplicantWithAccount() {
+    return insertApplicantWithAccount(Optional.empty());
```
shanemc-goog, line 115: inline doc the empty

server/test/support/ResourceCreator.java
```diff
@@ -112,9 +112,17 @@ public Account insertAccount() {
...
+  }
+
+  public Applicant insertApplicantWithAccount(Optional<String> accountEmail) {
     Applicant applicant = insertApplicant();
     Account account = insertAccount();
 
+    if (accountEmail.isPresent()) {
```
shanemc-goog, line 122: not necessary: but I always forget about ifPresent, and that it takes a consumer.  

accountEmail.ifPresent( account::setEmailAddress)

It def reads... different, so just mentioning it if you ever want to use it.

https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-

server/test/support/CfTestHelpers.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.mockito.Mockito.mockStatic;
+
+import com.google.common.collect.ImmutableMap;
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+import org.mockito.MockedStatic;
+import play.api.test.Helpers;
+
+public class CfTestHelpers {
```
shanemc-goog, line 12: There's only 1 class/test that use Cf in file names and the rest use CiviForm.  I'm fine with well known abbreviations but we don't seem to necessarily have that with Cf or use it..  Would we enforce using Cf going forward to establish this?

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import play.mvc.Http;
+import play.mvc.Result;
+import services.applicant.JsonPathProvider;
+import services.export.AbstractExporterTest;
+
+public class ProgramApplicationsApiControllerTest extends AbstractExporterTest {
+
+  private ApiKey apiKey;
```
shanemc-goog, line 28: style: move to after the static finals.

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+                /* nextPageToken= */ Optional.empty(),
+                /* pageSize= */ Optional.empty())
+            .url();
+
+    Result result = doRequest(requestUrl);
+    assertThat(result.status()).isEqualTo(HttpStatus.SC_OK);
+
+    DocumentContext resultJson = JsonPathProvider.getJsonPath().parse(contentAsString(result));
+    assertThat(resultJson.read("payload[0].application_id", Long.class))
+        .isEqualTo(applicationTwo.id);
```
shanemc-goog, line 62: Is the connection with the program returned and applicationTwo magic?

I don't see it mentioned in the test otherwise, or understand why it's applicationTwo and not applicationOne.  Can we make it less magic / errorprone ? :D

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertThat(resultJson.read("payload.length()", Integer.class)).isEqualTo(1);
```
shanemc-goog, line 63: Put this first so that read(payload[0]) doesn't give a confusing result on the evaluation above when this test would trigger for 0 results.

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+                /* nextPageToken= */ Optional.empty(),
+                /* pageSize= */ Optional.of(2))
+            .url();
+
+    Result result = doRequest(firstRequestUrl);
+    assertThat(result.status()).isEqualTo(HttpStatus.SC_OK);
+
+    DocumentContext resultJson = JsonPathProvider.getJsonPath().parse(contentAsString(result));
+    assertThat(resultJson.read("payload[0].application_id", Long.class))
+        .isEqualTo(applicationThree.id);
```
shanemc-goog, line 88: Related to the above, this magic is fragile.  Can we make the ordering constraint a clear part of the setup, or if there is an underlying invariant in the setup, comment that for readers of the test.

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertThat(resultJson.read("payload[1].application_id", Long.class))
+        .isEqualTo(applicationTwo.id);
+    assertThat(resultJson.read("payload.length()", Integer.class)).isEqualTo(2);
```
shanemc-goog, line 91: put above the payload data verification asserts

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertThat(result.status()).isEqualTo(HttpStatus.SC_OK);
+
+    DocumentContext resultJson = JsonPathProvider.getJsonPath().parse(contentAsString(result));
+    assertThat(resultJson.read("payload[0].application_id", Long.class))
+        .isEqualTo(applicationTwo.id);
+    assertThat(resultJson.read("payload.length()", Integer.class)).isEqualTo(1);
+  }
+
+  @Test
+  public void list_success_multiplePages() {
```
shanemc-goog, line 67: This is really two tests, and it'd be clearer to break them up.

1. multiple results in a page.
2. multiple pages (this doesn't need to verify the first page because of the above and existing tests.

shanemc-goog (changes requested):
Somehow the UI didn't include all the tests in my last pass

server/app/repository/ProgramRepository.java
```diff
@@ -209,18 +209,23 @@ public ImmutableList<Program> getAllProgramVersions(long programId) {
...
+   * the applications were created. Results are optionally limited to applications submitted after
+   * {@code submitTimeFrom} and/or before {@code submitTimeTo}, inclusive of both values.
    *
    * <p>If searchNameFragment is not an unsigned integer, the query will filter to applications with
    * email, first name, or last name that contain it.
    *
    * <p>If searchNameFragment is an unsigned integer, the query will filter to applications with an
    * application ID matching it.
+   *
+   * <p>Both offset-based and page number-based pagination are supported. For paginationSpecEither
```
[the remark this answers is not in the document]
bion: done

server/app/services/IdentifierBasedPaginationSpec.java
bion: good call, removed

server/app/services/IdentifierBasedPaginationSpec.java
bion: done

server/app/services/PageNumberBasedPaginationSpec.java
bion: done

bion: done and added to the getter

server/app/services/PageNumberBasedPaginationSpec.java
bion: done

server/app/services/PageNumberBasedPaginationSpec.java
bion: done and removed the optional

server/test/repository/ProgramRepositoryTest.java
bion: done

server/app/services/program/ProgramService.java
```diff
@@ -319,12 +332,34 @@ ImmutableList<Application> getSubmittedProgramApplications(long programId)
    * <p>If searchNameFragment is an unsigned integer, query will filter to applications with an
    * applicant ID matching it.
    *
-   * @param paginationSpec specification for paginating the results.
+   * @param paginationSpecEither either a OffsetBasedPagination or IdentifierBasedPaginationSpec
```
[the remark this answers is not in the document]
bion: done

server/app/services/program/ProgramService.java
```diff
@@ -319,12 +332,34 @@ ImmutableList<Application> getSubmittedProgramApplications(long programId)
...
+      throws ProgramNotFoundException;
+
+  /**
+   * Get all submitted applications for this program and all other previous and future versions of
+   * it where the application's submit time is in the specified range.
+   *
+   * @param paginationSpecEither either a OffsetBasedPagination or IdentifierBasedPaginationSpec
+   *     spec.
+   * @param submitTimeFrom specifies the oldest submission time to include.
+   * @param submitTimeTo specifies the newest submission time to include.
```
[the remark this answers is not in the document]
bion: done

server/build.sbt
```diff
@@ -43,7 +43,7 @@ lazy val root = (project in file("."))
 
       // Testing libraries
       "org.assertj" % "assertj-core" % "3.14.0" % Test,
-      "org.mockito" % "mockito-core" % "4.5.1",
+      "org.mockito" % "mockito-inline" % "4.5.1",
```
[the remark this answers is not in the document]
bion: yes. mockito-inline allows mocking statics

server/conf/application.conf
```diff
@@ -517,3 +517,9 @@ api_secret_salt = ${?CIVIFORM_API_SECRET_SALT}
 
 api_keys_ban_global_subnet = true
 api_keys_ban_global_subnet = ${?CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET}
+
+# Max and default page size for ProgramApplicationsApiController.list
+# This should be kept low enough so that the request completes before hitting the
+# load balancer timeout.
+api_applications_list_max_page_size = 1000
```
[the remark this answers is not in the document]
bion: > do we think different deployments will really want different system wide defaults, much more so than individuals wanting customization as they interact with a specific program?

I don't know. I made this value configurable so that if a deployment is having a hard time with an API consumer requesting pages that are too big, they can reduce the max size without code changes (and doing this much was easy). If more granular control is needed I think we should let feedback drive the requirements for it.

server/test/app/ApiAuthenticationTest.java
bion: done

server/test/support/ResourceCreator.java
bion: done

server/test/support/ResourceCreator.java
bion: i actually really like this, thanks for showing me!

server/test/support/CfTestHelpers.java
bion: I can go either way with a very mild preference for `Cf` for brevity. Thoughts?

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

jeanniefu (approved):

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: Added a docstring for `createFakeProgramWithEnumerator`

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: Added javadoc for `createFakeProgramWithEnumerator`

server/conf/application.conf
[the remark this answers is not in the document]
shanemc-goog: I'm ok leaving it, but it does feel like the wrong place for the configuration.  The folks using the Admin interface are unlikely to be aware of it.

server/app/services/IdentifierBasedPaginationSpec.java
```diff
@@ -32,7 +25,13 @@ public int getPageSize() {
     return this.pageSize;
   }
 
-  public Optional<T> getCurrentPageOffsetIdentifier() {
+  /**
+   * Get the offset identifier for this page of results. The identifier is a value that can be
```
shanemc-goog, line 29:
```suggestion
   * Get the offset identifier for this page of results.
   *
   * <p>The identifier is a value that can be
```

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
shanemc-goog: The test setup is still quite unreadable.

After looking at the createFake.. method I see that the magic is that each application has a timestamp of 2022-01-01, 2022-02-01, 2022-03-01.

That is hidden by the parent setup and not communicated by the object names, such that searching for [2022-01-15, 2022-02-15) has no readable result from this test.

A few ways to shore this up.
1. Comment on the relative order, "the three applicationX are a month apart successively", and use the actual timestamp of applicationTwo in specifying the fromDate/toDate. EG applicationTwo.submittion_time - 1 ---> toString format

2. Create alias/reference variables in this class from the parents variables, with clearer names; eg
applicationOldest, applicationMiddle, applicationYoungest Those then will convey the relevant distinguishing characteristic.
Having the search time directly associated with the resultant application would still be good here, instead of the magic values.

3. Just document the hidden timestamp setup as it's directly pertinent here. If we didn't have that setup, and we were creating the programs through utility method, the salient thing that would be included in this test are those dates, so we should communicate that information one way or another.

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
shanemc-goog: For here, comment at the start that there are three applications in reverse creation order to account for relevant test "setup" information.

shanemc-goog (changes requested):

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: done

server/test/controllers/api/ProgramApplicationsApiControllerTest.java
bion: Tried to do a combination of better variable names with an explanatory comment in the setup section. I balked at doing date math because it gets weird doing math on instants and then converting them to dates IMO.

shanemc-goog (approved):

PR #2518, opened 2022-05-19
Set authority_id for fake admins

bion:
Some features, such as creating API keys, require user profiles to have an authority_id, even in demo mode. This PR sets a randomized authority_id with the prefix "fake-admin-" so that the value is present in demo mode. The randomness is to prevent collisions when logging in repeatedly as a fake admin.

### DIFF
--- a/server/app/auth/FakeAdminClient.java
+++ b/server/app/auth/FakeAdminClient.java
@@ -13,7 +13,7 @@ import org.pac4j.core.util.HttpActionHelper;
 
 /**
  * This class implements a special client that allows logging in without logging in to a real AD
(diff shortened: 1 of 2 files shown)


gwendolyngoetz (approved):
This looks like it'll solve the problem for staging. Thanks!

Do you think we still need the [environment check in the ApiKeyService](https://github.com/seattle-uat/civiform/blob/1856236535ddc8316ec4ac1944cfc321ea6d0ca5/server/app/services/apikey/ApiKeyService.java#L320)?

kmcnellis (approved):

Issue #2010, opened 2022-03-02
TI's can search client list


elisekalstad:
TIs would like to be able to search for clients by name and date of birth within the TI dashboard.

Mock-ups needed.

swatkat1:
Hi @rahaghas I have a question for you- 
When you recommended the Search by DOB in a separate search box, did the text and the date search have a OR functionality or an AND functionality? 
Between the two searches should I add -AND it or 
OR it?

![Image](https://user-images.githubusercontent.com/101214724/176748517-2fc3f859-e0e2-420a-9042-be340fbfcd40.png)

rahaghas:
Would prefer an OR functionality, meaning the Search button would only apply to the Search by name field and date of birth would need it's own action

swatkat1:
Awesome Thank you @rahaghas

PR #2522, opened 2022-05-20
TI Dashboard Improvements

### LINKED ISSUE #2009: TIs can view additional client information within TI dashboard.
Trusted Intermediaries would like to be able to view additional client information within the TI dashboard. They would like to be able to view **client's name,** **date of birth,** and **address** (WHOOPS! I MEANT EMAIL ADDRESS!) within the TI dashboard. They would like to be able to **view the application status** for each client's application within the client card.

Will need mock-ups for TI view.

swatkat1:
### Description
*Adding DOB to Client TI Dashboard
*Added the ability to update the DOB for any clients whose DOB is missing
*search by DOB added to TI Dashboard

### Checklist
- [ -] Created tests which fail without the change (if possible)

## Release notes:
*Made changes to the TI Dashboard by adding Applicant DateOfBirth and Also the possibility to update DOB in the cases where its missing
*Added functionality to search by DOB on TI Dashboards.

### Issue(s)
Fixes #2009 #2010 

image


server/app/services/applicant/ApplicantData.java
```diff
@@ -149,4 +149,15 @@ public ImmutableMap<Path, String> getFailedUpdates() {
...
+  public Optional<String> getDOB() {
+    String dob = readString(WellKnownPaths.APPLICANT_DOB).get();
+    return Optional.of(dob);
+  }
+
+  public void setDOB(@Nullable String dob) {
+    if (!hasPath(WellKnownPaths.APPLICANT_DOB)) {
+      putString(WellKnownPaths.APPLICANT_DOB, dob);
+    }
+  }
```
bion, line 162: suggestion (blocking): add unit tests for these methods

bion: suggestion (blocking): use CameCase for these method names https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

server/app/services/applicant/ApplicantData.java
```diff
@@ -149,4 +149,15 @@ public ImmutableMap<Path, String> getFailedUpdates() {
   public boolean updateDidFailAt(Path path) {
     return getFailedUpdates().containsKey(path);
   }
+
+  public Optional<String> getDOB() {
+    String dob = readString(WellKnownPaths.APPLICANT_DOB).get();
+    return Optional.of(dob);
+  }
+
+  public void setDOB(@Nullable String dob) {
```
bion, line 158: suggestion (blocking): avoid `@Nullable` and prefer `Optional.ofNullable` instead

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -207,7 +206,23 @@ private Tag renderApplicantRow(Account applicant) {
             StyleUtils.even(Styles.BG_GRAY_100))
         .with(renderInfoCell(applicant))
         .with(renderApplicantInfoCell(applicant))
-        .with(renderActionsCell(applicant));
+        .with(renderActionsCell(applicant))
+        .with(renderDOBCell(applicant));
+  }
+
+  private Tag renderDOBCell(Account applicant) {
```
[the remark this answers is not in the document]
bion: suggestion (blocking): `Applicant` and `Account` are separate resources in the code base, it's confusing to name an `Account` variable `applicant`
```suggestion
  private Tag renderDOBCell(Account account) {
```

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -207,7 +206,23 @@ private Tag renderApplicantRow(Account applicant) {
...
+
+    Optional<Applicant> newestApplicant = applicant.newestApplicant();
+    if (newestApplicant.isEmpty()) {
+      return td().withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.PR_12);
+    }
+    String applicantDOB = "";
+    Optional<String> dob = newestApplicant.get().getApplicantData().getDOB();
+    if (dob.isPresent()) {
+      applicantDOB = dob.get();
+    }
```
bion, line 223: suggestion (blocking): use `orElse` for accessing an `Optional` with a default value:
```suggestion
    String applicantDob = newestApplicant.get().getApplicantData().getDob().orElse("");
```

server/app/repository/UserRepository.java
```diff
@@ -279,9 +280,9 @@ public void createNewApplicantForTrustedIntermediaryGroup(
     newAccount.save();
     Applicant applicant = new Applicant();
     applicant.setAccount(newAccount);
-    applicant
-        .getApplicantData()
-        .setUserName(form.getFirstName(), form.getMiddleName(), form.getLastName());
+    ApplicantData applicantData = applicant.getApplicantData();
+    applicantData.setUserName(form.getFirstName(), form.getMiddleName(), form.getLastName());
+    applicantData.setDOB(form.getDob());
```
bion, line 285: suggestion (blocking): validate the user-provided `dob` value in the form before writing it to the database, add tests to exercise the validation

server/app/services/WellKnownPaths.java
```diff
@@ -9,4 +9,5 @@ public class WellKnownPaths {
   public static Path APPLICANT_MIDDLE_NAME = Path.create("applicant.name.middle_name");
   public static Path APPLICANT_LAST_NAME = Path.create("applicant.name.last_name");
   public static Path APPLICANT_NAME = Path.create("applicant.name");
+  public static Path APPLICANT_DOB = Path.create("applicant.date_of_birth");
```
bion, line 12: question (blocking): confirm that in Seattle's system the DOB question has this path

bion (changes requested):

server/app/services/applicant/ApplicantData.java
```diff
@@ -149,4 +149,15 @@ public ImmutableMap<Path, String> getFailedUpdates() {
...
+    if (!hasPath(WellKnownPaths.APPLICANT_DOB)) {
```
clouser2, line 159: Not blocking: The current behavior silently doesn't update if the DOB has already been set. That behavior seems surprising to me. Would it be acceptable to instead throw some sort of error indicating that the DOB can't be set multiple times?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -262,7 +277,8 @@ private Tag renderApplicantTableHeader() {
         tr().withClasses(Styles.BORDER_B, Styles.BG_GRAY_200, Styles.TEXT_LEFT)
             .with(th("Info").withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.W_1_3))
             .with(th("Applications").withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.W_1_3))
-            .with(th("Actions").withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.W_1_4)));
+            .with(th("Actions").withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.W_1_3))
+            .with(th("DOB").withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.W_1_3)));
```
clouser2, line 281: Not blocking: Can this be de-dented?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -207,7 +206,23 @@ private Tag renderApplicantRow(Account applicant) {
...
-        .with(renderActionsCell(applicant));
+        .with(renderActionsCell(applicant))
+        .with(renderDOBCell(applicant));
+  }
+
+  private Tag renderDOBCell(Account applicant) {
+
+    Optional<Applicant> newestApplicant = applicant.newestApplicant();
+    if (newestApplicant.isEmpty()) {
+      return td().withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.PR_12);
```
clouser2, line 217: Not blocking: To avoid repeating the `td()` call and associated styles, could this instead be written as:

String applicantDOB = account.newestApplicant().map(applicant -> applicant.getApplicantData().getDob()).orElse("");

clouser2 (changes requested):

server/app/services/applicant/ApplicantData.java
```diff
@@ -149,4 +149,15 @@ public ImmutableMap<Path, String> getFailedUpdates() {
   public boolean updateDidFailAt(Path path) {
     return getFailedUpdates().containsKey(path);
   }
+
+  public Optional<String> getDOB() {
```
bion: suggestion (blocking): follow the style guide for method naming
```suggestion
  public Optional<String> getDob() {
```

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -85,10 +93,55 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+    if (form.hasErrors()) {
+      return redirectToDashboardWithUpdateDateOfBirthError(form.errors().get(0).message(), form);
+    }
+    if (Strings.isNullOrEmpty(form.get().getDob())) {
+      return redirectToDashboardWithUpdateDateOfBirthError("Date Of Birth is required.", form);
+    }
+
+    try {
+      Applicant applicant =
+          userRepository.lookupApplicant(applicantId).toCompletableFuture().join().get();
```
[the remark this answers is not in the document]
clouser2: Additionally, should we have explicit error handling if the applicant couldn't be found (`Optional.empty()` is returned)?

clouser2 (changes requested):
I like the new UI!!!

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -85,10 +93,55 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+
+    if (civiformProfile.isEmpty()) {
+      return unauthorized();
+    }
+    Optional<TrustedIntermediaryGroup> trustedIntermediaryGroup =
+        userRepository.getTrustedIntermediaryGroup(civiformProfile.get());
+    if (trustedIntermediaryGroup.isEmpty()) {
+      return notFound();
+    }
+    if (!trustedIntermediaryGroup.get().id.equals(tiId)) {
```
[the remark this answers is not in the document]
bion: +1 blocking

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -13,9 +13,11 @@
 import com.google.common.collect.ImmutableList;
 import com.typesafe.config.Config;
 import forms.AddApplicantToTrustedIntermediaryGroupForm;
```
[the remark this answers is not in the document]
swatkat1: Done!

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -85,10 +93,55 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+    }
+    if (!trustedIntermediaryGroup.get().id.equals(tiId)) {
+      return unauthorized();
+    }
+    Form<UpdateApplicantDOB> form =
+        formFactory.form(UpdateApplicantDOB.class).bindFromRequest(request);
+    if (form.hasErrors()) {
+      return redirectToDashboardWithUpdateDateOfBirthError(form.errors().get(0).message(), form);
+    }
+    if (Strings.isNullOrEmpty(form.get().getDob())) {
```
bion, line 121: issue (blocking): please validate the DOB for:
- being in the correct format
- that the date is in the past

server/app/controllers/ti/TrustedIntermediaryController.java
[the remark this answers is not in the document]
bion: +1 blocking

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -85,10 +93,55 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+      return redirectToDashboardWithUpdateDateOfBirthError("Date Of Birth is required.", form);
+    }
+
+    try {
+      Applicant applicant =
+          userRepository.lookupApplicant(applicantId).toCompletableFuture().join().get();
+      applicant.getApplicantData().setDateOfBirth(form.get().getDob());
+      userRepository.updateApplicant(applicant);
```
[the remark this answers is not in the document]
bion: +1 blocking

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -85,10 +93,55 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+      return redirectToDashboardWithUpdateDateOfBirthError(
+          "Date Of birth for this applicant exists.  Cannot Update DOB. "
+              + " Direct applicant to sign in and go to"
+              + " "
+              + trustedIntermediaryUrl,
+          form);
+    }
+    return redirect(
+        routes.TrustedIntermediaryController.dashboard(
+            Optional.empty(), Optional.empty(), Optional.empty()));
```
bion, line 142: suggestion (blocking): add param name comments for each of these empty optionals `/* paramName= */ Optional.empty()`

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -114,11 +167,15 @@ public Result addApplicant(Long id, Http.Request request) {
     if (Strings.isNullOrEmpty(form.get().getLastName())) {
       return redirectToDashboardWithError("Last name required.", form);
     }
+    if (Strings.isNullOrEmpty(form.get().getDob())) {
```
bion, line 170: issue (blocking): please validate the DOB for:

being in the correct format
that the date is in the past

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -114,11 +167,15 @@ public Result addApplicant(Long id, Http.Request request) {
...
+      return redirectToDashboardWithError("Date Of Birth required.", form);
+    }
     try {
       userRepository.createNewApplicantForTrustedIntermediaryGroup(
           form.get(), trustedIntermediaryGroup.get());
       return redirect(
-          routes.TrustedIntermediaryController.dashboard(Optional.empty(), Optional.empty()));
+          routes.TrustedIntermediaryController.dashboard(
+              Optional.empty(), Optional.empty(), Optional.empty()));
```
bion, line 178: suggestion (blocking): param name comments

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -134,12 +191,22 @@ public Result addApplicant(Long id, Http.Request request) {
   private Result redirectToDashboardWithError(
       String errorMessage, Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
     return redirect(
-            routes.TrustedIntermediaryController.dashboard(Optional.empty(), Optional.empty()))
+            routes.TrustedIntermediaryController.dashboard(
+                Optional.empty(), Optional.empty(), Optional.empty()))
```
bion, line 195: suggestion (blocking): param name comments

server/app/forms/SearchWithDateForm.java
```diff
@@ -0,0 +1,22 @@
+package forms;
+
+public class SearchWithDateForm {
```
bion, line 3: suggestion (blocking): please document this class and make it `final`

server/app/forms/UpdateApplicantDOB.java
```diff
@@ -0,0 +1,13 @@
+package forms;
+
+public class UpdateApplicantDOB {
```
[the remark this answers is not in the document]
bion: suggestion (blocking): please document and make `final`

server/app/models/Account.java
```diff
@@ -160,4 +161,11 @@ public String getApplicantName() {
         .map(u -> u.getApplicantData().getApplicantName().orElse("<Unnamed User>"))
         .orElse("<Unnamed User>");
   }
+
+  public Optional<LocalDate> getApplicantDateOfBirth() {
```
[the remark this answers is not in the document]
bion: +1 blocking

server/app/models/Account.java
```diff
@@ -160,4 +161,11 @@ public String getApplicantName() {
...
+    return this.getApplicants().stream()
+        .max(Comparator.comparing(Applicant::getWhenCreated))
+        .map(u -> u.getApplicantData().getDateOfBirth())
```
bion, line 168: nit: use method reference

```suggestion
        .map(Applicant::getApplicantData)
        .map(ApplicantData::getDateOfBirth)
```

server/app/models/Account.java
```diff
@@ -160,4 +161,11 @@ public String getApplicantName() {
         .map(u -> u.getApplicantData().getApplicantName().orElse("<Unnamed User>"))
         .orElse("<Unnamed User>");
   }
+
+  public Optional<LocalDate> getApplicantDateOfBirth() {
+    return this.getApplicants().stream()
```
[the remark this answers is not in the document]
bion: suggestion (blocking): factor out the common code into a separate method for retrieving the most recent applicant, document the method

server/app/models/TrustedIntermediaryGroup.java
```diff
@@ -53,7 +55,8 @@ public String getDescription() {
     return this.description;
   }
 
-  public ImmutableList<Account> getManagedAccounts(Optional<String> search) {
+  public ImmutableList<Account> getManagedAccounts(
+      Optional<String> search, Optional<String> searchDate) {
```
bion, line 59: suggestion (blocking): document this method and state what happens if search and searchDate are present vs not

bion, line 59: suggestion (blocking): rename `search` to `nameQuery`

server/app/models/TrustedIntermediaryGroup.java
```diff
@@ -66,6 +69,17 @@ public ImmutableList<Account> getManagedAccounts(Optional<String> search) {
                           .contains(search.get().toLowerCase(Locale.ROOT)))
               .collect(ImmutableList.toImmutableList());
     }
+    if (searchDate.isPresent()) {
+      LocalDate localDate =
```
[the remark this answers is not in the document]
bion: suggestion (blocking): parsing and validating it is of the correct format is not a responsibility of the model, please move that logic into a service or validator class

server/app/models/TrustedIntermediaryGroup.java
bion, line 59: suggestion (blocking): change `Optional<String> searchDate` to `Optional<LocalDate> searchDate`. Do not do parsing and validation in a model class

server/app/services/applicant/ApplicantData.java
```diff
@@ -150,4 +152,15 @@ public ImmutableMap<Path, String> getFailedUpdates() {
...
   }
+
+  public Optional<LocalDate> getDateOfBirth() {
+    return readDate(WellKnownPaths.APPLICANT_DOB);
+  }
+
+  public void setDateOfBirth(String dateString) throws DateOfBirthExistsException {
+    if (hasPath(WellKnownPaths.APPLICANT_DOB)) {
+      throw new DateOfBirthExistsException();
+    }
```
bion, line 163: question (blocking): why can the date of birth not change after initially set? what if a user needs to correct a mistake?

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -328,7 +328,7 @@ private void notifySubmitter(
     String tiDashLink =
         baseUrl
             + controllers.ti.routes.TrustedIntermediaryController.dashboard(
-                    Optional.empty(), Optional.empty())
+                    Optional.empty(), Optional.empty(), Optional.empty())
```
bion, line 331: suggestion (blocking): param comments for empty optionals

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -183,7 +183,7 @@ private DivTag maybeRenderTiButton(Optional<CiviFormProfile> profile, String use
       String tiDashboardText = "Trusted intermediary dashboard";
       String tiDashboardLink =
           controllers.ti.routes.TrustedIntermediaryController.dashboard(
-                  Optional.empty(), Optional.empty())
+                  Optional.empty(), Optional.empty(), Optional.empty())
```
bion, line 186: suggestion (blocking): param comments

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -38,10 +30,7 @@
 import views.components.FieldWithLabel;
 import views.components.LinkElement;
 import views.components.ToastMessage;
-import views.style.BaseStyles;
-import views.style.ReferenceClasses;
-import views.style.StyleUtils;
-import views.style.Styles;
+import views.style.*;
```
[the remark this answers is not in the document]
bion: +1 blocking

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -218,15 +221,50 @@ private TrTag renderTIRow(Account ti) {
...
+  }
+
+  private TdTag renderDateOfBirthCell(
+      Account account, TrustedIntermediaryGroup tiGroup, Http.Request request) {
+
+    Optional<Applicant> newestApplicant = account.newestApplicant();
+    if (newestApplicant.isEmpty()) {
+      return td().withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.PR_12);
+    }
+    Optional<LocalDate> current_DOB = newestApplicant.get().getApplicantData().getDateOfBirth();
```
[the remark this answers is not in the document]
bion: +1 blocking

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -218,15 +221,50 @@ private TrTag renderTIRow(Account ti) {
...
+    // for cases where DOB is already present, display the Date of Birth
+    // for other cases, show the form to update the DOB.
+    // Note: the update can be done only once.
```
bion, line 247: question (blocking): why?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -218,15 +221,50 @@ private TrTag renderTIRow(Account ti) {
...
+    if (current_DOB.isPresent()) {
+      LocalDate unformatedDate = current_DOB.get();
+      return td().with(
+              div(unformatedDate.format(DateTimeFormatter.ofPattern("MM-dd-yyyy")))
```
bion, line 251: suggestion (blocking): use [DateConverter](https://github.com/seattle-uat/civiform/blob/main/server/app/services/DateConverter.java#L14)

server/app/models/TrustedIntermediaryGroup.java
```diff
@@ -66,6 +69,17 @@ public ImmutableList<Account> getManagedAccounts(Optional<String> search) {
...
+          LocalDate.parse(searchDate.get(), DateTimeFormatter.ofPattern("yyyy-MM-dd"));
```
bion, line 74: suggestion (blocking): use [DateConverter](https://github.com/seattle-uat/civiform/blob/main/server/app/services/DateConverter.java#L14)

server/conf/routes
```diff
@@ -76,9 +76,10 @@ GET     /admin/tiGroups/:id/edit           controllers.admin.TrustedIntermediary
 POST    /admin/tiGroups/:id/addTi          controllers.admin.TrustedIntermediaryManagementController.addIntermediary(id: Long, request: Request)
 POST    /admin/tiGroups/:id/removeTi       controllers.admin.TrustedIntermediaryManagementController.removeIntermediary(id: Long, request: Request)
 
-GET     /admin/tiDash                      controllers.ti.TrustedIntermediaryController.dashboard(request: Request,  search: java.util.Optional[String], page: java.util.Optional[Integer])
-POST    /admin/tiGroups/:id/addApplicant   controllers.ti.TrustedIntermediaryController.addApplicant(id: Long, request: Request)
 
+GET     /admin/tiDash                      controllers.ti.TrustedIntermediaryController.dashboard(request: Request,  search: java.util.Optional[String], searchDate: java.util.Optional[String],  page: java.util.Optional[Integer])
+POST    /admin/tiGroups/:id/addApplicant   controllers.ti.TrustedIntermediaryController.addApplicant(id: Long, request: Request)
+POST     /admin/tiGroups/:id/updateDateOfBirth/:applicantId controllers.ti.TrustedIntermediaryController.updateDateOfBirth(id: Long, applicantId: Long, request: Request)
```
bion, line 82: suggestion (blocking): fix spacing here to be consistent with the rest of this file

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -63,4 +63,13 @@ public void withFailedUpdates() {
     assertThat(data.getFailedUpdates()).isEqualTo(ImmutableMap.of(samplePath, "invalid_value"));
     assertThrows(IllegalStateException.class, () -> data.asJsonString());
   }
+
+  @Test
```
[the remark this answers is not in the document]
bion: +1 blocking please add a browser test

bion (changes requested):
An overall need for this PR is to move toward better separation of concerns. Controllers should avoid directly implementing validation and data querying/collating logic, and should instead delegate to service classes. We don't have a `TrustedIntermediaryService` yet, so this may be a good opportunity to introduce one.

Similarly, model classes should avoid implementing complex query logic. If you need to write more complex queries, a repository class is the correct place for that.

server/test/services/applicant/ApplicantDataTest.java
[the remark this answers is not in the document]
swatkat1: I will be honest with you- I am very poor in ts and last time it took me a long time to get the broken browser tests working. Unless I can pair program, it will be hard for me do it alone. I have done my best in adding Unit testing and also testing my branch (switch between main and my current branch to make sure the experience is seamless)

server/app/views/applicant/TrustedIntermediaryDashboardView.java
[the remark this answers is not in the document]
nbeloglazov: We should enable that check in error prone: https://errorprone.info/bugpattern/WildcardImport MIght go through all "suggestion" checks and see if we want to make some of them errors.

server/app/services/ti/DateOfBirthExistsException.java
```diff
@@ -0,0 +1,7 @@
+package services.ti;
+
+public class DateOfBirthExistsException extends java.util.NoSuchElementException {
+  public DateOfBirthExistsException() {
```
nbeloglazov, line 4: Does this constructor brings any value?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: for a given applicant, the DOB is unique and can be updated when it isn't present. If it were so, searching by DOB will a difficult--> If T1 edits A1's DOB, and T2 edit A1's DOB again, T1 and T2 will have difficulty searching for A1's DOB.. DOB is should be updated once. I can check if Elise on this.

server/test/services/applicant/ApplicantDataTest.java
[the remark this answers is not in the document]
nbeloglazov: I'm happy to pair program. I can do tomorrow 10:30-12pm. I'm experienced with TS and somewhat with civiform browsertests. So we should be able to figure out with two of us.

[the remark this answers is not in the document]
swatkat1: Thanks @nbeloglazov :) I will send you an invite!

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,54 @@
+import {Page} from 'playwright'
+
+export class TiDashboard {
```
nbeloglazov, line 3: Add comment explaining that this class helps to interfact with Trusted intermediatery dashboard page. And requires that user logs in as TI first.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,54 @@
...
+  async fillFormForNewClients(client: ClientInformation) {
+    await this.page.fill('label:has-text("Email Address")', client.emailAddress)
+    await this.page.fill('label:has-text("First Name")', client.firstName)
+    await this.page.fill('label:has-text("Middle Name")', client.middleName)
+    await this.page.fill('label:has-text("Last Name")', client.lastName)
+    await this.page.fill('label:has-text("Date Of Birth")', client.dobDate)
+    await this.page.click('text ="Add"')
+  }
+
+  async checkInnerTableForClientInformation(client: ClientInformation) {
```
nbeloglazov, line 23: How about renaming it to `expectDashboardContainClient`? So that "expect" word matches `expect` functions used in jest for assertions.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,54 @@
...
+    const tableInnerText = await this.page.innerText('table')
+    expect(tableInnerText).toContain(this.convertToMMDDYYYY(client.dobDate))
+    expect(tableInnerText).toContain(client.emailAddress)
+    expect(tableInnerText).toContain(client.firstName)
+    expect(tableInnerText).toContain(client.lastName)
+  }
+  async checkInnterTableNotToConatain(client: ClientInformation) {
```
nbeloglazov, line 30: Typo: `Innter` => `Inner`
Also name of the function should match the previous function. They both should have similar names with the only difference being "No" or "Not" or something like that in this function.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,54 @@
...
+    const tableInnerText = await this.page.innerText('table')
+    expect(tableInnerText).not.toContain(this.convertToMMDDYYYY(client.dobDate))
+    expect(tableInnerText).not.toContain(client.emailAddress)
+    expect(tableInnerText).not.toContain(client.firstName)
+    expect(tableInnerText).not.toContain(client.lastName)
+  }
+  async searchByDob(dobDate: string) {
```
nbeloglazov, line 37: Let's spell out searchByDob to searchByDateOfBirth so that it's easier to read tests. "Dob" might not me immediately clear to whoever reads the test.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor(page: Page) {
+    this.page = page
+  }
+
+  async gotoTIDashboardPage(page: Page) {
+    await page.click('text="Trusted intermediary dashboard"')
+  }
+
+  async fillFormForNewClients(client: ClientInformation) {
```
nbeloglazov, line 14: `fillForm` sounds like it only fills while your function also submits the form creating the client. So maybe rename it to `fillAndSubmitFormForNewClient` or `createClient`?

nbeloglazov (changes requested):

server/app/services/ti/DateOfBirthExistsException.java
swatkat1: Removed this Exception as the PM wants this to be an editable field..

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: Checked with Elise - we want this to be editable always (whether DOB is present or not)

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: Not needed at it is binding with the input() tag

server/app/controllers/ti/TrustedIntermediaryController.java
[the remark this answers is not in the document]
swatkat1: I will add it in the next iteration

server/app/repository/SearchParameters.java
```diff
@@ -0,0 +1,25 @@
+package repository;
+
+import com.google.auto.value.AutoValue;
+import java.util.Optional;
+
+@AutoValue
+public abstract class SearchParameters {
```
nbeloglazov, line 7: SearchParameters is pretty generic name. Given that this class works specifically with search parameters for clients in TI. I can think of few options:
1. Rename it to TrustedIntermediaryClientSearchParams (long :/)
2. Introduce `ti` package and move it there as ClientSearchParams.

@clouser2 @swatkat1 are there precedents for passing search data around for other usecases?

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Optional<Applicant> testApplicant =
+        repo.lookupApplicantByEmail("sample2@fake.com").toCompletableFuture().join();
+    assertThat(testApplicant.get().getApplicantData().getDateOfBirth().get().toString())
+        .isEqualTo("2022-07-18");
+  }
+
+  @Test
+  public void testUpdateDOBFunctionWithExistingDob() {
+    Http.RequestBuilder requestBuilder =
+        addCSRFToken(Helpers.fakeRequest().bodyForm(ImmutableMap.of("dob", "2022-10-05")));
```
nbeloglazov, line 87: Can you move this experssion to line 98 so that when the test reads is 

1. Initialize form and create new applicant.
2. Initialize change-dob request and update DOB.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -0,0 +1,106 @@
...
+    repo.createNewApplicantForTrustedIntermediaryGroup(form, group);
+    Optional<Applicant> applicant =
+        repo.lookupApplicantByEmail("sample3@example.com").toCompletableFuture().join();
+    Result result =
+        tiController.updateDateOfBirth(group.id, applicant.get().id, requestBuilder.build());
+    assertThat(result.status()).isEqualTo(SEE_OTHER);
+    Optional<Applicant> finalApplicant =
+        repo.lookupApplicant(applicant.get().id).toCompletableFuture().join();
+    assertThat(finalApplicant.get().getApplicantData().getDateOfBirth().get().toString())
+        .isEqualTo("2022-07-10");
```
nbeloglazov, line 104: Shouldn't expected DOB be `2022-10-05` as that's what you passed to `updateDateOfBirth` method.

nbeloglazov (changes requested):

server/app/models/Account.java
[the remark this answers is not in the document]
swatkat1: I Cannot understand this comment. Please explain more

server/app/repository/SearchParameters.java
swatkat1: Thank you for you feedback. But I am taking this in the second iteration.

server/app/models/Account.java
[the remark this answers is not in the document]
clouser2: The javadoc string for Account.getApplicantName explicitly calls out why it’s using the “max” operator. I'd replicate that here as well.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
swatkat1: No the DB always stores DOB in yyyy-mm-dd format. This is followed everywhere.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -134,12 +191,22 @@ public Result addApplicant(Long id, Http.Request request) {
...
+        .flashing("providedDateOfBirth", form.get().getDob());
+  }
+
+  private Result redirectToDashboardWithUpdateDateOfBirthError(
+      String errorMessage, Form<UpdateApplicantDOB> form) {
+    return redirect(
+            routes.TrustedIntermediaryController.dashboard(
+                Optional.empty(), Optional.empty(), Optional.empty()))
+        .flashing("error", errorMessage)
+        .flashing("providedDateOfBirth", form.get().getDob());
```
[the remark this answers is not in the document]
swatkat1: [error] /usr/src/server/app/controllers/ti/TrustedIntermediaryController.java:222:1: incompatible types: java.util.Optional<forms.UpdateApplicantDob> cannot be converted to java.lang.String
[error] form.value()

Cannot fix in this iteration.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -218,15 +233,44 @@ private TrTag renderTIRow(Account ti) {
...
+                .withMethod("POST")
+                .withAction(
+                    routes.TrustedIntermediaryController.updateDateOfBirth(tiGroup.id, account.id)
+                        .url())
+                .with(
+                    input()
+                        .withId("date-of-birth-update")
+                        .withName("dob")
+                        .withType("date")
+                        .withValue(current_Dob.isPresent() ? current_Dob.get().toString() : ""),
```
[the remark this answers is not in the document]
swatkat1: I can't as I have to convert the currentDob to string.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -0,0 +1,106 @@
...
+  public void setup() {
+    profileFactory = instanceOf(ProfileFactory.class);
+    tiController = instanceOf(TrustedIntermediaryController.class);
+    Applicant managedApplicant = createApplicant();
+    createTIWithMockedProfile(managedApplicant);
+    repo = instanceOf(UserRepository.class);
+    data = profileFactory.createFakeTrustedIntermediary();
+  }
+
+  @Test(expected = NullPointerException.class)
```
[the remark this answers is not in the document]
swatkat1: Yes - 

[error] Test controllers.ti.TrustedIntermediaryControllerTest.addApplicantTestWithMissingDob failed: java.lang.NullPointerException: null, took 0.035 sec
[error]     at controllers.ti.TrustedIntermediaryControllerTest.addApplicantTestWithMissingDob(TrustedIntermediaryControllerTest.java:55)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]     at java.lang.reflect.Method.invoke(Method.java:566)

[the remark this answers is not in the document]
nbeloglazov: Sorry, I looked at the code and don't understand why we expected a NPE here. According to the code we should return redirection to dashboard and no exception should be thrown.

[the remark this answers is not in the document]
swatkat1: I completely Agree but think it is failing in this Step:    Form<AddApplicantToTrustedIntermediaryGroupForm> form =
        formFactory.form(AddApplicantToTrustedIntermediaryGroupForm.class).bindFromRequest(request); --> the form fields aren't all there.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -80,26 +83,26 @@ public Result dashboard(
     if (trustedIntermediaryGroup.isEmpty()) {
       return notFound();
     }
+    SearchParameters searchParameters =
+        SearchParameters.builder().setSearch(search).setSearchDate(searchDate).build();
     ImmutableList<Account> managedAccounts =
-        trustedIntermediaryGroup.get().getManagedAccounts(search, searchDate);
+        tiService.retriveSearchResult(searchParameters, trustedIntermediaryGroup);
```
clouser2, line 89: Consider naming this method `getManagedAccounts` or `filterManagedAccounts`. `retrieveSearchResult` seems a bit general to know what's being returned.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -110,36 +113,30 @@ public Result updateDateOfBirth(Long tiId, Long applicantId, Http.Request reques
...
-    } catch (DateOfBirthExistsException e) {
-      String trustedIntermediaryUrl = baseUrl + "/trustedIntermediaries";
-
-      return redirectToDashboardWithUpdateDateOfBirthError(
-          "Date Of birth for this applicant exists.  Cannot Update DOB. "
-              + " Direct applicant to sign in and go to"
-              + " "
-              + trustedIntermediaryUrl,
-          form);
+      tiService.checkFormForDobUpdate(form, applicantId);
```
clouser2, line 119: It appears that this method also updates the applicant data rather than just "checking it". Perhaps this could be renamed to `updateApplicantDateOfBirth`?

Also, minor nit on service argument order. Typically the database ID for the entity to be updated is the first argument.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -110,36 +113,30 @@ public Result updateDateOfBirth(Long tiId, Long applicantId, Http.Request reques
...
+    } catch (IncorrectDateFormatException e) {
+      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
```
clouser2, line 121: Generally I've found the codebase to be a bit inconsistent here, but one pattern we use rather than throwing exceptions for validation errors is to return an `ErrorAnd<THERETURNTYPE, CiviFormError>`. Then the caller can check if there are errors and perform similar logic to what you have here.

Probably for another time, but some of this validation logic could likely be embedded into the form itself.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -110,36 +113,30 @@ public Result updateDateOfBirth(Long tiId, Long applicantId, Http.Request reques
...
+    } catch (MissingDateOfBirthException e) {
```
clouser2, line 122: If you stick with the exception pattern, consider consolidating them into a single catch since the error handling is the same for each.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -110,36 +113,30 @@ public Result updateDateOfBirth(Long tiId, Long applicantId, Http.Request reques
...
+    } catch (FormHasErrorException e) {
+      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
+    } catch (DateOfBirthNotInPastException e) {
+      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
     }
+
     return redirect(
         routes.TrustedIntermediaryController.dashboard(
-            Optional.empty(), Optional.empty(), Optional.empty()));
+            /* paramName=  search*/
```
clouser2, line 134: Not blocking: Please add a space before the " */" here and elsewhere.

server/app/controllers/ti/TrustedIntermediaryController.java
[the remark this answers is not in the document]
clouser2: I believe this can be fixed by calling `form.value().get()`.

server/app/models/Account.java
```diff
@@ -156,16 +156,26 @@ public void removeAdministeredProgram(ProgramDefinition program) {
...
-        .orElse("<Unnamed User>");
+   Optional<Applicant> newestApplicant = newestApplicant();
+   if(newestApplicant.isPresent() && newestApplicant.get().getApplicantData().getApplicantName().isPresent())
+   {
+     return newestApplicant.get().getApplicantData().getApplicantName().get();
+   }
+    return "<Unnamed User>";
   }
-
+  /**
```
clouser2, line 166: Please add a newline between methods.

server/app/repository/SearchParameters.java
clouser2: +1 that we likely want a more specific name, such as TIAccountFilters or something along those lines.

Mind leaving a TODO to rename this to something different so it can be tracked against the larger issue for cleanups?

server/app/services/DateConverter.java
```diff
@@ -32,6 +32,10 @@ public String formatIso8601Date(Instant instant) {
     return DateTimeFormatter.ISO_DATE.withZone(zoneId).format(instant);
   }
 
+  public LocalDate parseStringtoLocalDate(String dateString) {
```
clouser2, line 35: Blocking: Please add javadocs for both of these methods.

server/app/services/WellKnownPaths.java
```diff
@@ -8,5 +8,5 @@ public class WellKnownPaths {
   public static Path APPLICANT_MIDDLE_NAME = Path.create("applicant.name.middle_name");
   public static Path APPLICANT_LAST_NAME = Path.create("applicant.name.last_name");
   public static Path APPLICANT_NAME = Path.create("applicant.name");
-  public static Path APPLICANT_DOB = Path.create("applicant.date_of_birth");
+  public static Path APPLICANT_DOB = Path.create("applicant.applicant.date_of_birth");
```
clouser2, line 11: Blocking: Why `applicant.applicant.date_of_birth`?

Isn't the path corresponding to the canonical question `applicant.applicant_date_of_birth`?

Canonical question definition: https://github.com/civiform/civiform/blob/16fc26e1bf0722fb29d253ecf137b8667d738528/server/app/tasks/DatabaseSeedTask.java#L77

If this is out of sync with the canonical question, it won't have the desired effect of auto-filling any applications that ask for the date of birth.

server/app/services/ti/DateOfBirthNotInPastException.java
```diff
@@ -0,0 +1,7 @@
+package services.ti;
+
+public class DateOfBirthNotInPastException extends Exception {
```
clouser2, line 3: If we keep the exception classes rather than returning `CiviFormError` from the service method, mark this and the other classes as `final` so that the decision to allow subclassing is explicit rather than implicit.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+import models.Account;
+import models.Applicant;
+import models.TrustedIntermediaryGroup;
+import play.data.Form;
+import repository.SearchParameters;
+import repository.UserRepository;
+import services.DateConverter;
+import services.applicant.exception.ApplicantNotFoundException;
+
+public class TrustedIntermediaryService {
```
clouser2, line 21: Please mark this class as final to make the decision to subclass explicit rather than implicit.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+  private final DateConverter dateConverter;
+  private final UserRepository userRepository;
+
+  @Inject
+  public TrustedIntermediaryService(DateConverter dateConverter, UserRepository userRepository) {
+    this.dateConverter = Preconditions.checkNotNull(dateConverter);
+    this.userRepository = Preconditions.checkNotNull(userRepository);
+  }
+
+  public ImmutableList<Account> retriveSearchResult(
```
clouser2, line 32: Suggestion, remove this method since it essentially has the same method signature as the "filter" method below.

Also note that calling trustedIntermediaryGroup.get() will throw if the optional is empty, so we need to handle that error (likely in the controller).

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+      SearchParameters searchParameters,
+      Optional<TrustedIntermediaryGroup> trustedIntermediaryGroup) {
+    return filterAccountsBySearchParams(searchParameters, trustedIntermediaryGroup.get());
+  }
+
+  public ImmutableList<Account> filterAccountsBySearchParams(
+      SearchParameters searchParameters, TrustedIntermediaryGroup tiGroup) {
+    ImmutableList<Account> allAccounts = tiGroup.getManagedAccounts();
+    allAccounts =
```
clouser2, line 41: Not blocking: Saving this off into a separate variable isn't really saving any lines of code and the majority of the logic is already indented. I'd suggest just returning the result of the stream directly rather than saving it.

Concretely:
```
return tiGroup.getManagedAccounts().stream()
  .filter(...)
  .collect(...);
```

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+        allAccounts.stream()
+            .filter(
+                account -> {
```
clouser2, line 44: Based on the method signatures, I'd expect this to return an account if it matches by name OR by date of birth. It appears that presently it will early out if a name parameter is provided that doesn't match the applicant's name.

I think this can be fixed fairly quickly by only returning true if it's a match.

Additionally, this could benefit from a specific regression test.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+    if (isDobInPast(dobDate.get())) {
+      throw new DateOfBirthNotInPastException("The Date of Birth cannot be in future");
+    }
+    Optional<Applicant> optionalApplicant = userRepository.lookupApplicantSync(applicantId);
+    if (optionalApplicant.isEmpty()) {
+      throw new ApplicantNotFoundException(applicantId);
+    }
+    Applicant applicant = optionalApplicant.get();
+    applicant.getApplicantData().setDateOfBirth(form.get().getDob());
+    userRepository.updateApplicant(applicant);
```
clouser2, line 91: Blocking: We discussed this offline (and it likely hasn't been incorporated yet). This method returns a `CompletableFuture` and we'll need to await the result.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,97 @@
...
+    if (Strings.isNullOrEmpty(form.get().getDob())) {
+      throw new MissingDateOfBirthException("Please Enter a valid Date of Birth");
+    }
+    Optional<LocalDate> dobDate;
+    try {
+      dobDate = Optional.ofNullable(dateConverter.parseStringtoLocalDate(form.get().getDob()));
+    } catch (DateTimeParseException e) {
+      throw e;
+    }
+    if (isDobInPast(dobDate.get())) {
```
clouser2, line 82: Blocking: We discussed this offline (and it likely hasn't been incorporated yet). This needs a "!", otherwise it'll return an error for valid dates.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -95,14 +104,16 @@ private FormTag renderSearchForm(
                 .url())
         .with(
             FieldWithLabel.input()
+                .setId("search")
                 .setFieldName("search")
-                .setValue(search.orElse(""))
-                .setLabelText("Search")
+                .setValue(searchParameters.search().orElse(""))
+                .setLabelText("SearchText")
```
clouser2, line 110: Blocking: I'd add a space here since this is user facing. Also, perhaps it should say something like "Search by Name"

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -231,40 +242,34 @@ private TrTag renderApplicantRow(
...
-                    input().withType("date").withName("dob").withPlaceholder("Click to Enter"),
+                    input()
+                        .withId("date-of-birth-update")
+                        .withName("dob")
+                        .withType("date")
+                        .withValue(currentDob.isPresent() ? currentDob.get().toString() : ""),
                     makeCsrfTokenInputTag(request),
-                    submitButton("Update")
-                        .withClasses(Styles.P_0, ApplicantStyles.BUTTON_PROGRAM_APPLY)));
+                    submitButton("add DOB")
```
clouser2, line 271: Even though we uppercase the text visually, I think we should capitalize the "Add"

server/conf/routes
```diff
@@ -65,21 +65,21 @@ POST    /admin/questions/:id/restore controllers.admin.AdminQuestionController.r
 POST    /admin/questions/:id/archive controllers.admin.AdminQuestionController.archive(request: Request, id: Long)
 
 # Routes for editing question localizations
-GET     /admin/questions/:id/translations/:locale/edit  controllers.admin.AdminQuestionTranslationsController.edit(request: Request, id: Long, locale: String)
-POST    /admin/questions/:id/translations/:locale       controllers.admin.AdminQuestionTranslationsController.update(request: Request, id: Long, locale: String)
+GET     /admin/questions/:id/translations/:locale/edit                    controllers.admin.AdminQuestionTranslationsController.edit(request: Request, id: Long, locale: String)
```
clouser2, line 68: Blocking: Looks like some whitsespace changes were introduced here unnecessarily? Mind reverting?

If I'm reading the diff incorrectly, please ignore.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -36,7 +36,7 @@ public void setup() {
   }
 
   @Test(expected = NullPointerException.class)
```
clouser2, line 38: Please leave a TODO against the cleanup issue so that we can track what needs to be done still.

Even if we keep this as expecting a null-pointer exception, I'd move it closer to where it's thrown so we can assert. As it stands, I'm not sure where it'd be thrown and it's unclear as to whether the assertions below are even triggered or not. You can use `assertThrows`, which we use throughout unit tests.

clouser2 (changes requested):
Took a look and the majority of the feedback is not blocking with a few remaining blocking issues.

server/conf/routes
swatkat1: image
Bion asked me to fix the alignment and I did it.. This is very odd why it is rendering like this.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,66 @@
...
+    await this.page.click('button:text("Search")')
+  }
+  convertToMMDDYYYY(dobDate: string): string {
+    //Input format :  '2021-10-10' YYYY-MM-DD O/p- MM-DD-YYYY
+    const arr = dobDate.split('-')
+    return arr[1] + '-' + arr[2] + '-' + arr[0]
+  }
+  async updateClientDateOfBirth(client: ClientInformation, newDobDate: string) {
+    await this.page.fill('#date-of-birth-update', newDobDate)
+    await this.page.click('text ="Add DOB"')
```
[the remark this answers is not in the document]
clouser2: Unresolving as the space still appears to be there. Is it intentional?

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
clouser2: Unresolving since this doesn't appear to have been addressed. Mind adding a TODO and consider using `assertThrows` instead of catching it at the top-level?

clouser2 (approved):
Approving since there are no more truly blocking issues. I unresolved a few conversations since they don't appear to have been incorporated (not sure if I'm reviewing an older version or not). Would you mind addressing those prior to merge?

Most of it is to add a TODO for the specific issues we intend to address in the cleanup bug. I generally find that without TODOs, it gets really hard to track in the issue itself.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -200,16 +198,11 @@ private Result redirectToDashboardWithError(
                 Optional.empty(),
                 /* paramName=  page */
                 Optional.empty()))
-        .flashing("error", errorMessage)
-        .flashing("providedFirstName", form.get().getFirstName())
```
clouser2, line 139: Blocking (I think): Won't remove these from the flash prevent re-rendering the form with the old values populated? (Used in TrustedIntermediaryDashboardView)

nbeloglazov (approved):
Nice!

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -23,9 +24,11 @@
 import play.i18n.MessagesApi;
 import play.mvc.Http;
 import play.mvc.Result;
+import repository.SearchParameters;
 import repository.UserRepository;
 import services.PaginationInfo;
-import services.ti.EmailAddressExistsException;
+import services.applicant.exception.ApplicantNotFoundException;
+import services.ti.*;
```
bion, line 31: suggestion (blocking): move wildcard import, as per [style guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements)

server/app/repository/SearchParameters.java
bion: suggestion (blocking): please change this name and document the class before merging. It is very generic and likely to cause confusion.

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -102,40 +102,40 @@ public Result dashboard(
...
-    } catch (MissingDateOfBirthException e) {
-      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
-    } catch (ApplicantNotFoundException e) {
-      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
-    } catch (FormHasErrorException e) {
-      redirectToDashboardWithUpdateDateOfBirthError(e.getLocalizedMessage(), form);
-    } catch (DateOfBirthNotInPastException e) {
+      tiService.updateApplicantDateOfBirth(accountId, form);
+
+    } catch (IncorrectDateFormatException
```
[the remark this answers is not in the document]
bion: @swatkat1 this was marked resolved without having been addresed

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -72,23 +83,62 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+    }
+
+    return redirect(
+        routes.TrustedIntermediaryController.dashboard(
+            /* paramName=  nameQuery */
+            Optional.empty(),
+            /* paramName=  searchDate */
+            Optional.empty(),
+            /* paramName=  page */
+            Optional.empty()));
```
bion, line 139: suggestion (blocking): 
```suggestion
            /* nameQuery=  */
            Optional.empty(),
            /* searchDate=  */
            Optional.empty(),
            /* page= */
            Optional.empty()));
```

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -72,23 +83,62 @@ public Result dashboard(Http.Request request, Optional<String> search, Optional<
...
+        userRepository.getTrustedIntermediaryGroup(civiformProfile.get());
+    if (trustedIntermediaryGroup.isEmpty()) {
+      return notFound();
+    }
+
+    Form<UpdateApplicantDob> form =
+        formFactory.form(UpdateApplicantDob.class).bindFromRequest(request);
+    try {
+      tiService.updateApplicantDateOfBirth(accountId, form);
+
```
bion, line 123: suggestion (blocking):
```suggestion
```

server/app/controllers/ti/TrustedIntermediaryController.java
bion: @swatkat1 this comment was resolved without having been addressed

server/app/controllers/ti/TrustedIntermediaryController.java
bion: @swatkat1 was this addressed?

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -134,12 +193,32 @@ public Result addApplicant(Long id, Http.Request request) {
...
+  private Result redirectToDashboardWithUpdateDateOfBirthError(String errorMessage) {
+    return redirect(
+            routes.TrustedIntermediaryController.dashboard(
+                    /* paramName=  nameQuery */
+                    Optional.empty(),
+                    /* paramName=  searchDate */
+                    Optional.empty(),
+                    /* paramName=  page */
+                    Optional.of(1))
```
bion, line 220: suggestion (blocking): fix comments to match the style used elsewhere in the codebase

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -183,7 +183,12 @@ private DivTag maybeRenderTiButton(Optional<CiviFormProfile> profile, String use
       String tiDashboardText = "Trusted intermediary dashboard";
       String tiDashboardLink =
           controllers.ti.routes.TrustedIntermediaryController.dashboard(
-                  Optional.empty(), Optional.empty())
+                  /* paramName=  nameQuery */
+                  Optional.empty(),
+                  /* paramName=  searchDate */
+                  Optional.empty(),
+                  /* paramName=  page */
+                  Optional.empty())
```
bion, line 191: suggestion (blocking): fix comments to match the style used elsewhere in the codebase

server/app/services/ti/TrustedIntermediaryService.java
bion: @swatkat1 this comment was resolved without being addressed

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,117 @@
...
+import models.Applicant;
+import models.TrustedIntermediaryGroup;
+import org.slf4j.LoggerFactory;
+import play.data.Form;
+import repository.SearchParameters;
+import repository.UserRepository;
+import services.DateConverter;
+import services.applicant.exception.ApplicantNotFoundException;
+
+public class TrustedIntermediaryService {
```
bion, line 22: suggestion (blocking): please add a unit test for this service

server/app/models/Account.java
```diff
@@ -155,9 +156,26 @@ public void removeAdministeredProgram(ProgramDefinition program) {
...
+    return "<Unnamed User>";
+  }
+
+  /**
+   * Returns the Date of Birth, as an Optional LocalDate, of the most-recently created Applicant
+   * associated with this Account. There is no particular reason for an Account to have more than
+   * one Applicant - this was a capability we built but did not use - so the ordering is somewhat
+   * arbitrary unnecessary.
+   */
+  public Optional<LocalDate> getApplicantDateOfBirth() {
```
bion, line 173: suggestion (blocking): please add a unit test for this new method

bion (changes requested):

Issue #1900, opened 2022-02-10
Fields with numbers - use dropdown instead of up/down arrow


glorialiou:
Number UI (up/down arrow) is confusing and has a small touch target. This can occur both for admins (e.g. when trying to set minimum/required # of choices) and for applicants (e.g. when trying to select how many dependents they have).

rahaghas:
Agree with removing the up/down arrow but don't see a need for incorporating a dropdown for number fields. Users should be able to accessibly enter a number without the use of a dropdown.

th174:
GOV.UK recommends against using `<input type="number"/>`, because it comes with various accessibility issues. Instead, it recommends this: `<input type="text" inputmode="numeric" pattern="[0-9]*">`

https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

Thoughts?

clouser2:
Removing the stepper styling can be achieved with some custom CSS (https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp). Note that Tailwind doesn't have any easy-to-use selector for disabling these controls right now.

I'm certainly open to the pattern of using inputmode="numeric". One note is that it appears that all of the accessibility issues from the article are marked fixed as of now. I believe the remaining benefits described in that article would be:
  * Browsers discarding non-number input
  * Scrolling would no longer increment the number

I don't think there's an easy decision here without tradeoffs. One could argue that scrolling to increment could be a benefit in use cases where the number field actually represents a number.

My inclination right now would be to start by visually removing the stepper control as @rahaghas described above. I'll defer to others as to whether we want to switch to `type="text" inputmode="numberic"`.

clouser2:
Spoke with @rahaghas, who mentioned that `type="text" inputmode="numeric"` would likely be preferable from a UX perspective as well. As such, I'll likely revert the stepper change and instead convert all `type=number` fields to follow this pattern.

Estimated effort: <1 hour.

clouser2:
Updated estimate: 1 day

[...]

PR #2531, opened 2022-05-23
Remove number steppers for input[type='number']

clouser2:
### Description
Note: This removes the steppers globally since the original concern was expressed both for applicant UI and admin-facing UI.

Before: Screen Shot 2022-05-23 at 10 42 07 AM
After: 
Screen Shot 2022-05-23 at 10 40 41 AM

### Checklist

### Issue(s)
#1900

### DIFF
--- a/server/app/assets/stylesheets/styles.css
+++ b/server/app/assets/stylesheets/styles.css
@@ -33,4 +33,14 @@
   .cf-accordion-visible > .cf-accordion-content {
     @apply h-auto mt-2 mb-1;
   }
+
+  /* Remove number steppers */
+  input::-webkit-outer-spin-button,
+  input::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+  input[type=number] {
+    -moz-appearance: textfield;
+  }
 }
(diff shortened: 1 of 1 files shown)


bion (approved):

shanemc-goog:
Just curious: do number inputs get spinners by default?

clouser2:
> Just curious: do number inputs get spinners by default?

I believe so, yes (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number).

server/app/assets/stylesheets/styles.css
```diff
@@ -33,4 +33,14 @@
   .cf-accordion-visible > .cf-accordion-content {
     @apply h-auto mt-2 mb-1;
   }
+
+  /* Remove number steppers (https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp) */
+  input::-webkit-outer-spin-button,
```
th174, line 38: This removes number steppers globally. Is that desired, or do we want to remove number steppers for the numerical type of question specifically?

If it's the latter, we might need to specify a CSS class.

th174 (approved):

server/app/assets/stylesheets/styles.css
clouser2: Good callout! I updated the PR description to call out that this will be a global change since the original issue referenced both the admin and applicant-facing UI.

Should we want to reintroduce this for a specific input at a later point, we can introduce the CSS class when needed.

PR #2544, opened 2022-05-24
Log API authentication failures

bion:
Log at WARN level with uniform message structure. Discussed in [design doc](https://docs.google.com/document/d/1g2XkMhDe6EaVGQabEes4KoDWpzL9F4S5pl2fUXoilSc/edit)

### DIFF
--- a/server/app/auth/ApiAuthenticator.java
+++ b/server/app/auth/ApiAuthenticator.java
@@ -3,6 +3,7 @@ package auth;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.time.Instant;
+import java.util.Optional;
 import javax.inject.Inject;
 import javax.inject.Provider;
 import models.ApiKey;
@@ -13,6 +14,8 @@ import org.pac4j.core.credentials.Credentials;
 import org.pac4j.core.credentials.UsernamePasswordCredentials;
 import org.pac4j.core.credentials.authenticator.Authenticator;
 import org.pac4j.core.exception.BadCredentialsException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import services.apikey.ApiKeyService;
 
 /**
@@ -44,6 +47,7 @@ import services.apikey.ApiKeyService;
  */
 public class ApiAuthenticator implements Authenticator {
 
+  private static final Logger logger = LoggerFactory.getLogger(ApiAuthenticator.class);
   private final Provider<ApiKeyService> apiKeyService;
 
   @Inject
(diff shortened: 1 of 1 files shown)


server/app/auth/ApiAuthenticator.java
```diff
@@ -74,18 +78,20 @@ public void validate(Credentials rawCredentials, WebContext context, SessionStor
...
     // requests with the invalid key do not put pressure on the database.
-    ApiKey apiKey =
-        apiKeyService
-            .get()
-            .findByKeyIdWithCache(keyId)
-            .orElseThrow(() -> new BadCredentialsException("API key does not exist"));
+    Optional<ApiKey> maybeApiKey = apiKeyService.get().findByKeyIdWithCache(keyId);
+
+    if (!maybeApiKey.isPresent()) {
+      unauthorized(context, "API key does not exist: " + keyId);
```
clouser2, line 84: High-level question on the design. API keys are meant to only be shown once to the API user. Consider the situation where there was a bug in findByKeyIdWithCache where the API key *did* exist, was valid, but wasn't being returned by the method. Here, someone with server logs access would be able to see the actual key. If we then fixed the bug, the key would effectively be "leaked".

This may not be a concern due to the presumable level of access someone with server logs access has, but I wanted to explicitly call it out.

My whole concern could be avoided if we didn't print the key ID here. In my opinion, it's not necessary information for debugging, since there's no real action for someone observing logs to take based on this.

clouser2: Same for the below updates as well.

clouser2 (approved):

server/app/auth/ApiAuthenticator.java
clouser2, line 84: I believe this is causing the unit test failure from CI since it's doing an exact match on the error message.

swatkat1 (approved):

server/app/auth/ApiAuthenticator.java
```diff
@@ -95,14 +101,24 @@ public void validate(Credentials rawCredentials, WebContext context, SessionStor
...
 
     String saltedCredentialsSecret = apiKeyService.get().salt(credentials.getPassword());
     if (!saltedCredentialsSecret.equals(apiKey.getSaltedKeySecret())) {
-      throw new BadCredentialsException("Invalid secret for key ID: " + keyId);
+      unauthorized(context, "Invalid secret for key ID: " + keyId);
     }
   }
+
+  private void unauthorized(WebContext context, String cause) {
+    logger.warn(
```
swatkat1, line 117: @bion  - Do we have a logger.xml to logs these? If not should we implement this  like the one you shared yesterday?

bion: There's already a logger configuration in the system at server/conf/logback.xml. This logger and all others in the system use the configuration there.

server/app/auth/ApiAuthenticator.java
bion: There's a few different pieces:

* API key: the database-backed resource that provides auth semantics for API calls
* API key ID: a random string that uniquely identifies an API key
* API key secret: a random string that, paired with the ID, allows an API caller to use an API key
* API key credentials: the combination of API key ID and secret, joined on ':' and base-64 encoded

We log the key ID here so that maintainers of a deployment can see which keys callers are attempting and failing to authenticate with. We never log the secret.

clouser2 (approved):

server/app/auth/ApiAuthenticator.java
bion: ah woops, yeah i meant to reply to your other comments. sorry!

PR #2559, opened 2022-05-24
Update nav to select the active nav section

### LINKED ISSUE #1906: Add tab UI to indicate toolbar navigation
See here for example screenshot / more info: https://docs.google.com/presentation/d/1goxiepPXPrW3wK8wUSNoCpcf7iXIohXfj96iTYiQxo0/edit#slide=id.gf083eb9b36_0_1333

See here for what it could look like (add underline and different color to indicate current tab): https://docs.google.com/presentation/d/1goxiepPXPrW3wK8wUSNoCpcf7iXIohXfj96iTYiQxo0/edit#slide=id.g101493a1a12_0_733

clouser2:
### Description
This implements the guidance from https://github.com/seattle-uat/civiform/issues/1906#issuecomment-1106881407, with the exception that there isn't a min-width set. The mock from that comment looks like there might be a larger redesign of the nav forthcoming. For now, keeping the nav as-is. The majority of the infrastructure to support this should remain the same.

Before:
Screen Shot 2022-05-24 at 2 51 31 PM

After:
Screen Shot 2022-05-24 at 2 51 10 PM

### Issue(s)
#1906


server/app/views/admin/AdminLayout.java
```diff
@@ -101,12 +122,35 @@ private ContainerTag renderNavBar() {
...
+
     return adminHeader
-        .with(headerLink("Programs", programLink))
-        .with(headerLink("Questions", questionLink))
-        .with(headerLink("Versions", versionLink))
-        .with(headerLink("Intermediaries", intermediaryLink))
-        .with(headerLink("API keys", apiKeysLink))
+        .with(
+            headerLink(
+                "Programs", programLink, activePage == NavPage.PROGRAMS ? activeNavStyle : ""))
```
clouser2, line 137: This pattern is somewhat inelegant in that we have a text description and an equality check for every link. I could imagine trying to come up with something more `switch`-based, but the fact that there are only 5 sections led me to think the benefit wasn't really worth it given the small scope.

server/app/views/admin/AdminLayout.java
```diff
@@ -101,12 +122,35 @@ private ContainerTag renderNavBar() {
...
     String apiKeysLink = controllers.admin.routes.AdminApiKeysController.index().url();
 
+    String activeNavStyle =
+        StyleUtils.joinStyles(
+            BaseStyles.TEXT_SEATTLE_BLUE,
+            Styles.FONT_MEDIUM,
+            Styles.BORDER_B_2,
+            BaseStyles.BORDER_SEATTLE_BLUE);
+
+    NavPage activePage = activeNavPage.orElseThrow();
```
clouser2, line 132: The decision to throw felt preferable to having an "unspecified" enum and not highlighting to me in that the page would fail to render when a layout value hadn't been specified. I assessed callsites to make sure we were setting it (and hopefully tests will exercise all of these paths).

bion: Suggestion (blocking): an alternative that would ensure it's set without potentially blocking the admin's ability to use the system could be to introduce:
```java
public class AdminLayoutFactory {
  @Inject
  public AdminLayoutFactory(ViewUtils viewUtils, Config configuration) {...}
  
  public AdminLayout getLayout(NavPage navPage) {
    return new AdminLayout(viewUtils, configuration, navPage);
  }
}
```

and then change:
```java
public class AdminLayout extends BaseHtmlLayout {
  
  // make the constructor package-private, remove the @Inject annotation, and require the navPage arg
  AdminLayout(ViewUtils viewUtils, Config configuration, NavPage navPage)
}
```

thoughts?

server/app/views/admin/AdminLayout.java
bion: Suggestion (non-blocking): an alternative could be to encapsulate this and the logic already in `headerLink` in a `AdminHeaderLink` class with a `render(NavPage navPage)` method

bion (changes requested):

server/app/views/admin/AdminLayout.java
bion: also instead of hand-writing the factory could instead use https://github.com/google/guice/wiki/AssistedInject

clouser2: Done. I shied away from using AssistedInject since we don't make very heavy use of Guice and the amount of boilerplate for the factory was relatively low.

server/app/views/admin/AdminLayout.java
clouser2: Indeed. One wrinkle to doing that or other helpers around the nav page is that we also use "headerLink" for rendering the logout button, which isn't really nav per se.

If we wanted style reuse, we could either add LOGOUT as an enum value. This feels like breaking the encapsulation of what the enum represents.

Another alternative would be to extract out the styles common to both. At that point, I think we'd have more boilerplate around all of this and it wouldn't necessarily improve readability. 

Assuming that seems reasonable to you, I'm going to keep as-is for now.

bion: sgtm, thanks for talking it through with me!

bion (approved):

goldblatt (approved):

clouser2:
Latest push was just to add this logic to the newly forked ProgramIndexViewV2.java.

PR #2568, opened 2022-05-25
[Admin UI redesign] Update the program list view (guarded behind query param flag)

clouser2:
### Description
This redesigns the UI for the program list view based on UXD mocks. Currently, this will only render when the URL contains the `v2=1` query parameter. The plan is to get this in front of a program admin for user testing, iterate, and remove the old UI.

Most changes here are simply those needed to get the styling in `ProgramIndexViewV2` to match those of the mocks.

I went back and forth between creating Java wrappers responsible for rendering the new button types (similar to `LinkElement`, but found the amount of plumbing and customization to be quite awkward. Instead, I landed more on a pattern more based on composition of:
1. Defining styling for the buttons in `Styles.java`
2. Defining a few new helpers in `BaseHtmlView.java` for making a button element with SVG content, making it a "link" button by attaching an onclick handler, etc.

Still TODO after initial feedback:
  * Updates to integration tests

Before:
Screen Shot 2022-05-25 at 9 45 42 AM

After: 
Screen Shot 2022-05-25 at 9 45 25 AM

### Issue(s)
#1238


server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -4,18 +4,19 @@
 import static j2html.TagCreator.div;
 import static j2html.TagCreator.each;
 import static j2html.TagCreator.h1;
-import static j2html.TagCreator.input;
-import static j2html.TagCreator.label;
 import static j2html.TagCreator.p;
+import static j2html.TagCreator.span;
```
clouser2, line 8: Note to the reviewer: The review flow collapsed this by default for me. Don't forget this file since this is where 99% of the changes are :).

server/app/views/components/Icons.java
```diff
@@ -182,7 +285,10 @@ public static ContainerTag svg(String pathString, int pixelSize) {
   }
 
   public static ContainerTag svg(String pathString, int width, int height) {
-    return svg(pathString).attr("viewBox", String.format("0 0 %1$d %2$d", width, height));
+    return svg(pathString)
+        .attr("viewBox", String.format("0 0 %1$d %2$d", width, height))
```
clouser2, line 289: Setting the viewBox to a specific height/width is insufficient to actually cause the SVG's bounds to match. Here, I explicitly set width / height of the SVG element, which is more consistent with what one would expect based on the method / argument names.

server/tailwind.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     transform: {
       java: (content) => {
         var output = []
-        for (const match of content.matchAll(/"([\w-/:]+)"/g)) {
+        for (const match of content.matchAll(/"([\w-/:.]+)"/g)) {
```
clouser2, line 17: Adding the "." here is subtle, but is necessary in order to enable tailwind styles with decimal components (e.g. space-x-3.5).

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -85,48 +87,64 @@ public Content render(
...
     HtmlBundle htmlBundle =
         layout
             .getBundle()
             .setTitle(pageTitle)
             .addMainContent(contentDiv)
-            .addModals(publishAllModal)
             .addFooterScripts(layout.viewUtils.makeLocalJsTag("admin_programs"));
+    if (maybePublishModal.isPresent()) {
+      htmlBundle = htmlBundle.addModals(maybePublishModal.get());
+    }
```
bion, line 100: suggestion (non-blocking): no need to reassign the variables, addModals mutates
```suggestion
    maybePublishModal.ifPresent(htmlBundle::addModals);
```

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -137,81 +155,168 @@ public ProgramDefinition getDisplayProgram(
...
-            .withClasses(Styles.W_FULL)
+
+    Tag draftRow = null;
+    if (draftProgram.isPresent()) {
+      List<Tag> draftRowActions = Lists.newArrayList();
+      List<Tag> draftRowExtraActions = Lists.newArrayList();
+      draftRowActions.add(renderEditLink(false, draftProgram.get(), request));
+      draftRowExtraActions.add(renderManageProgramAdminsLink(draftProgram.get()));
+      draftRowExtraActions.add(renderManageTranslationsLink(draftProgram.get()));
+      draftRow = renderProgramRow(false, draftProgram.get(), draftRowActions, draftRowExtraActions);
```
bion, line 265: suggestion (blocking): add comment for name of boolean param

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -137,81 +155,168 @@ public ProgramDefinition getDisplayProgram(
...
+      List<Tag> activeRowExtraActions = Lists.newArrayList();
+      activeRowActions.add(maybeRenderViewApplicationsLink(activeProgram.get(), profile.get()));
+      if (!draftProgram.isPresent()) {
+        activeRowActions.add(renderEditLink(true, activeProgram.get(), request));
+        activeRowExtraActions.add(renderManageProgramAdminsLink(activeProgram.get()));
+      }
+      activeRowActions.add(renderViewLink(activeProgram.get()));
+      activeRow =
+          renderProgramRow(
+              true,
```
bion, line 280: suggestion (blocking): add comment for name of boolean param

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -137,81 +155,168 @@ public ProgramDefinition getDisplayProgram(
...
-                Styles.BORDER, Styles.BORDER_GRAY_300, Styles.BG_WHITE, Styles.ROUNDED, Styles.P_4);
+                p(programTitleText)
+                    .withClasses(
+                        ReferenceClasses.ADMIN_PROGRAM_CARD_TITLE,
+                        Styles.W_1_4,
+                        Styles.PY_7,
+                        Styles.TEXT_BLACK,
+                        Styles.FONT_BOLD,
+                        Styles.TEXT_XL),
+                div().withClass(Styles.FLEX_GROW).with(draftRow, activeRow));
```
bion, line 299: suggestion (blocking): to avoid `null` values could assign this div higher up and call `with` at the end of the respective `if` statements for building these rows

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -137,81 +155,168 @@ public ProgramDefinition getDisplayProgram(
...
+      draftRow = renderProgramRow(false, draftProgram.get(), draftRowActions, draftRowExtraActions);
+    }
+
+    Tag activeRow = null;
+    if (activeProgram.isPresent()) {
+      List<Tag> activeRowActions = Lists.newArrayList();
+      List<Tag> activeRowExtraActions = Lists.newArrayList();
+      activeRowActions.add(maybeRenderViewApplicationsLink(activeProgram.get(), profile.get()));
+      if (!draftProgram.isPresent()) {
+        activeRowActions.add(renderEditLink(true, activeProgram.get(), request));
```
bion, line 274: suggestion (blocking): add comment for name of boolean param

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -230,108 +335,74 @@ private static Instant extractLastUpdated(
...
-    }
-    throw new IllegalArgumentException("Program neither active nor draft.");
+  Tag renderViewLink(ProgramDefinition program) {
+    String viewLink =
+        baseUrl
+            + controllers.applicant.routes.RedirectController.programByName(program.slug()).url();
+    ContainerTag button =
+        ButtonUtils.makeSvgTextButton("View", Icons.VISIBILITY_SVG_PATH)
+            .withClass(AdminStyles.TERTIARY_BUTTON_STYLES);
+    return ButtonUtils.asLinkButton(button, viewLink);
```
bion, line 345: issue (blocking): this design introduces a regression in that the the program deep link is no longer visible to the admin. The purpose of displaying it in the program list isn't to link to the application for the admin, but rather to display it to the admin so they can copy it for use outside of CiviForm (which is why the link is not clickable in the current UI).

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -230,108 +335,74 @@ private static Instant extractLastUpdated(
...
+                  activeProgram.id(), Optional.empty(), Optional.empty())
+              .url();
+
+      ContainerTag button =
+          ButtonUtils.makeSvgTextButton("Applications", Icons.TEXT_SNIPPET_SVG_PATH)
+              .withId("program-view-apps-link-" + activeProgram.id())
+              .withClass(AdminStyles.TERTIARY_BUTTON_STYLES);
+      return ButtonUtils.asLinkButton(button, editLink);
+    }
+    return null;
```
bion, line 397: suggestion (blocking): please avoid returning null from methods and return an Optional instead

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -230,108 +335,74 @@ private static Instant extractLastUpdated(
...
+          routes.AdminApplicationController.index(
+                  activeProgram.id(), Optional.empty(), Optional.empty())
+              .url();
+
+      ContainerTag button =
+          ButtonUtils.makeSvgTextButton("Applications", Icons.TEXT_SNIPPET_SVG_PATH)
+              .withId("program-view-apps-link-" + activeProgram.id())
+              .withClass(AdminStyles.TERTIARY_BUTTON_STYLES);
+      return ButtonUtils.asLinkButton(button, editLink);
+    }
```
bion, line 396: issue (non-blocking): I know the logic here is mostly copied-over, but we should consider refactoring it. At a minimum there's an issue here with N+1 queries for checking program authorization. Feel free to ignore for now but maybe add a TODO or file a tech-debt issue.

server/app/views/components/ButtonUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import j2html.tags.ContainerTag;
+import j2html.tags.Tag;
+import org.apache.commons.lang3.RandomStringUtils;
+import play.filters.csrf.CSRF;
+import play.mvc.Http;
+import scala.Option;
+import views.style.Styles;
+
+/** Contains helpers for creating buttons for use in the UI. */
+public final class ButtonUtils {
```
bion, line 17: issue (blocking): the views already have two classes that contain logic for creating different types of buttons, I'd prefer to refactor toward having one vs introducing a third

server/app/views/components/ButtonUtils.java
```diff
@@ -0,0 +1,49 @@
...
+            .withMethod("POST")
+            .withAction(href)
+            .with(input().isHidden().withValue(csrfToken).withName("csrfToken"));
+
+    return el.attr("form", formId).with(hiddenForm);
+  }
+
+  public static ContainerTag asLinkButton(ContainerTag el, String href) {
+    return el.attr("onclick", String.format("document.location.href = \"%s\"", href));
+  }
```
bion, line 41: issue (blocking): this is nearly a duplicate of https://github.com/seattle-uat/civiform/blob/main/server/app/views/BaseHtmlView.java#L66-L70

server/app/views/components/Icons.java
```diff
@@ -9,6 +9,8 @@
  */
 public class Icons {
 
+  public static final String ADD_SVG_PATH =
+      "M9.125 15.833V10.875H4.167V9.125H9.125V4.167H10.875V9.125H15.833V10.875H10.875V15.833Z";
```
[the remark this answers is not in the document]
bion: sgtm, though the autoformatter will edit them into multi-line strings. you may have to add magic comments to disable autoformatting them

server/app/views/components/Modal.java
```diff
@@ -97,6 +93,8 @@ public ModalBuilder setTriggerButtonText(String triggerButtonText) {
     }
 
     public ModalBuilder setTriggerButtonContent(Tag triggerButtonContent) {
+      Preconditions.checkState(
+          "button".equals(triggerButtonContent.getTagName()), "content must be of type button");
       this.triggerButtonContent = Optional.ofNullable(triggerButtonContent);
```
bion, line 98: note (non-blocking): the newer version of j2html has java types for each element which will eliminate the need for this sort of thing. working with Sami to get https://github.com/seattle-uat/civiform/pull/2281 merged would be a good fixit week item

server/app/views/style/Styles.java
```diff
@@ -699,6 +699,36 @@ public final class Styles {
...
+  public static final String SPACE_Y_2_5 = "space-y-2.5";
+  public static final String SPACE_Y_3 = "space-y-3";
+  public static final String SPACE_Y_3_5 = "space-y-3";
+  public static final String SPACE_Y_4 = "space-y-4";
+  public static final String SPACE_Y_5 = "space-y-5";
+  public static final String SPACE_Y_6 = "space-y-6";
+  public static final String SPACE_Y_7 = "space-y-7";
+  public static final String SPACE_Y_8 = "space-y-8";
+  public static final String SPACE_Y_9 = "space-y-9";
+  public static final String SPACE_Y_10 = "space-y-10";
```
bion, line 731: note (non-blocking): i'm surprised these work, when we generated this file we intended to include all available classes

bion (changes requested):
Looks really good, huge improvement to the UI!

server/app/views/components/Modal.java
syntapy: @bion I think I'm getting closed. pre recent merge there were just a couple types of browser tests failing. Anyhow, could you run the CI tests for me rn?

server/app/views/admin/programs/ProgramIndexViewV2.java
clouser2: Filed #2582 and left TODOs for the potential N+1 behavior.

clouser2 (commented):
Addressed all comments except for button construction duplication. Once I've pondered that, I'll send another response and re-request review.

server/app/views/components/Icons.java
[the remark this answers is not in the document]
clouser2: Sounds good. I'll give that a try in a follow-on PR.

server/app/views/components/Modal.java
clouser2: Agreed that would be a huge improvement here. @syntapy, I'm happy to help (and / or review) if you run into any issues.

I'll be out 5/27-5/31, but happy to coordinate with you on Wednesday!

server/app/views/components/ButtonUtils.java
clouser2: Discussed offline and agreed. My main goal is to reduce the number of ways we attempt to wrap creation of the underlying element. In my opinion, that pattern doesn't grow extremely well and eventually enough customizations are added to it (to support more scenarios) that it ends up being an almost-complete set of the functionality offered by the original element it was wrapping. For now, we were planning on the following:

1. Modify current PR so that everything in `ButtonUtils` moves to `BaseHtmlView`. Where appropriate (e.g. redirectButton), modify existing callers to pass in a constructed button element that the method augments (rather than constructing a new button). Don’t touch existing `BaseHtmlView.button()` helpers.
2. Post PR-submit, do a pass to remove unnecessary button methods from `BaseHtmlView`.
3. As more views are converted, move away from the `LinkElement` pattern in favor of using the helpers in `BaseHtmlView`.
4. If we find the number of helpers in `BaseHtmlView` to be unwieldy in the future, move out to a static helper class

server/app/views/components/ButtonUtils.java
clouser2: Agreed. We now use a variant of that helper.

clouser2 (commented):
All feedback (hopefully) incorporated.

browser-test/src/admin_program_list.test.ts
```diff
@@ -50,4 +50,5 @@ describe('Most recently updated program is at top of list.', () => {
 
     await endSession(browser)
   })
-})
\ No newline at end of file
+})
```
kmcnellis, line 53: [nit] no changes here but whitespace

server/app/views/components/Icons.java
kmcnellis: Maybe leave a comment to that effect? If you need to explain it to us here, others reading the code might have the same question

server/app/views/admin/programs/ProgramIndexViewV2.java
bion: lgtm

server/app/views/admin/programs/ProgramIndexViewV2.java
bion: > What do you think about "View" instead being changed to something that copies the program link to the clipboard?

that sounds fine, just want it to be super obvious how to get the deep link before switching to the new version of the page

server/app/views/style/Styles.java
bion: aha

bion (approved):

server/app/views/components/Icons.java
kmcnellis: Maybe leave a comment to that effect? If you need to explain it to us here, others reading the code might have the same question

kmcnellis (changes requested):

server/app/views/admin/programs/ProgramIndexViewV2.java
kmcnellis: I really wish github didn't do this :(

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -139,81 +154,176 @@ public ProgramDefinition getDisplayProgram(
...
+                    p().with(
+                            span(updatedPrefix),
+                            span(formattedUpdateTime).withClass(Styles.FONT_SEMIBOLD)),
+                    p().with(
+                            span(String.format("%d", blockCount)).withClass(Styles.FONT_SEMIBOLD),
+                            span(blockCount == 1 ? " screen, " : " screens, "),
+                            span(String.format("%d", questionCount))
+                                .withClass(Styles.FONT_SEMIBOLD),
+                            span(questionCount == 1 ? " question" : " questions"))),
+            div().withClass(Styles.FLEX_GROW),
```
kmcnellis, line 227: did you run the formatter?  These newlines seem inconsistent

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -232,108 +342,78 @@ private static Instant extractLastUpdated(
...
-          .setText(linkText)
-          .setStyles(Styles.MR_2)
-          .asAnchorText();
-    } else {
-      return div();
+  private Optional<Tag> maybeRenderViewApplicationsLink(
+      ProgramDefinition activeProgram, CiviFormProfile userProfile) {
+    // TODO(#2582): Determine if this has N+1 query behavior and fix if
+    // necessary.
+    boolean userIsAuthorized = true;
```
kmcnellis, line 390: It makes me feel better if authorization-type bools default to false, and get set to true only if the user is authorized.

kmcnellis (changes requested):

server/app/views/admin/programs/ProgramIndexViewV2.java
clouser2: Yah, I found it pretty surprising. It seems that collapsing makes it possible to hide the changes that actually matter. I'd be more comfortable if this was applied to binary-type files rather than text files.

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -139,81 +154,176 @@ public ProgramDefinition getDisplayProgram(
...
+    Optional<Instant> updatedTime = program.lastModifiedTime();
+    if (isActive) {
+      badgeText = "Active";
+      badgeBGColor = BaseStyles.BG_CIVIFORM_GREEN_LIGHT;
+      badgeFillColor = BaseStyles.TEXT_CIVIFORM_GREEN;
+      updatedPrefix = "Published on ";
+    }
+
+    String formattedUpdateTime =
+        updatedTime.isPresent() ? renderDateTime(updatedTime.get(), zoneId) : "unknown";
```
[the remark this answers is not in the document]
clouser2: `isPresent()` doesn't appear to take a lambda, so it's not possible to chain in this way from what I can tell.

The following does work:
`String formattedUpdateTime =
        updatedTime.map(t -> renderDateTime(t, zoneId)).orElse("unknown");`

I think these two forms are somewhat equivalent in readability. I'm still getting used to some of the optional syntax and am happy to go with the map syntax instead.

server/app/views/admin/programs/ProgramIndexViewV2.java
clouser2: I did run the formatter. The indentation seems fine to me, but I might be missing something.

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -232,108 +342,78 @@ private static Instant extractLastUpdated(
...
-      return "Active, with draft";
-    } else if (draftProgram.isPresent()) {
-      return "Draft";
-    } else if (activeProgram.isPresent()) {
-      return "Active";
-    }
-    throw new IllegalArgumentException("Program neither active nor draft.");
+  Tag renderViewLink(ProgramDefinition program) {
+    // TODO(#1238): Rather than navigating to the program link
+    // consider renaming the action and copying the link to the clipboard.
```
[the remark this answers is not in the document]
clouser2: Sounds good. Updated the TODO to reflect this once we figure out what is desirable.

server/app/views/admin/programs/ProgramIndexViewV2.java
clouser2: Works for me. Rather than defaulting to false, I explicitly set it in each case. Also note that if I tried defaulting to false, that would cause an empty catch block, which would trigger a warning :).

server/app/views/components/Icons.java
clouser2: Done.

server/app/views/admin/programs/ProgramIndexViewV2.java
kmcnellis: It's probably fine.  It's just that half of them have the div().withClass(classes) on the same line, and half split it up over 2 or 3 lines.

server/app/views/admin/programs/ProgramIndexViewV2.java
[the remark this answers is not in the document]
kmcnellis: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html
ifPresent is overloaded - there's a version with a consumer, and one without.  

Yeah, I don't see any real reason to go with the map syntax.  Just wanted to give you more options!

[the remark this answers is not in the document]
kmcnellis: https://www.baeldung.com/java-optional#if-present

kmcnellis (changes requested):
I've been looking at the changes locally, and I think there's a couple UI issues.  Feel free to address them here or in a new PR:
(this is in half-macbook screen size)
image

1) The dot before "draft" and "active" gets smaller with each instance of it
2) The "draft" and "active" pills are not all allgned in the same column
3) Sometime "edit" and "view" are one line, sometimes on 2
4) The drop-down menu is center aligned 
image

kmcnellis:
> I've been looking at the changes locally, and I think there's a couple UI issues. Feel free to address them here or in a new PR: (this is in half-macbook screen size) image
> 
> 1. The dot before "draft" and "active" gets smaller with each instance of it
> 2. The "draft" and "active" pills are not all allgned in the same column
> 3. Sometime "edit" and "view" are one line, sometimes on 2
> 4. The drop-down menu is center aligned
>    image

The dot thing might be intentional, but it looks accidental

server/app/views/admin/programs/ProgramIndexViewV2.java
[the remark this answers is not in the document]
clouser2: `ifPresent` takes a consumer, which has return type void. So this ends up not being compatible with trying to assign the result of applying the consumer. Either way, I'm fine with the map syntax :)

rahaghas:
Dot thing is definitely not intentional! Thanks for catching

clouser2:
> > I've been looking at the changes locally, and I think there's a couple UI issues. Feel free to address them here or in a new PR: (this is in half-macbook screen size) image
> > 
> > 1. The dot before "draft" and "active" gets smaller with each instance of it
> > 2. The "draft" and "active" pills are not all allgned in the same column
> > 3. Sometime "edit" and "view" are one line, sometimes on 2
> > 4. The drop-down menu is center aligned
> >    image
> 
> The dot thing might be intentional, but it looks accidental

Thanks for trying this out. These all very good issues that hopefully can be solved with a few media breakpoints. As I understand, we don't need to optimize the Admin UI for very small screen sizes, but half a mac screen is still pretty high resolution, so it'd be good to fix these up.

I left a TODO for right now capturing the issues you described. That way we can get the bulk in and iterate from there (which will hopefully make for easier reviews).

RE "Sometimes edit and view are one line, sometimes 2. That was intentional per Raha's mocks. I think the idea is that clicking edit when there's only a draft would create a new version. Once there's a draft available, editing only happens on the draft row, not the active row.

kmcnellis (approved):

server/app/views/components/Modal.java
syntapy: @bion @clouser2 It seems like I can just remove that line then right? 96 & 97?

syntapy: @clouser2 sorry I didn't see your comment until just a min ago

clouser2: Yah, removing lines 96 and 97 should be good. If Preconditions isn't used elsewhere in the file, you might be able to remove the import as well.

PR #2570, opened 2022-05-25
Fork ProgramIndexView into a separate ProgramIndexViewV2 in preparation for larger UI changes

clouser2:
### Description
In #2568, we'll redesigning the program list view. Initially, changes will be guarded behind the `v2` query parameter (for program admin user testing). I intend to fork the view entirely and manually keep in sync (for what should be a short period).

Presently `ProgramIndexViewV2` is identical to `ProgramIndexView`. This is intentional and split into a separate PR to aid in the diff generated by the following larger PR that changes the HTML.

### Issue(s)
#1238

### DIFF
--- a/server/app/controllers/admin/AdminProgramController.java
+++ b/server/app/controllers/admin/AdminProgramController.java
@@ -24,6 +24,7 @@ import services.program.ProgramNotFoundException;
 import services.program.ProgramService;
 import views.admin.programs.ProgramEditView;
 import views.admin.programs.ProgramIndexView;
+import views.admin.programs.ProgramIndexViewV2;
 import views.admin.programs.ProgramNewOneView;
 
 /** Controller for handling methods for admins managing program definitions. */
@@ -31,6 +32,7 @@ public class AdminProgramController extends CiviFormController {
 
   private final ProgramService service;
   private final ProgramIndexView listView;
+  private final ProgramIndexViewV2 listViewV2;
   private final ProgramNewOneView newOneView;
   private final ProgramEditView editView;
   private final FormFactory formFactory;
@@ -42,6 +44,7 @@ public class AdminProgramController extends CiviFormController {
   public AdminProgramController(
       ProgramService service,
       ProgramIndexView listView,
+      ProgramIndexViewV2 listViewV2,
       ProgramNewOneView newOneView,
       ProgramEditView editView,
       VersionRepository versionRepository,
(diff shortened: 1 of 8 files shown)


server/app/controllers/admin/AdminProgramController.java
```diff
@@ -65,6 +69,11 @@ public AdminProgramController(
   @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
   public Result index(Request request) {
     Optional<CiviFormProfile> profileMaybe = profileUtils.currentUserProfile(request);
+    // TODO(#1238): Remove the old view once the new rendering
+    // should be default.
+    if (request.queryString().containsKey("v2")) {
```
clouser2, line 74: Note to reviewer: The rationale here is that we can get this in front of a program admin to test live and iterate on the design. If there's a feature miss or issue, they needn't be blocked while we iterate.

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -0,0 +1,337 @@
...
+                h1(pageTitle).withClasses(Styles.MY_4),
+                div()
+                    .withClasses(Styles.FLEX, Styles.ITEMS_CENTER)
+                    .with(renderNewProgramButton())
+                    .with(div().withClass(Styles.FLEX_GROW))
+                    .condWith(programs.anyDraft(), publishAllModal.getButton()),
+                div()
+                    .withClasses(ReferenceClasses.ADMIN_PROGRAM_CARD_LIST, Styles.INVISIBLE)
+                    .with(
+                        p("Loading")
```
bion, line 79: suggestion (non-blocking): use i18n message rather than hard-coded string

bion (approved):

server/app/views/admin/programs/ProgramIndexViewV2.java
bion: feel free to ignore

clouser2: Given that this is in the Admin UI where we have quite a few unlocalized messages, I'm going to leave as-is. Should we decide to localize the Admin UI, this would naturally fall out of that effort.

bion: I was thinking that it's a bit more generic, and if we're going to introduce a loading or spinner anywhere in the applicant UI it'd be good to have it translated already. Totally find to leave as-is though.

goldblatt (approved):
adding @shanemc-goog here since i know he's been thinking of feature flagging. this LGTM

clouser2:
> adding @shanemc-goog here since i know he's been thinking of feature flagging. this LGTM

Definitely agree that this might be a good candidate when the feature flag system exists. I think this might inform the need for the ability for a specific user to force feature flag state. The goal here is to make sure that one of the users can see the new UI.

clouser2:
> > adding @shanemc-goog here since i know he's been thinking of feature flagging. this LGTM
> 
> Definitely agree that this might be a good candidate when the feature flag system exists. I think this might inform the need for the ability for a specific user to force feature flag state. The goal here is to make sure that one of the users can see the new UI.

Spoke with @shanemc-goog offline and there's no need to block this change.

PR #2575, opened 2022-05-25
No longer record videos for prober runs.

clouser2:
### Description
Disabling this so we don't use too much storage for our GitHub actions. We only really need one CI run in order to determine what the source of prober failures is.


gwendolyngoetz (approved):

bion (approved):

bion:
not needed since we're removing the Seattle deploy from this repo https://github.com/seattle-uat/civiform/pull/2579/files

PR #2581, opened 2022-05-26
Force check and close of warning message toast

gwendolyngoetz:
Staging still has the issue where tests fail if the warming message toast is in the way. This will check for it and click close if it is there.

While these two test files only have a single test each I went with creating/close the session and clicking the button within the beforeEach/afterEach. New tests will automatically get it and I think it's a good pattern to eventually move into the other tests.

There might be a way to encapsulate it into a single function to call, but my first attempt didn't work and I didn't have time to try again.


browser-test/src/admin_api_keys.test.ts
```diff
@@ -1,13 +1,31 @@
...
+  let browser : Browser
+  let page : Page
+
+  beforeEach(async () => {
+    var session = await startSession()
+
+    browser = session.browser
+    page = session.page
+
+    await closeWarningMessage(page)
```
bion, line 21: suggestion (non-blocking): could call `closeWarningMessage` at the end of `startSession` so it happens for every test

bion (approved):

browser-test/src/admin_api_keys.test.ts
gwendolyngoetz: I feel like there was a reason I didn't do that when I first look at this a couple weeks back, but it's easy to test.

gwendolyngoetz: @bion Tests still passed after moving it so it's in the startSession now

clouser2 (approved):

PR #2584, opened 2022-05-26
Add support for Feature Flags.

### LINKED ISSUE #1943: CiviForm lacks a feature flag or experimentation system
When new code is introduced to CiviForm, there is no way to enable or disable it, only deploy or rollback.

We should implement release decoupling for CiviForm using either an open source or custom-made feature flag or experiments system.

- [x] Feature Flag
- [ ] Set feature flags via url (ease of testing in staging etc.)
- [ ] Experiment System

shanemc-goog:
### Description
Adds Injectable annotations and a Module provider that surfaces the system application configuration for features that should be on or off across the entire application.

### Checklist

### Issue(s)
Fixes #1943

### DIFF
--- a/server/app/modules/FeatureFlagsModule.java
+++ b/server/app/modules/FeatureFlagsModule.java
@@ -1,0 +1,21 @@
+package modules;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import annotations.FeatureFlags.StatusTrackingEnabled;
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import com.typesafe.config.Config;
+
+public class FeatureFlagsModule extends AbstractModule {
+  @Override
+  public void configure() {
+  }
+
+  @Provides
+  @StatusTrackingEnabled
+  public boolean provideStatusTrackingEnabled(Config config) {
+    return checkNotNull(config).getBoolean("status_tracking_enabled");
+  }
+
+}
--- a/server/app/modules/FeatureFlagsModule.java
+++ b/server/app/modules/FeatureFlagsModule.java
@@ -2,20 +2,20 @@ package modules;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import annotations.FeatureFlags.StatusTrackingEnabled;
+import annotations.FeatureFlags.ApplicationStatusTrackingEnabled;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.typesafe.config.Config;
 
+/** Provides configuration backed values that indicate if application wide features are enabled. */
 public class FeatureFlagsModule extends AbstractModule {
-  @Override
-  public void configure() {
-  }
-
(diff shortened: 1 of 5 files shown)


server/conf/application.conf
```diff
@@ -526,3 +527,7 @@ api_keys_ban_global_subnet = ${?CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET}
 # load balancer timeout.
 api_applications_list_max_page_size = 1000
 api_applications_list_max_page_size = ${?CIVIFORM_API_APPLICATIONS_LIST_MAX_PAGE_SIZE}
+
+# Feature flag toggles.
+status_tracking_enabled = true
```
goldblatt, line 532: maybe would default to false here

shanemc-goog: totally.  Just playing around atm.  Will fix up and send out formally.

server/app/modules/FeatureFlagsModule.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class FeatureFlagsModule extends AbstractModule {
+  @Override
+  public void configure() {
+  }
+
+  @Provides
+  @StatusTrackingEnabled
+  public boolean provideStatusTrackingEnabled(Config config) {
+    return checkNotNull(config).getBoolean("status_tracking_enabled");
```
clouser2, line 18: This works, but is there a need to add Guice annotations when the config is already injected at the callsite? For example, the flag could be a static String and the callsites could evaluate against the config.

One argument in favor of Guice annotations would be it'd be easier to mock out data in unit tests.

server/conf/application.conf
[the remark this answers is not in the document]
shanemc-goog: yeah, this is just a draft for the general idea.

server/app/modules/FeatureFlagsModule.java
shanemc-goog: 1. I think it's good to abstract these mechanics out, especially as we add more Feature Flags and ones that will be long lived.  The values will also be treated more "readably" through the code base as class ctors will generally have to accept the injected value and save it as a class field.
2. I hope we can support url overrides in dev/staging and having that handled in one place is good.
3. Not all involved classes will have the config injected also I think.
4. Testing does seem easier and the "concern" is treated as a first party thing as you mention

clouser2: Definitely seems like reasonable rationale. Another nice thing that would fall out of this is if we wanted to eventually allow admins / specific user types to "force" experiments. Having a single Guice-injection site that can wrap that decision up seems preferable to having each consumer of the feature flag have to apply that logic.

shanemc-goog: yeah, def having wide eyed views into that future :D

clouser2 (approved):

bion (approved):

PR #2595, opened 2022-05-27
first draft fix 2547

### LINKED ISSUE #2547: Unable to open docs for 1 application
**Describe the bug**
There is one application (#13665) where none of the program admins can open any documents--the documents are for Government ID and Income Verification, they are JPEGs and PDFs. The staff can view the application, but when then go open the uploaded documents, the page goes blank and the whole application goes away.

Staff mentioned that this has happened before and it's usually when a resident has previously submitted an application. They then search for the original application and can pull the documents. But this one looks to be the first application that the resident submitted.

**To Reproduce**
1. Search for application #13665
2. Click to download the documents in the application (gov ID and income)
3. See that page goes blank and application disappears

swatkat1:
### Description
This is to solve the issue caused by the bug #2547 (currently affecting 3 applications)
If the applicant reuses the file uploads across programs, the file doesn't download for other programs.

### Issue(s)
Fixes #2547

### DIFF
--- a/server/app/controllers/FileController.java
+++ b/server/app/controllers/FileController.java
@@ -1,18 +1,24 @@
 package controllers;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import auth.Authorizers;
 import auth.ProfileUtils;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
+
+import com.google.common.collect.ImmutableList;
+import models.Application;
 import org.pac4j.play.java.Secure;
 import play.libs.concurrent.HttpExecutionContext;
 import play.mvc.Http.Request;
 import play.mvc.Result;
+import repository.ApplicationRepository;
 import services.cloud.StorageClient;
 import services.program.ProgramDefinition;
 import services.program.ProgramNotFoundException;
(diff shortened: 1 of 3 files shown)


server/app/controllers/FileController.java
```diff
@@ -64,13 +71,38 @@ public CompletionStage<Result> show(Request request, long applicantId, String fi
             });
   }
 
+  private long extractApplicantIdFromFileKey(String fileKey) {
+    String applicantString = fileKey.split("/")[0];
+    long applicantid = (long) Long.parseLong(applicantString.split("-")[1].trim());
+    return applicantid;
+  }
+
```
bion, line 79: issue (blocking): the file key has the *applicant* ID embedded in it, not the application ID https://github.com/seattle-uat/civiform/blob/main/server/app/services/cloud/FileNameFormatter.java#L35

bion (changes requested):

server/app/controllers/FileController.java
```diff
@@ -64,13 +71,38 @@ public CompletionStage<Result> show(Request request, long applicantId, String fi
...
     try {
+      boolean isAuthorized = false;
+      if (fileKey != null) {
+        // if the applicant has submitted the same file to both programA and program-B
+        // then the admin has authorization to access the file in programA
+        long applicationId = extractApplicantIdFromFileKey(fileKey);
+        ImmutableList<Application> applicationList =
+            this.applicationRepository.getAllApplications().stream()
+                .filter(application -> application.id == applicationId)
+                .collect(toImmutableList());
```
bion, line 91: issue (blocking): the file key has the applicant ID embedded in it, not the application ID https://github.com/seattle-uat/civiform/blob/main/server/app/services/cloud/FileNameFormatter.java#L35

bion (changes requested):

server/app/controllers/FileController.java
bion: a possible longer-term fix for this could be to add a column to the files table that stores an array of program IDs that can access the file. when the applicant submits an application the server can scan the application for file upload questions and ensure the program has access.

Issue #1879, opened 2022-02-03
Question input fields missing `aria-describedby` attribute referencing the question description


bion:

syntapy:
I'd like to work on this one

bion:
@syntapy when we chatted on slack it sounded like you're working on other stuff, I'm going to pick this one up since it's related to work I'm already doing

jeanniefu:
Is this the expected behavior? Let's say we have a single question with multiple input fields, e.g. for addresses:

**Enter your address. (question text)**
_Put your primary residence (question help text)_
address: [ ] city: [ ]  state: [ ]  zip: [ ] 

My understanding is that we want to add the `aria-describedby` attribute to each input (e.g. address, city, state, zip), and that that attribute should point to the "Put your primary residence (question help text)" id.

Also when there are validation errors, my understanding is that we want to tag both the question help text ID and the validation error ID (related: https://github.com/seattle-uat/civiform/issues/1878).

clouser2:
That sounds consistent with my understanding. The `aria-describedBy` attribute can take one or multiple elements from what I can tell ([docs](https://www.w3.org/TR/wai-aria/#aria-describedby)).

jeanniefu:
I also chatted with an a11y analyst, who said using aria-describedby sounds good. They also recommended I look into fieldset and legend ([link](https://www.w3.org/WAI/tutorials/forms/grouping/)), but my impression is that the styling of fieldset is quite restrictive ([link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#styling_with_css)). I will go with the approach I outlined earlier: https://github.com/seattle-uat/civiform/issues/1879#issuecomment-1132372657.

jeanniefu:
Actually I will see if fieldset + legend works for these cases. It sounds like we should be able to override the border and positioning, and restyle it however we want: [link](https://stackoverflow.com/questions/29224061/correct-way-to-group-checkboxes-and-radio-buttons).

PR #2596, opened 2022-05-27
Add aria-describedby for question description and errors for input fields

jeanniefu:
### Description
Adds question level `aria-describedby` attributes (e.g. question description and question-level errors) to each input of that question.

Previously, the screen reader would only read out field-level errors but not question-level errors (See https://github.com/seattle-uat/civiform/issues/1878#issuecomment-1136572406).

Note that this PR does not address radio buttons, checkboxes, and misses some corner cases for enumerators. I will follow up with separate fixes.

### Checklist

### Issue(s)
#1878, #1879

### DIFF
--- a/server/app/views/components/FieldWithLabel.java
+++ b/server/app/views/components/FieldWithLabel.java
@@ -53,7 +53,7 @@ public class FieldWithLabel {
   protected String labelText = "";
   protected String placeholderText = "";
   protected String screenReaderText = "";
-  protected ArrayList<String> ariaDescribedByIds;
+  protected ArrayList<String> ariaDescribedByIds = new ArrayList<String>();
   protected Messages messages;
   protected ImmutableSet<ValidationError> fieldErrors = ImmutableSet.of();
   protected boolean showFieldErrors = true;
@@ -239,7 +239,7 @@ public class FieldWithLabel {
    * Set the list of HTML tag IDs that should be used for a11y descriptions.
    */
   public FieldWithLabel setAriaDescribedByIds(ArrayList<String> ariaDescribedByIds) {
-    this.ariaDescribedByIds = ariaDescribedByIds;
+    this.ariaDescribedByIds = new ArrayList<String>(ariaDescribedByIds);
     return this;
   }
 
--- a/server/app/views/components/FieldWithLabel.java
+++ b/server/app/views/components/FieldWithLabel.java
@@ -24,7 +24,9 @@ import views.style.BaseStyles;
 import views.style.StyleUtils;
 import views.style.Styles;
 
-/** Utility class for rendering an input field with an optional label. */
+/**
+ * Utility class for rendering an input field with an optional label.
+ */
 public class FieldWithLabel {
 
   private static final ImmutableSet<String> STRING_TYPES =
@@ -35,7 +37,9 @@ public class FieldWithLabel {
   protected String fieldType = "text";
   protected String fieldValue = "";
 
-  /** For use with fields of type `number`. */
+  /**
+   * For use with fields of type `number`.
+   */
   protected OptionalLong fieldValueNumber = OptionalLong.empty();
 
   protected OptionalLong minValue = OptionalLong.empty();
@@ -46,6 +50,7 @@ public class FieldWithLabel {
   protected String labelText = "";
   protected String placeholderText = "";
   protected String screenReaderText = "";
+  protected String descriptionId = "";
   protected Messages messages;
   protected ImmutableSet<ValidationError> fieldErrors = ImmutableSet.of();
   protected boolean showFieldErrors = true;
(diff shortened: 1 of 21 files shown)


server/app/views/components/FieldWithLabel.java
```diff
@@ -221,6 +227,18 @@ public FieldWithLabel setScreenReaderText(String screenReaderText) {
...
   }
 
+  /** Set the list of HTML tag IDs that should be used for a11y descriptions. */
+  public FieldWithLabel setAriaDescribedByIds(List<String> ariaDescribedByIds) {
+    this.ariaDescribedByIds = new ArrayList<>(ariaDescribedByIds);
+    return this;
+  }
+
+  /** Set whether the question this field belongs to has errors. */
+  public FieldWithLabel setHasQuestionErrors(boolean hasQuestionErrors) {
```
clouser2, line 237: FieldWithLabel is used for rendering fields outside of the context of an application (e.g. within the Admin UI when defining question / program types). Putting questions-specific logic feels like it might be introducing a bit too much coupling with the applicant experience.

What about something like `forceIsInvalid()` to ensure that `aria-invalid` is set even if there aren't any field-specific validation errors?

server/app/views/components/FieldWithLabel.java
```diff
@@ -283,10 +301,16 @@ public ContainerTag getContainer() {
...
+    fieldTag.condAttr(hasFieldErrors || hasQuestionErrors, "aria-invalid", "true");
+
+    var allAriaDescribedByIds = new ArrayList<String>(ariaDescribedByIds);
     if (hasFieldErrors) {
-      fieldTag.attr("aria-invalid", "true");
-      fieldTag.attr("aria-describedBy", fieldErrorsId);
+      allAriaDescribedByIds.add(0, fieldErrorsId);
     }
+    fieldTag.condAttr(
+        !allAriaDescribedByIds.isEmpty(),
```
clouser2, line 311: Is a blank value acceptable for `aria-describedBy`? If so, we could unconditionally set this.

clouser2 (changes requested):
Love the direction overall! Had a few questions around whether when we want to associate top-level question errors with specific fields.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -71,27 +89,36 @@ public final Tag render(ApplicantQuestionRendererParams params) {
...
+      ariaDescribedByIds.add(0, requiredErrorId);
     }
 
     return div()
-        .withId(question.getContextualizedPath().toString())
+        .withId(id)
         .withClasses(Styles.MX_AUTO, Styles.MB_8, getReferenceClass(), getRequiredClass())
         .with(questionTextDiv)
-        .with(renderTag(params, validationErrors));
+        .with(renderTag(params, validationErrors, ariaDescribedByIds, hasQuestionErrors));
```
bion, line 122: suggestion (blocking): it's generally a good practice to avoid passing mutable collections around when avoidable. in this class you could make `ariaDescribedByIds` an `ImmutableList.Builder<String>` and change this line to
```suggestion
        .with(renderTag(params, validationErrors, ariaDescribedByIds.build(), hasQuestionErrors));
```

server/app/views/questiontypes/EnumeratorQuestionRenderer.java
```diff
@@ -83,6 +90,11 @@ protected Tag renderTag(
                             messages.at(
                                 MessageKey.ENUMERATOR_BUTTON_ADD_ENTITY.getKeyName(),
                                 localizedEntityType)))
+                    .condAttr(hasQuestionErrors, "aria-invalid", "true")
+                    .condAttr(
+                        !ariaDescribedByIds.isEmpty(),
+                        "aria-describedby",
```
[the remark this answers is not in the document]
bion: doesn't look like it https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

bion (changes requested):

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -71,27 +89,36 @@ public final Tag render(ApplicantQuestionRendererParams params) {
             String.format("Unhandled error display mode: %s", params.errorDisplayMode()));
     }
 
+    boolean hasQuestionErrors = false;
     ImmutableSet<ValidationErrorMessage> questionErrors =
         validationErrors.getOrDefault(question.getContextualizedPath(), ImmutableSet.of());
     if (!questionErrors.isEmpty()) {
       // Question error text
+      hasQuestionErrors = true;
```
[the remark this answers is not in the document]
jeanniefu: Good points! I checked with the a11y team, and they said we should NOT associate the question level errors and descriptions with the sub fields. 

They also recommend we group the composite fields (like address, name) using fieldset and legend, and associate the question level errors/descriptions with the fieldset. Related issue: https://github.com/seattle-uat/civiform/issues/2618. Also, we may want to change the way we handle required fields and errors based on a11y team's feedback (https://github.com/seattle-uat/civiform/issues/2629), which might also change my approach here.

I think I should change this PR to just affect the non-composite fields. Will comment again here once it's ready.

jeanniefu:
This was redone and submitted as #3354

PR #2598, opened 2022-05-31
Setup script for aws

azizova-leyla:
### Description

### Checklist

### Issue(s)


goldblatt (changes requested):
oo this does seem like maybe we should make this a base class and both of the setups should inherit? i think i probably didn't do that originally but did intend to?

azizova-leyla:
> oo this does seem like maybe we should make this a base class and both of the setups should inherit? i think i probably didn't do that originally but did intend to?

done, I think? let me know if this is what you had in mind.

goldblatt (approved):

PR #2603, opened 2022-05-31
Only format java and TS files that have changed

### LINKED ISSUE #2483: bin/fmt takes quite a long time
Presently, `bin/fmt` takes >1 minute to run. We should investigate ways to improve this. One possibility would be to only run on a subset of files (e.g. those with edits).

bion:
Attempts to speed up `bin/fmt` by reducing the number of files the autoformatters operate on. For java it uses `git diff --name-only main` to get the tracked files that have changes and  `git ls-files --others --exclude-standard` to get the untracked files that have changes. It still runs the shell formatter for all shell files in the system since it's harder to determine which files should be formatted as shell code, and it runs prettier every time since it's pretty fast to run even without filtering the input.

fixes #2483


bin/fmt
```diff
@@ -5,8 +5,13 @@
 source bin/lib.sh
 docker::set_project_name_dev
 
-bin/pull-image
+if [[ -z "${USE_LOCAL_CIVIFORM}" ]]; then
```
kmcnellis, line 8: I don't love pulling this out from the main pull-image.  Perhaps the bin/pull-image script can take an optional argument of the image name (and pull all if not provided)

bin/fmt
```diff
@@ -5,8 +5,13 @@
...
+  docker pull -q civiform/formatter:latest
+else
+  echo "Using local civiform and dependency docker images"
+fi
 
 docker run --rm -it \
   -v "$(pwd):/code" \
+  -u $(id -u ${USER}):$(id -g ${USER}) \
```
kmcnellis, line 16: What is this for?

formatter/fmt
```diff
@@ -0,0 +1,58 @@
...
+
+set -e
+set +x
+set -o pipefail
+
+cd /code
+
+# Gets the list of file paths for files that are tracked by git and have
+# changes relative to the main branch.
+function get_tracked_modified_files() {
```
kmcnellis, line 11: Make this optional - when running in a PR, it should check all files

formatter/fmt
```diff
@@ -0,0 +1,58 @@
...
+    get_tracked_modified_files
+    get_untracked_files
+  ) | grep -E "server\/.+\.java"
+}
+
+function get_modified_ts_files() {
+  (
+    get_tracked_modified_files
+    get_untracked_files
+  ) | grep -E ".+\.ts"
```
kmcnellis, line 32: Prettier modifies more than just .ts files (including json, js, yaml, toml, ts, html and more: https://prettier.io/docs/en/)

Also, I _ think _ prettier is pretty quick.  We might get 90% of the speedup just from java.

kmcnellis (changes requested):

formatter/fmt
bion: why?

bin/fmt
bion: Helps avoid file ownership problems when running the scripts by running the docker process as the same user as runs the `bin/fmt` script

bin/fmt
bion: done

bin/pull-image
```diff
@@ -1,13 +1,20 @@
 #! /usr/bin/env bash
 
-# DOC: Pull the latest development environment docker image from Docker Hub unless USE_LOCAL_CIVIFORM is set.
+# DOC: Pull the latest development environment docker image from Docker Hub
+# DOC: unless USE_LOCAL_CIVIFORM is set. May specify a specific what to pull
```
shanemc-goog, line 4: "specific what" reads weird.

also include "single" since you can't specify 2 things.

formatter/fmt
```diff
@@ -0,0 +1,47 @@
...
+# not excluded by the .gitignore file.
+function get_untracked_files() {
+  git ls-files --others --exclude-standard
+}
+
+function get_modified_java_files() {
+  (
+    get_tracked_modified_files
+    get_untracked_files
+  ) | grep -E "server\/.+\.java"
```
shanemc-goog, line 25: Probably want a ^ at the start

do we need server/ though?

formatter/fmt
```diff
@@ -0,0 +1,47 @@
...
+set -e
+set +x
+set -o pipefail
+
+cd /code
+
+# Gets the list of file paths for files that are tracked by git and have
+# changes relative to the main branch.
+function get_tracked_modified_files() {
+  git diff --name-only main
```
shanemc-goog, line 12: This lists what I assume is all the files for me... Could this be something that won't work for everyones setup?

bin/pull-image
bion: wups, typo. fixed

formatter/fmt
bion: oh hmm, what's your local git state? are you on main? a branch?

formatter/fmt
kmcnellis: For one, its faster than the other tests, so not a big deal to run it on every file.

For the other, If there's an a change made by the formatter in a PR test, then it will error.  So by checking all the files, we make sure that there's no errors that have snuck in. And if we add a new formatting option, we'd error out the test and know that our formatting was not correct with some file.

But I guess it's preference-  perhaps doing it this way and getting only updating the formatting of files when we change them for some other reason is OK too.

kmcnellis (approved):

formatter/fmt
shanemc-goog: Oh, I was on a renovate bot branch.. I made a new branch and it is WAI showing no files.

bin/pull-image
```diff
@@ -1,13 +1,20 @@
 #! /usr/bin/env bash
 
-# DOC: Pull the latest development environment docker image from Docker Hub unless USE_LOCAL_CIVIFORM is set.
+# DOC: Pull the latest development environment docker image from Docker Hub
+# DOC: unless USE_LOCAL_CIVIFORM is set. May specify what to pull with flags
```
shanemc-goog, line 4: You can only specify one right?  Include that here.

shanemc-goog (approved):

bin/pull-image
bion: Nope can specify multiple

bion: ohwait, crap yeah only one. thanks

formatter/fmt
bion: ah i think we actually do test all of them since we don't use bin/fmt in CI but rather https://github.com/seattle-uat/civiform/blob/faster-autoformatter/.github/workflows/format.yaml#L23-L26

PR #2605, opened 2022-05-31
Run CI browser tests in parallel batches

bion:
Runs the CI browser tests in parallel batches.

fixes https://github.com/seattle-uat/civiform/issues/1936

### DIFF
--- a/bin/run-browser-tests-ci-batch
+++ b/bin/run-browser-tests-ci-batch
@@ -10,10 +10,20 @@ import sys
 
 BATCH_COUNT = 4
 
-batch_number = sys.argv[1]
+def bad_batch_number_input(batch_num_input):
+    print(f"Invalid batch number: {batch_num_input}", file=sys.stderr)
+    exit(1)
+
+if len(sys.argv) < 2:
+    bad_batch_number_input("none")
+
+try:
+    batch_number = int(sys.argv[1])
+except ValueError:
+    bad_batch_number_input(sys.argv[1])
 
 if batch_number < 1 or batch_number > BATCH_COUNT:
-    print(f"Invalid batch number: {batch_number}")
+    bad_batch_number_input(batch_number)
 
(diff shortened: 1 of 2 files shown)


bin/run-browser-tests-ci-batch
```diff
@@ -0,0 +1,42 @@
...
+except ValueError:
+    bad_batch_number_input(sys.argv[1])
+
+if batch_number < 1 or batch_number > BATCH_COUNT:
+    bad_batch_number_input(batch_number)
+
+test_files = glob.glob('browser-test/src/*.test.ts')
+test_files = [file.replace("browser-test/", "") for file in test_files]
+test_files.sort()
+test_files_count = len(glob.glob('browser-test/src/*.test.ts'))
```
jeanniefu, line 31: Just curious, any reason to do this instead of just length of test_files?

bin/run-browser-tests-ci-batch
```diff
@@ -0,0 +1,42 @@
...
+batch_size = math.ceil(test_files_count / BATCH_COUNT)
+
+batch_contents = test_files[(batch_number - 1) * batch_size : batch_number * batch_size]
```
jeanniefu, line 34: If (batch_number * batch_size) is greater than the list size, does this error?

bin/run-browser-tests-ci-batch
bion: nope, that was a mistake. thanks!

bin/run-browser-tests-ci-batch
bion: nope, python will silently take the minimum of the provided value or the length of the list

bion:
```python
>>> [1, 2, 3][1:2000]
[2, 3]
```

jeanniefu: Cool, thanks!

swatkat1 (approved):

PR #2611, opened 2022-06-01
Fix bin/pull-image for macOS

bion:
macOS is on an older version of bash which does not support the `-v` flag in conditionals. This change swaps using `-v` for `! -z` in the `bin/pull-image` flag for compatibility.

### DIFF
--- a/bin/pull-image
+++ b/bin/pull-image
@@ -7,7 +7,7 @@
 
 source bin/lib.sh
 
-if [[ ! -z "${USE_LOCAL_CIVIFORM}" ]]; then
+if [[ -n "${USE_LOCAL_CIVIFORM}" ]]; then
   echo "Using local civiform and dependency docker images"
   exit 0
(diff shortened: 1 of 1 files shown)


bin/pull-image
```diff
@@ -7,7 +7,7 @@
 
 source bin/lib.sh
 
-if [[ -v USE_LOCAL_CIVIFORM ]]; then
+if [[ ! -z "${USE_LOCAL_CIVIFORM}" ]]; then
```
gwendolyngoetz, line 10: (blocking) Please change `! -z` to `-n`

clouser2 (approved):

gwendolyngoetz (changes requested):

bin/pull-image
bion: done

gwendolyngoetz (approved):

PR #2612, opened 2022-06-01
Enable gzip compression

syntapy:
### Description
Enables gzip compression as described [here](https://www.playframework.com/documentation/2.8.x/GzipEncoding)

Reduces CSS transfer size from 1,726.8 KiB down to 174.8 KiB, according to two lighthouse reports

Full lighthouse reports are
- [with no gzip](https://github.com/seattle-uat/civiform/files/8836291/with_no_gzip.pdf)
- [with gzip](https://github.com/seattle-uat/civiform/files/8836295/with_gzip.pdf)

### Checklist
- [x] Make sure its compressed
- [x] Check what a stack error trace looks like

### Issue(s)

Addresses https://github.com/seattle-uat/civiform/issues/1940


bion (changes requested):

server/conf/application.conf
```diff
@@ -307,6 +307,7 @@ play.filters {
   enabled += filters.HSTSFilter
   enabled += filters.LoggingFilter
   enabled += filters.ValidAccountFilter
+  enabled += filters.gzip.GzipFilter
```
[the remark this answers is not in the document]
clouser2: Out of curiosity, why only in prod mode?

When viewing the response in devtools, the response will be un-zipped into its full form.

server/conf/application.conf
```diff
@@ -307,6 +307,7 @@ play.filters {
   enabled += filters.HSTSFilter
   enabled += filters.LoggingFilter
   enabled += filters.ValidAccountFilter
```
clouser2, line 309: For posterity, it'd be great to include in the PR a before / after CSS size as well as a pointer to the docs for enabling this filter. That'll help future readers understand where we pulled this from.

clouser2 (approved):

server/conf/application.conf
clouser2: Disregard my request for a doc pointer since it was in fact already included in the PR details (whoops!).

Before / after CSS size would still be valuable in my opinion.

server/conf/application.conf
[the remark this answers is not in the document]
bion: just to reduce complexity and stack trace size when applicable. probably shouldn't be blocking, changed the comment

bion (approved):

syntapy:
I'm getting an error, need to work that out. Also should probably set the highest compression level

kmcnellis (approved):

kmcnellis:
>

kmcnellis:
(sorry, that was a mistake)

syntapy:
I need to try it out in lighthouse still or something

syntapy:
Realized I had gzip disabled in dev. After enabling it in dev I still get the same thing in the report

Here are full reports

[disabled_in_dev.pdf](https://github.com/seattle-uat/civiform/files/8819596/disabled_in_dev.pdf)
[enabled_in_dev.pdf](https://github.com/seattle-uat/civiform/files/8819600/enabled_in_dev.pdf)

server/conf/application.conf
```diff
@@ -308,6 +308,28 @@ play.filters {
...
 
+  # GZip filter configuration
+  gzip {
```
clouser2, line 312: I'd consider removing these settings (as they're presumably pretty similar to the default behavior we'd get out of the box). In my opinion, the main savings will come from having any gzip at all. If we find the defaults to be unacceptable, we can tweak them at that point.

server/conf/application.conf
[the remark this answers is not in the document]
clouser2: I'm not sure I understand how disabling gzip in dev would reduce complexity. It seems that it would be something we'd have to remember is different between the two. From the browser and server perspective, performing gzip would be more of a network-layer concern and is largely transparent on either side (e.g. even though the content is gzipped, the dev tools still render the un-gzipped content).

Personally, I'd prefer to keep things consistent between the two stacks. Having this enabled for dev also means we'll get more eyes on it should something go wrong. If we disable in dev, it's possible that the issues will only occur in deployments, which seems unfortunate.

clouser2 (approved):

server/conf/application.conf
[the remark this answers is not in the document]
kmcnellis: +1. I don't think that enabling in dev with have any real impact on complexity.  And it's always nice to keep it as similar to prod as possible

server/conf/application.conf
kmcnellis: Yes, please remove these

server/conf/application.conf
```diff
@@ -308,6 +308,28 @@ play.filters {
...
+    #
+    # In general, it is recommended to turn off the buffer and prevent generation of overlong chunks at the source.
+    bufferSize = 8k
+
+    # The maximum amount of content to buffer for gzipping in order to calculate the content length before falling back
+    # to chunked encoding.
+    chunkedThreshold = 100k
+
+    # The compression level to use, integer, -1 to 9, inclusive. See java.util.zip.Deflater.
+    compressionLevel = -1
```
kmcnellis, line 324: Keep this one, with a comment:
```
    compressionLevel = -1     # -1 is the default compression (level 6)
```

server/conf/application.conf
```diff
@@ -308,6 +308,28 @@ play.filters {
...
+
+    # The byte threshold for the response body size which controls if a response should be gzipped (e.g. 1k).
+    # If the body size cannot be determined, then it is assumed the response is over the threshold.
+    # Set to 0 if you want to compress all responses, no matter how large the response body size is.
+    threshold = 0
```
kmcnellis, line 329: Set this to a low number (500 or 1000 is probably good).  There's no need to spend the extra effort compressing small files that don't get much of a size gain.

```
    threshold = 1000     # Don't compress very small responses.
```

server/conf/application.dev.conf
```diff
@@ -19,4 +19,6 @@ play.filters {
     cookie.secure = false
   }
   disabled += filters.HSTSFilter
+
+  gzip: null
```
kmcnellis, line 23: Change this to `disabled += "play.filters.gzip.GzipFilter"`

kmcnellis (changes requested):

server/conf/application.dev.conf
bion: @clouser2 makes a good argument for leaving gzip on in dev, I rescind my request for this

kmcnellis:
With default compression (level 6)

With gzip, tailwinds is 173kb bytes
Without, it is 1.8 Mb

So it's around a 10x reduction in size.  Pretty good!

On compression level 9, it's 171kb - probably not worth the extra processing time (though it is minimal)

server/conf/application.conf
```diff
@@ -308,6 +308,28 @@ play.filters {
   enabled += filters.LoggingFilter
   enabled += filters.ValidAccountFilter
 
+  # GZip filter configuration
+  gzip {
+
+    # The maximum amount of data to send to the Gzip compressor in one go. Use `0` or `2147483647` to disable the buffering.
+    #
+    # In general, it is recommended to turn off the buffer and prevent generation of overlong chunks at the source.
+    bufferSize = 8k
```
kmcnellis, line 317: Please add an allow-list for file types.  Image formats (like png) are already compressed, and shouldn't be double-compressed (they can actually increase in size, and rarely decrease).

```
gzip {

    contentType {
        # If non empty, then a response will only be compressed if its content type is in this list.
        whiteList = [ "text/*", "application/javascript", "application/json" ]
    }
}
```

kmcnellis (changes requested):

server/conf/application.conf
clouser2: Good call! I'm actually surprised there wasn't a default allowlist along these lines in the framework!

server/conf/application.dev.conf
kmcnellis: SG

kmcnellis, line 23: Remove this.  It's not needed

kmcnellis (changes requested):

server/conf/application.conf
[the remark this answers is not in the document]
syntapy: I have re-enabled it in dev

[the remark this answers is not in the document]
syntapy: I added [this](https://github.com/seattle-uat/civiform/pull/2612/commits/3fc762ab824da8bc552909374c8dfa60cc6e9f5c) and got this trace

flat_earth_exception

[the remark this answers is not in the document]
syntapy: WIthout gzip that stack trace looks like

flat_earth_exception_nogzip

syntapy:
I finally have a report on how gzip effected transfer size and all. Please see the updated description

server/conf/application.conf
```diff
@@ -307,6 +307,16 @@ play.filters {
...
   enabled += filters.LoggingFilter
   enabled += filters.ValidAccountFilter
+  enabled += play.filters.gzip.GzipFilter
+
+  gzip {
+    threshold = 1000        # don't compress very small responses
+    compressionLevel = -1   # -1 is the default compression level (level 6)
+    contentType {
+      # If non empty, then a response will only be compressed if its content type is in this list.
+      blackList = [ "app/assets/Images/"]
```
clouser2, line 317: Blocking: I don't know that this line does what we want. The docs claim that this should be a MIME Type.

I *believe* you could do this instead:

```
blackList = ["image/*"]
```

Alternatively, we could only gzip well-known types (as in the example in the [docs](https://www.playframework.com/documentation/2.8.x/GzipEncoding#Controlling-which-responses-are-gzipped)). That'd mean we could remove the "blackList" entry and instead add:

```
whiteList = [ "text/*", "application/javascript", "application/json" ]
```

Either way has tradeoffs, so whatever you'd prefer.

clouser2 (approved):

server/conf/application.conf
kmcnellis: +1 for only gzipping well-known types

kmcnellis (approved):

server/conf/application.conf
syntapy: Oh shoot, @bion I didn't get to do these yet

kmcnellis: made https://github.com/seattle-uat/civiform/pull/2667

PR #2616, opened 2022-06-01
Add barebones deploy scripts to make it work

azizova-leyla:
Issues: https://github.com/seattle-uat/civiform/issues/2553


cloud/aws/bin/deploy
```diff
@@ -0,0 +1,24 @@
+#! /usr/bin/env bash
+set -e
+
+source "cloud/shared/bin/lib.sh"
+source "cloud/azure/bin/lib.sh"
```
kmcnellis, line 5: Should this be importing anything from azure?

kmcnellis (changes requested):

cloud/aws/bin/deploy
azizova-leyla: good catch. we don't need this.

bion (approved):

cloud/aws/bin/deploy
```diff
@@ -0,0 +1,23 @@
+#! /usr/bin/env bash
+set -e
```
goldblatt, line 2: we could possibly make these scripts python based?

goldblatt (approved):
could possibly make this python based but otherwise seems alright

kmcnellis (approved):

Issue #2589, opened 2022-05-26
De-duplicate logic for file upload views


clouser2:
Presently, we have Azure and AWS-based file upload strategies that contain a fair amount of duplication for rendering the applicant-facing UI. This duplication has led to bugs where functionality is only implemented in one strategy but not the other (#2487). We can likely consolidate a fair amount of the logic and tie it in with our other rendering strategy used for other question types.

kristinyim:
Going to be addressed in fix-it week.

clouser2:
Work here is complete!

PR #2634, opened 2022-06-02
De-duplicate file upload view rendering logic

clouser2:
### Description
Pares down `AwsFileUploadViewStrategy` / `AzureFileUploadViewStrategy` to only implementation-specific details and moves duplicated logic into the base `FileUploadViewStrategy` implementation.

Why? Due to the duplication, it was common for the user-facing behavior to diverge, leading to bugs (#2487 for example).

Notes:
* The `azure_delete` script is included unconditionally for file uploads. This was previously only included if the delete button was rendered. The network savings don't seem worth the extra code complexity
* `azure_upload` can reference the container for details rather than an unnecessary wrapping div ID

Follow-on PRs will:
* Potentially remove the divergent `isSample` plumbing used for admin previewing of the field. We can likely make this transparent without the need to add unnecessary plumbing
* Make file upload work for AWS in `bin/run-dev` (likely just a matter of making ports visible to the localstack container)
* Look into removing `CloudStorageViewStrategy` for dev-only endpoints related to file upload. Assuming we can make typical file upload work, this seems like unnecessary logic / divergent behavior that might not be worth implementing.

### Issue(s)
#2589


bion:
Screenshot from 2022-06-02 10-14-14
🤩

azizova-leyla (approved):

server/app/views/FileUploadViewStrategy.java
```diff
@@ -44,18 +52,89 @@ public abstract class FileUploadViewStrategy extends ApplicationBaseView {
...
+    Tag skipForms = renderDeleteAndContinueFileUploadForms(params);
+    Tag buttons = renderFileUploadBottomNavButtons(params);
+
+    return div(uploadForm, skipForms, buttons).with(each(extraScriptTags(), tag -> footer(tag)));
+  }
+
+  protected ContainerTag renderFileUploadFormElement(Params params, StorageUploadRequest request) {
+    return form()
+        .withId(BLOCK_FORM_ID)
+        .attr(ENCTYPE, "multipart/form-data")
```
bion, line 131: suggestion (blocking):
```suggestion
        .attr(ENCTYPE, MimeTypes.MULTIPART_FORM_DATA)
```

`MimeTypes` is in `org.eclipse.jetty.http`

bion (changes requested):
Overall looks good! Did you view the rendered page using both strategies?

server/app/views/FileUploadViewStrategy.java
```diff
@@ -44,18 +52,89 @@ public abstract class FileUploadViewStrategy extends ApplicationBaseView {
    * @param fileUploadQuestion The question that requires a file upload.
    * @return a container tag with the necessary fields
    */
-  public abstract ContainerTag signedFileUploadFields(
-      ApplicantQuestionRendererParams params, FileUploadQuestion fileUploadQuestion);
+  public final ContainerTag signedFileUploadFields(
+      ApplicantQuestionRendererParams params, FileUploadQuestion fileUploadQuestion) {
+    Optional<String> uploaded =
+        fileUploadQuestion.getFilename().map(f -> String.format("File uploaded: %s", f));
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/views/FileUploadViewStrategy.java
clouser2: This would involve bringing in a new build dependency in build.sbt. Just wanted to confirm that it still seems worth it from your perspective.

clouser2:
> Overall looks good! Did you view the rendered page using both strategies?

Yes. Additionally, my hope and dream is that browser test coverage is sufficient to ensure there were no breaking change.

bion (approved):

PR #2654, opened 2022-06-03
Fix form field ordering issue that breaks AWS S3 uploads.

### LINKED ISSUE #2653: AWS file uploads don't work on staging
**Describe the bug**
Attempting to upload a file on staging fails. This is fallout from the refatoring done in #2589, specifically #2634.

**To Reproduce**
1. Navigate to staging.seattle.civiform.com and apply to the "test program for single file upload"
2. Choose a file to upload and click upload
3. An error is shown:
Screen Shot 2022-06-03 at 4 04 13 PM

**Expected behavior**
File upload succeeds

clouser2:
### Description
We now move the input field generation into each specific provider and document that order is input for S3 (as well as the field name of "file").

### Issue(s)
Fixes #2653; #2589

### DIFF
--- a/server/app/views/AwsFileUploadViewStrategy.java
+++ b/server/app/views/AwsFileUploadViewStrategy.java
@@ -3,16 +3,21 @@ package views;
 import static j2html.TagCreator.input;
 
 import com.google.common.collect.ImmutableList;
+import j2html.attributes.Attr;
 import j2html.tags.ContainerTag;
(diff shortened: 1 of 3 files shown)


bion (approved):

jeanniefu (approved):

clouser2:
To be clear, this regression made me very sad :/.

PR #2630, opened 2022-06-02
Replace Styles.java calls with tailwind style literals

syntapy:
### Description

Reduces css transfer size from 169K to 21K according to lighthouse report, and on average over 7 lighthouse runs it reduces time to interactive from 4.7s to 3.1s

This was an idea @clouser2 came up with and is a much simpler solution compared to this [monster pr](https://github.com/seattle-uat/civiform/pull/1951)

It uses a `replace.js` (in the commit history of this PR) to replace calls to `Styles.java` with tailwind style literals. 

Though there is one call to `Styles.INVISIBLE` that the script misses, and so I manually replaced that with `"invisible"` in one commit of this PR

I also used following sed command to remove imports to `Styles.java` (again, a separate commit in this PR)

```find ./ -name '*.java' -exec sed -i 's/import views.style.Styles;//' {} \;```
#### Test process I used to make sure tailwind automatically rebuilds on code changes
- Find a bg color style thats not used anywhere in the program
- I used `"bg-red-600"` for that
- Changed `LoginForm.java` `bg-gray-100` to that style
- Refresh the page
- Make sure that bottom horizontal stripe on home page is red not gray

## Release notes:

Use tailwind style literals instead of storing style classes in a java class. This reduces the tailwind.css file by an order of magnitude without having to parse java code for style calls

### Checklist
- [ ] Fix broken styles
  - [x] Invisible border on some buttons (e.g. after submitting application)
  - [x] Gray focus border around radio buttons
    - Cases where should be blue or red
  - [x] Review button text is flush left
    - Should be centered 
- [x] Automatically rebuild styles on recompile or maybe use a tailwind watch command. Check that it works on
  - [x] `run-dev`
  - [x] `run-dev --azure`
  - [x] `run-browser-test-env`
  - [x] `run-browser-test-env --azure`
  - [x] any two of the above after restarting the containers

### Issue(s)
Fixes https://github.com/seattle-uat/civiform/issues/1940

PR #2660, opened 2022-06-06
Adding administrative-use-only text when editing screen name.

### LINKED ISSUE #1901: Add more information about what is applicant-facing vs. internal
In the add screens step, it is unclear whether and what information is shown on the applicant-facing side of things or whether it's internal. Suggestion is to add more information about whether things are applicant-facing or internal.

This could be relevant in other areas as well, the add screens step is just what was noted.

clouser2:
See proposal in https://github.com/seattle-uat/civiform/issues/1901#issuecomment-1136479248

Before:
Screen Shot 2022-06-06 at 9 10 59 AM

After:
Screen Shot 2022-06-06 at 9 11 09 AM

### Issue(s)
Fixes #1901


server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -547,6 +548,8 @@ private Modal blockDescriptionModal(Tag csrfTag, BlockForm blockForm, String blo
         .withId("block-edit-form")
         .with(
             div(
+                    h3("The following fields will only be visible to administrators")
+                        .withClass("mb-2"),
```
clouser2, line 552: Note: Using the "mb-2" directly rather than `Styles.MB_2` to not cause merge conflicts with #2630, where we'll be moving to the "mb-2" pattern anyway.

server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -547,6 +548,8 @@ private Modal blockDescriptionModal(Tag csrfTag, BlockForm blockForm, String blo
         .withId("block-edit-form")
         .with(
             div(
+                    h3("The following fields will only be visible to administrators")
```
kmcnellis, line 551: Did you consider changing it from "screen name" to something with a little less assumptions associated with it?  Like "Label" or "Nickname" ?

jeanniefu (approved):

server/app/views/admin/programs/ProgramBlockEditView.java
clouser2: @rahaghas was considering revamping the language around screens / blocks. We should likely do that as part of a more holistic change though. "Name / Description" are consistent with the terminology we use for programs and questions, so introducing separate language here would be a bit inconsistent. I'm definitely open to switching the language we use within the Admin UI, I just think we should do it as one atomic change. Filed #2668 for that.

kmcnellis: Thanks!

server/app/views/admin/programs/ProgramBlockEditView.java
shanemc-goog: I haven't looked at 2630, but if we need to come do cleanup after that CL then do a 

//TODO(Issue #2630): xxx

server/app/views/admin/programs/ProgramBlockEditView.java
shanemc-goog, line 551: Do we distinctly know what H level is above this?

a11y needs correct hierarchy.

Not that we have that as is, but being mindful here if we can do the right thing.

clouser2: Good point. I made it an h1 in this case since it's a modal dialog. Since it is contained within a modal, the main visible content to the screen reader would be this heading. As such, it should be h1. Used TW classes to decrease the size.

server/app/views/admin/programs/ProgramBlockEditView.java
clouser2: #2630 is the PR to inline Tailwind styles. The way my PR is written means there won't be any merge conflicts or need to do any cleanup after that PR. As such, I don't think there's a need for a TODO.

shanemc-goog (approved):

PR #2661, opened 2022-06-06
Open external links in new tab

### LINKED ISSUE #2128: Web links open in the same browser tab

gwendolyngoetz:
### Description
* Allows the link generator to make links open in the same tab or a new tab.
* Sets external links to open in new tab. 
* Adds an open in new tab icon for external links.
* Slightly increases the default font size of the static text component.

### Checklist

### Issue(s)
Fixes #2128

before-after


server/test/views/components/TextFormatterTest.java
```diff
@@ -25,11 +26,48 @@ public void urlsRenderCorrectly() {
...
+  public void urlsRenderToOpenInNewTabCorrectly() {
+    ImmutableList<DomContent> content =
+        TextFormatter.createLinksAndEscapeText(
+            "hello google.com http://internet.website", TextFormatter.UrlOpenAction.NewTab);
+
+    assertThat(content).hasSize(4);
+    assertThat(content.get(0).render()).isEqualTo(new Text("hello ").render());
+    assertThat(content.get(1).render())
+        .isEqualTo(
+            "<a href=\"http://google.com/\" class=\"text-seattle-blue\" target=\"_blank\">google.com<svg xmlns=\"http://www.w3.org/2000/svg\" fill=\"currentColor\" stroke=\"currentColor\" stroke-width=\"1%\" aria-hidden=\"true\" viewBox=\"0 0 24 24\" width=\"24\" height=\"24\" class=\"flex-shrink-0 h-5 w-auto inline ml-1 align-text-top\"><path d=\"M19 19H5V5H12V3H5C3.89 3 3 3.9 3 5V19C3 20.1 3.89 21 5 21H19C20.1 21 21 20.1 21 19V12H19V19ZM14 3V5H17.59L7.76 14.83L9.17 16.24L19 6.41V10H21V3H14Z\"></path></svg></a>");
```
clouser2, line 39: Up to you, but I'd consider a less strict assertion to avoid having to inline the svg content (which may change at some point).

Maybe we could assert that the content starts with everything through `<svg ` and then assert that the content ends with `</svg></a>`. I know we lose some fidelity in what's being asserted with that approach, so understand if you'd prefer to keep the full SVG content inlined.

Same for the assertion below.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -29,6 +29,7 @@
 import services.program.ProgramDefinition;
 import views.BaseHtmlView;
 import views.HtmlBundle;
+import views.components.Icons;
```
clouser2, line 32: Top-level comment: I'd consider adding before/after screenshots to the PR description. That'd make understanding the Tailwind style applications a bit easier.

clouser2 (approved):

server/test/views/components/TextFormatterTest.java
gwendolyngoetz: Good idea. Fixed

server/app/views/applicant/ProgramIndexView.java
gwendolyngoetz: Done :)

clouser2 (approved):

bion (approved):

PR #2667, opened 2022-06-06
Change gzip blacklist to whitelist using mime types

kmcnellis:
### Description
Use mimetypes to decide which files can be gzipped, instead of blacklisting the image asset directory

Leftover change from #2612 (https://github.com/seattle-uat/civiform/pull/2612#discussion_r889610985).

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -314,7 +314,7 @@ play.filters {
     compressionLevel = -1   # -1 is the default compression level (level 6)
     contentType {
       # If non empty, then a response will only be compressed if its content type is in this list.
-      blackList = [ "app/assets/Images/"] 
+      whiteList = [ "text/*", "application/javascript", "application/json" ]
     }
   }
 


syntapy:
Ok, a whiteList is definitely more elegantly exclusive than a blackList. But if I remember correctly, when I ran a lighthouse test with this exact snippet as per [comment])(https://github.com/seattle-uat/civiform/pull/2612#discussion_r888362901) it didn't compress. Right now maven is taking a while to download after switching so it will be a bit before I can try again

Reading the [doc](https://www.playframework.com/documentation/2.8.x/GzipEncoding#Controlling-which-responses-are-gzipped) it isn't exactly clear to me what `text/*` means, i.e. if thats a type indication or a path indication

syntapy:
Oh nevermind I just tried it, it worked!

clouser2 (approved):

syntapy:
> But if I remember correctly, when I ran a lighthouse test with this exact snippet as per https://github.com/seattle-uat/civiform/pull/2612#discussion_r888362901 it didn't compress. Right now maven is taking a while to download after switching so it will be a bit before I can try again
>
> Reading the [doc](https://www.playframework.com/documentation/2.8.x/GzipEncoding#Controlling-which-responses-are-gzipped) it isn't exactly clear to me what text/* means, i.e. if thats a type indication or a path indication

@kmcnellis Well that was kind of rude of me. I shouldn't have assumed you wouldn't check it first. Sorry! I feel bad now

kmcnellis:
> > But if I remember correctly, when I ran a lighthouse test with this exact snippet as per [#2612 (comment)](https://github.com/seattle-uat/civiform/pull/2612#discussion_r888362901) it didn't compress. Right now maven is taking a while to download after switching so it will be a bit before I can try again
> > Reading the [doc](https://www.playframework.com/documentation/2.8.x/GzipEncoding#Controlling-which-responses-are-gzipped) it isn't exactly clear to me what text/* means, i.e. if thats a type indication or a path indication
> 
> @kmcnellis Well that was kind of rude of me. I shouldn't have assumed you wouldn't check it first. Sorry! I feel bad now

No worries!

PR #2670, opened 2022-06-06
Make sure that the sbt config files in server/project get live-synced to the docker image

kmcnellis:
### Description
They were getting copied only when the container was built, then ignored when edited

### Checklist
- [ x] Ran each env


bion (approved):

clouser2 (approved):

clouser2:
Thanks for doing this!

PR #2684, opened 2022-06-08
Find or create canonical questions on server start

bion:
Binds a `DatabaseSeedTask` instance as an eager singleton, which has a constructor that implements seeding logic.

Runs the seeds in a serializable transaction and retries upon failure.

closes https://github.com/seattle-uat/civiform/issues/2484

### DIFF
--- a/server/app/modules/DatabaseSeedModule.java
+++ b/server/app/modules/DatabaseSeedModule.java
@@ -1,0 +1,12 @@
+package modules;
+
+import com.google.inject.AbstractModule;
+import tasks.DatabaseSeedTask;
+
+public class DatabaseSeedModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(DatabaseSeedTask.class).asEagerSingleton();
+  }
+}
--- a/server/app/modules/DatabaseSeedModule.java
+++ b/server/app/modules/DatabaseSeedModule.java
@@ -3,6 +3,10 @@ package modules;
 import com.google.inject.AbstractModule;
 import tasks.DatabaseSeedTask;
 
+/**
+ * Binds the {@link DatabaseSeedTask} as an eager singleton, which causes it to run at server start
+ * time.
+ */
 public class DatabaseSeedModule extends AbstractModule {
 
   @Override
--- a/server/app/modules/DatabaseSeedModule.java
+++ b/server/app/modules/DatabaseSeedModule.java
@@ -1,16 +1,35 @@
 package modules;
 
+import akka.actor.ActorSystem;
 import com.google.inject.AbstractModule;
+import com.google.inject.Inject;
+import java.time.Duration;
+import javax.inject.Provider;
+import scala.concurrent.ExecutionContext;
 import tasks.DatabaseSeedTask;
 
 /**
- * Binds the {@link DatabaseSeedTask} as an eager singleton, which causes it to run at server start
- * time.
+ * Binds the {@link DatabaseSeedScheduler} as an eager singleton, which causes it to run at server
+ * start time.
  */
 public class DatabaseSeedModule extends AbstractModule {
 
   @Override
   protected void configure() {
-    bind(DatabaseSeedTask.class).asEagerSingleton();
+    bind(DatabaseSeedScheduler.class).asEagerSingleton();
+  }
+
+  public static class DatabaseSeedScheduler {
+
+    @Inject
+    public DatabaseSeedScheduler(
+        ActorSystem actorSystem,
+        ExecutionContext executionContext,
+        Provider<DatabaseSeedTask> databaseSeedTaskProvider) {
+      actorSystem
+          .scheduler()
+          .scheduleOnce(
+              Duration.ofMillis(10), () -> databaseSeedTaskProvider.get().run(), executionContext);
+    }
   }
 }
--- a/server/app/repository/QuestionRepository.java
+++ b/server/app/repository/QuestionRepository.java
@@ -7,9 +7,11 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.ebean.DB;
 import io.ebean.Database;
+import io.ebean.SqlRow;
 import io.ebean.Transaction;
 import io.ebean.TxScope;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
@@ -171,6 +173,17 @@ public class QuestionRepository {
         .sorted(Comparator.comparing(question -> question.getQuestionDefinition().getName()))
         .map(Question::getQuestionDefinition)
         .collect(ImmutableList.toImmutableList());
+  }
+
+  public ImmutableSet<String> getQuestionNames() {
+    ImmutableSet.Builder<String> names = ImmutableSet.builder();
+    List<SqlRow> rows = database.sqlQuery("SELECT DISTINCT name FROM questions").findList();
+
+    for (SqlRow row : rows) {
+      names.add(row.getString("name"));
+    }
+
+    return names.build();
   }
 
   private static class ConflictDetector {
--- a/server/app/repository/QuestionRepository.java
+++ b/server/app/repository/QuestionRepository.java
@@ -7,11 +7,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.ebean.DB;
 import io.ebean.Database;
-import io.ebean.SqlRow;
 import io.ebean.Transaction;
 import io.ebean.TxScope;
 import java.util.Comparator;
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
@@ -176,14 +174,13 @@ public class QuestionRepository {
   }
 
   public ImmutableSet<String> getQuestionNames() {
-    ImmutableSet.Builder<String> names = ImmutableSet.builder();
-    List<SqlRow> rows = database.sqlQuery("SELECT DISTINCT name FROM questions").findList();
+    return database.sqlQuery("SELECT DISTINCT name FROM questions").findList().stream()
+        .map((row) -> row.getString("name"))
+        .collect(ImmutableSet.toImmutableSet());
+  }
 
-    for (SqlRow row : rows) {
-      names.add(row.getString("name"));
-    }
-
-    return names.build();
+  public boolean questionExists(String questionName) {
+    return database.find(Question.class).where().eq("name", questionName).exists();
   }
 
   private static class ConflictDetector {
--- a/server/app/tasks/DatabaseSeedTask.java
+++ b/server/app/tasks/DatabaseSeedTask.java
@@ -100,7 +100,7 @@ public final class DatabaseSeedTask {
    * the database, inserting the definitions in {@code CANONICAL_QUESTIONS} if any aren't found.
    */
   private void seedCanonicalQuestions() {
-    if (!isMissingCanonicalQuestions()) {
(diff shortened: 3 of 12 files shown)


server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+            String.format(
+                "Unable to create canonical question \"%s\" due to %s",
+                questionDefinition.getName(), errorMessages));
+      } else {
+        LOGGER.info("Created canonical question \"%s\"", questionDefinition.getName());
+      }
+    }
+  }
+
+  private void inSerializableTransaction(Runnable fn, int tryCount) {
```
bion, line 130: Note to reviewers:

I can imagine this maybe (??) causing livelock if multiple servers start at near enough the same time and try to seed at near enough the same time, I think limiting the number of retries should mean that in such a circumstance eventually all of the servers except one will run out of retries in which case the last server standing wins and gets to commit its transaction.

Other things we could do here to reduce this problem:

A. Use the [akka scheduler](https://www.playframework.com/documentation/2.8.x/ScheduledTasks) to reschedule retries instead of recursing. Could reschedule with some random jitter to try and reduce the likelihood of servers scheduling their retries at the same time.

B. Schedule the initial attempt at a random delay. The downside there is that the first request to a rebuilt dev server won't trigger seeding so if you clear your questions in dev and reload the questions index it won't have the canonical questions until you reload again.

A and B aren't mutually exclusive.

server/app/repository/QuestionRepository.java
```diff
@@ -173,6 +175,17 @@ public ImmutableList<QuestionDefinition> getAllQuestionsForTag(QuestionTag tag)
         .collect(ImmutableList.toImmutableList());
   }
 
+  public ImmutableSet<String> getQuestionNames() {
+    ImmutableSet.Builder<String> names = ImmutableSet.builder();
+    List<SqlRow> rows = database.sqlQuery("SELECT DISTINCT name FROM questions").findList();
```
clouser2, line 180: Any specific reason to use raw SQL rather than the ORM?

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+import services.ErrorAnd;
+import services.LocalizedStrings;
+import services.question.QuestionService;
+import services.question.types.DateQuestionDefinition;
+import services.question.types.NameQuestionDefinition;
+import services.question.types.QuestionDefinition;
+
+/**
+ * Task for seeding the database. All of its work is done in the constructor, which is its only
+ * public method.
```
clouser2, line 34: Why does the work need to be done in the constructor? Is that a constraint imposed by the fact that it's a "Task" or is it something different?

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+ * with a transaction scope of SERIALIZABLE. If the transaction fails it retries up to {@code
+ * MAX_RETRIES} times.
+ */
+public final class DatabaseSeedTask {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DatabaseSeedTask.class);
+  private static final int MAX_RETRIES = 10;
+  private static final ImmutableList<QuestionDefinition> CANONICAL_QUESTIONS =
+      ImmutableList.of(
+          new NameQuestionDefinition(
```
clouser2, line 48: Not blocking and way outside the scope of this PR: Is there a builder pattern established for this that we could leverage rather than needing to comment-document each of the fields below?

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+      inSerializableTransaction(fn, ++tryCount);
+    } finally {
+      if (transaction.isActive()) {
+        transaction.end();
+      }
+    }
+  }
+
+  private boolean isMissingCanonicalQuestions() {
+    var questionService = questionServiceProvider.get();
```
clouser2, line 154: Not blocking: Having a separate variable here doesn't seem necessary as it's only used once.

What about:

`ImmutableSet<String> existingQuestionNames = questionServiceProvider.get().getQuestionNames();`

server/app/tasks/DatabaseSeedTask.java
clouser2: Just to confirm, this task wouldn't block the server being considered healthy, correct? If we were to block serving requests on new server startup / restart, it'd be unfortunate for a livelock situation to cause perceived slowness in responding to requests.

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+      transaction.commit();
+    } catch (NonUniqueResultException | SerializableConflictException | RollbackException e) {
+      LOGGER.warn("Database seed transaction failed: %s", e);
+
+      if (tryCount > MAX_RETRIES) {
+        throw new RuntimeException(e);
+      }
+
+      transaction.end();
+      inSerializableTransaction(fn, ++tryCount);
```
clouser2, line 145: Related to the above. Rather than rescheduling, could we Thread.sleep for a random backoff interval? Or is this occurring on the main thread?

Retry-in-a-tight-loop patterns generally make me wary.

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+        database.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE));
+
+    try {
+      fn.run();
+      transaction.commit();
+    } catch (NonUniqueResultException | SerializableConflictException | RollbackException e) {
+      LOGGER.warn("Database seed transaction failed: %s", e);
+
+      if (tryCount > MAX_RETRIES) {
+        throw new RuntimeException(e);
```
clouser2, line 141: Will this cause the server to crash?

I could actually see that being desired behavior. I could also see an argument for saying "let's let this fail and accept the 500s that would come from functionality that depends on it". I just think that decision needs to be made explicit in a comment.

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+      ErrorAnd<QuestionDefinition, CiviFormError> result =
+          questionService.create(questionDefinition);
+
+      if (result.isError()) {
+        String errorMessages =
+            result.getErrors().stream()
+                .map(CiviFormError::message)
+                .collect(Collectors.joining(", "));
+
+        LOGGER.error(
```
clouser2, line 120: What should we do in this case since downstream functionality will depend on it? Check-fail the server? Continue on and accept that downstream features that depend on these questions existing will throw errors?

clouser2 (changes requested):

server/app/repository/QuestionRepository.java
bion: I wasn't able to figure out how to do it with the ORM in a way that didn't fetch all columns of the table

server/app/tasks/DatabaseSeedTask.java
bion: It's a constraint imposed by Play's method of running code at server start time, which is to bind it as an eager singleton in DI: https://www.playframework.com/documentation/2.8.x/ScheduledTasks#Starting-tasks-when-your-app-starts

server/app/tasks/DatabaseSeedTask.java
bion: [There is](https://github.com/seattle-uat/civiform/blob/main/server/app/services/question/types/QuestionDefinitionBuilder.java#L18). I decided against using it because `QuestionDefinitionBuilder#build()` requires handling a `UnsupportedQuestionTypeException` which would make for messier code in this class IMO. Thoughts?

server/app/tasks/DatabaseSeedTask.java
bion: done

server/app/tasks/DatabaseSeedTask.java
```diff
@@ -0,0 +1,162 @@
...
+    this.versionRepository = checkNotNull(versionRepository);
+    this.database = DB.getDefault();
+
+    this.run();
+  }
+
+  private void run() {
+    if (isMissingCanonicalQuestions()) {
+      // Ensure a draft version exists to avoid transaction collisions.
+      versionRepository.get().getDraftVersion();
```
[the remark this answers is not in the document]
bion: Initializing the draft has its own transaction logic which ensures that only one draft ever exists in the system at the time. If there is already a thread-local, required transaction when `getDraftVersion` runs, it results in a stack overflow due to `getDraftVersion`'s retry logic. Calling `getDraftVersion` here ensures a draft is available while keeping the transactions separate.

server/app/tasks/DatabaseSeedTask.java
bion: I believe it is happening on the main thread, which, per your question above, would block the server from becoming healthy. I'm not 100% on the behavior there so I'm going to experiment with it some and let you know.

kmcnellis (changes requested):

server/app/repository/QuestionRepository.java
```diff
@@ -173,6 +175,17 @@ public ImmutableList<QuestionDefinition> getAllQuestionsForTag(QuestionTag tag)
...
+
+    for (SqlRow row : rows) {
+      names.add(row.getString("name"));
```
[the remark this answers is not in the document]
bion: yep, done

server/app/tasks/DatabaseSeedTask.java
bion: It will not.

server/app/tasks/DatabaseSeedTask.java
bion: Added some discussion of this to the top-level javadoc

server/app/tasks/DatabaseSeedTask.java
bion: I think it should continue on after failing. This code will only run when the server starts for the first time with a completely fresh database. I'm leery of designing it in such a way that it could prevent the server from operating at all if something goes wrong.

server/app/repository/QuestionRepository.java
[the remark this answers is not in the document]
bion: done

server/app/tasks/DatabaseSeedTask.java
bion: done, added a new method `unsafeBuild` to get around error handling

kmcnellis (approved):

server/app/modules/DatabaseSeedModule.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Inject
+    public DatabaseSeedScheduler(
+        ActorSystem actorSystem,
+        ExecutionContext executionContext,
+        Provider<DatabaseSeedTask> databaseSeedTaskProvider) {
+      actorSystem
+          .scheduler()
+          .scheduleOnce(
+              Duration.ofMillis(10), () -> databaseSeedTaskProvider.get().run(), executionContext);
```
clouser2, line 32: Not blocking: I'd either choose a randomish value with jitter or don't have a delay at all. My rationale is that the goal of adding a sleep is to avoid synchronous calls from multiple servers starting up at the same time.

server/app/tasks/DatabaseSeedTask.java
clouser2: Thanks for updating that!

clouser2 (approved):

server/app/modules/DatabaseSeedModule.java
bion: discussed out-of-band

Issue #2573, opened 2022-05-25
FR: Applicant view shows previous applications.


shanemc-goog:
Currently applicants can't see if they're applied to a program.

Showing the information is good for user visibility, but also can reduce re-applications and will be necessary for the feature that will show the status of the application.

First iteration can just show which programs have been applied too.

Subsequent ones could show the data.

kristinyim:
@cassandralfernandes is going to own this as part of her status tracking PRD

clouser2:
I'll take a look at this next week and sync on what the MVP is here. I imagine we'll want this to look visually consistent initially with the current applicant experience.

clouser2:
Rather than building out a separate UI, I wonder if we could instead leverage the existing programs page for this. For logged-in users, we separate programs for which there is an active draft application from other programs:
<img width="721" alt="Screen Shot 2022-05-26 at 3 36 48 PM" src="https://user-images.githubusercontent.com/1870301/170591017-a3263cef-a921-4481-8c2d-341109498a3b.png">

If a user has already completed an application for the program, we put it in the "New Programs" section and let the user re-apply.

A potential low-effort path forward could be to:
* Put a stamp on the program card if the user has already completed the application saying "last applied 5/23"
* Change the text from "Apply" to "Re-apply".
* Consider removing the "New Programs" and "Continue" sections in favor of a single section, trying to sort the programs by some TBD heuristic. OR move previously applied programs into this section and keep at the top.

clouser2:
@rahaghas for feedback / guidance on the above.

shanemc-goog:
Sounds good to me, Seattle only has 5 programs so I think an organization like that works, and there are mocks for a larger redesign that can iterate on this.

@rahaghas what are your thoughts here?

rahaghas:
I think this is great, and we heard from TIs in Seattle saying they would want exactly this organization.

PR #2695, opened 2022-06-09
Applicant-facing program index page contains info for previously submitted applications

clouser2:
### Description
Previously, we were only showing information for in-progress applications. This PR separates out programs that already have a submitted application and also attach the date the application was submitted. Prior to this, already-applied-for programs would appear in the "New Programs" section.

This implements the proposal described in https://github.com/seattle-uat/civiform/issues/2573#issuecomment-1139121336.

Screen Shot 2022-06-09 at 10 41 54 AM

The majority of the complexity contained within this PR is in `ApplicantServiceImpl`. Previously, we were doing the DB query in `UserRepository` and then performing program deduplication to account for the fact that applications could exist for previous versions of the program. Rather than having those as separate concerns, I lumped it all into `ApplicantServiceImpl.relevantProgramsForApplicant`.

Note: We'll need to wait for translations from #2691 to be live before merging this.

While here, I also fixed a port mapping issue that prevented login from `bin/run-dev`.

### Checklist

### Issue(s)
#2573


server/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -68,56 +61,16 @@ public CompletionStage<Result> index(Request request, long applicantId) {
...
-              Set<String> adminNames =
-                  programsWithDraftApplications.stream()
-                      .map(ProgramDefinition::adminName)
-                      .collect(Collectors.toSet());
-              ImmutableList<ProgramDefinition> dedupedActivePrograms =
-                  allPrograms.get(LifecycleStage.ACTIVE).stream()
-                      .filter(
-                          programDefinition -> !adminNames.contains(programDefinition.adminName()))
-                      .collect(ImmutableList.toImmutableList());
-              // Deduplicate programs with draft applications. This is a temporary fix for a bug
```
clouser2, line 85: Note to the reviewer: This deduplication / logging moved into ApplicantServiceImpl.relevantProgramsForApplicant.

server/app/repository/ApplicationRepository.java
```diff
@@ -204,4 +205,28 @@ public CompletionStage<Optional<Application>> getApplication(long applicationId)
...
+   *
+   * <p>The {@link Program} associated with the application is eagerly loaded.
+   */
+  public CompletionStage<ImmutableSet<Application>> getApplicationsForApplicant(
+      long applicantId, ImmutableSet<LifecycleStage> stages) {
+    return supplyAsync(
+        () -> {
+          return database
+              .find(Application.class)
+              .alias("a")
```
shanemc-goog, line 220: this looks unecessary.

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,138 @@ public ImmutableList<Application> getAllApplications() {
...
+    Map<String, Map<LifecycleStage, Optional<Application>>> mostRecentApplicationsByProgram =
+        applications.stream()
+            .collect(
+                Collectors.groupingBy(
+                    a -> {
+                      return a.getProgram().getProgramDefinition().adminName();
+                    },
+                    Collectors.groupingBy(
+                        Application::getLifecycleStage,
+                        // In practice, we don't expect an applicant to have multiple applications
```
shanemc-goog, line 411: This isn't the case for obsolete. maybe just note that.

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,138 @@ public ImmutableList<Application> getAllApplications() {
...
+                        Collectors.maxBy(Comparator.comparingLong(a -> a.id)))));
+
+    ImmutableList.Builder<ApplicantProgramData> inProgressPrograms = ImmutableList.builder();
+    ImmutableList.Builder<ApplicantProgramData> submittedPrograms = ImmutableList.builder();
+    ImmutableList.Builder<ApplicantProgramData> unappliedPrograms = ImmutableList.builder();
+
+    mostRecentApplicationsByProgram.forEach(
+        (programName, appByStage) -> {
+          Optional<Application> maybeDraftApp =
+              appByStage.getOrDefault(LifecycleStage.DRAFT, Optional.empty());
```
shanemc-goog, line 423: consider `Optional.ofNullable(appByStage.get(LifecycleStage.DRAFT);`
instead.

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,138 @@ public ImmutableList<Application> getAllApplications() {
...
+          Optional<Application> maybeSubmittedApp =
+              appByStage.getOrDefault(LifecycleStage.ACTIVE, Optional.empty());
+          Optional<Instant> maybeSubmitTime = maybeSubmittedApp.map(Application::getSubmitTime);
+          if (maybeDraftApp.isPresent()) {
+            // We want to ensure that any generated links points to the version
+            // of the program associated with the draft, not the most recent version.
```
shanemc-goog, line 429: I don't think this comment is necessary, the code seems fairly obvious and readable without restating how the system modeling works. 

if you keep the comment be clearer about lifecyclstage / table since stages are used in both modelings. eg "draft application" "most recent program version"

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,138 @@ public ImmutableList<Application> getAllApplications() {
...
+            // As such, we use the program definition associated with the application.
+            inProgressPrograms.add(
+                ApplicantProgramData.create(
+                    maybeDraftApp.get().getProgram().getProgramDefinition(), maybeSubmitTime));
```
shanemc-goog, line 433: You're using the submitted apps time not the draft's time.

server/app/repository/ApplicationRepository.java
clouser2: Nice catch!

server/app/services/applicant/ApplicantService.java
```diff
@@ -82,24 +84,59 @@ CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProgramServ
...
-   * Return all applications, including applications from previous versions, with program,
-   * applicant, and account associations eager loaded.
+   * Return all applications for all applicants, including applications from previous versions, with
+   * program, applicant, and account associations eager loaded.
    */
   ImmutableList<Application> getAllApplications();
+
+  /**
+   * Return all programs that are appropriate to serve to an applicant. Appropriate programs are
+   * those where the applicant: * Has a draft application * Has previously applied * Any other
```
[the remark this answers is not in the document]
clouser2: Thanks. I totally forgot about javadocs using HTML.

server/app/services/applicant/ApplicantService.java
```diff
@@ -82,24 +84,59 @@ CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProgramServ
...
+   * programs that are public.
```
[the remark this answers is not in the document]
clouser2: Good question. For "draft" applications, we were showing them (even prior to this PR).

The change from this PR would be that we would show programs for which the applicant had applied in the past even if they were discontinued).

I'll follow up with Cassandra about whether we should hide programs even if they completed / draft applications.

@cassandralfernandes , mind providing feedback?

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,138 @@ public ImmutableList<Application> getAllApplications() {
...
+          checkForDuplicateDrafts(applications);
+          return relevantProgramsForApplicant(activePrograms, applications);
+        },
+        httpExecutionContext.current());
+  }
+
+  private RelevantPrograms relevantProgramsForApplicant(
+      ImmutableList<ProgramDefinition> activePrograms, ImmutableSet<Application> applications) {
+    ImmutableMap<String, ProgramDefinition> activeProgramNames =
+        ImmutableMap.copyOf(
```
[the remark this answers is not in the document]
clouser2: I should add a comment for this. I explicitly collected to the ImmutableMap to avoid a case where the activePrograms stream contained multiple entries with the same adminName. Shouldn't happen in practice, but did in my unit tests. The ImmutableMap collector will fail since immutablemaps can't be built with duplicate keys. Here, the last one in would win.

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: Reworded, thanks for noticing.

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: `appsByStage.get(LifecycleStage.DRAFT)` would throw if the key didn't exist. This could be the case if there weren't any draft applications for the program, but there was a previously submitted one (e.g. `LifecycleStage.ACTIVE`).

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: Removed the comment.

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: That's correct. We still want to show when the program was submitted, even if there's a draft version. Additionally, the draft version doesn't have a "last edited" timestamp, as I understand.

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
clouser2: Spoke with @cassandralfernandes and we're going to circle back on a decision here. For now, I'm going to preserve the existing behavior when a program is hidden from the index of 1) showing draft applications and 2) not showing submitted applications.

I left a TODO to track resolving this.

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,139 @@ public ImmutableList<Application> getAllApplications() {
...
+                    maybeDraftApp.get().getProgram().getProgramDefinition(), maybeSubmitTime));
+            programNamesWithApplications.add(programName);
+          } else if (maybeSubmittedApp.isPresent() && activeProgramNames.containsKey(programName)) {
+            ProgramDefinition programDef = activeProgramNames.get(programName);
+            submittedPrograms.add(ApplicantProgramData.create(programDef, maybeSubmitTime));
+            programNamesWithApplications.add(programName);
+          }
+        });
+
+    Set<String> missingActivePrograms =
```
shanemc-goog, line 445: nit: missing -> unapplied

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,139 @@ public ImmutableList<Application> getAllApplications() {
...
+        Sets.difference(activeProgramNames.keySet(), programNamesWithApplications);
+    missingActivePrograms.forEach(
+        programName -> {
+          unappliedPrograms.add(
+              ApplicantProgramData.create(activeProgramNames.get(programName), Optional.empty()));
+        });
+
+    // Ensure each list is ordered by database ID for consistent ordering.
```
shanemc-goog, line 453: Per the above comment, the program.id is not the most correct ordering given the system version can be reverted.

server/test/views/TranslationUtilsTest.java
```diff
@@ -0,0 +1,49 @@
+package views;
```
shanemc-goog, line 1: likewise, I don't see that this is used in the PR

shanemc-goog (changes requested):
I took a deeper look.

There's two deeper issues around 

1. If we should handle violations of system invariants such as if there are multiple Active programs with the same name.

2. That program.id order is not a fully correct proxy for application order.

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
shanemc-goog: The TODO doesn't show in application code only test, put it here too

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: > Will the UI will distinguish a draft with a previously submitted application?

Yes - it will be in the "In progress" section and contain the title "Submitted 2/2/2022".

> I'm not quite understanding why it makes sense to have the wrong latestApplicationSubmitTime associated with an application. Would it be equally correct to name the field latestSubmittedApplicationTime? If so that is clearer naming imo.

To be clear, the information displayed in the UI is the latest *submitted* application time, so attaching the submitted time for a distinct application to a draft application is still correct with respect to what we want to display in the UI.

That said, I agree that the naming is confusing. Happy to rename.

server/app/services/applicant/ApplicantServiceImpl.java
clouser2: Agreed. That said, I think this is the current behavior (which I was attempting to preserve).

We gather the list of active programs based on "natural order" from the active version (e.g. no ordering applied to the query) ([code](https://github.com/seattle-uat/civiform/blob/e1f02bb0b6975f308745365a40ad32203e830ef0/server/app/models/Version.java#L78-L80)). The natural order for the many to many would be the Primary Key, which uses the program ID ([code](https://github.com/seattle-uat/civiform/blob/e1f02bb0b6975f308745365a40ad32203e830ef0/server/conf/evolutions/default/17.sql#L19)).

Given this, would you prefer we use a different ordering? I'm not sure what we'd want to use since there's not a clear signal. Using alphabetical seems like it'd be a big change from what we have now.

server/app/services/applicant/ApplicantService.java
```diff
@@ -82,24 +84,64 @@ CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProgramServ
...
+   * <ul>
+   *   <li>Has a draft application
+   *   <li>Has previously applied
+   *   <li>Any other programs that are public
+   * </ul>
+   */
+  CompletionStage<RelevantPrograms> relevantProgramsForApplicant(long applicantId);
+
+  @AutoValue
+  public abstract static class ApplicantProgramData {
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/services/applicant/ApplicantService.java
```diff
@@ -82,24 +84,64 @@ CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProgramServ
...
+    static ApplicantProgramData create(
+        ProgramDefinition program, Optional<Instant> latestApplicationSubmitTime) {
+      return new AutoValue_ApplicantService_ApplicantProgramData(
+          program, latestApplicationSubmitTime);
+    }
+  }
+
+  @AutoValue
+  public abstract static class RelevantPrograms {
+    public abstract ImmutableList<ApplicantProgramData> inProgress();
```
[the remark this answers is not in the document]
shanemc-goog: Ah, I thought it might be permissible to only set some of these values such that we'd want default otherwise.

Curious, why have the builder rather than the default create method that requires all 3 be passed in?

server/app/services/applicant/ApplicantServiceImpl.java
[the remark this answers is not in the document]
shanemc-goog: Sorry didn't realize you were maintaining existing functionality.

hmmmm, that's rough.  Ok let's leave it, the comment helps suss out the oddness the existing code was trying to handle.


( I think the existing code would show the duplicate name as two programs oddly enough...)

server/app/services/applicant/ApplicantServiceImpl.java
```diff
@@ -358,6 +367,139 @@ public ImmutableList<Application> getAllApplications() {
...
+    // case, the ImmutableMap collector would throw since ImmutableMap builders don't allow
+    // construction where the same key is provided twice. Using Collectors.toMap would just
+    // use the last provided key.
+    ImmutableMap<String, ProgramDefinition> activeProgramNames =
+        ImmutableMap.copyOf(
+            activePrograms.stream()
+                .collect(Collectors.toMap(ProgramDefinition::adminName, pdef -> pdef)));
+
+    // When new versions of Programs are created, they have distinct IDs but retain the
+    // same adminName. In order to find the most recent draft / active application,
```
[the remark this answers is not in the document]
shanemc-goog: ok, I'll defer to you.  I've managed to get into some git hub state where it's hard to track changes to really say more here.

server/app/services/applicant/ApplicantServiceImpl.java
shanemc-goog: ahhh, right.  I missed that fact.   If only there was Optional.ofNullableOrUnwrapOptionalsYolo()

server/app/views/TranslationUtils.java
```diff
@@ -0,0 +1,67 @@
+package views;
```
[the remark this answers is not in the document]
shanemc-goog: thanks, my searching didn't find it on my own :(

server/app/views/TranslationUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import services.MessageKey;
+
+/** Utility class for managing translated content. */
+public final class TranslationUtils {
+  /**
+   * Helper to split a translated string with a single argument into the components preceeding the
+   * argument and the components after the argument.
+   *
+   * <p>This makes it possible to visually style subcomponents of the translated message separately
+   * when different languages may put the argument in a different position within the translated
```
shanemc-goog, line 17: This won't work for RtL languages I think, though I don't think we currently need to or do it otherwise.  Can you make an issue and a todo so it's easier to find when we need to support one?

shanemc-goog (approved):

clouser2:
We have translations available for all but two of the languages and we're going to proceed with merging. Since approval was last given on 6/15, it'd probably be a good idea to have another quick look at this for correctness, @shanemc-goog.

clouser2:
I looked at the overall diff to ensure that this still seems correct. Additionally, I had two diffs side by side, the first being the diff against main and the second being the diff "since my last review" from June. In both cases, the diffs looked the same. This also caught the fact that I had failed to preserve some javadocs when merging over the months. Those have now been restored.

The only new dependency we've added on this is in RedirectController.java. From my perspective, that's the only new change that needs to be freshly reviewed.

server/app/services/applicant/ApplicantService.java
```diff
@@ -465,6 +469,175 @@ public ImmutableList<Application> getApplications(TimeFilter submitTimeFilter) {
...
+
+  private ImmutableList<ApplicantProgramData> sortByProgramId(
+      ImmutableList<ApplicantProgramData> programs) {
+    return programs.stream()
+        .sorted(Comparator.comparing(p -> p.program().id()))
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  private void checkForDuplicateDrafts(ImmutableSet<Application> applications) {
+    // Add logging to better understand a bug where some applicants were seeing
```
shanemc-goog, line 590: nit: Use active language.

The logging is to better understand....

Also perhaps this should just be the method javadoc since it doesn't check/throw like our other check methods doo (checkNotNull) etc.  maybe rename, logDuplicateDrafts etc

server/app/services/applicant/ApplicantService.java
```diff
@@ -465,6 +469,175 @@ public ImmutableList<Application> getApplications(TimeFilter submitTimeFilter) {
...
+      if (draftApplications.size() > 1) {
+        String joinedProgramIds =
+            String.join(
+                ", ",
+                draftApplications.stream()
+                    .map(a -> String.format("%d", a.getProgram().getProgramDefinition().id()))
+                    .collect(ImmutableList.toImmutableList()));
+        logger.debug(
+            String.format(
+                "DEBUG LOG ID: 98afa07855eb8e69338b5af13236a6b7. Program"
```
shanemc-goog, line 615: what is this random ID?  It's not clear it's utility or why is needs to be SUPER unique rather than search-filter unique.

server/app/views/TranslationUtils.java
shanemc-goog: I thought I replied, but don't see it on this thread..

What was the outcome? 

My concern was that parts 1 and 2 could get flipped in a RtL language such that we'd highlight the opposite part.

Adding a bug is fine.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -140,8 +148,7 @@ private DivTag topContent(String titleText, String infoTextLine1, String infoTex
 
   private DivTag mainContent(
       Messages messages,
-      ImmutableList<ProgramDefinition> draftPrograms,
-      ImmutableList<ProgramDefinition> activePrograms,
+      ApplicantService.RelevantPrograms relevantPrograms,
```
shanemc-goog, line 151: nit: what do you think about naming this applicationPrograms, or similar.

"relevantPrograms" feels too program focused, which then makes reading relevantPrograms.inProgress harder as the inprogress means "draft application" and not "draft program" which I think is not very clear.

shanemc-goog (changes requested):
just naming/readability thoughts mostly

server/app/services/applicant/ApplicantService.java
clouser2: Done. Used more active language, moved to javadoc, and renamed to `logDuplicateDrafts`.

server/app/services/applicant/ApplicantService.java
clouser2: It's copied forward from where it was originally in ApplicantProgramsController. I'm not planning to change it.

server/app/views/TranslationUtils.java
clouser2: I still think my June 16th comment is relevant for this question. Regarding the specific scenario of part 1 and part 2 getting flipped in a RtL language, I don't think that would be an issue since we only have two components:
  1) The "static" text
  2) The dynamic value

We would apply visual styling to one of these but not the other and the helper accounts for the dynamic value being in a different location (e.g. before instead of after) in an RtL language.

I'm not sure what to put in the bug since I don't actually think that this is introducing a new issue for languages of that sort.

server/app/views/applicant/ProgramIndexView.java
clouser2: Done. Also renamed the type to `ApplicationPrograms`.

shanemc-goog (approved):

PR #2698, opened 2022-06-10
Jacoco Code coverage

### LINKED ISSUE #1934: CiviForm has unknown unit test coverage
We estimate our level of unit test coverage to be poor, but do not have a metric yet since we do not have a unit test code coverage tool. We should investigate a good open source option and add it.

swatkat1:
### Description
Adding code coverage using Jacoco plugin. The plugin creates two types of report
*html report under server/jacoco/report/html/index.html (can be viewed locally)
*a cvs file which is shared with the jacoco-badge-generator(link: https://github.com/marketplace/actions/jacoco-badge-generator#inputs) which helps us create a badge to show in README 

### Checklist

### Issue(s)
Fixes #1934


.gitignore
```diff
@@ -1,6 +1,8 @@
 /project/
 flakes
 logs/
+*.html
```
kmcnellis, line 4: I don't thing adding *.html is a good idea. Better just stick with server/jacoco

docker-compose.dev.yml
```diff
@@ -12,6 +12,7 @@ x-sbt-volumes:
     - node_modules-data:/usr/src/server/node_modules
     - project-data:/usr/src/server/project/project
     - project-target-data:/usr/src/server/project/target
+    - project-coverage:/usr/src/server/target/scala-2.13/jacoco/report
```
kmcnellis, line 15: Remove project-coverage

server/build.sbt
```diff
@@ -145,12 +146,23 @@ lazy val root = (project in file("."))
   )
   .settings(excludeTailwindGeneration: _*)
 
+  //jacoco report setting
+jacocoReportSettings := JacocoReportSettings()
+  .withTitle("Jacoco")
+  .withFormats(JacocoReportFormats.CSV)
+//html for local views
+jacocoReportSettings := JacocoReportSettings()
```
kmcnellis, line 154: I'm pretty sure this is overriding the CSV version.  I think you can do `withFormats(JacocoReportFormats.CSV, JacocoReportFormats.HTML)`

kmcnellis: (see https://github.com/sbt/sbt-jacoco/blob/eb82c08f694e9e79c810906e8ad6420f8c4987a5/src/sbt-test/sbt-jacoco/report-settings/build.sbt)

kmcnellis (changes requested):

.gitignore
swatkat1: done.

docker-compose.dev.yml
clouser2: @kmcnellis is the rationale that we'd want to mirror the local filesystem for Jacoco reports rather than putting in on a shared Docker volume?

If that's the rationale, I think I'd agree. That way one could just have the coverage report show up in their local filesystem after a run rather than needing to start a Docker container with the volume attached in order to view it.

server/build.sbt
```diff
@@ -145,12 +146,23 @@ lazy val root = (project in file("."))
   )
   .settings(excludeTailwindGeneration: _*)
 
+  //jacoco report setting
+jacocoReportSettings := JacocoReportSettings()
+  .withTitle("Jacoco")
```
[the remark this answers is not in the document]
swatkat1: Given the fact that I tried many tools and only Jacoco worked worked as expected- shouldn't we give some credit to it :) But to your point- code-coverage does sound good :)

docker-compose.dev.yml
kmcnellis: Yes, to both of you.  I think this was just an artifact from the first approach

.github/workflows/tests.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       run: docker build -t civiform-dev --cache-from docker.io/civiform/civiform-dev:latest ./
     - name: Run tests
       run: USE_LOCAL_CIVIFORM=true bin/run-test-ci
+    - name: jacoco-badge-generator
```
kmcnellis, line 31: Nit: rename this to run_junit_test_and_coverage (or something like that

kmcnellis (approved):

.github/workflows/tests.yml
swatkat1: The Step "name: Run tests" - actually runs the test and creates the report.. This step uses the report to generate the badge..

clouser2 (approved):

PR #2699, opened 2022-06-10
Enable the sbt remote cache

kmcnellis:
### Description
Runs a full sbt compile on the docker image build, and saves the assets locally (outside of the server directory, so docker volumes don't shadow/override it).  Then on sbt load, pulls from the "remote" (really local) cache into our build directory. 

Seems to reuse ~80% of assets, and speeds up the build more than 50%

### Checklist
- [ ] Ran it locally

### DIFF
--- a/bin/build-dev
+++ b/bin/build-dev
@@ -7,7 +7,7 @@ source bin/lib.sh
 # Build the new development image.
 docker build \
   -t civiform-dev \
-  --cache-from civiform/civiform-browser-test:latest \
+  --cache-from civiform/civiform-dev:latest \
   --build-arg BUILDKIT_INLINE_CACHE=1 \
   .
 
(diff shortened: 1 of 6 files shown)


bin/build-dev
```diff
@@ -7,7 +7,7 @@ source bin/lib.sh
 # Build the new development image.
 docker build \
   -t civiform-dev \
-  --cache-from civiform/civiform-browser-test:latest \
+  --cache-from civiform/civiform-dev:latest \
```
bion, line 10: oof, good catch

bion (changes requested):
Please add a comment in the build.sbt file giving a general overview of how this works and where the different parts live

kmcnellis:
> Please add a comment in the build.sbt file giving a general overview of how this works and where the different parts live

done

bion (approved):

clouser2 (approved):

PR #2702, opened 2022-06-11
Assets should be fingerprinted and cached for longer periods

kmcnellis:
### Description
Use the SBT-web pipeline to gzip and fingerprint all assets.  Then we can enable a longer cache on these immutable files.  When they change, the non-cached html will request the file with the new digest, and skip the browser cache.

This should improve page-navigation load speeds, as the browser doesn't need to re-verify extra files.

Not enabled in dev.

### Checklist
- [ ] Tested locally


server/app/filters/DisableCachingFilter.java
```diff
@@ -22,11 +29,29 @@ public EssentialAction apply(EssentialAction next) {
...
+                    result -> {
+                      String path = request.uri();
+                      Integer status = result.status();
+                      ImmutableList<String> assets = staticAssetPaths;
+
+                      if (environment.isDev()) {
+                        // Must revalidate status asset caches in dev mode
+                        assets = ImmutableList.<String>of();
+                      }
+                      if (assets.stream().anyMatch(m -> path.contains(m))) {
```
clouser2, line 41: Consider using `startsWith` rather than `contains`.

I think this would be more accurate (e.g. not catch resources like `some-dynamic-admin-path/public/...`) and more performant (not have to scan the whole string to return false). The performance concern is unlikely to matter, but the correctness concern might.

server/build.sbt
```diff
@@ -110,6 +115,9 @@ lazy val root = (project in file("."))
     // After 2 transitive steps, do more aggressive invalidation
     // https://github.com/sbt/zinc/issues/911
     incOptions := incOptions.value.withTransitiveStep(2),
+    pipelineStages := Seq(digest, gzip), // plugins to use for assets
```
clouser2, line 118: Does this mean that the GzipFilter is no longer necessary?

clouser2 (approved):
Cache headers always make me wary. That said, it'll be nice to not make the browser re-request the content on every request :D

server/app/filters/DisableCachingFilter.java
```diff
@@ -11,8 +12,14 @@
 public class DisableCachingFilter extends EssentialFilter {
   private final Executor exec;
 
+  private final ImmutableList<String> staticAssetPaths =
+      ImmutableList.of("/public/", "/assets/", "favicon.ico", "/favicon.ico");
+
+  @Inject private play.Environment environment;
```
bion, line 18: suggestion (blocking): we use constructor injection everywhere else in the app to call `checkNotNull`, please use that pattern here as well

server/app/filters/DisableCachingFilter.java
```diff
@@ -11,8 +12,14 @@
 public class DisableCachingFilter extends EssentialFilter {
   private final Executor exec;
 
+  private final ImmutableList<String> staticAssetPaths =
```
bion, line 15: suggestion (blocking): make it a static constant
```suggestion
  private static final ImmutableList<String> STATIC_ASSET_PATHS =
```

server/app/filters/DisableCachingFilter.java
```diff
@@ -22,11 +29,29 @@ public EssentialAction apply(EssentialAction next) {
...
                 .map(
-                    result ->
-                        result
-                            .withHeader("Cache-Control", "no-cache, must-revalidate")
-                            .withHeader("Pragma", "no-cache")
-                            .withHeader("Expires", "0"),
+                    result -> {
+                      String path = request.uri();
+                      Integer status = result.status();
+                      ImmutableList<String> assets = staticAssetPaths;
```
bion, line 35: suggestion (blocking): rather than reassigning a variable, do `if (!environment.isDev() && ...)` on L41

server/app/filters/DisableCachingFilter.java
bion: Agree that doing startsWith is preferrable behavior for correctness. Could do
```java
private static final Splitter PATH_SPLITTER = Splitter.on("/');
...
Optional<String> pathPrefix = Iterables.getFirst(PATH_SPLITTER.splitToList(path));
if (pathPrefix.map(ASSET_PATH_PREFIXES::contains).orElse(false)) {
```
with that approach could also change the hard-coded list of path prefixes to an `ImmutableSet`

bion (changes requested):

server/app/filters/DisableCachingFilter.java
```diff
@@ -22,11 +29,29 @@ public EssentialAction apply(EssentialAction next) {
...
+                          // In prod/staging, static assets are fingerprinted,
+                          // so we can cache for a longer time.
+                          // Cache for 2 weeks.
+                          return result.withHeader(
+                              "Cache-Control", "public, max-age=1209600, immutable");
+                        }
+                      }
+                      // Don't cache anything else.
+                      return result.withHeader(
+                          "Cache-Control", "no-store, max-age=0, must-revalidate");
```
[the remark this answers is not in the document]
kmcnellis: Those are unnecessary and should no longer be used.  They have been deprecated for years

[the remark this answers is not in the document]
kmcnellis: Honestly, we also don't really need max-age or must-revalidate either. no-store is the strongest, and should always prevent all caching, regardless of what everything else says.   It's not invalid, though. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#use_cases

server/build.sbt
kmcnellis: Gzip filter is still needed for dyanmically generated files.  The gzip here is only for the static assets, and actually builds the gzip into the built package, so there's no overhead at run time.

server/app/filters/DisableCachingFilter.java
kmcnellis: done

server/app/filters/DisableCachingFilter.java
kmcnellis: Done. Yeah, sorry.  This was from play's example.

server/app/filters/DisableCachingFilter.java
```diff
@@ -22,11 +34,28 @@ public EssentialAction apply(EssentialAction next) {
...
-                            .withHeader("Cache-Control", "no-cache, must-revalidate")
-                            .withHeader("Pragma", "no-cache")
-                            .withHeader("Expires", "0"),
+                    result -> {
+                      Optional<String> pathPrefix = Iterables.getFirst(PATH_SPLITTER.splitToList(request.uri()));
+                      Integer status = result.status();
+
+                      if (environment.isDev()) {
+                        // Must revalidate status asset caches in dev mode
+                        // pathPrefix = Optional.empty();
```
bion, line 43: question (blocking): commented out?

server/app/filters/DisableCachingFilter.java
```diff
@@ -22,11 +34,28 @@ public EssentialAction apply(EssentialAction next) {
...
+                      }
+                      if (pathPrefix.map(ASSET_PATH_PREFIXES::contains).orElse(false)) {
+                        // Only cache when Status is OK https://web.dev/uses-long-cache-ttl/
+                        if (status == 200 || status == 203 || status == 206) {
```
bion, line 47: suggestion (blocking): use a single `if`.

(optional) could use another constant: `OK_STATUS_CODES = ImmutableSet.of(200, 203, 206);` and then

```java
private static final ImmutableSet<Integer> OK_STATUS_CODES = ImmutableSet.of(200, 203, 206);
...
if (pathPrefix.map(ASSET_PATH_PREFIXES::contains).orElse(false)
    // Only cache when Status is OK https://web.dev/uses-long-cache-ttl/
    && OK_STATUS_CODES.contains(status)) {
```

bion (changes requested):

server/app/filters/DisableCachingFilter.java
kmcnellis: This is out of date - it was disabled for testing

server/app/filters/DisableCachingFilter.java
kmcnellis: done

bion (approved):

PR #2704, opened 2022-06-13
Convert "View" to "Copy link" in the new program index UI

clouser2:
The interactivity was added in Typescript rather than inline on the button construction in Java since it was complicated and could benefit from the type system.

Two wrinkles I ran into when trying to implement this were:
1) We aren't using Typescript / Javascript modules, so it wasn't possible to take a dependency on the `ToastController` type (or global var `toastController`).
2) The JavaScript clipboard API may have potential permission errors (unlikely, but possible). In this case, the displayed toast contains the link.

Screen Shot 2022-06-13 at 11 51 57 AM

### Issue(s)
#1238


server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -351,17 +351,13 @@ private static Instant extractLastUpdated(
...
-    String viewLink =
+  Tag renderCopyProgramLink(ProgramDefinition program) {
+    String programLink =
         baseUrl
             + controllers.applicant.routes.RedirectController.programByName(program.slug()).url();
-    ContainerTag button =
-        makeSvgTextButton("View", Icons.VISIBILITY_SVG_PATH)
-            .withClass(AdminStyles.TERTIARY_BUTTON_STYLES);
-    return asRedirectButton(button, viewLink);
+    return makeSvgTextButton("Copy link", Icons.CONTENT_COPY_SVG_PATH)
```
bion, line 358: suggestion (blocking): maybe `Copy shareable link`?

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -26,6 +27,21 @@ class AdminPrograms {
         cardsPlaceholder.classList.add('hidden')
       }
     }
+
+    const withCopyableProgramLink = Array.from(
+      cardsParent.querySelectorAll(
+        `[${AdminPrograms.PROGRAM_LINK_ATTRIBUTE}]`
+      ) as any as Array<HTMLElement>
```
bion, line 34: question (blocking): what type does this get if there's no `as`?

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -26,6 +27,21 @@ class AdminPrograms {
...
+    )
+    withCopyableProgramLink.forEach((el) => {
+      const programLink = el.getAttribute(AdminPrograms.PROGRAM_LINK_ATTRIBUTE)
+      if (!programLink) {
+        return
```
bion, line 39: suggestion (blocking): log a warning that there's no program link where we expect it to be

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -61,6 +77,44 @@ class AdminPrograms {
       lastUpdatedMillis,
     }
   }
+
+  static async tryCopyToClipboard(programLink: string): Promise<boolean> {
```
bion, line 81: suggestion (blocking): please add a doc comment explaining the return value

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -61,6 +77,44 @@ class AdminPrograms {
...
+    if (!window.navigator['clipboard']) {
+      return false
+    }
+    try {
+      ;(await window.navigator['clipboard'].writeText(
+        programLink
+      )) as unknown as Promise<void>
```
bion, line 88: question (blocking): we're not using the return type, why cast the return value?

bion (changes requested):

server/app/views/admin/programs/ProgramIndexViewV2.java
bion: We want program admins to understand:

1. the link is a deep link, and this the only place to get it (copying the URL after visiting a specific program links to a specific version)
1. the link is human-readable (i.e. `/programs/utility-discount-program/` instead of `/programs/12/blocks/1`) which is what all users see when actually visiting a program with their browser

bion (changes requested):

server/app/views/admin/programs/ProgramIndexViewV2.java
clouser2: Spoke with @rahaghas , who suggested "Share link". I additionally added some hover text that is a riff on the text we originally had within the original version of the UI.

If it's acceptable to y'all, I'd suggest going with this, getting feedback, and iterating.

clouser2:
> I think saying "Share link" should suffice to make it clear that it can be used outside of Civiform. I don't think that hover text is necessary - it would be more useful for the hover text to make it clear that the link will be copied to the user's clipboard

Sounds good. I removed the hover text.

PR #2707, opened 2022-06-13
Change jacoco settings to global scope

kmcnellis:
They weren't being picked up in Test/


swatkat1:
Hi Kiana, Thanks for looking into this. I think we still should be able to view the HTML report locally and that why I had added the docker cp command to copy the entire report folder.. I will go ahead and add it back in the run-test-coverage file.

PR #2708, opened 2022-06-13
Seed canonical questions before prober runs

bion:
Running probers does not cause an app restart, but it does wipe the database. As a result the canonical questions are missing since they are automatically created when missing only on server start. This change adds a controller action that runs the database seed task and updates the browser tests to do so after wiping the database.

### DIFF
--- a/server/app/controllers/dev/DatabaseSeedController.java
+++ b/server/app/controllers/dev/DatabaseSeedController.java
@@ -38,11 +38,13 @@ import services.question.types.NameQuestionDefinition;
 import services.question.types.QuestionDefinition;
 import services.question.types.RadioButtonQuestionDefinition;
 import services.question.types.TextQuestionDefinition;
+import tasks.DatabaseSeedTask;
 import views.dev.DatabaseSeedView;
 
 /** Controller for seeding the database with test content to develop against. */
 public class DatabaseSeedController extends DevController {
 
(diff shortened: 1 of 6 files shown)


server/app/controllers/dev/DatabaseSeedController.java
```diff
@@ -78,9 +82,20 @@ public Result index(Request request) {
             request, activeAndDraftPrograms, questionDefinitions, request.flash().get("success")));
   }
 
+  public Result seedCanonicalQuestions() {
```
clouser2, line 85: High-level: Can we put this logic into `seed` since we want these questions to be around even for dev instances?

If I understand correctly, that would mean we wouldn't need to call a separate URL from browser tests. I believe we call `/dev/seed` every time we start a Jest run.

gwendolyngoetz (approved):

server/app/controllers/dev/DatabaseSeedController.java
bion: We do visit that URL at the beginning of the run, but only to call `clear`, the browser tests don't use the seed command.

I added the canonical questions to the existing seed action anyway since it doesn't hurt and we generally do want to ensure they're there.

gwendolyngoetz (approved):

PR #2717, opened 2022-06-14
Moved from Cicirello to Codecov for badge display

### LINKED ISSUE #1934: CiviForm has unknown unit test coverage

swatkat1:
### Description
Unless we commit the badge using GitActions, the Cicirello is unable to find the badge to be displayed in README. 
I have investigated the codecov and with basic functionalities, we can upload our coverage.xml and get the codecov badge displayed in our README.

- Extended the README -Added Codecov coverage badge

### Issue(s)
Fixes #1934


.github/workflows/tests.yaml
```diff
@@ -28,11 +28,11 @@ jobs:
       run: docker build -t civiform-dev --cache-from docker.io/civiform/civiform-dev:latest ./
     - name: Run tests
       run: USE_LOCAL_CIVIFORM=true bin/run-test-ci
-    - name: jacoco-badge-generator
-      uses: cicirello/jacoco-badge-generator@v2
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
```
clouser2, line 32: I believe there was discussion around not using the CodeCov tool. @goldblatt should have more context. Can you confirm that there aren't any concerns there?

I'll review the rest of this PR as if that's already been figured out.

clouser2 (changes requested):
Using "request changes" to signify we should address the top-level comment.

.github/workflows/tests.yaml
swatkat1: She said yes to using Codecov in the last meeting. I will check with her once more.

swatkat1: She said yes to using Codecov in the last meeting. I will check with her once more.

README.md
```diff
@@ -1,6 +1,7 @@
 ![CI status](https://github.com/seattle-uat/civiform/actions/workflows/push.yaml/badge.svg)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/6008/badge)](https://bestpractices.coreinfrastructure.org/projects/6008)
 [![Seattle Staging Deploy](https://github.com/seattle-uat/civiform-deploy/actions/workflows/deploy-staging.yml/badge.svg?branch=main)](https://github.com/seattle-uat/civiform-deploy/actions/workflows/deploy-staging.yml)
+[![coverage](https://codecov.io/gh/seattle-uat/civiform/branch/main/graph/badge.svg?token=K7RXZGHECF)](https://codecov.io/gh/seattle-uat/civiform)
```
clouser2, line 4: What is the token for? Is it a secret? Can we avoid adding the token?

clouser2 (changes requested):

bion:
If it's a secret it needs to be retired and replaced since it's now leaked
into git history

README.md
swatkat1: Agreed! not needed.. found another way.

.github/workflows/tests.yaml
```diff
@@ -28,11 +28,11 @@ jobs:
       run: docker build -t civiform-dev --cache-from docker.io/civiform/civiform-dev:latest ./
     - name: Run tests
       run: USE_LOCAL_CIVIFORM=true bin/run-test-ci
-    - name: jacoco-badge-generator
-      uses: cicirello/jacoco-badge-generator@v2
+    - name: Upload coverage to Codecov
```
kmcnellis, line 31: This should probably only run on push to main, otherwise we'll have a lot of uploaded reports, potentially conflicting

kmcnellis, line 31: This should probably only run on push to main, otherwise we'll have a lot of uploaded reports, potentially conflicting

kmcnellis (changes requested):

.github/workflows/tests.yaml
kmcnellis, line 32: Why was it bad to save the generated badge to the repository?  It would reduce dependance on third-party tools

kmcnellis (changes requested):

.github/workflows/tests.yaml
swatkat1: No it won't.. It is sorted by branch name.. Also this will generate a report comparing the current branch with Main.. so the engineer would know if his/her coverage is pulling down/up the main coverage numbers

.github/workflows/tests.yaml
swatkat1: Our repo has a lot of protections (like Reviewers approval required) hence this approach did not work.

README.md
clouser2: Blocking: Inlining bion@'s comment:

> If it's a secret it needs to be retired and replaced since it's now leaked
into git history

Assuming it wasn't a secret, then this comment is not blocking.

clouser2 (approved):
Looks good modulo confirmation of using the tool and the single blocking comment I left.

README.md
swatkat1: Regarding your comment on the leak of the Token - this token is not the uploading token (which is a different and needs to be kept secret) this is just the token to access the badge. Also Since we are a public Repo - use of token is not needed for us. I am trying to find more info on how to reset the badge token but so far, it looks like nothing important is lost. I will keep digging.

.github/workflows/tests.yaml
goldblatt: cool ya i think we are okay the fear would be another attack on codecov that would allow the secrets in this particular file to be read. @swatkat1 will you add checking https://about.codecov.io/security-update/ to the oncall rotaiton docs please?

goldblatt (approved):

.github/workflows/tests.yaml
kmcnellis: nice!

kmcnellis (approved):
Let's try to run this without the secrets and then looks good!

PR #2732, opened 2022-06-15
Update README to remove coverage badge

swatkat1:
### Description
Removed the Coverage badge from README as it was not working as intended. Cicirello requires the badge to commited inthe branch using Github action and it get more complicated as our branches are protected and commits will not be pushed without reviewers. Hence reverting this change. I am investigating other tools to show the badge.

### Checklist

### Issue(s)


clouser2 (approved):
Mind updating the description to indicate why we're removing it? Are we intending to readd it to the README at a later point?

swatkat1:
Removed the Coverage badge from README as it was not working as intended. Cicirello requires the badge to commited inthe branch using Github action and it get more complicated as our branches are protected and commits will not be pushed without reviewers. Hence reverting this change. I am investigating other tools to show the badge.

Issue #1743, opened 2021-12-09
CSV extracts have customizable date range and fields


elisekalstad:
Program admins would like to specify the date range for the CSV exports as they batch process applications. It would be useful to have a customizable export function for program admins to select the appropriate date range and columns/fields they'd like to include. 

This will cut down on processing/filtering time. It will allow program admins more control of their experience to align with existing workflows.

clouser2:
I'll take a look at this. Initially, I'm going to focus on customizable date ranges. I'd like to better understand the use case for customizable fields before pursuing that too far. As I understand, customizable date ranges is more important.

Once I've investigated the export code a bit further, I'll provide a time estimate.

bion:
@clouser2 

1. sounds good!
2. also sounds reasonable to me. another option would be to have the button open a modal, which we already have decent support for. could potentially be less work than creating a new page entirely. i'd defer to you for what makes sense for implementation efficiency and @rahaghas for UX though.

rahaghas:
Here's my thinking:

2 buttons: 1 is a date filter, which filters the list of applications. Another is a button that says "Download applications". When that second button is clicked, you're offered options. CSV or JSON? All applications or just the ones between the date ranges you've currently selected?

Let me know if this sounds feasible and I can update this thread with a mock.

clouser2:
> Here's my thinking:
> 
> 2 buttons: 1 is a date filter, which filters the list of applications. Another is a button that says "Download applications". When that second button is clicked, you're offered options. CSV or JSON? All applications or just the ones between the date ranges you've currently selected?
> 
> Let me know if this sounds feasible and I can update this thread with a mock.

For the demography export, the program admin is looking at the list of current programs rather than the list of applications. I'm not sure that the program admin will be able to actually view applications since the demography report is meant to contain aggregate data. As such, I don't think the above suggestion would be feasible (for demography exports).

For non-demography exports (e.g. viewed through the Applications UI for Program Admins), something like this may be workable. That said, it would involve a bit of server-side refactoring.

bion:
Generally speaking, short of forbidding it altogether we want to guide the admin away from downloading an unbounded collection since the size is causing the download to sometimes time out on prod

clouser2:
[...]

PR #2734, opened 2022-06-15
Ensure CSV exports write the header row even if there are no entries. Clean up internal implementation.

clouser2:
### Description
Noticed as part of #1743. The interface for exporting CSVs didn't have clear separation of concerns or object lifecycle. We had logic to write the header row when writing the first record. While here, I took the opportunity to clean up the interface a bit:
  * Removed ExporterFactory, since it only has a single callsite
  * A `CSVExporter` wraps a `CSVPrinter` and implements `AutoClosable`. This means that the header row is written as part of the constructor (similar to the interface of `CSVPrinter`). This also means that we no longer have to pass a `Writer` to every invocation of a record, hopefully making the lifecycle clearer. 

### Checklist

### Issue(s)
 #1743


clouser2:
Note: I expect the formatter to fail since it presently has a bug where it runs on deleted files. Sending a PR out for that shortly.

server/app/services/export/CsvExporter.java
```diff
@@ -22,59 +21,39 @@
  * {@link Path} indexing into an applicant's data, and CsvExporter takes the path and reads the
  * answer from {@link ReadOnlyApplicantProgramService} if present.
  */
-public class CsvExporter {
-  public static final CSVFormat DEFAULT_CSV_FORMAT =
```
clouser2, line 26: Note to the reviewer: This was unnecessarily shared between the implementation and the test.

server/app/services/export/CsvExporter.java
```diff
@@ -22,59 +21,39 @@
...
-      CSVFormat.DEFAULT.builder().setHeader().setSkipHeaderRecord(true).build();
-
+public final class CsvExporter implements AutoCloseable {
```
bion, line 24: suggestion (blocking): update javadoc to mention the need to close

server/app/services/export/CsvExporter.java
```diff
@@ -22,59 +21,39 @@
...
+    this.printer = new CSVPrinter(writer, format);
   }
 
   /**
-   * The CSV exporter will write the headers on first call to services.export(). It does not store
-   * the writer between calls. Since it is intended for many applications, this function is intended
-   * to be called several times.
+   * The CSV exporter does not store the writer between calls. Since it is intended for many
+   * applications, this function is intended to be called several times.
    */
```
bion, line 48: suggestion (blocking): the comment for this method beginning with "The CSV exporter does not store the writer between calls." is a little strange now that the method doesn't take a writer. Please reword or remove

server/app/services/export/ExporterService.java
```diff
@@ -127,49 +127,49 @@ private CsvExportConfig generateDefaultCsvExportConfig(
...
-        .anyMatch(exportDefinition -> exportDefinition.csvConfig().isPresent())) {
-      csvExporter = exporterFactory.csvExporter(program.toProgram());
-    } else {
-      csvExporter = exporterFactory.csvExporter(generateDefaultCsvConfig(programId));
-    }
+    Optional<CsvExportConfig> maybeExportConfig =
+        program.exportDefinitions().stream()
+            .filter(exportDefinition -> exportDefinition.csvConfig().isPresent())
+            .map(exportDefinition -> exportDefinition.csvConfig().get())
+            .findFirst();
```
bion, line 134: suggestion (non-blocking): i'm pretty sure that admin-specified, per-program export configuration was a planned feature that never got implemented. could remove this logic altogether if we wanted i think, though that may be refactor outside the scope of this PR

bion (changes requested):

bion (approved):

server/app/services/export/CsvExporter.java
```diff
@@ -21,60 +21,40 @@
  * CsvExporter takes a list of {@link Column}s and exports the data specified. A column contains a
  * {@link Path} indexing into an applicant's data, and CsvExporter takes the path and reads the
  * answer from {@link ReadOnlyApplicantProgramService} if present.
+ *
+ * <p>The close method should be called when complete in order for the underlying {@link CsvPrinter}
```
kmcnellis, line 25: According to the java docs, close is called automatically if you have a try-with-resources.  This comment seems a bit incorrect.  Could you modify it to say you need to either close() or  try-with-resources?

kmcnellis (approved):

PR #2737, opened 2022-06-15
Browser tests batch script doesn't fail if the subprocess fails

gwendolyngoetz:
The ci browser tests batch script is failing silently. The python script always returns 0 so it doesn't trigger the failure on the github action step.

[Example](https://github.com/seattle-uat/civiform/runs/6905881953) see "run_all_tests / run_browser_tests_aws (4)"

[Python docs](https://docs.python.org/3/library/subprocess.html) indicate that the check parameter will cause the subprocess to throw an exception if the exit code is not 0 which in turn has the main process fail with exit code 1.

### DIFF
--- a/bin/run-browser-tests-ci-batch
+++ b/bin/run-browser-tests-ci-batch
@@ -41,4 +41,4 @@ for item in batch_contents:
 
 shell_command = ["./bin/run-browser-tests-ci"] + batch_contents
 
-subprocess.run(shell_command)
+subprocess.run(shell_command, check=True)
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -36,6 +36,7 @@ jobs:
 
   run_browser_tests_aws:
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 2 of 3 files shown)


bin/run-browser-tests-ci-batch
```diff
@@ -41,4 +41,4 @@ for item in batch_contents:
 
 shell_command = ["./bin/run-browser-tests-ci"] + batch_contents
 
-subprocess.run(shell_command)
+subprocess.run(shell_command, check=True)
```
clouser2, line 44: Not blocking: I believe this can be replaced with the equivalent 
 `subprocess.check_call(shell_command)` (https://docs.python.org/3/library/subprocess.html#subprocess.check_call).

clouser2 (approved):

bin/run-browser-tests-ci-batch
gwendolyngoetz: The docs say that run is the newer and preferred method. https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

bion (approved):

PR #2740, opened 2022-06-15
Rewrite deploy script for aws in python

azizova-leyla:
cloud/aws/bin/deploy -> cloud/aws/bin/deploy.py
cloud/shared/bin/lib/terraform -> cloud/shared/bin/lib/terraform.py


cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
+#! /usr/bin/env python3
+import subprocess
```
clouser2, line 2: Do you mind updating the PR description to indicate where the original versions of the bash script can be found (to make it easier to compare)?

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
...
+sys.path.append(os.getcwd())
+
+from cloud.shared.bin.lib import civiform_mode
+from cloud.shared.bin.lib import terraform
+
+# Keep in sync with cloud/shared/bin/lib.sh
+os.environ['TF_VAR_FILENAME'] = 'setup.auto.tfvars'
+os.environ['BACKEND_VARS_FILENAME'] = 'backend_vars'
+
+subprocess.check_call(['cloud/shared/bin/lib/tf_apply_setup.py'])
```
clouser2, line 15: No changes requested, just noting for posterity:

While `.run()` is apparently the preferred way to call `subprocess` nowadays. That said, it defaults the `check` arg to False, which feels like a pretty unsafe default to me.

Personally, I prefer `check_call` since it's more explicit about what's happening.

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
...
+
+if civiform_mode.is_dev():
+    terraform.copy_backend_override()
+
+if not terraform.perform_apply():
+    print('Terraform deployment failed.')
```
clouser2, line 21: The previous version was calling out::error. Did this write to stderr?

If so, I'd suggest doing an `import sys` above and `sys.stderr.write(...)`. An alternative would be:

`print('...', file=sys.stderr)`

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
...
+    # TODO: write and upload logs.
```
clouser2, line 22: If possible, I'd suggest using an issue number associated with the TODO. E.g. `TODO(#1234)`

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
...
+    raise 1
```
clouser2, line 23: I'd raise an actual exception here like:

`raise ValueError('Terraform deployment failed.')`

If we raised an error, it'd be more Pythonic and likely would remove the need for the print statement above (but you'd probably still want deploy logging).

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,26 @@
+#! /usr/bin/env python3
+import subprocess
+import os
+import sys
+
+sys.path.append(os.getcwd())
```
clouser2, line 6: Can this be modified by an upstream caller?

cloud/shared/bin/deploy
```diff
@@ -12,4 +12,10 @@ if [[ "${IMAGE_TAG}" == "latest" ]]; then
 fi
 
 export TF_VAR_image_tag="${IMAGE_TAG}"
-exec "cloud/${CIVIFORM_CLOUD_PROVIDER}/bin/deploy" "$@"
+if [[ -f "cloud/${CIVIFORM_CLOUD_PROVIDER}/bin/deploy.py" ]]; then
```
clouser2, line 15: Is this fork necessary? Post-this PR, the Python script will always be present. I'd suggest also deleting the bash version here as well, since it wouldn't need to be used anymore.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
+import subprocess
+import os
+import shutil
+
+from cloud.shared.bin.lib import civiform_mode
+
+
+def perform_apply():
+    '''Generates terraform variable files and runs terraform init and apply.'''
```
clouser2, line 9: The original version claims that it initializes a storage bucket as well. Was that comment just incorrect or do we need to also move it here?

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+
+    TERRAFORM_PLAN_OUT_FILE = "terraform_plan"
+    TERRAFORM_TEMPLATE_DIR = os.getenv("TERRAFORM_TEMPLATE_DIR")
```
clouser2, line 12: For these three variables, there's no need to use UPPER_CASE. Instead we can use the Python default lower_underscore_case.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+import os
+import shutil
+
+from cloud.shared.bin.lib import civiform_mode
+
+
+def perform_apply():
+    '''Generates terraform variable files and runs terraform init and apply.'''
+
+    TERRAFORM_PLAN_OUT_FILE = "terraform_plan"
```
clouser2, line 11: Style nit: Use single quotes in order to be internally consistent within this file.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+    TERRAFORM_TEMPLATE_DIR = os.getenv("TERRAFORM_TEMPLATE_DIR")
+    TF_VARS_FILENAME = os.getenv('TF_VAR_FILENAME')
+
+    terraform_cmd = ['terraform', f'-chdir={TERRAFORM_TEMPLATE_DIR}']
+
+    if civiform_mode.is_dev():
+        subprocess.check_call(
+            terraform_cmd + ['init', '-upgrade', '-reconfigure'])
```
clouser2, line 19: Not blocking: Rather than using lists here, I'd consider specifying this as a single string and using `shlex.split`.

Concretely, this one would be something like:

```
subprocess.check_call(shlex.join(f'{terraform_cmd} init -upgrade -reconfigure'))
```

That said, if you think the list approach feels more workable, I'm also fine leaving as-is.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
```
clouser2, line 20: Was the call to setup_tf_shared_state that was in the original version now unnecessary?

If we're going to eventually use this for Azure deployments as well, I'd leave a TODO.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+                'init', '-input=false', '-upgrade',
+                f'-backend-config={os.getenv("BACKEND_VARS_FILENAME")}'
+            ])
+
+    if os.path.exists(os.path.join(TERRAFORM_TEMPLATE_DIR, TF_VARS_FILENAME)):
+        print(
+            f'{TF_VARS_FILENAME} exists in {TERRAFORM_TEMPLATE_DIR} directory')
+    else:
+        print(
+            f'Aborting the script. {TF_VARS_FILENAME} does not exists in {TERRAFORM_TEMPLATE_DIR} directory'
```
clouser2, line 32: Minor typo: s/exists/exist/

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        raise 1
```
clouser2, line 34: I'd remove the print statement and instead raise an error with this message.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+        ])
+
+    if civiform_mode.is_test():
+        return True
+
+    terraform_apply_cmd = terraform_cmd + ['apply', '-input=false', '-json']
+    if not civiform_mode.is_dev():
+        subprocess.check_call(
+            terraform_apply_cmd +
+            ['-auto-approve', f'{TERRAFORM_PLAN_OUT_FILE}'])
```
clouser2, line 49: No need for an f-string here and in a few other places. The underlying variable is already a string.

cloud/shared/bin/lib/terraform.py
```diff
@@ -0,0 +1,71 @@
...
+        terraform_cmd + [
+            'plan', '-input=false', f'-out={TERRAFORM_PLAN_OUT_FILE}',
+            f'-var-file={os.getenv("TF_VAR_FILENAME")}'
+        ])
+
+    if civiform_mode.is_test():
+        return True
+
+    terraform_apply_cmd = terraform_cmd + ['apply', '-input=false', '-json']
+    if not civiform_mode.is_dev():
```
clouser2, line 46: Consider flipping the conditionals since we already have an else statement.

clouser2 (changes requested):

cloud/aws/bin/deploy.py
azizova-leyla: Right now this is the first python script that will get run. we should move that into deploy-specific deploy script once we have those. Added a todo.

cloud/aws/bin/deploy.py
azizova-leyla: done

cloud/aws/bin/deploy.py
azizova-leyla: done

cloud/shared/bin/deploy
azizova-leyla: this is shared script, it is invoked for both azure and aws. Added todo to cleanup.

cloud/shared/bin/lib/terraform.py
azizova-leyla: we're setting it up as part of setup for aws, and should do the same for azure. I don't think it should be in this script at all. Added a comment so we don't forget.

cloud/shared/bin/lib/terraform.py
azizova-leyla: That's the bucket that was azure specific. See todo on top.

cloud/aws/bin/deploy.py
clouser2: Sounds good. One last request for posterity. Mind also adding a brief comment as to why it's necessary to do this in the middle of import statements?

Potential example:
```
# We modify the system path here so that the imports below can be resolved.
```

clouser2: Sounds good. One last request for posterity. Mind also adding a brief comment as to why it's necessary to do this in the middle of import statements?

Potential example:
```
# We modify the system path here so that the imports below can be resolved.
```

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,27 @@
...
+import sys
+
+# TODO(#2743): move this to deploy specific script.
+sys.path.append(os.getcwd())
+
+from cloud.shared.bin.lib import civiform_mode
+from cloud.shared.bin.lib import terraform
+
+# Keep in sync with cloud/shared/bin/lib.sh
+os.environ['TF_VAR_FILENAME'] = 'setup.auto.tfvars'
```
clouser2, line 13: On second glance, we might be able to do away with the environment variable setting entirely if downstream scripts are written in Python. The reason being that Python scripts could just encode these in helper functions / packages.

I think we're likely not there yet, but just noting that we can hopefully get rid of these in the future.

cloud/shared/bin/deploy
clouser2: Gotcha. Thanks for the context :)

clouser2 (approved):

goldblatt (changes requested):

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,28 @@
...
+
+# We modify the system path here so that the imports below can be resolved.
+# TODO(#2743): move this to deploy specific script.
+sys.path.append(os.getcwd())
+
+from cloud.shared.bin.lib import civiform_mode
+from cloud.shared.bin.lib import terraform
+
+# Keep in sync with cloud/shared/bin/lib.sh
+os.environ['TF_VAR_FILENAME'] = 'setup.auto.tfvars'
```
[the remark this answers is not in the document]
goldblatt: ya would recommend using the config loader class here

cloud/aws/bin/deploy.py
```diff
@@ -0,0 +1,28 @@
...
+os.environ['BACKEND_VARS_FILENAME'] = 'backend_vars'
+
+subprocess.check_call(['cloud/shared/bin/lib/tf_apply_setup.py'])
+
+if civiform_mode.is_dev():
+    terraform.copy_backend_override()
+
+if not terraform.perform_apply():
```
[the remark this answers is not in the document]
azizova-leyla: I don't see us using it anywhere

cloud/shared/bin/lib/tf_apply_setup.py
```diff
@@ -11,20 +11,24 @@
...
+            f"Found the following validation errors: {new_line}{f'{new_line}'.join(validation_errors)}"
+        )
 
-# Write the passthrough vars to a temporary file
-tf_var_writter = TfVarWriter(terraform_tfvars_filename)
-variables_to_write = config_loader.get_terraform_variables()
-tf_var_writter.write_variables(variables_to_write)
+    template_dir = config_loader.get_template_dir()
+
+    terraform_tfvars_filename = f"{template_dir}/setup.auto.tfvars"
```
goldblatt, line 28: we probably should os.join this

cloud/shared/bin/lib/run_tf_apply_setup.py
```diff
@@ -0,0 +1,5 @@
+#! /usr/bin/env python3
+
+from cloud.shared.bin.lib.tf_apply_setup import load_config
+
+load_config()
```
goldblatt, line 5: this naming feels a lil weird naming wise

goldblatt (approved):

PR #2749, opened 2022-06-16
Add date range filtering to the demographics export

clouser2:
Since this report is slow to generate, adding time filters will help admins pare down the returned data to only that which is needed. Follow-on work will be to add date range filtering to the program-specific CSV exports.

Since the demography export button is shown on the program list (not application list), clicking the button now pops up a modal with the filters available ([discussion](https://github.com/seattle-uat/civiform/issues/1743#issuecomment-1156731269)).

Screen Shot 2022-06-16 at 3 08 32 PM

Note: I didn't add browser tests for this since we don't presently have the ability to generate fake applications at specific times.

### Checklist

### Issue(s)
#1743


server/app/controllers/admin/BadRequestException.java
```diff
@@ -0,0 +1,12 @@
+package controllers.admin;
+
+/**
+ * Thrown when a request is malformed. This class is final to ensure that it works with {@link
+ * controllers.ErrorHandler} and subclasses aren't thrown with the same expected behavior.
+ */
+public final class BadRequestException extends RuntimeException {
```
bion, line 7: suggestion (blocking): how about moving https://github.com/seattle-uat/civiform/blob/main/server/app/controllers/api/BadApiRequestException.java#L8-L13 out of the api package and using it here as well?

server/app/repository/TimeFilter.java
```diff
@@ -0,0 +1,26 @@
+package repository;
+
+import com.google.auto.value.AutoValue;
+import java.time.Instant;
+import java.util.Optional;
+
+/** Allows filtering data based on an optional fromTime (inclusive) / toTime (exclusive).ß */
```
bion, line 7:
```suggestion
/** Allows filtering data based on an optional fromTime (inclusive) / toTime (exclusive). */
```

server/conf/routes
```diff
@@ -83,7 +83,7 @@ GET     /admin/programs/:programId/applications/all                       contro
 GET     /admin/programs/:programId/applications/singleVersion             controllers.admin.AdminApplicationController.downloadSingleVersion(request: Request, programId: Long)
 GET     /admin/programs/:programId/applications/:applicationId            controllers.admin.AdminApplicationController.show(request: Request, programId: Long, applicationId: Long)
 GET     /admin/programs/:programId/applications/:applicationId/download   controllers.admin.AdminApplicationController.download(request: Request, programId: Long, applicationId: Long)
-GET     /admin/demographics                                               controllers.admin.AdminApplicationController.downloadDemographics()
+GET     /admin/demographics                                               controllers.admin.AdminApplicationController.downloadDemographics(request: Request)
```
bion, line 86: suggestion (blocking): you can specify optional query params here with `java.util.Optional[Integer]` i.e.
```suggestion
GET     /admin/demographics                                               controllers.admin.AdminApplicationController.downloadDemographics(request: Request, from_date: java.util.Optional[String], to_date: java.util.Optional[String])
```

server/app/models/Application.java
```diff
@@ -108,4 +109,10 @@ public Application setSubmitTimeToNow() {
     this.submitTime = Instant.now();
     return this;
   }
+
+  @VisibleForTesting
+  public Application setSubmitTimeForTest(Instant v) {
+    this.submitTime = v;
+    return this;
+  }
```
bion, line 117: suggestion (blocking): there's test support code [here](https://github.com/seattle-uat/civiform/blob/main/server/test/support/CfTestHelpers.java#L14-L26) for mocking the return value of `Instant.now()` so you can use `setSubmitTimeToNow`

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -143,15 +150,34 @@ public Result downloadSingleVersion(Http.Request request, long programId)
...
    * Demographics information is collected from answers to a collection of questions specially
    * marked by CiviForm admins.
    */
   @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
-  public Result downloadDemographics() {
+  public Result downloadDemographics(Http.Request request) {
+    TimeFilter submitTimeFilter =
+        TimeFilter.builder()
+            .setFromTime(parseDateFromQuery(request, Filters.FROM_DATE_QUERY_PARAM))
+            .setToTime(parseDateFromQuery(request, Filters.TO_DATE_QUERY_PARAM))
```
bion, line 177: suggestion (blocking): making the query params part of the method signature rather than pulling them off the `request` object makes it a little easier to see what the HTTP interface is. see comment on the `routes` file and [example usage](https://github.com/seattle-uat/civiform/blob/main/server/app/controllers/api/ProgramApplicationsApiController.java#L58-L64)

server/app/controllers/admin/Filters.java
```diff
@@ -0,0 +1,7 @@
+package controllers.admin;
+
+/** Class to hold application filter constants shared between different views. */
+public final class Filters {
```
bion, line 4: suggestion (blocking): i don't think this class will be necessary if query params are specified in the route and method signature for the action

bion (changes requested):

server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -89,24 +96,55 @@ public Content render(
...
+
+    ContainerTag downloadDemographicCsvModalContent =
+        div()
+            .withClasses(Styles.PX_8)
+            .with(
+                form()
+                    .withMethod("GET")
+                    .withAction(routes.AdminApplicationController.downloadDemographics().url())
+                    .with(
+                        p("This will download all applications for all programs and can take"
```
azizova-leyla, line 123: this is a bit hard to read. Rephrase into something like:
"This will download all applications for all programs. Use filters below to select date range for the data. Warning: if you select a large date range or leave it blank it might take up to (10? 30?) minutes to export data."

server/test/repository/ApplicationRepositoryTest.java
```diff
@@ -142,6 +145,65 @@ public void submitApplication_noDrafts() {
...
+    Applicant applicant = saveApplicant("Alice");
+    Application app =
+        repo.submitApplication(applicant, program, Optional.empty()).toCompletableFuture().join();
+    app.refresh();
+    app.setSubmitTimeForTest(submitTime).save();
+    return app;
+  }
+
+  @Test
+  public void getApplications() {
```
azizova-leyla, line 160: can you add a test where there are 0 applications after the filter

azizova-leyla (changes requested):

bion (approved):
Brett I often admire how you organize code and the care you take for separation of concerns. This PR being another example!

clouser2:
> Brett I often admire how you organize code and the care you take for separation of concerns. This PR being another example!

:). Thank you for the kind words!

azizova-leyla (approved):

PR #2755, opened 2022-06-16
Fixing pagination rendering when there are no pages (previously was "Page 1 of 0")

clouser2:
### Description
Noticed this while investigating adding date filtering to the existing Program Admin applications view. I could also see a case for rendering "Page 0 of 0" and rendering "No results" further down. If there's a strong preference, I can convert to that.

Also, I looked into making the buttons show as disabled when there were no results since they currently give the illusion that clicking them will do something. I didn't pursue that since the styles being used are "private" to LinkElement and I didn't have a good sense of how / where we'd want to break those out. As it stands, this currently seems like a small usability improvement.

Before:
Screen Shot 2022-06-16 at 4 11 27 PM

After: 
Screen Shot 2022-06-16 at 4 11 11 PM

### Issue(s)
#1743

### DIFF
--- a/server/app/views/BaseHtmlView.java
+++ b/server/app/views/BaseHtmlView.java
@@ -99,8 +99,10 @@ public abstract class BaseHtmlView {
       div.with(
           new LinkElement().setText("←").setHref(linkForPage.apply(page - 1).url()).asButton());
     }
+    String paginationText =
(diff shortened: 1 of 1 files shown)


bion (approved):
good catch!

PR #2770, opened 2022-06-21
Add Status data model, serialization

### LINKED ISSUE #2902: Status Tracking: Implement Status Definition data model and database field
For Status Tracking we need the java data classes and to add them to the Program database table.

shanemc-goog:
### Description

Data modeling for Status Tracking configuration.

Note this is the data the CiviForm Admin will specify and Program Admins will use.  This won't be the modeling for how the specified Application Status is recorded, but this will flow into that.

- Adds `Status` a single Status definition and `StatusDefinitions` a wrapper for multiple Statuses and other future information.
- Adds `StatusDefinitions` as a column/Field in Program.
- Evolution adds `StatusDefinitions` including a default of an empty object `{"statuses": []}`
  For verification, when a new program is created the no-status data is represented in the db as `{"statuses": []}`
- Serialization works and looks like
``` 
{"statuses": [
    {"status": "Approved",
      "email_body": "I'm an email!",
       "status_localized": {
           "isRequired": true,
           "translations": {"en_US": "Approved"}},
      "email_body_localized": {
          "isRequired": true,
          "translations": {"en_US": "I'm an email!"}}
    }
]}
```

## Release notes:
Adds Program.status_definitions Database modeling for upcoming Status Tracking feature

### Checklist

### Issue(s)
Fixes #2902


shanemc-goog:
@bion I'd appreciate an eye on the db setup modeling especially since it's hard to correct any issues with that after the fact.

server/app/services/program/StatusDefinitions.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+import services.LocalizedStrings;
+
+public class StatusDefinitions {
```
clouser2, line 11: Please add a JavaDoc describing the use and purpose.

clouser2, line 11: Consider marking this class as final to prevent subclassing (or at least make the decision to allow subclassing explicit in the future when needed).

server/app/services/program/StatusDefinitions.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @JsonProperty("statuses")
+  private ImmutableList<Status> statuses;
+
+  @JsonCreator
+  public StatusDefinitions(@JsonProperty("statuses") ImmutableList<Status> statuses) {
+    this.statuses = ImmutableList.copyOf(statuses);
+  }
+
+  public StatusDefinitions() {
```
clouser2, line 21: Not blocking: Is a default constructor necessary? If not, I'd just err on the side of having callers specify an empty list explicitly (my rationale is that this would be less action-at-a-distance of what values the class is initialized with).

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -207,6 +207,7 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
             .build()
             .toProgram();
     return ErrorAnd.of(
+        // Note: This method seems heavy-handed for non question updates.
```
clouser2, line 210: Nit: I'd consider removing this comment since there's nothing actionable from it. If this is something we should fix or an idea for a future improvement, we should either 1) file an issue + TODO or 2) expand the comment to include a potential solution or time when we should fix.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -345,6 +346,31 @@ public ProgramDefinition moveBlock(
         .join();
   }
 
+  @Override
+  @Transactional
+  public StatusDefinitions getStatuses(long programId) throws ProgramNotFoundException {
```
clouser2, line 351: ProgramDefinition already contains the status information. Given that, wouldn't a caller likely use one of the overloads of `getProgramDefinition`?

clouser2 (changes requested):

server/app/services/program/StatusDefinitions.java
```diff
@@ -0,0 +1,71 @@
...
+import services.LocalizedStrings;
+
+public class StatusDefinitions {
+
+  @JsonProperty("statuses")
+  private ImmutableList<Status> statuses;
+
+  @JsonCreator
+  public StatusDefinitions(@JsonProperty("statuses") ImmutableList<Status> statuses) {
+    this.statuses = ImmutableList.copyOf(statuses);
```
bion, line 18: suggestion (blocking): since the argument is immutable there's no need to copy it

bion (changes requested):

server/app/services/program/ProgramServiceImpl.java
shanemc-goog: Yeah... There's so much stuff like this given the rapid development that a TODO for all of it feels like too much too :D

I removed the usage comments and kept the method one.

server/app/services/program/StatusDefinitions.java
shanemc-goog: Making it final gives a warning about needing toString. Not exactly sure what that's about, but not worth the time right now.  We don't have a pattern of making these final so may want to do that enmasse

shanemc-goog: done

server/app/services/program/StatusDefinitions.java
shanemc-goog: The definition allows for it not having set data (eg statuses) so I think an empty constructor that affords that fits.

server/app/services/program/ProgramServiceImpl.java
shanemc-goog: Ah thanks, removed.

server/app/services/program/StatusDefinitions.java
shanemc-goog: done

server/test/services/program/ProgramServiceImplTest.java
```diff
@@ -1160,4 +1161,45 @@ public void getProgramDefinitionAsync_reordersBlocksOnRead() throws Exception {
...
+            .setLocalizedEmailBodyText(LocalizedStrings.of(Locale.US, "I'm a US email!"))
+            .build();
+
+    ErrorAnd<ProgramDefinition, CiviFormError> result =
+        ps.setStatuses(program.id, new StatusDefinitions(ImmutableList.of(status)));
+
+    assertThat(result.isError()).isFalse();
+    StatusDefinitions gotStatusDef = result.getResult().statusDefinitions();
+    assertThat(gotStatusDef.getStatuses()).containsExactly(status);
+  }
```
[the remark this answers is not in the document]
shanemc-goog: yessss, I didn't know we had that level of testing, added.

bion (approved):

server/app/services/program/StatusDefinitions.java
clouser2: My preference is weak. I generally prefer being more verbose at initialization rather than having to click-through to see what the default would be. My style probably is a little bit less Java-ey though :).

clouser2 (approved):

PR #2772, opened 2022-06-21
Write TFVars before setup script

goldblatt:
### Description
Making some fixes to the tfvars script so that we create it before we do any setup script (AWS setupscript uses terraform). Also grabs the backend_override so that the dev upload works

### Checklist

### Issue(s)

### DIFF
--- a/cloud/shared/bin/lib/setup.py
+++ b/cloud/shared/bin/lib/setup.py
@@ -52,13 +52,13 @@ image_tag = config_loader.get_config_var("IMAGE_TAG")
 log_args = f"\"{image_tag}\" {current_user}"
 
 print("Writing TF Vars file")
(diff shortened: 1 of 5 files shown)


cloud/shared/bin/lib/setup.py
```diff
@@ -51,6 +51,15 @@
 image_tag = config_loader.get_config_var("IMAGE_TAG")
 log_args = f"\"{image_tag}\" {current_user}"
 
+print("Writing TF Vars file")
+terraform_tfvars_path = f"{template_dir}/{config_loader.tfvars_filename}"
```
azizova-leyla, line 55: os.path.join

azizova-leyla (approved):

PR #2775, opened 2022-06-22
Add script for creating releases

bion:
Adds a bin script for creating new releases. See https://docs.google.com/document/d/1ELRj03KzsXzp1ezzuD3c03eZsQiLOjERDUxp5zbFuMQ/edit?resourcekey=0-cmpNQbrx_PjLHFAZKpXVmA#heading=h.64x49zep4dz for context


.gitignore
```diff
@@ -45,3 +45,4 @@ test-support/sbt_cache/
 **/.ammonite/
 **/metals.sbt
 **/pom.xml
+/gh-release-token.txt
```
clouser2, line 48: Maybe a more general **/gh-release-token.txt to prevent us from including this file from ANY directory.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+# Requirements:
+#   Must have creds to login to docker hub as "civiform" user.
+#   Must have a GitHub API token with repo and read:org scopes located at "gh-release-token.txt"
+#
+# When successful, this script:
+#   1. tags the git commit for the release with the release number
+#   2. tags the docker image for the release with the release number
+#   3. creates a new draft release on github
+
+if not os.path.exists("gh-release-token.txt"):
```
clouser2, line 21: Super style nit: Consider using single quotes throughout. There's no true recommendation either way, but empirically from the Python code I've seen in the wild, single-quotes tend to be more prevalent.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+    print("gh-release-token.txt not found in root directory", file=sys.stderr)
```
clouser2, line 22: Not blocking: Consider using a multi-line string and combining the print statements:

```
print('''
gh-release-token.txt not found in root directory
Create one with repo....
''')
```

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+
+if len(sys.argv) != 3:
+    print(
+        "Usage: bin/create-release COMMIT_SHA RELEASE_NUMBER", file=sys.stderr)
+    exit(1)
+
+
+def shell_cmd(command_string):
+    try:
+        process = subprocess.run(
```
clouser2, line 36: Blocking: Either specify the `check=True` param here or use `subprocess.check_call()`. This will validate that the called process exited non-0.

The Python docs claim that `subprocess.run()` is the preferred method to call, but the default value for `check` is False. Personally, I think that's a pretty strong reason NOT to use `subprocess.run()` :).

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+            shlex.split(command_string), capture_output=True)
+    except FileNotFoundError:
```
clouser2, line 38: Not blocking: When is this error raised?

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+with open("gh-release-token.txt", "r") as file:
+    GH_TOKEN = file.read().rstrip()
+
+if not GH_TOKEN.startswith("ghp_"):
+    print("Invalid token found in gh-release-token.txt", file=sys.stderr)
+    exit(1)
+
+check_gh_auth_response = requests.get(
+    "https://api.github.com/repos/seattle-uat/civiform",
+    headers={"Authorization": "token " + GH_TOKEN})
```
clouser2, line 56: Style nit, not blocking: You can use f-strings to interpolate this.

```
headers={'Authorization': f'token {GH_TOKEN}'}
```

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+if check_gh_auth_response.status_code != 200:
+    print(
+        "GitHub API auth failed, satus code: " +
```
clouser2, line 59: Style nit, not blocking: Minor typo of `satus`.

Same comment as above regarding f-strings if you choose to incorporate it.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+if not "Login Succeeded" in shell_cmd(
+        "docker login") or not "civiform" in shell_cmd("docker info"):
+    print("Unable to login to docker hub")
+    exit(1)
+
+################################################################################
+# Set release variables
+################################################################################
+
+COMMIT_SHA = sys.argv[1]
```
clouser2, line 77: Consider initializing these above closer to the place where we check arg length and print usage information.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+        return ''
+    return process.stdout.decode('ascii') if process else ''
+
+
+################################################################################
+# Resolve and validate GitHub token, check auth
+################################################################################
+
+with open("gh-release-token.txt", "r") as file:
+    GH_TOKEN = file.read().rstrip()
```
clouser2, line 48: Not blocking: Any particular reason to use UPPER_CAMEL_CASE here rather than lower_camel_case? If I recall correctly, the former is typically used to represent package-level constants rather than things that are dynamically initialized.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+    exit(1)
+
+################################################################################
+# Validate commit SHA
+################################################################################
+
+branches_containing_commit_sha = shell_cmd(
+    "git --no-pager branch --contains " + COMMIT_SHA)
+branches_containing_commit_sha = [
+    re.sub('\*', '', x).strip()
```
clouser2, line 97: Blocking: This regular expression is non-obvious. I'd add a quick comment indicating sample format to make it easier to determine what this is actually substituting.

bin/create-release
```diff
@@ -0,0 +1,197 @@
...
+# Find snapshot tag that matches commit SHA
+################################################################################
+
+# Git defaults to using the first 7 chars of the full SHA for the short SHA.
+SHORT_SHA = shell_cmd("git rev-parse --short " + COMMIT_SHA)[:7]
+
+print(f"Searching for an image tag with short SHA {SHORT_SHA}")
+
+tags_url = "https://registry.hub.docker.com/v2/repositories/civiform/civiform/tags?order=-last_activity&page_size=100"
+snapshot_tag = None
```
clouser2, line 114: Nit, not blocking: Re-initialize snapshot_tag as the first line in the loop. This will prevent weird bugs where we go through the loop a second time and for some reason use a previous snapshot_tag version.

That may not actually be a problem here, but generally I prefer this type of pattern.

clouser2 (changes requested):
Nice!

bin/create-release
bion: done

bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: oh jeez i should have left a comment since now i can't remember, it came up while i was writing this script

bin/create-release
bion: i was using constant case for every variable that doesn't change after initialization. would you prefer i downcase this and the other vars (e.g. the release number, sha, etc)?

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

.gitignore
bion: done

bin/create-release
clouser2: Definitely up to you, but what about removing it for now? Then, if we find it to be a problem when running the script in the future, we can re-add and capture why it's necessary?

bin/create-release
clouser2: My preference is weak and your rationale makes sense. I'm fine keeping it this way, it was just a new pattern to me.

bin/create-release
clouser2: Thanks!

clouser2 (approved):

.github/release.yml
```diff
@@ -0,0 +1,19 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+      - deploy
+      - infrastructure
```
kmcnellis, line 6: Can we add documentation on these tags to the contributor guide?

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+# Requirements:
+#   Must have creds to login to docker hub as 'civiform' user.
+#   Must have a GitHub API token with repo and read:org scopes located at 'gh-release-token.txt'
+#
+# When successful, this script:
+#   1. tags the git commit for the release with the release number
+#   2. tags the docker image for the release with the release number
+#   3. creates a new draft release on github
+
+if not os.path.exists('gh-release-token.txt'):
```
kmcnellis, line 21: [feature request] Is it possible to run this in a github action so that individuals don't need to create access tokens?

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+    print(
+        '''
+        gh-release-token.txt not found in root directory
```
kmcnellis, line 24: This isn't in the root directory (that's `/`, I believe) - it would be in the top-level directory of the civiform repository

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+COMMIT_SHA = sys.argv[1]
+RELEASE_NUMBER = sys.argv[2]
+RELEASER_NAME = shell_cmd('git config --get user.name')
+RELEASER_EMAIL = shell_cmd('git config --get user.email')
+
+################################################################################
+# Resolve and validate GitHub token, check auth
+################################################################################
+
+with open('gh-release-token.txt', 'r') as file:
```
kmcnellis, line 58: We shoud put this file name in a variable if it's used in multiple spots

bin/create-release
kmcnellis, line 21: Why bother to have an extra "path exists" check instead of just using error handling around the "open" command below?

bin/create-release
kmcnellis: We should follow the style guide for this: https://google.github.io/styleguide/pyguide.html#316-naming

Global/class level constants get CAPS_WITH_UNDER, as do "module-level constants" (https://google.github.io/styleguide/pyguide.html#254-decision)

kmcnellis: I don't believe this counts, in the sense that it's not a fixed value for the module

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+if not GH_TOKEN.startswith('ghp_'):
+    print('Invalid token found in gh-release-token.txt', file=sys.stderr)
+    exit(1)
+
+check_gh_auth_response = requests.get(
+    'https://api.github.com/repos/seattle-uat/civiform',
+    headers={'Authorization': f'token {GH_TOKEN}'})
+if check_gh_auth_response.status_code != 200:
+    print(
+        f'GitHub API auth failed, satus code: {str(check_gh_auth_response.status_code)}',
```
kmcnellis, line 70: spelling. "satus"->"status"

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+        file=sys.stderr)
+    exit(1)
+
+################################################################################
+# Check docker hub auth
+################################################################################
+
+if not 'Login Succeeded' in shell_cmd(
```
kmcnellis, line 78: I'm not a fan of hiding this pair of stateful shell queries inside an "if" statement, but it's personal preference.  I think it's easier to follow the logic if you split them (like "Unable to login" and then "Not logged in as the civiform user")

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+if not os.path.exists('gh-release-token.txt'):
+    print(
+        '''
+        gh-release-token.txt not found in root directory
+        Create one with repo and read:org scopes at https://docs.github.com/en/enterprise-server@3.4/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
+    ''',
+        file=sys.stderr)
+    exit(1)
+
+if len(sys.argv) != 3:
```
kmcnellis, line 30: This should be the first validation performed

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+    print(
+        'Usage: bin/create-release COMMIT_SHA RELEASE_NUMBER', file=sys.stderr)
+    exit(1)
+
+
+def shell_cmd(command_string):
```
kmcnellis, line 36: Please create a main function, so that we're not defining functions inside the flow of logic.

```
def main():
    print("Hello World!")

if __name__ == "__main__":
    main()
```
https://google.github.io/styleguide/pyguide.html#317-main

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+
+if not re.match('^v\d+\.\d+\.\d+$', RELEASE_NUMBER):
+    print('Invalid version number: ' + RELEASE_NUMBER, file=sys.stderr)
+    exit(1)
+
+################################################################################
+# Validate commit SHA
+################################################################################
+
+branches_containing_commit_sha = shell_cmd(
```
kmcnellis, line 95: Can you add a comment explaining why you are inspecting the branches?

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+    'git --no-pager branch --contains ' + COMMIT_SHA)
+branches_containing_commit_sha = [
+    # The output of the command is a list of branch names, with a '*' at the
+    # beginning of the current branch. This regex and the strip transform
+    # the list to only branch names with no trailing or preceding whitespace.
+    re.sub('\*', '', x).strip()
+    for x in branches_containing_commit_sha.split('\n')
+]
+if not 'main' in branches_containing_commit_sha:
```
kmcnellis, line 104: Really, shouldn't we be verifying that origin/main contains the SHA?

bin/create-release
kmcnellis: Like "Validate commit SHA exists in main branch"

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+    print('Invalid commit SHA: ' + COMMIT_SHA, file=sys.stderr)
+    exit(1)
+
+################################################################################
+# Find snapshot tag that matches commit SHA
```
kmcnellis, line 109: "Find Dockerhub snapshot tag that matches commit SHA"

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+
+    if response.status_code != 200:
+        print(
+            f'Docker Hub request failed with status code: {str(response.status_code)}',
+            file=sys.stderr)
+        exit(2)
+
+    response_json = response.json()
+
+    for tag_info in response_json['results']:
```
kmcnellis, line 130: Is this relying on the github action that publishes images to tag them with the commit sha?  If so, can we document the dependancies? (Maybe an "if change" type block)

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+
+    tags_url = response_json['next']
+
+print(f'Found existing server image: {snapshot_tag}')
+
+################################################################################
+# Add tag to git history
+################################################################################
+
+print('Adding tag to git history')
```
kmcnellis, line 152: What do you mean by adding it to git history?  That wording is confusing to me

bin/create-release
```diff
@@ -0,0 +1,200 @@
...
+    f"git tag -a {RELEASE_NUMBER} -m 'Releaser: {RELEASER_EMAIL}' {COMMIT_SHA}")
+shell_cmd(f'git push origin {RELEASE_NUMBER}')
+
+################################################################################
+# Tag Docker image with release
+################################################################################
+
+print('Tagging Docker image and pushing to Docker Hub')
+
+shell_cmd(f'docker pull civiform/civiform:{snapshot_tag}')
```
kmcnellis, line 164: I really wish we didn't have to do this.  It takes a long time to pull images sometimes.

kmcnellis: Maybe there's some way github actions could do it for us?  Maybe something for a future iteration

kmcnellis (changes requested):

.github/release.yml
bion: yep, will be updating the docs after this PR goes in

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: reworded to 'Adding git tag and pushing to GitHub'

bin/create-release
bion: yeah it's a bummer, i'm not sure of a way around it

bin/create-release
bion: downcased it

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
bion: done

bin/create-release
```diff
@@ -0,0 +1,219 @@
...
+    if not 'origin/main' in branches_containing_commit_sha:
+        print('Invalid commit SHA: ' + commit_sha, file=sys.stderr)
+        exit(1)
+
+    ################################################################################
+    # Find Docker Hub snapshot tag that matches commit SHA
+    ################################################################################
+
+    # Git defaults to using the first 7 chars of the full SHA for the short SHA.
+    SHORT_SHA = shell_cmd('git rev-parse --short ' + commit_sha)[:7]
```
kmcnellis, line 126: nit: lowercase

bin/create-release
```diff
@@ -0,0 +1,219 @@
...
+        headers=create_release_headers)
+    if response.status_code != 201:
+        print(
+            'GitHub request failed with status code: ' +
+            str(response.status_code),
+            file=sys.stderr)
+        print(response.json()['message'], file=sys.stderr)
+        exit(2)
+
+    RELEASE_URL = response.json()['html_url']
```
kmcnellis, line 214: nit: lowercase

kmcnellis (approved):

bin/create-release
bion: done

bin/create-release
bion: done

PR #2777, opened 2022-06-22
Add eslint to catch incorrect javascript patterns

kmcnellis:
### Description
Add eslint to catch potentially-problematic patterns in javascript, using https://github.com/google/eslint-config-google to follow the google js style guice


server/app/assets/javascripts/admin_applications.ts
```diff
@@ -18,7 +18,7 @@ class AdminApplications {
     this.displayFrame = frame
     this.cards = document.querySelectorAll(
       AdminApplications.CARD_SELECTOR,
-    ) as any as Array<HTMLElement>
+    ) as any as Array<HTMLElement> // eslint-disable-line @typescript-eslint/no-explicit-any
```
nbeloglazov, line 21: I think in cases like this we should use `as unknown as Array<HTMLElement>` instead of any. And it will help avoiding eslint error.

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -96,6 +96,7 @@ class AdminPrograms {
 
   static async copyProgramLinkToClipboard(programLink: string) {
     const succeeded = await AdminPrograms.tryCopyToClipboard(programLink)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const toastController = window['toastController'] as any
```
nbeloglazov, line 100: Can we import ToastController and do window['toastController'] as ToastController?

.eslintrc.json
```diff
@@ -0,0 +1,29 @@
...
+  "env": {
+    "browser": true,
+    "commonjs": true
+  },
+  "plugins": ["@typescript-eslint"],
+  "rules": {
+    "require-jsdoc": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "valid-jsdoc": "warn",
+    "@typescript-eslint/no-unused-vars": "off", // overlaps with standard eslint no-unused-vars.
```
nbeloglazov, line 19: We should also enable no-floating-promise rule: https://typescript-eslint.io/rules/no-floating-promises
Maybe in a follow up PR.

We could also try enabling strict checks and see whether it adds a lot of annoying errors.
https://typescript-eslint.io/docs/linting/configs#strict

nbeloglazov (changes requested):
Looks good!

browser-test/src/support/admin_questions.ts
```diff
@@ -33,7 +33,7 @@ export class AdminQuestions {
     await this.expectAdminQuestionsPage()
   }
 
-  async goToViewQuestionPage(questionName: string) {
+  async goToViewQuestionPage(_questionName: string) {
```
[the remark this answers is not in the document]
kmcnellis: Without this change, eslint marks those parameters as unused, and errors out.  The underscore marks it as a "I know it's not used, but please leave it there anyway" parameter

browser-test/src/support/admin_questions.ts
```diff
@@ -344,25 +344,25 @@ export class AdminQuestions {
   }
 
   async updateAllQuestions(questions: string[]) {
-    for (var i in questions) {
+    for (const i in questions) {
```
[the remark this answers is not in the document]
kmcnellis: I am not changing the actual logic at all.  Just fixing linting errors.  I do agree, though!

server/app/assets/javascripts/admin_applications.ts
kmcnellis: Same here.  I'm trying to change as little as possible.  A good follow up to this PR would be to go back and remove these disables wherever possible.

server/app/assets/javascripts/admin_programs.ts
kmcnellis: That is a good suggestion for a future PR

.eslintrc.json
kmcnellis: +1 to doing it after this goes in

nbeloglazov (approved):

PR #2795, opened 2022-06-24
Record api key usage

bion:
We want to keep a count of how many times an API key has been consumed and what the last IP address to authenticate with it was. (see [design](https://docs.google.com/document/d/1g2XkMhDe6EaVGQabEes4KoDWpzL9F4S5pl2fUXoilSc/edit)).

This PR adds a method for doing that and runs it as part of the list applications API handler.


server/app/controllers/api/CiviFormApiController.java
```diff
@@ -42,16 +48,24 @@ public Result checkAuth() {
...
   }
 
+  protected ApiKey getApiKey(Http.Request request) {
+    return profileUtils
+        .currentApiKey(request)
+        .orElseThrow(() -> new AccountNonexistentException("No API key found for profile"));
+  }
+
+  protected CompletionStage<ApiKey> recordApiKeyUsage(Http.Request request) {
+    return apiKeyRepository.recordApiKeyUsage(
```
clouser2, line 65: Blocking: This pattern requires each API endpoint to remember to call recordApiKeyUsage, which seems like it might be a source of potential errors as new endpoints are added. What about pushing the logic up to something that doesn't require each endpoint to remember to call it?

The current logic will throw if we can't record the API key usage, so pushing it up to more common logic shouldn't change any behavior around errors-for-failed-increments. The only thing we lose is a bit of asynchronous nature where the increment in parallel with the program definition lookup. For API calls, I assume this performance hit would be extremely minor.

Given the above, three potential options for where this logic could reside are:
1) ApiAuthenticator - We could make the increment call after we validate the cached key. Downside of putting it here is that if we wanted to track key usage by method, this wouldn't easily know what "route" the request was for.
2) CiviformApiController - The base class could have an abstract method for executing the request (e.g. `executeInternal`) and a final method `execute`. Implementors would put actual logic in `executeInternal` and the base class's `execute` would be responsible for incrementing the API key. This way we'd ensure logic was applied consistently. Additionally, it'd make it easier to add metrics on what route the API key was used for.
3) Refactor profileUtils.currentApiKey to increment.

clouser2 (changes requested):

server/app/controllers/api/CiviFormApiController.java
bion: I ended up implementing it using a filter. No unit test but a browser test covers the behavior. PTAL and LMK what you think!

browser-test/browser-test-compose.yml
```diff
@@ -31,4 +31,5 @@ services:
       - DB_JDBC_STRING=jdbc:postgresql://database:5432/postgres
       - BASE_URL=http://civiform:9000
       - CIVIFORM_TIME_ZONE_ID
+      - CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET=false
```
clouser2, line 34: Not blocking: For consistency, would it make sense to add this to the other compose files (e.g. `bin/run-dev` and `bin/sbt-test`)?

If so, I'd consider adding this to `application.dev.conf` instead.

browser-test/src/admin_api_keys.test.ts
```diff
@@ -35,13 +35,23 @@ describe('Managing API keys', () => {
...
       programSlugs: ['api-using-program'],
     })
 
     expect(typeof credentials).toEqual('string')
 
     await adminApiKeys.expectApiKeyIsActive('Test API key')
+
+    let apiResponse = await adminApiKeys.callCheckAuth(credentials)
+    expect(apiResponse.status).toEqual(200)
+    await adminApiKeys.expectKeyCallCount('test-api-key', 1)
```
clouser2, line 48: Blocking: Since the key count is updated from a background thread asynchronously, it's possible that we could run into an instance where the call count hasn't been updated by the time we arrive at `expectKeyCallCount`. This could lead to a low-level of flake.

Given the expected asynchronous nature, I'd suggest converting the `expectKeyCallCount` / `expectLastCallIpAddressToBeSet` methods to poll (or wrap the expectation in code that polls until the desired state is met).

browser-test/package.json
```diff
@@ -16,5 +16,8 @@
     "ts-node": "10.8.1",
     "ts-node-dev": "2.0.0",
     "typescript": "4.7.4"
+  },
+  "dependencies": {
```
clouser2, line 20: Not blocking: Two issues around consistency with other dependencies.

1. Any reason we shouldn't put this in devDependencies?
2. Should we use a specific version number rather than using "compatible-with" semantics?

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,74 @@
...
+import play.api.libs.concurrent.AkkaSchedulerProvider;
+import play.mvc.EssentialAction;
+import play.mvc.EssentialFilter;
+import play.mvc.Http;
+import scala.concurrent.ExecutionContext;
+import services.apikey.ApiKeyService;
+
+/**
+ * This filter looks for requests with paths that begin with /api and record the usage data for the
+ * relevant API key if it has one. The API key update is scheduled to run in a background through to
```
clouser2, line 20: Not blocking: Should this be "in a background thread"?

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,74 @@
...
+ * reduce latency and to ensure issues with recording the usage do not cause API requests to fail.
```
clouser2, line 21: Not blocking: I'd expand the comment either here or where we make the background call noting that both eventual consistency as well as potentially failed increments are both acceptable (the latter in the case of many concurrent API calls occurring in short succession).

Also, to confirm, lots of API key calls updating the database could lead to DB contention. Presumably, that's acceptable and not something we're worrying about for V1?

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  public EssentialAction apply(EssentialAction next) {
+    return EssentialAction.of(
+        (Http.RequestHeader request) ->
+            next.apply(request)
+                .map(
+                    result -> {
+                      if (request.path().startsWith("/api")) {
```
clouser2, line 49: Not blocking: Nit: I'd suggest switching the path matching to "/api/" to ensure that a newly added and unrelated route doesn't accidentally match this unnecessarily. While I can't think of anything along these lines off the top of my head, it's likely a good idea to be more specific to avoid head-scratching in the future.

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,74 @@
...
+                            profileUtilsProvider.get().currentApiKeyId(request);
+
+                        // If the key ID is not present then the request was not
+                        // authenticated and does not need to be recorded.
+                        if (maybeApiKeyId.isPresent()) {
+                          String remoteAddress = request.remoteAddress();
+
+                          akkaSchedulerProvider
+                              .get()
+                              .scheduleOnce(
```
clouser2, line 60: Not blocking: Probably complete overkill and not necessary here, but it might make sense to debounce / batch requests here to reduce database contention when there's a large influx of API requests.

That'd look something like having some sort of background-thread-local-variable that's incremented here, then scheduling a task (that's debounced every 1s or so) to update the current state. It could then:
1) Lock
2) Consult the background-thread-local-variable and increment by that amount
3) Unlock

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,74 @@
...
+    this.exec = checkNotNull(exec);
+    this.profileUtilsProvider = checkNotNull(profileUtilsProvider);
+  }
+
+  @Override
+  public EssentialAction apply(EssentialAction next) {
+    return EssentialAction.of(
+        (Http.RequestHeader request) ->
+            next.apply(request)
+                .map(
```
clouser2, line 47: Not blocking: Does an exception thrown in a filter presumably block a subsequent filter from running?

Given that our semantics for this are best-effort, might it make sense to wrap this whole stanza in a try/catch (and log) so that failure to retrieve the API key / schedule the background work doesn't block the response from being sent to the client?

clouser2 (changes requested):
Overall looks great! Just had a few questions specifically around edge-case semantics.

server/app/filters/ApiKeyUsageFilter.java
bion: done

server/app/filters/ApiKeyUsageFilter.java
bion: done

server/app/filters/ApiKeyUsageFilter.java
bion: done

browser-test/package.json
bion: done

browser-test/src/admin_api_keys.test.ts
bion: done

browser-test/browser-test-compose.yml
bion: The default is to not allow it, and I'm inclined to leave it that way. I'm overriding it for browser tests because we don't know what IP address it will come from

server/app/filters/ApiKeyUsageFilter.java
bion: That'd definitely be an improvement, but I don't think any environment is going to see enough traffic to make the complexity tradeoff worth it. It's also ok if the count isn't perfectly accurate

server/app/filters/ApiKeyUsageFilter.java
bion: > Also, to confirm, lots of API key calls updating the database could lead to DB contention. Presumably, that's acceptable and not something we're worrying about for V1?

yep, not a concern for the foreseeable future

browser-test/src/admin_api_keys.test.ts
```diff
@@ -34,14 +34,24 @@ describe('Managing API keys', () => {
 
     const credentials = await adminApiKeys.createApiKey({
       name: 'Test API key',
-      expiration: '2022-01-31',
-      subnet: '8.8.8.8/32',
+      expiration: '2100-01-01',
```
clouser2, line 37: Not blocking: Was this failing before the change or was it passing due to an application logic bug?

server/app/filters/ApiKeyUsageFilter.java
```diff
@@ -0,0 +1,84 @@
...
+import play.mvc.EssentialAction;
+import play.mvc.EssentialFilter;
+import play.mvc.Http;
+import scala.concurrent.ExecutionContext;
+import services.apikey.ApiKeyService;
+
+/**
+ * This filter looks for requests with paths that begin with /api and record the usage data for the
+ * relevant API key if it has one. The API key update is scheduled to run in a background thread
+ * through to reduce latency and to ensure issues with recording the usage do not cause API requests
```
clouser2, line 23: Not blocking: Should this be thread?

browser-test/browser-test-compose.yml
clouser2: Sounds good. Given that, I'd add a comment stating that for posterity.

server/app/filters/ApiKeyUsageFilter.java
clouser2: Sounds good to me. Can always add this if the need arises.

clouser2 (approved):

browser-test/src/admin_api_keys.test.ts
clouser2: Phew. That's pretty much the best outcome :D

clouser2 (approved):

PR #2797, opened 2022-06-24
Make SVG icons into an enum rather than const. Also add /dev/icons view to determine the size of each icon.

clouser2:
### Description

Moving this to an enum makes it easier to categorize what icons we actually have. Additionally, I've been dealing with issues where I try and scale an existing SVG path up and down and running into issues where the content gets clipped. This is due to the fact that the SVG paths we have have inherent sizes. We'll need to set the viewBox property to this inherent size in order to properly scale the icon up/down using the width/height properties. Having the `/dev/icons` view is a first step to determining which icons need to be updated (and is also generally useful for seeing what icons are available).

See https://css-tricks.com/scale-svg/.

### DIFF
--- a/server/app/assets/javascripts/dev_icons.ts
+++ b/server/app/assets/javascripts/dev_icons.ts
@@ -1,3 +1,5 @@
+/** This handler is responsible for annotating each rendered SVG icon
+ * with its size as calculated by the browser.  */
 window.addEventListener('load', () => {
   Array.from(document.querySelectorAll('tr')).forEach((rowEl) => {
     const svgEl = rowEl.querySelector('svg')
--- a/server/app/assets/javascripts/dev_icons.ts
+++ b/server/app/assets/javascripts/dev_icons.ts
@@ -1,0 +1,11 @@
+window.addEventListener('load', () => {
+  Array.from(document.querySelectorAll('tr')).forEach((rowEl) => {
+    const svgEl = rowEl.querySelector('svg')
+    if (!svgEl) {
+      return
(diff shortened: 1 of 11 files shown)


server/app/assets/javascripts/dev_icons.ts
```diff
@@ -0,0 +1,11 @@
+window.addEventListener('load', () => {
+  Array.from(document.querySelectorAll('tr')).forEach((rowEl) => {
```
bion, line 2: suggestion (blocking): please document this file

server/app/views/dev/IconsView.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.inject.Inject;
+import j2html.tags.ContainerTag;
+import j2html.tags.Tag;
+import play.twirl.api.Content;
+import views.BaseHtmlLayout;
+import views.BaseHtmlView;
+import views.HtmlBundle;
+import views.components.Icons;
+
+public final class IconsView extends BaseHtmlView {
```
bion, line 21: suggestion (blocking): please document this class

bion (changes requested):

server/app/views/dev/IconsView.java
clouser2: Done.

server/app/assets/javascripts/dev_icons.ts
clouser2: Done.

PR #2812, opened 2022-06-28
Using TimeFilter in API requests for consistency.

clouser2:
### Description

Follow-on to #2749, where the TimeFilter was originally introduced.

### Issue(s)

#1743

### DIFF
--- a/server/app/controllers/admin/AdminApplicationController.java
+++ b/server/app/controllers/admin/AdminApplicationController.java
@@ -282,7 +282,7 @@ public class AdminApplicationController extends CiviFormController {
     var paginationSpec = new PageNumberBasedPaginationSpec(PAGE_SIZE, page.orElse(1));
     PaginationResult<Application> applications =
         programService.getSubmittedProgramApplicationsAllVersions(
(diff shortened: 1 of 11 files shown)


server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -282,7 +282,7 @@ public Result index(
     var paginationSpec = new PageNumberBasedPaginationSpec(PAGE_SIZE, page.orElse(1));
     PaginationResult<Application> applications =
         programService.getSubmittedProgramApplicationsAllVersions(
-            programId, F.Either.Right(paginationSpec), search);
+            programId, F.Either.Right(paginationSpec), search, TimeFilter.builder().build());
```
bion, line 285: suggestion (non-blocking): add `public static final EMPTY = TimeFilter.builder().build();` to `TimeFilter` and use that

bion (approved):

server/app/controllers/admin/AdminApplicationController.java
clouser2: Done.

PR #2814, opened 2022-06-28
Remove unused export definition code

clouser2:
### Description

Note: Confirmed that the following SQL returned 0 rows, indicating there are no configured export definitions:

```
select export_definitions from programs where export_definitions != '[]';
```

### Issue(s)

#1743


bion:
@clouser2 when this rolls out, a server will start with the new code while the old servers run the old code. Queries generated by the old code will include the column to be removed, which will cause them to fail. Since the deployment automatically stalls out if error rates spike that could potentially keep the deployment from completing.

I'm not sure what the best way to handle this is, but before we move forward with this we should experiment with simulating the deployment by running a couple local servers in production mode, updating one to use the new code which will trigger the evolution, and then seeing what happens with the older ones.

clouser2:
> @clouser2 when this rolls out, a server will start with the new code while the old servers run the old code. Queries generated by the old code will include the column to be removed, which will cause them to fail. Since the deployment automatically stalls out if error rates spike that could potentially keep the deployment from completing.
> 
> I'm not sure what the best way to handle this is, but before we move forward with this we should experiment with simulating the deployment by running a couple local servers in production mode, updating one to use the new code which will trigger the evolution, and then seeing what happens with the older ones.

You're definitely correct and I think this points to a more general issue with how we evolve our DB schema with respect to rollout and versioning. A simple fix for right now would be to omit the DB evolution from this PR and add it at a later point after we're certain that all servers are running the updated code. As a general approach, this won't scale very well as we build the application code to assume an up-to-date shcema.

bion (approved):

PR #2819, opened 2022-06-28
Don't create duplicate stored files

bion:
When an applicant uploads a file for a given program-block, the file is stored in cloud storage keyed by the applicant-program-block-originalFileName -- a four-part composite key. If they upload another file (or the same file) with the same filename for the same program-block, the bytes in cloud storage are overwritten, but CiviForm creates a duplicate `StoredFile` entry with a colliding `name` value, resulting in two (or more) database entries for a single file in cloud storage.

This change prevents CiviForm from creating a duplicate as a result of this applicant behavior.

Fixes https://github.com/seattle-uat/civiform/issues/2818

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramBlocksController.java
+++ b/server/app/controllers/applicant/ApplicantProgramBlocksController.java
@@ -1,6 +1,7 @@
 package controllers.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.failedFuture;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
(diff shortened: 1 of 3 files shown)


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -294,22 +295,23 @@ public CompletionStage<Result> updateFile(
                     originalFileName.get());
               }
 
-              updateFileRecord(key.get(), originalFileName);
-              return applicantService.stageAndUpdateIfValid(
```
bion, line 298: note to reviewer: the previous behavior made an async call to insert the file record but did not handle the resulting promise, which made silent failure possible if the code to insert the record failed for some reason

server/test/controllers/applicant/ApplicantProgramBlocksControllerTest.java
```diff
@@ -494,6 +497,44 @@ public void updateFile_completedProgram_redirectsToReviewPage() {
     assertThat(result.redirectLocation()).hasValue(reviewRoute);
   }
 
+  @Test
+  public void updateFile_storedFileAlreadyExists_doesNotCreateDuplicateStoredFile() {
+    var storedFileRepo = instanceOf(StoredFileRepository.class);
```
clouser2, line 502: Not blocking: Consider adding a comment that this is a regression test and link the GitHub issue.

clouser2 (approved):
Looks good!

server/app/controllers/applicant/ApplicantProgramBlocksController.java
gwendolyngoetz: @bion Would it have been possible that a file made it's way into S3, but didn't get recorded in the database? I can pull metadata from S3 if we need to check or rebuild missing records.

PR #2826, opened 2022-06-29
Add filtering for program admin CSV / JSON exports

### LINKED ISSUE #1743: CSV extracts have customizable date range and fields

clouser2:
### Description

Adds submission time filtering to CSV / JSON exports, similar to what was added for the demography export in #2749. To accomodate this, several UI changes were made, outlined in https://github.com/seattle-uat/civiform/issues/1743#issuecomment-1170212003 (caveat that search bar improvements outlined were split into #2831).

Concretely, we:
* Allow filtering applications by time as well as the search box
* Move the download CSV buttons up to inline with the "Filter" button as a secondary action
* Pop a modal (similar to demography export) and prompt as to whether should occur with filters applied or not
* A few margin / positioning tweaks from real-time feedback with @rahaghas 

Before:
Screen Shot 2022-07-01 at 1 51 39 PM

After:
Screen Shot 2022-07-01 at 1 58 02 PM
Screen Shot 2022-07-11 at 10 10 37 AM

### Checklist

### Issue(s)

Fixes #1743


browser-test/src/application_download.test.ts
```diff
@@ -168,21 +186,21 @@ describe('normal application flow', () => {
     const demographicsCsvContent = await adminPrograms.getDemographicsCsv()
 
     expect(demographicsCsvContent).toContain(
-      'Opaque ID,Program,Submitter Email (Opaque),TI Organization,Create time,Submit time,csvcurrency (currency),csvdate (date),dropdowncsvdownload (selection),namecsvdownload (first_name),namecsvdownload (middle_name),namecsvdownload (last_name)',
+      'Opaque ID,Program,Submitter Email (Opaque),TI Organization,Create time,Submit time,name (first_name),name (middle_name),name (last_name),csvcurrency (currency),csvdate (date),dropdowncsvdownload (selection)',
```
clouser2, line 189: Note to the reviewer: The field ordering issues are here presumably due to the canonical question already being in the database. When publishing, I believe we clone questions that haven't changed. As such, it'd have a later ID and probably would end up being rendered later in the list.

browser-test/src/support/admin_programs.ts
```diff
@@ -212,7 +212,7 @@ export class AdminPrograms {
     await this.page.click('#update-block-button:not([disabled])')
 
     for (const questionName of questionNames) {
-      await this.page.click(`button:text("${questionName}")`)
+      await this.page.click(`button >> text="${questionName}"`)
```
clouser2, line 215: Note to the reviewer: Now that we use the "Name" canonical question, the previous selector would have caught the "Edit Screen Name and Description" button, since the `text()` selector asserts that the element contains the text, not that it's an exact match.

server/app/views/BaseHtmlView.java
```diff
@@ -114,38 +112,6 @@ protected ContainerTag renderPaginationDiv(
     return div.with(br());
   }
 
-  protected ContainerTag renderSearchForm(
```
clouser2, line 117: Note to the reviewer: There was only one other callsite of this and the contained logic wasn't super interesting / general. I preemptively removed it since I wanted to add more controls to the form (time filters) and it didn't currently support it.

Rather than trying to make this more general, I thought it better to remove the layer of indirection.

server/app/views/admin/programs/ProgramIndexViewV2.java
```diff
@@ -141,7 +141,7 @@ private Modal renderDemographicsCsvModal() {
                                     .withClasses(Styles.ML_3, Styles.INLINE_FLEX),
                                 FieldWithLabel.date()
                                     .setFieldName("untilDate")
-                                    .setLabelText("To:")
```
clouser2, line 144: Note to the reviewer: I forgot to rename To -> Until when originally adding time filters to the demography export.

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -265,12 +304,25 @@ public Result show(Http.Request request, long programId, long applicationId)
...
     if (page.isEmpty()) {
-      return redirect(routes.AdminApplicationController.index(programId, search, Optional.of(1)));
+      return redirect(
+          routes.AdminApplicationController.index(
+              programId, search, Optional.of(1), fromDate, untilDate));
     }
 
+    TimeFilter submitTimeFilter =
+        TimeFilter.builder()
+            .setFromTime(parseDateFromQuery(dateConverter, fromDate))
```
kmcnellis, line 322: I'd be careful here with time zones. Is it doing midnight local (browser), midnight server time, UTC, or the current time.  Since we're not asking for a timestamp or timezone, it's unclear.

server/test/services/export/JsonExporterTest.java
```diff
@@ -24,7 +25,9 @@ public void testAllQuestionTypesWithoutEnumerators() throws Exception {
         exporter
             .export(
                 fakeProgram.getProgramDefinition(),
-                IdentifierBasedPaginationSpec.MAX_PAGE_SIZE_SPEC_LONG)
+                IdentifierBasedPaginationSpec.MAX_PAGE_SIZE_SPEC_LONG,
+                Optional.empty(),
+                TimeFilter.EMPTY)
```
kmcnellis, line 30: I'd like to see more tests in here

server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -95,31 +114,135 @@ public Content render(
...
+                                        controllers.admin.routes.AdminApplicationController
+                                            .downloadAllJson(
+                                                program.id(),
+                                                Optional.empty(),
+                                                Optional.empty(),
+                                                Optional.empty(),
+                                                Optional.empty())
+                                            .url())
+                                    .withType("submit"))));
+    return Modal.builder(modalId, modalContent)
```
kmcnellis, line 239: Should we disable downloads when the current filters haven't been applied?   (or give some indication that the changes to filters won't be reflected in the download)

kmcnellis (approved):

server/test/services/export/JsonExporterTest.java
clouser2: The underlying implementation calls `ProgramRepository.getApplicationsForAllProgramVersions`, which already has coverage for all of the filtering ([code](https://github.com/seattle-uat/civiform/blob/5a7404a06e426cca2eda1422dbcf28b9d83c2e2e/server/test/repository/ProgramRepositoryTest.java#L180-L418)). Most of this is a pass-through.

Similar to the demography export, we do a smoke-test of the "plumbing" of the query parameters all the way through to the exported data through browser testing.

Please let me know if this doesn't work for you and I can try and find a good balance of additional tests in a subsequent PR.

Issue #2712, opened 2022-06-14
English translations are not surfaced in the Admin's usage of CiviForm


shanemc-goog:
**Describe the bug**
For an English language CiviForm Admin, we effectively store the english language content twice and in a way that can become out of sync without an Admin necessarily realizing it. (This technically is true for an admin using any language we translate too.)

This is caused by CiviForm store the name for Questions and Programs separate as "admin name" from having en-US as a translation language.  When the Admin is viewing the program or question list or is building Programs they see the admin name, however users will see their locale translation name which is the en-US one.

This means it's possible for the Admin to change english visible text in the translations, but not see it elsewhere in english.

goldblatt:
this needs product work in terms of the impact on the apps behaviour and which user roles (log in as applicant and using tool and logging in as an admin and using thing). Also should read up on i18n and how it works in terms of en/en_us

goldblatt:
@swatkat1 and @gwendolyngoetz FYI. we might want to run a diff between the en and en_us translations so we can know the impact of this currently so we can prio work going forward

elisekalstad:
Curious if this is the reason for the error message "Page is not fully translated into this language".

PR #2275, opened 2022-04-08
converted TIDashboard link to button and made NEXT as SAVE AND NEXT

swatkat1:
### Description
This PR has two changes:
1. The NEXT button the user sees while filling the applications is not clear if the application till the point was saved or not. So the change is to make the NEXT button as "SAVE AND NEXT".
2. The second change is is the TI home page. The TI Dahsboard link is not clearly highlighted as a button, so I have modified it to be like a SUBMIT button style. 

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary
Screen shot:
image

image

Please note - I am unable to test the TI changes locally. 
### Issue(s)

Fixes #1742 ; Fixes #2098

PR #2827, opened 2022-06-30
Syncing messages/messages.en-US files

gwendolyngoetz:
* Fixing the button.nextScreen text to be "Save and next" which is what the messages.en-US file was changed to a while back
* Diffing the messages and messages.en-US files and updating messages.en-US where out of sync

Related to #2275 and #2712


gwendolyngoetz:
Sorry, the second commit is ugly. I ended up reordering messages.en-US to be in the same order as messages because it was becoming a challenge to match everything.

bion (approved):

PR #2831, opened 2022-06-30
Program admin applicant search improvements to allow full name search

clouser2:
### Description
Noticed this as part of pending PR #2826.

Previously, program admins could search by the applicant's first name, last name, email, or application ID. This augments the functionality to allow additionally searching using the following forms:
* First Last
* Last First
* Last, First

A couple of additional improvements were also added, including:
* Searching by email is case insensitive (even if the stored email has mixed casing)
* Updated the search field label to indicate that email is a valid search option
* Adding unit test coverage that wasn't previously present

### Checklist

### Issue(s)

#1743

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -31,6 +31,8 @@ import play.libs.F;
 import services.IdentifierBasedPaginationSpec;
 import services.PageNumberBasedPaginationSpec;
 import services.PaginationResult;
+import services.Path;
+import services.WellKnownPaths;
 import services.program.ProgramDefinition;
 import services.program.ProgramNotFoundException;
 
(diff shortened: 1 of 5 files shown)


server/app/repository/ProgramRepository.java
```diff
@@ -247,23 +249,20 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
...
                 .or()
-                .eq("submitter_email", search)
-                .raw(
-                    getApplicationObjectPath("applicant", "name", "first_name") + " ILIKE ?",
-                    "%" + search + "%")
-                .raw(
-                    getApplicationObjectPath("applicant", "name", "last_name") + " ILIKE ?",
-                    "%" + search + "%")
+                .ieq("submitter_email", search)
+                .raw(firstNamePath + " || ' ' || " + lastNamePath + " ILIKE ?", "%" + search + "%")
```
clouser2, line 263: Note to the reviewer: JSONB allows for concatenated strings to be passed to ILIKE, similar to standard string columns: https://www.postgresql.org/docs/9.5/functions-json.html

server/test/support/ResourceCreator.java
```diff
@@ -120,6 +124,7 @@ public Applicant insertApplicantWithAccount(Optional<String> accountEmail) {
     Account account = insertAccount();
 
     accountEmail.ifPresent(account::setEmailAddress);
+    account.save();
```
clouser2, line 127: Note to the reviewer: Not calling this before meant that the call to setEmailAddress above was ineffective.

bion (approved):

PR #2833, opened 2022-06-30
Small fixes for AWS deploy

kmcnellis:
### Description

 - Make deploys more reliable by adjusting the auto-generated passwords/user names to work regardless of the prefix and special characters chosen at random.
 - Adds terraform files to the gitignore
 - Don't try to delete the terraform state bucket (It causes unrecoverable errors) when changing prefix
   - Instead, run `cloud/deploys/dev_aws/bin/terraform-setup state rm module.setup.aws_s3_bucket.backend_state_bucket` to remove it from the state
 - Add scripts to run terraform directly, when needed. 
 - Pipe SES_SENDER_EMAIL into terraform.

### DIFF
--- a/cloud/aws/modules/setup/secrets.tf
+++ b/cloud/aws/modules/setup/secrets.tf
@@ -8,8 +8,11 @@ resource "aws_kms_key" "civiform_kms_key" {
 
 # Create a random generated password to use for postgres_password.
 resource "random_password" "postgres_username" {
-  length  = 16
+  length  = 7
   special = false
+  keepers = {
+    version = 1
+  }
 }
 
 # Creating a AWS secret for postgres_username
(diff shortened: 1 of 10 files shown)


cloud/aws/modules/setup/secrets.tf
```diff
@@ -21,13 +24,18 @@ resource "aws_secretsmanager_secret" "postgres_username_secret" {
...
-  secret_string = random_password.postgres_username.result
+  secret_string = "db_admin_${random_password.postgres_username.result}"
 }
 
 # Create a random generated password to use for postgres_password.
 resource "random_password" "postgres_password" {
-  length  = 16
-  special = false
+  length           = 40
+  special          = false
```
azizova-leyla, line 33: this needs to be true for override_special to take effect

azizova-leyla (changes requested):

cloud/aws/modules/setup/secrets.tf
kmcnellis: done

azizova-leyla (approved):

Issue #1937, opened 2022-02-14
Refactor browser tests


cassandralfernandes:
Browser test coverage is uneven and poorly organized. We should refactor so that

1. every question type and every question type configuration option is covered by a browser test
2. browser tests and focused, as much as feasible, to individual features (right now single tests will test a wide variety of behaviors)
3. testing patterns and expectations are clear for new contributors

PR #2851, opened 2022-07-06
Add a browser level test for the checkbox question type

jeanniefu:
### Description
Add a browser level test for the checkbox question type. For all question types, I propose we have a browser level test that covers these high level scenarios:
- Single question of that type:
  - Test that submits successfully
  - Test that errors when required and was not answered
  - Test any question type specific errors
- Multiple questions:
  - Test that submits successfully
  - Test that an unanswered optional question allows submission
  - Test that first invalid Q prevents submission
  - Test that second invalid Q prevents submission

After this looks good and we get alignment, I will send out additional PRs to address all other missing question types.

### Checklist

### Issue(s)
#1937

### DIFF
--- a/browser-test/src/checkbox.test.ts
+++ b/browser-test/src/checkbox.test.ts
@@ -74,7 +74,6 @@ describe('Checkbox question for applicant flow', () => {
       // Check checkbox error and required error are present.
       expect(await pageObject.isHidden(checkBoxError)).toEqual(false)
       const checkboxId = '.cf-question-checkbox'
-      // Contains this substring. Is this brittle? Would have to update text every time. Or can I add an ID just for testing?
       expect(await pageObject.innerText(checkboxId)).toContain(
         'This question is required.',
       )
@@ -150,7 +149,7 @@ describe('Checkbox question for applicant flow', () => {
       await applicantQuestions.submitFromReviewPage(programName)
     })
 
-    it('unanswered optional question submits successfully', async () => {
+    it('with unanswered optional question submits successfully', async () => {
       await loginAsGuest(pageObject)
       await selectApplicantLanguage(pageObject, 'English')
 
--- a/browser-test/src/checkbox.test.ts
+++ b/browser-test/src/checkbox.test.ts
@@ -1,0 +1,205 @@
+import {
+  AdminPrograms,
+  AdminQuestions,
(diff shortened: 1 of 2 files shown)


browser-test/src/checkbox.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      const checkBoxError = '.cf-applicant-question-errors'
+      expect(await pageObject.isHidden(checkBoxError)).toEqual(true)
+
+      // Click next without selecting anything.
+      await applicantQuestions.clickNext()
+
+      // Check checkbox error and required error are present.
+      expect(await pageObject.isHidden(checkBoxError)).toEqual(false)
+      const checkboxId = '.cf-question-checkbox'
+      expect(await pageObject.innerText(checkboxId)).toContain(
```
jeanniefu, line 77: This is a bit brittle, because if the text changed, we'd have to update it here too. Alternatively, we could add an id to the required error and check that instead. It felt a bit weird to add an id just for testing though.

browser-test/src/checkbox.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        'orange',
+      ])
+      await applicantQuestions.clickNext()
+
+      // Check error is shown.
+      expect(await pageObject.isHidden(checkBoxError)).toEqual(false)
+    })
+  })
+
+  describe('multiple checkbox questions', () => {
```
jeanniefu, line 104: Actually, I feel like this type of test should not belong in the question type specific browser tests. This feels independent of what question type is being used. Instead, I'd like to create just one of these tests in application_navigation.test.ts, and remove it from all the question type specific tests. WDYT?

shanemc-goog: I believe we have these because FE validation is bespoke for each type; at one point some questions handled multiples of their types and others did not.

browser-test/src/checkbox.test.ts
shanemc-goog: I think this is fine. The developer will easily catch it when they run the tests before merging.

Ideally we could reference a shared value but because it spans languages we can't.

jeanniefu: Ok sounds good, thanks for your input!

PR #2860, opened 2022-07-08
Improve paths for browser test videos.

nbeloglazov:
Currently we always try to use current test name as directory for storing video. This doesn't work for tests which initialize context in `beforeAll`. This PR groups videos by test file directories. So instead of

```
tmp/video/undefined/video1.webm
tmp/video/Name of test case 1/video2.webm
tmp/video/Name of test case 2/video3.webm
```

we get

```
tmp/video/first.test.ts/video1.webm
tmp/video/second.test.ts/Name of test case 1/video2.webm
tmp/video/second.test.ts/Name of test case 1/video3.webm
```


clouser2 (approved):
Nice!

PR #2868, opened 2022-07-11
Switch from apprunner to fargate

### LINKED ISSUE #2822: Switch to Fargate in AWS terraform

azizova-leyla:
Creating basic fargate config.  This works, will make it better in subsequent PRs.

### Issue(s)
Fixes #2822


cloud/aws/templates/aws_oidc/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "aws_region" {
 variable "civiform_image_repo" {
   type        = string
   description = "Public ECR repository with Civiform images"
-  default     = "public.ecr.aws/t1q6b4h2/universal-application-tool"
+  default     = "civiform/civiform"
```
kmcnellis, line 10: Is this correct?  If we are using the docker repo, then the description should change.

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -0,0 +1,107 @@
+module "ecs_cluster" {
+  source = "cn-terraform/ecs-cluster/aws"
+  name   = var.app_prefix
```
kmcnellis, line 3: Can you add the tags block to these?  They all support it

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -0,0 +1,107 @@
...
+      "awslogs-create-group"  = "true",
+    }
+    secretOptions = null
+  }
+}
+
+module "ecs_fargate_service" {
+  source      = "cn-terraform/ecs-fargate-service/aws"
+  name_prefix = var.app_prefix
+  # TODO: create and validate this certificate in terraform
```
kmcnellis, line 89: What is this certificate?

kmcnellis (changes requested):

cloud/aws/templates/aws_oidc/variables.tf
azizova-leyla: I think it will technically work with both. Changed to Dockehub to avoid confusion.

cloud/aws/templates/aws_oidc/app.tf
azizova-leyla: The arn that's there right now is just a certificate I created manually. We should use https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate instead.

cloud/aws/templates/aws_oidc/app.tf
azizova-leyla: Done

kmcnellis (approved):

PR #2870, opened 2022-07-11
File migration fix

bion:
If a file was created before the migration to using StoredFileAcls, programReadAcls will be null on initial load. This caused a NullPointerException when loading the record in the migration task. This PR sets the internal state of the ACLs to an empty collection so the migration task can populate them.

### DIFF
--- a/server/app/auth/StoredFileAcls.java
+++ b/server/app/auth/StoredFileAcls.java
@@ -1,11 +1,10 @@
 package auth;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableSet;
 import java.util.HashSet;
+import javax.annotation.Nullable;
 import models.Account;
 import services.program.ProgramDefinition;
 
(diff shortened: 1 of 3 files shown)


server/app/auth/StoredFileAcls.java
```diff
@@ -25,8 +24,20 @@ public StoredFileAcls() {
...
-    this.programReadAcls = checkNotNull(programReadAcls);
+  public StoredFileAcls(
+      @Nullable @JsonProperty("programReadAcls") HashSet<String> programReadAcls) {
+    // If the file was created before the migration to using StoredFileAcls,
+    // programReadAcls will be null on initial load. In this case we initialize
+    // the internal state of the ACLs to an empty collection so the migration
+    // task can populate them.
+    // It is safest to leave this code in, even after the migration, in the
+    // event that somehow a StoredFile is created without serializing an
+    // instance StoredFileAcls.
```
shanemc-goog, line 35: I don't think having the first part of this comment is particularly useful given the second part basically makes allowances for unknown reasons that would cause it.

I'd just do something more like "If there aren't read acls for some reason use an empty acl" and log the unexpected situation.

server/test/repository/StoredFileRepositoryTest.java
```diff
@@ -64,4 +67,23 @@ public void lookupFile() {
 
     assertThat(result).isEqualTo(file);
   }
+
+  /**
+   * This is a regression test to ensure that files created before the migration to stored file ACLs
```
shanemc-goog, line 72: Related, if this is going to stick around after the migration then it should be commented to make sense when someone reads it in 6 months.

PR #2872, opened 2022-07-11
Use the underlying exception when completable futures error

kmcnellis:
### Description
This unwraps completion exceptions and uses the custom civiform exception handler for logins.
Also updates the ProfileMergeConflictException to store the error message in the exception, rather than using the non-functioning redirect action from pac4j.

Previously, errors were getting wrapped and rendered useless.  This enables the underlying exception message to be displayed.

### DIFF
--- a/server/app/controllers/ErrorHandler.java
+++ b/server/app/controllers/ErrorHandler.java
@@ -55,6 +55,13 @@ public class ErrorHandler extends DefaultHttpErrorHandler {
 
   @Override
   public CompletionStage<Result> onServerError(RequestHeader request, Throwable exception) {
+
+    // Unwrap exceptions thrown within a CompletableFuture, to handle the
+    // original error stack trace.
+    if (exception instanceof CompletionException) {
+      exception = exception.getCause();
+    }
+
     // Exceptions that reach here will generate 500s. Here we convert certain ones to different user
     // visible states. Note: there are methods on the parent that handle dev and prod separately.
     Optional<Throwable> match = findThrowableByTypes(exception, BAD_REQUEST_EXCEPTION_TYPES);
@@ -67,11 +74,6 @@ public class ErrorHandler extends DefaultHttpErrorHandler {
 
     if (match.isPresent()) {
       return CompletableFuture.completedFuture(Results.unauthorized());
-    }
-
-    // Unwrap CompletionException
-    if (exception instanceof CompletionException) {
-      return super.onServerError(request, exception.getCause());
     }
 
     return super.onServerError(request, exception);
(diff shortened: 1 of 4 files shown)


server/app/controllers/ErrorHandler.java
```diff
@@ -68,6 +69,11 @@ public CompletionStage<Result> onServerError(RequestHeader request, Throwable ex
       return CompletableFuture.completedFuture(Results.unauthorized());
     }
 
+    // Unwrap CompletionException
```
shanemc-goog, line 72:
```suggestion
    // Unwrap CompletionException.
```

server/app/controllers/LoginController.java
```diff
@@ -47,12 +47,13 @@ public class LoginController extends Controller {
...
       @AdminAuthClient @Nullable IndirectClient adminClient,
       @ApplicantAuthClient @Nullable IndirectClient applicantClient,
+      CiviFormHttpActionAdapter civiFormHttpActionAdapter,
       SessionStore sessionStore,
       Config config) {
     this.adminClient = adminClient;
     this.applicantClient = applicantClient;
     this.sessionStore = Preconditions.checkNotNull(sessionStore);
-    this.httpActionAdapter = PlayHttpActionAdapter.INSTANCE;
+    this.httpActionAdapter = civiFormHttpActionAdapter;
```
shanemc-goog, line 56: What is this change doing?

server/app/controllers/ErrorHandler.java
shanemc-goog, line 72: This 'what' is pretty readable, can you document the 'why' also?

I'm trying to figure out if this changes the information revealed to the user in prod mode and also wondering what information that is.  An end user seeing an exception isn't the most useful thing (as compared to a nice broken robot) and we should be careful about leaking sensitive configuration information through exceptions.

Do you know how that changes (if at all) with this?   EG I know for a lot of error users see the general red screen with a hashcode that an IT manager could look for in the logs.

kmcnellis: This doesn't affect that visibility at all.  If something was shown before to a user, it still will be (and that seems to usually be a rather vague "oops an error occurred" 
image
 
 
 All this does is unwrap any exceptions thrown inside a completable future, (which are rather unhelpfully wrapped in a completionexception), and shows the error message for underlying error that was thrown

server/app/controllers/LoginController.java
kmcnellis: This just aligns the login error handler with the custom one Bion made for the API.  CiviFormHttpActionAdapter inherits from PlayHttpActionAdapter, and currently only overrides responses for API requests.  I mainly did this so that we don't have 2 different paths for errors.

shanemc-goog (approved):

PR #2874, opened 2022-07-12
Improve application index and preview look on mobile

nbeloglazov:
#### Applications index page
Add padding so that text doesn't start immediately at the border. 
Before: 
image
After:
image

### Application preview page
Reworked question DOM layout. Each question now consists of 2 columns. The left column contains question text with optional answer underneath. The right column contains optional answer timestamp with optional action link (edit/continue) vertically centered. That layout works better on mobile and avoids overlapping of text (see on screenshots). Most screens didn't change, only the second before/after pair is affected.

Before:
image
After: 
image

Before:
image

After:
image

Before:
image

After:
image

Before:
image

After:
image

### DIFF
--- a/server/app/views/applicant/ApplicantProgramSummaryView.java
+++ b/server/app/views/applicant/ApplicantProgramSummaryView.java
@@ -166,7 +166,7 @@ public final class ApplicantProgramSummaryView extends BaseHtmlView {
     }
 
     ContainerTag actionAndTimestampDiv =
-        div().withClasses(Styles.PR_2, Styles.FLEX, Styles.FLEX_COL);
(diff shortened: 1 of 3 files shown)


server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -200,10 +200,8 @@ private ContainerTag renderQuestionSummary(
               .setHref(editLink)
               .setText(editText)
               .setStyles(
-                  Styles.ABSOLUTE,
```
azizova-leyla, line 203: this absolute has been bothering me, thank you so much for fixing!

azizova-leyla (approved):
Thank you for fixing it! This is exactly the type of things that are great to fix as you go.
FYI: we have larger redesign ideas captured in https://github.com/seattle-uat/civiform/issues/1920.

PR #2880, opened 2022-07-12
Update the devex: recreate civiform container on ctrl-c 

kmcnellis:
### Description

- Recreates the civiform container each time you ctrl-c out of it (but the other emulators stay running to improve restart time).
- Also tell the dev container to avoid re-downloading dependancies on first start (rely on the ones built-in to the container).
- Shows output when pulling images from docker hub

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -5,7 +5,8 @@ version: '3.4'
 # Shared yaml extensions (https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields)
 # Volume mapping shared between containers.
 x-sbt-volumes:
-  volumes: &sbt-volumes
+  volumes:
+    &sbt-volumes
     - ./server:/usr/src/server
     # Un-shadow the build directories under server.
     - target:/usr/src/server/target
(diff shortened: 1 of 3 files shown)


clouser2 (approved):
Thanks for fixing this up!

docker-compose.dev.yml
```diff
@@ -23,10 +23,11 @@ services:
     volumes: *sbt-volumes
     stdin_open: true # docker run -i
     tty: true # docker run -t
+    restart: always
     ports:
       - 9000:9000
       - 8457:8457
-    command: -jvm-debug "*:8457" ~run -Dconfig.file=conf/application.dev.conf
+    command: -jvm-debug "*:8457" ~run -Dsbt.offline -Dconfig.file=conf/application.dev.conf
```
nbeloglazov, line 30: Can you add comment clarifying why sbt.offline is helpful in this scenario? Will it prevent pulling dependencies from maven repo?

kmcnellis: Yes - I'm not 100% sure it will always skip them, but it anecdotally seems to work

clouser2 (approved):

docker-compose.dev.yml
nbeloglazov: How will it work in the case when someone adds are new dependency locally and runs `run-dev` expecting that it will fetch it and compile app with it? Will the engineer have to remove `sbt.offline`?

kmcnellis: It looks like it's just a hint, and will download the dependancy anyway

nbeloglazov (approved):

PR #2884, opened 2022-07-12
Validate the APP_PREFIX config variable only contains alphanumeric characters (and hyphen)

### LINKED ISSUE #2806: In deploy scripts for AWS don't allow names with a dot
It breaks s3 bucket certificates, and might have other issues. Prefer - or _.

Add a check in our scripts for this.

clouser2:
### Description

This implements the top-level validation described in https://github.com/seattle-uat/civiform/issues/2806#issuecomment-1182525408.

While here, I made several additional changes encountered while attempting to test and validate this change. They were bundled into this PR since the PR itself doesn't really make much sense without the changes to support testing:
* Attempted to simplify the multi-return-value semantics of `ConfigLoader.validate_config` since we can already represent whether a config is valid by returning an empty set of validation errors
* Updated the set of paths our `validate_variable_definitions` file runs on to include the variables from AWS / Azure configs. This approach doesn't seem super scalable (filed #2887)
* Run several additional Python tests (including `config_loader_test.py`) as part of PR checks. Again, this isn't scalable and we should invest into running ALL Python unit tests as part of PR checks (filed #2888)
* Switched to use single quotes in `config_loader.py` for internal file consistency

Tested:
* Introduced an invalid configuration in `variable_definitions.json` and confirmed the PR check fails ([commit](https://github.com/seattle-uat/civiform/pull/2884/commits/7abcd71b06d3680495a56d4bf3b1c5529566a082) / [failing GitHub check](https://github.com/seattle-uat/civiform/runs/7326467483?check_suite_focus=true))
* Introduced an invalid value for `APP_PREFIX` in my local `civiform_config.sh` and observed:
  ```
  Found the following validation errors:
        [APP_PREFIX] only lowercase alphanumeric characters and hyphens allowed
  ```

## Release notes:

Validate the APP_PREFIX config variable only contains alphanumeric characters (and hyphen)

### Checklist

### Issue(s)

Fixes #2806


cloud/shared/bin/lib/config_loader.py
```diff
@@ -63,27 +64,34 @@ def get_env_variables(self, variable_definitions: dict):
...
-                        f"{config_value} not supported enum for {name}")
+                        f'[{name}] \'{config_value}\' is not a supported enum value. Want a value in [{", ".join(definition.get("values"))}]'
+                    )
 
-        return is_valid, validation_errors
+            value_regex = definition.get('value_regex', None)
+            if config_value is not None and value_regex:
+                validation_error = definition.get(
+                    'value_regex_error_message', None)
+                if not validation_error:
```
azizova-leyla, line 88: why is this **not** validation_error? I'd imagine if validation error is empty we're good?

cloud/shared/validate_variable_definitions.py
```diff
@@ -28,7 +29,9 @@ def load_repo_variable_definitions_files(self):
         # add their paths here.
         cwd = os.getcwd()
         definition_file_paths = [
-            cwd + "/cloud/shared/variable_definitions.json"
+            f"{cwd}/cloud/shared/variable_definitions.json",
```
azizova-leyla, line 32: nit: single quotes?

cloud/shared/bin/lib/config_loader.py
```diff
@@ -63,27 +64,34 @@ def get_env_variables(self, variable_definitions: dict):
...
         for name, definition in variable_definitions.items():
-            is_required = definition.get("required", False)
+            is_required = definition.get('required', False)
             config_value = configs.get(name, None)
 
             if is_required and config_value is None:
-                is_valid = False
-                validation_errors.append(
-                    f"{name} is required, but not provided")
+                validation_errors.append(f'[{name}] required, but not provided')
```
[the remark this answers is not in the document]
clouser2: Good suggestion! Rather than returning, I continue in order to process additional variable definitions. This also allowed me to remove the if config_value is not None checks below.

cloud/shared/bin/lib/config_loader.py
clouser2: Renamed to `validation_error_message`. This is saying that we'll throw an error if there's no `validation_error_message` value defined.

This is actually something that shouldn't happen assuming that the validate_variable_definitions test has run on the config. As part of #2887, we should be able to remove this entirely. I left a TODO to that effect.

cloud/shared/validate_variable_definitions.py
clouser2: I agree that I prefer single-quotes. I left this as double-quotes for consistency with the rest of the file. It'd be nice to do a pass on all of our Python files in a single PR and convert to preferring one over the other, but probably not super high priority.

clouser2 (commented):
Feedback incorporated :)

azizova-leyla (approved):

PR #2886, opened 2022-07-13
Add "unsupported browser" page

### LINKED ISSUE #512: If user is on IE, show banner telling them to upgrade to Edge/etc
@AWBaumann has reported that IE11, which is not HTML5-compatible, is making life hard for us. We agreed it's fine to not support IE11 given that it's going EoL soon, which will let us use more HTML5 components.

However, many public schools and libraries will still have IE11 (or older) installed, and many of our applicants will be using it. So, if we detect the user is on any version of Internet Explorer, we should show a banner telling them to upgrade to Edge or another browser. The good news is that pretty much every modern Windows PC will have IE *and* Edge, so users shouldn't have too hard a time going to Edge.

nbeloglazov:
### Description

Implement filter that redirects users with old browsers (IE9-11) to a page which tells them to use one of the modern browsers. Here is how it looks: 

image

## Release notes:

Users with legacy browsers such as IE11 and below are now shown a  page asking them to use one of the modern browsers (Edge, Firefox, Safari, Chrome).

### Checklist

### Issue(s)

Fixes #512


nbeloglazov:
I didn't use translations for the text on the page. Wasn't sure if I should given that this page is shown before user has a chance to choose language themselves. Do we rely on the detected language coming from `Accept-Language` header?

server/app/filters/UnsupportedBrowserFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import play.mvc.EssentialAction;
+import play.mvc.EssentialFilter;
+import play.mvc.Http;
+import play.mvc.Results;
+
+/**
+ * Filter that redirects requests from unsupported browsers (e.g. IE11) to a page that tells user to
+ * use newer browser.
+ */
+public class UnsupportedBrowserFilter extends EssentialFilter {
```
clouser2, line 15: Not blocking: Consider preemptively marking this as final to make the choice to allow subclassing explicit rather than implicit.

server/app/controllers/SupportController.java
```diff
@@ -9,20 +9,30 @@
 import play.mvc.Http;
 import play.mvc.Result;
 import views.support.UnconfirmedIdcsEmailBugView;
+import views.support.UnsupportedBrowserView;
 
 public class SupportController extends Controller {
```
clouser2, line 14: Not blocking: Consider preemptively marking this as final to make the choice to allow subclassing explicit rather than implicit.

server/app/views/support/UnsupportedBrowserView.java
```diff
@@ -0,0 +1,58 @@
...
+import j2html.tags.ContainerTag;
+import javax.inject.Inject;
+import play.twirl.api.Content;
+import views.BaseHtmlView;
+import views.HtmlBundle;
+import views.applicant.ApplicantLayout;
+import views.style.StyleUtils;
+import views.style.Styles;
+
+public class UnsupportedBrowserView extends BaseHtmlView {
```
clouser2, line 17: Not blocking: Consider preemptively marking this as final to make the choice to allow subclassing explicit rather than implicit.

server/app/views/support/UnsupportedBrowserView.java
```diff
@@ -0,0 +1,58 @@
...
+        div().withClasses(Styles.CONTAINER, Styles.MX_AUTO, Styles.MAX_W_XL, Styles.P_4);
+    container.with(
+        h1("Unsupported browser").withClasses(Styles.TEXT_CENTER, Styles.MB_4),
+        div(
+            "Please use one of the browsers listed below. If you are using Windows you should have"
+                + " Microsoft Edge installed on your computer."));
+    ContainerTag browsers =
+        div().withClasses(Styles.FLEX, Styles.FLEX_WRAP, Styles.JUSTIFY_AROUND, Styles.MT_6);
+    browsers.with(
+        createBrowserIcon("Microsoft Edge", "edge-logo-128px.png"),
```
clouser2, line 40: Are we allowed to include or reference these images? Are they in the public domain or available for including?

To be clear, I have no idea what is or isn't acceptable here, just thought I'd flag to make sure we had made an explicit decision.

server/app/views/support/UnsupportedBrowserView.java
```diff
@@ -0,0 +1,58 @@
...
+    this.applicantLayout = checkNotNull(applicantLayout);
+  }
+
+  public Content render() {
+    HtmlBundle bundle = applicantLayout.getBundle();
+
+    bundle.setTitle("The browser you are currently using is not supported");
+    ContainerTag container =
+        div().withClasses(Styles.CONTAINER, Styles.MX_AUTO, Styles.MAX_W_XL, Styles.P_4);
+    container.with(
```
clouser2, line 32: Not blocking: Feel free to chain these since they're unconditionally added.

e.g.:

```
DivTag container =
  div()
    .withClasses(...)
    .with(
      h1(...),
      div(...)
    );
```

Same for the below initialization of the `browsers` var. If so inclined, you can inline all of this into one big initialization, but that's definitely a style preference.

server/app/views/support/UnsupportedBrowserView.java
```diff
@@ -0,0 +1,58 @@
...
+    container.with(browsers);
+
+    bundle.addMainContent(container);
+    return applicantLayout.render(bundle);
+  }
+
+  private static ContainerTag createBrowserIcon(String name, String logo) {
+    return div(
+            img()
+                .withSrc("/assets/images/" + logo)
```
clouser2, line 53: Blocking: Add alt text for each of these for a11y.

server/test/filters/UnsupportedBrowserFilterTest.java
```diff
@@ -0,0 +1,90 @@
...
+          + " Version/15.4 Safari/605.1.15";
+
+  @Test
+  public void testMissingUserAgentDoesNotTriggerRedirect() throws Exception {
+    Result res = runFilter(fakeRequest(), Results.ok(""));
+    assertThat(res.status()).isEqualTo(200);
+  }
+
+  @Test
+  public void testOldIETriggerRedirect() throws Exception {
```
clouser2, line 46: Not blocking: Nit. You can use a parameterized test here so that you'll get individual test names for each of the things being iterated (https://github.com/seattle-uat/civiform/blob/7fea10a885d891f8926bf1fbd24ca0148f9e7732/server/test/repository/ProgramRepositoryTest.java#L234).

A guide to the general JUnit mechanism: https://www.baeldung.com/parameterized-tests-junit-5

Same feedback applies to the other browser tests.

server/app/filters/UnsupportedBrowserFilter.java
```diff
@@ -0,0 +1,39 @@
...
+  @Override
+  public EssentialAction apply(EssentialAction next) {
+    return EssentialAction.of(
+        request -> {
+          if (UnsupportedBrowserFilter.isUnsupportedBrowser(request)) {
+            Call unsupportedBrowser =
+                controllers.routes.SupportController.handleUnsupportedBrowser();
+            // Don't redirect if request is for a static asset or for the unsupported browser page
```
clouser2, line 23: Nit and not blocking: Consider moving this clause above the isUnsupportedBrowser check and early out.

server/app/filters/UnsupportedBrowserFilter.java
```diff
@@ -0,0 +1,39 @@
...
+              return Accumulator.done(Results.redirect(unsupportedBrowser));
+            }
+          }
+          return next.apply(request);
+        });
+  }
+
+  private static boolean isUnsupportedBrowser(Http.RequestHeader request) {
+    Optional<String> userAgent = request.getHeaders().get(Http.HeaderNames.USER_AGENT);
+    // Trident/[5-7].0 browsers are IE9-IE11. Everything else consider supported.
```
clouser2, line 36: Blocking: User agent matching tends to lead to a lot of false positives / false negatives. If possible, I'd consider checking to see if there are any libraries out there that handle this and making use of them.

That said, the regex here is straightforward. I just worry about what needs to be matched changing over time and us forgetting to update, which would lead to an unsupported browser page even if we actually did support it :(.

clouser2 (changes requested):
Nice! I'm a fan of the approach and structure, just a few open questions around 1) whether there's a library we can use and 2) whether it's acceptable for us to make use of the browser logos.

server/test/filters/UnsupportedBrowserFilterTest.java
nbeloglazov: I don't think I can use it here unfortunately. We are using Junit4 for unit tests `org.junit.*` packages. Juni4 allows you to parameterize whole test class only, it doesn't allow to parameterize individual test methods.

The link you are mentioned is junit5 which uses `org.junit.jupiter.*` packages. If I were to use it - we should probably migrate all existing test to junit5 and mandate usage of junit5. And that requires some SBT [plugin](https://github.com/sbt/sbt-jupiter-interface) it seems.

nbeloglazov: Oh, we are actually using some third-party plugin that you referenced: JUnitParamsRunner. So it's not junit5 thing. Ok, I'll rewrite the test then.

server/app/filters/UnsupportedBrowserFilter.java
nbeloglazov: Do you think that will be faster? Because it will remove all static requests faster and we have a lot of those?

server/test/filters/UnsupportedBrowserFilterTest.java
clouser2: Apologies for the poor link :(

server/app/filters/UnsupportedBrowserFilter.java
nbeloglazov: I looked at some libraries, but couldn't find something suitable for java. I found this: https://github.com/nielsbasjes/yauaa but i had few issues:
1. Required adding a bunch of logging libraries like log4j or slf4j. I don't fully understand logging landscape in java to evaluate what it might mean for the app.
2. Required non-trivial initialization. You need to initialize instance to parse UA string and it says it may take some time and they warn against using static members: [doc](https://yauaa.basjes.nl/using/#avoid-using-it-as-a-static-member). So would need to do something around singleton injection...

Generally wasn't sure how fast the UA parsing/detection would be when using full-blown library. And here we are more sensitive to latency as this code runs for every single request. Hm, talking about that, I should pre-compile regex.

I agree that if we start excluding more browsers and this logic starts to grow - we should use some library.

server/app/filters/UnsupportedBrowserFilter.java
nbeloglazov: Done.

server/app/views/support/UnsupportedBrowserView.java
nbeloglazov: Done.

server/app/views/support/UnsupportedBrowserView.java
nbeloglazov: Done.

server/app/views/support/UnsupportedBrowserView.java
nbeloglazov: Done.

gwendolyngoetz:
More of a tangent, but when I can find who can give me access to the Seattle tag manager data I'll provide some stats on current usage.

gwendolyngoetz (commented):
Non-blocking nit: Any chance we can get the browser logos closer to the same size? The Safari one stands out as being much smaller.

server/app/filters/UnsupportedBrowserFilter.java
clouser2: My original rationale was more of a readability / style preference to avoid an additional level of indentation.

Having an early out feels a bit more in fitting with the idea that "none of the logic in this filter is applied for static assets".

Totally up to you on that though, my preference is weak.

clouser2 (approved):

nbeloglazov:
@gwendolyngoetz updated safari logo
What's Seattle tag manager? Is it google analytics or something like that? 
I think we'll also be able to see useragents from server logs. Though those will require some parsing.

PR #2895, opened 2022-07-13
Remove dependency on junit 5.

nbeloglazov:
### Description

We use Junit 4 to execute tests and AssertJ for performing assertions. Yes we include Junit 5 as test dependency. Some tests use `assertThrows` provided by it. AssertJ has similar method `assertThatThrownBy` and should be used instead for consistency. Plus it removes potential confusion that project is using Junit 5.


clouser2 (approved):
Nice cleanup!

Issue #2788, opened 2022-06-23
Allow questions to be deleted


inghamn:
With multiple people working on many programs, questions end up being created but never used.  These can often be entered in error or even duplicates of questions that already existed and used in other programs.

In order to clean up the question bank, it would be nice to permit the deletion of questions that are not used in any programs.

cassandralfernandes:
Currently, you can archive questions that are not part of a program but the feature is not intuitive. Cliff, I've added this as an agenda topic to our next Bloomington + Google check in to understand more about your use case and whether or not this issue could be solved with archiving

azizova-leyla:
Assigning this to Brett to add indicators to the UI when a question is being in programs, and better messaging about archival options.

clouser2:
I plan to start work on this tomorrow. I'll check in with @rahaghas later today to collect feedback on small wins here.

clouser2:
Concretely, here's where the archive option shows up: ![Screen Shot 2022-07-12 at 2 49 44 PM](https://user-images.githubusercontent.com/1870301/178601563-f9fcf008-52cb-45f5-9a66-bb0cac0d1a5e.png)

Note that this presently only appears if the question isn't referenced by any active programs. Archiving prevents the question from showing up in the question bank. The CiviForm admin has the option to un-archive one of these questions.

A few hopefully small tweaks we can make to improve the usability here are:
* Split archived questions off into an entirely separate table / section (add a note that these won't be visible in the question bank). As it stands, previously archived questions still show in the list in their previous position. The act of archiving a question warrants putting it below the fold of other questions in my opinion.
* Add another column to the table indicating the number of programs the question is actively used in (which would hopefully make it more apparent why the archive option is not available). Make this a badge with a tooltip listing the programs that the question is referenced by (extra credit)
* Indicate 1) whether the question is archived and 2) number of programs it's referenced by in the page rendered after clicking "View" as well.
* Once the "what programs is this question referenced by" functionality is built out, we could also attempt to surface this in the question bank, though I'm not sure of the best avenue

@rahaghas What do you think about this as a starting point?

clouser2:
Confirmed with @rahaghas on the UX side that the plan above seems like a reasonable start.

clouser2:
Updates have emerged since my latest plan.

## Text to use for referencing programs counts

Surfacing a count of programs a question is used in is slightly complicated. A few examples:
[...]

PR #2903, opened 2022-07-14
Prevent archiving / restoring questions that aren't in a valid state.

clouser2:
### Description

Presently, the `/archive` and `/restore` handlers don't do any validation beyond the question ID existing. This means that a question that is still referenced by a program can be archived.

This could definitely occur for a stale tab scenario where an admin has the archive option, then references the question from a separate program (rendering it unarchivable).

Additionally, we should always rely on validation before we mutate state rather than relying on links being generated at the appropriate time.

Currently, this returns a 400 error. As part of #2788, I may adjust the controller to redirect to the main question page with an error message.

## Release notes:

Fix issue where UI could allow accidental archival of questions that were still referenced.

### Checklist

### Issue(s)

#2788

### DIFF
--- a/server/app/services/question/QuestionServiceImpl.java
+++ b/server/app/services/question/QuestionServiceImpl.java
@@ -118,8 +118,9 @@ public final class QuestionServiceImpl implements QuestionService {
     }
     ActiveAndDraftQuestions activeQuestions =
         readOnlyQuestionService().getActiveAndDraftQuestions();
-    if (activeQuestions.getDeletionStatus(question.get().getQuestionDefinition().getName())
-        != DeletionStatus.PENDING_DELETION) {
+    if (!activeQuestions
+        .getDeletionStatus(question.get().getQuestionDefinition().getName())
+        .equals(DeletionStatus.PENDING_DELETION)) {
       throw new InvalidUpdateException("Question is not restorable.");
     }
     Version draftVersion = versionRepositoryProvider.get().getDraftVersion();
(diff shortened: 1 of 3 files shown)


server/app/services/question/QuestionServiceImpl.java
```diff
@@ -112,6 +116,12 @@ public void restoreQuestion(Long id) throws InvalidUpdateException {
     if (question.isEmpty()) {
       throw new InvalidUpdateException("Did not find question.");
     }
+    ActiveAndDraftQuestions activeQuestions =
+        readOnlyQuestionService().getActiveAndDraftQuestions();
+    if (activeQuestions.getDeletionStatus(question.get().getQuestionDefinition().getName())
+        != DeletionStatus.PENDING_DELETION) {
```
bion, line 122: suggestion (blocking): use .equals for enums

server/app/services/question/QuestionServiceImpl.java
```diff
@@ -126,6 +136,12 @@ public void archiveQuestion(Long id) throws InvalidUpdateException {
     if (question.isEmpty()) {
       throw new InvalidUpdateException("Did not find question.");
     }
+    ActiveAndDraftQuestions activeQuestions =
+        readOnlyQuestionService().getActiveAndDraftQuestions();
+    if (activeQuestions.getDeletionStatus(question.get().getQuestionDefinition().getName())
+        != DeletionStatus.DELETABLE) {
```
bion, line 142: suggestion (blocking): use .equals for enums

bion (changes requested):
Does the UI allow a way of doing this? If so we should update the UI as well

clouser2:
> Does the UI allow a way of doing this? If so we should update the UI as well

Not presently assuming everything occurs in a single tab. Multi-tab scenarios would allow this though. In a future PR, I'll likely redirect to the question page with an error message in the flash.

server/app/services/question/QuestionServiceImpl.java
clouser2: Done.

server/app/services/question/QuestionServiceImpl.java
clouser2: Done.

bion (approved):

PR #2915, opened 2022-07-18
Redirect to draft question when trying to edit published.

### LINKED ISSUE #2497: Question -> New Version can be done multiple times overwriting previous ones.
**Describe the bug**
Creating a New Version of an active question can be done multiple times with each subsequent one replacing previous changes.

**To Reproduce**
1. Open the Admin question page in 2 tabs
2. In Tab 1 click New Version on a question.  Make changes to that questions description then Save
Notice: the changes are shown in the Question view.
3. IN Tab 2, repeat Step 2
4. Reload tab 1s view of the Questions and notice it's changes no longer show.

The database will correctly reflect 1 "draft" question

**Expected behavior**
When New Version is selected and there's a draft in progress, load the existing draft instead of trying to make a new one.

There are very few Admin's and their mental model will want the existing draft and not to see an error that they need to reload first.

nbeloglazov:
### Description

It's possible for global admins to edit question in parallel: in one tab create a new version of an existing question and then, in the second tab, try to edit the original question. Currently if that happens the admin sees the original question data even though a new draft version exists. This PR handles this case and redirects admin to the draft version.

## Release notes:

Redirect to the draft version if exists when trying to edit a published question.

### Checklist

### Issue(s)

Fixes #2497

### DIFF
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -272,7 +272,8 @@ describe('normal question lifecycle', () => {
 
     // Update the question to create new draft version.
     await adminQuestions.gotoQuestionNewVersionPage('name-q')
-    // Save edit url as later we'll use it to imitate user having multiple tabs open.
+    // The ID in the URL after clicking new version corresponds to the active question form (e.g. ID=15).
+    // After a draft is created, the ID will reflect the newly created draft version (e.g. ID=16).
     const editUrl = page.url()
     const newQuestionText = await adminQuestions.updateQuestionText(
       'second version',
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -258,7 +258,6 @@ describe('normal question lifecycle', () => {
 
   it('redirects to draft question when trying to edit original question', async () => {
     const {page} = await startSession()
-    page.setDefaultTimeout(4000)
     await loginAsAdmin(page)
 
     const adminQuestions = new AdminQuestions(page)
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -255,4 +255,36 @@ describe('normal question lifecycle', () => {
       ),
     ).toBeTruthy()
   })
+
+  it('redirects to draft question when trying to edit original question', async () => {
+    const {page} = await startSession()
+    page.setDefaultTimeout(4000)
+    await loginAsAdmin(page)
+
+    const adminQuestions = new AdminQuestions(page)
+    const adminPrograms = new AdminPrograms(page)
+
+    await adminQuestions.gotoAdminQuestionsPage()
+    await adminQuestions.addNameQuestion({questionName: 'name-q'})
+
+    const programName = 'test program'
+    await adminPrograms.addProgram(programName)
+    await adminPrograms.publishProgram(programName)
+
+    // Update the question to create new draft version.
+    await adminQuestions.gotoQuestionNewVersionPage('name-q')
+    // Save edit url as later we'll use it to imitate user having multiple tabs open.
+    const editUrl = page.url()
+    const newQuestionText = await adminQuestions.updateQuestionText(
+      'second version',
+    )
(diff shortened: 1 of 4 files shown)


server/test/controllers/admin/AdminQuestionControllerTest.java
```diff
@@ -188,6 +188,24 @@ public void edit_invalidIDReturnsBadRequest() {
...
   }
 
+  @Test
+  public void edit_returnsRedirectWhenEditingQuestionWithExistingDraft() {
+    Question originalNameQuestion = testQuestionBank.applicantName();
+    Question draftNameQuestion =
+        testQuestionBank.maybeSave(
+            this.createNameQuestionDuplicate(originalNameQuestion), LifecycleStage.DRAFT);
+
+    // sanity check that thew new question has different id.
```
jeanniefu, line 198: nit: typo

jeanniefu (approved):
Thanks for this fix!

browser-test/src/question_lifecycle.test.ts
```diff
@@ -255,4 +255,36 @@ describe('normal question lifecycle', () => {
       ),
     ).toBeTruthy()
   })
+
+  it('redirects to draft question when trying to edit original question', async () => {
+    const {page} = await startSession()
+    page.setDefaultTimeout(4000)
```
clouser2, line 261: Is the timeout override necessary?

server/test/controllers/admin/AdminQuestionControllerTest.java
nbeloglazov: Thanks, done.

browser-test/src/question_lifecycle.test.ts
```diff
@@ -255,4 +255,36 @@ describe('normal question lifecycle', () => {
...
+    await adminQuestions.addNameQuestion({questionName: 'name-q'})
+
+    const programName = 'test program'
+    await adminPrograms.addProgram(programName)
+    await adminPrograms.publishProgram(programName)
+
+    // Update the question to create new draft version.
+    await adminQuestions.gotoQuestionNewVersionPage('name-q')
+    // Save edit url as later we'll use it to imitate user having multiple tabs open.
+    const editUrl = page.url()
```
clouser2, line 277: Should we navigate to the question page and grab the URL prior to going to the new version page? As this currently reads, it makes me think we're saving off the URL of the new question version rather than the old question version.

browser-test/src/support/admin_questions.ts
```diff
@@ -163,15 +163,26 @@ export class AdminQuestions {
     expect(tableInnerText).not.toContain(questionName)
   }
 
-  async gotoQuestionEditPage(questionName: string) {
+  private async gotoQuestionEditOrNewVersionPage(
+    questionName: string,
```
clouser2, line 167: Not blocking and a style nit: Since there are two arguments of the same type, it's pretty easy to get argument order mixed up. What about destructuring so that callers need to pass in "named arguments"?

clouser2 (changes requested):
Looks good. Only questions are around whether the browser test is actually exercising the new codepath.

browser-test/src/question_lifecycle.test.ts
nbeloglazov: Hmm. Assuming that published question id is `10` then "new version url" will be `/admin/questions/10/edit` which is the same as "old question edit url".

browser-test/src/support/admin_questions.ts
nbeloglazov: Done.

browser-test/src/question_lifecycle.test.ts
nbeloglazov: Maybe not. Copied it from another test. Removed it and let's see if it becomes flaky.

browser-test/src/question_lifecycle.test.ts
clouser2: Right you are. That's not entirely obvious. Would you mind updating the comment to reflect this, maybe something like:
The ID in the URL after clicking new version corresponds to the active question form. After a draft is created, the ID will reflect the newly created draft version.

clouser2 (approved):

PR #2917, opened 2022-07-18
Use secret manager in Fargate

### LINKED ISSUE #2873: Use secrets from secret manager in deploy
We should be able to specify ARN of the secret in the config.

azizova-leyla:
### Description

Switch to connect to secrets manager directly, instead of passing values in.
Caveats: 
* Secret values must be present for tasks to start up.
* Setting desired count to 0 for now, so this will NOT bring the server up. Set it higher to test things.
* More work to be done here

## Release notes:

In AWS deploy scripts use Secret Manager to provide values to Fargate instance. 

### Checklist

### Issue(s)

Fixes #2873


cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -34,13 +35,88 @@ module "td" {
...
+    {
+      name      = "ADFS_SECRET"
+      valueFrom = aws_secretsmanager_secret_version.adfs_secret_secret_version.arn
+    },
+    {
+      name      = "ADFS_CLIENT_ID"
+      valueFrom = aws_secretsmanager_secret_version.applicant_oidc_client_id_secret_version.arn
+    },
+    {
+      name      = "APPLICANT_OIDC_DISCOVERY_URI"
```
kmcnellis, line 103: Discovery URI and client ID don't need to be a secret - they are also shared with anyone who tries to login, so they are very easy to discover.

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -96,8 +165,9 @@ module "td" {
 }
 
 module "ecs_fargate_service" {
-  source      = "cn-terraform/ecs-fargate-service/aws"
-  name_prefix = var.app_prefix
+  source        = "cn-terraform/ecs-fargate-service/aws"
+  name_prefix   = var.app_prefix
+  desired_count = 0 # TODO: set this to actual value
```
kmcnellis, line 170: Why are we setting this to 0 / not enabling it to actually bring up servers?  It seems a little confusing to do that, if all that it takes to actually get a server is to set this to non-zero.  It's not like anyone is using these scripts for anything but testing right now

kmcnellis (approved):

cloud/aws/templates/aws_oidc/app.tf
azizova-leyla: because if the task fails it will retry indefinitely, blowing through dockerhub quota for pulling images. 
We need to stop it from retrying indefinitely and then set this to a reasonable number.

cloud/aws/templates/aws_oidc/app.tf
azizova-leyla: missed this comment. Agree, those don't need to be secrets.

PR #2924, opened 2022-07-19
Only build dependency docker images when they change, simplify flags

kmcnellis:
### Description

Update the CI infra to:
- Build the dependency docker images when they change
- Build in separate actions
- Build the dev-oidc image
- Simplify the flags for pushing to dockerhub (make them opt-in only)

### Checklist

- [ ] Tested as much as possible using the "on pull_request" hook


.github/workflows/build.yaml
```diff
@@ -0,0 +1,149 @@
+name: ci
+
+on:
```
azizova-leyla, line 3: this file will still push the image, I think it will be easier to understand if this was "push" and what's left in the other file was "test_deploy"

azizova-leyla (commented):
Thank you so much for making push optional. I did not like commenting out things.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,149 @@
...
+        run: bin/build-dev
+
+  build_formatter:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - id: file_change
+        uses: trilom/file-changes-action@v1.2.4
+        if: ${{ github.event_name == 'push' }}
```
bion, line 62: issue (blocking): here and with the other jobs. the file_change step only runs if the event is push, but the later step that depends on output from file_change runs if the event is not push

bion (changes requested):

.github/workflows/build.yaml
kmcnellis: The later step is `        if: ${{ github.event_name != 'push' }} ||  contains(toJSON(steps.file_changes.outputs.files), 'formatter/')`

AKA, if it's not a push event (it was manually triggered), always build the container

kmcnellis: Added a comment

.github/workflows/build.yaml
kmcnellis: Renamed

azizova-leyla (approved):

Issue #2752, opened 2022-06-16
Status Tracking: Create Program status configuration UI


shanemc-goog:
- [x] Somewhere in the Program editor, provide an area to add/remove/edit statuses. Mocks TBD
- [ ] Upon clicking a Submit button send the configuration to the handler in #2751 

Note: This should be gated on the Status Tracking Feature Flag.  All other status tracking features will trigger based on their being status configuration and not on the feature flag also; so this will serve as the main control for enabling statuses.

clouser2:
The main work here is complete with assorted TODOs to clean up.

PR #2796, opened 2022-06-24
Add statuses listing in civiform admin UI

clouser2:
### Description

This is not currently populated by any real data, nor are there any backing handlers for the create/edit view. This will be added in future PRs. All UI is guarded behind the `application_status_tracking_enabled` feature flag.

Manage status page:
Screen Shot 2022-06-27 at 9 22 48 AM

Create status popup:
Screen Shot 2022-06-27 at 9 22 58 AM

Edit status popup:
Screen Shot 2022-06-27 at 9 23 07 AM

Entrypoint for updated program index view:
Screen Shot 2022-06-27 at 9 23 23 AM

Entrypoint for existing program index view (this will hopefully be cleaned up prior to application status tracking launching):
Screen Shot 2022-06-27 at 9 31 01 AM

Release notes: N/A

### Checklist
- [X] Created tests which fail without the change (if possible)

### Issue(s)
#2752

PR #2938, opened 2022-07-19
CiviForm admin handlers for creating, editing, and deleting program statuses

clouser2:
### Description

This PR adds basic create / edit / delete semantics for the CiviFom admin program status editing experience. Status editing / creating happens as a modal popup and form validation errors are displayed on the popped up form.

Note: Presently, when submitting a form, a POST request is made to the `/statuses` URL. Upon rendering the page, the relevant modal will display errors, but won't be displayed by default. I left a TODO for that since it's definitely a necessary change, but out of the scope of this work.

Next up, I'll add browser tests for the functionality.

Delete modal screenshot:
Screen Shot 2022-07-20 at 2 02 02 PM

Manage status page:
Screen Shot 2022-06-27 at 9 22 48 AM

Create status popup:
Screen Shot 2022-06-27 at 9 22 58 AM

Edit status popup:
Screen Shot 2022-06-27 at 9 23 07 AM

## Release notes:

N/A.

### Checklist

### Issue(s)

#2752


server/conf/routes
```diff
@@ -25,6 +25,8 @@ POST    /admin/programs/:programId/admins       controllers.admin.ProgramAdminMa
 
 # Routes for a CiviForm admin to manage application statuses for a given program.
 GET     /admin/programs/:programId/statuses controllers.admin.AdminProgramStatusesController.index(request: Request, programId: Long)
+POST    /admin/programs/:programId/statuses controllers.admin.AdminProgramStatusesController.index(request: Request, programId: Long)
```
shanemc-goog, line 28: I don't think "index" actions should be POSTs, as well we shouldn't double up GET/POST on the same handler.  I don't think that's RESTful or something I've seen us do otherwise; though I could be wrong.

server/app/forms/admin/ProgramStatusesEditForm.java
```diff
@@ -0,0 +1,77 @@
...
+  private String emailBody;
+
+  public ProgramStatusesEditForm() {
+    this.originalStatusText = "";
+    this.statusText = "";
+    this.emailBody = "";
+  }
+
+  public String getOriginalStatusText() {
+    return originalStatusText != null ? originalStatusText : "";
```
shanemc-goog, line 34: We're using null a lot here when I don't think it is necessary; eg required by the framework or a dependency.

We should use Optional unless we can't, EG for non required DB values etc.

shanemc-goog:
Can you include screenshots for the create/edit/etc panels?

shanemc-goog (changes requested):

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,209 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+        validated = validateEditRequest(request, program);
+      } catch (MissingStatusException e) {
+        return redirect(routes.AdminProgramStatusesController.index(programId).url())
+            .flashing(
+                "error",
+                "The status being edited no longer exists and may have been modified in a separate"
+                    + " window.");
+      }
+      maybeEditForm = Optional.of(validated.getLeft());
+      if (!validated.getLeft().hasErrors() && validated.getRight().isPresent()) {
```
shanemc-goog, line 77: Unpack left/right asap and use them for readability.  Pair<> is usually discouraged as it lacks readability, the preference is typically a private AutoValue, but here unpacking it into readable left/right may work.

server/conf/routes
clouser2: Since application statuses don't have a "true" identifier that we want to put in a URL, I don't think we'd want to create anything REST-ful for statuses.

GET/POST on the same handler is definitely something I've seen done before since you're updating the entity being referred to (in this case the program's statuses). This approach falls out from the fact that:
  1) We want to have a single page with modal forms for editing individual stauses (from agreed-on UX)
  2) When form validation errors occur, it'd be unfortunate to keep the user on a different URL than they started and still render the same content (e.g. URL is `/update` but still renders full index page
  3) REST-ful APIs in fact would use the same collection identifier with different HTTP verbs to indicate whether it corresponded to a CREATE or UPDATE (https://restfulapi.net/http-methods/). In this case, we merge CREATE / UPDATE into a single `POST` rather than having an overload that accepts a request with verb `PATCH`.

Let me know what you think. I'm also happy to create two separate `AdminProgramStatusesController.[index|update]` methods if you think that'd aid in readability. Under the hood, they'll still call largely the same underlying logic in the controller and the same view.

clouser2:
>

clouser2:
> Can you include screenshots for the create/edit/etc panels?

Sure. I've re-included the screenshots from the original #2796.

server/app/forms/admin/ProgramStatusesEditForm.java
clouser2: Empirically, the framework returns `null` if the form parameter is missing entirely from the request body ([docs](https://www.playframework.com/documentation/2.8.x/JavaForms)). We aren't in control of how this data is populated since we're using the `form(...)` helper and binding from the request.

clouser2 (commented):
Attempted to answer the larger feedback questions. I'll fix the other two comments shortly.

server/app/controllers/admin/AdminProgramStatusesController.java
clouser2: Good suggestion for readability. My general preference is discouraging Pair<> for external APIs and that it's fine to use for internal private methods (assuming the pair isn't of the same type). If we desire AutoValue, I can look into that, but I'm not sure it provides a lot of benefit here.

I'd love it if there were easier lighter-weight ways to do this in Java.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,209 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+        Result result = redirect(routes.AdminProgramStatusesController.index(programId).url());
+        ErrorAnd<ProgramDefinition, CiviFormError> setStatusResult =
+            service.setStatuses(programId, validated.getRight().get());
+        if (setStatusResult.isError()) {
+          String errorMessage = joinErrors(setStatusResult.getErrors());
+          result = result.flashing("error", errorMessage);
```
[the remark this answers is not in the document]
clouser2: That's more readable, thanks!

clouser2 (commented):
Feedback incorporated. Still need to iron out if the high-level design needs adjustment or not.

server/conf/routes
clouser2: Note that we could use another URL to support the POST, but we'd eventually want to redirect back to the GET version of the URL. Performing this redirect would mean we'd lose the POST data and we'd need to preserve it by putting it in the `.flash` ([example](https://github.com/seattle-uat/civiform/blob/cfbc4a9eff6099cf7adf6c2283f55c037eb033dd/server/app/controllers/ti/TrustedIntermediaryController.java#L136-L143)).

This approach felt more error-prone to me if we end up updating what POST data is required / etc. My belief is that very very few people outside of devs on a project are actually looking at the specific URL as they're submitting forms, so I want to make sure there's a clear win for the user here.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
       return notFound("status tracking is not enabled");
     }
-    return ok(statusesView.render(service.getProgramDefinition(programId)));
+    requestChecker.throwIfProgramNotDraft(programId);
+    ProgramDefinition program = service.getProgramDefinition(programId);
+    int previousStatusCount = program.statusDefinitions().getStatuses().size();
+
+    Optional<Form<ProgramStatusesEditForm>> maybeEditForm = Optional.empty();
+    if (request.method().equalsIgnoreCase(HttpVerbs.POST)) {
+      Form<ProgramStatusesEditForm> editForm;
```
shanemc-goog, line 66: Stick a final keyword in front of these to ensure they're set before reading.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+      try {
+        Pair<Form<ProgramStatusesEditForm>, Optional<StatusDefinitions>> validated =
+            validateEditRequest(request, program);
+        editForm = validated.getLeft();
+        updatedStatus = validated.getRight();
+      } catch (MissingStatusException e) {
+        return redirect(routes.AdminProgramStatusesController.index(programId).url())
+            .flashing(
+                "error",
+                "The status being edited no longer exists and may have been modified in a separate"
```
shanemc-goog, line 77: This will show as a toast?  Can we give the user something actionable to do to resolve the issue? EG reload the page.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
     if (!statusTrackingEnabled) {
       return notFound("status tracking is not enabled");
     }
-    return ok(statusesView.render(service.getProgramDefinition(programId)));
+    requestChecker.throwIfProgramNotDraft(programId);
+    ProgramDefinition program = service.getProgramDefinition(programId);
+    int previousStatusCount = program.statusDefinitions().getStatuses().size();
+
+    Optional<Form<ProgramStatusesEditForm>> maybeEditForm = Optional.empty();
+    if (request.method().equalsIgnoreCase(HttpVerbs.POST)) {
```
shanemc-goog, line 65: suggestion: Invert this and return early so we can unnest what is most of the method.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+            previousStatusCount == updatedStatus.get().getStatuses().size()
+                ? "Status updated"
+                : "Status created");
+      }
+    }
+
+    return ok(statusesView.render(request, service.getProgramDefinition(programId), maybeEditForm));
+  }
+
+  private Pair<Form<ProgramStatusesEditForm>, Optional<StatusDefinitions>> validateEditRequest(
```
shanemc-goog, line 100: Please doc this. The method is doing a lot and returning data (not just validating) but it's unclear what that is.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+    Form<ProgramStatusesEditForm> form =
+        formFactory
+            .form(ProgramStatusesEditForm.class)
+            .bindFromRequest(request, ProgramStatusesEditForm.FIELD_NAMES.toArray(new String[0]));
+    if (form.value().isEmpty()) {
+      return Pair.of(form, Optional.empty());
+    }
+    ProgramStatusesEditForm value = form.value().get();
+    StatusDefinitions current = program.statusDefinitions();
+    if (value.getOriginalStatusText().isEmpty()) {
```
shanemc-goog, line 111: This means we're creating a new status right?  Please comment that.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+      Http.Request request, ProgramDefinition program) throws MissingStatusException {
+    Form<ProgramStatusesEditForm> form =
+        formFactory
+            .form(ProgramStatusesEditForm.class)
+            .bindFromRequest(request, ProgramStatusesEditForm.FIELD_NAMES.toArray(new String[0]));
+    if (form.value().isEmpty()) {
+      return Pair.of(form, Optional.empty());
+    }
+    ProgramStatusesEditForm value = form.value().get();
+    StatusDefinitions current = program.statusDefinitions();
```
shanemc-goog, line 110: nit: current -> currentDefs/Definitions

there's a few things that could be current that I think this aids reading.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+      } catch (DuplicateStatusException e) {
+        form = form.withError(ProgramStatusesEditForm.STATUS_TEXT_FORM_NAME, e.userFacingMessage);
+      }
+    } else {
+      try {
+        current.setStatuses(
+            replaceStatus(
+                current.getStatuses(),
+                value.getOriginalStatusText(),
+                (existingStatus) -> {
```
shanemc-goog, line 133: This feels a bit convoluted as well this is the only call site so it's unclear why a function param is needed.

Passing in value, and having replaceStatus do this seems a lot cleaner, or passing in the 3 value data items explicitly which only adds 1 param.

replaceStatus(current.getStatuses(), value);

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+    return ImmutableList.<StatusDefinitions.Status>builder()
+        .addAll(statuses)
+        .add(newStatus)
+        .build();
+  }
+
+  private static ImmutableList<StatusDefinitions.Status> replaceStatus(
+      ImmutableList<StatusDefinitions.Status> statuses,
+      String replaceStatusName,
+      Function<StatusDefinitions.Status, StatusDefinitions.Status> replacerFunc)
```
shanemc-goog, line 210: This method and this param need javadocs

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+        .map(i -> statuses.get(i))
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  private static ImmutableMap<String, Integer> statusNameToIndexMap(
+      ImmutableList<StatusDefinitions.Status> statuses) {
+    return IntStream.range(0, statuses.size())
+        .boxed()
+        .collect(
+            ImmutableMap.toImmutableMap(i -> statuses.get(i).statusText(), i -> i, (s1, s2) -> s1));
```
shanemc-goog, line 253: Why allow for duplicates [ s1, s2 -> s1]  and silently throwing them away?

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+    if (!statusNameToIndex.containsKey(replaceStatusName)) {
+      throw new MissingStatusException();
+    }
+    StatusDefinitions.Status editedStatus =
+        replacerFunc.apply(statuses.get(statusNameToIndex.get(replaceStatusName)));
+    if (!replaceStatusName.equals(editedStatus.statusText())
+        && statusNameToIndex.containsKey(editedStatus.statusText())) {
+      throw new DuplicateStatusException(editedStatus.statusText());
+    }
+    return IntStream.range(0, statuses.size())
```
shanemc-goog, line 222: I think it's a lot cleaner to just copy the ImmutableList to a List, replace the index, then convert back

shanemc-goog (changes requested):

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -34,6 +57,210 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
+      } catch (DuplicateStatusException e) {
+        form = form.withError(ProgramStatusesEditForm.STATUS_TEXT_FORM_NAME, e.userFacingMessage);
+      }
+    }
+
+    return form.hasErrors() ? Pair.of(form, Optional.empty()) : Pair.of(form, Optional.of(current));
+  }
+
+  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  public Result delete(Http.Request request, long programId) throws ProgramNotFoundException {
```
shanemc-goog, line 153: Please doc public methods, especially when their an API for the frontend.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -4,28 +4,51 @@
...
+import services.CiviFormError;
+import services.ErrorAnd;
+import services.LocalizedStrings;
+import services.program.ProgramDefinition;
 import services.program.ProgramNotFoundException;
 import services.program.ProgramService;
+import services.program.StatusDefinitions;
 import views.admin.programs.ProgramStatusesView;
 
 public final class AdminProgramStatusesController extends CiviFormController {
```
shanemc-goog, line 33: While I'm putting the comment on the class, I think the controller needs to document its API better.  EG nothing declares what Adding a status does (eg append), or replace (maintains position, and changes the admin visible text / non-translated content)

clouser2 (commented):
Thanks for all of the feedback, I think things are more readable! Hopefully I've incorporated the majority :D

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -242,84 +287,60 @@ private static String randomModalId() {
   }
 
   /**
-   * Returns the form if it matches the current status object being rendered. The page has multiple
-   * inline modals for editing and creating statuses.
+   * Returns whether the form matches the current status object being rendered. The page has
+   * multiple inline modals for editing and creating statuses.
    */
-  private Optional<Form<ProgramStatusesEditForm>> formForCurrentStatus(
+  private boolean isFormForCurrentStatus(
```
shanemc-goog, line 293:
```suggestion
  private static boolean isFormForCurrentStatus(
```

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -47,31 +48,45 @@ public final class ProgramStatusesView extends BaseHtmlView {
...
-      Optional<Form<ProgramStatusesEditForm>> maybeEditForm) {
-    Modal createStatusModal =
-        makeStatusEditModal(
-            request,
-            program,
-            Optional.empty(),
-            formForCurrentStatus(Optional.empty(), maybeEditForm));
+      Optional<Form<ProgramStatusesForm>> maybeStatusForm) {
+    final Form<ProgramStatusesForm> createStatusForm;
+    if (isFormForCurrentStatus(Optional.empty(), maybeStatusForm)) {
```
shanemc-goog, line 76: isFormForCurrentStatus doesn't need to accept an Optional Status as it will always return false, and as such doesn't need to be called in this situation. 

That said, this if will never be true, so what's up with the if/else?

shanemc-goog (changes requested):

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -52,215 +51,131 @@ public AdminProgramStatusesController(
...
-    return form.hasErrors() ? Pair.of(form, Optional.empty()) : Pair.of(form, Optional.of(current));
+    return ok(
+        statusesView.render(
+            request,
+            service.getProgramDefinition(programId),
+            /* maybeStatusForm= */ Optional.empty()));
   }
 
+  /**
+   * Creates a new status or updates an existing status associated with the program. Creation is
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -52,215 +51,131 @@ public AdminProgramStatusesController(
...
+    final String indexUrl = routes.AdminProgramStatusesController.index(programId).url();
+    if (mutateResult.isError()) {
+      return redirect(indexUrl).flashing("error", joinErrors(mutateResult.getErrors()));
     }
-
-    return result;
+    return redirect(indexUrl)
+        .flashing(
+            "success",
+            previousStatusCount == mutateResult.getResult().statusDefinitions().getStatuses().size()
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -52,215 +51,131 @@ public AdminProgramStatusesController(
...
+      throws ProgramNotFoundException, DuplicateStatusException {
+    // An empty "configuredStatusText" parameter indicates that a new
+    // status should be created.
+    if (formData.getConfiguredStatusText().isEmpty()) {
+      return service.appendStatus(
+          program.id(),
+          StatusDefinitions.Status.builder()
+              .setStatusText(formData.getStatusText())
+              .setLocalizedStatusText(LocalizedStrings.withDefaultValue(formData.getStatusText()))
+              .setEmailBodyText(formData.getEmailBody())
```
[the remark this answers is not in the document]
clouser2: Done for now. We should talk about this separately. I'm not sure I'm seeing clear value to having email be marked as optional when an empty string signifies an unset value. It'll only complicate our logic for email sending, where we wouldn't want to send an empty email anyway.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -52,215 +51,131 @@ public AdminProgramStatusesController(
...
+              .setLocalizedEmailBodyText(
+                  Optional.of(LocalizedStrings.withDefaultValue(formData.getEmailBody())))
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -52,215 +51,131 @@ public AdminProgramStatusesController(
...
-        .boxed()
-        .collect(
-            ImmutableMap.toImmutableMap(i -> statuses.get(i).statusText(), i -> i, (s1, s2) -> s1));
-  }
-
-  private static final class MissingStatusException extends Exception {}
-
-  private static final class DuplicateStatusException extends Exception {
-    final String userFacingMessage;
+    final String deleteStatusText = rawDeleteStatusText.trim();
```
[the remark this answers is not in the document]
clouser2: Good catch. Agreed with "sneaking in as a real status". We likely could move to a model where we trim the status entered by the user in the form before serializing, but always use the raw value from the DB in the relevant fields. That would allow us to still perform the operation for "bad" data while having the controllers try to put the "right" value in.

I left a TODO for this since it's pretty low priority in my opinion and I'd like to iterate on smaller PRs once the initial functionality is in (if that's acceptable from your end).

server/app/forms/admin/ProgramStatusesForm.java
```diff
@@ -0,0 +1,83 @@
...
+  public String getConfiguredStatusText() {
+    return checkNotNull(configuredStatusText);
+  }
+
+  public void setConfiguredStatusText(String value) {
+    configuredStatusText = checkNotNull(value).trim();
+  }
+
+  public String getStatusText() {
+    return checkNotNull(statusText);
```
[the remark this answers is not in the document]
clouser2: Removed here and elsewhere.

server/app/services/program/DuplicateStatusException.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * DuplicateStatusException is thrown when an {@link StatusDefinitions.Status} with the same
+ * statusText already exists in the list of a program's configured statuses for application review.
+ */
+public final class DuplicateStatusException extends Exception {
+  private final String userFacingMessage;
+
+  public DuplicateStatusException(String statusName) {
+    super();
```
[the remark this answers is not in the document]
clouser2: I'm not clear what you mean by using an exception as data?

Exception's default constructor would have an empty message. My goal in creating a distinct userFacingMessage was to make it clear that the content shown was acceptable to be seen by the user. If you'd prefer, I can go the traditional approach and use the exception message when displaying to the user.

server/app/services/program/StatusDefinitions.java
```diff
@@ -77,7 +77,7 @@ public abstract static class Builder {
       public abstract Builder setEmailBodyText(String value);
 
       @JsonProperty("email_body_localized")
-      public abstract Builder setLocalizedEmailBodyText(LocalizedStrings value);
+      public abstract Builder setLocalizedEmailBodyText(Optional<LocalizedStrings> value);
```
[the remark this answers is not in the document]
clouser2: We no longer call this with Optional.empty() from the controllers.

server/app/views/admin/programs/ProgramStatusesView.java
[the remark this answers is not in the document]
clouser2: Done.

clouser2: It does in the case that we're rendering the creation form. There isn't a configured status (e.g. Optional.empty()) and there is a form that has the configuredStatusText field set to empty.

> That said, this if will never be true, so what's up with the if/else?
It will in the situation described above.

server/app/views/admin/programs/ProgramStatusesView.java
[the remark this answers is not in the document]
clouser2: Done.

clouser2: Done.

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -242,84 +287,60 @@ private static String randomModalId() {
...
+   * Returns whether the form matches the current status object being rendered. The page has
+   * multiple inline modals for editing and creating statuses.
    */
-  private Optional<Form<ProgramStatusesEditForm>> formForCurrentStatus(
+  private boolean isFormForCurrentStatus(
       Optional<StatusDefinitions.Status> maybeStatus,
-      Optional<Form<ProgramStatusesEditForm>> maybeEditForm) {
-    String wantOriginalStatusText =
+      Optional<Form<ProgramStatusesForm>> maybeEditForm) {
+    String wantconfiguredStatusText =
```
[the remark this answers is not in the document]
clouser2: Done.

clouser2 (commented):
Feedback incorporated.

server/app/views/admin/programs/ProgramStatusesView.java
shanemc-goog: I thin it'd be a lot cleaner to have a method that knowingly determines if it's a create action rather than encoding create as empty/"" value, and then isFormForCurrentStatus does a check for an actual value.

boolean isCreationForm(form) {
  return maybeEditForm.get().value().get().getConfiguredStatusText().isEmpty();
}

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
shanemc-goog: I don't think we want to store empty strings and empty localization in the DB.

"" representing missing content is seemingly a Form artifact and not a necessity or desire, can we not account for that in the Form (the getters return Optional?) , so no down stream code has to care about those FE->BE semantics and specifically has to address that they are optional value?

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -347,21 +352,99 @@ public ProgramDefinition moveBlock(
...
+          ImmutableSet.of(
+              CiviFormError.of(
+                  "The status being edited no longer exists and may have been modified in a"
+                      + " separate window.")));
+    }
+    List<StatusDefinitions.Status> statusesCopy =
+        Lists.newArrayList(program.statusDefinitions().getStatuses());
+    StatusDefinitions.Status editedStatus =
+        statusReplacer.apply(statusesCopy.get(statusNameToIndex.get(toReplaceStatusName)));
+    if (!toReplaceStatusName.equals(editedStatus.statusText())
```
shanemc-goog, line 401:
```suggestion
    // If the status name was changed and it matches another status, issue an error.
    if (!toReplaceStatusName.equals(editedStatus.statusText())
```

shanemc-goog (changes requested):
Looking good, just a few small things.

server/app/services/program/DuplicateStatusException.java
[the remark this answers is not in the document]
clouser2: Done, though with a bit awkwardness since the call to `super()` needs to be first.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -125,29 +122,35 @@ private ErrorAnd<ProgramDefinition, CiviFormError> createOrEditStatusFromFormDat
...
-              .setLocalizedEmailBodyText(
-                  Optional.of(LocalizedStrings.withDefaultValue(formData.getEmailBody())))
-              .build());
+              .setLocalizedStatusText(LocalizedStrings.withDefaultValue(formData.getStatusText()));
+      if (!formData.getEmailBody().isBlank()) {
+        newStatusBuilder =
+            newStatusBuilder
+                .setEmailBodyText(formData.getEmailBody())
+                .setLocalizedEmailBodyText(
+                    Optional.of(LocalizedStrings.withDefaultValue(formData.getEmailBody())));
```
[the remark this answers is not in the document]
clouser2: Still needed since the builder doesn't have a non-optional overload, and adding one caused the issues we discussed prior with Jackson.

server/app/services/program/ProgramServiceImpl.java
clouser2: Done.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -347,21 +352,99 @@ public ProgramDefinition moveBlock(
...
+    if (!statusNameToIndex.containsKey(toRemoveStatusName)) {
+      return ErrorAnd.error(
+          ImmutableSet.of(
+              CiviFormError.of(
+                  "The status being deleted no longer exists and may have been deleted in a"
+                      + " separate window.")));
+    }
+    List<StatusDefinitions.Status> statusesCopy =
+        Lists.newArrayList(program.statusDefinitions().getStatuses());
+    statusesCopy.remove((int) statusNameToIndex.get(toRemoveStatusName));
```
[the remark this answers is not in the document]
clouser2: It was a need to unbox. Thanks for calling out. I was definitely sad about that one.

server/app/services/program/StatusDefinitions.java
```diff
@@ -36,9 +39,20 @@ public ImmutableList<Status> getStatuses() {
    * <p>The order of the items will be maintained and used as the natural order of the statuses.
    */
   public void setStatuses(ImmutableList<Status> statuses) {
+    assertStatusNamesNonEmptyAndUnique(statuses);
     this.statuses = statuses;
   }
 
+  private static void assertStatusNamesNonEmptyAndUnique(ImmutableList<Status> statuses) {
+    Preconditions.checkState(
+        statuses.stream().map(Status::statusText).collect(ImmutableSet.toImmutableSet()).size()
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -287,16 +287,16 @@ private static String randomModalId() {
   }
 
   /**
-   * Returns whether the form matches the current status object being rendered. The page has
-   * multiple inline modals for editing and creating statuses.
+   * Returns whether the form matches the status object being rendered. The page has multiple inline
+   * modals for editing and creating statuses.
    */
-  private boolean isFormForCurrentStatus(
+  private static boolean isFormForStatus(
```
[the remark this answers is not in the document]
clouser2: Swapped order. The form is also optional in the case of rendering the initial view. Something that may be able to be cleaned up in a future iteration.

clouser2 (commented):
Feedback incorporated.

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
clouser2: That's not what's happening here (please see the comment above). Empty string for "configuredStatusText" is our signifier that this represents the create form. As such, we're not storing empty strings in the DB. Additionally, we perform validation that the actual status text is non-empty in the validation for the form itself.

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
shanemc-goog: You'd said that we no longer call `setLocalizedEmailBodyText `with Optional.empty, so doesn't that mean that the setter doesn't need an optional form which was added in this PR?

https://github.com/civiform/civiform/pull/2938#discussion_r931666519
"We no longer call this with Optional.empty() from the controllers"

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
shanemc-goog: Sorry, to be clear I meant the email text and its localization.  We ideally don't want to store empty representations of them when they aren't set, and if we could have the Form provide them as appropriately created Optional that seems nice.

Not sure if we're miscommunicating here

server/app/views/admin/programs/ProgramStatusesView.java
clouser2: I have a slight preference for a single method (so that the matching logic is consolidated to a single method). In my opinion, that potentially reduces the maintenance burden if we decide to change semantics on what form fields mean. That said, it's nebulous and I don't feel strongly enough to keep pushing for it. Introduced the method you described.

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
clouser2: Good callout. I switched to only calling it when it's present in the existing status.

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
clouser2: Gotcha. Done.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -142,11 +142,17 @@ private ErrorAnd<ProgramDefinition, CiviFormError> createOrEditStatusFromFormDat
...
+                  // English localized text from this handler and
+                  // disabling the English translations UI.
                   // Note: We preserve the existing localized status / email body
                   // text so that existing translated content isn't destroyed upon
                   // editing status.
-                  .setLocalizedStatusText(existingStatus.localizedStatusText())
-                  .setLocalizedEmailBodyText(existingStatus.localizedEmailBodyText());
+                  .setLocalizedStatusText(existingStatus.localizedStatusText());
+          if (existingStatus.localizedEmailBodyText().isPresent()) {
+            builder =
```
shanemc-goog, line 153: you don't need builder = here.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -142,11 +142,17 @@ private ErrorAnd<ProgramDefinition, CiviFormError> createOrEditStatusFromFormDat
...
+                builder.setLocalizedEmailBodyText(existingStatus.localizedEmailBodyText().get());
+          }
           if (!formData.getEmailBody().isBlank()) {
             builder = builder.setEmailBodyText(formData.getEmailBody());
```
shanemc-goog, line 157: Don't need builder = here.

shanemc-goog (changes requested):

shanemc-goog (approved):

server/app/controllers/admin/AdminProgramStatusesController.java
clouser2: Done.

server/app/controllers/admin/AdminProgramStatusesController.java
clouser2: Done.

PR #2940, opened 2022-07-20
Omit non-server dependencies in release notes

bion:


kmcnellis (approved):
We should consider including infra changes once a city is using our scripts

PR #2942, opened 2022-07-20
Use a matrix to run CI builds

kmcnellis:
### Description

Switch to using a matrix to run dockerfile builds, to avoid as much duplicated logic as possible.  Also avoid some potential race conditions around building dependancy images only on file changes (we don't want to cancel a workflow that does build the image in favor of one that skips the build).

also update the workflow names to be more descriptive

### Checklist

- [ ]  Tested as much as possible using the "on pull_request" hook


bion (approved):
nice work!

PR #2943, opened 2022-07-20
Fix get_current_user function in AWS setup.py

nbeloglazov:
### Description

The function should return string and not Process python object. The implementation is copied from the azure implementation.


azizova-leyla (approved):

nbeloglazov:
I'm going to merge it even though azure deploy is failing. It's failing because I'm PR'ing from another repo and unrelated to the change itself.

PR #2949, opened 2022-07-21
Remove --ipc param when running browser tests.

nbeloglazov:
### Description

Having `--ipc=host` has issues for engineers using linux. Reading docs, ipc is related to memory sharing and shouldn't be needed for browser tests. Tried on macos and linux - tests work fine with the param removed.


swatkat1 (approved):

gwendolyngoetz (commented):
That's how I've run them the entire time on Ubuntu without any known problems.

PR #2954, opened 2022-07-21
If config has fewer than 2 languages don't show language selector

### LINKED ISSUE #2919: Disable language selector screen if there's only one language

**Additional context**

**Done when**

azizova-leyla:
### Description

If config has less than 2 languages don't show language selector. 
Added some tests. Those files should be better tested, but leaving that out of this CL.

## Release notes:

If application is configured to only be available in one language through CIVIFORM_SUPPORTED_LANGUAGES don't show language selector.

### Checklist

### Issue(s)

Fixes #2919

### DIFF
--- a/server/app/controllers/HomeController.java
+++ b/server/app/controllers/HomeController.java
@@ -5,6 +5,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import auth.CiviFormProfile;
 import auth.ProfileUtils;
 import java.util.Locale;
+import com.google.common.base.Strings;
+import com.typesafe.config.Config;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -28,19 +30,24 @@ public class HomeController extends Controller {
   private final MessagesApi messagesApi;
   private final HttpExecutionContext httpExecutionContext;
   private final Langs langs;
+  private final Optional<String> faviconURL;
 
   @Inject
   public HomeController(
+      Config configuration,
       LoginForm form,
       ProfileUtils profileUtils,
       MessagesApi messagesApi,
       HttpExecutionContext httpExecutionContext,
       Langs langs) {
+    checkNotNull(configuration);
     this.loginForm = checkNotNull(form);
     this.profileUtils = checkNotNull(profileUtils);
     this.messagesApi = checkNotNull(messagesApi);
     this.httpExecutionContext = checkNotNull(httpExecutionContext);
     this.langs = checkNotNull(langs);
+    this.faviconURL =
+        Optional.ofNullable(Strings.emptyToNull(configuration.getString("whitelabel.favicon_url")));
   }
 
   public CompletionStage<Result> index(Http.Request request) {
@@ -95,6 +102,16 @@ public class HomeController extends Controller {
 
   public Result playIndex() {
     return ok("public index");
+  }
+
+  // Redirect any browsers who, by default, request favicon from root, to the
+  // specified favicon link.
+  // https://stackoverflow.com/questions/56222166/prevent-browser-from-trying-to-load-favicon-from-root-directory)
+  public Result favicon() {
+    if (faviconURL.isPresent()) {
+      return found(faviconURL.get()); // http 302
+    }
+    return notFound();
   }
 
   @Secure
--- a/server/app/controllers/HomeController.java
+++ b/server/app/controllers/HomeController.java
@@ -4,7 +4,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import auth.CiviFormProfile;
 import auth.ProfileUtils;
-import java.util.Locale;
 import com.google.common.base.Strings;
 import com.typesafe.config.Config;
 import java.util.Optional;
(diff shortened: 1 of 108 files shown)


server/app/controllers/applicant/RedirectController.java
```diff
@@ -78,6 +84,13 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
               }
 
               Applicant applicant = applicantFuture.join();
+              ApplicantData data = applicant.getApplicantData();
```
nbeloglazov, line 87: Reading this method - it's not clear why we are setting preferred locale specifically here, in "programByName" action. To be setting locale sounds like something general, applied to various "entries" to the app.

server/app/controllers/HomeController.java
```diff
@@ -62,6 +67,12 @@ public CompletionStage<Result> index(Http.Request request) {
                 // If the applicant has not yet set their preferred language, redirect to
                 // the information controller to ask for preferred language.
                 ApplicantData data = applicant.getApplicantData();
+                if (langs.availables().size() <= 1 && !data.hasPreferredLocale()) {
+                  data.setPreferredLocale(
+                      langs.availables().isEmpty()
+                          ? Locale.US
+                          : langs.availables().get(0).toLocale());
+                }
```
nbeloglazov, line 75: Do we have a util file where we can share "setPreferredLocale" method with the other controller?

server/app/controllers/HomeController.java
```diff
@@ -62,6 +67,12 @@ public CompletionStage<Result> index(Http.Request request) {
                 // If the applicant has not yet set their preferred language, redirect to
                 // the information controller to ask for preferred language.
                 ApplicantData data = applicant.getApplicantData();
+                if (langs.availables().size() <= 1 && !data.hasPreferredLocale()) {
```
nbeloglazov, line 70: Hm, I think it worth a comment here saying that we are setting preferred locale when a single language is available so that step of language selection can be automatically skipped. Otherwise it's a bit unclear why is it useful to "choose" locale for user.

server/test/controllers/HomeControllerAuthenticatedTest.java
```diff
@@ -53,4 +62,56 @@ public void testAuthenticatedSecurePage() {
...
+            instanceOf(HttpExecutionContext.class),
+            mockLangs);
+    Result result =
+        controller.index(addCSRFToken(fakeRequest()).build()).toCompletableFuture().join();
+    assertThat(result.redirectLocation())
+        .contains(
+            controllers.applicant.routes.ApplicantProgramsController.index(applicant.id).url());
+  }
+
+  @Test
```
clouser2, line 98: I'd also test where the locale is set to something other than the single-configured language (edge case, but good to specify the behavior).

server/app/controllers/HomeController.java
azizova-leyla: added util file.

server/app/controllers/HomeController.java
```diff
@@ -62,6 +67,12 @@ public CompletionStage<Result> index(Http.Request request) {
...
+                  data.setPreferredLocale(
```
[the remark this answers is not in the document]
azizova-leyla: good catch. added updateApplicant in the util method.

server/test/controllers/HomeControllerAuthenticatedTest.java
azizova-leyla: done.

server/app/controllers/HomeController.java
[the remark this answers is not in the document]
azizova-leyla: I think it's okay to leave preferred locale as is for the person. It's highly unlikely that they would be able to set it to some other language (it would have to available at some point, and the removed).

server/app/controllers/applicant/RedirectController.java
azizova-leyla: this is deep link case. applicant would get a link directly to that program and this is the method that would be called. agree that the naming is not super clear.

server/app/controllers/LanguageUtils.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public Applicant maybeSetDefaultLocale(Applicant applicant) {
+    ApplicantData data = applicant.getApplicantData();
+    if (data.hasPreferredLocale() || langs.availables().size() > 1) {
+      return applicant;
+    }
+    data.setPreferredLocale(
+        langs.availables().isEmpty() ? Locale.US : langs.availables().get(0).toLocale());
+    userRepository.updateApplicant(applicant);
```
clouser2, line 29: This returns a future. We should await its result.

server/app/controllers/LanguageUtils.java
```diff
@@ -0,0 +1,32 @@
...
+    this.langs = checkNotNull(langs);
+  }
+
+  public Applicant maybeSetDefaultLocale(Applicant applicant) {
+    ApplicantData data = applicant.getApplicantData();
+    if (data.hasPreferredLocale() || langs.availables().size() > 1) {
+      return applicant;
+    }
+    data.setPreferredLocale(
+        langs.availables().isEmpty() ? Locale.US : langs.availables().get(0).toLocale());
```
clouser2, line 28: Mind reusing `LocalizedStrings.DEFAULT_LOCALE` or pulling the constant elsewhere?

clouser2 (approved):

nbeloglazov (approved):

Issue #2955, opened 2022-07-22
Enable strict typescript compilation checking for browser tests


nbeloglazov:
Browser tests contain [tsconfig.json](https://github.com/seattle-uat/civiform/blob/main/browser-test/src/tsconfig.json) file which enables strict type checking. But that file currently not wired or used. So strict type checking is not enabled. Enabling adds a bunch of errors, example:

```
src/text.test.ts:184:37 - error TS7005: Variable 'pageObject' implicitly has an 'any' type.
184       await selectApplicantLanguage(pageObject, 'English')
                                        ~~~~~~~~~~

src/text.test.ts:186:13 - error TS7005: Variable 'applicantQuestions' implicitly has an 'any' type.
186       await applicantQuestions.applyProgram(programName)
                ~~~~~~~~~~~~~~~~~~
src/text.test.ts:187:13 - error TS7005: Variable 'applicantQuestions' implicitly has an 'any' type.
187       await applicantQuestions.answerTextQuestion('I love CiviForm!', 0)
                ~~~~~~~~~~~~~~~~~~
src/text.test.ts:188:13 - error TS7005: Variable 'applicantQuestions' implicitly has an 'any' type.
188       await applicantQuestions.answerTextQuestion(
                ~~~~~~~~~~~~~~~~~~

src/text.test.ts:192:13 - error TS7005: Variable 'applicantQuestions' implicitly has an 'any' type.
192       await applicantQuestions.clickNext()
                ~~~~~~~~~~~~~~~~~~
src/text.test.ts:195:20 - error TS7005: Variable 'pageObject' implicitly has an 'any' type.
195       expect(await pageObject.innerText(textId)).toContain(

```

We should fix those errors and wire up tsconfig.json

PR #2956, opened 2022-07-22
Use tsconfig.json for TS transformation when running browser tests

nbeloglazov:
### Description

With this change browser tests will use configuration files (`package.json`, `jest.config.js`) from local directory rather than using the ones provided in the image. That makes sure that changes to configs immediately take effect. 

Additionally this change sets up `ts-jest` to use `tsconfig.json` provided in the tests. Until now `tsconfig.json` hasn't been used at all.

Related to #2955

### DIFF
--- a/bin/run-browser-tests
+++ b/bin/run-browser-tests
@@ -7,9 +7,7 @@ docker::set_project_name_browser_tests
 bin/pull-image
 
 docker run --rm -it \
-  -v "$(pwd)/browser-test/src:/usr/src/civiform-browser-tests/src" \
-  -v "$(pwd)/browser-test/bin:/usr/src/civiform-browser-tests/bin" \
-  -v "$(pwd)/browser-test/tmp:/usr/src/civiform-browser-tests/tmp" \
+  -v "$(pwd)/browser-test:/usr/src/civiform-browser-tests" \
   -e RECORD_VIDEO="${RECORD_VIDEO}" \
   --network "${DOCKER_NETWORK_NAME}" \
   civiform/civiform-browser-test:latest \
--- a/browser-test/bin/wait_for_server_start_and_run_tests.sh
+++ b/browser-test/bin/wait_for_server_start_and_run_tests.sh
@@ -25,6 +25,8 @@ for arg; do
   set -- "$@" "$arg"
 done
 
+yarn install
+
 if (($debug == 1)); then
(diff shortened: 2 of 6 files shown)


bin/run-browser-tests
```diff
@@ -7,9 +7,7 @@ docker::set_project_name_browser_tests
 bin/pull-image
 
 docker run --rm -it \
-  -v "$(pwd)/browser-test/src:/usr/src/civiform-browser-tests/src" \
-  -v "$(pwd)/browser-test/bin:/usr/src/civiform-browser-tests/bin" \
-  -v "$(pwd)/browser-test/tmp:/usr/src/civiform-browser-tests/tmp" \
+  -v "$(pwd)/browser-test:/usr/src/civiform-browser-tests" \
```
clouser2, line 10: This all seems reasonable to me, though I'm still learning Docker as I go. @kmcnellis might be able to provide a more thorough review on correctness here than I.

browser-test/bin/wait_for_server_start_and_run_tests.sh
```diff
@@ -25,6 +25,8 @@ for arg; do
   set -- "$@" "$arg"
 done
 
+yarn install
```
clouser2, line 28: Mind adding a comment as to why this is needed?

Can it go inside the $debug `if` below?

clouser2 (approved):

nbeloglazov:
Oops, I enabled auto-merge and it got merged automatically.

browser-test/bin/wait_for_server_start_and_run_tests.sh
nbeloglazov: I moved it from "build image" phase to "run test" phase. The rationale is that having it during "build image" doesn't really optimize for anything AFAIK. And having it in "run test" phase means that local changes to dependencies will be taken into effect. Like if someone adds a new dependency for TS browser tests - by running `bin/run-browser-tests` it will fetch dependencies and use them. With the old configuration changes to dependencies require rebuilding image locally. 

Also it matches better `server` where you can change dependencies in `build.sbt` and it will take effect immediately when running `run-dev` without need to rebuild image.

nbeloglazov:
@kmcnellis please take a look as Brett suggested and I'll address comments in a follow up PR

PR #2960, opened 2022-07-22
Improve typing in browser tests

nbeloglazov:
### Description

Adds explicit types to variables which otherwise treated as `any`. This required to enable strict type checking. Plus it might also improve auto-suggestion when using IDEs. 

### Issue(s)

Related #2955


jeanniefu (approved):
Thanks for these fixes!

PR #2963, opened 2022-07-22
Add test coverage for ActiveAndDraftQuestions and extend publish draft test coverage

clouser2:
### Description

As part of #2788, I'll be modifying the logic for how we determine how many programs reference a question. A follow-on from this will hopefully simplify the logic for whether a question can be archived. Prior to this refactoring, I've added tests for the existing functionality in `ActiveAndDraftQuestions` and left TODOs for places where the resulting behavior actually is a bug in the intended design. These will be fixed in a follow-on PR.

I also augmented the correctness tests for publishing a draft version (which I'll also be touching as part of #2788.

While here, I also:
  * Added `Versions.addQuestion` to update the M2M from the Version side. This made tests much simpler to write and I eventually intend to modify VersionRepository to call this when publishing a version.
  * Marked VersionRepository as final since there's no need to allow subclassing currently.
  * Fixed a minor null-pointer exception exposed when writing tests for getActiveQuestionDefinition / getDraftQuestionDefinition.
 
## Release notes:

N/A.

### Checklist

### Issue(s)

#2788

### DIFF
--- a/server/test/repository/VersionRepositoryTest.java
+++ b/server/test/repository/VersionRepositoryTest.java
@@ -54,12 +54,12 @@ public class VersionRepositoryTest extends ResetPostgres {
             .withBlock("Screen 1")
             .withRequiredQuestion(secondQuestion)
             .build();
-    Question secondQuestionUpdated = resourceCreator.insertQuestion("second-question");
-    secondQuestionUpdated.addVersion(versionRepository.getDraftVersion()).save();
+    Question secondQuestionDraft = resourceCreator.insertQuestion("second-question");
+    secondQuestionDraft.addVersion(versionRepository.getDraftVersion()).save();
     Program secondProgramDraft =
         ProgramBuilder.newDraftProgram("bar")
             .withBlock("Screen 1")
-            .withRequiredQuestion(secondQuestionUpdated)
+            .withRequiredQuestion(secondQuestionDraft)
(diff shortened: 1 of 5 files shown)


kmcnellis (approved):

server/test/repository/VersionRepositoryTest.java
```diff
@@ -39,17 +39,57 @@ public void setupVersionRepository() {
...
+        ProgramBuilder.newActiveProgram("foo")
+            .withBlock("Screen 1")
+            .withRequiredQuestion(firstQuestion)
+            .build();
+    Program secondProgramActive =
+        ProgramBuilder.newActiveProgram("bar")
+            .withBlock("Screen 1")
+            .withRequiredQuestion(secondQuestion)
+            .build();
+    Question secondQuestionUpdated = resourceCreator.insertQuestion("second-question");
```
shanemc-goog, line 57: nit: Updated -> Draft for parity with the Draft program that immediately follow

shanemc-goog (changes requested):

shanemc-goog (approved):
Just the one nit, but LGTM

server/test/repository/VersionRepositoryTest.java
clouser2: Done!

PR #2965, opened 2022-07-22
Build dev images for arm

kmcnellis:
### Description

Use the base Alpine image (since eclipse-temurin doesn't build alpine for arm), and docker buildx to enable multiplatform builds for the dev & formatter docker images.
 
## Release notes:

n/a

### Checklist

### DIFF
--- a/bin/build-dev
+++ b/bin/build-dev
@@ -3,9 +3,12 @@
 # DOC: Build a new development docker image and optionally push to Docker Hub if PUSH_IMAGE is set to anything.
 
 source bin/lib.sh
+export DOCKER_BUILDKIT=1
+
+PLATFORM="${PLATFORM:-"linux/amd64"}"
 
 # Build the new development image.
-docker build \
+docker buildx build --platform="${PLATFORM}" \
   -t civiform-dev:latest \
   --cache-from civiform/civiform-dev:latest \
   --build-arg BUILDKIT_INLINE_CACHE=1 \
--- a/bin/build-dev
+++ b/bin/build-dev
@@ -5,7 +5,7 @@
 source bin/lib.sh
 export DOCKER_BUILDKIT=1
 
-PLATFORM="${PLATFORM:-"linux/amd64"}"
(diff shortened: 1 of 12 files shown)


bin/build-dev
```diff
@@ -3,9 +3,12 @@
 # DOC: Build a new development docker image and optionally push to Docker Hub if PUSH_IMAGE is set to anything.
 
 source bin/lib.sh
+export DOCKER_BUILDKIT=1
+
+PLATFORM="${PLATFORM:-"linux/amd64"}"
```
nbeloglazov, line 8: Can you add comment explaining what this line is doing? Is that bash string manipulation/substring?

prod.Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM eclipse-temurin:11.0.15_10-jdk-alpine AS stage1
+# syntax=docker/dockerfile:1
+FROM --platform=$BUILDPLATFORM eclipse-temurin:11.0.15_10-jdk-alpine AS stage1
```
nbeloglazov, line 2: Given that everywhere else you are building images from scratch, would it make sense to build here one too for consistency? Or there are certain benefits in using eclipse-temurin here?

bin/build-dev
kmcnellis: This just sets a default for the variable

prod.Dockerfile
kmcnellis: I'm not entirely sure, but given that this is the production image, we do want it to be as secure as possible.  I think it's reasonable to have a dependance on the package management system in our development images, given that those are just running locally, but there's a bit of an extra layer of confidence, security, and validation in using officially vetted docker images for our production instance.  The things they do for the official images are in this dockerfile for jdk: https://github.com/adoptium/containers/blob/main/11/jdk/alpine/Dockerfile.releases.full and this for jre: https://github.com/adoptium/containers/blob/main/11/jre/alpine/Dockerfile.releases.full 

They take several steps to be more secure, like checksums, and pinning specific versions of java

azizova-leyla (approved):

PR #2928, opened 2022-07-19
Add a flag to allow global admins to be program admins

azizova-leyla:
### Description

Please explain the changes you made here.

## Release notes:

New config value is introduced: ALLOW_GLOBAL_ADMINS_BE_PROGRAM_ADMINS. When set to true it allows Civiform Global Admin to also be a Program Admin. 

### Checklist

- [x] Created tests which fail without the change (if possible)
- [x] Extended the README / documentation, if necessary

### Issue(s)

Fixes #2916

PR #2967, opened 2022-07-25
Revert "Add a flag to allow global admins to be program admins (#2928)" 

azizova-leyla:
I've kept the test that was missing to validate expected behavior.

## Release notes:

Removing config value: ALLOW_GLOBAL_ADMINS_BE_PROGRAM_ADMINS. After consideration, this is not a use case that we need to support.

### DIFF
--- a/server/app/services/role/RoleService.java
+++ b/server/app/services/role/RoleService.java
@@ -57,7 +57,7 @@ public class RoleService {
     ProgramDefinition program = programService.getProgramDefinition(programId);
     // Filter out CiviForm admins from the list of emails - a CiviForm admin cannot be a program
     // admin.
-    ImmutableSet<String> sysAdminEmails =
+    ImmutableSet<String> globalAdminEmails =
         getGlobalAdmins().stream()
             .map(Account::getEmailAddress)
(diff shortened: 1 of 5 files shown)


kmcnellis:
Was this something that the Seattle deployment still wanted?  If it's behind a flag, it's still optional

azizova-leyla:
No, Seattle doesn't need this feature.

server/app/services/role/RoleService.java
```diff
@@ -62,18 +55,18 @@ public Optional<CiviFormError> makeProgramAdmins(
...
-    ImmutableSet<String> globalAdminEmails =
-        allowGlobalAdmins
-            ? ImmutableSet.of()
-            : getGlobalAdmins().stream()
-                .map(Account::getEmailAddress)
-                .filter(address -> !Strings.isNullOrEmpty(address))
-                .collect(toImmutableSet());
+    // Filter out CiviForm admins from the list of emails - a CiviForm admin cannot be a program
+    // admin.
+    ImmutableSet<String> sysAdminEmails =
```
clouser2, line 60: Can we keep the variable name as "globalAdminEmails"?

clouser2 (approved):

PR #2974, opened 2022-07-25
Add question creation / last modified timestamps and ensure that they aren't updated as part of publishing a version

clouser2:
### Description

Why? New UI mocks for the question UI call for showing the time a draft version of the question was created / edited.

Additionally, this is a useful proxy for determining if an action updated the underlying DB model.

As a test, I updated the logic for publishing a draft Version to ensure that the associated Questions aren't updated when updating the M2M for similar reasons to doing this for Program (#2524).  

## Release notes:

Adding internal modification tracking information not exposed in the UI. This adds the `create_time` and `last_modified_time` columns to the `questions` table.

### Checklist

### Issue(s)

#2788


shanemc-goog:
FYI: Everything should have release notes pretty much from discussing with Bion.

server/app/repository/VersionRepository.java
```diff
@@ -95,7 +95,7 @@ public void publishNewSynchronizedVersion() {
               activeQuestion ->
                   !draftQuestionNames.contains(activeQuestion.getQuestionDefinition().getName()))
           // For each active question not associated with the draft, associate it with the draft.
-          .forEach(activeQuestionNotInDraft -> activeQuestionNotInDraft.addVersion(draft).save());
+          .forEach(activeQuestionNotInDraft -> draft.addQuestion(activeQuestionNotInDraft));
```
shanemc-goog, line 98: Add a comment since the directionality is important.  Also do we not need to save?

server/conf/evolutions/default/41.sql
```diff
@@ -0,0 +1,9 @@
+# --- Add creation and modification timestamps to questions.
+
+# --- !Ups
+alter table questions add column create_time timestamp;
+alter table questions add column last_modified_time timestamp;
+
+# --- !Downs
+alter table questions drop column if exists create_time;
+alter table questions drop column if exists last_modified_time;
```
shanemc-goog, line 9: missing newline says github

shanemc-goog (changes requested):
Left the comments outside the review on accident.

clouser2:
> FYI: Everything should have release notes pretty much from discussing with Bion.

Added. Out of curiosity, why? In this case, I guess we could say something like "adding internal modification tracking information not exposed in the UI". For something where I were fixing whitespace or adding test coverage, it doesn't seem that there's anything useful to add.

server/app/repository/VersionRepository.java
clouser2: Whoops, forgot to add that. Replicated the comment when adding programs to the version. The M2M will be updated when saving the version. Fortunately, all of our test coverage confirms that the behavior is the same before and after :D.

server/conf/evolutions/default/41.sql
clouser2: Done.

shanemc-goog:
> Out of curiosity, why?

Per the release [design doc](https://docs.google.com/document/d/1ELRj03KzsXzp1ezzuD3c03eZsQiLOjERDUxp5zbFuMQ/edit?resourcekey=0-cmpNQbrx_PjLHFAZKpXVmA#) pretty much anything that isn't clean up / refactor should have them it seems.  We're erroring on over communicating rather than under.

In particular you're changing the DB schema, and at least contributing to a new feature, so that's worth pointing out

clouser2:
> > Out of curiosity, why?
> 
> Per the release [design doc](https://docs.google.com/document/d/1ELRj03KzsXzp1ezzuD3c03eZsQiLOjERDUxp5zbFuMQ/edit?resourcekey=0-cmpNQbrx_PjLHFAZKpXVmA#) pretty much anything that isn't clean up / refactor should have them it seems. We're erroring on over communicating rather than under.
> 
> In particular you're changing the DB schema, and at least contributing to a new feature, so that's worth pointing out

Acknowledged and updated.

server/app/repository/VersionRepository.java
shanemc-goog: Ah I see!  The .save is kinda hidden here so I didn't see it either.

shanemc-goog (changes requested):
LGTM. Just need the release notes updated.

shanemc-goog (approved):

shanemc-goog:
> Acknowledged and updated

thanks, and the "feature" part of that comment was wrong I was mashing together this and status tracking.

Issue #2388, opened 2022-04-28
Setup a11y testing framework


jeanniefu:
We should add automated testing for accessibility. Accessibility folks recommended some tools that we can look into: axe, lighthouse

jeanniefu:
Breakdown of tasks:
1) Add a11y testing for all question types
2) Add a11y testing for the applicant flow (e.g. each page)
3) Update docs on a11y testing and how to fix issues

PR #2978, opened 2022-07-26
Always set FieldWithLabel id

jeanniefu:
### Description
Always set FieldWithLabel id. If missing, generates a random one.

We used to always generate a random ID if one was not set, but this was accidentally removed due to some refactoring in #2281.

One known issue this causes is that labels are no longer linked to inputs, which is a problem for screen readers.

## Release notes:

Fix screenreader bug where labels and inputs are not linked with ID.

### Checklist

### Issue(s)

#2388

### DIFF
--- a/server/app/views/components/FieldWithLabel.java
+++ b/server/app/views/components/FieldWithLabel.java
@@ -317,7 +317,6 @@ public class FieldWithLabel {
 
   /** Public final tag getters * */
   public DivTag getTextareaTag() throws RuntimeException {
-    genRandIdIfEmpty();
     if (isTagTypeTextarea()) {
       TextareaTag textareaFieldTag = TagCreator.textarea();
       applyAttributesFromSet(textareaFieldTag);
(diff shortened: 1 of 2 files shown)


server/app/views/components/FieldWithLabel.java
```diff
@@ -317,7 +317,6 @@ private DivTag getNonNumberInputTag() throws RuntimeException {
 
   /** Public final tag getters * */
   public DivTag getTextareaTag() throws RuntimeException {
-    genRandIdIfEmpty();
```
clouser2, line 320: Why remove this call?

clouser2 (approved):

clouser2:
Thanks for fixing!

server/app/views/components/FieldWithLabel.java
jeanniefu: It'll be called instead in the shared applyAttrsAndGenLabel(...) function below. That shared function (or checkboxApplyAttrsAndGenLabel) will be called by all the public getXTag specific functions.

clouser2: Aha, I'd totally missed that one. Also, the addition of so many tests gave me a lot of confidence that it worked, so thanks for that as well :).

PR #2983, opened 2022-07-26
Status Tracking: Display status options on Admin's Application review page.

### LINKED ISSUE #2911: Status Tracking: Update Application Details page to have a status selector
Show the available statuses (when present) to the Program Admin

shanemc-goog:
### Description

When statuses are configured for an applications program they are shown in a dropdown when the application is viewed by an Admin.

This only displays them, setting will come next.

Note: creating statuses is launch feature flag guarded so this won't be surfaced to users.
image (1)

## Release notes:

N/A - Part of a feature flagged upcoming feature

### Checklist

### Issue(s)

Fixes #2911


browser-test/src/admin_application_statuses.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  const programWithoutStatusesName = 'test program without statuses'
+
+  beforeAll(async () => {
+    const {page} = await startSession()
+    pageObject = page
+  })
+
+  describe('without program statuses', () => {
+    beforeAll(async () => {
+      // Timeout for clicks and element fills. If your selector fails to locate
```
clouser2, line 25: Did you find that this call was necessary for the test? We only set it for one other test (presumably for a specific peculiarity of that test, though it doesn't appear to be well-documented).

If it's not needed right now, I'd prefer leaving it out until we encounter an issue. If we do leave it, I'd suggest documenting the specific selector / codepath that makes it necessary.

browser-test/src/support/admin_programs.ts
```diff
@@ -453,6 +453,10 @@ export class AdminPrograms {
     ).not.toBeNull()
   }
 
+  async expectStatusSelectorVisible() {
+    await this.page.isVisible('#status-selector')
```
clouser2, line 457: Since browser tests are trying to assert things as the user sees them, I think it'd be ideal to avoid doing selections by ID wherever possible.

Instead, could we assert that some text isn't visible on the page?

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -294,6 +293,7 @@ public Result show(Http.Request request, long programId, long applicationId)
             .join();
     ImmutableList<Block> blocks = roApplicantService.getAllActiveBlocks();
     ImmutableList<AnswerData> answers = roApplicantService.getSummaryData();
+    var statusDefinitions = program.statusDefinitions();
```
clouser2, line 296: Nit: This variable is used from a single place. Consider just inlining the call for readability. As it stands, I don't think there's readability benefit since the variable name is the same as the method being called. I'd make the same argument for programId and programName. That said, I can also see an argument for doing it here since everything else is initialized before the call. Up to you.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -151,4 +164,33 @@ private DivTag renderAnswer(long programId, AnswerData answerData) {
                 .withClasses(
                     Styles.FLEX_AUTO, Styles.TEXT_RIGHT, Styles.FONT_LIGHT, Styles.TEXT_XS));
   }
+
+  private static SelectTag renderStatusOptionsDropdown(StatusDefinitions statusDefinitions) {
+    SelectTag dropdownTag =
+        select()
+            .withId("select-status")
```
clouser2, line 171: If we update the test assertions to operate based on a user-visible element, then I think we don't need to set an ID here.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -151,4 +164,33 @@ private DivTag renderAnswer(long programId, AnswerData answerData) {
                 .withClasses(
                     Styles.FLEX_AUTO, Styles.TEXT_RIGHT, Styles.FONT_LIGHT, Styles.TEXT_XS));
   }
+
+  private static SelectTag renderStatusOptionsDropdown(StatusDefinitions statusDefinitions) {
+    SelectTag dropdownTag =
```
clouser2, line 169: From a usability standpoint, we may want to add a label to the dropdown so that the user knows that this corresponds to a status (or add help text or some other affordance). I'd check with Raha on this. If we went the label route, a benefit would be that we could update the selector in the browser tests to select by that instead.

clouser2 (changes requested):
Nothing major. The only high-level question is around whether we should add a label to the dropdown, which may aid in usability and simplify browser testing.

server/app/controllers/admin/AdminApplicationController.java
shanemc-goog: inlined this one, Kept the others as I should have used programName in another spot.  programId is a method param, so tricky to "use the right one" if we want to pull from the definition.

server/app/views/admin/programs/ProgramApplicationView.java
shanemc-goog: ack, see the other comment about testing the no-statuses case.

server/app/views/admin/programs/ProgramApplicationView.java
shanemc-goog: Do you mean using SelectWithLabel ?

What's the main benefit of label here for screen readers?

clouser2: Not necessarily SelectWithLabel, since that tends to display pretty large and might not be suitable from a styling standpoint here.

My main concern is that someone might not know what the dropdown actually represents. Having a label indicating that it's for an application status may aid in understanding. It likely also will be a bit preferable from an a11y standpoint as well.

browser-test/src/support/admin_programs.ts
```diff
@@ -472,6 +472,12 @@ export class AdminPrograms {
     ).not.toBeNull()
   }
 
+  async expectStatusSelectorVisible() {
+    await this.page.isVisible(
```
clouser2, line 476: Can you add an explicit return here?

clouser2 (approved):

Issue #2979, opened 2022-07-26
Make sure there's limit on file uploads


azizova-leyla:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Additional context**

**Done when**
What conditions need to be met before this issue can be closed?

clouser2:
I'll add more information shortly. Setting a content-policy should allow us to enforce this limit: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html

Still working on understanding whether this is something set server-side or on the bucket itself.

clouser2:
It looks like this is something we'll set server-side when generating the presigned URL. I can't find any documentation around how to put this on the bucket itself.

Concrete proposal here from discussion would be to:
  * Add a new configurable setting for maximum file upload size (backed by an environment variable).
  * When generating the presigned URL, set a content limit restriction on the object (https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html)
  * Consider adding an attribute the HTML file input limiting the size of the upload (so someone is notified quickly on failure)
  * Add help text indicating the size limit
  * Validate what the S3 behavior is if a file larger than the configured limit is provided. Does it call back the server URL and indicate failure, fail silently, etc? If it does callback, add a server-side log for this.

clouser2:
Looks like we do have a limit on the file size already configured. It's presently set to 1GB (https://github.com/civiform/civiform/blob/7c060e3664f39a6668be0dac550d62dd043994e4/server/app/services/cloud/aws/SignedS3UploadRequest.java#L244).

azizova-leyla:
From Seattle's prod:
![image](https://user-images.githubusercontent.com/1741747/181125606-8d38d22e-bc66-46d2-babb-9f57ee4ce655.png)

azizova-leyla:
I'll drop default limit down to 100mb which should cover all existing files with plenty of room.
We should make file limit explicit in UI before dropping it lower.

LSistla1:
P2

gwendolyngoetz:
Updated usage 2023-09-25

![image](https://github.com/civiform/civiform/assets/195162/188a65bd-d396-43e6-8e8c-bec3ccdccaeb)

caitlinshk:
[...]

PR #2986, opened 2022-07-26
add file limit to config

azizova-leyla:
### Description

Setting file upload limit to 100MB and making it configurable. This PR is not updating UI to notify user of the limit, that will come later.

## Release notes:

Adding a flag to control file upload limit in S3. Setting default to 100MB. 

### Checklist

### Issue(s)

Quick fix for #2979


server/app/services/cloud/aws/SimpleStorage.java
```diff
@@ -50,8 +52,8 @@ public SimpleStorage(
       ApplicationLifecycle appLifecycle) {
     this.region = checkNotNull(region).get();
     this.credentials = checkNotNull(credentials);
-    this.bucket = checkNotNull(config).getString(AWS_S3_BUCKET_CONF_PATH);
-
+    this.config = checkNotNull(config);
```
clouser2, line 55: Not blocking: I'd either store the file limit as it's own var and initialize it here or remove "bucket" as a field and read it's value from the config in a similar way.

clouser2 (approved):

server/app/services/cloud/aws/SimpleStorage.java
azizova-leyla: done.

PR #2989, opened 2022-07-27
Add axe accessibility tests to browser tests for question types

jeanniefu:
### Description

Add axe (https://github.com/dequelabs/axe-core) accessibility browser tests. This also adds a custom expect matcher, that outputs the accessibility violations on failure. Here is an example of how the test failure message looks:

axe failure example

This PR just adds it for all question types. I will send a separate PR to cover the overall applicant's experience (e.g. a11y testing per page). I will also update the docs with debugging instructions.

## Release notes:

Add Axe accessibility testing framework to browser tests.

### Checklist

### Issue(s)
#2388

### DIFF
--- a/browser-test/src/end_to_end_enumerators.test.ts
+++ b/browser-test/src/end_to_end_enumerators.test.ts
@@ -134,9 +134,13 @@ describe('End to end enumerator test', () => {
     await applicantQuestions.answerNameQuestion('Porky', 'Pig')
     await applicantQuestions.clickNext()
 
+    // Validate that enumerators are accessible
+    await applicantQuestions.validateAccessibility()
+
     // Put in two things in the enumerator question
     await applicantQuestions.addEnumeratorAnswer('Bugs')
     await applicantQuestions.addEnumeratorAnswer('Daffy')
+
     await applicantQuestions.clickNext()
 
     // FIRST REPEATED ENTITY
--- a/browser-test/src/end_to_end_enumerators.test.ts
+++ b/browser-test/src/end_to_end_enumerators.test.ts
@@ -1,25 +1,35 @@
+import {Page} from 'playwright'
 import {
   startSession,
   loginAsAdmin,
   loginAsGuest,
+  logout,
   AdminQuestions,
   AdminPrograms,
   ApplicantQuestions,
   selectApplicantLanguage,
-  endSession,
+  resetSession,
   validateAccessibility,
   waitForPageJsLoad,
 } from './support'
 
 describe('End to end enumerator test', () => {
   const programName = 'ete enumerator program'
+  let pageObject: Page
 
-  it('Create nested enumerator and repeated questions', async () => {
-    const {browser, page} = await startSession()
+  beforeAll(async () => {
+    const {page} = await startSession()
+    pageObject = page
+  })
 
-    await loginAsAdmin(page)
-    const adminQuestions = new AdminQuestions(page)
-    const adminPrograms = new AdminPrograms(page)
+  afterEach(async () => {
+    await resetSession(pageObject)
+  })
+
+  it('Create nested enumerator and repeated questions as admin', async () => {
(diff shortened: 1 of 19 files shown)


browser-test/src/support/setup-jest.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Custom matcher that outputs accessibility violations using axe.
+expect.extend({
+  toHaveNoA11yViolations(results: axe.AxeResults) {
+    const numViolations = results.violations.length
+    const pass = numViolations == 0
+    if (pass) {
+      return {
+        message: () => `Expected axe accessibility violations, found none`,
```
clouser2, line 22: Not blocking: Nit: Since there's no string interpolation, regular quote characters can be used. The backtick signifies that it's a template string. The effect at runtime will be the same, but when I saw a backtick, I was looking for the interpolated variables by reflex.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -134,9 +134,13 @@ describe('End to end enumerator test', () => {
     await applicantQuestions.answerNameQuestion('Porky', 'Pig')
     await applicantQuestions.clickNext()
 
+    // Validate that enumerators are accessible
+    await applicantQuestions.validateAccessibility()
```
clouser2: I'd leave a comment here as to why we don't validate accessibility as a separate test like in the other files.

clouser2 (approved):
Nice! I like the factoring and custom matcher :D

browser-test/src/support/setup-jest.ts
```diff
@@ -0,0 +1,37 @@
...
+declare global {
+  namespace jest {
+    interface Expect extends CustomMatchers {}
+    interface Matchers<R> extends CustomMatchers<R> {}
+    interface InverseAsymmetricMatchers extends CustomMatchers {}
+  }
+}
+
+// Custom matcher that outputs accessibility violations using axe.
+expect.extend({
```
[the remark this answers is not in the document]
jeanniefu: Good idea, added!

browser-test/src/support/setup-jest.ts
```diff
@@ -0,0 +1,37 @@
...
+  toHaveNoA11yViolations(results: axe.AxeResults) {
+    const numViolations = results.violations.length
+    const pass = numViolations == 0
```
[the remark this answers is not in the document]
jeanniefu: Done!

browser-test/src/support/setup-jest.ts
jeanniefu: Thanks for the context! Switched to single quote.

browser-test/src/end_to_end_enumerators.test.ts
jeanniefu: Ah no good reason to not have it as a separate test :). Refactored this to match the others a bit more.

jeanniefu (commented):
Thanks so much for the review! I refactored the enumerator test file, and also moved `validateAccessibility()` into index.ts since it isn't specific to applicant questions.

browser-test/src/file.test.ts
```diff
@@ -92,6 +93,16 @@ describe('file upload applicant flow', () => {
       const error = await pageObject.$('.cf-fileupload-error')
       expect(await error?.isHidden()).toEqual(false)
     })
+
+    // TODO(#2988) Enable test once a11y issues are fixed.
+    // it('has no accessiblity violations', async () => {
```
gwendolyngoetz, line 98: Non-blocking: Rather than commenting this out, you might try the [it.skip](https://jestjs.io/docs/api#testskipname-fn) method. It'll show that it was skipped in the test output and maybe make it easier to remember to come back and enabled later.

jeanniefu: Didn't know this existed. This is perfect, thanks!!

PR #2993, opened 2022-07-28
Move back to bellsoft image for arm

kmcnellis:
### Description

OpenJDK doesn't successfully compile for arm (it randomly times out).  Move back to the bellsoft image re: https://github.com/adoptium/containers/issues/158#issuecomment-1046890267
## Release notes:

n/a

### Checklist

### Issue(s)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,8 @@
 # workaround uses an aarch64 (arm64) image instead when an optional platform argument is set to arm64.
 # Docker's BuildKit skips unused stages so the image for the platform that isn't used will not be built.
 
-ARG JDK_VERSION="11.0.15"
-FROM eclipse-temurin:${JDK_VERSION}-jdk-alpine as amd64
-FROM bellsoft/liberica-openjdk-alpine:${JDK_VERSION} as arm64
+FROM eclipse-temurin:11.0.16_8-jdk-alpine as amd64
+FROM bellsoft/liberica-openjdk-alpine:11.0.16-8 as arm64
 
 FROM ${TARGETARCH}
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,5 +1,13 @@
 # syntax=docker/dockerfile:1
-FROM alpine
+
+# The eclipse-temurin image and the standard openJDK11 fails to run on M1 Macs because it is incompatible with ARM architecture. This
+# workaround uses an aarch64 (arm64) image instead when an optional platform argument is set to arm64.
+# Docker's BuildKit skips unused stages so the image for the platform that isn't used will not be built.
+
+FROM eclipse-temurin:11.0.15_10-jdk-alpine as amd64
+FROM bellsoft/liberica-openjdk-alpine:11.0.16-8 as arm64
```
gwendolyngoetz, line 8: I wonder if we should keep temurin and bellsoft set at the same version number. My thought is that for devs we've seen minor issues when people (ok, often me) have been on different versions of docker-compose. Having ARM Macs vs Intel Macs/PCs with different version could give us troubleshooting headaches.

azizova-leyla (approved):
It works! Thank you for digging into this. Please address Gwen's comment, but lgtm otherwise.

Dockerfile
kmcnellis: We could pin them to the same version, but we'll have to make sure that the version is published to both places before updating it

kmcnellis: Correction: no.  They don't publish exactly the same versions

gwendolyngoetz: OK. Thanks

PR #2994, opened 2022-07-28
Remove OIDC TF variables that were moved to secret manager.

nbeloglazov:
### Description

The following variables are now created in AWS Secret manager and are not passed from terraform config: `APPLICANT_OIDC_CLIENT_ID`, `APPLICANT_OIDC_CLIENT_SECRET`, `APPLICANT_OIDC_DISCOVERY_URI`.

Also removes `aws_apprunner_vpc_connector` resource that no longer used since deployment was moved to fargate.

### DIFF
--- a/cloud/aws/templates/aws_oidc/variable_definitions.json
+++ b/cloud/aws/templates/aws_oidc/variable_definitions.json
@@ -31,24 +31,6 @@
     "tfvar": true,
     "type": "string"
   },
-  "APPLICANT_OIDC_CLIENT_ID": {
-    "required": true,
-    "secret": true,
-    "tfvar": true,
-    "type": "string"
-  },
-  "APPLICANT_OIDC_CLIENT_SECRET": {
-    "required": true,
-    "secret": true,
-    "tfvar": true,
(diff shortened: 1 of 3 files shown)


cloud/aws/templates/aws_oidc/variable_definitions.json
```diff
@@ -31,24 +31,6 @@
...
-    "tfvar": true,
-    "type": "string"
-  },
-  "APPLICANT_OIDC_CLIENT_SECRET": {
-    "required": true,
-    "secret": true,
-    "tfvar": true,
-    "type": "string"
-  },
-  "APPLICANT_OIDC_DISCOVERY_URI": {
```
azizova-leyla, line 46: discovery url shouldn't be a secret. we should remove it from secrets and keep it here.

nbeloglazov: Done. I'll remove it from secrets and switch to this in follow up PR to keep this one no-op.

azizova-leyla (approved):

PR #2998, opened 2022-07-28
Optimize the docker builds to reuse as much cached content as possible

kmcnellis:
### Description

Change the order of copying files, and add to the docker ignore, so that we don't have to re-download sbt or node unless they change.

Also move the docker volume creation for build directories into the image, to avoid overwriting them.

## Release notes:

n/a

### Checklist


Dockerfile
```diff
@@ -25,16 +25,26 @@ RUN set -o pipefail && \
...
+RUN mkdir -p "${PROJECT_LOC}"
+WORKDIR "${PROJECT_LOC}"
+
+RUN sbt update
+
+COPY "${PROJECT_NAME}"/package* .
+RUN npm install
+
+# Don't copy the project source until here, so everything above gets cached.
 COPY "${PROJECT_NAME}" "${PROJECT_LOC}"
```
nbeloglazov, line 37: Tbh I don't understand it. Wouldn't copying to the PROJECT_LOC overwrite everything that that folder had until that call?

Dockerfile
```diff
@@ -25,16 +25,26 @@ RUN set -o pipefail && \
...
-  sbt compile pushRemoteCache -Dconfig.file=conf/application.dev.conf
+RUN sbt update compile pushRemoteCache -Dconfig.file=conf/application.dev.conf
 
+COPY entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
 
+# Save build results to anonymous volumes for reuse
+VOLUME [ "/usr/src/server/target","/usr/src/server/project/project", "/usr/src/server/project/target", "/usr/src/server/node_modules", "/usr/src/server/.bsp/","/usr/src/server/public/stylesheets/" ]
+# The server code is it's own volume
+VOLUME ["/usr/src/server"]
```
nbeloglazov, line 48: Why is it important that server code is in separate volume? Also when running this image we are going to always (?) map `/usr/src/server` to local directory?

nbeloglazov (changes requested):

Dockerfile
kmcnellis: We map it locally, but not in the CI-mode tests.  Those rely on the built-in server directory

kmcnellis: It's not particularly important to make the server volume here, but it is important to do it after the build volumes

Dockerfile
kmcnellis: Yes, but that's not a problem.  We are using the npm cache when downloading the files, and entrypoint re-runs npm install. And we just need to run sbt so that it downloads all it's own dependancies.

Dockerfile
```diff
@@ -25,16 +25,26 @@ RUN set -o pipefail && \
   wget -qO - "${SBT_URL}" | tar xz -C "${INSTALL_DIR}" && \
   echo -ne "- with sbt ${SBT_VERSION}\n" >> /root/.built
 
+RUN mkdir -p "${PROJECT_LOC}"
+WORKDIR "${PROJECT_LOC}"
+
+RUN sbt update
+
+COPY "${PROJECT_NAME}"/package* .
```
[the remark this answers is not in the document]
kmcnellis: done.

nbeloglazov (approved):

PR #3003, opened 2022-07-29
Make applicant_oidc_discovery_uri a regular variable instead of secret

nbeloglazov:
### Description

`applicant_oidc_discovery_uri` is passed from civiform_config.sh and shouldn't be stored as secret. 

Additionally make BASE_URL required param over CUSTOM_HOSTNAME as the latter doesn't seem to provide useful functionality.


azizova-leyla (approved):
What about admin discovery uri? it should also be a variable, not a secret.

nbeloglazov:
Only `applicant_oidc_discovery_uri` in the sample config https://github.com/civiform/civiform-deploy/blob/main/civiform_config.example.sh So I decided to migrate only that variable as wasn't sure about ADFS. Asking Kiana about ADFS variables. WIll migrate it as well once it's in the config.

PR #3010, opened 2022-07-29
Update localization handling when updating program statuses

clouser2:
### Description

This bundles a few changes related to program status localization based on discussions. Namely:
  * Removing StatusDefinitions.emailBodyText (non-localized version) in lieu of the English-localized version in localizedEmailBodyText. This field is unnecessary since we'll always be consulting the Locale.US key in localizedEmailBodyText, similar to what we already do for a program's display name.
  * When updating statuses from the "edit status" controller, always update the English content while preserving any existing translations.
  * When setting the email body to blank from the "edit status" controller, treat this as disabling email and clear ALL translated content for the status.

Note: For "old" data that has the "emailBodyText" field present on a serialized program status, Jackson will throw an error for an unrecognized field. We *could* annotate the class to allow unknown fields when deserializing, but that seems like overkill since we don't actually have any production databases with a status set (only dev environments). As such, once this is merged, I'll send out an announcement on Slack that anyone using program statuses will need to reset their Postgres database to avoid this.

## Release notes:

Various fixes for program status translations. The feature is not yet launched, so these should have no effect. 

### Checklist

### Issue(s)

#2752


clouser2:
> The first bullet doesn't state an outcome just a problem, I think it's meant to say we're not using it anymore and instead defaulting to the system default language?
> 
> ```
> StatusDefinitions.emailBodyText (non-localized version). This field is unnecessary since we'll always be consulting the Locale.US key in localizedEmailBodyText, similar to what we already do for a program's display name.
> ```

Updated.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -128,32 +128,41 @@ private ErrorAnd<ProgramDefinition, CiviFormError> createOrEditStatusFromFormDat
...
+            // Only carry forward translated content if a non-empty email body
+            // has been provided. Any other translations will be lost when
+            // this occurs.
+            if (existingStatus.localizedEmailBodyText().isPresent()) {
+              // Preserve any existing translations.
+              LocalizedStrings updated =
+                  existingStatus
+                      .localizedEmailBodyText()
+                      .get()
+                      .updateTranslation(LocalizedStrings.DEFAULT_LOCALE, formData.getEmailBody());
```
shanemc-goog, line 160: I think we should make a updateDefaultTranslation() method so callers don't have to manage this concern correctly.

I'm a little iffy on how much we want to "enable" changing just the default, but we are in effect supporting it so I think making that parallel withDefaultValue() makes sense and just clearly doc it.

server/test/controllers/admin/AdminProgramStatusesControllerTest.java
```diff
@@ -51,13 +52,28 @@ public class AdminProgramStatusesControllerTest extends ResetPostgres {
       StatusDefinitions.Status.builder()
           .setStatusText("Rejected")
           .setLocalizedStatusText(LocalizedStrings.withDefaultValue("Rejected"))
-          .setEmailBodyText(Optional.of("Rejected email body"))
           .setLocalizedEmailBodyText(
               Optional.of(LocalizedStrings.withDefaultValue("Rejected email body")))
           .build();
 
+  private static final StatusDefinitions.Status WITH_TRANSLATIONS_STATUS =
```
shanemc-goog, line 59: nit/opt: WITH_STATUS_TRANSLATIONS reads better to me.

shanemc-goog (changes requested):
LGTM just some small stuff

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -128,32 +128,41 @@ private ErrorAnd<ProgramDefinition, CiviFormError> createOrEditStatusFromFormDat
...
+            newStatusBuilder.setLocalizedEmailBodyText(
+                Optional.of(LocalizedStrings.withDefaultValue(formData.getEmailBody())));
       }
       return service.appendStatus(program.id(), newStatusBuilder.build());
     }
     return service.editStatus(
         program.id(),
         formData.getConfiguredStatusText(),
         (existingStatus) -> {
+          // TODO(#2752): Disable the English translations UI
```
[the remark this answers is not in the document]
clouser2: Gotcha. I was hoping the formatter would catch/fix that if we want to approach the 100-character limit. I generally try to wrap based on the code around it since I find it more readable.

I very much don't have a strong enough opinion here to push back though :).

server/app/controllers/admin/AdminProgramStatusesController.java
clouser2: Sounds good. Added this along with relevant unit tests.

server/test/controllers/admin/AdminProgramStatusesControllerTest.java
clouser2: Done.

shanemc-goog (approved):

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
shanemc-goog: Ah interesting pattern.  I believe the general style guides prefer fitting to 100 as a default for less vertical space.

PR #3028, opened 2022-08-02
Remove auto-approve from setup.py

nbeloglazov:
### Description

Currently setup.py auto-approves and applies any changes when mode is staging or prod. It's dangerous as user should review those changes before applying them. The auto-approve seems to have been added for automation (CI tests) which currently are not implemented. In future, when we implement the tests, we can add this option back guarded by some flag or environment variable used only in tests.

### DIFF
--- a/cloud/shared/bin/lib/setup.py
+++ b/cloud/shared/bin/lib/setup.py
@@ -17,106 +17,107 @@ Setup.py sets up and runs the initial terraform deployment. It's broken into
 The script generates a .tfvars file that is used to deploy via terraform.
 """
 
-###############################################################################
-# Load and Validate Inputs
-###############################################################################
 
-## Load the Config and Definitions
-config_loader = ConfigLoader()
+def main():
+    ###############################################################################
+    # Load and Validate Inputs
+    ###############################################################################
 
-validation_errors = config_loader.load_config()
-if validation_errors:
-    new_line = '\n\t'
-    exit(
-        f"Found the following validation errors: {new_line}{f'{new_line}'.join(validation_errors)}"
-    )
+    ## Load the Config and Definitions
+    config_loader = ConfigLoader()
 
-###############################################################################
-# Load Setup Class for the specific template directory
-###############################################################################
-
-terraform_template_dir = config_loader.get_template_dir()
-app_prefix = config_loader.app_prefix
-Setup = load_setup_class(terraform_template_dir)
-
-template_setup = Setup(config_loader)
-template_setup.setup_log_file()
-current_user = template_setup.get_current_user()
-
-image_tag = config_loader.get_config_var("IMAGE_TAG")
-log_args = f"\"{image_tag}\" {current_user}"
-
-print("Writing TF Vars file")
-terraform_tfvars_path = os.path.join(
-    terraform_template_dir, config_loader.tfvars_filename)
-
-# Write the passthrough vars to a temporary file
-tf_var_writter = TfVarWriter(terraform_tfvars_path)
-conf_variables = config_loader.get_terraform_variables()
-tf_var_writter.write_variables(conf_variables)
-
-try:
-    print("Starting pre-terraform setup")
-    template_setup.pre_terraform_setup()
+    validation_errors = config_loader.load_config()
+    if validation_errors:
+        new_line = '\n\t'
+        exit(
+            f"Found the following validation errors: {new_line}{f'{new_line}'.join(validation_errors)}"
+        )
 
     ###############################################################################
-    # Terraform Init/Plan/Apply
+    # Load Setup Class for the specific template directory
     ###############################################################################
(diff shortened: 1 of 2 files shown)


cloud/shared/bin/lib/setup.py
```diff
@@ -17,106 +17,107 @@
...
-# Write the passthrough vars to a temporary file
-tf_var_writter = TfVarWriter(terraform_tfvars_path)
-conf_variables = config_loader.get_terraform_variables()
-tf_var_writter.write_variables(conf_variables)
-
-try:
-    print("Starting pre-terraform setup")
-    template_setup.pre_terraform_setup()
 
+def main():
```
azizova-leyla, line 21: I don't know enough of python, so maybe this is not an issue, but would this always run? my understanding of why this wasn't in the main is this can be called from various other scripts, and it it's inside a function it doesn't always gets executed? @clouser2 for opinions since he wrote lots of python.

cloud/shared/bin/lib/setup.py
```diff
@@ -17,106 +17,107 @@
...
+    Setup = load_setup_class(terraform_template_dir)
+
+    template_setup = Setup(config_loader)
+    template_setup.setup_log_file()
+    current_user = template_setup.get_current_user()
+
+    image_tag = config_loader.get_config_var("IMAGE_TAG")
+    log_args = f"\"{image_tag}\" {current_user}"
+
+    terraform_tfvars_path = os.path.join(
```
clouser2, line 50: Nit: The `print("Writing TF Vars file")` statement appears to have been removed. Was that intentional?

cloud/shared/bin/lib/setup.py
```diff
@@ -17,106 +17,107 @@
...
+
+    # Write the passthrough vars to a temporary file
+    tf_var_writter = TfVarWriter(terraform_tfvars_path)
+    conf_variables = config_loader.get_terraform_variables()
+    tf_var_writter.write_variables(conf_variables)
+
+    try:
+        print("Starting pre-terraform setup")
+        template_setup.pre_terraform_setup()
+        # return
```
clouser2, line 61: Nit: This seems unnecessary. Please remove.

cloud/shared/bin/lib/setup.py
```diff
@@ -17,106 +17,107 @@
...
+        ###############################################################################
```
clouser2, line 62: Nit: Mind indenting the comment stanza? Not sure what happened here.

Same for the below comment stanzas.

cloud/shared/bin/lib/setup.py
```diff
@@ -17,106 +17,107 @@
 The script generates a .tfvars file that is used to deploy via terraform.
 """
 
-###############################################################################
```
clouser2, line 20: High level comment for next time, nothing to do here.

If possible, can we do things like this as independent PRs? The diff for moving to a `main()`-based approach is the majority of this PR. Finding the place where auto-approve was removed required looking at a fair number of removed lines on the left and ensuring that it wasn't on the right side of the diff.

clouser2 (approved):

cloud/shared/bin/lib/setup.py
clouser2: TL;DR - The `main()` function will still be run as part of invoking the script. This change is consistent with the Google Python style guide and also makes things more testable.

@nbeloglazov outlined the benefits here, but adding a bit more context as well since Python's import system is interesting. I'm definitely a fan of the new approach since we can test things better and it improves control flow (as mentioned).  

Python's importability is fun. As part of importing a package, all top-level code in the file is run as part of the import. Top-level code is anything not explicitly in a lambda (e.g. a function definition, class definition, etc).

Prior to this PR, we would have run all of the code every time the module was imported, which is somewhat suboptimal.

Of course, a Python file sometimes represents a script to be run (as is the case here). When the file is being imported as part of a top-level invocation (e.g. `python my_script.py`), the `__name__` variable is set to `__main__`. This allows the script to execute some code if it's being executed as part of an invocation.

A convention that's followed even for shell scripts is the one from this PR where a most of the logic is encapsulated in a `main()` function. Note at the bottom that the following stanza causes the code to be executed (as before) when run from the command line:

```
if __name__ == '__main__':
  main()
```

This is consistent with the Google Python style guide (https://google.github.io/styleguide/pyguide.html#317-main).

azizova-leyla: Thank you for background. super helpful!

azizova-leyla (approved):

cloud/shared/bin/lib/setup.py
nbeloglazov: Yes, I should have done it like that in the first place. Moved "main-ifying" to separate PR: https://github.com/civiform/civiform/pull/3041

PR #3037, opened 2022-08-03
Prevent running destroy on setup state.

nbeloglazov:
### Description

Currently, when running `destroy`, terraform almost successfully destroys the main environment. But it fails when destroying resources that were created during setup (e.g. s3 bucket that holds terraform state). That bucket is marked as non-destructible via lifecycle. So better not to run destroy on setup state at all.


cloud/aws/templates/aws_oidc/bin/destroy.py
```diff
@@ -17,4 +17,5 @@ def pre_terraform_destroy(self):
             self._make_backend_override()
 
     def post_terraform_destroy(self):
-        self._tf_run_for_aws(is_destroy=True)
```
azizova-leyla, line 20: if you remove this - wouldn't it just do nothing? Maybe remove the command altogether if it doesn't make sense here?

nbeloglazov: Yes, it would do nothing. This method is shared with azure and it does something there. So I can't just remove the method as it's being called in the shared setup.py.
As for what should we do with the state object during destroy - I don't know really. I'm ok just leaving it there and, if needed, manually delete from s3.

azizova-leyla: I see. Can you print something to that regard?

azizova-leyla (approved):

cloud/aws/templates/aws_oidc/bin/destroy.py
nbeloglazov: Added message saying that bucket hasn't been destroyed and link to the bucket for manual removal.

PR #3039, opened 2022-08-03
Surface the number of programs referencing a question

clouser2:
### Description

This adds a count of active and draft programs referencing a given question. This will hopefully:
1) Make it clearer to the user when it's possible to archive a given question
2) Make it easier to see where a question is used

In order to achieve this:
  * A "preview" mode was introduced for publishing a version. This allows us to simulate the final set of programs in questions that would be in the ACTIVE version if "Publish all drafts" was used. Using the existing DRAFT version for this information isn't feasible since it only contains the newly edited programs/questions and won't contain programs/questions that are just carried forward.
  * Updated ActiveAndDraftQuestions to accept a VersionRepository rather than two distinct versions. This was mostly done for readability and making it easier to encapsulate the logic to calculate referenced programs from `ActiveAndDraftQuestions`.
  * `ActiveAndDraftQuestions` now calculates a set of programs referencing a given question using the "preview" of what changes are in the draft version. 

Another minor cleanup was to mark the `Version` class as final in order to prevent subclassing. This required adjusting a unit test that was unnecessarily subclassing using an anonymous class.

Future work (left TODOs for this) will be to:
  * Allow viewing a published program's configuration. Currently there's no way to allow this
  * Update the modal to contain links to the specific program block that references a question
  * Update the calculation of whether a question is deletable to make use of program references rather than calculating it using bespoke logic. Consulting the "preview" of the draft version avoids replicating simulation of what would be in the draft version if it were to be published

References surfaced on the question list page:
Screen Shot 2022-08-03 at 8 44 55 AM

Modal showing the specific programs referencing a question:
Screen Shot 2022-08-03 at 8 45 03 AM

## Release notes:

Surfacing the number of programs referencing a given question from the question list view.

### Checklist

### Issue(s)

#2788


clouser2 (commented):
Note: While the diff is ~800 lines, ~500 of them are test-related.

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -24,39 +33,95 @@ public final class ActiveAndDraftQuestions {
           String, Pair<Optional<QuestionDefinition>, Optional<QuestionDefinition>>>
       versionedByName;
   private final ImmutableMap<String, DeletionStatus> deletionStatusByName;
+  private final Optional<ImmutableMap<String, ImmutableSet<ProgramDefinition>>>
+      referencingDraftProgramsByName;
+  private final ImmutableMap<String, ImmutableSet<ProgramDefinition>>
+      referencingActiveProgramsByName;
 
-  public ActiveAndDraftQuestions(Version active, Version draft) {
-    ImmutableMap.Builder<String, QuestionDefinition> activeToName = ImmutableMap.builder();
```
clouser2, line 29: Note to the reviewer: A good portion of the changes here were actually in using stream classes for better readability rather than having a bunch of ImmutableMap.Builders that were appended in a for-loop and built later on.

rahaghas (approved):
lgtm

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -24,39 +33,95 @@ public final class ActiveAndDraftQuestions {
...
-        versionedByNameBuilder = ImmutableMap.builder();
-    for (String name : Sets.union(activeNames.keySet(), draftNames.keySet())) {
-      versionedByNameBuilder.put(
-          name,
-          Pair.create(
-              Optional.ofNullable(activeNames.get(name)),
-              Optional.ofNullable(draftNames.get(name))));
+  public ActiveAndDraftQuestions(VersionRepository repository) {
+    Version active = repository.getActiveVersion();
+    Version draft = repository.getDraftVersion();
```
bion, line 43: suggestion (blocking): c'tors generally should just initialize resources, but shouldn't themselves do much logic or any i/o. The previous version of this code wasn't great, but rather than moving it toward doing performing queries in the c'tor I recommend moving the query and collation logic into a private method that gets called in the public methods if it hasn't run yet i.e. the optional values it sets are empty.

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -24,39 +33,95 @@ public final class ActiveAndDraftQuestions {
...
+              .map(ProgramQuestionDefinition::id)
+              .filter(questionIdToNameLookup::containsKey)
+              .map(questionIdToNameLookup::get)
+              .collect(ImmutableList.toImmutableList());
+      for (String questionName : programQuestionNames) {
+        if (!result.containsKey(questionName)) {
+          result.put(questionName, Sets.newHashSet());
+        }
+        result.get(questionName).add(program.getProgramDefinition());
       }
```
bion, line 120: suggestion (non-blocking): instead of having the stream and then the loop could do `.forEach(questionName -> {...})` instead of `.collect` with the `for` loop

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -94,4 +159,39 @@ public Optional<QuestionDefinition> getDraftQuestionDefinition(String name) {
...
+
+  /** Contains sets of programs in the active and draft versions that reference a given question. */
+  @AutoValue
+  public abstract static class ReferencingPrograms {
+
+    /**
+     * Returns a set of references to the question in the DRAFT version. This returns
+     * Optional.empty() if there are no edited programs or questions in the DRAFT version.
+     */
+    public abstract Optional<ImmutableSet<ProgramDefinition>> draftReferences();
```
bion, line 179: suggestion (non-blocking): thoughts on using an empty set rather than an empty optional if there are no edited programs or questions in the DRAFT version? seems an empty optional and an optional containing an empty set are semantically identical atm

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -290,13 +289,9 @@ private Modal makeStatusDeleteModal(
...
-    return Modal.builder(randomModalId(), content).setModalTitle("Delete this status").build();
-  }
-
-  private static String randomModalId() {
-    // We prepend a "a-" since element IDs must start with an alphabetic character, whereas UUIDs
-    // can start with a numeric character.
-    return "a-" + UUID.randomUUID().toString();
+    return Modal.builder(Modal.randomModalId(), content)
+        .setModalTitle("Delete this status")
+        .build();
```
bion, line 294: question (non-blocking): this looks unrelated to the other changes?

server/app/views/components/Modal.java
```diff
@@ -76,6 +77,12 @@ private DivTag getModalHeader() {
                 .withClasses(ReferenceClasses.MODAL_CLOSE, BaseStyles.MODAL_CLOSE_BUTTON));
   }
 
+  public static String randomModalId() {
+    // We prepend a "a-" since element IDs must start with an alphabetic character, whereas UUIDs
+    // can start with a numeric character.
+    return "a-" + UUID.randomUUID().toString();
```
bion, line 83: suggestion (non-blocking): might reduce a couple milliseconds of thinking for someone reading the rendered result by prepending `uuid-` instead

bion (changes requested):

clouser2 (commented):
All feedback addressed. The main point we might want to iterate is the discussion of whether to do collation in the constructor or not. I've tried to outline some tradeoffs and tried to implement a halfway solution. Please let me know what you think.

server/app/repository/VersionRepository.java
```diff
@@ -53,12 +53,24 @@ public VersionRepository(ProgramRepository programRepository) {
...
+   * Simulates publishing a new version of all programs and questions. All DRAFT programs/questions
+   * will become ACTIVE, and all ACTIVE programs/questions without a draft will be copied to the
+   * next version. This method will not mutate the database and will return an updated Version
+   * corresponding to what would be the new ACTIVE version.
+   */
+  public Version previewPublishNewSynchronizedVersion() {
+    return publishNewSynchronizedVersion(true);
+  }
+
+  private Version publishNewSynchronizedVersion(boolean dryRun) {
```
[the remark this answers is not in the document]
bion: thanks!

server/app/services/question/ActiveAndDraftQuestions.java
bion: Yeah, just brought it up as an option. If you this reads better to you then sgtm

server/app/services/question/ActiveAndDraftQuestions.java
bion: Ah I see, that's reasonable

server/app/services/question/ActiveAndDraftQuestions.java
bion: Doing the IO in the static method is definitely an improvement, and way less of a lift in comparison to rewriting the whole thing to be lazy.

**I'm ok with merging as-is.**

That being said, to address the multithreading issue you brought up, I think that could be resolved by making the private method that performs the I/O and collates the results [`synchronized`](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html).

Something like:

```java
  private synchronized void ensureResults() {
    // load data from DB, collate it and set the variables
  }
  
  public boolean draftVersionHasAnyEdits() {
    ensureResults();
    return draftVersionHasAnyEdits;
  }
```

This wouldn't address your concern of reduced immutability though, though the mutability would be internal only.

bion (approved):

server/app/services/question/ActiveAndDraftQuestions.java
clouser2: Sounds good. I'm going to merge this as-is and try and put together a follow-on PR that moves all initialization of this data to an actual function that initializes a plain-old-data-object.

PR #3041, opened 2022-08-03
Move top-level code to main() functions in setup.py and destroy.py

nbeloglazov:
### Description

This helps with local development, for example to abort execution early by adding "return" and generally having a main function is more conventional compared to having all code on top-level.

### DIFF
--- a/cloud/shared/bin/lib/destroy.py
+++ b/cloud/shared/bin/lib/destroy.py
@@ -7,28 +7,34 @@ from setup_class_loader import load_destroy_class
 Destroy.py destroys the setup
 """
 
-###############################################################################
-# Load and Validate Inputs
-###############################################################################
 
-## Load the Config and Definitions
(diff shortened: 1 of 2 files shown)


cloud/shared/bin/lib/destroy.py
```diff
@@ -7,28 +7,34 @@
...
-
-template_dir = config_loader.get_template_dir()
-Destroy = load_destroy_class(template_dir)
-
-template_destroy = Destroy(config_loader)
-template_destroy.pre_terraform_destroy()
-terraform.perform_apply(config_loader, is_destroy=True)
-template_destroy.post_terraform_destroy()
+
+def main():
```
clouser2, line 11: Looks good modulo the comments from https://github.com/civiform/civiform/pull/3028#pullrequestreview-1060577560.

clouser2 (approved):

PR #3047, opened 2022-08-03
Fix expect Program state utilities to actually expect.

shanemc-goog:
### Description

Fix 2 expect utilities to actually call expect

## Release notes:

NA

### Checklist

### Issue(s)

### DIFF
--- a/browser-test/src/support/admin_programs.ts
+++ b/browser-test/src/support/admin_programs.ts
@@ -160,11 +160,11 @@ export class AdminPrograms {
 
   // TODO(clouser): More fine-grained selectors for this and active.
   async expectDraftProgram(programName: string) {
-    await this.page.isVisible(this.programCardSelector(programName, 'Draft'))
+    expect(await this.page.isVisible(this.programCardSelector(programName, 'Draft')))
   }
 
   async expectActiveProgram(programName: string) {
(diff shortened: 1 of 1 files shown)


browser-test/src/support/admin_programs.ts
```diff
@@ -160,11 +160,17 @@ export class AdminPrograms {
 
   // TODO(clouser): More fine-grained selectors for this and active.
```
clouser2, line 161: While here, can you remove the TODO against my name? I think I forgot to remove it prior to originally merging the PR.

browser-test/src/support/admin_programs.ts
```diff
@@ -160,11 +160,17 @@ export class AdminPrograms {
...
   async expectDraftProgram(programName: string) {
-    await this.page.isVisible(this.programCardSelector(programName, 'Draft'))
+    expect(
```
clouser2, line 163: Thanks for fixing! That was a complete miss when I added these. I assume that I thought `isVisible()` would wait UNTIL something was visible, not return the current visibility state :(.

clouser2 (approved):

browser-test/src/support/admin_programs.ts
shanemc-goog: done

PR #3051, opened 2022-08-04
Replace assertThrows with assertThatThrownBy

nbeloglazov:
### Description

We use AssertJ in java unit tests to perform assertions, including for asserting exceptions. There are few places where we use Junit's `assertThrows` though. This PR migrates those calls to use `assertThatThrownBy`. Improves consistency. Downside is that it's harder to remember compared to simple `assertThrows` :/


clouser2 (approved):
Nice cleanup :)

PR #3052, opened 2022-08-04
Add missing awaits to browser tests

nbeloglazov:
### Description

Found them using eslint. Once https://github.com/civiform/civiform/pull/2777 is in it will be a presubmit error.


jeanniefu (approved):
Thanks for fixing!

nbeloglazov:
Some browser test are failing. Seems to be flakiness as tests are different from the ones being updated in this PR. So merging the PR.

PR #3056, opened 2022-08-04
Using deep link does not create duplicate applications

### LINKED ISSUE #2969: Error Submitting Due to more than one Draft Application
**Describe the bug**

User reported they were unable to submit an application and got a generic error show in a banner.

From the logs:
```
r.ApplicationRepository - java.util.concurrent.CompletionException: java.lang.RuntimeException: Found more than one DRAFT application for applicant 18874, program 830.
java.util.concurrent.CompletionException: java.lang.RuntimeException: Found more than one DRAFT application for applicant 18874, program 830.
at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:645)
at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.RuntimeException: Found more than one DRAFT application for applicant 18874, program 830.
at repository.ApplicationRepository.submitApplicationInternal(ApplicationRepository.java:98)
at repository.ApplicationRepository.lambda$submitApplication$1(ApplicationRepository.java:78)
at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
... 2 more
```

I fixed this issue by setting the lifecycle_stage on the older application to 'obsolete', but that won't fix the source that is causing the issue. Unfortunately I don't have anymore info at this time on what may have contributed towards an application getting in this state.

**To Reproduce**

Currently unknown

**Expected behavior**
Applications submit successfully. Duplication applications can't get created.

- Browser: Chrome

**Additional context**

bion:
### Description

This fixes a bug wherein an applicant could have multiple drafts for the same program. This would happen when:

1. applicant answers a question for version A of a program, creating draft application A
1. CF admin publishes a new version of the program, version B
1. applicant navigates to the program using its deep link (e.g. from the index page), the deep link (previously) always links to the most recent version of the program, now version B
1. applicant answers a question for version B of the program, creating a second draft
1. applicant attempts to submit their application, encounter an error due to logic that asserts only one draft of a program at a time

The fix implemented here is to change the logic for deep links to redirect to the version of the program the applicant already has a draft application for, if one exists.

## Release notes:

Bugfix: duplicate draft applications when applicant uses deep links to updated program.

### Checklist

### Issue(s)

fixes #2969

### DIFF
--- a/server/app/controllers/api/ProgramApplicationsApiController.java
+++ b/server/app/controllers/api/ProgramApplicationsApiController.java
@@ -28,7 +28,7 @@ import services.program.ProgramNotFoundException;
 import services.program.ProgramService;
 
 /** API controller for admin access to a specific program's applications. */
-public class ProgramApplicationsApiController extends CiviFormApiController {
+public final class ProgramApplicationsApiController extends CiviFormApiController {
 
   public static final String PROGRAM_SLUG_PARAM_NAME = "programSlug";
   public static final String FROM_DATE_PARAM_NAME = "fromDate";
--- a/server/app/controllers/api/ProgramApplicationsApiController.java
+++ b/server/app/controllers/api/ProgramApplicationsApiController.java
@@ -91,7 +91,7 @@ public class ProgramApplicationsApiController extends CiviFormApiController {
             .orElse(new IdentifierBasedPaginationSpec<>(pageSize, Long.MAX_VALUE));
 
     return programService
-        .getProgramDefinitionAsync(programSlug)
+        .getActiveProgramDefinitionAsync(programSlug)
         .thenApplyAsync(
             programDefinition -> {
               PaginationResult<Application> paginationResult;
--- a/server/app/controllers/applicant/RedirectController.java
+++ b/server/app/controllers/applicant/RedirectController.java
@@ -62,7 +62,7 @@ public final class RedirectController extends CiviFormController {
     this.languageUtils = checkNotNull(languageUtils);
   }
 
-  public CompletionStage<Result> programByName(Http.Request request, String programSlug) {
+  public CompletionStage<Result> programBySlug(Http.Request request, String programSlug) {
     Optional<CiviFormProfile> profile = profileUtils.currentUserProfile(request);
 
     if (profile.isEmpty()) {
--- a/server/app/controllers/applicant/RedirectController.java
+++ b/server/app/controllers/applicant/RedirectController.java
@@ -15,7 +15,6 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.atomic.AtomicLong;
 import javax.inject.Inject;
 import models.Applicant;
 import models.LifecycleStage;
@@ -80,7 +79,7 @@ public final class RedirectController extends CiviFormController {
             (Applicant applicant) -> {
               // Attempt to set default language for the applicant.
               applicant = languageUtils.maybeSetDefaultLocale(applicant);
-              final AtomicLong applicantId = new AtomicLong(applicant.id);
+              final long applicantId = applicant.id;
 
               // If the applicant has not yet set their preferred language, redirect to
               // the information controller to ask for preferred language.
@@ -88,12 +87,12 @@ public final class RedirectController extends CiviFormController {
                 return CompletableFuture.completedFuture(
                     redirect(
                             controllers.applicant.routes.ApplicantInformationController.edit(
-                                applicantId.get()))
+                                applicantId))
                         .withSession(
                             request.session().adding(REDIRECT_TO_SESSION_KEY, request.uri())));
               }
 
-              return getProgramVersionForApplicant(applicantId.get(), programSlug)
+              return getProgramVersionForApplicant(applicantId, programSlug)
                   .thenComposeAsync(
                       (Optional<ProgramDefinition> programForExistingApplication) -> {
                         // Check to see if the applicant already has an application
@@ -104,11 +103,10 @@ public final class RedirectController extends CiviFormController {
                               redirect(
                                   controllers.applicant.routes.ApplicantProgramReviewController
                                       .preview(
-                                          applicantId.get(),
-                                          programForExistingApplication.get().id())));
+                                          applicantId, programForExistingApplication.get().id())));
                         }
 
-                        return redirectToActiveProgram(applicantId.get(), programSlug);
+                        return redirectToActiveProgram(applicantId, programSlug);
                       },
                       httpContext.current());
             },
--- a/server/app/controllers/applicant/RedirectController.java
+++ b/server/app/controllers/applicant/RedirectController.java
@@ -5,38 +5,42 @@ import static controllers.CallbackController.REDIRECT_TO_SESSION_KEY;
 
 import auth.CiviFormProfile;
 import auth.ProfileUtils;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import controllers.CiviFormController;
 import controllers.LanguageUtils;
 import controllers.routes;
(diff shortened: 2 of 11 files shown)


server/app/controllers/applicant/RedirectController.java
```diff
@@ -69,37 +72,71 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
...
+                            controllers.applicant.routes.ApplicantInformationController.edit(
+                                applicantId))
+                        .withSession(
+                            request.session().adding(REDIRECT_TO_SESSION_KEY, request.uri())));
               }
 
-              return redirect(
-                  controllers.applicant.routes.ApplicantProgramReviewController.preview(
-                      applicant.id, programFuture.join().id));
+              return getProgramVersionForApplicant(applicantId, programSlug)
```
azizova-leyla, line 95: it feels potentially error prone to do the check in RedirectController. if we ever have another entry point the check will need to be duplicated. Is it possible to abstract it more and use the same helper here and in the normal flow?

bion: > if we ever have another entry point

It's hard to think of a scenario where we would -- the deep link is intended to be the canonical front door to a program for the applicant.

> Is it possible to abstract it more and use the same helper here and in the normal flow?

The tricky part here is "in the normal flow". There's a number of different routes that include the explicit program version ID, so we'd need to guard each of them. Doable, but a much larger scope of effort I don't see how an applicant would even get to one of those routes without being redirected by this controller.

I'm happy to discuss this more. If we decide we want to put deeper guards against this happening in the future I suggest we do it as a follow up though so that this fix can go out sooner.

server/app/controllers/api/ProgramApplicationsApiController.java
```diff
@@ -91,7 +91,7 @@ public CompletionStage<Result> list(
             .orElse(new IdentifierBasedPaginationSpec<>(pageSize, Long.MAX_VALUE));
 
     return programService
```
clouser2, line 93: Not blocking: Semi-unrelated to this change, but part of a pet project of mine. Assuming it doesn't cause any compile errors, mind marking this class as final to make the choice for subclassing explicit rather than implicit?

server/app/controllers/applicant/RedirectController.java
```diff
@@ -46,20 +49,20 @@ public RedirectController(
...
     this.profileUtils = checkNotNull(profileUtils);
-    this.programRepository = checkNotNull(programRepository);
+    this.programService = checkNotNull(programService);
     this.upsellView = checkNotNull(upsellView);
     this.messagesApi = checkNotNull(messagesApi);
     this.languageUtils = checkNotNull(languageUtils);
   }
 
-  public CompletionStage<Result> programByName(Http.Request request, String programName) {
+  public CompletionStage<Result> programByName(Http.Request request, String programSlug) {
```
clouser2, line 65: Blocking: Can this be renamed to `programBySlug()`? Otherwise, I'm confused about the variable name change.

server/app/controllers/applicant/RedirectController.java
```diff
@@ -69,37 +72,71 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
...
-
-              Applicant applicant = applicantFuture.join();
+    return profile
+        .get()
+        .getApplicant()
+        .thenComposeAsync(
+            (Applicant applicant) -> {
               // Attempt to set default language for the applicant.
               applicant = languageUtils.maybeSetDefaultLocale(applicant);
+              final long applicantId = applicant.id;
```
clouser2, line 82: Not blocking: Any particular reason to stash this off here rather than just accessing it on the Applicant object at the callsite?

server/app/controllers/applicant/RedirectController.java
```diff
@@ -69,37 +72,71 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
...
+  private CompletionStage<Optional<ProgramDefinition>> getProgramVersionForApplicant(
+      long applicantId, String programSlug) {
+    return applicantService
+        .relevantPrograms(applicantId)
+        .thenApplyAsync(
+            (ImmutableMap<LifecycleStage, ImmutableList<ProgramDefinition>> relevantPrograms) ->
+                relevantPrograms.values().stream()
+                    .flatMap(Collection::stream)
+                    .filter(program -> program.slug().equals(programSlug))
+                    .findAny(),
```
clouser2, line 136: In the case where the applicant DOES have existing draft applications in multiple program versions, it seems that this call might return a "random" program version. In reality, we should likely make the version returned here the same as the one from @azizova-leyla  bug fix (e.g. the most recent program version). I also may be misunderstanding things here :).

server/app/controllers/applicant/RedirectController.java
```diff
@@ -69,37 +72,71 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
...
+                        .withSession(
+                            request.session().adding(REDIRECT_TO_SESSION_KEY, request.uri())));
               }
 
-              return redirect(
-                  controllers.applicant.routes.ApplicantProgramReviewController.preview(
-                      applicant.id, programFuture.join().id));
+              return getProgramVersionForApplicant(applicantId, programSlug)
+                  .thenComposeAsync(
+                      (Optional<ProgramDefinition> programForExistingApplication) -> {
```
azizova-leyla, line 97: I don't see a filter for draft applications anywhere. How does this work? Wouldn't ` relevantPrograms() `return all active programs as well and the else clause will never get executed?

clouser2 (changes requested):

server/app/controllers/applicant/RedirectController.java
bion: variables defined in an outer scope and referenced in lambdas must be final, applicant can't be final because we reassign it on the line above

server/app/controllers/applicant/RedirectController.java
bion: Super good point, thanks for the catch

server/app/controllers/applicant/RedirectController.java
bion: done

clouser2 (approved):

server/app/controllers/api/ProgramApplicationsApiController.java
bion: done

server/app/controllers/applicant/RedirectController.java
bion: fixed while pairing

server/app/controllers/applicant/RedirectController.java
```diff
@@ -69,37 +72,74 @@ public CompletionStage<Result> programByName(Http.Request request, String progra
...
+                    controllers.applicant.routes.ApplicantProgramReviewController.preview(
+                        applicantId, activeProgramDefinition.id())),
+            httpContext.current());
+  }
+
+  private CompletionStage<Optional<ProgramDefinition>> getProgramVersionForApplicant(
+      long applicantId, String programSlug) {
+    // Find all applicant's DRAFT applications for programs of the same slug
+    // redirect to the newest program version with a DRAFT application.
+    // TODO(#2573): clean this up, don't compare on program ID
```
azizova-leyla, line 131: how's #2573 relevant here? you're not comparing on program ids.

azizova-leyla (approved):

server/app/controllers/applicant/RedirectController.java
bion: We are on line 138.

Brett's PR for #2573 has query code that will make this less janky and allow us to compare on the actual applications

PR #3060, opened 2022-08-04
add prod db params

### LINKED ISSUE #3054: Make db prod config
Need to set a few more params to make rds in prod mode - replication, prevent deletion, availability zones.

azizova-leyla:
### Description

Add availability zones, deletion protection, final snapshot and encryption to database.

## Release notes:

### Checklist

### Issue(s)

Fixes #3054

### DIFF
--- a/cloud/aws/templates/aws_oidc/main.tf
+++ b/cloud/aws/templates/aws_oidc/main.tf
@@ -28,7 +28,6 @@ resource "aws_db_instance" "civiform" {
     Name = "${var.app_prefix} Civiform Database"
     Type = "Civiform Database"
   }
-  availability_zone               = data.aws_availability_zones.available.names
   deletion_protection             = local.deletion_protection
   instance_class                  = var.postgres_instance_class
   allocated_storage               = var.postgres_storage_gb
--- a/cloud/aws/templates/aws_oidc/main.tf
+++ b/cloud/aws/templates/aws_oidc/main.tf
@@ -41,7 +41,7 @@ resource "aws_db_instance" "civiform" {
   parameter_group_name            = aws_db_parameter_group.civiform.name
   publicly_accessible             = false
   skip_final_snapshot             = local.skip_final_snapshot
-  final_snapshot_identifier       = "{var.app_prefix}-civiform-db-finalsnapshot"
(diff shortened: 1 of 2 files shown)


cloud/aws/templates/aws_oidc/main.tf
```diff
@@ -22,7 +28,8 @@ resource "aws_db_instance" "civiform" {
     Name = "${var.app_prefix} Civiform Database"
     Type = "Civiform Database"
   }
-
+  availability_zone               = data.aws_availability_zones.available.names
+  deletion_protection             = local.deletion_protection
```
nbeloglazov, line 32: I'm guessing it means that it will be impossible to run `destroy` on prod environments, right? Should we fail `destroy` early if civiform_mode is prod? 
I think we can suggest either
1. Print something like "destroying of prod environment is not supported. Get a cup of coffee and prepare for 1h long manual deleting of stuff in aws console."
2. Go unmark database (and probably few other places) as deletion-protected and then run destroy script again.

I don't think we need to do anything at this point. Just thinking.

cloud/aws/templates/aws_oidc/main.tf
```diff
@@ -22,7 +28,8 @@ resource "aws_db_instance" "civiform" {
     Name = "${var.app_prefix} Civiform Database"
     Type = "Civiform Database"
   }
-
+  availability_zone               = data.aws_availability_zones.available.names
```
nbeloglazov, line 31: The name name suggests that this param takes a single zone. But `available.names` returns a list. Is that ok?

nbeloglazov (approved):

cloud/aws/templates/aws_oidc/main.tf
azizova-leyla: you're correct, it can only be one and its picked automatically, so we don't need to set it.

cloud/aws/templates/aws_oidc/main.tf
azizova-leyla: Correct. destroy is meant to be there for testing only. Prod should be HARD to destroy. I like your "get a cup of coffee" idea, I'll add it to destroy script.

nbeloglazov (approved):

PR #3062, opened 2022-08-05
Simplify setup.py and few other things

nbeloglazov:
### Description

Currently setup.py calls terraform using the same code as in terraform.py. This PR updates setup.py to use terraform.py instead. The only difference is that setup.py calls `terraform init` with `-migrate-state` option, while terraform.py doesn't use that option. Given that setup.py is expected to be called on empty repo there no need to migrate anything as state should be immediately created on s3. Here are docs for the [backend initialization](https://www.terraform.io/cli/commands/init#backend-initialization).

Additional changes:
1. Add main() to AWS' deploy.py.
2. Change post_terraform_setup() to throw error by default as it should be called only when post-terraform setup is enabled by the subclass.
3. Make "you are using incorrect image tag" message a bit more helpful.

### Checklist

### DIFF
--- a/cloud/shared/bin/deploy
+++ b/cloud/shared/bin/deploy
@@ -7,7 +7,7 @@ source cloud/shared/bin/lib.sh
 cloud/shared/bin/validate_cloud_provider
 
 if [[ "${IMAGE_TAG}" == "latest" ]]; then
-  out::error "--tag must reference a specific image, not 'latest'."
+  out::error "--tag must reference a specific image, not 'latest'. Check available versions on https://hub.docker.com/r/civiform/civiform"
   exit 1
 fi
 
--- a/cloud/aws/bin/deploy.py
+++ b/cloud/aws/bin/deploy.py
@@ -1,23 +1,27 @@
 #! /usr/bin/env python3
-import subprocess
-import os
 import sys
 
 from cloud.shared.bin.lib import terraform
 from cloud.shared.bin.lib import tf_apply_setup
 from cloud.aws.bin.lib import backend_setup
 
-config_loader = tf_apply_setup.load_config()
 
-if config_loader.is_dev():
-    terraform.copy_backend_override(config_loader)
-else:
-    backend_setup.setup_backend_config(config_loader)
+def main():
+    config_loader = tf_apply_setup.load_config()
 
-if not terraform.perform_apply(config_loader):
-    sys.stderr.write('Terraform deployment failed.')
-    # TODO(#2606): write and upload logs.
-    raise ValueError('Terraform deployment failed.')
+    if config_loader.is_dev():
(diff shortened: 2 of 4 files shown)


cloud/shared/bin/deploy
```diff
@@ -7,7 +7,7 @@ source cloud/shared/bin/lib.sh
 cloud/shared/bin/validate_cloud_provider
 
 if [[ "${IMAGE_TAG}" == "latest" ]]; then
```
clouser2, line 9: Not blocking: Technically there's no longer any reason that this check still needs to exist since we allow deploying from latest in https://github.com/civiform/civiform-deploy/pull/14.

That said, it's still a moving tag, so we might want to continue not supporting it so that there's not drift between the label when the Git repository was checked out and when the image is being referenced.

Since our civiform-deploy script has the ability to resolve a tag to a specific SHA version of the digest, we could modify this code to ONLY accept a tag of the form `sha256:<digest>`. That would give us the benefits of knowing that our Git repository is always in sync with what's being deployed. We'd also know that the image used doesn't ever have the ability to change (due to a tag update).

The main benefit of the above is more predictability about what's actually being run
If you agree with that general sentiment, I'm happy to file a bug.

clouser2 (approved):

azizova-leyla (approved):

cloud/shared/bin/deploy
nbeloglazov: Are you suggesting that deploy engineers use the sha hash instead of versions? Versions feel more natural.

clouser2: I'm suggesting that the "main" deployment repo only accept sha hash. Deployers would use the deploys cripts from `civiform-deploy`, which would accept versions, tags, or sha hashes.

The main thing I'm trying to get at it here is that tags / versions are immutable (the tag can move to different versions, etc), whereas a sha is immutable. We pipe this through to various downstream systems and using a floating tag for that seems like it could lead to some unintended updates. Contrast that with a sha hash, which is immutable.

nbeloglazov: You meant that "tags / versions are mutable", right? Not immutable? 
Mentally I treat tags and versions as immutable so they correspond to hashes for me. But if that's a concern then I don't mind. It's going to be a bit less annoying to not being use latest during development, but shouldn't be that cumbersome.

How will I easily find a sha of a release? Today I go to https://hub.docker.com/r/civiform/civiform/tags Should I use its digest instead? 

Hm, also it might make debugging prod more challenging as now you can't easily see what version it is running by looking at the "image" in task definition: https://us-west-2.console.aws.amazon.com/ecs/v2/task-definitions/mikita5-td/7/json?region=us-west-2 Though there it's "latest" in real prod it will be proper version.

PR #3090, opened 2022-08-05
Build deletion status off of program references and allow archiving questions in the draft version

clouser2:
### Description

This reuses the "program references" feature added in #3039. Allowing archive / restore for questions added in the draft version seemed like a feature that would match user expectations.

## Release notes:

Adds ability to archive questions created in an unpublished version.

### Checklist

### Issue(s)

#2788


server/app/models/Version.java
```diff
@@ -171,4 +179,29 @@ public boolean removeTombstoneForQuestion(Question question) {
     }
     return this.tombstonedQuestionNames.remove(question.getQuestionDefinition().getName());
   }
+
+  @VisibleForTesting
+  public boolean addTombstoneForProgramForTest(Program program) {
```
bion, line 184: suggestion (blocking): since this is public, it doesn't need the testing annotation

bion, line 184: suggestion (blocking): since this is public, it doesn't need the testing annotation -- either make it private or remove the annotation

bion, line 184: suggestion (blocking): please document this method's return type

server/app/models/Version.java
```diff
@@ -171,4 +179,29 @@ public boolean removeTombstoneForQuestion(Question question) {
...
+    if (this.tombstonedProgramNames == null) {
+      this.tombstonedProgramNames = new ArrayList<>();
+    }
+    if (this.programIsTombstoned(program.getProgramDefinition().adminName())) {
+      return false;
+    }
+    return this.tombstonedProgramNames.add(program.getProgramDefinition().adminName());
+  }
+
+  public boolean removeTombstoneForProgram(Program program) {
```
bion, line 194: suggestion (blocking): please document this method

server/app/models/Version.java
```diff
@@ -171,4 +179,29 @@ public boolean removeTombstoneForQuestion(Question question) {
...
+    if (this.tombstonedProgramNames == null) {
+      this.tombstonedProgramNames = new ArrayList<>();
+    }
+    return this.tombstonedProgramNames.remove(program.getProgramDefinition().adminName());
+  }
+
+  public boolean hasAnyChanges() {
```
bion, line 201: suggestion (blocking): please document this method

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -70,25 +69,35 @@ private ActiveAndDraftQuestions(Version active, Version draft, Version withDraft
                           Optional.ofNullable(draftNames.get(name)));
                     }));
 
-    draftVersionHasAnyEdits = draft.getPrograms().size() > 0 || draft.getQuestions().size() > 0;
+    draftVersionHasAnyEdits = draft.hasAnyChanges();
```
bion, line 72: good encapsulation!

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -70,25 +69,35 @@ private ActiveAndDraftQuestions(Version active, Version draft, Version withDraft
...
             .collect(
                 ImmutableMap.toImmutableMap(
                     questionName -> questionName,
                     questionName -> {
-                      if (draft.getTombstonedQuestionNames().contains(questionName)) {
-                        return DeletionStatus.PENDING_DELETION;
-                      } else if (isNotDeletable(
-                          active, draft, activeNames, draftNames, questionName)) {
+                      final ImmutableMap<String, ImmutableSet<ProgramDefinition>>
+                          referencesToExamine =
```
bion, line 89: suggestion (non-blocking): i don't think the compiler requires this to be `final` so i recommend removing `final`

bion (changes requested):

server/app/models/Version.java
clouser2: Sounds good. I kept the "ForTest" suffix in the method name. We can remove this if we ever find a need for this to be used outside of testing.

clouser2: Done.

clouser2: Done for this and oher existing undocumented methods. Attempted to keep it consistent with the language used in `program/questionIsTombstoned`.

server/app/models/Version.java
clouser2: Done.

server/app/models/Version.java
clouser2: Done.

server/app/services/question/ActiveAndDraftQuestions.java
clouser2: Thanks :).

server/app/services/question/ActiveAndDraftQuestions.java
clouser2: Removed. I added it as final more as a hint to myself that we would only initialize the variable once. We don't really do that elsewhere so probably not worth starting here :D.

bion (approved):

PR #3093, opened 2022-08-08
Ask user for secret values during initial AWS deployments

### LINKED ISSUE #2948: Allow users to enter secret values via CLI
We create secrets in Terraform, it would be convenient to then ask the user to input their secret values and use aws cli to upload the values.

nbeloglazov:
### Description

After the initial AWS deployment we create secrets with empty values in AWS secret manager. CiviForm app requires non-empty values and won't start with empty. This PR adds a code that asks user to provide values for the secrets and sets them in the secret manager. The code doesn't ask for values if they secret is already non-empty. 

### Checklist

### Issue(s)

Fixes #2948


cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -0,0 +1,34 @@
...
+from cloud.shared.bin.lib.config_loader import ConfigLoader
+
+
+class AwsCli:
+    """Wrapper class that encapsulates calls to AWS CLI."""
+
+    def __init__(self, config: ConfigLoader):
+        self.config: ConfigLoader = config
+
+    def get_secret_value(self, secret_name: str) -> str:
```
azizova-leyla, line 15: it doesn't sit well with me to get secret value here. 
Can we instead allow users to skip entering a value by entering empty string?

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -0,0 +1,34 @@
...
+        res = self._call_cli(
+            [
+                "secretsmanager", "get-secret-value",
```
azizova-leyla, line 18: nit: prefer single quotes.

cloud/aws/templates/aws_oidc/bin/setup.py
```diff
@@ -30,3 +51,31 @@ def pre_terraform_setup(self):
...
+
+    def requires_post_terraform_setup(self):
+        return True
+
+    def post_terraform_setup(self):
+        for name, doc in SECRETS.items():
+            self._maybe_set_secret_value(
+                f"{self.config.app_prefix}-{name}", doc)
+
+    def _maybe_set_secret_value(self, secret_name: str, documentation: str):
```
azizova-leyla, line 63: do we really need types? @clouser2 wdyt?

azizova-leyla (changes requested):

cloud/aws/templates/aws_oidc/bin/aws_cli.py
nbeloglazov: I think it makes script more bullet-proof to accidental changes. That way if value is set user doesn't accidentally overrides it as they need to go to the AWS console, look at it and then modify it. It is helpful in cases when setup is run multiple times. I know you believe that setup should be run only once, but I think we need to be prepared for cases where setup needs to be run multiple times. 

And your worry is that we read the secret and it can accidentally be exposed somewhere? I can update the code to limit exposure a bit more by having is_secret_empty() method so that secret value never leaves AwsCli class .

cloud/aws/templates/aws_oidc/bin/setup.py
nbeloglazov: Types are useful for my local development. It allows intellij provide better auto-suggestions. And it helps reading code. Not sure about enabling type-checking on presubmit though. Ideally - yes, but it might require a bunch of gymnastics to make sure all of code passes type-checking.

What is your concern with having types? Inconsistent code style and potentially out-of-date types?

cloud/aws/templates/aws_oidc/bin/aws_cli.py
nbeloglazov: Done.

cloud/aws/templates/aws_oidc/bin/aws_cli.py
azizova-leyla: yes, my worry is the secret can be exposed this way. Let's do is_secret_empty(). I understand that the person running this script has access to secrets manager and can look things up, but I'd rather not expose more than we need.

cloud/aws/templates/aws_oidc/bin/setup.py
```diff
@@ -1,27 +1,48 @@
...
-import shutil
+from getpass import getpass
+from typing import Dict
 
+from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
 from cloud.aws.templates.aws_oidc.bin.aws_template import AwsSetupTemplate
 from cloud.aws.bin.lib import backend_setup
+from cloud.shared.bin.lib.config_loader import ConfigLoader
+
+# TODO: move these to variable_definitions.json and read docs from there.
```
clouser2, line 11: I have a slight preference to mark the TODO against a specific issue (e.g. `TODO(#1234)`) so it's easier to remember what works still needs to be done for a specific issue.

cloud/aws/templates/aws_oidc/bin/setup.py
```diff
@@ -1,27 +1,48 @@
...
+}
 
 
 class Setup(AwsSetupTemplate):
 
-    def get_current_user(self):
-        get_current_command = "aws sts get-caller-identity --query UserId --output text"
+    def __init__(self, config: ConfigLoader):
+        super().__init__(config)
+        self.aws_cli = AwsCli(config)
```
clouser2, line 31: Not blocking: If this shouldn't be accessed outside of the class, I'd make it "private" (e.g. `self._aws_cli`).

cloud/aws/templates/aws_oidc/bin/setup.py
clouser2: I'm indifferent on type hints :). There's definite benefit and having them in a few isolated places likely will increase readability.

clouser2 (approved):

cloud/aws/templates/aws_oidc/bin/setup.py
nbeloglazov: Done.

cloud/aws/templates/aws_oidc/bin/setup.py
nbeloglazov: Done.

azizova-leyla (approved):

PR #3101, opened 2022-08-08
Add Dob to TI dashboard and TI Controller Updates

### LINKED ISSUE #2009: TIs can view additional client information within TI dashboard.

swatkat1:
### Description

Adding a column in TI dashboard for the applicant's DOB.
As part of updating the TIController, this PR moves the logic of Validation and Error-Handling to the TrustedIntermediaryService class and packages all form errors in the TIClientCreationResult object. This follows the similar implementation of APIKeyService class.  

## Release notes:
Adding a Date of Birth column to the TI Dashboard and Implementing Service class to TI route.

### Checklist

### Issue(s)

Fixes #2009 

### Screenshot:
image


bion:
@swatkat1 this PR has a lot more than adding a column, please update the title, description, and release notes

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  convertToMMDDYYYY(dobDate: string): string {
+    const [year, month, day] = dobDate.split('-')
+    return `${month}-${day}-${year}`
+  }
+}
+
+// This class helps to test the TrustesIntermediary dashboard changes
+// We should be logged as a TI to use this class
```
bion, line 43: issue (blocking): this comment is identical to the one on the class

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -49,13 +51,15 @@ public TrustedIntermediaryController(
...
-      Config config) {
+      Config config,
+      TrustedIntermediaryService tiService) {
     this.profileUtils = Preconditions.checkNotNull(profileUtils);
     this.tiDashboardView = Preconditions.checkNotNull(trustedIntermediaryDashboardView);
     this.userRepository = Preconditions.checkNotNull(userRepository);
     this.formFactory = Preconditions.checkNotNull(formFactory);
     this.messagesApi = Preconditions.checkNotNull(messagesApi);
     this.baseUrl = Preconditions.checkNotNull(config).getString("base_url");
+    this.tiService = tiService;
```
bion, line 62: suggestion (blocking): check not null

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -105,21 +109,19 @@ public Result addApplicant(Long id, Http.Request request) {
...
+      if (tiClientCreationResult.isSuccessful()) {
+        return redirect(
+            routes.TrustedIntermediaryController.dashboard(
+                /* search= */ Optional.empty(), /* page= */ Optional.empty()));
+      }
+    }
+    // Only for EmailAddressException, the expection is thrown from the Service class and handled in
+    // the controller.
+    // For all other errors, the error is packaged in the TIClientResult object
+    catch (EmailAddressExistsException e) {
```
bion, line 124:
```suggestion
    // Only for EmailAddressException, the expection is thrown from the Service class and handled in
    // the controller.
    // For all other errors, the error is packaged in the TIClientResult object
    } catch (EmailAddressExistsException e) {
```

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -129,17 +131,30 @@ public Result addApplicant(Long id, Http.Request request) {
...
+        tiClientCreationResult.getForm().get());
+  }
+
+  private String getFormErrors(Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
+    StringBuilder errorMessage = new StringBuilder();
+    form.errors().stream()
+        .forEach(
+            validationError ->
+                errorMessage.append(
+                    System.getProperty("line.separator") + validationError.message()));
```
bion, line 145: question (blocking): why use `System.getProperty("line.separator")` here? the server's operating system shouldn't have any impact on what we display in the browser

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -129,17 +131,30 @@ public Result addApplicant(Long id, Http.Request request) {
...
+    return errorMessage.toString();
   }
 
   private Result redirectToDashboardWithError(
       String errorMessage, Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
     return redirect(
-            routes.TrustedIntermediaryController.dashboard(Optional.empty(), Optional.empty()))
+            routes.TrustedIntermediaryController.dashboard(Optional.empty(), Optional.of(1)).url())
```
bion, line 152: suggestion (blocking): please add param comments for the optionals here

server/app/services/DateConverter.java
```diff
@@ -32,6 +32,16 @@ public String formatIso8601Date(Instant instant) {
     return DateTimeFormatter.ISO_DATE.withZone(zoneId).format(instant);
   }
 
+  /**
+   * Converts a DateString in "YYYY-MM-dd" format to LocalDate type Throws DateTimeParseException
+   */
+  public LocalDate parseStringtoLocalDate(String dateString) {
```
bion, line 38: nit (blocking): 
```suggestion
  public LocalDate parseStringToLocalDate(String dateString) {
```

server/app/services/ti/TIClientCreationResult.java
```diff
@@ -0,0 +1,34 @@
+package services.ti;
+
+import forms.AddApplicantToTrustedIntermediaryGroupForm;
+import java.util.Optional;
+import play.data.Form;
+
+public class TIClientCreationResult {
+
+  private final Optional<Form<AddApplicantToTrustedIntermediaryGroupForm>> form;
+  // private final Optional<String> errorMessage;
```
bion, line 10: suggestion (blocking): please remove commented code

server/app/services/ti/TIClientCreationResult.java
```diff
@@ -0,0 +1,34 @@
+package services.ti;
+
+import forms.AddApplicantToTrustedIntermediaryGroupForm;
+import java.util.Optional;
+import play.data.Form;
+
+public class TIClientCreationResult {
```
bion, line 7: suggestion (blocking): please document this class and make it `final`

bion: question (blocking): it looks like this class is just a wrapper around `Optional<Form<AddApplicantToTrustedIntermediaryGroupForm>>` without storing any other state or providing other functionality. Is it needed?

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,80 @@
...
+import forms.AddApplicantToTrustedIntermediaryGroupForm;
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import javax.inject.Inject;
+import models.TrustedIntermediaryGroup;
+import play.data.Form;
+import repository.UserRepository;
+import services.DateConverter;
+
+public class TrustedIntermediaryService {
```
bion, line 13: suggestion (blocking): please document this class and make it `final`

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,80 @@
...
+  private final UserRepository userRepository;
+  private final DateConverter dateConverter;
+
+  @Inject
+  public TrustedIntermediaryService(UserRepository userRepository, DateConverter dateConverter) {
+    this.userRepository = userRepository;
+    this.dateConverter = dateConverter;
```
bion, line 20: suggestion (blocking): check not null

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public TIClientCreationResult addNewClient(
+      Form<AddApplicantToTrustedIntermediaryGroupForm> form,
+      TrustedIntermediaryGroup trustedIntermediaryGroup)
+      throws EmailAddressExistsException {
```
bion, line 26: suggestion (blocking): why not handle `EmailAddressExistsException` here and add an error message to the form?

swatkat1:
@bion - changed the title, description and the release notes.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private Form<AddApplicantToTrustedIntermediaryGroupForm> validateDateOfBirth(
+      Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
+    if (Strings.isNullOrEmpty(form.value().get().getDob())) {
+      return form.withError("dob", "Date of Birth required");
+    }
+    LocalDate currentDob = null;
+    try {
+      currentDob = dateConverter.parseStringtoLocalDate(form.value().get().getDob());
```
bion, line 71: suggestion (blocking): rather than adding a new method, please use `parseIso8601DateToStartOfDateInstant` here

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,80 @@
...
+      throws EmailAddressExistsException {
+    form = validateEmailAddress(form);
+    form = validateFirstName(form);
+    form = validateLastName(form);
+    form = validateDateOfBirth(form);
+    if (form.hasErrors()) {
+      return TIClientCreationResult.failure(form);
+    }
+    userRepository.createNewApplicantForTrustedIntermediaryGroup(
+        form.get(), trustedIntermediaryGroup);
```
bion, line 35: thought (non-blocking): it's outside of the scope of this PR, but a good code improvement for the future would be to move the logic in `UserRepository` that's related to TIs into this new service you've created. That would fit the intention of keeping repositories focused on query logic only.

bion (changes requested):
This is a big step in the direction of better separation of concerns!

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: Good catch!!

browser-test/src/support/ti_dashboard.ts
swatkat1: Updated

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: When multiple errors are present in a form, all the errors were concatenated and shown to the end user. After discussing with @clouser2  - we decided we would add a line break.

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: Sure

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: Done!

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: But it will return as an instant and not as a LocalDate

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: The old code showed a redirection tot he sign-in page on EmailExistsException. I though through about it and decided, if an email account exists, we should let the TI know of the conflicts and may be follow the old code. 
     return redirectToDashboardWithError(
          "Email address already in use.  Cannot create applicant if an account already exists. "
              + " Direct applicant to sign in and go to"
              + " "
              + trustedIntermediaryUrl,
          form);

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: I agree with you! we should move a bit of code around and address it in the subsequent PRs.

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: Done!

server/app/services/ti/TIClientCreationResult.java
swatkat1: Yes in future, We may add the form for updating Client DOB.. it is needed...

server/app/controllers/ti/TrustedIntermediaryController.java
bion: Ah, you can just use `"\n"` instead

server/app/services/ti/TIClientCreationResult.java
swatkat1: Done!

server/app/services/ti/TrustedIntermediaryService.java
bion: If the email already exists, how does redirecting to the sign-in page help the TI?

server/app/services/ti/TIClientCreationResult.java
bion: So the plan is for it to wrap two separate forms in the future?

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: Done!

server/app/services/DateConverter.java
swatkat1: good catch!

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: hmm.. may be... So should I just add this validation in the form and show to the as an form field error?

server/app/services/ti/TrustedIntermediaryService.java
bion: If you need a `LocalDate` please factor out the parsing logic from the existing method (`LocalDate.parse(dateString, DateTimeFormatter.ISO_DATE)`) so we don't have multiple implementations of parsing the same date format

server/app/services/ti/TrustedIntermediaryService.java
bion: I think that makes the most sense

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: If I understand it correctly - call parseStringToLocalDate() inside the parseIso8601DateToStartOfDateInstant method?

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: It is done! I have also added a unit test for validating this logic.

server/app/services/DateConverter.java
```diff
@@ -32,14 +32,24 @@ public String formatIso8601Date(Instant instant) {
     return DateTimeFormatter.ISO_DATE.withZone(zoneId).format(instant);
   }
 
+  /**
+   * Converts a DateString in "YYYY-MM-dd" format to LocalDate type Throws DateTimeParseException
+   */
+  public LocalDate parseStringToLocalDate(String dateString) {
+    return LocalDate.parse(dateString, DateTimeFormatter.ofPattern("yyyy-MM-dd"));
```
bion, line 39: suggestion (blocking): please use the built-in format:
```suggestion
    return LocalDate.parse(dateString, DateTimeFormatter.ISO_DATE);
```

server/app/services/DateConverter.java
```diff
@@ -32,14 +32,24 @@ public String formatIso8601Date(Instant instant) {
     return DateTimeFormatter.ISO_DATE.withZone(zoneId).format(instant);
   }
 
+  /**
+   * Converts a DateString in "YYYY-MM-dd" format to LocalDate type Throws DateTimeParseException
+   */
+  public LocalDate parseStringToLocalDate(String dateString) {
```
bion, line 38: suggestion (blocking): please match the naming convention of the other method
```suggestion
  public LocalDate parseIso8601DateToLocalDate(String dateString) {
```

server/app/services/ti/TrustedIntermediaryService.java
bion: Yep! Please see comments in the DateConverter.java file

server/app/services/ti/TrustedIntermediaryService.java
bion: Yep! Please see comments in the DateConverter.java file

server/app/services/ti/TrustedIntermediaryService.java
bion: Great!

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: Done! thank you

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,94 @@
...
+
+    return TIClientCreationResult.success();
+  }
+
+  private Form<AddApplicantToTrustedIntermediaryGroupForm> validateEmailAddress(
+      Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
+    if (Strings.isNullOrEmpty(form.value().get().getEmailAddress())) {
+      return form.withError("emailAddress", "Email Address required");
```
clouser2, line 63: For future maintainability, would you mind exposing the field names a `public static final String` objects on the form class that is being parsed?

server/test/services/ti/TrustedIntermediaryServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import play.data.Form;
+import play.data.FormFactory;
+import play.mvc.Http;
+import repository.UserRepository;
+
+public class TrustedIntermediaryServiceTest extends WithMockedProfiles {
+
+  UserRepository repo;
```
clouser2, line 24: Mind marking all of these as `private`?

server/app/services/ti/TIClientCreationResult.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Holds state relevant to the result of attempting to create a Client for TrustedIntermediary.
+ *
+ * <p>If the creation attempt was successful, the response is a successful object with an empty
+ * form.
+ *
+ * <p>If the creation attempt was not successful, the result contains the same form object with the
+ * form validation errors.
+ */
+public final class TIClientCreationResult {
```
bion, line 16: issue (blocking): it's still not clear to me why this class is necessary vs returning the `Optional<FormClass>` from the service

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import models.TrustedIntermediaryGroup;
+import org.junit.Before;
+import org.junit.Test;
+import play.mvc.Http;
+import play.mvc.Result;
+import play.test.Helpers;
+import repository.UserRepository;
+
+public class TrustedIntermediaryControllerTest extends WithMockedProfiles {
+  UserRepository repo;
```
clouser2, line 22: Please mark these all as `private`.

server/app/services/DateConverter.java
swatkat1: Done! thank you. I also updated the documentation.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -0,0 +1,100 @@
...
+  private Form<AddApplicantToTrustedIntermediaryGroupForm> validateDateOfBirth(
+      Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
+    if (Strings.isNullOrEmpty(form.value().get().getDob())) {
+      return form.withError("dob", "Date of Birth required");
+    }
+    LocalDate currentDob = null;
+    try {
+      currentDob = dateConverter.parseStringToLocalDate(form.value().get().getDob());
+    } catch (DateTimeParseException e) {
+      return form.withError("dob", "Date of Birth must be in MM-dd-yyyy format");
```
bion, line 93: issue (blocking): is this validation message correct? i'm seeing the date in a different format everywhere else in this code

clouser2 (changes requested):
Minor issues for the most part and a hope for a bit stricter validation in the browser tests. Other than those, this looks great!

server/app/services/ti/TIClientCreationResult.java
swatkat1: Yes..

server/app/services/DateConverter.java
```diff
@@ -32,14 +32,24 @@ public String formatIso8601Date(Instant instant) {
...
+    return LocalDate.parse(dateString, DateTimeFormatter.ISO_DATE);
+  }
+  /** Returns the current LocalDate based on the specified time-zone(zoneId) */
```
bion, line 44:
```suggestion
  }
  
  /** Returns the current LocalDate based on the specified time-zone(zoneId) */
```

server/app/services/DateConverter.java
```diff
@@ -32,14 +32,24 @@ public String formatIso8601Date(Instant instant) {
...
+  public LocalDate getCurrentDateForZoneId() {
+    return LocalDate.now(this.zoneId);
+  }
```
bion, line 47:
```suggestion
  }
  
```

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -226,7 +227,24 @@ private TrTag renderApplicantRow(Account applicant) {
...
+    Optional<Applicant> newestApplicant = applicantAccount.newestApplicant();
+    if (newestApplicant.isEmpty()) {
+      return td().withClasses(BaseStyles.TABLE_CELL_STYLES, Styles.PR_12);
+    }
+    String currentDob =
+        newestApplicant
+            .get()
+            .getApplicantData()
+            .getDateOfBirth()
+            .map(LocalDate::toString)
```
bion, line 244: suggestion (blocking): would prefer a method in DateConverter that specifies the format rather than implicitly relying on the default format of `LocalDate::toString`

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -63,4 +63,14 @@ public void withFailedUpdates() {
     assertThat(data.getFailedUpdates()).isEqualTo(ImmutableMap.of(samplePath, "invalid_value"));
     assertThrows(IllegalStateException.class, () -> data.asJsonString());
   }
+
+  @Test
+  public void checkDateOfBirth() {
+    ApplicantData data = new ApplicantData();
+    String sampleDOB = "2022-01-05";
```
bion, line 70: suggestion (blocking): please see the style guide on case of abbreviations: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

bion (changes requested):

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -22,7 +24,24 @@ describe('Trusted intermediaries', () => {
...
+    const client: ClientInformation = {
+      emailAddress: 'fake@sample.com',
+      firstName: 'first',
+      middleName: 'middle',
+      lastName: 'last',
+      dobDate: '2021-05-10',
+    }
+    await tiDashboard.createClient(client)
+    await tiDashboard.expectDashboardContainClient(client)
+  })
```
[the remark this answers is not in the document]
swatkat1: I tried it and with the current code, you are right, we can't.

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: So the date is entered as "MM-dd-yyyy" format through the date picker. But our J2HTML date picker converts it from mm-dd-yyyy format to yyyy-mm-dd format. check how we enter date:
image

server/app/services/ti/TIClientCreationResult.java
swatkat1: This is needed in the very next PR i am going to work on. I will be adding a form for Updating DOB and this class will also hold the form object which contains errors

browser-test/src/trusted_intermediary.test.ts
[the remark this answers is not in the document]
swatkat1: It can't be done. But i have manually switched between the two branches (main and my branch) and checked compatibility.

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,52 @@
...
+    await this.page.fill('label:has-text("Email Address")', client.emailAddress)
+    await this.page.fill('label:has-text("First Name")', client.firstName)
+    await this.page.fill('label:has-text("Middle Name")', client.middleName)
+    await this.page.fill('label:has-text("Last Name")', client.lastName)
+    await this.page.fill('label:has-text("Date Of Birth")', client.dobDate)
+    await this.page.click('text="Add"')
+  }
+
+  async expectDashboardContainClient(client: ClientInformation) {
+    expect(
```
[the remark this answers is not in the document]
swatkat1: Can you please share the command to do this?

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -63,4 +63,14 @@ public void withFailedUpdates() {
     assertThat(data.getFailedUpdates()).isEqualTo(ImmutableMap.of(samplePath, "invalid_value"));
     assertThrows(IllegalStateException.class, () -> data.asJsonString());
   }
+
+  @Test
```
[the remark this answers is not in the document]
swatkat1: Can't be done as the client cannot be created with a dob.

browser-test/src/support/ti_dashboard.ts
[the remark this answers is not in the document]
swatkat1: Mind sharing the command.

server/app/services/ti/TIClientCreationResult.java
bion: In that case the service class will presumably have a new method updateClient which can return an optional of the other form, no?

server/test/services/applicant/ApplicantDataTest.java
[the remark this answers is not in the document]
bion: There will be clients in the database that don't have a DOB when this code is released. You can add a test that simulates that by creating the client accounts directly, a good place to put the code for that is `support.ResourceCreator`

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: done!

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: The whole point of toString() is just this. I don't see a need for another method.

server/test/services/applicant/ApplicantDataTest.java
swatkat1: I will update.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
bion: `LocalDate::toString` formats the date slightly differently than the one format used throughout the app

bion: See example: https://github.com/civiform/civiform/blob/main/server/app/services/applicant/question/DateQuestion.java#L57

It'd be good to move to using `DateConverter` for formatting all dates in the system so that we can be sure they're displayed and parsed consistently.

server/test/services/applicant/ApplicantDataTest.java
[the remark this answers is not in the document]
swatkat1: I was able to do it without using ResourceCreator.

server/app/services/ti/TIClientCreationResult.java
```diff
@@ -0,0 +1,42 @@
...
+  public static TIClientCreationResult failure(
+      Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
+    return new TIClientCreationResult(Optional.of(form));
+  }
+
+  public boolean isSuccessful() {
+    return form.isEmpty();
+  }
+
+  public Optional<Form<AddApplicantToTrustedIntermediaryGroupForm>> getForm() {
```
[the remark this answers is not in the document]
swatkat1: The Optional<> is needed to check if it is successful or not. I am following the pattern as specified in ApiKeyCreationResult. If it needs to change, please document them.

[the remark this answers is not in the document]
bion: A major difference from `ApiKeyCreationResult` is that `ApiKeyService` needs to return data to the caller upon success, whereas that doesn't appear to be the case here

browser-test/src/support/ti_dashboard.ts
```diff
@@ -0,0 +1,52 @@
+import {Page} from 'playwright'
+
+// This class is to test Civiform in the TI path
```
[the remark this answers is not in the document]
swatkat1: Done

server/app/services/ti/TIClientCreationResult.java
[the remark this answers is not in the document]
swatkat1: In the next iteration, where we will be filtering the applicants by Name/Dob, this class can also store the search results. This is the first step to that bigger change.

server/app/services/DateConverter.java
```diff
@@ -32,6 +32,16 @@ public String formatIso8601Date(Instant instant) {
     return DateTimeFormatter.ISO_DATE.withZone(zoneId).format(instant);
   }
 
+  /**
+   * Converts a DateString in "YYYY-MM-dd" format to LocalDate type Throws DateTimeParseException
+   */
+  public LocalDate parseStringToLocalDate(String dateString) {
```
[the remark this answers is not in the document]
swatkat1: Done

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
swatkat1: Done

server/app/services/ti/TIClientCreationResult.java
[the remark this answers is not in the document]
swatkat1: @bion and @clouser2  - to your point- will it be better if I get rid of the result class and simply return the Optional<FormObject> from the service class. If in future i need a result class, i will add it. Does this sounds good?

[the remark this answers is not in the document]
bion: @swatkat1 yes, that sounds good. For returning search results I'd expect that to be a third method on the service class that has its own return value -- I don't see a reason for the create, update, and search/list methods to have the same return value.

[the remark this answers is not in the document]
clouser2: Not 100% sure I agree given that we return an initialized Optional on success. If we go that route, I'd be very clear in the docstring that an empty result implies success while a non-empty result implies error.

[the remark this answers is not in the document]
bion: Better would probably be always returning the form, the controller can inspect it to see if it has errors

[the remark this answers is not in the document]
swatkat1: ok..

server/app/services/ti/TIClientCreationResult.java
swatkat1: I have deleted this class

server/test/services/ti/TrustedIntermediaryServiceTest.java
swatkat1: done!

server/app/services/ti/TIClientCreationResult.java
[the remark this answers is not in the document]
swatkat1: this is done!

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: Done

server/app/services/DateConverter.java
```diff
@@ -53,4 +68,8 @@ public String renderDate(Instant time) {
     ZonedDateTime dateTime = time.atZone(zoneId);
     return dateTime.format(DateTimeFormatter.ofPattern("yyyy/MM/dd"));
   }
+  /** Formats an {@link LocalDate} to a String. */
+  public String renderDate(LocalDate date) {
+    return date.format(DateTimeFormatter.ofPattern("yyyy/MM/dd"));
```
bion, line 73: suggestion (blocking): extract a constant for `DateTimeFormatter.ofPattern("yyyy/MM/dd")` and use it here and on L69

server/test/services/ti/TrustedIntermediaryServiceTest.java
```diff
@@ -0,0 +1,242 @@
...
+    formFactory = instanceOf(FormFactory.class);
+    profileFactory = instanceOf(ProfileFactory.class);
+    Applicant managedApplicant = createApplicant();
+    createTIWithMockedProfile(managedApplicant);
+    repo = instanceOf(UserRepository.class);
+    profileFactory.createFakeTrustedIntermediary();
+  }
+
+  @Test
+  public void testAddClientWithInvalidDOB() {
```
bion, line 42: suggestion (blocking): please correct the abbreviation to `...Dob`

bion, line 42: suggestion (blocking): please correct the abbreviation to `...Dob`

bion, line 42: suggestion (blocking): please use the test method naming convention `methodUnderTest_contextForTheTest` e.g. `addClient_withInvalidDob`

server/test/services/applicant/ApplicantDataTest.java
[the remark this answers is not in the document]
bion: @swatkat1 which test covers this case?

bion (changes requested):

browser-test/src/support/ti_dashboard.ts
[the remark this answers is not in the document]
clouser2: I'm not sure I follow regarding what command you're asking about. I'd imagine this could be converted to something that:
  1) Finds all of the applicant rows through a query selector.
  2) For each row, does a subquery for the relevant field inside (similar to what each one is doing now).
  3) Confirms that ALL of the entries match and returns true if so, false otherwise.

Still not blocking, but likely would improve our confidence that everything's working.

clouser2 (approved):

bion (changes requested):
Please see last comments

bion (approved):

PR #3106, opened 2022-08-09
Added screenshots to landing page test with `jest-image-snapshot`

### LINKED ISSUE #2387: Add screenshot diffing tests
We should look into setting up testing where screenshots are automatically taken and compared to previous screenshots. This can help us catch unintentional UI changes, and verify intentional ones.

th174:
### Description

Added screenshots to all UI browser tests with jest-image-snapshot
You can save and validate a screenshot with await validateScreenshot(pageObject);
This will automatically save a screenshot if non exists, and compare it with the previous screenshot if one does exist.

## Release notes:

### Checklist

### Issue(s)

Fixes #2387


th174:
I can't debug the tests because I can't run `run-browser-tests` locally, due to my M1 Mac (arm64) not being able to run the browser tests which require (amd64). Please feel free to pick up this pull request if you don't have hardware incompatibility issues.

browser-test/src/support/index.ts
```diff
@@ -205,3 +212,19 @@ export const validateAccessibility = async (page: Page) => {
...
+  matchImageSnapshotOptions?: MatchImageSnapshotOptions,
+) => {
+  expect(
+    await page.screenshot({
+      ...pageScreenshotOptions,
+    }),
+  ).toMatchImageSnapshot({
+    failureThreshold: 0.03,
+    failureThresholdType: 'percent',
+    ...matchImageSnapshotOptions,
```
nbeloglazov, line 228: Can we group screenshots by test files? Currently they all stored in one giant directory which might become harder to work with the more screenshots we get. Putting them to per-test-file directories would help a bit.

Additionally we should anticipate mobile tests as well. Currently all screenshots are taken on desktop. But civiform needs to be well-optimized for mobile and that will be important part of the browser testing. We could add "desktop/mobile" suffix to screenshots from the start or put them in separate "desktop"/"mobile" directories so that later we won't have to move a lot of files.

bin/run-browser-tests
```diff
@@ -13,6 +13,7 @@ docker volume create browser-tests-node-modules
 # Additionally we map node_modules to a separate volume so that it doesn't
 # conflict with node_modules created locally.
 docker run --rm -it \
+  --platform linux/x86_64 \
```
nbeloglazov, line 16: Can you clarify why is this needed? I thought platform is the part of the image itself? Though I don't understand docker that well.

bin/run-browser-tests-local
```diff
@@ -4,4 +4,5 @@
 
 source bin/lib.sh
 
+cd browser-test/ || exit
```
nbeloglazov, line 7: Thanks for fixing it!
Should you go back to the root folder after test completed?

nbeloglazov (changes requested):
High-level comment. It might be too late to separate, but for the purpose of making review simpler I would split it into 2 or more PRs. First PR adds infrastructure (screenshot functions, updates scripts, etc) and adds screenshots to 1-2 tests as examples.

And later PRs just add screenshots calls. So that reviewer can be more focused reviewing just screenshots (and there are a lot of them).

Would it be possible to do or too cumbersome to split now?

bin/run-browser-tests-local
th174: Changing directory only changes the directory for the scope of the script itself. The terminal calling the script isn't affected, so there's no need to cd back.

bin/run-browser-tests
th174: This was an attempt to get the browser tests to run on my M1 arm64 Mac. It does allow the docker image to startup, but the test fails later anyway. I'll remove this.

browser-test/src/support/index.ts
th174: Yes, you can specify a file path parameter for saving the images. They're currently already grouped according to the test name, but we can also put them in directories as well.

th174:
> High-level comment. It might be too late to separate, but for the purpose of making review simpler I would split it into 2 or more PRs. First PR adds infrastructure (screenshot functions, updates scripts, etc) and adds screenshots to 1-2 tests as examples.

> And later PRs just add screenshots calls. So that reviewer can be more focused reviewing just screenshots (and there are a lot of them).

> Would it be possible to do or too cumbersome to split now?

I'll make this PR just the infrastructure, and I'll leave adding the screenshots to other people as I can't test the screenshots myself anyway.

browser-test/src/support/index.ts
nbeloglazov, line 228: Let's group screenshots by test directories matching test names. And skip "ts" in the direcory name. So for your test it will be be `__snapshots__/landing_page_test/the-landing-page-it-has-login-options-1-snap.png`
Though "snap" part also seems reduntant. Not sure what it supposed to indicate in their default schema.

browser-test/src/support/index.ts
```diff
@@ -205,3 +212,19 @@ export const validateAccessibility = async (page: Page) => {
 
   expect(results).toHaveNoA11yViolations()
 }
+
+export const validateScreenshot = async (
```
nbeloglazov, line 216: Let's add a function comment explaining that it takes a screenshot and saves it to a file `__snapshots__/test_file_name/name-of-the-test.png`. And that depending on whether screenshot is already present it is either diffing or adding a new screenshot. I think this function will be the main place where engineers learn about how screenshots work so it's important to have a good explanation here.

nbeloglazov (changes requested):
Overall looks good. One issue I want to resolve before submitting is the question of running local vs docker. As we talked in slack, running local is likely always produce wrong results specific to developer machine and we might restrict taking screenshots only when running within docker. I'm going to bring it up during eng standup today as not all folks probably seen the thread.

browser-test/src/landing_page.test.ts
```diff
@@ -1,10 +1,11 @@
-import {startSession, endSession} from './support'
+import {endSession, startSession, validateScreenshot} from './support'
 
 describe('the landing page', () => {
   it('it has login options', async () => {
     const {browser, page} = await startSession()
 
     expect(await page.textContent('html')).toContain('Continue as guest')
+    await validateScreenshot(page, {fullPage: true})
```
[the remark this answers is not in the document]
th174: It takes a screenshot of the viewport by default, that is, the visible area in the browser.

nbeloglazov:
Discussed it during the standup. Let's add a feature that disables screenshot checks when running with `bin/run-browser-tests-local` so that they ran and can be updated only when running from docker.

browser-test/src/support/index.ts
th174: Done

browser-test/src/support/index.ts
```diff
@@ -205,3 +212,38 @@ export const validateAccessibility = async (page: Page) => {
...
+ * If the screenshot already exists, compare the new screenshot with the
+ * existing screenshot, and save a pixel diff instead if the two don't match
+ */
+export const validateScreenshot = async (
+  page: Page,
+  pageScreenshotOptions?: PageScreenshotOptions,
+  matchImageSnapshotOptions?: MatchImageSnapshotOptions,
+) => {
+  // Do not make image snapshots when running locally
+  if (isLocalDevEnvironment()) {
```
nbeloglazov, line 228: Looking at the implementation of isLocalDevEnvironment, it returns true for civiform:9000 and localhost:9999. 
civiform:9000 is used in run-browser-tests [here](https://github.com/civiform/civiform/blob/main/browser-test/bin/wait_for_server_start_and_run_tests.sh#L7)
localhost:9999 is used in [run-browser-tests-local](https://github.com/civiform/civiform/blob/main/bin/run-browser-tests-local#L7). 
And we need to disable it only for the second case but not the first one.

nbeloglazov: Let's add another environment variable like DISABLE_SCREENSHOTS and set it in local run. We might need to set it in https://github.com/civiform/civiform/blob/main/bin/run-prober as well. So it will be a good to have a dedicated param that can be set to disable screenshots.

nbeloglazov (approved):
I cloned PR and tested it locally. Overall works good. Few notes. Some of them can be addressed in a follow up PR
1. Add __ diff_output __ directory to .gitignore. 
2. Error message when screenshot failes is pretty unreadable because it prints whole image in base64. We should disable it somehow and leave just "here is the diff file". I hope that jest has options for that. Also I would love if we could print something like "to update screenshots re-run tests with some option". 
3. Screenshots live in "browser-test/src/__image_snapshots__" folder. `src` isn't appropriate folder as it's not really source code. Also personally __ prefix and suffix doesn't match our directories as we never use such prefixes. But that's less critical.

th174:
> Error message when screenshot failes is pretty unreadable because it prints whole image in base64. We should disable it somehow and leave just "here is the diff file". I hope that jest has options for that. Also I would love if we could print something like "to update screenshots re-run tests with some option".

That actually is the default option. I specifically turned on the option to print the whole image in base64 because it's useful to examine diffs if tests fail in CI. You can copy and paste the base64 image into your browser URL bar and view the diff, whereas trying to view the image locally onto the CI server is much more difficult.

nbeloglazov:
I see. Good point about CI. I guess we have two different usecases then. For CI/github actions we should look into uploading screenshots [as artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). I haven't tried it and don't know how it looks in practice, but the doc mentions screenshots. But we can do it as a follow up.

th174:
Uploading artifacts looks like a simple workflow step, so I went ahead and made the change

nbeloglazov:
Thanks! I guess we can test whether it actually works only after you submit it and the action will run few times?

th174:
Right, I'll have to submit it, and then we can test by intentionally making a PR with a failing test, to see if the artifacts get generated.

PR #3112, opened 2022-08-09
Change label from implicit to explicit for radio buttons

jeanniefu:
### Description

Explicit labels (where `label` and `input` are linked by IDs) are better supported by assistive technology than implicit labels (where the `label` element wraps the `input` element). See [link](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly). This PR does not affect styling.

Also updates the input id to be randomly generated instead of using the text. The text is not unique per page (e.g. we could have 2 different radio questions with the same answer).

Before:
AT5vCqyrUdRx2Zu

After (no change in UI):
7q4KfYwsFXgxh6z

## Release notes:

Change label from implicit to explicit for radio buttons for accessibility.

### Checklist


jeanniefu:
Actually I need to fix the unit test first. Reverting back to draft.

bion (approved):

PR #3115, opened 2022-08-09
Add FeatureFlag overrides utility.

### LINKED ISSUE #3071: Create way to change feature flag state for browser tests
Feature Flags allow us to guard new functionality behind configuration variables, as well offer different optional experiences.

As these are configured via the system configuration there is no way to vary them in browser tests to ensure correct behavior for the different flag values.

We need to create such a mechanism.

shanemc-goog:
### Description

Adds the ability to override feature flag values via HTTP, intended to be used by browser tests to test different values for a feature false and human evaluation of features in staging environments.

This is explicitly not intended for production use, and its utility is guarded by an environment variable.

1. Adds FeatureFlagOverrides to store/supply the overrides, and FeatureFlagOverridesController to set them.
2. FeatureFlagsModule is the sole consumer of the overrides and provider of the feature flags effective value.

Note: the consumers of @ApplicationStatusTrackingEnabled had to be updated to use a Provider as Controller and therefor View classes aren't instantiated every request so otherwise cache the value of the flag.

## Release notes:

Creates a mechanism to set feature flag values via HTTP for development and testing.

### Checklist

### Issue(s)

Fixes #3071


shanemc-goog:
Note: Tests incoming, wanted a first look review on the code.

server/conf/routes
```diff
@@ -100,6 +100,9 @@ GET     /admin/apiKeys/new                          controllers.admin.AdminApiKe
 POST    /admin/apiKeys                              controllers.admin.AdminApiKeysController.create(request: Request)
 POST    /admin/apiKeys/:apiKeyId/retire             controllers.admin.AdminApiKeysController.retire(request: Request, apiKeyId: Long)
 
+GET     /admin/feature/:flag/enable                 controllers.FeatureFlagOverrideController.enable(request: Request, flag: String)
```
clouser2, line 103: Since this is only intended for use in non-production (i.e. dev) environments, what about grouping them with the existing `/dev/` routes (e.g. `/dev/features/enable`?

server/app/controllers/FeatureFlagOverrideController.java
```diff
@@ -0,0 +1,32 @@
...
+import auth.Authorizers;
+import javax.inject.Inject;
+import org.pac4j.play.java.Secure;
+import play.mvc.Http.Request;
+import play.mvc.Result;
+
+/**
+ * Allows for overriding of feature flags by the CiviForm Admin via HTTP request.
+ */
+public class FeatureFlagOverrideController extends CiviFormController {
```
clouser2, line 11: How about putting the controllers / views in a `/dev` directory, similar to what we do for the `/dev/seed/` routes?

clouser2, line 11: If possible, can you mark this as `final` to make the decision to allow subclassing explicit rather than implicit?

server/app/controllers/FeatureFlagOverrideController.java
```diff
@@ -0,0 +1,32 @@
...
+  private final FeatureFlagOverrides overrides;
+
+  @Inject
+  public FeatureFlagOverrideController(FeatureFlagOverrides overrides) {
+    this.overrides = overrides;
+  }
+
+  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  public Result enable(Request request, String flagName) {
+    overrides.setOverride(flagName, "true");
```
clouser2, line 21: Right now, the only use case for overrides is to always use a boolean. Until we have a use case otherwise, would it make sense for setOverride to take a string key and a boolean value?

server/app/controllers/FeatureFlagOverrideController.java
```diff
@@ -0,0 +1,32 @@
...
+public class FeatureFlagOverrideController extends CiviFormController {
+  private final FeatureFlagOverrides overrides;
+
+  @Inject
+  public FeatureFlagOverrideController(FeatureFlagOverrides overrides) {
+    this.overrides = overrides;
+  }
+
+  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  public Result enable(Request request, String flagName) {
```
clouser2, line 20: Here and in the disable handler, I'd extend `DevController` and call `isDevOrStagingEnvironment()` for consistency.

server/app/annotations/FeatureFlagOverrides.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  /**
+   * Returns the override value for {@flag} if it is present and overrides are enabled.
+   *
+   * Parsing follows {@link Boolean#parseBoolean(String)} rules.
+   */
+  public Optional<Boolean> getOverrideBoolean(String flag) {
+    logger.error(overrides.toString());
```
clouser2, line 51: Why do we want this logger statement when retrieving a single value? I'd assume that this will be called many times and just end up flooding the console.

server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -47,14 +51,14 @@ public final class ProgramIndexView extends BaseHtmlView {
...
-  private final boolean statusTrackingEnabled;
+  private final Provider<Boolean> statusTrackingEnabled;
 
   @Inject
   public ProgramIndexView(
       AdminLayoutFactory layoutFactory,
       Config config,
       DateConverter dateConverter,
-      @ApplicationStatusTrackingEnabled boolean statusTrackingEnabled) {
+      @ApplicationStatusTrackingEnabled Provider<Boolean> statusTrackingEnabled) {
```
clouser2, line 58: There are two other places that inject this value that also need to be updated:
* AdminProgramStatusesController
* ProgramTranslationView

server/conf/application.conf
```diff
@@ -598,3 +598,10 @@ api_applications_list_max_page_size = ${?CIVIFORM_API_APPLICATIONS_LIST_MAX_PAGE
 # In development features.
 application_status_tracking_enabled = false
 application_status_tracking_enabled = ${?CIVIFORM_APPLICATION_STATUS_TRACKING_ENABLED}
+
+# If enabled allows for setting of feature flags via HTTP.
```
clouser2, line 602: Agree with this sentiment entirely.

To further reduce the chances that this is accidentally set, I'd set the entry to false here and allow the environment variable override from application.dev.conf and application.test.conf.

Something I think that would make the most sense long-term is for us to have an application.prod.conf and always set this to false there.

clouser2 (changes requested):

server/app/controllers/FeatureFlagOverrideController.java
shanemc-goog: I was tacitly trying to follow the configuration class which I think just does bare strings, and relies on the user doing the right thing wrt to the value in the configuration file and code getting with the right type...

I don't have a strong opinion either way though if you think overloads in that class are better.

server/app/controllers/FeatureFlagOverrideController.java
shanemc-goog: done!

server/app/controllers/FeatureFlagOverrideController.java
shanemc-goog: done

server/app/annotations/FeatureFlagOverrides.java
shanemc-goog: it was left over debugging, removed.

server/app/views/admin/programs/ProgramIndexView.java
shanemc-goog: done!!

server/conf/application.conf
shanemc-goog: Great idea, moved it.

shanemc-goog (commented):
I did all the non file moving stuff and route changes.  will do them next.

server/conf/routes
shanemc-goog: done

server/app/controllers/FeatureFlagOverrideController.java
shanemc-goog: done

shanemc-goog (commented):
PTAL, Added tests and fixed up an issue with GuiceApplicationBuilder in my new test and in DevSeedController in #3127

server/conf/application.test.conf
```diff
@@ -16,6 +16,10 @@ play.evolutions.db.default.enabled = "true"
 azure.blob.container = "super cool blob container name"
 azure.blob.account = "my awesome azure account name"
 
+auth.applicant_idp = "disabled"
```
clouser2, line 19: Not blocking: Was this change necessary?

clouser2 (approved):

server/conf/application.test.conf
shanemc-goog: It was just lower down and I moved it up.

PR #3125, opened 2022-08-10
add option for civiform that auto approves terraform

azizova-leyla:
### Description

Needed to automatically deploy staging.

## Release notes:

### Checklist

### Issue(s)

### DIFF
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -37,6 +37,14 @@ class ConfigLoader:
     def civiform_mode(self):
         return os.environ['CIVIFORM_MODE']
 
+    @property
+    def use_local_backend(self):
+        return os.environ['USE_LOCAL_BACKEND']
+
+    @property
+    def skip_confirmations(self):
+        return os.environ['SKIP_CONFIRMATIONS']
+
     def load_config(self):
         self._load_config()
         return self.validate_config()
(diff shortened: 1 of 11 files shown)


cloud/shared/variable_definitions.json
```diff
@@ -60,7 +60,7 @@
     "secret": false,
     "tfvar": true,
     "type": "enum",
-    "values": ["prod", "staging", "dev", "test"]
+    "values": ["prod", "staging", "ci", "dev", "test"]
```
nbeloglazov, line 63: We should write down somewhere what is the difference between all these modes. For example I don't understand now why we need dev when we can use staging. The only thing that `dev` does right now, it seems, is that it stores terraform state locally instead of s3. Which doesn't seem to be that useful, at least to me. 

And test is specific to azure and just produces plan without actually running, right?

azizova-leyla: yeah, I've been thinking about that too. Not sure where is the best place to write this down. as of right now:
prod - will set some things to be indestructible, use for real prod
staging - same as prod, but easier to destroy if needed
ci - skip confirmation checks
dev - store tfstate locally
test - validate things and exit early

I could alternatively just have prod and prod/staging/test here and add new flags for auto-approve and local-state.

nbeloglazov: I'd vote for having new flag as you suggested. It will make it more explicit.

cloud/shared/bin/lib/config_loader.py
```diff
@@ -37,6 +37,14 @@ def aws_region(self):
     def civiform_mode(self):
         return os.environ['CIVIFORM_MODE']
 
+    @property
+    def use_local_backend(self):
+        return os.environ['USE_LOCAL_BACKEND']
```
nbeloglazov, line 42: I think this will throw error if the variable is not defined. Maybe use `os.environ.get('USE_LOCAL_BACKEND', False)`?

azizova-leyla: done

cloud/shared/variable_definitions.json
azizova-leyla: done

nbeloglazov (approved):

PR #3131, opened 2022-08-11
Print some few links at the end of AWS setup and deploy

nbeloglazov:
### Description

1. The links should help deployer verify whether deployment was successful or not. 
2. At the end of setup we remind to update domain CNAME record to point to the freshly created load balancer url where civiform app is running.
3. Move second run of `terraform apply` to azure-specific post-terraform step. It's not needed by AWS and messes up "print final message" code which is done in post-terraform setup. 

### Checklist

### DIFF
--- a/cloud/aws/bin/deploy.py
+++ b/cloud/aws/bin/deploy.py
@@ -1,6 +1,8 @@
 #! /usr/bin/env python3
 import sys
 
+from cloud.aws.templates.aws_oidc.bin import resources
+from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
 from cloud.shared.bin.lib import terraform
 from cloud.shared.bin.lib import tf_apply_setup
 from cloud.aws.bin.lib import backend_setup
@@ -21,6 +23,14 @@ def main():
 
     if config_loader.is_test():
         print('Test completed')
+
+    aws_cli = AwsCli(config_loader)
+    fargate_service = f'{config_loader.app_prefix}-{resources.FARGATE_SERVICE}'
+    tasks_url = aws_cli.get_url_of_fargate_tasks(
+        config_loader.app_prefix, fargate_service)
+    print()
+    print('Deployment finished. You can monitor civiform tasks status here:')
+    print(tasks_url)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


cloud/aws/bin/deploy.py
```diff
@@ -22,6 +24,14 @@ def main():
     if config_loader.is_test():
         print('Test completed')
 
+    aws_cli = AwsCli(config_loader)
+    fargate_service = f'{config_loader.app_prefix}-{resources.FARGATE_SERVICE}'
+    tasks_url = aws_cli.get_url_of_fargate_tasks(
```
azizova-leyla, line 29: inline?

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -53,9 +53,18 @@ def restart_ecs_service(self, cluster: str, service_name: str):
                 f'--service={service_name}', f'--cluster={cluster}'
             ])
 
+    def get_load_balancer_dns(self, name: str) -> str:
```
azizova-leyla, line 56: we can get this as output variable from terraform https://registry.terraform.io/modules/cn-terraform/ecs-fargate-service/aws/latest?tab=outputs aws_lb_lb_dns_name

See https://www.terraform.io/cli/commands/output

azizova-leyla (changes requested):

cloud/aws/bin/deploy.py
nbeloglazov: Done.

cloud/aws/templates/aws_oidc/bin/aws_cli.py
nbeloglazov: Yes we can. I don't see much benefit though. We already have to use AWS CLI for various stuff so for consistency I think it's better to use it. One benefit of using terraform is that we don't need to rely on LB name which is defined in terraform and hardcoded in python code. It doesn't seem like a that big of a problem though.

cloud/azure/templates/azure_saml_ses/bin/setup.py
```diff
@@ -56,6 +57,8 @@ def setup_log_file(self):
     def post_terraform_setup(self):
         self._get_adfs_user_inputs()
         self._configure_slot_settings()
+        # Why are running terraform again for azure?
```
azizova-leyla, line 60: we're running it because get_adfs_user_inputs updates secret variables.

azizova-leyla (approved):

PR #3133, opened 2022-08-11
Refactors templates handling in python deploy script

nbeloglazov:
### Description

There are changes in this PR. 

1. Make `SetupTemplate` class to be explicit base class for Setup and Destroy AWS/Azure classed by adding `pre_terraform_destroy` and `post_terraform_destroy` methods there. Both AWS Destroy and Azure Destroy classes inherit from `SetupTemplate`. Maybe we should redo inheritance in this case. For example have single AWS and single Azure classes which encapsulate both setup and destroy logic.

2. Create instances of AWS/Azure Setup/Destroy classes in setup_class_loaders instead of returning class objects. That allows to specify return type and improve type inference for IDEs. And generally hides that non-standard reflection logic within that file.

### Checklist

### DIFF
--- a/cloud/shared/bin/lib/setup_template.py
+++ b/cloud/shared/bin/lib/setup_template.py
@@ -1,14 +1,13 @@
-#! /usr/bin/env python3
-import shutil
-import tempfile
-import shutil
-
-from cloud.shared.bin.lib.config_loader import ConfigLoader
 """
 Template Setup
 
 These functions need to be defined for every template.
 """
+
+import tempfile
+import shutil
+
+from cloud.shared.bin.lib.config_loader import ConfigLoader
 
 
 class SetupTemplate:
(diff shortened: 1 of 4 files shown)


cloud/shared/bin/lib/setup_template.py
```diff
@@ -34,12 +33,6 @@ def setup_log_file(self):
         _, self.log_file_path = tempfile.mkstemp()
         print(" - TODO: Setup log file here.")
 
-    def _make_backend_override(self):
-        current_directory = self.config.get_template_dir()
-        shutil.copy2(
```
clouser2, line 39: Was removal of this code necessary? Or was it unused?

clouser2 (approved):
This looks great! Nice cleanup.

cloud/shared/bin/lib/setup_template.py
nbeloglazov: It's a duplicate of the same function 20 lines above.

PR #3134, opened 2022-08-11
Reinstate status tracking tests and update Override functionality to be session stored.

### LINKED ISSUE #3071: Create way to change feature flag state for browser tests

shanemc-goog:
### Description

Replaced the

- FeatureFlag system to be an injected utility (no longer per feature annotation)
- The override system to be stored in Session Cookies and not as a system wide configuration.  This provides more flexibility for tests as well in replicated environments like Seattle Staging.

Update Status Tracking tests, which use a feature flag, to enable the flag in the test rather than assume it is.

## Release notes:

NA testing only.

### Checklist

### Issue(s)

Fixes #3071


browser-test/src/admin_application_statuses.test.ts
```diff
@@ -25,9 +27,16 @@ describe.skip('view program statuses', () => {
...
+    // Need to navigate back to the Admin page after enabling the feature flag.
+    await gotoEndpoint(pageObject, '')
   })
 
-  describe('without program statuses', () => {
+  afterAll(async () => {
+    await disableFeatureFlag(pageObject, 'application_status_tracking_enabled')
+  })
+
+  describe.skip('without program statuses', () => {
```
shanemc-goog, line 39:
```suggestion
  describe('without program statuses', () => {
```

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -103,7 +112,7 @@ describe.skip('view program statuses', () => {
       expect(await adminPrograms.isStatusSelectorVisible()).toBe(true)
     })
 
-    it('shows default option as placeholder', async () => {
+    it.skip('shows default option as placeholder', async () => {
```
shanemc-goog, line 115:
```suggestion
    it('shows default option as placeholder', async () => {
```

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -9,11 +9,22 @@ import {
...
+  enableFeatureFlag,
+  gotoEndpoint,
+  disableFeatureFlag,
 } from './support'
 import {Page} from 'playwright'
+import EmptyFunction = jest.EmptyFunction
 
-// TODO(#3071): Re-enable when the feature flag is controllable in tests.
-describe.skip('view program statuses', () => {
+const describeIf = (
```
clouser2, line 19: If we expect this to be a long-term solution, I'd suggest moving the definition to something in `support`.

browser-test/src/support/index.ts
```diff
@@ -180,6 +180,14 @@ export const seedCanonicalQuestions = async (page: Page) => {
   await page.click('#canonical-questions')
 }
 
+export const enableFeatureFlag = async (page: Page, flag: string) => {
+  await page.goto(BASE_URL + '/dev/feature/$flag/enable')
+}
+
+export const disableFeatureFlag = async (page: Page, flag: string) => {
+  await page.goto(BASE_URL + '/dev/feature/$flag/disable')
```
clouser2, line 188: I believe this won't actually interpolate the String. I believe the correct interpolation syntax uses a backtick and `${`:

```
`/dev/feature/${flag}/disable`
```

browser-test/src/support/index.ts
```diff
@@ -180,6 +180,14 @@ export const seedCanonicalQuestions = async (page: Page) => {
...
+}
```
clouser2, line 189: The above interpolation leads me to believe that the calls to enableFeatureFlag / disableFeatureFlag are a no-op.

Given the current implementation where probers would entirely skip the tests, do we actually need to have the enable/disable as part of this PR?

At a high level, I think something that might be missing is once again disabling status tracking in the relevant `.conf` file. We may want to add a new one specifically for browser tests that inherits from whatever is there right now and disables the feature. That would hopefully make it explicit around whether the feature flag forcing mechanism was working or not.

clouser2 (changes requested):

shanemc-goog:
Going to take an alternative approach based on Brett's suggestion

browser-test/src/support/index.ts
shanemc-goog: You were correct, I fixed that bug.

We also need to disable the .conf value which is a few more steps.  I made #3237 for that

browser-test/src/support/index.ts
shanemc-goog: fixed! thank you!

browser-test/src/admin_application_statuses.test.ts
shanemc-goog: removed

server/app/controllers/dev/FeatureFlagOverrideController.java
```diff
@@ -1,40 +1,43 @@
...
 
-  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  @Secure(authorizers = Labels.ANY_ADMIN)
   public Result disable(Request request, String flagName) {
     if (!isDevOrStagingEnvironment()) {
       return notFound();
     }
-    overrides.setOverride(flagName, "false");
-    return ok();
+    String redirectTo = request.getHeaders().get("referer").orElse("/");
```
bion, line 40: nit: use `play.mvc.Http.REFERER` constant instead

server/app/controllers/dev/FeatureFlagOverrideController.java
```diff
@@ -1,40 +1,43 @@
...
 
-  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  @Secure(authorizers = Labels.ANY_ADMIN)
   public Result enable(Request request, String flagName) {
     if (!isDevOrStagingEnvironment()) {
       return notFound();
     }
-    overrides.setOverride(flagName, "true");
-    return ok();
+    String redirectTo = request.getHeaders().get("referer").orElse("/");
```
bion, line 30: nit: use `play.mvc.Http.REFERER` constant instead

server/app/featureflags/FeatureFlags.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * Returns the current setting for {@code flag} from session cookie if present or then the system
+   * {@link Config}.
+   */
+  private boolean getFlagEnabled(Request request, String flag) {
+    if (!areOverridesEnabled()) {
+      return false;
+    }
```
bion, line 43: issue (blocking): the config var is called "feature_flag_overrides_enabled" which suggests that only overriding is enabled/disabled, not that globally all feature flags are disabled.

server/app/featureflags/FeatureFlags.java
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      return request
+          .session()
+          .get(flag)
+          .map(Boolean::parseBoolean)
+          .orElseGet(() -> config.getBoolean(flag));
+    } catch (ConfigException.Missing ignore) {
+      // Ignore if the Config doesn't have the value configured.
+      return false;
```
bion, line 52: suggestion (blocking): please log a warning for the missing flag to help reduce confusion when there are typos

bion (changes requested):

shanemc-goog (commented):
PTAL

server/app/featureflags/FeatureFlags.java
shanemc-goog: done! good catch, added a test

server/app/featureflags/FeatureFlags.java
shanemc-goog: done!

server/app/controllers/dev/FeatureFlagOverrideController.java
shanemc-goog: done

server/app/controllers/dev/FeatureFlagOverrideController.java
shanemc-goog: done

server/app/controllers/dev/FeatureFlagOverrideController.java
```diff
@@ -1,40 +1,44 @@
...
   @Inject
-  public FeatureFlagOverrideController(
-      FeatureFlagOverrides overrides, Environment environment, Config configuration) {
+  public FeatureFlagOverrideController(Environment environment, Config configuration) {
     super(environment, configuration);
-    this.overrides = overrides;
   }
 
-  @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
+  @Secure(authorizers = Labels.ANY_ADMIN)
```
clouser2, line 26: Definitely doesn't matter right now, but we might want to allow anyone to call this since we already protect it based on whether we're a dev environment or not. As currently configured, we can't force an experiment as an applicant.

server/app/featureflags/FeatureFlags.java
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Returns the current setting for {@code flag} from session cookie if present or then the system
+   * {@link Config}.
+   */
+  private boolean getFlagEnabled(Request request, String flag) {
+    Optional<Boolean> configValue = Optional.empty();
+    if (config.hasPath(flag)) {
+      configValue = Optional.of(config.getBoolean(flag));
+    } else {
+      logger.warn("Feature flag requested for unconfigured flag: {}", flag);
```
clouser2, line 49: What about returning false explicitly here rather than continuing on? I think there may be a subtle bug where an unrecognized path with an override in the session would return the session value. For unrecognized paths, we should just always return false (or even throw an error in my opinion.

If that's the case, then I think we can make configValue just be a boolean, removing the need for the `.orElse` calls below.

clouser2 (approved):

server/app/featureflags/FeatureFlags.java
bion: +1

bion (approved):

server/app/featureflags/FeatureFlags.java
shanemc-goog: done.

My light reasoning for the former case didn't hold up I realized, good call.  I updated the tests a bit too to be verbose/clearer on the 3 moving parts here, config, overrides control, actual overrides.

PR #3136, opened 2022-08-11
Store playwright browsers within node_modules directory for docker tests

nbeloglazov:
### Description

This way wherever a new version of playwright is downloaded - it will update browsers as well. It helps to prevent cases where browsers downloaded to some ephemeral place which will be destroyed after docker container is killed while node_modules remain as they are mounted to a persistent volume.

### Checklist

### DIFF
--- a/bin/run-browser-tests-ci
+++ b/bin/run-browser-tests-ci
@@ -8,7 +8,6 @@ source bin/lib.sh
 docker::set_project_name_browser_tests
 
 docker volume create browser-tests-node-modules
-
 docker run \
   -v "$(pwd)/browser-test:/usr/src/civiform-browser-tests" \
   -v "browser-tests-node-modules:/usr/src/civiform-browser-tests/node_modules" \
--- a/bin/run-browser-tests-ci
+++ b/bin/run-browser-tests-ci
@@ -8,9 +8,12 @@ source bin/lib.sh
 docker::set_project_name_browser_tests
 
 docker volume create browser-tests-node-modules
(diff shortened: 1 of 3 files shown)


browser-test/bin/wait_for_server_start_and_run_tests.sh
```diff
@@ -5,6 +5,10 @@ set -euo pipefail
 START_TIME=$(date +%s)
 DEADLINE=$(($START_TIME + 500))
 SERVER_URL="http://civiform:9000"
+# Store playwright browsers within node_modules directory. This way playwright
+# library and browsers placed together and less likely go out of sync if
+# there are manipulations with docker volumes.
+PLAYWRIGHT_BROWSERS_PATH=0
```
kmcnellis, line 11: Can you link to the documentation on this? https://playwright.dev/docs/browsers#managing-browser-binaries-1

bin/run-browser-tests-ci
```diff
@@ -8,6 +8,7 @@ source bin/lib.sh
 docker::set_project_name_browser_tests
 
 docker volume create browser-tests-node-modules
+
```
kmcnellis, line 11: (Super nit- feel free to ignore) Probably better to skip this white space addition, since there's no other changes to this file

kmcnellis (approved):

browser-test/bin/wait_for_server_start_and_run_tests.sh
nbeloglazov: Done.

PR #3143, opened 2022-08-12
Allow updating statuses in the translation UI

### LINKED ISSUE #2754: Status Tracking: Add statuses and email content to the Programs translation manager
Update the "Manage Translations" panel to include statuses and their email content when specified.

- [ ] Send the updated content to the server to be saved.

clouser2:
### Description

In #3124, we render statuses based on the configured program, but attempting to save changes doesn't actually update the statuses in the database. This change achieves that.

As part of this, the `ProgramTranslationForm` needed to be refactored since there are a dynamic number of statuses for a program and we can't easily have a unique field name. I instead exposed this through using a `DynamicForm` (there's precedent of using this for API key CiviForm Admin UI) and wrapping up the field handling in that class so downstream code can just get a set of `LocalizationUpdate.StatusUpdate` objects.

Some notes:
  * We continue to not really do any validation in the form itself (outside of expecting fields to be present). This responsibility lies in ProgramService
  * For simplicity, if any of the program status is out of sync with the configured statuses, we fail the request and reload the translation UI with an error message. We could try and "patch" the statuses we could find, but that seems to add unnecessary complexity for the uncommon "stale tab" scenario.
  * LocalizedStrings previously had no way to clear translations for a given locale. I added that for this specific use case and left a TODO against cleanup bug to consolidate our other usages to support this.

## Testing notes:

The added browser test is temporarily skipped until feature flags can be controlled in browser tests. Prior to disabling the test, I allowed CI to run with the test enabled, which was successful (https://github.com/civiform/civiform/runs/7812004315?check_suite_focus=true).

Additionally, I set `application_status_tracking_enabled = false` in `application.dev.conf` and `application.test.conf` to simulate prober environment state and confirmed that tests pased.

## Release notes:
N/A

### Checklist

### Issue(s)

Fixes #2754; #3144


server/app/forms/translation/ProgramTranslationForm.java
```diff
@@ -1,40 +1,153 @@
...
+ */
+public final class ProgramTranslationForm {
+
+  public static final String DISPLAY_NAME_FORM_NAME = "displayName";
+  public static final String DISPLAY_DESCRIPTION_FORM_NAME = "displayDescription";
+
+  private final DynamicForm form;
+  private final int maxStatusTranslations;
+
+  private static final Lang DEFAULT_LANG = new Lang(LocalizedStrings.DEFAULT_LOCALE);
```
shanemc-goog, line 34: move up with the other consts.

sending this on its own as GH is being weird and curious if/where it shows up after sending.

shanemc-goog (changes requested):

server/app/forms/translation/ProgramTranslationForm.java
shanemc-goog: ah it shows now!

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
...
+      throws ProgramNotFoundException, OutOfDateStatusesException {
     ProgramDefinition programDefinition = getProgramDefinition(programId);
     ImmutableSet.Builder<CiviFormError> errorsBuilder = ImmutableSet.builder();
-    validateProgramText(errorsBuilder, "display name", displayName);
-    validateProgramText(errorsBuilder, "display description", displayDescription);
+    validateProgramText(errorsBuilder, "display name", localizationUpdate.localizedDisplayName());
+    validateProgramText(
+        errorsBuilder, "display description", localizationUpdate.localizedDisplayDescription());
+
+    // It is permissible to set the localized text / email for a status to empty, which falls back
```
shanemc-goog, line 244: This comment doesn't seem to address the if/throw and is more confusing than informative.  Maybe rephrase what's here and put this inside the utility method.

EG, what i means to be in sync is not clear from this code's context (being in sync seems good though), such that "empty status" semantics are more so hard to reason about.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
             .join());
   }
 
+  private boolean localizationStatusesAreInSyncWithConfiguredStatuses(
```
shanemc-goog, line 221: Doc what it means to be in sync

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
...
+            .collect(ImmutableList.toImmutableList());
+    ImmutableList<String> configuredStatusNames =
+        program.statusDefinitions().getStatuses().stream()
+            .map(StatusDefinitions.Status::statusText)
+            .collect(ImmutableList.toImmutableList());
+    return localizationStatusNames.equals(configuredStatusNames);
+  }
+
   @Override
   public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
```
shanemc-goog, line 235: I believe we need an @Transactional on this so that the read/verify/update doesn't get changed in the middle.

(I have no idea why it's lowercasing the T there)

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
...
     ImmutableSet.Builder<CiviFormError> errorsBuilder = ImmutableSet.builder();
-    validateProgramText(errorsBuilder, "display name", displayName);
-    validateProgramText(errorsBuilder, "display description", displayDescription);
+    validateProgramText(errorsBuilder, "display name", localizationUpdate.localizedDisplayName());
+    validateProgramText(
+        errorsBuilder, "display description", localizationUpdate.localizedDisplayDescription());
+
+    // It is permissible to set the localized text / email for a status to empty, which falls back
+    // on the English text.
+    if (!localizationStatusesAreInSyncWithConfiguredStatuses(
```
shanemc-goog, line 246: Suggestion: Make this a validateX() method and let it throw the exception like the above validate ones

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
...
+        ImmutableList.builder();
+    for (int statusIdx = 0;
+        statusIdx < programDefinition.statusDefinitions().getStatuses().size();
+        statusIdx++) {
+      LocalizationUpdate.StatusUpdate statusUpdateData =
+          localizationUpdate.statuses().get(statusIdx);
+      StatusDefinitions.Status existingStatus =
+          programDefinition.statusDefinitions().getStatuses().get(statusIdx);
+      StatusDefinitions.Status.Builder updateBuilder =
+          StatusDefinitions.Status.builder()
```
shanemc-goog, line 263: suggestion: If we add [toBuilder()](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance) to the AutoValue we can instead replace these two lines with `existinStatus.toBuilder()` which I think is a lot more readable in general.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -218,14 +218,67 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateProgramDefinition(
...
+              .setStatusText(existingStatus.statusText())
+              .setLocalizedStatusText(
+                  existingStatus
+                      .localizedStatusText()
+                      .updateTranslation(locale, statusUpdateData.localizedStatusText()));
+      if (existingStatus.localizedEmailBodyText().isPresent()) {
```
shanemc-goog, line 269: doc this if/else/throw, I think it's

// If the status has email content, update the localization to whatever was provided; otherwise if there's a localization update when there is no email content to localize, that indicates a mismatch between the frontend and the database.

server/app/views/admin/programs/ProgramTranslationView.java
```diff
@@ -102,55 +101,65 @@ private ImmutableList<DomContent> formFields(
...
-                                    .setValue(translationForm.getDisplayDescription())
+                                    .setValue(updateData.localizedDisplayDescription())
                                     .getInputTag(),
                                 defaultLocaleTextHint(program.localizedDescription())))));
     if (statusTrackingEnabled.get()) {
       String programStatusesLink =
           controllers.admin.routes.AdminProgramStatusesController.index(program.id()).url();
 
-      for (StatusDefinitions.Status s : program.statusDefinitions().getStatuses()) {
+      Preconditions.checkState(
```
shanemc-goog, line 122: If this fails it's an indication of a coding error right and not being out of sync with the FE?  We're reading the program and making the form/updates all in the same server request correct?

server/app/views/admin/programs/ProgramTranslationView.java
```diff
@@ -102,55 +101,65 @@ private ImmutableList<DomContent> formFields(
...
+          updateData.statuses().size() == program.statusDefinitions().getStatuses().size());
+      for (int statusIdx = 0; statusIdx < updateData.statuses().size(); statusIdx++) {
+        StatusDefinitions.Status configuredStatus =
+            program.statusDefinitions().getStatuses().get(statusIdx);
+        LocalizationUpdate.StatusUpdate statusUpdateData = updateData.statuses().get(statusIdx);
         ImmutableList.Builder<DomContent> fieldsBuilder =
             ImmutableList.<DomContent>builder()
                 .add(
+                    input()
+                        .isHidden()
```
shanemc-goog, line 132: How does the hidden input associated with the visible input?

server/test/forms/translation/ProgramTranslationFormTest.java
```diff
@@ -0,0 +1,195 @@
...
+                            .build()))
+                .build());
+  }
+
+  @Test
+  public void bindFromRequest_extraStatusesInFormBodyBeyondSpecifiedAreIgnored() throws Exception {
+    Request request = fakeRequest().bodyForm(REQUEST_DATA_WITH_TWO_TRANSLATIONS).build();
+
+    ProgramTranslationForm form =
+        ProgramTranslationForm.bindFromRequest(request, instanceOf(FormFactory.class), 1);
```
shanemc-goog, line 65: inline comment the magic numbers in these. 1 here 3 below.

server/test/forms/translation/ProgramTranslationFormTest.java
```diff
@@ -0,0 +1,195 @@
...
+                    ImmutableList.of(
+                        LocalizationUpdate.StatusUpdate.builder()
+                            .setConfiguredStatusText("first status configured text")
+                            .setLocalizedStatusText(Optional.of("first status text"))
+                            .setLocalizedEmailBody(Optional.of("first status email"))
+                            .build(),
+                        LocalizationUpdate.StatusUpdate.builder()
+                            .setConfiguredStatusText("second status configured text")
+                            .setLocalizedStatusText(Optional.of("second status text"))
+                            .setLocalizedEmailBody(Optional.of("second status email"))
```
shanemc-goog, line 55: remove this line to evaluate the missing email scenario?

server/test/forms/translation/ProgramTranslationFormTest.java
```diff
@@ -0,0 +1,195 @@
...
+                        LocalizationUpdate.StatusUpdate.builder()
+                            .setConfiguredStatusText("first status configured text")
+                            .setLocalizedStatusText(Optional.of("first status text"))
+                            .setLocalizedEmailBody(Optional.of("first status email"))
+                            .build()))
+                .build());
+  }
+
+  @Test
+  public void bindFromRequest_notEnoughStatusesInFormBodyAreOmmitted() throws Exception {
```
shanemc-goog, line 82: I'm not really following what this test is testing.  The form body has 3 items (which indicates the DB had 3 correct?), and the request has 2, so the 3rd is dropped?

browser-test/src/admin_program_translation.test.ts
```diff
@@ -47,7 +62,106 @@ describe('Admin can manage translations', () => {
     await endSession(browser)
   })
 
-  it('create a question and add translations', async () => {
+  // TODO(#3071): Re-enable when the feature flag is controllable in tests.
+  it.skip('creates a program with statuses and adds translations for program statuses', async () => {
```
shanemc-goog, line 66: skip isn't on the one above, just double checking this passed locally/GH

server/app/controllers/admin/AdminProgramTranslationsController.java
```diff
@@ -107,32 +108,28 @@ public Result update(Http.Request request, long programId, String locale)
...
-      return redirect(routes.AdminProgramController.index().url());
-    } catch (ProgramNotFoundException e) {
-      return notFound(String.format("Program ID %d not found.", programId));
+      return redirect(routes.AdminProgramController.index().url())
+          .flashing(
+              "success",
+              String.format(
+                  "Program translations updated for %s",
+                  localeToUpdate.getDisplayLanguage(LocalizedStrings.DEFAULT_LOCALE)));
+    } catch (OutOfDateStatusesException e) {
```
shanemc-goog, line 130: This catch should only be around updateLocalization

server/app/forms/translation/ProgramTranslationForm.java
```diff
@@ -1,40 +1,153 @@
...
+            })
+        .filter(maybeStatusUpdate -> maybeStatusUpdate.isPresent())
+        .map(maybeStatusUpdate -> (LocalizationUpdate.StatusUpdate) maybeStatusUpdate.get())
+        .collect(ImmutableList.toImmutableList());
   }
 
-  public void setDisplayName(String displayName) {
-    this.displayName = displayName;
+  public static String configuredStatusFieldName(int index) {
+    return String.format("status-%d-configured", index);
```
shanemc-goog, line 143: suggestion: these 3 name patterns are effectively a triple but they don't read as such or discernibly to me at least.

If I understand their relationship, here's a stab at more aligned names.
configured -> configured-status
statusText ->localized-status
email -> localized-email

shanemc-goog (changes requested):

server/app/services/program/ProgramServiceImpl.java
clouser2: YEESSSSS! Done. I imagine there are a lot more places we could likely do that. I'll hunt them down in a follow-on PR.

server/app/services/program/ProgramServiceImpl.java
clouser2: Yah, your explanation is correct. Added.

server/app/views/admin/programs/ProgramTranslationView.java
clouser2: Exactly. I made this a precondition check for that reason. If you'd prefer, we can throw an OutOfDateStatusesException and update the controller to handle that (though it should be unexpected). For now, I've left as-is, but I'm happy to switch to that pattern (or something else) if you'd prefer.

server/app/views/admin/programs/ProgramTranslationView.java
clouser2: Added a comment. Also renamed the method to `statusKeyToUpdateFieldName` per earlier feedback.

server/test/forms/translation/ProgramTranslationFormTest.java
clouser2: I'm not sure I follow. In this case, the request data indicates that there's a configured email (by it's presence). The "empty-or-unrovided-email" scenario is already tested in `bindFromRequest_emptyStatusOrEmailIsConsideredNotProvided`

server/test/forms/translation/ProgramTranslationFormTest.java
clouser2: Done. I forget the need to do this since my editor displays the variable name inline while I'm editing code.

server/test/forms/translation/ProgramTranslationFormTest.java
clouser2: This form is only responsible for parsing the form body data and doesn't have any true idea about what's configured inside the database outside of the argument passed to `bindFromRequest`.

In this case, we're testing an out-of-date-tab situation where the configured statuses count is 3 and there are only 2 provided in the form body. The goal is to ensure that trying to parse a non-present 3rd status wouldn't crash anything and would just return 2 at the end. Upstream callers would be responsible for caring about whether having 2 updates was acceptable or not (that's now documented in the implementation).

I updated the comment.

shanemc-goog (approved):

shanemc-goog:
LGTM, just the discussion on a few threads

server/app/views/admin/programs/ProgramTranslationView.java
clouser2: Ah, yes. They're all siblings in the form, the main thing grouping them together is the index. So there will be the following inputs (much DOM ommitted for clarity):

````
<form>
  <input name="status-key-to-update-0" value="first-status" hidden />
  <input name="localized-status-0" value="localized-first-status" />
  <input name="localized-email-0" value="localized-first-status-email-body" />
  
  <input name="status-key-to-update-1" value="second-status" hidden />
  <input name="localized-status-1" value="localized-second-status" />
  <input name="localized-email-1" value="localized-second-status-email-body" />
</form>
```

clouser2: Left a comment to try to clarify this.

clouser2 (commented):
I've addressed the comments on the open threads. Given the approval, I'm going to merge. If there's any additional feedback, I'm happy to address in a quick follow-on PR.

PR #3145, opened 2022-08-12
Fix few warnings in TS codebase

nbeloglazov:


jeanniefu (approved):

kmcnellis (commented):
Looks like quite a few tests are failing

kmcnellis (commented):
Looks like quite a few tests are failing

nbeloglazov:
Yeah :( Seems like we use old Typescript with typings that don't have `NodeListOf.forEach` method. Might have to undo it or explicitly convert to array.

kmcnellis (approved):

Issue #3148, opened 2022-08-12
SVG icons don't have consistent pixel sizing


clouser2:
Throughout the codebase, we set explicit width / height on SVGs. Each of the paths that's already present doesn't contain the original "size" represented by the SVG. This means that if two paths were copied based on a different pixel size, we'll get inconsistent clipping / centering when trying to render them at different sizes (see TODO referencing this bug).

We should do a pass through the codebase and import SVGs based on a consistent size (and include it in the HTML we output in order to make it easier to scale the icons). See https://css-tricks.com/scale-svg/.

nbeloglazov:
Adjusted slightly script that calculates inferred size of the icons and most of them are 20 and 24px. So should be pretty safe to bake-in sizes of each svg into its enum value.

![image](https://user-images.githubusercontent.com/252053/185500320-cad20802-08ff-4dd7-af9b-1bc05cc804fe.png)

PR #3150, opened 2022-08-12
Warn that translations will be cleared if the email is cleared when editing a status

clouser2:
### Description

In addition to the warning, I:
  * Repointed a TODO for SVG resizing at a more appropriate tech debt issue (#3148)
  * Removed a TODO that isn't necessary (clearing state can be achieved with a page refresh)

Screenshot:
Screen Shot 2022-08-12 at 2 42 02 PM

### Testing notes:

Status tracking browser tests are temporarily skipped until feature flags can be controlled within browser tests. I temporarily allowed CI to run with the tests enabled, which was successful (https://github.com/civiform/civiform/runs/7814428678?check_suite_focus=true).

## Release notes:

N/A

### Checklist

### Issue(s)

#2752; #3148


browser-test/src/support/admin_program_statuses.ts
```diff
@@ -141,6 +141,22 @@ export class AdminProgramStatuses {
...
+    await this.page.click(
+      this.programStatusItemSelector(statusName) + ' button:has-text("Edit")',
+    )
+
+    const modal = await waitForAnyModal(this.page)
+    const innerText = await modal.innerText()
+    expect(innerText).toContain('Edit this status')
+
+    // Close the modal prior to any assertions to avoid affecting
+    // subsequent tests.
```
shanemc-goog, line 154: wrap up

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -321,9 +333,8 @@ private Modal makeStatusUpdateModal(
       Http.Request request,
       ProgramDefinition program,
       Form<ProgramStatusesForm> form,
-      boolean displayOnLoad) {
-    // TODO(#2752): If an email is already configured, add a warning that setting it to empty
-    // will clear any other localized text.
+      boolean displayOnLoad,
+      Optional<DomContent> extraContent) {
```
shanemc-goog, line 337: `extraContent` This feels unnecessarily vague, If we make it `boolean showEmailDeletionWarning` it adds clarity and readability.

shanemc-goog (changes requested):

browser-test/src/support/admin_program_statuses.ts
clouser2: Done.

server/app/views/admin/programs/ProgramStatusesView.java
clouser2: Done. One thing to note is that adding a second boolean has a different tradeoff in that it increases the risk of mixing variable order up. That said, I think introducing an AutoValue for this is too heavyweight so I'm not going to pursue that approach.

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -226,12 +225,13 @@ private Pair<DivTag, ImmutableList<Modal>> renderStatusItem(
...
-            ? Optional.of(renderEmailTranslationWarning())
-            : Optional.empty();
     Modal editStatusModal =
-        makeStatusUpdateModal(request, program, statusEditForm, displayOnLoad, extraContent);
+        makeStatusUpdateModal(
+            request,
+            program,
+            statusEditForm,
+            displayOnLoad,
+            status.localizedEmailBodyText().isPresent());
```
shanemc-goog, line 234: prefix comment this too for readability as the evaluation doesn't map to the params utility.

shanemc-goog (changes requested):

server/app/views/admin/programs/ProgramStatusesView.java
shanemc-goog: ack. That's not unique here though and there are better ways to address it than what we had if it's of concern.   

If the method can be made static we could use AutoBuilder!!

shanemc-goog (approved):
LGTM, Just the one comment

server/app/views/admin/programs/ProgramStatusesView.java
clouser2: Done.

clouser2:
After talking with @rahaghas , we decided to update the language to:

Please be aware that clearing the email body will also clear any associated translations

I'll send out a follow-on PR for that shortly.

PR #3152, opened 2022-08-12
No longer trimming configuredStatusText since it corresponds directly to what is in the DB

clouser2:
### Description

It makes sense to trim the text from user-provided status texts and emails. However, the `configuredStatusText` field used to identify the original status when editing / deleting is essentially a key into what is in the DB directly. Additionally, it's populated based on the DB value. Performing a trim here will cause issues if a status makes it's way into a program that hasn't been trimmed.

## Release notes:

N/A

### Checklist

### Issue(s)

#2752


clouser2:
> Do we trim the original status value input?

Yes, here: https://github.com/civiform/civiform/blob/99e4c9f70f4a27aad9bf18b2f62c5358f7e25b3f/server/app/forms/admin/ProgramStatusesForm.java#L64

shanemc-goog (approved):

PR #3154, opened 2022-08-14
Added DisplayableMessage wrapper for toast messages, allowing the caller to specify the severity of the toast message

### LINKED ISSUE #1176: Some toasts that are "success" messages are rendered like "errors"
Currently our "error" toasts and "success" toasts look identical and they both look like errors (red). We should display happier looking toasts for success messages.

th174:
### Description

This change creates a standard interface for reporting handling error messages as toast messages by adding a wrapper class which contains both a message and a severity. It also uses Optional.empty() when no message is present, replacing the dubious practice of setting the message to empty string.

## Release notes:

### Checklist

### Issue(s)

Fixes #1176


server/app/controllers/DisplayableMessage.java
```diff
@@ -0,0 +1,39 @@
+package controllers;
+
+import views.components.ToastMessage.ToastType;
+
+public class DisplayableMessage {
```
clouser2, line 5: Please mark this class as `final` to make the decision to allow subclassing explicit rather than implicit.

server/app/controllers/DisplayableMessage.java
```diff
@@ -0,0 +1,39 @@
+package controllers;
+
+import views.components.ToastMessage.ToastType;
+
```
clouser2, line 4: Please add a javadoc string for the type as well as the nested enum.

server/app/controllers/DisplayableMessage.java
```diff
@@ -0,0 +1,39 @@
...
+  public DisplayableMessage(String message, Severity severity) {
+    this.message = message;
+    this.severity = severity;
+  }
+
+  public String getMessage() {
+    return message;
+  }
+
+  public Severity getSeverity() {
```
clouser2, line 19: Is there a strong reason to create a distinct enum type for this or could we use `ToastType` instead?

server/app/controllers/DisplayableMessage.java
```diff
@@ -0,0 +1,39 @@
+package controllers;
+
+import views.components.ToastMessage.ToastType;
```
clouser2, line 3: At a high level, can you help me understand why we wouldn't convert our APIs to pass `ToastMessage`s around everywhere?

I definitely agree that the transition from a string to a strong type is an improvement, I'm just trying to figure out what `DisplayableMessage` is adding on top of `ToastMessage`.

clouser2 (changes requested):
One open question on why we're adding a further wrapper to `ToastMessage` rather than using this type throughout the codebase.

server/app/controllers/DisplayableMessage.java
th174: I think ToastMessage is more of a UX component, and DisplayableMessage is more of a data type. MVC patterns typically try to separate the responsibility between data and its visual representation.

ToastMessage contains properties relating to UX representation like duration and `canDismiss`, while DisplayableMessage is focused on the data only.

Additionally, DisplayableMessage allows us the flexibility to display messages using some other UX component apart from a ToastMessage if we need to at some point in the future.

If you think this is overengineered, I totally understand, and I can change it to just use ToastMessage throughout

server/app/controllers/DisplayableMessage.java
th174: Thanks will do

server/app/controllers/DisplayableMessage.java
clouser2: That makes sense to me. In this case, each controller ends up translating downstream calls which return errors to a DisplayableMessage and passing that to a view which then converts that to a Toast. In my opinion, our controllers are already tightly bound to the view itself, so it might not add a lot of value.

My main concern is that we may end up slowly having to add the concept of dismissable and the other view-related logic to DisplayableMessage over time.

I understand your rationale and I'll defer to you. You may want to consult @shanemc-goog or @bion for a second opinion.

clouser2 (approved):

server/app/controllers/DisplayableMessage.java
th174: Requested approval from @bion to see what he thinks

server/app/controllers/DisplayableMessage.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** Model for user feedback messages that appear on the screen with a certain severity. */
```
bion, line 5: suggestion (blocking): "model" in the context of this code base refers to EBean models. I get that this is meant to be a data class representing a domain object, i.e. a model, but would prefer to avoid overloading the term in the code itself (including comments and docs). Maybe use "data class" instead of model?

server/app/controllers/DisplayableMessage.java
bion: `DisplayableMessage` is pretty tightly bound to the toast concept -- the Severity `enum` refers directly to the toast UI component with its `toastType` field.

IMO this is a bit premature and we should factor out a common data class if/when there is a clear pattern for reuse across multiple components. See the [LocalizedStrings](https://github.com/civiform/civiform/blob/main/server/app/services/LocalizedStrings.java#L20) class for an example.

For further context, there are already other types of errors that we want to return to the user that do not fit the data scheme of `DisplayableMessage`, such as inline form field validation messages.

bion (changes requested):

server/app/controllers/DisplayableMessage.java
th174: Makes sense. I'll just go with using the ToastMessage directly then.

server/app/controllers/admin/AdminProgramBlocksController.java
```diff
@@ -113,7 +115,7 @@ public Result edit(Request request, long programId, long blockId) {
     try {
       ProgramDefinition program = programService.getProgramDefinition(programId);
       BlockDefinition block = program.getBlockDefinition(blockId);
-      return renderEditViewWithMessage(request, program, block, "");
+      return renderEditViewWithMessage(request, program, block, Optional.empty());
```
bion, line 118: suggestion (blocking): please add a param name comment for the empty optional i.e. `/* message= */ Optional.empty()`

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -1,6 +1,7 @@
 package controllers.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static views.components.ToastMessage.ToastType.*;
```
bion, line 4: issue (blocking): [no wildcard imports ](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)

server/app/views/components/ToastMessage.java
```diff
@@ -32,20 +32,26 @@ enum ToastType {
   /** If true this message will not be shown if a user has already seen and dismissed it. */
   private boolean canIgnore = false;
 
+  public ToastMessage(String message, ToastType severity) {
+    this.message = message;
+    this.type = severity;
```
bion, line 37: suggestion (blocking): add `checkNotNull`

bion (changes requested):

server/app/views/components/ToastMessage.java
th174: Done.

server/app/controllers/applicant/ApplicantProgramReviewController.java
th174: Done.

server/app/controllers/admin/AdminProgramBlocksController.java
th174: Done.

bion (approved):

PR #3170, opened 2022-08-15
add test for aws oidc template

azizova-leyla:
### Description

Copies config from https://github.com/civiform/staging-aws-deploy/blob/main/civiform_config.sh and runs setup and deploy in test mode.

## Release notes:

### Checklist

### Issue(s)

### DIFF
--- a/.github/workflows/aws-oidc-deploy-test.yaml
+++ b/.github/workflows/aws-oidc-deploy-test.yaml
@@ -30,18 +30,18 @@ jobs:
           #  the changes work with actual data
           curl https://raw.githubusercontent.com/civiform/staging-aws-deploy/main/civiform_config.sh > civiform_config.sh
 
-      - name: Source config, set CIVIFORM_MODE to be "test"
+      - name: Source config, set CIVIFORM_MODE to be "test", Run setup with latest tag
         run: |
           source civiform_config.sh
           export CIVIFORM_MODE="test"
           echo "${LATEST_SNAPSHOT_TAG}"
-
-      - name: Run setup with latest tag
-        run: |
           cloud/shared/bin/setup  "--tag=${LATEST_SNAPSHOT_TAG}"
           exit $?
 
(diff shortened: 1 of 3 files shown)


.github/workflows/aws-oidc-deploy-test.yaml
```diff
@@ -0,0 +1,50 @@
...
+          #  the changes work with actual data
+          curl https://raw.githubusercontent.com/civiform/staging-aws-deploy/main/civiform_config.sh > civiform_config.sh
+
+      - name: Source config, set CIVIFORM_MODE to be "test", Run setup with latest tag
+        run: |
+          source civiform_config.sh
+          export CIVIFORM_MODE="test"
+          export APP_PREFIX="$(git rev-parse --short HEAD)"
+          echo "${LATEST_SNAPSHOT_TAG}"
+          cloud/shared/bin/setup  "--tag=${LATEST_SNAPSHOT_TAG}"
```
nbeloglazov, line 41: Running setup sources `civiform_config.sh`. So it will override CIVIFORM_MODE that you set previously?

azizova-leyla: it runs https://github.com/civiform/civiform/blob/main/cloud/shared/bin/setup that does not source civiform_config.sh that's why it's sourced here explicitly.

nbeloglazov: It runs lib.sh which in turn sources the config: https://github.com/civiform/civiform/blob/main/cloud/shared/bin/lib.sh#L16 
Though it sources config from the template folder and there is no config there AFAIU as you created it in the root directory. So that source will actually fail? Can we test this action somehow (maybe it already ran) or we need to submit and see?

nbeloglazov: Oh, sorry, I misread the code.

nbeloglazov (approved):

PR #3171, opened 2022-08-15
Logout users from the central provider (with flag)

### LINKED ISSUE #1770: Call to IDCS logout endpoint
**Describe the bug**
Selecting "logout" does not trigger an endpoint logout in IDCS. For example, if I log in to CiviForm and then click "logout", I am redirected back to the CiviForm login homepage. But if I then click "login", I am not asked for my username or password. It automatically takes me to my account. 

The IDCS team recommends calling the IDCS logout endpoint to make sure we have a clean logout ending application session and IDCS session as well.

**To Reproduce**
1. Go to https://civiform.seattle.gov
2. Click on 'Login"
3. Enter Username and Password
4. Selected 'Logout'
5. Return to home login page
6. Selected 'Login'
7. See that CiviForm does not ask for username or password; user is still logged in.

**Expected behavior**
Clicking logout would force a user to enter their username and password when trying to log back in.

kmcnellis:
### Description

Allows users to be logged out of the central OIDC auth provider (as opposed to just the local civiform session), so that when the "logout" button is pressed, then "login" is pressed, the user is re-prompted to login.  

When "logout" is pressed, it redirects to the OIDC provider, which then redirects back to the base_url

## Release notes:

Allows users to be logged out of the central OIDC auth provider, so that when the "logout" button is pressed, then "login" is pressed, the user is re-prompted to login.  

Configured with the following variables: 

```
export APPLICANT_OIDC_PROVIDER_LOGOUT=true  # Enable central logout
export APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM="returnTo" # Customize the redirect query parameter in the logout request
export APPLICANT_OIDC_LOGOUT_CLIENT_PARAM="client_id" # Add the client id & customize it's query parameter in the logout request
```

### Checklist

### Issue(s)

Fixes #1770


server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+      this.extraParams = Map.of();
+    }
+
+    this.baseUrl =
+        getConfigurationValue("base_url")
+            .orElseThrow(() -> new RuntimeException("base_url must be set"));
+  }
+
+  /** Helper function for retriving values from the application.conf, */
+  protected final Optional<String> getConfigurationValue(String name) {
```
clouser2, line 52: Can this be marked static and avoid the need to store the complete configuration in the "civiformConfiguration" instance variable?

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+    if (civiformConfiguration.hasPath(name)) {
+      return Optional.ofNullable(civiformConfiguration.getString(name));
+    }
+    return Optional.empty();
+  }
+
+  // Override, since parent checks that the profile is an instance of OidcProfile.
+  // Does not supply idtoken hint.
```
clouser2, line 60: Would you mind expanding these comments a bit more, specifically around:
  * Why is it bad or incorrect to assert the profile is an instance of a specific class?
  *  What does not supplying an idtoken hint mean?

Is the current implementation essentially the base classe's implementation copied and modified or is different in other ways? I mainly ask to determine how we would prevent this from diverging as the superclasses implementation changes over time.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public Optional<RedirectionAction> getLogoutAction(
+      final WebContext context,
+      final SessionStore sessionStore,
+      final UserProfile currentProfile,
+      final String targetUrl) {
+    final String logoutUrl = configuration.findLogoutUrl();
+    logger.debug("Logging user out: " + currentProfile.toString());
```
clouser2, line 68: Is this logging statement necessary? I mainly ask because I don't know what UserProfile.toString() returns and whether we should think of it as safe to put in logs. For example, if it had some sort of secret value, it'd be unfortunate for it to make its way to logs.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+    if (CommonHelper.isNotBlank(logoutUrl) && currentProfile instanceof CiviFormProfileData) {
+      try {
+        final URI endSessionEndpoint = new URI(logoutUrl);
+        LogoutRequest logoutRequest =
+            new CustomOidcLogoutRequest(
+                endSessionEndpoint,
+                this.postLogoutRedirectParam.orElse(null),
```
clouser2, line 75: Not blocking: Typically, I believe we tend to not use "this" as a qualifier unless it's necessary (or in a constructor). Same for "extraParams" below.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+                new URI(targetUrl),
+                this.extraParams);
+        logger.debug("logoutRequest: " + logoutRequest.toURI().toString());
+
+        return Optional.of(
+            HttpActionHelper.buildRedirectUrlAction(context, logoutRequest.toURI().toString()));
+      } catch (final URISyntaxException e) {
+        throw new TechnicalException(e);
```
clouser2, line 83: Not blocking: Can we build the endSessionEndpoint / targetUrl `URI`s in a more scoped try/catch?

Perhaps also add a brief comment as to why we wrap it in a TechnicalException rather than letting it be thrown.

server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.google.common.base.Strings;
+import com.nimbusds.openid.connect.sdk.LogoutRequest;
+import java.net.URI;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class CustomOidcLogoutRequest extends LogoutRequest {
```
clouser2, line 11: Not blocking: Consider marking this "final" to make the decision to subclass explicit rather than implicit.

server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -0,0 +1,87 @@
...
+  /** The optional post-logout redirection param and URI. */
```
clouser2, line 12: I believe this comment is meant to apply to both of the below variables but it only got attached to the topmost one. Likely an artifact of bin/fmt. Can you add a separate comment for the URI below?

server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -0,0 +1,87 @@
...
+  private final String postLogoutRedirectParam;
+
+  private final URI postLogoutRedirectURI;
+
+  private final Map<String, String> extraParams;
```
clouser2, line 17: Can we use ImmutableMap here to provide a further guarantee that the key/values in the map itself are unmodifiable?

server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -0,0 +1,87 @@
...
+  public CustomOidcLogoutRequest(
+      final URI uri,
+      final String postLogoutRedirectParam,
+      final URI postLogoutRedirectURI,
+      final Map<String, String> extraParams) {
+
+    super(uri, /* idTokenHint */ null, postLogoutRedirectURI, /* state */ null);
+
+    if (Strings.isNullOrEmpty(postLogoutRedirectParam)) {
+      // default to OIDC spec
```
clouser2, line 32: Not blocking: Consider linking to the entry in the spec to make it easy for one to quickly find this information again.

server/app/auth/oidc/OidcProvider.java
```diff
@@ -179,7 +179,9 @@ public OidcClient get() {
     Optional<String> providerName = getProviderName();
     OidcConfiguration config = new OidcConfiguration();
 
+    config.setLogoutUrl("https://civiform-staging.us.auth0.com/v2/logout");
```
clouser2, line 182: This seems like something that was missed. If I understand correctly, we don't want to reference specific URLs from server implementation.

server/app/auth/oidc/OidcProvider.java
```diff
@@ -197,17 +199,19 @@ public OidcClient get() {
...
+        new CiviformOidcLogoutActionBuilder(civiformConfig, config, clientID));
+
     try {
       client.init();
     } catch (Exception e) {
       logger.error("Error while initilizing OIDC provider", e);
       throw e;
     }
-
+    logger.debug("Client: " + client.toString());
```
clouser2, line 214: Same question / concern as above. Should we be logging this information and are confident that it won't contain any sensitive data?

clouser2 (changes requested):
Looks pretty reasonable. Most of my comments were around clarifying documentation and getting a better understanding of why certain params / classes are needed.

cloud/aws/templates/aws_oidc/variable_definitions.json
```diff
@@ -67,7 +67,7 @@
     "secret": false,
     "tfvar": true,
     "type": "string",
-    "value_regex": "[^\\s]+",
+    "value_regex": "[^\\s]*",
```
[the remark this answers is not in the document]
kmcnellis: If users provide an empty string (but define the variable), the old regex failed.  But the code correctly handles that condition (with a default variable)/

cloud/aws/templates/aws_oidc/variables.tf
```diff
@@ -213,6 +213,24 @@ variable "civiform_applicant_idp" {
...
 
+variable "applicant_oidc_provider_logout" {
+  type        = bool
+  description = "If the applicant OIDC logout should also perform a central logout from the auth provider"
+  default     = false
+}
+
+variable "applicant_oidc_post_logout_redirect_param" {
+  type        = string
+  description = "What query parameter to use for sending the redirect uri to the central OIDC provider for logout. Defaults to post_logout_redirect_uri"
```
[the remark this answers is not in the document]
kmcnellis: done

cloud/aws/templates/aws_oidc/variables.tf
```diff
@@ -213,6 +213,24 @@ variable "civiform_applicant_idp" {
   default     = ""
 }
 
+variable "applicant_oidc_provider_logout" {
+  type        = bool
+  description = "If the applicant OIDC logout should also perform a central logout from the auth provider"
+  default     = false
```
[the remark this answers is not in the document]
kmcnellis: I like the idea of forcing it.  Removed the default

cloud/azure/modules/app/variables.tf
```diff
@@ -186,6 +186,25 @@ variable "civiform_applicant_idp" {
   default     = "login-radius"
 }
 
+
+variable "applicant_oidc_provider_logout" {
```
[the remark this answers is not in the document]
kmcnellis: done

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -143,6 +143,9 @@ module "td" {
     STAGING_APPLICANT_LIST               = var.staging_applicant_notification_mailing_list
     APPLICANT_OIDC_PROVIDER_NAME         = var.applicant_oidc_provider_name
     CIVIFORM_APPLICANT_IDP               = var.civiform_applicant_idp
+    APPLICANT_OIDC_PROVIDER_LOGOUT       = var.applicant_oidc_provider_logout
+    APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM = var.applicant_oidc_post_logout_redirect_param
+    APPLICANT_OIDC_LOGOUT_CLIENT_PARAM   = var.applicant_oidc_logout_client_param
```
bion, line 148: suggestion (blocking): i think terraform fmt will realign these

cloud/azure/modules/app/locals.tf
```diff
@@ -56,6 +56,9 @@ locals {
     ADFS_DISCOVERY_URI       = data.azurerm_key_vault_secret.adfs_discovery_uri.value
     ADFS_GLOBAL_ADMIN_GROUP  = var.adfs_admin_group
     CIVIFORM_APPLICANT_IDP   = var.civiform_applicant_idp
+    APPLICANT_OIDC_PROVIDER_LOGOUT            = var.applicant_oidc_provider_logout
+    APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM = var.applicant_oidc_post_logout_redirect_param
+    APPLICANT_OIDC_LOGOUT_CLIENT_PARAM        = var.applicant_oidc_logout_client_param
```
bion, line 61: suggestion (blocking): i think terraform fmt will realign these

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+import org.pac4j.oidc.logout.OidcLogoutActionBuilder;
+
+/**
+ * Custom OidcLogoutActionBuilder for CiviFormProfileData (since it extends CommonProfile, not
+ * OidcProfile). Also allows for divergence from the [oidc
+ * spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) if your provider requires it
+ * (e.g. Auth0).
+ *
+ * <p>Does not provide the recommended id_token_hint, since these are not stored by the civiform
+ * profile.
```
bion, line 29: question (non-blocking): should we store this?

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>If the oidc_logout_client_id_param config variable is set, also adds the client_id to the
+ * logout request.
+ */
+public final class CiviformOidcLogoutActionBuilder extends OidcLogoutActionBuilder {
+
+  private final Optional<String> postLogoutRedirectParam;
+  private final ImmutableMap<String, String> extraParams;
+
+  public CiviformOidcLogoutActionBuilder(
+      Config civiformConfiguration, final OidcConfiguration oidcConfiguration, String clientID) {
```
bion, line 43: nit: `final` probably isn't necessary?

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+  /** Helper function for retriving values from the application.conf, */
+  static final Optional<String> getConfigurationValue(Config civiformConfiguration, String name) {
+    if (civiformConfiguration.hasPath(name)) {
+      return Optional.ofNullable(civiformConfiguration.getString(name));
+    }
+    return Optional.empty();
+  }
+
+  /**
+   * Override the parent's getLogoutAction, since it check that the profile is an instance of
```
bion, line 67:
```suggestion
   * Override the parent's getLogoutAction, since it checks that the profile is an instance of
```

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * <p>Also use the custom CustomOidcLogoutRequest to build the URL.
+   */
+  @Override
+  public Optional<RedirectionAction> getLogoutAction(
+      final WebContext context,
+      final SessionStore sessionStore,
+      final UserProfile currentProfile,
+      final String targetUrl) {
+    final String logoutUrl = configuration.findLogoutUrl();
```
bion, line 78: nit: generally don't need to use `final` for variables and params unless the compiler demands it

server/app/auth/oidc/OidcProvider.java
```diff
@@ -197,19 +199,19 @@ public OidcClient get() {
...
+
     try {
       client.init();
     } catch (Exception e) {
       logger.error("Error while initilizing OIDC provider", e);
       throw e;
     }
-
     var providerMetadata = client.getConfiguration().getProviderMetadata();
-    logger.debug("Provider metadata: " + providerMetadata.toString());
```
bion, line 212: why remove this?

server/app/modules/SecurityModule.java
```diff
@@ -77,8 +77,12 @@ protected void configure() {
 
     // you can logout by hitting the logout endpoint, you'll be redirected to root page.
     LogoutController logoutController = new LogoutController();
-    logoutController.setDefaultUrl(routes.HomeController.index().url());
+    logoutController.setDefaultUrl(baseUrl + routes.HomeController.index().url());
+    logoutController.setLocalLogout(true);
     logoutController.setDestroySession(true);
+
+    Boolean logoutProvider = configuration.getBoolean("auth.oidc_provider_logout");
+    logoutController.setCentralLogout(logoutProvider);
```
bion, line 85: nit: the naming of this variable is confusing since it's a bool but the name calls it a provider?

server/conf/logback.xml
```diff
@@ -31,7 +31,8 @@
   <logger name="s3client" level="INFO" />
   <logger name="io.ebean" level="INFO" />
   <logger name="org.pac4j" level="INFO" />
-  <logger name="auth" level="INFO" />
+  <logger name="auth" level="DEBUG" />
+  <logger name="org.pac4j" level="DEBUG" />
```
bion, line 35: suggestion (blocking): this will probably be too much auth-related logs to be a good default in dev

bion (changes requested):
Is it feasible to build an image and test this on Seattle staging to ensure it works for both IDCS and ADFS before merging? cc @gwendolyngoetz

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: I don't think it's worth storing this until one the civic entities' identity providers requires it in order to logout correctly.

cloud/azure/modules/app/locals.tf
kmcnellis: done

cloud/aws/templates/aws_oidc/variables.tf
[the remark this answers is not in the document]
kmcnellis: Unfortunately our tests for terraform validation don't currently allow variables without defaults

bion:
@kmcnellis Is it feasible to build an image and test this on Seattle staging to ensure it works for both IDCS and ADFS before merging? cc @gwendolyngoetz

cloud/aws/templates/aws_oidc/variable_definitions.json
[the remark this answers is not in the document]
clouser2: Ah, I see. So to clarify, the downstream code treats a present key with an empty value the same as if the key wasn't present?

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
clouser2: Sounds reasonable.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * <p>Uses the post_logout_redirect_uri parameter by default, but allows overriding to a different
+ * value using the auth.oidc_post_logout_param config variable
+ *
+ * <p>If the oidc_logout_client_id_param config variable is set, also adds the client_id to the
+ * logout request.
+ */
+public final class CiviformOidcLogoutActionBuilder extends OidcLogoutActionBuilder {
+
+  private Optional<String> postLogoutRedirectParam;
```
clouser2, line 39: Not blocking: Both of these variables can be marked final, I believe.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+
+    if (clientIdParam.isPresent()) {
+      this.extraParams = ImmutableMap.of(clientIdParam.get(), clientID);
+    } else {
+      this.extraParams = ImmutableMap.of();
+    }
+  }
+
+  /** Helper function for retriving values from the application.conf, */
+  static final Optional<String> getConfigurationValue(Config civiformConfiguration, String name) {
```
clouser2, line 59: Not blocking: Can this be marked as `private`? Also, since the class is marked `final`, there's no need to add the specifier here.

server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -0,0 +1,103 @@
...
+   * Create new OIDC logout request with a optional redirect url, optional client id, and other
+   * params.
+   */
+  public CustomOidcLogoutRequest(
+      final URI uri,
+      final String postLogoutRedirectParam,
+      final URI postLogoutRedirectURI,
+      final ImmutableMap<String, String> extraParams) {
+
+    super(uri, /* idTokenHint */ null, postLogoutRedirectURI, /* state */ null);
```
clouser2, line 45: Does this work? The base class constructor appears to throw an IllegalArgumentException when the provided hint is null.

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  @Override
+  public Optional<RedirectionAction> getLogoutAction(
+      WebContext context, SessionStore sessionStore, UserProfile currentProfile, String targetUrl) {
+    String logoutUrl = configuration.findLogoutUrl();
+    if (CommonHelper.isNotBlank(logoutUrl) && currentProfile instanceof CiviFormProfileData) {
+      try {
+        URI endSessionEndpoint = new URI(logoutUrl);
+        LogoutRequest logoutRequest =
+            new CustomOidcLogoutRequest(
```
clouser2, line 80: It appears that the base implementation forks based on whether target URL is blank or not (https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/logout/OidcLogoutActionBuilder.java).

Generally, I'm a little hesitant to fork code, especially in a library specifically for security. If some sort of security patch were added in the base class, we wouldn't pick it up nor would we have any way of knowing we needed to.

If we could support ID tokens in the CiviformProfile, is there any way we'd be able to use the base implementation and not require custom subclassing? As I understand, it might not be possible since we wouldn't have a way of passing additional query parameters in the callback.

If we stay with this implementation, I'd prefer we open an issue against pac4j / etc to support this specific scenario since it avoids the need to code fork.

server/app/auth/oidc/OidcProvider.java
clouser2: Blocking: This still appears to be present.

clouser2 (changes requested):

server/app/auth/oidc/CustomOidcLogoutRequest.java
kmcnellis: I'm not sure where you see that?  I see this in the docs:
```
idTokenHint - The ID token hint (recommended), null if not specified.
```
https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/5.25/com/nimbusds/openid/connect/sdk/LogoutRequest.html

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: The basic issue here is that Auth0 (and maybe other providers) don't comply with the Oauth spec for logout.  It's not really pac4j's fault. 

However, the big problem (and bigger refactor) is that the parent uses the line `if (CommonHelper.isNotBlank(logoutUrl) && currentProfile instanceof OidcProfile)`.  We don't inherit from OidcProfile, so it's always false.

The fork you mention is not really a fork in logic.  It uses a shortcut constructor that calls the main one, but just inserts 'null' so you don't have to.

kmcnellis: (in order to fix the oidc profile thing, we'd need to remove all our saml code, or branch the logic somehow)

cloud/aws/templates/aws_oidc/variable_definitions.json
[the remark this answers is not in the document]
kmcnellis: Correct

cloud/deploys/dev_aws/civiform_config.example.sh
```diff
@@ -0,0 +1,52 @@
+# More details and documentation for these are consolidated in the deploy repo:
```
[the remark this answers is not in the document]
kmcnellis: removed

server/app/auth/oidc/OidcProvider.java
kmcnellis: It's fixed now

cloud/aws/templates/aws_oidc/app.tf
kmcnellis: done

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: done

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: removed

server/conf/logback.xml
kmcnellis: removed

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: removed

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
clouser2: I definitely agree that the logic now is consistent with what we want. My worry is that the underlying Pac4J library will drift over time and we'll silently fall out of sync with the implementation.

I definitely agree that the non-standards-compliant implementations complicate this. I don't think we need to block this PR over it necessarily, but I do think we might want to have a larger discussion about if there are other hooks in Pac4J that we could use to not have to fork the code. Or perhaps we can file a bug to support these scenarios, since I imagine the Pac4J maintainers likely would prefer having customization hooks to potentially forking the implementation.

Again, this concern needn't be considered blocking in any way for the PR, I'm just trying to voice my concern of drift over time.

server/app/auth/oidc/CustomOidcLogoutRequest.java
clouser2: Here's where I found it: https://github.com/hidglobal/oauth-2.0-sdk-with-openid-connect-extensions/blob/master/src/main/java/com/nimbusds/openid/connect/sdk/LogoutRequest.java#L90

I think I might have been looking at the wrong implementation.

clouser2 (approved):

server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
kmcnellis: ➕

azizova-leyla:
Note (no action from you required): cloud changes need to be migrated to the new repo.

PR #3175, opened 2022-08-16
add civiform prefix to all resource names

azizova-leyla:
### Description

It will be easier to find resources in AWS console.

## Release notes:

### Checklist

### Issue(s)


nbeloglazov (changes requested):
You need to update resources.py as well for variables that come from the app.tf: 
https://github.com/civiform/civiform/blob/main/cloud/aws/templates/aws_oidc/bin/resources.py

Also there are much more resources. Do you really want to update all of them? Like secrets, database, vpc, etc. Maybe we can require that "civiform" be a part of "app_prefix" itself? Currently we default it to "my-app".

azizova-leyla:
Added civiform to all resources (hopefully). I would prefer to code it in rather then requiring in the app-prefix.

nbeloglazov (approved):

PR #3177, opened 2022-08-16
Removing hover state from application status row.

clouser2:
### Description

Per feedback from @rahaghas, there's no reason to change the background color when hovering over an application status row since clicking it wouldn't do anything.

## Release notes:

N/A

### Checklist

### Issue(s)

#2752


rahaghas (approved):

shanemc-goog:
So I think given it's a "Feature" there should be release notes, but this is also making a change to an unlaunched feature, so it's not particularly notable I think.

@bion what kind of release note would you want to see here?

shanemc-goog (approved):

PR #3186, opened 2022-08-17
Add ability to move question up and down within its block

### LINKED ISSUE #1928: Allow admins to reorder questions more easily
Idea: Drag and drop to reorder questions on the program creation page instead of having to delete and readd (current state)

Relevant slide from UX audit: https://docs.google.com/presentation/d/1goxiepPXPrW3wK8wUSNoCpcf7iXIohXfj96iTYiQxo0/edit#slide=id.gf083eb9b36_0_1337

nbeloglazov:
### Description

Adds up and down arrow buttons when editing questions in block. Buttons are hidden for first and last questions. Server-side implementation takes a new position and not just "up/down" enum as we might implement drag-and-drop for questions to reorder and that way we won't need to update server-side.

## Release notes:

Add ability to move question up and down within its block.

### Checklist

### Issue(s)

Fixes #1928


browser-test/src/admin_program_creation.test.ts
```diff
@@ -76,4 +78,64 @@ describe('Create program with enumerator and repeated questions', () => {
 
     await endSession(browser)
   })
+
+  fit('change questions order within block', async () => {
```
bion, line 82:
```suggestion
  it('change questions order within block', async () => {
```

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
```diff
@@ -144,4 +145,33 @@ public Result setOptional(
     return redirect(
         controllers.admin.routes.AdminProgramBlocksController.edit(programId, blockDefinitionId));
   }
+
+  /** POST endpoint for changing position of a question in its block. */
+  @Secure(authorizers = Labels.CIVIFORM_ADMIN)
+  public Result move(
+      Request request, long programId, long blockDefinitionId, long questionDefinitionId)
+      throws InvalidQuestionPositionException {
```
bion, line 153: suggestion (blocking): if I understand correctly, this error could be thrown as a result of bad user input. If so, handle the error here and return `badRequest()` instead of throwing and causing a 500

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -708,6 +710,50 @@ public ProgramDefinition setProgramQuestionDefinitionOptionality(
...
+    questions.add(newPosition, question);
+
+    try {
+      return updateProgramDefinitionWithBlockDefinition(
+          programDefinition,
+          blockDefinition.toBuilder()
+              .setProgramQuestionDefinitions(ImmutableList.copyOf(questions))
+              .build());
+    } catch (IllegalPredicateOrderingException e) {
+      // Changing a question position within block should not affect predicates.
```
bion, line 751: nit: maybe add "Changing a question position within block should not affect predicates because predicates cannot depend on questions within the same block."

server/conf/routes
```diff
@@ -53,11 +53,11 @@ POST    /admin/programs/:programId/blocks/:blockDefinitionId/delete    controlle
 GET     /admin/programs/:programId/blocks/:blockDefinitionId/edit/predicates/edit    controllers.admin.AdminProgramBlockPredicatesController.edit(request: Request, programId: Long, blockDefinitionId: Long)
 POST    /admin/programs/:programId/blocks/:blockDefinitionId/edit/predicates         controllers.admin.AdminProgramBlockPredicatesController.update(request: Request, programId: Long, blockDefinitionId: Long)
 POST    /admin/programs/:programId/blocks/:blockDefinitionId/edit/predicates/delete  controllers.admin.AdminProgramBlockPredicatesController.destroy(programId: Long, blockDefinitionId: Long)
-
```
bion, line 56: nit: please keep this newline for separating routes to different controllers

bion (changes requested):
great work! just a few small comments

browser-test/src/admin_program_creation.test.ts
nbeloglazov: Thanks for catching it. We should enable eslint check for that.

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
nbeloglazov: Done.

server/app/services/program/ProgramServiceImpl.java
nbeloglazov: Done.

server/conf/routes
nbeloglazov: Done.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -708,6 +710,51 @@ public ProgramDefinition setProgramQuestionDefinitionOptionality(
...
+      if (question.id() == questionDefinitionId) {
+        it.remove();
+        break;
+      }
+    }
+    if (question == null) {
+      throw new ProgramQuestionDefinitionNotFoundException(
+          programId, blockDefinitionId, questionDefinitionId);
+    }
+    questions.add(newPosition, question);
```
clouser2, line 742: Would the following be equivalent (though technically less performant due to the double iteration):
```
Optional<Question> toMove = questions.stream()
  .filter(q -> q.id() == questionDefinitionId)
  .findFirst();
List<Question> otherQuestions = questions.stream()
  .filter(q -> q.id() != questionDefinitionId)
  .collect(Collectors.toList());
otherQuestions.add(newPosition, question);

```

server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -391,7 +397,9 @@ private DivTag renderQuestion(
       long blockDefinitionId,
       QuestionDefinition questionDefinition,
       boolean canRemove,
-      boolean isOptional) {
+      boolean isOptional,
+      int questionIndex,
+      int questionsCount) {
```
clouser2, line 402: Not blocking: This is pre-existing, but seeing two booleans followed by two ints makes it seem like it'd be really easy to mix up the argument order in this function :(.

clouser2 (commented):
I like it!

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
```diff
@@ -144,4 +145,34 @@ public Result setOptional(
...
   }
+
+  /** POST endpoint for changing position of a question in its block. */
+  @Secure(authorizers = Labels.CIVIFORM_ADMIN)
+  public Result move(
+      Request request, long programId, long blockDefinitionId, long questionDefinitionId) {
+    requestChecker.throwIfProgramNotDraft(programId);
+
+    DynamicForm requestData = formFactory.form().bindFromRequest(request);
+    int newPosition = Integer.parseInt(requestData.get("position"));
```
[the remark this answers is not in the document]
nbeloglazov: Is sharing constants through views standard practice? I do want to share constant, just not sure what is the best place for it.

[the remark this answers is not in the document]
nbeloglazov: Done. Passing null to parseInt throws NumberFormatException as well. Checked it.

server/app/services/program/InvalidQuestionPositionException.java
```diff
@@ -0,0 +1,10 @@
+package services.program;
+
+public class InvalidQuestionPositionException extends Exception {
```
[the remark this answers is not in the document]
nbeloglazov: Done.

server/app/views/admin/programs/ProgramBlockEditView.java
nbeloglazov: Is there a way to address it? Create dedicated value object?

server/app/services/program/ProgramServiceImpl.java
nbeloglazov: Sure. I don't think performance here is critical given that number of questions should be low.

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
```diff
@@ -144,4 +145,34 @@ public Result setOptional(
...
+
+    try {
+      programService.setProgramQuestionDefinitionPosition(
+          programId, blockDefinitionId, questionDefinitionId, newPosition);
+    } catch (ProgramNotFoundException e) {
```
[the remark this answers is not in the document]
nbeloglazov: Done.

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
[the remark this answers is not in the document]
clouser2: Unfortunately sharing constants isn't widely done throughout the codebase, but I think we should try and move towards this model. IMO, the controller already has a dependency on the view, so adding the constant there doesn't introduce any new dependencies.

Another way this can be achieved is by creating a distinct form object with getters/setters and putting the constants there. In this case though, it'd be a single-field form, so likely not really worth adding.

server/app/views/admin/programs/ProgramBlockEditView.java
clouser2: Yah, the only way I can think of is a dedicated value object. Probably not worth it here. In a perfect world, I'd love something similar to named arguments or destructuring. Definitely not worth blocking this PR over though :D

clouser2 (approved):

bion (approved):

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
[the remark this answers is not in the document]
nbeloglazov: Done. Added constant to view. IN this specific that controller doesn't depend on that view. But I think adding this dependency isn't bad as generally they are related and controllers can depend on view.

PR #3187, opened 2022-08-17
Fix browsertest M1 builds and optimize node modules

kmcnellis:
### Description
- Removes the explicit use of tini  (replaces it with the docker --init flag)  in order to fix the m1 qemu image error: 
`qemu-x86_64: Could not open ‘/lib64/ld-linux-x86-64.so.2’: No such file or directory`
- Moves the node install to the image build and save it to an anonymous volume to prevent overwriting/contamination with local file system.
- Deletes package-lock, since this uses yarn to install packages

## Release notes:

n/a

### Checklist

### Issue(s)

Fixes browsertest for m1 macs


azizova-leyla (approved):

browser-test/bin/wait_for_server_start_and_run_tests.sh
```diff
@@ -11,6 +11,10 @@ SERVER_URL="http://civiform:9000"
 # https://playwright.dev/docs/browsers#managing-browser-binaries-1
 PLAYWRIGHT_BROWSERS_PATH=0
```
nbeloglazov, line 12: That's no longer needed if you set it inside image? And this script is run only from within container.

nbeloglazov (approved):

browser-test/bin/wait_for_server_start_and_run_tests.sh
kmcnellis: removed

PR #3192, opened 2022-08-17
add CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET to terraform

azizova-leyla:
### Description

Needs to be set to false in staging to make our browser tests pass.

## Release notes:

### Checklist

### Issue(s)

### DIFF
--- a/cloud/shared/variable_definitions.json
+++ b/cloud/shared/variable_definitions.json
@@ -115,5 +115,11 @@
     "secret": false,
     "tfvar": true,
     "type": "bool"
+  },
+  "CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET": {
+    "required": false,
+    "secret": false,
+    "tfvar": true,
+    "type": "bool"
   }
 }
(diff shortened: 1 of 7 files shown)


cloud/shared/variable_definitions.json
```diff
@@ -115,5 +115,11 @@
     "secret": false,
     "tfvar": true,
     "type": "bool"
+  },
+  "CIVIFORM_API_KEYS_BAN_GLOBAL_SUBNET": {
```
nbeloglazov, line 119: Did you try deploying without providing this variable in config? What value is passed to fargate tasks, 'true'? Just want to make sure that empty string doesn't override the default true value of the terraform variable.

nbeloglazov (approved):

cloud/shared/variable_definitions.json
azizova-leyla: verified that it's defaulting to true when not set

PR #3193, opened 2022-08-17
Displaying default text in the translation UI alongside the text rather than below it

clouser2:
### Description

Adjusting "default text" shown in the translation UI per feedback:
  * Default text -> English text
  * Rendering the English text to the right of the content to be translated rather than below.

### Screenshots
Program translations:
Screen Shot 2022-08-17 at 3 43 51 PM

Question translations:
Screen Shot 2022-08-17 at 3 43 39 PM

## Release notes:

Adjusts the translation UI to render the English version of the text side-by-side with the inputs rather than below.

### Checklist

### Issue(s)

#2754


shanemc-goog:
How about making the input block as tall as the default content as it likely will be around that tall.  Otherwise we're going to hide a lot of the inputed content for emails

server/app/views/admin/TranslationFormView.java
```diff
@@ -103,10 +106,29 @@ protected final FormTag renderTranslationForm(
...
+                .withClasses(Styles.PX_2, Styles.PY_1, Styles.TEXT_SM, Styles.BG_GRAY_100)
+                .with(
+                    p("English text:").withClass(Styles.FONT_MEDIUM),
+                    p().with(defaultLocaleTextContent(localizedStrings))));
+  }
+
+  private final ImmutableList<DomContent> defaultLocaleTextContent(
+      LocalizedStrings localizedStrings) {
+    // Convert newlines to break statements.
+    ImmutableList<DomContent> content =
```
shanemc-goog, line 125: The `pre` element will handle this for us right without modifying their content?

What do we do for the static question type and other block inputs?

shanemc-goog (changes requested):

clouser2:
> How about making the input block as tall as the default content as it likely will be around that tall. Otherwise we're going to hide a lot of the inputed content for emails

@rahaghas, thoughts on this one? One note is that we use a textarea which can be adjusted via the user to increase the size. I could see a world where we get the number of lines in the english text and set the text area rows to that value. If we decide to go this route, I'll leave a TODO and address independently.

server/app/views/admin/TranslationFormView.java
clouser2: Good call on using `<pre>`. For static questions, we actually run it through our own pre-processor that creates accordions, etc. I don't think we'll want to do the same here given that we're only sending plain-text email. `<pre>` is a great way to preserve the original content and I switched to that!

shanemc-goog (approved):

rahaghas:
Agree with Shane - Input blocks should be taller- perhaps we can define two types in input blocks in general so we have a rule around it, one for short text and one for tall. I'm ok with defining one height for tall and allowing users to scroll within that if the content is longer

clouser2:
> Agree with Shane - Input blocks should be taller- perhaps we can define two types in input blocks in general so we have a rule around it, one for short text and one for tall. I'm ok with defining one height for tall and allowing users to scroll within that if the content is longer

Sounds good. To be clear, I think we'd want to apply this to `textarea` inputs only. How many rows would you like to appear for "short" vs "tall"?

rahaghas:
Agreed about applying to textarea only. Maybe 10?

PR #3198, opened 2022-08-18
Enforce consistent sizing in icons svg

nbeloglazov:
### Description

SVG stored in Icons.java all developed for certain sizes (20, 24, 48). This size has been lost when adding SVG to Icons enum. Currently callers of icons have to know that size and pass it explicitly, which is error-prone and unnecessary. Instead callers should not care about original SVG size, they should care only about what size they want to render in their specific case on a page. They should treat `SvgTag` returned by Icons.svg() as any other DOM element that needs to be styled using classes. Example recommended usage:

```
Icons.svg(Icons.ICON_NAME).withClasses(Styles.H_4, Styles.W_4)
```

This PR includes several changes:
1. Remove all svg() methods that take separate width and height arguments. At the moment all icons are square and that's very likely going to be that way going forward. Having 2x methods is harder to read. 
2. Add original size to each enum value. If size passed by the user doesn't match the original size - the error is thrown. That's just a sanity check to make sure that today all users pass correct size.
3. Add `Icons.svg(Icon)` method that doesn't have size and make it the recommended way to use svg elements.
4. Fix few places where we use incorrect icon size. It differs by 2px-4px and hardly noticeable. But here is a somewhat noticeable diff for "plus" icon in question bank (now icon is better centered vertically):

before:
image

after:
image

In later PRs I'm planning to update all callers of `Icons.svg(icon, size)` to use `Icons.svg(icon)` and set width/height using classes.

### Checklist

### Issue(s)

Issue #3148

### DIFF
--- a/server/app/views/BaseHtmlView.java
+++ b/server/app/views/BaseHtmlView.java
@@ -77,7 +77,10 @@ public abstract class BaseHtmlView {
   protected static ButtonTag makeSvgTextButton(String buttonText, Icons icon) {
     return TagCreator.button()
         .with(
-            Icons.svg(icon, 20).withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0),
+            Icons.svg(icon, 20)
+                .withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0)
+                // Can't set 18px using Tailwind CSS classes.
+                .withStyle("width: 18px; height: 18px;"),
             span(buttonText).withClass(Styles.TEXT_LEFT));
   }
 
--- a/server/app/views/BaseHtmlView.java
+++ b/server/app/views/BaseHtmlView.java
@@ -77,7 +77,7 @@ public abstract class BaseHtmlView {
   protected static ButtonTag makeSvgTextButton(String buttonText, Icons icon) {
     return TagCreator.button()
         .with(
-            Icons.svg(icon, 18).withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0),
+            Icons.svg(icon, 20).withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0),
             span(buttonText).withClass(Styles.TEXT_LEFT));
   }
 
--- a/server/app/views/components/Icons.java
+++ b/server/app/views/components/Icons.java
@@ -1,5 +1,7 @@
 package views.components;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import services.question.types.QuestionType;
 
 /**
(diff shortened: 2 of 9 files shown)


server/app/views/BaseHtmlView.java
```diff
@@ -77,7 +77,7 @@ protected static ButtonTag asRedirectButton(ButtonTag buttonEl, String redirectU
   protected static ButtonTag makeSvgTextButton(String buttonText, Icons icon) {
     return TagCreator.button()
         .with(
-            Icons.svg(icon, 18).withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0),
+            Icons.svg(icon, 20).withClasses(Styles.ML_1, Styles.INLINE_BLOCK, Styles.FLEX_SHRINK_0),
```
clouser2, line 80: The intention was that SVG buttons would render their icons as 18px (from mocks). Is it possible to add the appropriate styling here such that it will render as 18px?

server/app/views/components/Icons.java
```diff
@@ -312,34 +366,30 @@ public static SvgTag questionTypeSvg(QuestionType type, int width, int height) {
...
-
-  public static SvgTag svg(Icons icon, int width, int height) {
-    return svg(icon.path, width, height);
-  }
-
-  public static SvgTag svg(String pathString, int width, int height) {
+  /** Don't use it. Use {@link #questionTypeSvg(QuestionType type) instead} */
+  public static SvgTag svg(Icons icon, int size) {
+    if (icon.size != size) {
+      throw new IllegalArgumentException(
```
clouser2, line 375: Our ability to detect this is only as good as our test coverage. Given that you'll be converting so that this type of error isn't possible in a future PR, can we downgrade this to a log statement to prevent a gap in our test coverage from potentially causing a user-facing error that is recoverable?

clouser2 (changes requested):
This is a great improvement and moves towards a lot better encapsulation / easy to use icons. Thanks for taking the time to fix this :D

server/app/views/components/Icons.java
nbeloglazov: This check was useful for me to catch errors in existing code by running test. I agree that having exceptions in prod will cause more problems than preventing actual major issue. Downgraded to logging.

server/app/views/BaseHtmlView.java
nbeloglazov: This is one is tricky. We are heavily using tailwindcss and it doesn't have size for 18px, only 16px and 20px: https://tailwindcss.com/docs/width#fixed-widths Do you know if have examples of setting size explicitly on elements, without using tailwind?

clouser2: Once we're on Tailwind 3, we can set the value explicitly like this: https://tailwindcss.com/docs/width#arbitrary-values. That said, we're not using Tailwind 3 yet.

The Tailwind docs semi-imply that we could add another size if needed, but that might not be worth the time to add. I'd be in favor of setting an inline style for this one (e.g. `style="width:18px; height:18px"`)

nbeloglazov: Sg, set width/height using styles.

clouser2 (approved):

PR #3200, opened 2022-08-18
Upload videos on browser test failures when running on CI

nbeloglazov:
### Description

This way it can help us understand the root cause of intermittent azure failures.

### Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -75,7 +75,7 @@ jobs:
       - name: Upload test videos on failure
         uses: actions/upload-artifact@v3
         with:
-          name: test videos directory
+          name: tests videos
           path: browser-test/tmp
         if: failure()
       - name: Print logs on failure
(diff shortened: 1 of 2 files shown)


swatkat1 (approved):
this is Awesome!!

.github/workflows/tests.yaml
```diff
@@ -118,6 +124,12 @@ jobs:
         with:
           name: image diff output directory
           path: browser-test/diff_output
+      - name: Upload test videos on failure
+        uses: actions/upload-artifact@v3
```
swatkat1, line 128: Just one question throguh- where will the videos be uploaded to? Will it be in github?

nbeloglazov: Yes, when a test fails it will upload videos and they will be available on "Summary" page of the github action that failed. Here is the example: https://github.com/civiform/civiform/actions/runs/2886074984 (scroll all the way down).

PR #3204, opened 2022-08-19
Closes #3027: Change "new version" to "edit" on questions tab

### LINKED ISSUE #3027: CiviForm admin UX: Change "new version" to "edit" on questions tab
"New version" on the questions tab is confusing and not consistent with the language we use on the programs tab 

Change "new version" to "edit" on questions tab 

**Additional context**
From admin UXR feedback sheet

**Screenshot**
Screen Shot 2022-08-02 at 3 52 23 PM

ckirksey3:
### Description

- Updates the text of the edit link from "New Version" to "Edit"
- Updates related typescript tests.
- Modifies .gitignore to include IntelliJ config file

Before:
Screen Shot 2022-08-19 at 12 47 29 PM

After:
Screen Shot 2022-08-19 at 12 48 51 PM

## Release notes:

Changed "New Version" to "Edit" on questions tab.

### Checklist

### Issue(s)

Fixes #3027


browser-test/src/support/admin_questions.ts
```diff
@@ -158,7 +158,7 @@ export class AdminQuestions {
     ).toContain('View')
     expect(
       await this.page.innerText(this.selectQuestionTableRow(questionName)),
-    ).toContain('New Version')
+    ).toContain('Edit')
```
clouser2, line 161: Can we make this assert a bit more scoped? I believe this will match "Edit draft ->" as well since it's asserting that the contents only includes the text "Edit". We may need to add a bit more scoped of a selector to try and extract out a list of displayed links and then ensure one of them includes the exact text of "Edit".

clouser2 (changes requested):
Looks great and thanks for the screenshot!

One comment about updating our assertions to be more tightly scoped now that the new text overlaps a bit with "Edit Draft"

browser-test/src/support/admin_questions.ts
ckirksey3: I tried playing with this and am having trouble scoping it down. For instance, this doesn't seem to match:
```
.toEqual('Edit →')
```
Do you have a recommendation for how to best scope it down?

clouser2: > I tried playing with this and am having trouble scoping it down. For instance, this doesn't seem to match:
> 
> ```
> .toEqual('Edit →')
> ```
> 
> Do you have a recommendation for how to best scope it down?

We may have to dive into the DOM a bit more. This may not be better, but we could adjust the selector used for innerText to select that last `td` in the row (https://playwright.dev/docs/selectors#n-th-element-selector). Then we could further subselect all of the `<a>` tags within that row and assert that "Edit" appears.

```
private questionTableRowLinks(questionName: string) {
  return this.selectQuestionTableRow(questionName) + ' td >> nth=-1 > a'
}
```

Then that wouldh ave all "a" tags, which could be used to assert the contents.

browser-test/src/support/admin_questions.ts
```diff
@@ -226,35 +226,15 @@ export class AdminQuestions {
...
-    buttonText,
-  }: {
-    questionName: string
-    buttonText: string
-  }) {
+  async gotoQuestionEditPage(questionName: string) {
     await this.gotoAdminQuestionsPage()
     await this.page.click(
-      this.selectWithinQuestionTableRow(questionName, `:text("${buttonText}")`),
+      this.selectWithinQuestionTableRow(questionName, `:text("Edit")`),
```
clouser2, line 232: Not blocking: Can we add the -> character here without breaking tests?

clouser2 (approved):

browser-test/src/support/admin_questions.ts
ckirksey3: Updated!

ckirksey3: Ok, that worked on the local sample test I ran, but caused some other tests to break so I'm rescinding it.

Issue #2912, opened 2022-07-16
Status Tracking: When a status is selected present confirmation dialog


shanemc-goog:
When the program admin selects a new status show the confirmation dialog.

clouser2:
@shanemc-goog / @rahaghas 

Were there any mocks indicating what the confirmation dialog would say? I'm also happy to just propose text and address it in the PR review.

clouser2:
Presumably, we want the behavior when the user cancels the confirm dialog to be to restore the previously selected status and do nothing. We likely also want the confirmation message to include the previous and to-be-set status as well.

Typically, we do all of this through a modal. However, modals have preset static content for the most part. We could do some tricks to adjust modal content. In this case, though, it feels like we might be able to get away with using the standard Javascript confirmation API. It's not as pretty as a modal, but we can dynamically build the content and it makes for a better flow for preventing updates. We have precedent for using this in the applicant-facing experience when removing an enumerated entity.

For expediency, I plan to start there and we can adjust if we think this is too divergent from the experience we want.

shanemc-goog:
Mocks are [here](https://docs.google.com/presentation/d/12wEhWKPvPh-e3cdWAnL_NdT_I-YRnl1nYBTr9zes1vE/edit?resourcekey=0-x_YmnOany2v3hZr_fgLgJw#slide=id.gf43218f635_0_131)

shanemc-goog:
I'm also recalling that Status and Notes are set differently, so will need a path for each, changing status is priority though.

clouser2:
Thanks for the mocks. I think "Notes" aren't in scope for V1, correct?

rahaghas:
I had it in the V1 mocks, but idk. @cassandralfernandes ?

clouser2:
Regarding the confirmation modal shown. I'll scrap my idea from https://github.com/civiform/civiform/issues/2912#issuecomment-1217053343 since I'd forgotten that we wanted to allow triggering / not triggering the email.

We don't create dynamic modals throughout the codebase, so we'll need to come up with a solution. The simplest would be to create a modal for *each* status and show the correct one depending on the selection. That's a little bit wasteful but is also consistent with what we do elsewhere.

[...]

PR #3211, opened 2022-08-19
Add edit note dialog to the program admin application page

clouser2:
### Description

This is present when application statuses are available. Presently, it's wired to a controller that immediately returns success (with a TODO to implement).

While here, also fixed an issue with `FieldWithLabel` where `setRows` / `setCols` wasn't actually setting the attributes correctly.

### Testing notes
Temporarily enabled status tracking and let CI run on it, which succeeded:
https://github.com/civiform/civiform/runs/7926402730?check_suite_focus=true

### Screenshots

Button in the UI:
Screen Shot 2022-08-19 at 1 17 32 PM

Modal shown:
Screen Shot 2022-08-19 at 1 17 38 PM

## Release notes:

Adding "Edit note" dialog to the program admin application page when application status tracking is enabled.

### Checklist

### Issue(s)

#2912


rahaghas (approved):

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -54,14 +54,15 @@ describe.skip('view program statuses', () => {
...
 
-    it('does not Show status options', async () => {
+    it('does not show status options or edit note', async () => {
       await loginAsProgramAdmin(pageObject)
 
       await adminPrograms.viewApplications(programWithoutStatusesName)
 
       await adminPrograms.viewApplicationForApplicant(userDisplayName())
 
       expect(await adminPrograms.isStatusSelectorVisible()).toBe(false)
```
shanemc-goog, line 64: related to the discussion we had on the previous PR, I think the setup here should be in the `describe` `beforeAll`, and then each expect in its own `it`.  Then they can fail individually and we get clear feedback around which is failing, and without this selector failure shadowing if the Edit Note is working or not.

server/conf/routes
```diff
@@ -95,6 +95,7 @@ GET     /admin/programs/:programId/applications/all                       contro
 GET     /admin/programs/:programId/applications/singleVersion             controllers.admin.AdminApplicationController.downloadSingleVersion(request: Request, programId: Long)
 GET     /admin/programs/:programId/applications/:applicationId            controllers.admin.AdminApplicationController.show(request: Request, programId: Long, applicationId: Long)
 POST     /admin/programs/:programId/applications/:applicationId/updateStatus            controllers.admin.AdminApplicationController.updateStatus(request: Request, programId: Long, applicationId: Long)
+POST     /admin/programs/:programId/applications/:applicationId/editNote            controllers.admin.AdminApplicationController.editNote(request: Request, programId: Long, applicationId: Long)
```
shanemc-goog, line 98: ? should we use update like with updateStatus?  edit feels better in english but I wonder if we are meant to adhere to a smaller set of RESTful verbs.

clouser2: Converted to updateNote for parity. IMO, paths aren't what give RESTfulness, it's the associated HTTP verb. A truly "RESTful" setup here would be a `POST` to `/admin/programs/:programId/applications/:applicationId/notes`. We don't end up following that convention elsewhere, so likely not worth converting to it here.

shanemc-goog: Ah true true.  I think I was miss using RESTful to get at having common/readable terms in our public controller methods.

shanemc-goog (approved):

shanemc-goog:
FYI I merged in the update that turns the browser tests back on. You likely won't notice a diff even if there is one as the default is still true, but I have a issue to fix that.

clouser2:
> FYI I merged in the update that turns the browser tests back on. You likely won't notice a diff even if there is one as the default is still true, but I have a issue to fix that.

Thanks for the info. Did a fresh sync and will merge if tests continue to pass.

PR #3214, opened 2022-08-19
Lower screenshot threshold to 1%

nbeloglazov:
### Description

Size of screenshots is pretty big so having even 3% threshold is too big to catch changes in icons or text. Setting it to 1% for now. Tried setting to 0% but getting some rendering differences when running locally vs on github action. Can't figure out the root cause of the differences as both should be running tests from the same docker image. 

### Checklist


jeanniefu (approved):
This SGTM. Thanks!

PR #3216, opened 2022-08-21
Add ApplicationEvent database table and repository

### LINKED ISSUE #3020: Status Tracking: Data modeling for status changes

```java
  private Account actor;
}

  NOTE_CHANGE("note_change");
  
}

//
  
    private String statusText;
    private Boolean emailSent;
  }

    String note;
  }
}
```

shanemc-goog:
### Description

Add data modeling for ApplicationEvent table which captures an event that occurs to an Application, initially this is setting the status or a note.

The modeling supports store or different event specific details in isolation from each other.

## Release notes:

Adds a new `application_events` database table to support the upcoming Status Tracking feature.

### Checklist

### Issue(s)

Fixes #3020


server/app/models/ApplicationEvent.java
```diff
@@ -0,0 +1,81 @@
...
+  @Constraints.Required private ApplicationEventDetails.Type eventType;
+  // The Account that triggered the event.
+  @ManyToOne
+  @JoinColumn(name = "actor_id")
+  private Account actor;
+  // Details of the event specific to the eventType.
+  @Constraints.Required @DbJson private ApplicationEventDetails details;
+  @WhenCreated private Instant createTime;
+
+  public ApplicationEvent() {}
```
clouser2, line 29: Are there any callers of the empty constructor? Is this needed for EBean? If not, I'd mark it as `private` since it doesn't seem like there'd be any reason to have this.

server/app/models/ApplicationEvent.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public ApplicationEvent(
+      Application application,
+      Account actor,
+      ApplicationEventDetails.Type eventType,
+      ApplicationEventDetails details) {
+    this.application = application;
```
clouser2, line 36: We don't do this in our other models, but I think we could definitely benefit from `checkNotNull()` assertions for all of these. Same would go for all of the setters that EBean (presumably) uses. This will help from setting invalid values into the model.

server/app/services/application/ApplicationEventDetails.java
```diff
@@ -0,0 +1,101 @@
...
+
+      @JsonProperty("email_sent")
+      public abstract Builder setEmailSent(Boolean emailSent);
+
+      public abstract StatusEvent build();
+    }
+  }
+
+  @AutoValue
+  public abstract static class NoteEvent {
```
clouser2, line 92: As I understand right now (for V1), we're only going to display a single note in the application UI and not display the list of notes. Just wanted to check in how you see that being modeled here.

server/app/services/application/ApplicationEventDetails.java
```diff
@@ -0,0 +1,101 @@
...
+ * into a database column.
+ */
+@AutoValue
+@JsonDeserialize(builder = AutoValue_ApplicationEventDetails.Builder.class)
+public abstract class ApplicationEventDetails {
+  /** The event affecting the application. */
+  @JsonProperty("event_type")
+  public abstract Type eventType();
+
+  // Only one of the following Event fields should be set.
```
clouser2, line 26: Is there anything we can build into the class to enforce this constraint?

One thing that might make sense would be to forgo having the AutoValue.Builder below since it gives the illusion that multiple fields could be set. Instead, we could have two static methods:

```
public abstract class ApplicationEventDetails {
...
...

  public static ApplicationEventDetails createStatusEvent(StatusEvent ev) {
    return new ApplicationEventDetails(EventType.STATUS, Optional.of(ev), Optional.empty());
  }
}
```

A downside to this is that as the number of events grows, the number of arguments to the constructor will also grow. However, constraining this complexity to the class itself would help prevent misuse where multiple fields are set.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.CompletionStage;
+import javax.inject.Inject;
+import models.Application;
+import models.ApplicationEvent;
+
+/**
+ * ApplicationEventRepository performs operations on {@link ApplicationEvent} that often involve
+ * other EBean models or asynchronous handling.
+ */
+public class ApplicationEventRepository {
```
clouser2, line 18: Please mark this as `final` to ensure that the decision to subclass is explicit rather than implicit.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,54 @@
...
+  private final Database database;
+  private final DatabaseExecutionContext executionContext;
+
+  @Inject
+  public ApplicationEventRepository(DatabaseExecutionContext executionContext) {
+    this.database = DB.getDefault();
```
clouser2, line 24: Can we add `checkNotNull` here for consistency?

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,54 @@
...
+    this.executionContext = checkNotNull(executionContext);
+  }
+
+  /** Insert a new {@link ApplicationEvent} record asynchronously. */
+  public CompletionStage<ApplicationEvent> insert(ApplicationEvent event) {
```
clouser2, line 29: At present, the only consumer of this API will be the program admin, where calls can be fully synchronous. Can we convert this to return `ApplicationEvent` and do the toCompletableFuture().join() call in the implementation?

Same goes for getEvents until we add the applicant-facing overload of the method. It's entirely possible that for applicants, we'll just want a `getCurrentStatus` or something along those lines, so we might even end up having a different API entirely exposed for that.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,54 @@
...
+    return supplyAsync(
+        () -> {
+          database.insert(event);
+          return event;
```
clouser2, line 33: Call event.refresh() after the insert completes. We do this in other repositories since EBean doesn't refresh the underlying model (i.e. setting the ID field).

server/test/repository/ApplicationEventRepositoryTest.java
```diff
@@ -0,0 +1,53 @@
...
+    ApplicationEventDetails details =
+        ApplicationEventDetails.builder()
+            .setEventType(ApplicationEventDetails.Type.STATUS_CHANGE)
+            .setStatusEvent(
+                StatusEvent.builder().setStatusText("Status").setEmailSent(true).build())
+            .build();
+    ApplicationEvent event =
+        new ApplicationEvent(
+            application, actor, ApplicationEventDetails.Type.STATUS_CHANGE, details);
+    repo.insert(event).toCompletableFuture().join();
```
clouser2, line 41: Mind asserting on the return value here? One interesting thing to assert in my opinion is that the event doesn't have an ID before and does have an ID after.

clouser2 (changes requested):

server/app/models/ApplicationEvent.java
shanemc-goog: removed.

server/app/repository/ApplicationEventRepository.java
shanemc-goog: done.

server/app/repository/ApplicationEventRepository.java
shanemc-goog: done.

server/app/repository/ApplicationEventRepository.java
shanemc-goog: done, but removed all the completion stuff and named the method Sync per existing patterns in other Repositories.

server/app/repository/ApplicationEventRepository.java
shanemc-goog: FYI @clouser2   TIL that @ Id and time annotations are updated automatically since ebean is setting them, so it's not strictly necessary here, but added it anyway for good measure.

refresh seemingly pulls in externally changed data including relationships.

server/app/services/application/ApplicationEventDetails.java
shanemc-goog: Yes! I forgot about that.  I added builder validation to ensure only one is set.

server/app/services/application/ApplicationEventDetails.java
shanemc-goog: This table is just a log, we can apply business logic on top of it.


As is, the most recent value is the relevant one, and the previous are history which we might surface later.

if/when we want to have multiple notes, we'd likely need to add a note id here and track that through them, we can write a data migrator to go update all the existing notes to have the same ID to bootstrap.

server/test/repository/ApplicationEventRepositoryTest.java
shanemc-goog: done I think.

I expanded the test a bit and broke out insert and get since I realized the insert return was usable alone.

server/app/models/ApplicationEvent.java
shanemc-goog: @bion curious your thoughts given the status quo.

Seemingly the "NOT NULL" database annotations are used to enforce this.

I don't see a problem doing it here also, but wonder if it borders on redundant/errorprone having the restriction in 2 places.

clouser2: While perhaps redundant, I'd put it in the category of "more actionable stack traces", since the error will appear at the time the bad data is provided rather than potentially further down (or even in a separate class) when the actual db operation is attempted.

server/app/repository/ApplicationEventRepository.java
clouser2: Ah, thanks! I think I'd been bitten in the past by M2M relationships not being updated or something along those lines. Thanks for adding it here. I think the consistency with the rest of the codebase will help us not to have to think about those things in the future (since database performance isn't yet an issue).

server/app/services/application/ApplicationEventDetails.java
clouser2: Ack!

clouser2 (approved):

server/app/services/application/ApplicationEventDetails.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @AutoValue
+  @JsonDeserialize(builder = AutoValue_ApplicationEventDetails_StatusEvent.Builder.class)
+  public abstract static class StatusEvent {
+    /**
+     * The text of the StatusDefinitions.Status applied to the application in the default locale.
+     */
+    @JsonProperty("status_text")
+    public abstract String statusText();
```
bion, line 82:
```suggestion
    public abstract String statusText();

```

server/app/services/application/ApplicationEventDetails.java
```diff
@@ -0,0 +1,113 @@
...
+      public abstract StatusEvent build();
+    }
+  }
+
+  @AutoValue
+  public abstract static class NoteEvent {
+    @JsonCreator
+    public static NoteEvent create(@JsonProperty("note") String note) {
+      return new AutoValue_ApplicationEventDetails_NoteEvent(note);
+    }
```
bion, line 108:
```suggestion
    }

```

bion (changes requested):

server/app/models/ApplicationEvent.java
bion: I'd also vote for the "belt and suspenders" approach of enforcing it both in the application code as well as in the DB schema

server/app/services/application/ApplicationEventDetails.java
shanemc-goog: done

server/app/services/application/ApplicationEventDetails.java
shanemc-goog: done

server/app/models/ApplicationEvent.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.persistence.Entity;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+import play.data.validation.Constraints;
+import services.application.ApplicationEventDetails;
+
+@Entity
+@Table(name = "application_events")
+public class ApplicationEvent extends BaseModel {
```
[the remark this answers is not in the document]
shanemc-goog: done

server/app/models/ApplicationEvent.java
```diff
@@ -0,0 +1,79 @@
...
+
+  // The Application the event is on.
+  @ManyToOne private Application application;
+  // The {@code ApplicationEventDetails.Type} of the event.
+  @Constraints.Required private ApplicationEventDetails.Type eventType;
+  // The Account that triggered the event.
+  @ManyToOne
+  @JoinColumn(name = "actor_id")
+  private Account actor;
```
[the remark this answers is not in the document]
shanemc-goog: done. actor -> creator

server/app/models/ApplicationEvent.java
shanemc-goog: added the setter/ctor checks.  AutoValue already rejects null by default so the serializer doesn't need to worry about that.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,41 @@
...
+  /** Insert a new {@link ApplicationEvent} record asynchronously. */
+  public ApplicationEvent insertSync(ApplicationEvent event) {
+    database.insert(event);
+    event.refresh();
+    return event;
+  }
+
+  /**
+   * Returns all {@link ApplicationEvent} records for the {@link Application} with id {@code
+   * applicationId} asynchronously.
```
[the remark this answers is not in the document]
shanemc-goog: done. the previous code was async btw

server/app/repository/ApplicationEventRepository.java
```diff
@@ -0,0 +1,41 @@
...
+   */
+  public ImmutableList<ApplicationEvent> getEvents(Long applicationId) {
+    return ImmutableList.copyOf(
+        database
+            .find(ApplicationEvent.class)
+            .where()
+            .eq("application_id", applicationId)
+            .findList());
```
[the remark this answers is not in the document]
shanemc-goog: done

server/conf/evolutions/default/42.sql
```diff
@@ -0,0 +1,18 @@
+ --- Create the ApplicationEvents table.
+
+# --- !Ups
+
+CREATE TABLE IF NOT EXISTS application_events (
+  id BIGSERIAL PRIMARY KEY NOT NULL,
+  application_id BIGINT,
+  actor_id BIGINT,
```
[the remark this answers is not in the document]
shanemc-goog: done.

Issue #3032, opened 2022-08-02
CiviForm admin UX: Add "create new question" button to the question bank 


cassandralfernandes:
**Is your feature request related to a problem? Please describe.**
It is not clear to the user how navigate from the program page to create a new question

**Describe the solution you'd like**
Add "create new question" button to the question bank that takes the user to the questions tab 

**Additional context**
From admin UXR feedback sheet 
We need to confirm whether the button would take the user to a new tab or remain in the same tab

**Screenshot*
<img width="1533" alt="Screen Shot 2022-08-02 at 4 07 45 PM" src="https://user-images.githubusercontent.com/99513234/182489744-fc17a6f0-c396-4f60-ab54-1b68b16b796c.png">

**mocks**
https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13ec4604693_0_91

clouser2:
The question bank doesn't dynamically refresh if a new question is added. Given this, if we opened "Add question" in a new tab, the admin wouldn't see the new contents until the previous tab was refreshed.

@rahaghas Given this, does it make sense to start by opening it in the same tab?

rahaghas:
Yes, what I would love if we could do for this one is have a version of the question tab that hides all the tabs at the top and just includes an "x" in the corner that takes you back to the program you were just editing. How difficult would this be? Including a mock:
<img width="742" alt="Screen Shot 2022-08-22 at 1 07 45 PM" src="https://user-images.githubusercontent.com/103459221/185979177-0d37ac41-2210-4141-91c4-d21182b9e7ce.png">

clouser2:
Credit goes to @nbeloglazov for suggesting an experience where the "source" of showing the "Create question" page is persisted. @rahaghas and I discussed and think it would make sense to do the following:

**Program builder page**
* Add a "Create new question" button in the question bank with the existing dropdown behavior
* Once a selection type is created, the same tab is navigated to the existing "New checkbox question" UI

**New question UI**
* Upon successfully creating the question, navigate back to the program builder page rather than the questions list page
* Add a "Cancel" button at the bottom of the form that also navigates back to the program builder page.

A couple of edge cases to be wary of:
  * If there were form errors, we re-render the create question page. We'll want to ensure that the destination continues to be persisted even in the face of that type of problem.

rahaghas:
[...]

PR #3224, opened 2022-08-22
Displaying a "Create new question" button in the question bank

clouser2:
### Description

Implements the first part of the proposal from https://github.com/civiform/civiform/issues/3032#issuecomment-1222723590. In a forthcoming PR, we'll add the redirect logic so that successful question creation redirects to the program builder.

Note: The placement of the text / button is slightly different from the mocks per @rahaghas 's approval.

### Screenshots

See updated screenshot in the "Files changed" section.

## Release notes:

Displaying a "Create new question" button in the question bank

### Checklist

### Issue(s)

#3032


browser-test/src/admin_program_creation.test.ts
```diff
@@ -66,9 +67,7 @@ describe('program creation', () => {
       '.cf-program-question:has-text("apc-name") >> .cf-remove-question-button',
     )
     await page.click('button:text("apc-enumerator")')
-    expect(await page.innerText('id=question-bank-questions')).toBe(
-      'Add Question',
-    )
+    expect(await page.innerText('id=question-bank-questions')).toBe('')
```
clouser2, line 70: Note to the reviewer: This change and the other is due to the DOM restructuring in QuestionBank. We were applying this ID to the entire question bank container rather than something containing the list of questions, thus causing us to over-specify in these sort of expectations.

We now have a screenshot test if we truly care about asserting that the 'Add Question' text is there.

server/app/views/components/CreateQuestionButton.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Contains a helper method to render the create question button with it's corresponding dropdown.
+ */
+public final class CreateQuestionButton {
+  /** Renders the "Create new question" button with a dropdown for each question type. */
+  public static DivTag renderCreateQuestionButton() {
+    String parentId = "create-question-button";
+    String dropdownId = parentId + "-dropdown";
+    DivTag linkButton =
+        new LinkElement().setId(parentId).setText("Create new question").asButtonNoHref();
```
nbeloglazov, line 22: This is pre-existing code but still few comments. No need to do anything. But if you have background on any of these - would be good for me to learn.
1. Why do we use link as button here while in other places we use regular <button> element.
2. `asButtonNoHref` implies that this element cast as Button (whatever it means) while in fact it actually wraps in a <div> tag.

server/app/views/components/QuestionBank.java
```diff
@@ -108,17 +108,31 @@ private FormTag questionBankPanel() {
...
+        div()
+            .with(
+                div()
+                    .with(
+                        p("Not finding a question you're looking for in this list?"),
+                        div()
+                            .withClass(Styles.FLEX)
+                            .with(
+                                div().withClass(Styles.FLEX_GROW),
+                                CreateQuestionButton.renderCreateQuestionButton()))));
```
nbeloglazov, line 124: Screenshot shows that button rendered underneath "Not finding a question" text while mock has screenshot to be on the right and text on the left.

browser-test/src/admin_program_creation.test.ts
```diff
@@ -126,6 +125,43 @@ describe('program creation', () => {
...
+    await adminPrograms.addProgram(programName)
+    await adminPrograms.goToManageQuestionsPage(programName)
+    await page.click('#create-question-button')
+    await page.click('#create-text-question')
+    await waitForPageJsLoad(page)
+
+    const [questionName, questionText] = [
+      'new-from-question-bank',
+      'Question text',
+    ]
```
nbeloglazov, line 144: That looks a bit odd. 
```
const questionName = ...;
const questionText = ...;
```
is not much longer and easier to read.

nbeloglazov (changes requested):

server/app/views/components/CreateQuestionButton.java
clouser2: Good callout. `LinkElement` is a helper that I'm trying to move us away from for some of the reasons you call out. I switched this to just create a button tag with the appropriate common styling. Calling this a link when it's really a button with a dropout is confusing :).

server/app/views/components/QuestionBank.java
clouser2: Indeed. This was something @rahaghas and I live-built and were ok with the position being slightly different from the mocks. I called this out in the description.

browser-test/src/admin_program_creation.test.ts
clouser2: Indeed. Prior to formatting, this was all on one line, which made it a bit easier to read. I agree that the formatting made this more confusing than helpful. Switched to two separate initializations.

clouser2 (commented):
Feedback incorporated.

clouser2:
> So, as it's implemented at the moment, user will need to go back to the program after they add question, right? Until you implement the "go back to the program" functionality we might surprise users.

Correct. Having the button there will still arguably be an improvement from status quo. It'll only really be an issue if the redirect behavior takes a while to implement and we cut a release without it. I have that PR mostly ready to follow this one.

PR #3228, opened 2022-08-22
Add current screenshots for admin question page

### LINKED ISSUE #3033: CiviForm admin UX: update layout of questions tab
The question list on the questions tab does not have the same layout as the programs tab and it is confusing 

Update question list to be consistent with the program list UI. Draft and active should be two separate list items within the card

**Additional context**
From admin UX feedback sheet

Screen Shot 2022-08-02 at 4 11 01 PM

[Mock](https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13ec4604693_0_182)

nbeloglazov:
### Description

Screenshot is taken after question is created and also when it is updated and in draft form. Also updated question_lifecycle.test.ts to always clear DB upon starting session. It makes screenshots stable by removing order dependency between test cases.

I didn't see noticeable increase in per-test case duration from adding `dropTable` call. 

### Checklist

### Issue(s)

Fixes #3033


browser-test/src/question_lifecycle.test.ts
```diff
@@ -4,21 +4,19 @@ import {
...
 } from './support'
 import {QuestionType} from './support/admin_questions'
+import {BASE_URL} from './support/config'
 
 describe('normal question lifecycle', () => {
-  beforeAll(async () => {
-    const {page} = await startSession()
-    await dropTables(page)
+  it('canonical question seeding works', async () => {
+    const {browser, page} = await startSession(/* clearDb= */ true)
```
jeanniefu, line 16: Why move this out of beforeAll?

browser-test/src/question_lifecycle.test.ts
```diff
@@ -84,10 +82,18 @@ describe('normal question lifecycle', () => {
 
       await adminQuestions.expectActiveQuestions(allQuestions)
 
+      // Take screenshot of questions being published and active.
+      await adminQuestions.gotoAdminQuestionsPage()
+      await validateScreenshot(page)
```
jeanniefu, line 87: nit, blocking: Could we specify a custom file name here and below (e.g. address-active, address-draft)?

Right now, I feel like it's a bit hard to tell what the screenshot is testing by just looking at the file name.

jeanniefu (changes requested):

browser-test/src/question_lifecycle.test.ts
jeanniefu: Although, I also see value in keeping as is since it's easier to trace it to the test that produced it. My personal preference would be to rename the screenshots, but up to you!

jeanniefu: Btw, https://github.com/civiform/civiform/pull/3233 will add the ability to set a custom name more easily!

browser-test/src/question_lifecycle.test.ts
nbeloglazov: There is no need in beforeAll now. This test drop tables on each startSession invocation and `seedCanonicalQuestions` is needed only by one test. And even there I think it's not really useful but sure, why not have test that verifies that seeding works.

browser-test/src/question_lifecycle.test.ts
nbeloglazov: Thanks for refactoring. Updated screenshots.

browser-test/src/question_lifecycle.test.ts
jeanniefu: Oh sorry, I meant why not keep the startSession() call (and maybe some other shared setup) in a beforeEach?

jeanniefu (approved):
1 non blocking comment

browser-test/src/question_lifecycle.test.ts
nbeloglazov: Ah, ok. Looks like this test used different style - starting session on each test case, while other tests start session once and reset it between test cases. I refactored this test to follow same style. And moved 'clearDb' logic to the resetSession method.

PR #3233, opened 2022-08-22
Add applicant-facing screenshot tests for all question types

jeanniefu:
### Description

* Adds a screenshot test of each question type with and without error messages (applicant view).
* Updates screenshot tests to require specifying a custom file name. The current ones are too verbose and not always descriptive.
* Uploads the image diff outputs on failure when run in GitHub.

## Release notes:

N/A

### Checklist


browser-test/src/support/index.ts
```diff
@@ -222,10 +222,13 @@ export const validateAccessibility = async (page: Page) => {
...
  * __snapshots__/test_file_name/name-of-the-test-1-snap.png
  * If the screenshot already exists, compare the new screenshot with the
- * existing screenshot, and save a pixel diff instead if the two don't match
+ * existing screenshot, and save a pixel diff instead if the two don't match.
+ * You can optionally specify customFileName, which will replace the default generated file name but
+ * preserve the default path.
  */
 export const validateScreenshot = async (
   page: Page,
+  customFileName?: string,
```
nbeloglazov, line 231: I suggest making that field mandatory for consistency. There is only one usage right now that doesn't use custom name so it's easy to migrate.

nbeloglazov (approved):

browser-test/src/support/index.ts
jeanniefu: Sounds good, done!

PR #3272, opened 2022-08-25
Create a focused browser tests compose configuration

### LINKED ISSUE #3237: Cleanup: remove status = true tracking from .conf files
We enable status tracking for unit tests and browser tests by default.

Ideally only the relevant tests should have the feature enabled, but as is the entire system enables it which means there's no general testing for the feature being off (and not having unintended coinsequences)

Browser Tests are written to work with the setting but need to remove it form the dev.conf file

Unit tests for the feature assume it is on, and need to be adjusted before removing the setting from the test.conf file.

shanemc-goog:
### Description

Browser tests use the same configuration as dev servers and we have a need to alter that now.

With launch features flags we typically want them on by default to more easily evaluate in dev servers but browser tests shouldn't as it's not the default production state.

This adds a browser test config that inherits from the dev one and undoes what the dev one enables.

## Release notes:

NA doesn't affect production code

### Checklist

### Issue(s)

Fixes #3237


browser-test/browser-test-compose.dev.yml
```diff
@@ -13,7 +13,7 @@ services:
     tty: true # docker run -t
     ports:
       - 9457:9457
-    command: -jvm-debug "*:9457" ~run -Dconfig.file=conf/application.dev.conf
+    command: -jvm-debug "*:9457" ~run -Dconfig.file=conf/application.dev-browser-tests.conf
```
clouser2, line 16: Not blocking: What about `application.browser_tests.conf`?

server/conf/application.dev-browser-tests.conf
```diff
@@ -0,0 +1,9 @@
+# Configuration for the dev browser test CiviForm server.
+include "application.dev.conf"
+
+# The dev server turns somethings on by default so
```
clouser2, line 4: This statement is going to be hard to enforce due to the fact that the configuration inherits from `application.dev.conf`.

What about having this inherit from `application.conf` and forking the current values? There's opportunity for drift there, but less surprise about things "being enabled by default" in dev.

clouser2 (approved):

server/conf/application.dev-browser-tests.conf
shanemc-goog: I thought about that, but was a little worried it be to much change from the status quo,  as well the naming wasn't clear.

If we made the dev server the one with a specific config (EG setting the feature flags for now) , I wasn't quite sure what to name that.  thoughts?

application.dev.conf -> application.dev_base.conf

Then inherit that in both:
- application.dev_server.conf
- application.dev_browser_tests_server.conf (effectively empty for now)

??

clouser2: Having an application.dev_base.conf would only exist in order to share the few common variables. In my opinion, the fact that they're mostly the same doesn't necessarily mean we should share them.

It feels like the reason the naming is challenging is because we're trying to build a relationship between the two conceptually that doesn't necessarily exist.

shanemc-goog: I think you're saying we should duplicate the configuration rather than inherit, is that the case?

In reality they are the same though because their intent is to start the server in the same manner especially wrt to networking, fake components etc. the relationship is they're both trying to run a "dev configured" server environment, one for unit tests one for browser tests.

So duplicating feels like it could be error prone as most changes would have to be done twice.

clouser2: Fine to leave as-is if you'd like. I just have a sense that they happen to be the same right now, even though we might have different expectations of them as time goes on. When that time comes, we could revisit whether forking or duplicating makes more sense.

shanemc-goog: I def like the idea of having them both branch off of the common thing, more like your suggestion I think, rather than assuming one is the "default", but not sure what naming won't be confusing....

browser-test/browser-test-compose.dev.yml
shanemc-goog: I don't see a clear pattern on the two and the directory uses -

shanemc-goog (commented):
Submitting as is to get the utility going.  We can change the layering later if we feel like it.

PR #3275, opened 2022-08-25
When creating a question from the program builder, redirect to the program builder on success.

clouser2:
### Description

Implements the second part of the proposal from https://github.com/civiform/civiform/issues/3032#issuecomment-1222723590. Namely, we propagate a redirect URL when creating a question. Upon successfully creating a question, the redirect URL is hit.

Additionally, a "cancel" button was added to the new question UI that also navigates back to the redirect URL.

### Screenshots
Screen Shot 2022-08-25 at 10 01 47 AM

## Release notes:

When creating a question from the program builder, redirect to the program builder on success.

### Checklist

### Issue(s)


server/app/forms/QuestionForm.java
```diff
@@ -53,49 +59,64 @@ protected QuestionForm(QuestionDefinition qd) {
     }
   }
 
-  public String getQuestionName() {
+  public final String getQuestionName() {
```
bion, line 62: question (non-blocking): why set all of these `final`?

bion (approved):

server/app/forms/QuestionForm.java
clouser2: I set these to final since this is an abstract class and there's no need to allow subclasses to arbitrarily override the implementation. If that need should arise, we can unmark them as final. It's the same logic as marking a class as "final" since the defaults would otherwise leave them "open" to reimplementation.

browser-test/src/support/admin_programs.ts
```diff
@@ -213,6 +213,20 @@ export class AdminPrograms {
     )
   }
 
+  async expectProgramBlockEditPageWithSuccessToast({
```
nbeloglazov, line 216: Personally I would create 2 functions: expectSuccessToast and expectProgramBlockEditPage. They seem to be checking 2 orthogonal things. Also expectSuccessToast can be useful in other places.

server/app/forms/QuestionForm.java
```diff
@@ -53,49 +59,64 @@ protected QuestionForm(QuestionDefinition qd) {
...
-  public void setQuestionHelpText(String questionHelpText) {
+  public final void setQuestionHelpText(String questionHelpText) {
     this.questionHelpText = checkNotNull(questionHelpText);
   }
 
+  public final String getRedirectUrl() {
+    if (redirectUrl == null) {
+      return "";
+    }
+    if (!redirectUrl.isBlank() && URI.create(redirectUrl).isAbsolute()) {
```
nbeloglazov, line 109: Interesting, you decided to do validation in getter instead of setter? Is that to be safer in case someone else sets value outside of the main `setRedirecUrl`? It will make debugging harder though.

nbeloglazov (approved):

browser-test/src/support/admin_programs.ts
clouser2: Done :D

server/app/forms/QuestionForm.java
clouser2: Added a comment. Correct. The goal is to prevent any bad values from finding their way out of the class and the getter seemed the safest place to do this. My concern with doing this in the setter is that there are multiple ways to get these values set that may change over time (e.g. a new constructor that may initialize the value directly, Play claiming it supports building forms with reflection, etc). For this one, it seemed better to play it safe by preventing the value from getting out. In all of these cases, the real way we'd see any errors is through someone directly manipulating HTTP requests.

PR #3277, opened 2022-08-25
Adding warnings when no email will be sent as part of updating a program's status

clouser2:
### Screenshots

No email configured for the status:
Screen Shot 2022-08-25 at 10 30 58 AM

No email provided by applicant: 
Screen Shot 2022-08-25 at 10 34 07 AM

## Release notes:

Adding warnings when no email will be sent as part of updating a program's status

### Checklist

### Issue(s)

### DIFF
--- a/browser-test/src/admin_application_statuses.test.ts
+++ b/browser-test/src/admin_application_statuses.test.ts
@@ -1,4 +1,5 @@
 import {
+  dismissModal,
   startSession,
   logout,
   loginAsGuest,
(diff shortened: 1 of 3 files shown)


browser-test/src/admin_application_statuses.test.ts
```diff
@@ -114,16 +119,43 @@ describe('view program statuses', () => {
...
         expect(await adminPrograms.getStatusOption()).toBe('Choose an option:')
         await adminPrograms.expectUpdateStatusToast()
         // TODO(#3020): Assert that the selected status has been updated.
       })
+
+      it('when no email is configured for the status, a warning is shown', async () => {
+        const modal = await adminPrograms.setStatusOptionAndAwaitModal(
+          noEmailStatusName,
+        )
+        expect(await modal.innerText()).toContain(
```
shanemc-goog, line 141: Do we have a test the the email checkbox is shown when there is an email to be sent?

shanemc-goog (approved):

browser-test/src/admin_application_statuses.test.ts
clouser2: Nope, which is what the TODO tracks. We can't currently add a unit test since we can't test the positive logged-in case from the controller.

We also can't do a browser test since we don't support logging in as an applicant (and thus configuring an email).

PR #3279, opened 2022-08-25
Use fieldset and legend for questions that include multiple input fields

### LINKED ISSUE #2618: Composite form fields should be grouped

Severity (Critical/Serious/Moderate/Good to have): Serious

- Form fields like checkboxes and radio buttons **should** be grouped. The a11y team recommended we use fieldset and legend (see **first** resource link below for an example).
- Address **can** be grouped (see **second** resource link below for an example).
- In the future, if we add a phone number field, we should group these fields (see **third** resource link below for an example)
- Should also group name (We currently have first, middle, last as separate fields)

Resources: 
* https://www.w3.org/WAI/tutorials/forms/grouping/
* https://www.w3.org/TR/WCAG20-TECHS/H71.html
* https://www.w3.org/TR/WCAG20-TECHS/ARIA17.html

jeanniefu:
### Description

Updates all composite questions (e.g. single question with multiple `input` fields) to use a `fieldset` and `legend` for better accessibility with screen readers. This allows the overall question to be associated with each of its individual inputs.

Using the below as an example: without this change, a screen reader user interacting with a single radio button input (e.g. "English") would have trouble associating it with its parent label ("Please select your preferred language"). With this change, each individual radio button now belongs to a properly labelled `fieldset` group, so a screen reader user is easily able to link the radio button to the overall question. See [link](https://www.w3.org/WAI/tutorials/forms/grouping/) for more context.
Screen Shot 2022-08-25 at 4 16 23 PM

This updates all applicant facing questions (e.g. language selection page, and all questions within an application).

Note that there are UI diffs due to slightly different padding following the `fieldset`. I was unable to debug why this happens, but I think it is acceptable. We already have slight variation for padding following different question types (ranging from 34px to 48px for my setup), and this change still keeps the padding within that range.

## Release notes:
[Accessibility] Update composite questions to use a fieldset and legend.

### Checklist

### Issue(s)

Fixes #2618


server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -23,10 +27,18 @@
...
+    COMPOSITE
+  }
+
   protected final ApplicantQuestion question;
+  private final InputFieldType inputFieldType;
 
-  ApplicantQuestionRendererImpl(ApplicantQuestion question) {
+  ApplicantQuestionRendererImpl(ApplicantQuestion question, InputFieldType inputFieldType) {
     this.question = checkNotNull(question);
+    this.inputFieldType = inputFieldType;
```
clouser2, line 41: I believe enums can be null in Java. Can you add a checkNotNull(inputFieldType) here for consistency with other constructors?

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -78,23 +82,53 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
...
     }
 
-    return div()
-        .withId(question.getContextualizedPath().toString())
-        .withClasses(Styles.MX_AUTO, Styles.MB_8, getReferenceClass(), getRequiredClass())
-        .with(questionTextDiv)
-        .with(renderTag(params, validationErrors));
+    if (inputFieldType == InputFieldType.COMPOSITE) {
+      // Group fields with fieldset and legend for a11y.
+      LegendTag legend =
```
clouser2, line 100: Is it possible to avoid forking of this extent due to the different tag type? It makes it a bit difficult to ensure that the majority of the implementation is the same.

I believe that J2HTML has base classes for tags that can contain child elements vs not (https://javadoc.io/static/com.j2html/j2html/1.5.0/j2html/tags/ContainerTag.html).

As I see it, we have a "questionPrimaryText" element which could be either a `<legend>` or a `<div>`. Additionally, we have a container element that could either be a `<fieldset>` or a `<div>`. Could we fork these two based on whether something was composite or not and then apply the additional styling on those?

If the restructuring doesn't feel worth it or possible, feel free to ignore.

server/test/views/questiontypes/ApplicantQuestionRendererFactoryTest.java
```diff
@@ -42,4 +52,19 @@ public void rendererExistsForAllTypes(QuestionType type) throws UnsupportedQuest
     assertThat(renderedContent).contains("Sample question text");
     assertThat(renderedContent).doesNotContain("help text");
   }
+
+  @Test
+  @Parameters(method = "compositeTypes")
```
clouser2, line 57: Nit: When we add new question types it might be possible to forget to update here. Consider having the parameterization pass every question type in.

Then we can check if it's in the set of compositeTypes and assert that fieldset is in it. Otherwise, we'd assert that fieldset isn't in it.

clouser2 (approved):
Looks good!

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Ah thanks, I didn't realize that. Done!

server/test/views/questiontypes/ApplicantQuestionRendererFactoryTest.java
jeanniefu: Sounds good, done!

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Yes, thank you for the suggestion! I refactored this as best I could to make it easier to keep the implementations in sync.

One issue I ran into was when trying to do ContainerTag.with() - error: "unchecked call to with(....) as a member of the raw type j2html.tags.ContainerTag". So I had to keep some of the logic duplicated in the if/else blocks. I'm not sure if there's a way to be able to share that logic.

clouser2 (approved):

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -78,23 +83,51 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
...
               .withClasses(Styles.P_1, Styles.TEXT_RED_600)
               .withText("*" + requiredQuestionMessage));
     }
 
+    // Composite fields should be rendered with fieldset and legend for screen reader users.
+    ContainerTag questionPrimaryTextTag;
+    ImmutableList<DomContent> questionTextDoms =
+        TextFormatter.createLinksAndEscapeText(
+            question.getQuestionText(), TextFormatter.UrlOpenAction.NewTab);
+    if (inputFieldType == InputFieldType.COMPOSITE) {
```
bion, line 104: nit: here and on L113 use `<KnownEnum>.equals(inputFieldType)`

bion: suggestion (non-blocking): here and L113 consider using a single `switch` statement instead of these two `if` statements

bion (changes requested):

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Thanks! Updated to use a switch statement, and had to include a default case (otherwise I get a "variable questionPrimaryTextTag might not have been initialized" error)

I also changed the unit test to use a switch statement and exhaustively list out the enum types, since I think that makes the test clearer.

bion: great!

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -78,23 +83,63 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
...
+        questionTag =
+            div()
+                .with(questionPrimaryTextTag)
+                .with(questionSecondaryTextDiv)
+                .with(renderTag(params, validationErrors));
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Unhandled input field type: %s", inputFieldType));
+    }
```
bion, line 138: suggestion (non-blocking): why not have a single switch and set both variables in it?

bion (approved):

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Yeah I was trying to maximize code sharing so it would be easier to keep in sync, but I think all I'm doing is saving one line (`.withClasses(ReferenceClasses.APPLICANT_QUESTION_TEXT, ApplicantStyles.QUESTION_TEXT)`). Sounds good to remove the extra switch block.

PR #3294, opened 2022-08-29
Add ProfileUtils tester to unlock testing being an admin/program ACL.

shanemc-goog:
### Description

Created a ProfileUtils fake for testing.  Admin Application features are guarded behind ACLs that we currently don't have a pattern for creating the data or faking otherwise.  This utility just bypasses all checks so subsequent code can be tested.

## Release notes:

NA/ testing only

### Checklist

### Issue(s)

### DIFF
--- a/server/test/controllers/admin/AdminApplicationControllerTest.java
+++ b/server/test/controllers/admin/AdminApplicationControllerTest.java
@@ -77,10 +77,7 @@ public class AdminApplicationControllerTest extends ResetPostgres {
     controller = makeNoOpProfileController();
     Program program = ProgramBuilder.newActiveProgram().build();
     Applicant applicant = resourceCreator.insertApplicantWithAccount();
-    applicant.refresh();
-    Application application =
-        Application.create(applicant, program, LifecycleStage.ACTIVE).setSubmitTimeToNow();
-    application.refresh();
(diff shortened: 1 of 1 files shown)


server/test/controllers/admin/AdminApplicationControllerTest.java
```diff
@@ -126,4 +169,60 @@ public void updateNote_notAdmin() throws Exception {
...
+        instanceOf(ProfileUtilsNoOpTester.class),
+        instanceOf(MessagesApi.class),
+        instanceOf(DateConverter.class),
+        Providers.of(LocalDateTime.now(ZoneId.systemDefault())),
+        instanceOf(FeatureFlags.class));
+  }
+
+  // A test version of ProfileUtils that disable functionality that is hard
+  // to otherwise test around.
+  static class ProfileUtilsNoOpTester extends ProfileUtils {
```
clouser2, line 193: I believe we have need for this outside of the class as well. Mind adding it to some test support package?

clouser2 (approved):

server/test/controllers/admin/AdminApplicationControllerTest.java
shanemc-goog: per offline: Will extract it later after it's used more here.  Need to assess the broader usage too, I think only a File class uses ProfileUtils.

PR #3296, opened 2022-08-29
drop tables before canonical questions test

azizova-leyla:
### Description

Fixing the broken test.

## Release notes:

### Checklist

### Issue(s)

### DIFF
--- a/browser-test/src/question_lifecycle.test.ts
+++ b/browser-test/src/question_lifecycle.test.ts
@@ -7,6 +7,7 @@ import {
   waitForPageJsLoad,
   validateScreenshot,
   resetSession,
+  dropTables,
 } from './support'
 import {QuestionType} from './support/admin_questions'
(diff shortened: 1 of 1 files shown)


browser-test/src/question_lifecycle.test.ts
```diff
@@ -25,6 +26,7 @@ describe('normal question lifecycle', () => {
   })
 
   it('canonical question seeding works', async () => {
+    await dropTables(page)
```
nbeloglazov, line 29: Can you add comment saying that this is needed when running on staging where it's not guaranteed that DB is empty when test starts?

azizova-leyla: this was failing on local for me as well. it might be dependent on the order in which tests are run, so it's not always fails, but it's definitely not just staging issue.

nbeloglazov (approved):

Issue #3269, opened 2022-08-25
Status Tracking: Allow filters for the Application list


shanemc-goog:
The entire Application List view is being revamped so we'll need to do a low-waste take at it in the current UI.

A MVP should allow for a filter selector showing
- all options available in the Applications that could be listed
- An option to show those with no status that can have a status selected.

Note: For applications on programs without statuses there is no status option available.

clouser2:
@rahaghas I was thinking for the V1, we could add the filter below the search box and before the Download / Filter buttons:
![Screen Shot 2022-08-25 at 11 29 00 AM](https://user-images.githubusercontent.com/1870301/186741583-818de1c8-11a0-46f9-b967-9c1bfdef6d8b.png)

I think the ideal experience would be a multi-select control of some sort. That said, I feel like checkboxes could take up a lot screen real estate. Additionally, we don't have any sort of dropdown-containing-checkboxes control that we can reuse. Building that would likely be a large amount of effort.

If we don't expect a need to choose more than one status when filtering, our V1 could simply be a select with a dropdown of the options.

What do you think?

clouser2:
As part of this work, an interesting challenge around our data modeling and applying it to filters emerged.

## Background

**Status change modeling** - Presently, any time the status is changed, an `ApplicationEvent` with a type of `status_change` ([code](https://github.com/civiform/civiform/blob/4f0821e44f89b43ee9a0ab86a75b29faabfd83e2/server/app/models/ApplicationEvent.java)) is added. As such, there are many `ApplicationEvent` entities associated with a given `Application`.

**Application filtering requirements** - Presently, we filter the set of applications shown to the program admin entirely in SQL ([code](https://github.com/civiform/civiform/blob/4f0821e44f89b43ee9a0ab86a75b29faabfd83e2/server/app/repository/ProgramRepository.java#L227)) since there can conceivably be O(thousands) of applications. Due to this (per @bion), the decision was made to do filtering at the database level rather than the application level since it would be more performant and consume less memory.

## Problem

Filtering based on application status given the current model would require joining the filtered set of applications with a query that:
  * Retrieves all `ApplicationEvent`s corresponding to a status change
  * Groups by application id and only retrieves the latest value

Expressing this in SQL may be possible, but very likely won't be performant since the cardinality of the right side of the join will consist of all application events for all applications. The belief is that the query planner would likely end up just doing N subqueries for all events corresponding to the filtered set of applications.

[...]

Issue #3264, opened 2022-08-25
Status Tracking: Receive Note Change FE request and save to DB


shanemc-goog:
Send the note from the FE to the BE and save it to the DB.

Sending the request is soon to be done.
The Repository is present, but need the Service and and Controller hookup

clouser2:
Also, update the note shown in the program admin UI.

shanemc-goog:
Just need to update the browser test that verifies the note was updated.

PR #3306, opened 2022-08-30
Surface latest status within the UI, backed by a DB trigger to populate the value

### LINKED ISSUE #3020: Status Tracking: Data modeling for status changes

### LINKED ISSUE #3267: Status Tracking: Reflect current Status in Status selector
The Application view status selector should show the current database status and not allow for the empty/unset value

clouser2:
### Description

Adds a `Application.latest_status` column corresponding to the most recent `ApplicationEvent` with type `status_change`. This value is precomputed via a database trigger.

The rationale for using a database trigger is documented [here](https://github.com/civiform/civiform/issues/3269#issuecomment-1232193163) (also linked in the database evolution).

This value is then consumed to populate the selected status value when the program admin is viewing an application. It is additionally used to in the status change confirmation dialog as the "previous status value".

This value will also be used in a subsequent PR to add filtering of applications via relevant statuses.

## Release notes:

Adds a column indicating the application's current status that has been set by a program admin. 

### Checklist

### Issue(s)
Fixes #3020; Fixes #3267; #3264; #3269

### DIFF
--- a/server/app/models/Application.java
+++ b/server/app/models/Application.java
@@ -120,7 +120,14 @@ public class Application extends BaseModel {
     return this;
   }
 
-  /** Returns the latest application status value associated with the applicaton. */
+  /**
+   * Returns the latest application status value associated with the application.
+   *
+   * <p>This value is updated by DB triggers defined in conf/evolutions/default/44.sql which set the
+   * status to the latest ApplicationEventDetails event for the application with a type of
+   * "status_change". Attempts to update the status manually will be overridden by the trigger (and
+   * has associated tests confirming this).
+   */
   public Optional<String> getLatestStatus() {
     return Optional.ofNullable(latestStatus);
   }
--- a/server/app/models/Application.java
+++ b/server/app/models/Application.java
@@ -121,12 +121,15 @@ public class Application extends BaseModel {
   }
 
   /**
-   * Returns the latest application status value associated with the application.
+   * Returns the latest application status text value associated with the application.
    *
    * <p>This value is updated by DB triggers defined in conf/evolutions/default/44.sql which set the
    * status to the latest ApplicationEventDetails event for the application with a type of
    * "status_change". Attempts to update the status manually will be overridden by the trigger (and
    * has associated tests confirming this).
+   *
+   * <p>If information about the actual event that set this status is desired, make use of
+   * getApplicationEvents instead.
    */
   public Optional<String> getLatestStatus() {
     return Optional.ofNullable(latestStatus);
--- a/server/app/models/Application.java
+++ b/server/app/models/Application.java
@@ -1,5 +1,6 @@
 package models;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.ebean.annotation.DbJson;
 import io.ebean.annotation.WhenCreated;
 import java.time.Instant;
@@ -44,6 +45,7 @@ public class Application extends BaseModel {
   private Instant submitTime;
   private String preferredLocale;
   private String submitterEmail;
+  private String latestStatus;
 
   public Application(Applicant applicant, Program program, LifecycleStage lifecycleStage) {
     this.applicant = applicant;
@@ -116,5 +118,19 @@ public class Application extends BaseModel {
   public Application setSubmitTimeToNow() {
     this.submitTime = Instant.now();
     return this;
+  }
+
+  /** Returns the latest application status value associated with the applicaton. */
+  public Optional<String> getLatestStatus() {
+    return Optional.ofNullable(latestStatus);
+  }
+
+  /**
+   * This is visible only for tests to manipulate the latest status directly in order to ensure that
+   * updates to it are overridden by the configured database trigger.
+   */
+  @VisibleForTesting
+  void setLatestStatusForTest(String latestStatus) {
+    this.latestStatus = latestStatus;
   }
 }
--- a/server/app/views/admin/programs/ProgramApplicationView.java
+++ b/server/app/views/admin/programs/ProgramApplicationView.java
@@ -301,8 +301,7 @@ public final class ProgramApplicationView extends BaseHtmlView {
       String applicantNameWithApplicationId,
       StatusDefinitions.Status status,
       Http.Request request) {
-    // TODO(#3020): Populate the modal content with the previous configured status.
-    String previousStatus = "Unset";
+    String previousStatus = application.getLatestStatus().orElse("Unset");
     FormTag modalContent =
         form()
             .withAction(
--- a/server/app/views/admin/programs/ProgramApplicationView.java
+++ b/server/app/views/admin/programs/ProgramApplicationView.java
@@ -23,7 +23,6 @@ import j2html.tags.specialized.ATag;
 import j2html.tags.specialized.ButtonTag;
 import j2html.tags.specialized.DivTag;
 import j2html.tags.specialized.FormTag;
-import j2html.tags.specialized.OptionTag;
 import j2html.tags.specialized.SelectTag;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -126,7 +125,7 @@ public final class ProgramApplicationView extends BaseHtmlView {
                         div()
                             .withClasses(Styles.FLEX, Styles.MR_4, Styles.SPACE_X_2)
                             .with(
-                                renderStatusOptionsSelector(statusDefinitions),
+                                renderStatusOptionsSelector(application, statusDefinitions),
                                 updateNoteModal.getButton()))
                     .with(renderDownloadButton(programId, application.id)))
             .with(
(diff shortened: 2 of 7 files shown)


server/conf/evolutions/default/44.sql
```diff
@@ -0,0 +1,66 @@
...
+# ---   https://github.com/civiform/civiform/issues/3269#issuecomment-1232193163
+# --- Relevant documentation:
+# --- Trigger functions: https://www.postgresql.org/docs/current/sql-createtrigger.html
+# --- Create trigger statement: https://www.postgresql.org/docs/current/sql-createtrigger.html
+
+# --- !Ups
+
+alter table applications add column if not exists latest_status varchar;
+
+-- The latest status associated with an application is set to NULL if no application event entries
```
shanemc-goog, line 13: Start the comment with what the method does, then have the exceptions/qualifiers second.   The comment kinda reads backwards in general to me, it ends with the intent and leads with the exception.

EG we should be specific that this returns the status text and not the status details json/ etc.

server/conf/evolutions/default/44.sql
```diff
@@ -0,0 +1,66 @@
...
+-- exist. Otherwise, the text from the latest event corresponding to a status change is used.
+CREATE OR REPLACE FUNCTION retrieve_latest_status(app_id applications.id%TYPE) RETURNS text AS $$
```
shanemc-goog, line 15: Functions are in a global namespace I assume so I think these names should be more specific to their usage.  Triggers too?

Here: retrieve_latest_application_status

next one: process_latests_status_on_application_change

etc

server/conf/evolutions/default/44.sql
```diff
@@ -0,0 +1,66 @@
...
+-- A function that sets the "latest_status" column anytime an application is updated.
+-- Note: The "NEW" variable corresponds to the application record to be inserted / updated.
+CREATE OR REPLACE FUNCTION process_application_change() RETURNS TRIGGER AS $$
+  BEGIN
+    NEW.latest_status := retrieve_latest_status(NEW.id);;
+    RETURN NEW;;
+  END;;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER application_change BEFORE INSERT OR UPDATE ON applications
```
shanemc-goog, line 38: I'm not sure why we need to trigger this on insert or update.

For insert it should be null which we can make the default in the add column statement.

For Update, won't we copy forward the existing value with each update?  Or Is it that every update overwrites this to null and then the trigger needs to fix it?

Eitherway can you doc the data-flow/rationale clearer?

server/conf/evolutions/default/44.sql
```diff
@@ -0,0 +1,66 @@
...
+    FOR EACH ROW EXECUTE FUNCTION process_application_change();
+
+-- A function that sets the "latest_status" column on the referenced application record anytime
+-- an event record is updated.
+-- Note: The "NEW" variable corresponds to the application event record to be inserted / updated.
+CREATE OR REPLACE FUNCTION process_application_event_change() RETURNS TRIGGER AS $$
+  BEGIN
+    IF ((NEW.details->>'event_type') = 'STATUS_CHANGE') THEN
+      UPDATE applications
+        SET latest_status = retrieve_latest_status(NEW.application_id)
```
shanemc-goog, line 48: This function is done ON the Events table so the right side here is unnecessarily re-querying the Events table I think; can't it be something like

NEW.details->'status_event'->>'status_text'

server/app/models/Application.java
```diff
@@ -117,4 +119,25 @@ public Application setSubmitTimeToNow() {
...
+
+  /**
+   * Returns the latest application status value associated with the application.
+   *
+   * <p>This value is updated by DB triggers defined in conf/evolutions/default/44.sql which set the
+   * status to the latest ApplicationEventDetails event for the application with a type of
+   * "status_change". Attempts to update the status manually will be overridden by the trigger (and
+   * has associated tests confirming this).
+   */
+  public Optional<String> getLatestStatus() {
```
shanemc-goog, line 131: Is there a reason to not make the full StatusEvent available here?  EG we can't as easily see the value for sendEmail or subsequent data we add.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -251,16 +251,16 @@ private DivTag renderStatusOptionsSelector(StatusDefinitions statusDefinitions)
...
-        .getStatuses()
+    statusDefinitions.getStatuses().stream()
+        .map(StatusDefinitions.Status::statusText)
         .forEach(
-            status -> {
-              String value = status.statusText();
-              OptionTag optionTag = option(value).withValue(value);
-              dropdownTag.with(optionTag);
+            statusText -> {
+              boolean isCurrentStatus = statusText.equals(application.getLatestStatus().orElse(""));
```
shanemc-goog, line 261: suggestion: pull `application.getLatestStatus().orElse("")` into a readable var before the stream operation

shanemc-goog (changes requested):

shanemc-goog:
Looks great!  just documentation / clarity stuff

server/conf/evolutions/default/44.sql
clouser2: Good call. Done.

server/conf/evolutions/default/44.sql
clouser2: Technically yes. However, say we were modifying an old event (either through a DB migration or through new functionality we add). In that case, we still want to set the latest status based on the most recent event. Given that, the query is necessary. Added docs that attempt to clarify this.

server/conf/evolutions/default/44.sql
clouser2: Moved both trigger definitions down side-by-side and attempted to document. As currently implemented, all is as you said. However, this trigger is specifically here to guard against application code attempting to change that behavior by explicitly modifying the latest_status column. The trigger guards against that being supported (and has unit tests to assert that). We can drop this in the future if we have a real need to modify the column in application code. However, that'll open us up to all of the concerns around ensuring that the column is manually kept in sync within application code.

bion (approved):
Thank you for documenting this so carefully on the referenced issue. And thanks @shanemc-goog for the great PR comments helping document the code itself. I think this approach will provide good scalability and performance while also keeping the likelihood of an inconsistent database state low.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -258,7 +258,8 @@ private DivTag renderStatusOptionsSelector(
         .map(StatusDefinitions.Status::statusText)
         .forEach(
             statusText -> {
-              boolean isCurrentStatus = statusText.equals(application.getLatestStatus().orElse(""));
+              String latestStatusText = application.getLatestStatus().orElse("");
```
shanemc-goog, line 261: This can go above the entire stream operation so it's capture outside of the forEach context and not done each time.

server/conf/evolutions/default/44.sql
```diff
@@ -28,39 +28,46 @@ $$ LANGUAGE SQL;
...
 -- an event record is updated.
 -- Note: The "NEW" variable corresponds to the application event record to be inserted / updated.
-CREATE OR REPLACE FUNCTION process_application_event_change() RETURNS TRIGGER AS $$
+CREATE OR REPLACE FUNCTION process_latest_status_on_application_event_change() RETURNS TRIGGER AS $$
   BEGIN
     IF ((NEW.details->>'event_type') = 'STATUS_CHANGE') THEN
       UPDATE applications
-        SET latest_status = retrieve_latest_status(NEW.application_id)
+        -- Don't always use the status event being modified since we care only about setting the
+        -- latest_status column based on the most recent application event, which isn't necessarily
```
shanemc-goog, line 46: Ah, because this happens on updates too we want to find the latest as it might not be the latest?

That's ok, though a little unfortunate as we should never be updating the event log rows, but to cover all the bases I see why its here.

shanemc-goog (commented):
just questions, LGTM

shanemc-goog (approved):
Can you do the unacknowledged doc update too?

server/conf/evolutions/default/44.sql
clouser2: Good call. Done!

server/conf/evolutions/default/44.sql
clouser2: Yah, exactly. Unfortunate and probably will never happen, but still necessary.

server/app/views/admin/programs/ProgramApplicationView.java
clouser2: Done.

PR #3309, opened 2022-08-30
Pass the Application Note to the BE and save to the DB

### LINKED ISSUE #3020: Status Tracking: Data modeling for status changes

### LINKED ISSUE #3264: Status Tracking: Receive Note Change FE request and save to DB

shanemc-goog:
### Description

Save the Application Note into the database.

Add an Id and name to the existing UI elements and have the Controller/Service save it as an ApplicationEvent.

## Release notes:

Program Application Notes are now saved into the database. 

### Checklist

### Issue(s)

Addresses #3020 #3264

### DIFF
--- a/server/app/controllers/admin/AdminApplicationController.java
+++ b/server/app/controllers/admin/AdminApplicationController.java
@@ -2,6 +2,7 @@ package controllers.admin;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static views.admin.programs.ProgramApplicationView.NEW_STATUS;
+import static views.admin.programs.ProgramApplicationView.NOTE;
 import static views.admin.programs.ProgramApplicationView.SEND_EMAIL;
 
 import annotations.BindingAnnotations.Now;
@@ -403,10 +404,21 @@ public final class AdminApplicationController extends CiviFormController {
     if (!applicationMaybe.isPresent()) {
       return notFound(String.format("Application %d does not exist.", applicationId));
     }
+    Application application = applicationMaybe.get();
 
-    // TODO(#3020): Actually edit the note rather than unconditionally returning success.
-    return redirect(
-            routes.AdminApplicationController.show(programId, applicationMaybe.get().id).url())
+    Map<String, String> formData = formFactory.form().bindFromRequest(request).rawData();
+    Optional<String> maybeNote = Optional.ofNullable(formData.get(NOTE));
+    if (maybeNote.isEmpty()) {
+      return badRequest("A note is not present.");
+    }
+    String note = maybeNote.get();
+
+    programAdminApplicationService.setNote(
+        application,
+        ApplicationEventDetails.NoteEvent.create(note),
+        profileUtils.currentUserProfile(request).get().getAccount().join());
+
+    return redirect(routes.AdminApplicationController.show(programId, application.id).url())
         .flashing("success", "Application note updated");
   }
 
--- a/server/app/services/applications/ProgramAdminApplicationService.java
+++ b/server/app/services/applications/ProgramAdminApplicationService.java
@@ -59,9 +59,7 @@ public final class ProgramAdminApplicationService {
             .setEventType(ApplicationEventDetails.Type.STATUS_CHANGE)
             .setStatusEvent(newStatus)
             .build();
-    ApplicationEvent event =
-        new ApplicationEvent(
-            application, admin, ApplicationEventDetails.Type.STATUS_CHANGE, details);
+    ApplicationEvent event = new ApplicationEvent(application, admin, details);
     eventRepository.insertSync(event);
   }
 
@@ -76,8 +74,7 @@ public final class ProgramAdminApplicationService {
             .setEventType(ApplicationEventDetails.Type.NOTE_CHANGE)
             .setNoteEvent(note)
             .build();
-    ApplicationEvent event =
-        new ApplicationEvent(application, admin, ApplicationEventDetails.Type.NOTE_CHANGE, details);
+    ApplicationEvent event = new ApplicationEvent(application, admin, details);
     eventRepository.insertSync(event);
   }
 }
--- a/server/app/services/applications/ProgramAdminApplicationService.java
+++ b/server/app/services/applications/ProgramAdminApplicationService.java
@@ -49,7 +49,7 @@ public final class ProgramAdminApplicationService {
   }
 
   /**
-   * Saves a new {@link ApplicationEventDetails.Type.STATUS_CHANGE} event.
+   * Sets the status on the {@code Application}.
    *
    * @param admin The Account that instigated the change.
    */
@@ -66,7 +66,7 @@ public final class ProgramAdminApplicationService {
   }
 
   /**
-   * Saves a new {@link ApplicationEventDetails.Type.NOTE_CHANGE} event.
+   * Sets the note on the {@code Application}.
    *
    * @param admin The Account that instigated the change.
    */
--- a/server/app/services/applications/ProgramAdminApplicationService.java
+++ b/server/app/services/applications/ProgramAdminApplicationService.java
@@ -10,6 +10,7 @@ import models.ApplicationEvent;
 import repository.ApplicationEventRepository;
 import repository.ApplicationRepository;
 import services.application.ApplicationEventDetails;
+import services.application.ApplicationEventDetails.NoteEvent;
 import services.application.ApplicationEventDetails.StatusEvent;
 import services.program.ProgramDefinition;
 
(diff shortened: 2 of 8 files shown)


server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -2,6 +2,7 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static views.admin.programs.ProgramApplicationView.NEW_STATUS;
+import static views.admin.programs.ProgramApplicationView.NOTE;
```
clouser2, line 5: Leaving comment here for general PR description: This doesn't address all of the open TODOs in #3020, so I'd recommend not marking it as fixing the issue. For example, the changes in PR #3306 will be required first.

In that PR, I repointed the "show the current note" TODOs against #3264, so I'd also suggest not marking this as fixing that issue. I couldn't find an issue in the status tracking tracker that was along the lines of "show the current note" that would be appropriate to point it against. I'd honestly just use #3264 for that, since setting an application note and being able to observe the changes seems like the smallest useful unit for a user.

Totally fine as a separate PR if you'd prefer, I'd just suggest not marking this PR as fixing the issues since there's still remaining work.

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -58,4 +64,20 @@ public void setStatus(Application application, StatusEvent newStatus, Account ad
             application, admin, ApplicationEventDetails.Type.STATUS_CHANGE, details);
     eventRepository.insertSync(event);
   }
+
+  /**
+   * Saves a new {@link ApplicationEventDetails.Type.NOTE_CHANGE} event.
+   *
+   * @param admin The Account that instigated the change.
+   */
+  public void setNote(Application application, NoteEvent note, Account admin) {
```
clouser2, line 73: Not blocking: Reading the docstring next to the method name is a bit incongrous (docstring=Saves a new and methodname=setNote). setNote implies there's a single note, while we model it as a new event of a given type. The UI will only display the latest note (at least as I understand). That said, at the service level, it feels like we're tying the method name to that implementation a little too strongly.

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -58,4 +64,20 @@ public void setStatus(Application application, StatusEvent newStatus, Account ad
...
+    ApplicationEventDetails details =
+        ApplicationEventDetails.builder()
+            .setEventType(ApplicationEventDetails.Type.NOTE_CHANGE)
+            .setNoteEvent(note)
+            .build();
+    ApplicationEvent event =
+        new ApplicationEvent(application, admin, ApplicationEventDetails.Type.NOTE_CHANGE, details);
```
clouser2, line 80: Didn't notice originally, but why do we have to specify the ApplicationEventDetails.Type both in the ApplicationEventDetails builder as well as the ApplicationEvent constructor? Could we remove the argument from the constructor and instead infer it from the provided details? That would cut out a class of potential mismatch errors that might come our way and simplify usage.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -269,17 +270,24 @@ private Modal renderUpdateNoteConfirmationModal(
       long programId, Application application, Http.Request request) {
     ButtonTag triggerButton =
         makeSvgTextButton("Edit note", Icons.EDIT).withClasses(AdminStyles.TERTIARY_BUTTON_STYLES);
+    // Give each form a unique id based on the application.
+    String formId = String.format("note-form-%d", application.id);
```
clouser2, line 274: If we're not selecting based on this ID anywhere, trying to make it human readable doesn't provide a lot of benefit in my opinion. It also makes it potentially more likely to run into conflicts if we were to either 1) copy/paste or 2) Have a similar element with the same name elsewhere in the DOM. Both of these are unlikely, obviously.

That said, if it doesn't need to have human-readable meaning, I'd prefer making that explicit through initializing the form with a random ID.

We could use a random UUID. It'd make sense in my opinion to pull Modal.randomModalId() into a helper function in something in app/views for this purpose.

server/test/controllers/admin/AdminApplicationControllerTest.java
```diff
@@ -363,6 +363,55 @@ public void updateNote_notAdmin() throws Exception {
...
+    // Execute.
+    Result result = controller.updateNote(request, program.id, application.id);
+
+    // Verify.
+    assertThat(result.status()).isEqualTo(BAD_REQUEST);
+    assertThat(contentAsString(result)).contains("A note is not present");
+  }
+
+  @Test
+  public void updateNote_succeeds() throws Exception {
```
clouser2, line 386: Mind adding a test with a note of the empty string "" succeeding? I think supporting this is definitely desirable since it's effectively "clearing" any existing notes. I just think it'll be helpful to have a test case that confirms explicit support for that value (and that it's not, for example, treated as a bad request).

clouser2 (changes requested):

server/app/controllers/admin/AdminApplicationController.java
shanemc-goog: I don't believe having Fixes closes the Issue, It's more as "This addresses issue X".

AFAIK our Fixes section is just self-imposed organization we have created similar to the Release Notes section and the checklist.  It's meaningless aside from humans who read it.

server/app/services/applications/ProgramAdminApplicationService.java
shanemc-goog: They both have different utility, one is at the database level to query on, the other is that I think it's reasonable to want to pass the Details object rather than the entire Event.  I might be wrong there, but felt better to start with than add subsequently.

server/app/controllers/admin/AdminApplicationController.java
clouser2: I don't think that's quite correct. Empirically, GitHub closes issues prefixed with the "Fixes" keyword, I've made use of this in many PRs. To just reference another issue, not prefixing with "Fixes" does the trick (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

I've observed this behavior with PRs. See https://github.com/civiform/civiform/pull/3289 for an example where the issue was automatically marked as fixed once the merge had completed and https://github.com/civiform/civiform/pull/3207 for an example where the issue didn't have the "Fixes" prefix and was still referenced (but not closed).

server/app/services/applications/ProgramAdminApplicationService.java
clouser2: Put another way, is there ever a case where we expect the two to mismatch? If not, then I don't see the need for it in the constructor. EBean's population of an ApplicationEvent won't use the constructor, I don't believe. It will instead use the provided getters/setters. Constructor initialization is only for our application code, where I don't think there's a need to specify them separately.

server/app/services/applications/ProgramAdminApplicationService.java
shanemc-goog: At some layer we should be imposing business logic on top of the data and I think that's appropriate in the Service layer as it's the lowest non database oriented layer.

There is logically 1 active status or note, there are also historical ones that the Service may later reveal to the a Controller.

As you've brought up before, we don't have great lines as such, often using the data class directly in controllers, but I don't think that is a good thing to not try to impost the business logic of viewing the database in a Service.

Thoughts?

server/app/controllers/admin/AdminApplicationController.java
shanemc-goog: Ah confusing.  Seemingly you can't get it right if two parallel PRs address an Issue and collectively fix it.

updated

server/app/services/applications/ProgramAdminApplicationService.java
shanemc-goog: Ah sorry, I missed the second infer part of your original comment, remove the param in the ctor.

server/app/views/admin/programs/ProgramApplicationView.java
shanemc-goog: Done. Modal is already here so used it directly.

server/test/controllers/admin/AdminApplicationControllerTest.java
shanemc-goog: agreed and done!

clouser2 (approved):
Looks good!

server/app/services/applications/ProgramAdminApplicationService.java
shanemc-goog: Hmm, that's a good approach, at least we can _try_ to enforce business logic through methods and their docs.

I'd prefer to do that in a separate PR though.

PR #3311, opened 2022-08-31
Refactor asRedirectButton method

nbeloglazov:
### Description

1. Rename it to `asRedirectElement` and update types to take any element. This is needed as we are planning to make some other elements, for example divs that act as rows, clickable.

2. Handle clicks using JS instead of inline `onclick` attribute. That gives us more flexibility in future as we can add extra JS in listener, for example call to `e.stopPropagation()`

### Checklist

### Issue(s)

Issue #3033

### DIFF
--- a/server/app/views/BaseHtmlView.java
+++ b/server/app/views/BaseHtmlView.java
@@ -68,12 +68,21 @@ public abstract class BaseHtmlView {
   }
 
   protected static ButtonTag redirectButton(String id, String text, String redirectUrl) {
-    return asRedirectButton(
+    return asRedirectElement(
         TagCreator.button(text).withId(id).withClasses(Styles.M_2), redirectUrl);
   }
 
-  protected static ButtonTag asRedirectButton(ButtonTag buttonEl, String redirectUrl) {
-    return buttonEl.attr("onclick", String.format("window.location = '%s';", redirectUrl));
+  /**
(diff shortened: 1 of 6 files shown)


server/app/views/BaseHtmlView.java
```diff
@@ -68,12 +68,21 @@ protected static ButtonTag submitButton(String id, String textContents) {
...
+  /**
+   * Turns provided element into a clickable element. Upon click the user will be redirected to the
+   * provided url. It's up to caller of this method to style element appropriately to make it clear
+   * that the element is clickable. For example add hover effect and change cursor style.
+   *
+   * @return The element itself.
+   */
+  protected static <T extends Tag> T asRedirectElement(T element, String redirectUrl) {
+    // Attribute `data-redirect-to` is handled in JS by main.ts file.
+    element.attr("data-redirect-to", redirectUrl);
```
clouser2, line 84: Not blocking: Can save a line by returning the result of element.attr(), since J2HTML elements support chaining.

clouser2 (approved):
Nice!

server/app/views/BaseHtmlView.java
nbeloglazov: That how it was initially. But with introduction of T it no longer works. Some generics shenanigans because `.attr()` also returns generic T according to function declaration and compiler somehow things that it returns `Tag`

bion (approved):

Issue #3222, opened 2022-08-22
Dockerize cloud


azizova-leyla:
Put all deploy scripts in a docker container, so user only needs to have docker installed.

azizova-leyla:
Counting this as fixed. We're building docker images for cloud.

PR #3318, opened 2022-08-31
Add new RUN command as a start of all deploy scripts

azizova-leyla:
### Description

New pattern of calling deploy scripts:
1. Call cloud/shared/run.py --command=[setup, deploy, destroy] --tag=IMAGE_TAG --config=CONFIG_FILE
2. Inside run.py load config, validate variables and then call out to actual command.

This centralize config loading to one file, and will be used in the dockerized version.

I'm planning to delete bash versions of scripts from cloud/shared/bin later, leaving for now so all current scripts work.

Also fixed actions that were not setting --tag properly.

## Release notes:

### Checklist

### Issue(s)

First step to [#3222](https://github.com/civiform/civiform/issues/3222)


cloud/shared/bin/deploy.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
+import importlib
+import os
+from cloud.shared.bin.lib.config_loader import ConfigLoader
```
bion, line 4: nit: `ConfigLoader` isn't used in this file

cloud/shared/bin/destroy.py
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env python3
+
+from cloud.shared.bin.lib.config_loader import ConfigLoader
```
bion, line 3: nit: `ConfigLoader` isn't used in this file

cloud/shared/bin/lib/config_loader.py
```diff
@@ -45,20 +47,35 @@ def use_local_backend(self):
...
+            exit(f'Cannot find file {full_config_path}')
+        print(f'Getting config from {full_config_path}')
+        command = shlex.split(
+            f'env -i bash -c "source {full_config_path} && env"')
+        proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+        for line in proc.stdout:
+            (key, _, value) = line.decode().partition("=")
+            os.environ[key] = value.strip()
+        proc.communicate()
+
```
bion, line 67: suggestion (blocking): this could use a comment to describe what's happening, and potentially extracted to a method with a more descriptive method name

cloud/shared/bin/lib/config_loader.py
```diff
@@ -124,7 +139,10 @@ def get_cloud_provider(self):
         return self.configs.get('CIVIFORM_CLOUD_PROVIDER')
 
     def get_template_dir(self):
-        return self.configs.get('TERRAFORM_TEMPLATE_DIR')
+        template_dir = self.configs.get('TERRAFORM_TEMPLATE_DIR')
+        if template_dir is None or not os.path.exists(template_dir):
+            exit(f'Cound not find template directory {template_dir}')
```
bion, line 144:
```suggestion
            exit(f'Could not find template directory {template_dir}')
```

cloud/shared/bin/run.py
```diff
@@ -0,0 +1,63 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--tag', help='Civiform image tag. Required for Setup and Deploy.')
+    parser.add_argument(
+        '--command',
+        help='Command to run. If ommited will validate config and exit.')
+    parser.add_argument(
+        '--config',
+        default='civiform_config.sh',
+        help='Path to civiform config file.')
```
bion, line 24: nit: to disambiguate with `application.conf` maybe say "Path to CiviForm deployment config file"

bion, line 24: nit: to disambiguate with `application.conf` maybe say "Path to CiviForm deployment config file"

bion (changes requested):

bion (approved):

cloud/shared/bin/destroy.py
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env python3
+
+from cloud.shared.bin.lib import terraform
+from cloud.shared.bin.lib.setup_class_loader import get_config_specific_destroy
+"""
+Destroy.py destroys Civiform deployment.
+"""
```
nbeloglazov, line 7: Module description should be at the beginning of the file
https://google.github.io/styleguide/pyguide.html#382-modules

cloud/shared/bin/destroy.py
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env python3
```
nbeloglazov, line 1: You don't need shebang here given that this file is not main (and won't be run from command line)
https://google.github.io/styleguide/pyguide.html#37-shebang-line

cloud/shared/bin/deploy.py
```diff
@@ -0,0 +1,16 @@
+import subprocess
+import importlib
+import os
+
+
+def run(config):
```
nbeloglazov, line 6: Can you add type hint here:

```
def run(config: ConfigLoader): 
   ...
```

This `run` function is not directly called anywhere so it might be challenging to figure out what `config` here is. 

Same for all other `run` functions in this PR.

cloud/shared/bin/lib/config_loader.py
```diff
@@ -45,20 +47,37 @@ def use_local_backend(self):
...
 
-    def _load_config(self):
+    def _get_config_values_from_sh_file(self, config_file):
+        ## 1. Export all variables from the config into clean environment
+        ## 2. Set values in current environment
+        if not os.path.exists(config_file):
+            exit(f'Cannot find file {config_file}')
+        print(f'Getting config from {config_file}')
+        command = shlex.split(f'env -i bash -c "source {config_file} && env"')
+        proc = subprocess.Popen(command, stdout=subprocess.PIPE)
```
nbeloglazov, line 61: Interesting trick of reading shell variables without needing to pre-source it.

cloud/shared/bin/lib/config_loader.py
```diff
@@ -45,20 +47,37 @@ def use_local_backend(self):
...
+        for line in proc.stdout:
+            (key, _, value) = line.decode().partition("=")
+            os.environ[key] = value.strip()
```
nbeloglazov, line 64: Hm. Do we actually need to expose these variables back to environment? Or can pass around as config?

cloud/shared/bin/run.py
```diff
@@ -0,0 +1,63 @@
+#! /usr/bin/env python3
+
+import argparse
+import os
+import sys
+import importlib
+
+sys.path.append(os.getcwd())
```
nbeloglazov, line 8: Why is this here between imports? Can you add comment explaining this unconventional call?

cloud/shared/bin/run.py
```diff
@@ -0,0 +1,63 @@
...
+
+    args = parser.parse_args()
+    if args.tag:
+        os.environ['TF_VAR_image_tag'] = args.tag
+    elif args.command is not None and args.command != 'destroy':
+        exit('--tag is required')
+
+    os.environ['TF_VAR_FILENAME'] = "setup.auto.tfvars"
+    os.environ['BACKEND_VARS_FILENAME'] = 'backend_vars'
+    os.environ['TERRAFORM_PLAN_OUT_FILE'] = 'terraform_plan'
```
nbeloglazov, line 34: FWIW it's odd to see these variables here. Would be better to set them only when we call terraform, and maybe pass them just to the call itself, if it's possible, rather than setting global variables. But it's probably requires more work.

nbeloglazov (changes requested):

cloud/shared/bin/lib/config_loader.py
azizova-leyla: Currently ConfigLoader reads some values from env variables, and I don't want to change that in this PR.
Next big change is to make config yaml and all of this code will be changed. At that point ConfigLoader will be one place to get all the values.
It will likely require rewrite of some bash scripts that we still have.

nbeloglazov (approved):

PR #3320, opened 2022-08-31
Adding application status as a filter option for program admins.

### LINKED ISSUE #3269: Status Tracking: Allow filters for the Application list

clouser2:
### Description

Adds a dropdown option for filtering by application status.

### Screenshots
Screen Shot 2022-08-30 at 10 56 19 AM

## Release notes:

Allow program admins to filter applications by application status if present.

### Checklist

### Issue(s)

Fixes #3269


server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -124,17 +133,21 @@ public Content render(
   }
 
   private FormTag renderSearchForm(
-      ProgramDefinition program, ButtonTag downloadButton, RenderFilterParams filterParams) {
+      ProgramDefinition program,
+      ImmutableList<String> allApplicationStatuses,
```
shanemc-goog, line 137: This won't show program status unless an application has it right?  So initially no Statuses would be present right?

Not a blocker but let's run it by Raha too

server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -161,7 +174,27 @@ private FormTag renderSearchForm(
...
+            allApplicationStatuses.size() > 0,
+            new SelectWithLabel()
+                .setFieldName(APPLICATION_STATUS_PARAM)
+                .setLabelText("Application status")
+                .setValue(filterParams.selectedApplicationStatus().orElse(""))
+                .setOptions(
+                    ImmutableMap.<String, String>builder()
+                        .put("Any application status", "")
+                        .put(
+                            "Only applications without a status",
```
shanemc-goog, line 188: What do you think about putting a "divider" selection to distinguish between these special options and the real one?

Just thinking "-------"

server/test/repository/ProgramRepositoryTest.java
```diff
@@ -305,6 +311,153 @@ private Application makeApplicationWithName(
...
+        adminAccount, thirdStatusApplication, FIRST_STATUS, SECOND_STATUS, THIRD_STATUS);
+
+    Application noStatusApplication =
+        resourceCreator.insertActiveApplication(resourceCreator.insertApplicant(), program);
+
+    Application backToNoStatusApplication =
+        resourceCreator.insertActiveApplication(resourceCreator.insertApplicant(), program);
+    // Application has transitioned through statuses and arrived back at an empty status (indicated
+    // by null).
+    createStatusEvents(adminAccount, backToNoStatusApplication, SECOND_STATUS, null);
```
shanemc-goog, line 363: 1. We shouldn't use null unless forced too so use Optional not null, or another pattern.  We're following the google java guides which dissuade null unless there's no option.

2. you only use the status text off of the status, so just pass that and empty string as needed directly.

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -457,12 +464,24 @@ public Result index(
...
             applications,
             RenderFilterParams.builder()
                 .setSearch(search)
                 .setFromDate(fromDate)
                 .setUntilDate(untilDate)
+                .setSelectedApplicationStatus(applicationStatus)
                 .build()));
   }
+
+  private ImmutableList<String> getAllApplicationStatusesForProgram(ProgramDefinition program) {
```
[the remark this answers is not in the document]
clouser2: Done. My general preference is to use a strong type instead of a database ID since I think it conveys more meaning. That said, in this case, we're actually calling a method that ends up being for all versions of the program, not just the one for the given ID, so it probably was ending up being more confusing than helpful.

server/app/repository/ProgramRepository.java
```diff
@@ -259,6 +259,15 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
       }
     }
 
+    String toMatchStatus = filters.applicationStatus().orElse("");
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/repository/SubmittedApplicationFilter.java
```diff
@@ -24,6 +24,13 @@ public abstract class SubmittedApplicationFilter {
   /** Filters to applications that were submitted within the provided date range. */
   public abstract TimeFilter submitTimeFilter();
 
+  /**
+   * If specified and non-empty, returns applications that match the provided status. If
+   * NO_STATUS_FILTERS_OPTION_UUID is provided, it indicates a desire to match only applications
```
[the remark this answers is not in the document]
clouser2: Done.

server/app/views/admin/programs/ProgramApplicationListView.java
clouser2: This is a list of all unique application statuses for all versions of a program and isn't directly tied to what applications have the statuses set. Put another way, the source of this information comes from ProgramDefinitions, not Applications.

I renamed to `allPossibleProgramApplicationStatuses` in an effort to clarify this.

As such, I don't think there's anything we need to run by UX since I imagine we'd want to show all statuses that could be selected, not just the ones that currently are set on an application.

server/test/repository/ProgramRepositoryTest.java
clouser2: Good call to not use null. Unless you feel strongly, I have a personal preference to pass an `Optional<StatusDefinitions.Status>` rather than a string since I think passing the type better documents the intended type of parameter than an arbitrary string (even though a string is sufficient).

clouser2:
Feedback incorporated!

shanemc-goog (approved):
woohoo!

Issue #3323, opened 2022-09-01
Standardize browser handling in browser tests


nbeloglazov:
At the moment there are 2 different styles of working with browser: launch new instance of browser on each test case and destroy it afterwards or launch browser once and reuse its context between test cases, resetting civiform play session. Both approaches have downsides: launching browser each time slows down tests and reusing same context breaks video recording: we get single video for multiple test cases. Additionally it requires extra work from eng to remember destroy browser or reset session.

We should centralize browser handling and update existing tests to follow the new pattern. Proposed pattern is to create a new function `createTestContext` that returns an object containing `page`, helper objects like `adminPrograms`. That function will register appropriate `before/afterAll`, `before/afterEach` handlers to ensure that browser context is reset correctly between tests. Additionally it will ensure that db is reset between tests. 

Example usage: 

```js
describe('some test', () => {
  const ctx = createTestContext()

  it('should do foo', async () => {
    await ctx.page.click('#some-button')
  })
})
```

nbeloglazov:
Here is more detailed description on how browser context/session is reset between tests. 

First, we need to close browser context after each test case so that its video is saved as a separate file. If we reuse browser context between multiple test cases we end up with longer videos that span multiple test cases and hard to understand (can't see at which point next test starts).

Second, we need to make sure that browser is initialized and ready to use during all `beforeAll/afterAll` invocations. 

Resetting session/context means that users shouldn't expect that `ctx.page` will be on certain page or logged in when they implement `afterAll`. It's important to note as it's reasonable to expect that when you logged in in `beforeAll` - then you'll be logged in during the test and in `afterAll`. In fact each test will get fresh session and need to log in in `beforeEach`. 

Here is example of how sessions reset between tests (different numbers mean different sessions). The example assumes that there can be multiple before/afterAll and nested describes.

```
describe()
    [1] beforeAll()
    describe()
        [1] beforeAll()
        // first test
        [2] beforeEach()
        [2] beforeEach()
        [2] it()
        [2] afterEach()
        [2] afterEach()
        // second test
        [3] beforeEach()
        [3] beforeEach()
        [3] it()
        [3] afterEach()
        [3] afterEach()
        [4] afterAll()
    describe()
        [4] beforeAll()
        // third test
        [5] beforeEach()
        [5] it()
        [5] afterEach()
        [6] afterAll()
    [6] afterAll()
```

@jeanniefu this should answer the question why we have fresh objects. 
[...]

PR #3324, opened 2022-09-01
Introduce createTestContext() function

nbeloglazov:
### Description

The function simplifies managing browser-related, session stuff: it resets session between tests and launches/shutdowns browser once per test file. Using this function test writers don't need to worry about initialization of browser. Example usage:

```ts
describe('some test', () => {
  const ctx = createTestContext();

  it('does something', async () => {
    // do something with ctx.page
  })
})
```

By default it clears database between tests. Many tests rely on database not being cleared so for those tests I disable DB clearing. In future we should rework them so that each re-creates data from scratch. Doing it now will increase test runtime significantly. Once we have some batch approach to create data we can revisit it.

### Checklist

### Issue(s)

Issue #3323

### DIFF
--- a/browser-test/src/support/index.ts
+++ b/browser-test/src/support/index.ts
@@ -9,6 +9,15 @@ import {
 import * as path from 'path'
 import {MatchImageSnapshotOptions} from 'jest-image-snapshot'
 import {waitForPageJsLoad} from './wait'
+import {
+  BASE_URL,
+  DISABLE_SCREENSHOTS,
+  TEST_USER_LOGIN,
+  TEST_USER_PASSWORD,
+} from './config'
+import {AdminQuestions} from './admin_questions'
+import {AdminPrograms} from './admin_programs'
+
 export {AdminApiKeys} from './admin_api_keys'
 export {AdminQuestions} from './admin_questions'
 export {AdminPredicates} from './admin_predicates'
@@ -20,12 +29,6 @@ export {ClientInformation, TIDashboard} from './ti_dashboard'
 export {ApplicantQuestions} from './applicant_questions'
 export {NotFoundPage} from './error_pages'
 export {clickAndWaitForModal, dismissModal, waitForPageJsLoad} from './wait'
-import {
-  BASE_URL,
-  TEST_USER_LOGIN,
-  TEST_USER_PASSWORD,
-  DISABLE_SCREENSHOTS,
-} from './config'
 
 export const isLocalDevEnvironment = () => {
   return (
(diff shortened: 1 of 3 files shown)


browser-test/src/support/index.ts
```diff
@@ -82,6 +89,96 @@ export const startSession = async (): Promise<{
...
+  adminQuestions: AdminQuestions
+  adminPrograms: AdminPrograms
+}
+
+/**
+ * Launches a browser and returns context that contains objects needed to
+ * interact with the browser. Example usage:
+ *
+ * describe('some test', () => {
+ *   const ctx = createBrowserContext()
```
jeanniefu, line 113: createTestContext

browser-test/src/support/index.ts
```diff
@@ -82,6 +89,96 @@ export const startSession = async (): Promise<{
   return {browser, context, page}
 }
 
+/**
+ * Object containing properties and methods for interacting with browser and
+ * app. See docs for createBrowserContext() method for more info.
```
jeanniefu, line 94: createTestContext

browser-test/src/support/index.ts
```diff
@@ -82,6 +89,96 @@ export const startSession = async (): Promise<{
...
+    await resetContext()
+  })
+
+  beforeEach(async () => {
+    await resetContext()
+  })
+
+  afterEach(async () => {
+    if (clearDb) {
+      await dropTables(ctx.page)
```
jeanniefu, line 167: Some of our tests have nested describes, e.g.
```
describe('Applicant flow', () => {
  describe('Checkbox q', () => {
    it ('test case 1', () => {...})
    it ('test case 2', () => {...})
  })

  describe('Address q', () => {
    it ('test case 1', () => {...})
    it ('test case 2', () => {...})
  })
})
```

In this case, is the intention that the `const ctx = createTestContext()` goes in the innermost describe?

browser-test/src/support/index.ts
```diff
@@ -82,6 +89,96 @@ export const startSession = async (): Promise<{
...
+    }
+    // resetting context here so that afterAll() functions of current describe()
+    // block and beforeAll() functions of the next describe() block have fresh
```
jeanniefu, line 170: Why is it important that the afterAll and beforeAll have a fresh page object?

browser-test/src/support/index.ts
nbeloglazov: Thanks. Done

browser-test/src/support/index.ts
nbeloglazov: Thanks. Done

browser-test/src/support/index.ts
nbeloglazov: No, `const ctx = createTestContext()` goes at the very beginning of the top-most describe. Added comment in the jsdoc about it.

browser-test/src/support/index.ts
nbeloglazov: Well, it flows naturally from the fact that we need to reset state in `afterEach` at which point follow up `afterAll` and `beforeAll` will get a fresh state.

browser-test/src/support/index.ts
jeanniefu: Ah I'm sorry, I was misunderstanding how beforeEach worked. I thought it would only be on the outer level describe, but I see now that it will run before each nested "it". Ok this makes sense then, so we drop the tables after each test.

browser-test/src/support/index.ts
jeanniefu: Ok, sounds good

jeanniefu (approved):

PR #3328, opened 2022-09-01
Change buttons layout on admin question page

nbeloglazov:
### Description

This is part of #3033

1. Show only "edit" button and hide all other buttons under triple dot button.
2. Remove "view" button and instead make whole row clickable. Upon click it opens "edit" version of the question. It works as "view" as well if user doesn't actually update question and saves it.
3. Add "See list" clickable link to see active and draft programs that use given question. Previously bold text "x active and y draft programs" was clickable instead.
4. Update close button on modal dialogs to use proper X icon rather thatn letter `x`. 

## Release notes:

Update layout of admin questions page to make it easier to use.

### Checklist

### Issue(s)

Issue #3033


browser-test/src/support/admin_questions.ts
```diff
@@ -179,9 +179,6 @@ export class AdminQuestions {
 
     expect(tableInnerText).toContain(questionName)
     expect(tableInnerText).toContain(questionText)
-    expect(
```
clouser2, line 182: Should this assertion be changed to something else?

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -238,19 +241,22 @@ private Pair<TdTag, Optional<Modal>> renderReferencingPrograms(
...
-    if (maybeReferencingProgramsModal.isPresent()) {
-      referencingProgramsCountContainer =
-          a().withId(maybeReferencingProgramsModal.get().getTriggerButtonId())
-              .withClasses(Styles.DECORATION_SOLID, Styles.CURSOR_POINTER)
-              .with(referencingProgramsCount);
-    }
-
     TdTag tag =
-        td().with(p().with(span("Used across "), referencingProgramsCountContainer))
+        td().with(p().with(span("Used across "), referencingProgramsCount))
```
clouser2, line 245: I'd confirm with @rahaghas about having a "View List" link versus linkifying the count (as before).

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -343,51 +349,79 @@ private Optional<ButtonTag> renderQuestionTranslationLink(QuestionDefinition def
...
     Pair<DomContent, Optional<Modal>> archiveOptionsAndModal =
         renderArchiveOptions(questionForArchive, activeAndDraftQuestions, request);
-    DomContent archiveOptions = archiveOptionsAndModal.getLeft();
-    Optional<Modal> maybeArchiveModal = archiveOptionsAndModal.getRight();
-    td.with(archiveOptions);
-    return Pair.of(td, maybeArchiveModal);
+    extraActions.add(archiveOptionsAndModal.getLeft());
+
+    // Build extra actions button and menu.
+    String extraActionsButtonId = "extra-actions-" + questionForArchive;
```
clouser2, line 394: Since the actual ID needn't be human readable, what about using `Modal.randomModalId` to generate a unique identifier? Side note is that we could probably pull the "random ID" function up to something higher level than just modal.

clouser2 (approved):

browser-test/src/support/admin_questions.ts
nbeloglazov: I don't think so. The fact that it's active is being tested by the presence of "New version" button. I think it's enough. And it will change soon anyway as we going to render both active and draft rows.

server/app/views/admin/questions/QuestionsListView.java
nbeloglazov: Mocks contain "View list" link: https://github.com/civiform/civiform/issues/3033#issuecomment-1223383317 Do you think I need to clarify that? I don't have context on the original implementation of linkifying the count.

clouser2: Ah, those are likely way more up to date than my knowledge. Probably safe to go with what you have :D

server/app/views/admin/questions/QuestionsListView.java
nbeloglazov: Done.

PR #3329, opened 2022-09-01
Update code naming for TI notification in the submit flow

shanemc-goog:
### Description

We have generic naming in our notification flow for what is in effect a TI flow; which is confusing and errorprone.

Also the current naming of "submitter" is ambiguous as the code flows explicitly mean "not the applicant", but applicants most often are the ones submitting applications. 

The code flows _could_ work if the submitter was a program admin, but the product does not allow for that, so I didn't aim to find a term that encompassed "not an applicant" and went with TI specifically.

The effective code flow root of the changes is [ApplicationService.submitApplication](https://github.com/civiform/civiform/pull/3329/files#diff-163cd8467b46ba09a916370044f5c51a2955bc56c2ee4650fb62ca66ccdd0d26R274)

## Release notes:

NA, just code renaming.

### Checklist

### Issue(s)

### DIFF
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -63,23 +63,23 @@ public class ApplicationRepository {
    * and create a new application in the active state.
    */
   public CompletionStage<Application> submitApplication(
-      Applicant applicant, Program program, Optional<String> submitterEmail) {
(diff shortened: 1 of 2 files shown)


bion (approved):
Great improvement, thanks!

PR #3330, opened 2022-09-01
Add debug logging of sent emails

### LINKED ISSUE #2716: Log email content in dev mode
Email is not testable outside of prod instances.  It _should_ be working in staging, but practically doesn't seem to be, and there's no mechanism for local dev.

We should at least log the email in dev mode for easy of local development.

shanemc-goog:
### Description

The emails we send are unobservable for local development. This debug logs them and adds a log configuration example that can be enabled to show them for development.

Upon submitting an application examples of the logging are

[debug] s.c.a.SimpleEmail - Sending email:
SendEmailRequest(Source=noreply@fake.identity, Destination=Destination(ToAddresses=[fake-local-admin-3@example.com, fake-local-admin-5@example.com, fake-local-admin-7@example.com, fake-local-admin-10@example.com]), Message=Message(Subject=Content(Data=New application 13 submitted), Body=Body(Text=Content(Data=Applicant 8 submitted a new application 13 to program Mock program.
View the application at http://localhost:9000/admin/programs/2/applications/13.))))

[debug] s.c.a.SimpleEmail - Sending email:
SendEmailRequest(Source=noreply@fake.identity, Destination=Destination(ToAddresses=[shane@example.com]), Message=Message(Subject=Content(Data=Your application to program Mock program is received), Body=Body(Text=Content(Data=Your application to program Mock program has been received. Your applicant ID is 8 and the application ID is 13.
Log in to CiviForm at http://localhost:9000.))))

## Release notes:

NA - logging that's disabled by default.

### Checklist

### Issue(s)

Fixes #2716

### DIFF
--- a/server/app/services/cloud/aws/SimpleEmail.java
+++ b/server/app/services/cloud/aws/SimpleEmail.java
@@ -59,6 +59,7 @@ public class SimpleEmail {
 
   public void send(ImmutableList<String> toAddresses, String subject, String bodyText) {
     if (toAddresses.isEmpty()) {
+      logger.warn("Attempting to send email to no recipients");
       return;
     }
 
@@ -73,6 +74,9 @@ public class SimpleEmail {
 
       SendEmailRequest emailRequest =
           SendEmailRequest.builder().destination(destination).message(msg).source(sender).build();
+      if (logger.isDebugEnabled()) {
+        logger.debug("Sending email:\n{}", emailRequest);
+      }
       client.get().sendEmail(emailRequest);
     } catch (SesException e) {
       logger.error(e.toString());
--- a/.github/workflows/scorecards.yaml
+++ b/.github/workflows/scorecards.yaml
@@ -50,6 +50,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: 'Upload to code-scanning'
-        uses: github/codeql-action/upload-sarif@v2.1.21
+        uses: github/codeql-action/upload-sarif@v2.1.22
         with:
           sarif_file: results.sarif
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -63,23 +63,23 @@ public class ApplicationRepository {
    * and create a new application in the active state.
    */
   public CompletionStage<Application> submitApplication(
-      Applicant applicant, Program program, Optional<String> submitterEmail) {
+      Applicant applicant, Program program, Optional<String> tiSubmitterEmail) {
     return supplyAsync(
-        () -> submitApplicationInternal(applicant, program, submitterEmail),
+        () -> submitApplicationInternal(applicant, program, tiSubmitterEmail),
         executionContext.current());
   }
 
   public CompletionStage<Optional<Application>> submitApplication(
-      long applicantId, long programId, Optional<String> submitterEmail) {
+      long applicantId, long programId, Optional<String> tiSubmitterEmail) {
     return this.perform(
         applicantId,
         programId,
         (ApplicationArguments appArgs) ->
-            submitApplicationInternal(appArgs.applicant, appArgs.program, submitterEmail));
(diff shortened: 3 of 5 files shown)


server/app/services/cloud/aws/SimpleEmail.java
```diff
@@ -73,6 +74,9 @@ public void send(ImmutableList<String> toAddresses, String subject, String bodyT
 
       SendEmailRequest emailRequest =
           SendEmailRequest.builder().destination(destination).message(msg).source(sender).build();
+      if (logger.isDebugEnabled()) {
+        logger.debug("Sending email:\n{}", emailRequest);
```
clouser2, line 78: Definitely correct me if I'm wrong here. Based on my understanding, these logs will show up for all providers. Presumably the toString() of the email request object would contain both the payload of the email as well as the email address. If that's the case, wouldn't these be considered PII and not something we'd want to log?

Perhaps the answer is just that "we'll never turn on debugging logging for prod", but I'd be a lot more comfortable if we instead logged only the application ID or something that didn't contain an email address.

Relatedly, I was poking around with @nbeloglazov yesterday, and we found that localstack has a debug endpoint for sent emails (https://docs.localstack.cloud/aws/ses/). We could presumably use this to assert email-related code for bin/run-dev and browser tests.

clouser2 (changes requested):
The implementation itself seems reasonable, but I had a high level question / concern around logging the email contents and whether we could approach this in a different way.

server/app/services/cloud/aws/SimpleEmail.java
shanemc-goog: yes and yes.  (adding an example in the desc too)

At the end of the day the system and the people running it could log everything if they desired.  The defaults in the logging conf already hide DEBUG logging that would expose PII otherwise, one in particular logs all SQL queries and results which would include emails.  So this seems quite in line with that.  It's also completely possible there's already a deeper logging target that we could set to debug to see this content. 

Adding the endpoint seems great, but also seemingly a larger investment than we have prioritized now, and likewise only works appropriately if correctly configured and used in dev, which is the same as this PR.

Feel free to have bion make a call.

shanemc-goog: Also the problem with not logging the content is we can never dev test new email content.  So eitherway we need to be able to see the full payload locally.

shanemc-goog: Actually I'll just add @bion  :D

clouser2: That rationale all makes sense. I just wanted to be purposeful about when we wanted to add logging such as this. SQL logging at debug level definitely seems concerning for the same reasons to me, but that's probably part of a larger cleanup.

I think the concern from my perspective is that the person viewing logs doesn't necessarily have to be the same as the person "running the system".

Regarding adding the endpoint and prioritization, I perhaps was being a bit too vague. The endpoint already exists and is exposed by localstack. On a run-dev environment, I "logged" in as f@example.com and submitted an application. I then navigated to http://localstack:4566/_localstack/ses and saw a JSON representation of the email that was sent. IMO, this is a stronger test that the email was sent than a log statement prior to calling the endpoint. Totally fine to defer this, I just wanted to call out that there's no new implementation required to have localstack debugging.

bion (changes requested):

server/app/services/cloud/aws/SimpleEmail.java
shanemc-goog: Reverting as we determined offline that all email content is already logged to http://localstack:4566/_localstack/ses (note you need to setup a /etc/hosts for that to work)

shanemc-goog:
Reverting as we determined offline that all email content is already logged to http://localstack:4566/_localstack/ses (note you need to setup a /etc/hosts for that to work)


Will make sure this is documented in our dev docs

PR #3333, opened 2022-09-01
Updating assertions to expect 'Guest', since we log in as guest when completing applications

clouser2:
### Description

Fixes failing Seattle staging prober runs ([example](https://github.com/seattle-civiform/civiform-deploy/runs/8135343650?check_suite_focus=true)). This was due to the fact that we use `userDisplayName`, which returns "Guest" for non-prober runs and the test account's name when running against Seattle staging.

The test code had copied this from elsewhere, which meant that it was expecting the test user when run as a prober, even though it had completed the application as a guest.

## Release notes:

N/A.

### Checklist

### DIFF
--- a/browser-test/src/admin_application_statuses.test.ts
+++ b/browser-test/src/admin_application_statuses.test.ts
@@ -8,7 +8,6 @@ import {
   selectApplicantLanguage,
   ApplicantQuestions,
   AdminPrograms,
-  userDisplayName,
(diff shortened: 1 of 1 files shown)


clouser2:
FYI @swatkat1. Once submitted, this should fix probers.

swatkat1 (approved):

Issue #3326, opened 2022-09-01
Add browser test support for logged-in applicant experience


clouser2:
Currently, the logged-in applicant experience is ONLY used when the `TEST_USER_LOGIN` / `TEST_USER_PASSWORD` environment variables are set in the test runner ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/browser-test/src/support/index.ts#L144-L161)). Otherwise, we silently fall back to logging in as the guest.

This means that any test that make use of `loginAsTestUser` may actually be logging in as guest and not actually exercising the code paths we intend to.

Presently, we have four environments where we run tests:

1. **CI browser test runs** - These don't set the environment variable and hence don't exercise the "logged-in" flow ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/bin/run-browser-tests-ci)).
2. **Seattle staging prober run** - The environment variables are set and passed via GitHub secrets ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/bin/run-prober)). Additionally, our current "login" code is purpose-built for the Seattle staging auth provider ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/browser-test/src/support/index.ts#L146-L156)).
3. **AWS staging prober run** - These don't set the environment variable and don't exercise the "logged-in" flow ([code](https://github.com/civiform/civiform-staging-deploy/blob/34f4f20e9bbb37cd577ee6bf3a19523b9efb6a1a/bin/run-prober)). Note that this invocation uses the "latest" browser-test image and the script is in a separate repo.
4. **Local browser test runs** - The environment variables aren't set in either of our local run-browser-tests script and hence don't test the logged-in experience ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/bin/run-browser-tests)).

For status tracking, we presently will only send notification emails to the applicant if they've logged in ([code](https://github.com/civiform/civiform/blob/ea9541238973e92f3a6c3723265be3c887d5e9f7/browser-test/src/admin_application_statuses.test.ts#L165-L166)). As it stands, due to the above, there presently is no way to test this case since we don't currently support logged-in applicants for browser tests (except in Seattle staging prober runs).

[...]

PR #3334, opened 2022-09-01
Making logged-in experience work in bin/run-browser-tests, bin/run-browser-tests-local, and bin/run-browser-tests-ci

clouser2:
### Description

Implements "Stage 1" of the changes described in https://github.com/civiform/civiform/issues/3326#issuecomment-1234629983.

Namely:
  * Browser test code has a different strategy for logging in as a test user for fake-oidc (bin/run-browser-tests and bin/run-browser-tests-ci), seattle-staging (bin/run-prober), and aws-staging (will be used by AWS staging prober in phase 2) 
  * If a strategy isn't provided, fall back on the previous behavior. Once AWS staging and Seattle staging have been updated, we can remove the fallback strategy (tracked by TODOs)
  * Adjust the fake-oidc provider to allow for redirect URIs of civiform:9000

### Testing

* bin/run-browser-tests works and manually confirmed that code which calls `loginAsTestUser` now attempts login as the user specified in the environment variable rather than "Guest".

## Release notes:

Infrastructure changes to improve testing of the logged-in applicant experience.

### Checklist

### Issue(s)

#3326


browser-test/browser-test-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       - 10000:10000
 
   fake-idcs:
-    image: docker.io/civiform/oidc-provider:latest
+    image: civiform/oidc-provider
```
clouser2, line 14: Note to the reviewer: This follows the pattern we have for other images that we build locally in CI and want to use for the test run. Prior to this change, we built a local OIDC provider but then always pulled the version from DockerHub (and hence CI runs wouldn't reflect local changes).

server/app/auth/oidc/applicant/IdcsProvider.java
```diff
@@ -72,7 +72,7 @@ protected String getResponseType() {
     // https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#responsetypes
     // TODO: turn this into it's own fake provider or make the local provider allow
     // 'token'.
-    if (baseUrl.contains("localhost:")) {
+    if (baseUrl.contains("localhost:") || baseUrl.startsWith("http://civiform:")) {
```
shanemc-goog, line 75: can you explicitly comment what the two fake domains support?

Previously it's effectively all "local fake IDCS" but now a reader isn't going to know where the lines are in our testing between the names.

test-support/test_oidc_provider.js
```diff
@@ -7,30 +7,32 @@ const configuration = {
...
-      // "native" because we're on localhost.
-      application_type: 'native',
+      application_type: 'web',
       scopes: ['openid', 'profile'],
       redirect_uris: [
         'http://localhost:9000/callback/OidcClient',
         'http://localhost:9000/callback/AdClient',
         'http://localhost:19001/callback/OidcClient',
         'http://localhost:19001/callback/AdClient',
+        'http://civiform:9000/callback/OidcClient',
```
shanemc-goog, line 17: Likewise can you suss out where localhost and civiform are separately relevant?  This file in generally is hard to follow how we use it.

Pointing people to the invalidate below or making a file level comment for both works.

test-support/test_oidc_provider.js
```diff
@@ -7,30 +7,32 @@ const configuration = {
...
       async claims() {
         return {
           sub: id,
           // pretend to be IDCS which uses this key for user email.
-          user_emailid: id + '@example.com',
+          user_emailid: email,
           // lie about verification for tests.
           email_verified: true,
-          user_displayname: 'first middle last',
+          user_displayname: email,
```
shanemc-goog, line 35: can you make this a string with more than 1 part? "user $email", etc.

  It'll help visual testing since names will often be multipart.

browser-test/src/support/index.ts
```diff
@@ -266,34 +268,106 @@ export const setLangEsUS = async (page: Page) => {
...
-    await page.fill('input[name=password]', TEST_USER_PASSWORD)
-    await page.click('button:has-text("Login"):not([disabled])')
-    await page.waitForNavigation({waitUntil: 'networkidle'})
-  } else {
-    await page.click('#guest')
+  switch (TEST_USER_AUTH_STRATEGY) {
+    case 'fake-oidc':
+      await loginAsTestUserFakeOidc(page)
+      break
+    case 'aws-staging':
```
shanemc-goog, line 275: I don't see that these are used in this PR, when will they be used/verified?

shanemc-goog (changes requested):
LGTM just comment stuff

test-support/test_oidc_provider.js
```diff
@@ -44,6 +46,20 @@ const configuration = {
...
 const oidc = new Provider('http://localhost:' + oidcPort, configuration)
 
+const {invalidate: orig} = oidc.Client.Schema.prototype
+
+// By default, redirect URLs must be https and cannot refer to localhost. While these are correct
+// for production implementations, this is intended to be used ONLY for testing. For browser tests,
+// internal Docker networking is used, which uses HTTP. For running the server locally, redirects
+// are pointed towards localhost.
+// From https://github.com/panva/node-oidc-provider/blob/0fcc112e0a95b3b2dae4eba6da812253277567c9/recipes/implicit_http_localhost.md
+oidc.Client.Schema.prototype.invalidate = function invalidate(message, code) {
```
bion, line 56: thanks for the clear doc!

bion (approved):

server/app/auth/oidc/applicant/IdcsProvider.java
clouser2: Done.

test-support/test_oidc_provider.js
clouser2: Attempted to clarify these. I *think* there's an opportunity to get rid of the 19001 callbacks entirely, but I'll do that separately after this has been submitted.

test-support/test_oidc_provider.js
clouser2: This one's actually kind of fun. For AWS staging which uses Auth0, the returned "name" attribute will always be ONLY the email address (`foo@example.com`). For Seattle staging, it's a `first middle last` setup. We have server-side code that handles each of these cases [here](https://github.com/civiform/civiform/blob/5bc4314e36da0b4b2d02f60c2bfc8fdeb8ab94d9/server/app/services/applicant/ApplicantData.java#L87-L109).

Given that values here are pretty flexible, I'd personally prefer to go for the "simpler" approach of a single part, since that's an equally valid scenario.

I left a comment as such. If you feel strongly, let me know and I can try and switch to multi-part. Given that we only have a single ID and we want to be able to differentiate between different logged-in guests, I'm not sure what we'd want to use for the second part.

clouser2 (commented):
Feedback mostly incorporated. Main open question is around whether to return a multi-part display name or not.

test-support/test_oidc_provider.js
shanemc-goog: Progress is better, so ok, but for manual testing having multiple name parts will allow for dev testing flow issues that might break up the content or wrap it oddly.

shanemc-goog (approved):

Issue #2891, opened 2022-07-13
Reset database between browser tests in a single run


jeanniefu:
Some browser tests call `dropTable()` before each test case: [code](https://github.com/seattle-uat/civiform/blob/9a42804c84531f87117f7cb12b7fde211f1a8fe2/browser-test/src/application_download.test.ts#L19-L23)

We also globally delete the database before an entire test run using globalSetup: [code](https://github.com/seattle-uat/civiform/blob/9a42804c84531f87117f7cb12b7fde211f1a8fe2/browser-test/jest.config.js#L7). (Can consider changing this to [setupFiles](https://jestjs.io/docs/configuration#setupfiles-array) if we want to clear between runs)

In an ideal state, each test case should be hermetic (database updates in one test should not be able to affect the state of another test), so we should clear the database between each test case. In a less than ideal but better than what we have state, we could clear the database between each test suite.

We should evaluate the performance of now vs. clearing DB between test suites vs. clearing DB between test cases. Unknown how much time clearing DB takes.

nbeloglazov:
While migrating tests to TestContext became a bit more familiar with existing tests. Some tests fill db with some data (questions, applications) and `beforeAll` and then each test case relies on that data being available. Migrating those tests to do it in `beforeEach` has pretty bad performance hit. For example some tests go from 20s to 100s. So we need to come up with an efficient way to prepare and share DB state between test cases.

As I see we have 2 separate but somewhat related problems:
1. Quickly seed DB with data for a test. Currently we do it by clicking through UI and it's kinda slow.
2. Share seeded DB between test cases.

For (1) we could do something like sending JSON with data and saving it to DB. But I don't think we can fully get rid of manually clicking through UI to prepare the desired initial DB state for a test. 

For (2) we could do something along `dev-dump`, `dev-restore-db` scripts: introduce a new TS function `prepareDB` that users can call and pass callback that builds necessary DB state. This callback will be ran once, in `beforeAll` and DB state is saved somewhere. Then in `beforeEach` that state will be restored. That way we can be sure that test cases don't affect each other as each test cases gets a new DB state. So test fill look something like:

```ts
describe('some test', () => {
  const ctx = createTestContext()
  
  describe('some specific scenario', () => {
    prepareDb(async () => {
      const {adminQuestions, adminPrograms} = ctx
      adminQuestions.createQuestion(...)
      adminQuestions.createQuestion(...)
      adminPrograms.createProgram(...)
    })

    it('check question', async () => {
[...]

PR #3336, opened 2022-09-02
Migrate applicant question tests to use TestContext

nbeloglazov:
### Description

See #3323 for context.

### Checklist

### DIFF
--- a/browser-test/src/checkbox.test.ts
+++ b/browser-test/src/checkbox.test.ts
@@ -1,40 +1,23 @@
-import {Page} from 'playwright'
 import {
-  AdminPrograms,
-  AdminQuestions,
-  ApplicantQuestions,
+  createTestContext,
   loginAsAdmin,
   loginAsGuest,
   logout,
   selectApplicantLanguage,
-  startSession,
-  resetSession,
   validateAccessibility,
   validateScreenshot,
 } from './support'
 
 describe('Checkbox question for applicant flow', () => {
-  let pageObject: Page
-
-  beforeAll(async () => {
-    const {page} = await startSession()
-    pageObject = page
-  })
-
-  afterEach(async () => {
-    await resetSession(pageObject)
-  })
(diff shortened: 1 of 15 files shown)


browser-test/src/support/index.ts
```diff
@@ -116,6 +118,7 @@ export interface TestContext {
  *   const ctx = createTestContext()
  *
  *   it('should do foo', async () => {
+  const {page, applicantQuestions} = ctx
```
jeanniefu, line 121: nit: fix formatting

browser-test/src/checkbox.test.ts
```diff
@@ -1,40 +1,23 @@
...
-  afterEach(async () => {
-    await resetSession(pageObject)
-  })
+  const ctx = createTestContext(/* clearDb= */ false)
 
   describe('single checkbox question', () => {
-    let applicantQuestions: ApplicantQuestions
     const programName = 'test program for single checkbox'
 
     beforeAll(async () => {
```
jeanniefu, line 17: not-blocking, and no action needed. just for discussion - In theory, we should be able to just change this to beforeEach for all the test files and set clearDB=true and that should work right? Do you know if there's a large performance hit?

jeanniefu (approved):
Thank you!

browser-test/src/checkbox.test.ts
nbeloglazov: Yes, replacing `beforeAll` with `beforeEach` will allow us to clean DB between tests. But as you said it does introduce performance hit and pretty nasty :/ 

I just wrote down some thoughts about that here: https://github.com/civiform/civiform/issues/2891#issuecomment-1235813976

browser-test/src/support/index.ts
nbeloglazov: Thanks. Done.

PR #3343, opened 2022-09-02
Publish modal shows edited questions and programs

### LINKED ISSUE #2114: Add summary of what is being published when an Admin Publishes All Programs
When Admins publish changes they must publish all pending changes.  The user flow for publishing however doesn't indicate what draft changes are being published though.

clouser2:
### Description

Implements the mocked dialog from #2114 that shows the list of questions / programs that will be published when clicking "Publish all drafts". Additionally, adds browser test coverage of the new dialog and adds screenshot coverage.

I added support for taking a screenshot of a specific element rather than the whole page since this dialog appears as a modal on top of the programs page, which has timestamps.

While here, I also made ActiveAndDraftPrograms a bit more consistent in initialization with ActiveAndDraftQuestions and added notes that it actually has N+1 query behavior in order to sync question definitions. I think we might be able to use the "simulated publish" Version that's used in ActiveAndDraftQuestions, but that's out of scope for this bug.

### Screenshots
Screen Shot 2022-09-06 at 8 55 47 AM

## Release notes:

When CiviForm admins publish drafts, surface a list of the questions / programs that will be edited.
 
### Checklist

### Issue(s)

Fixes #2114


clouser2:
@rahaghas If you get a chance, can you give feedback on if the content in the screenshot is acceptable for this? I'm happy to address feedback prior to submission or as a fast follow :).

rahaghas (approved):

clouser2:
> Maybe just change "Edited" to "Draft" to be consistent with our language in other spots. And let's change the title language to "All draft programs will be published" questions are only published as part of programs, so I don't this we need to say that.

Incorporated and screenshots updated!

browser-test/src/civiform_admin_publish.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  let adminPrograms: AdminPrograms
+  let adminQuestions: AdminQuestions
+
+  const programNoQuestions = 'publish-test-program-no-questions'
+  const programWithQuestion = 'publish-test-program-with-question'
+  const questionName = 'publish-test-address-q'
+
+  beforeAll(async () => {
+    const {page} = await startSession()
+    pageObject = page
```
bion, line 22: question (blocking): why reassign?

bion (changes requested):

browser-test/src/civiform_admin_publish.test.ts
clouser2: startSession returns an object containing multiple values. Presently, the variable we use to track the page (similar to other tests) is called pageObject. We need to store the result off in a single variable, then assign it to the variable.

bion: could you rename the variable defined in this file to `page` and edit L21 to be `{page} = await startSession()`?

clouser2: Unfortunately no. That turns out to be a compiler error:
```
src/civiform_admin_publish.test.ts:21:12 - error TS2809: Declaration or statement expected. This '=' follows a block of statements, so if you intended to write a destructuring assignment, you might need to wrap the the whole assignment in parentheses.
```

Switching the statement to:

```
({page} = await startSession()
```

works. That said, that pattern is not obvious to me and I think it might be better to be consistent with what we do elsewhere:

```
const session = await startSession()
pageObject = session.page

bion (approved):

PR #3354, opened 2022-09-06
Attach aria labels for descriptions and errors for applicant questions

### LINKED ISSUE #1879: Question input fields missing `aria-describedby` attribute referencing the question description

### LINKED ISSUE #3356: Change applicant question div IDs to use a string that does not contain square brackets and periods
Currently, the HTML id for applicant question divs use the contextualized path as a string, which could contain square brackets and periods: [code link](https://github.com/civiform/civiform/blob/main/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java#L137).

According to the note [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) it is a best practice for IDs to only use letters, numbers, -, and _.

jeanniefu:
### Description

For composite questions, question level errors/descriptions are associated with the fieldset that wraps the inputs, and input level errors are associated with the specific input. For non-composite questions (e.g. with a single input field), we associate the question level errors/description with the input directly.

I turned on my screen reader and manually verified all changes.

File upload error requires a client-side change, which will be done in a followup PR.

## Release notes:

[a11y] Read out applicant question errors and descriptions for screen reader users.

### Checklist

### Issue(s)

#1878, fixes #1879, fixes #3356


jeanniefu:
My original super old PR is here: https://github.com/civiform/civiform/pull/2596. I tried to address all the relevant feedback there with this new approach.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -36,10 +37,19 @@ protected enum InputFieldType {
...
+  // HTML id tags for various elements within this question.
+  private final String id;
+  private final String descriptionId;
+  private final String requiredErrorId;
+  private final String validationErrorId;
 
   ApplicantQuestionRendererImpl(ApplicantQuestion question, InputFieldType inputFieldType) {
     this.question = checkNotNull(question);
     this.inputFieldType = checkNotNull(inputFieldType);
+    id = question.getContextualizedPath().toString();
```
bion, line 49: suggestion (blocking): can we convert this to a value that doesn't contain brackets and periods? according to the note [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) it is a best practice for IDs to only use letters, numbers, `-`, and `_`

bion (changes requested):
This is a huge improvement, great work @jeanniefu! just a few small things

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Yes, that sounds good! This is actually existing behavior (I just moved it from line 137 to here), so I think it doesn't need to be blocking (unless you have concerns about this PR adding more IDs that don't follow that guideline).

I filed https://github.com/civiform/civiform/issues/3356 and can send out a separate PR for that fix.

jeanniefu: Actually based on Brett's suggestion in the bug to use a random ID (which sounds good to me), I'll just update here since that's a one line change.

bion (approved):

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -48,16 +59,22 @@ private String getRequiredClass() {
 
   protected abstract DivTag renderTag(
       ApplicantQuestionRendererParams params,
-      ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> validationErrors);
+      ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> validationErrors,
+      ImmutableList<String> ariaDescribedByIds,
+      boolean hasErrors);
```
clouser2, line 64: I'm not following why `hasErrors` is needed. Isn't the presence of any key in the `validationErrors` map a sufficient signal that any errors exist?

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -48,16 +59,22 @@ private String getRequiredClass() {
...
 
   @Override
   public final DivTag render(ApplicantQuestionRendererParams params) {
+    boolean hasErrors = false;
+    ImmutableList.Builder<String> ariaDescribedByBuilder = ImmutableList.builder();
```
clouser2, line 69: Not blocking: You can chain the initial addition:

```
ImmutableList.Builder<String> ariaDescribedByBuilder = ImmutableList.builder()
  .add(descriptionId);
```

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -36,10 +38,19 @@ protected enum InputFieldType {
 
   protected final ApplicantQuestion question;
   private final InputFieldType inputFieldType;
+  // HTML id tags for various elements within this question.
+  private final String id;
```
clouser2, line 42: Not blocking: I'd consider qualifying this as `questionId` or something along those lines.

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -71,11 +75,12 @@ protected DivTag renderTag(
                             .setFieldName(addressQuestion.getCityPath().toString())
                             .setLabelText(messages.at(MessageKey.ADDRESS_LABEL_CITY.getKeyName()))
                             .setValue(addressQuestion.getCityValue().orElse(""))
-                            .addReferenceClass(ReferenceClasses.ADDRESS_CITY)
                             .setFieldErrors(
                                 messages,
                                 validationErrors.getOrDefault(
                                     addressQuestion.getCityPath(), ImmutableSet.of()))
+                            .setAriaInvalid(hasErrors)
```
clouser2, line 82: I'm a bit confused here and elsewhere around why `hasErrors` is piped around and why we're setting the aria-invalid attribute here based on a global concept of whether errors exist. Wouldn't this only be invalid if the above argument to `setFieldErrors` was non-empty?

If that's the case, is it possible we could bundle up all of the logic for setting this attribute into FieldWithLabel and transparently set the aria-invalid attribute if the field errors are non-empty? That'd also allow us to remove the `setAriaInvalid` method from the class.

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -48,6 +51,7 @@ protected DivTag renderTag(
                         messages,
                         validationErrors.getOrDefault(
                             addressQuestion.getStreetPath(), ImmutableSet.of()))
+                    .setAriaInvalid(hasErrors)
```
clouser2, line 54: Is there a missing call to add the described by IDs here and in other renderers?

clouser2 (changes requested):

server/app/views/questiontypes/AddressQuestionRenderer.java
clouser2: Oh, I think I had missed that this would be considered a "COMPOSITE" field and the described by would be associated with the top level field? If so, I wonder if we should have separate methods for rendering composite fields versus single fields since the described by IDs would be extraneous here.

I could also see this being a case where we'd add some inheritance in order to make it explicit whether a question type was composite or singular. As it stands, it's a bit hard to follow why the parameter is ignored in some question types and used in others.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Added a lengthier comment in the other thread

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
clouser2: I think this can be removed by also adding the type annotation to the builder method (e.g.):

```
ImmutableList.Builder<String> ariaDescribedByBuilder = ImmutableList.<String>builder()
  .add(...);
```

server/app/views/questiontypes/AddressQuestionRenderer.java
clouser2: For the purposes of this PR, I'm definitely up for any of the options. My general feeling is that separate classes (e.g. option #3) might be the most maintainable long-term since it has very clearly defined boundaries and hopefully no unused code / calls.

server/app/views/questiontypes/AddressQuestionRenderer.java
clouser2: I believe top-level question errors are surfaced as an entry in the map at the top-level question's path. We extract that [here](https://github.com/civiform/civiform/blob/main/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java#L82-L97). If a "required" question doesn't end up surfacing an error there, that seems like a bit of debt that we could clean up. If we wanted to separate it from this PR, we could perhaps create a copy of the map with a required error and pass it along to downstream callers so that they could continue to use the map as the "source-of-truth".

> My impression was that validationErrors held a map for validation errors across all questions. Is it actually per question? 

I believe it's a map constructed based on a specific question instance ([code](https://github.com/civiform/civiform/blob/29dd8296eefb3e59aad3ad97cb41fa4d9e6a5ef9/server/app/services/applicant/question/Question.java#L51)).

Re: `setAriaInvalid`, sounds good to me!

server/app/views/questiontypes/AddressQuestionRenderer.java
jeanniefu: Agreed that (3) sounds like the most correct approach. I'll send out a separate PR to refactor so this one stays smaller.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
jeanniefu: Thank you! Done.

server/app/views/questiontypes/AddressQuestionRenderer.java
jeanniefu: In that first link you provided, it actually uses [isRequiredButWasSkippedInCurrentProgram()](https://github.com/civiform/civiform/blob/main/server/app/services/applicant/question/ApplicantQuestion.java#L108:18), which uses various properties of the question object instead of the map. I agree though that we should do a cleanup so that it's represented in that error map. Seems confusing that the error logic is fragmented. I created #3393, and will do a fast follow fix.

jeanniefu:
I'm gonna submit this now so it's fixed before our a11y audit starts. I'll send fast follow up fixes to address Brett's comments. Thanks for the thorough review!

PR #3358, opened 2022-09-07
Status Tracking: Default the application note editor to the existing note

shanemc-goog:
### Description

Reads the latest note value for an Application and seeds the note editor with that content.

## Release notes:

Status Tracking: Shows the current application note in the note editor.

### Checklist

### Issue(s)

### DIFF
--- a/server/app/models/Application.java
+++ b/server/app/models/Application.java
@@ -10,6 +10,7 @@ import java.util.Optional;
 import javax.persistence.Entity;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
+import javax.persistence.OrderBy;
 import javax.persistence.Table;
 import play.data.validation.Constraints;
 import services.applicant.ApplicantData;
@@ -34,6 +35,7 @@ public class Application extends BaseModel {
   @ManyToOne private Program program;
 
   @OneToMany(mappedBy = "application")
+  @OrderBy("createTime desc")
   private List<ApplicationEvent> applicationEvents;
 
   @Constraints.Required private LifecycleStage lifecycleStage;
--- a/server/test/services/applications/ProgramAdminApplicationServiceTest.java
+++ b/server/test/services/applications/ProgramAdminApplicationServiceTest.java
@@ -74,13 +74,12 @@ public class ProgramAdminApplicationServiceTest extends ResetPostgres {
     String note = "Application note";
     ProgramDefinition program = ProgramBuilder.newActiveProgram("some-program").buildDefinition();
 
-    Account account = resourceCreator.insertAccount();
     Applicant applicant = resourceCreator.insertApplicantWithAccount();
     Application application =
         Application.create(applicant, program.toProgram(), LifecycleStage.ACTIVE)
             .setSubmitTimeToNow();
 
-    // Execute
+    // Execute, verify.
     assertThat(service.getNote(application)).isEmpty();
   }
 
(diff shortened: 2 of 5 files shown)


shanemc-goog (commented):
baseline PR

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -161,7 +161,34 @@ describe('view program statuses', () => {
...
+      await adminPrograms.expectNoteUpdatedToast()
+
+      // Reload the note editor.
+      await adminPrograms.viewApplications(programWithStatusesName)
+      await adminPrograms.viewApplicationForApplicant('Guest')
+
+      expect(await adminPrograms.getNoteContent()).toBe(noteText)
+    })
+
+    it('allows updating a note', async () => {
```
clouser2, line 179: Not blocking: I'm not sure having two distinct tests here really has more value than combining them into a single test that:
  1) Edits the note
  2) Confirms that the displayed content matches the edit
  3) Edit the note again
  4) Confirms that the updated content matches the edit

As it stands, the first test is already testing the "update" case as part of test setup.

server/app/models/Application.java
```diff
@@ -34,6 +35,7 @@ public class Application extends BaseModel {
   @ManyToOne private Program program;
 
   @OneToMany(mappedBy = "application")
+  @OrderBy("createTime desc")
```
clouser2, line 38: Not blocking: Would it make sense to add an index on createTime?

server/test/services/applications/ProgramAdminApplicationServiceTest.java
```diff
@@ -66,4 +68,36 @@ public void getApplication_emptyAdminName() {
...
+    ProgramDefinition program = ProgramBuilder.newActiveProgram("some-program").buildDefinition();
+
+    Account account = resourceCreator.insertAccount();
+    Applicant applicant = resourceCreator.insertApplicantWithAccount();
+    Application application =
+        Application.create(applicant, program.toProgram(), LifecycleStage.ACTIVE)
+            .setSubmitTimeToNow();
+
+    service.setNote(application, NoteEvent.create("first note"), account);
+    service.setNote(application, NoteEvent.create(note), account);
```
clouser2, line 97: I *think* you might end up with a potentially flaky test here since EBean serializes timestamps truncated to millisecond precision. When I was developing similar tests, they passed on my machine, then failed in CI (presumably since the tests ran faster).

If you want to account for this in the test and avoid that type of flake, you could add a sleep for 2ms. I don't really think that would have a dramatic impact on our test runtime and would hopefully reduce the flake rate.

Context: https://github.com/civiform/civiform/blob/08868c0bc54327e3ed4289c9827ba5c471f109e8/server/test/models/QuestionTest.java#L210-L215

clouser2 (approved):

server/app/models/Application.java
shanemc-goog: I hadn't thought about it either way.  Given we don't search on this and I don't expect there to be many events per application I lean towards an index being wasteful than much benefit.

thoughts?

server/test/services/applications/ProgramAdminApplicationServiceTest.java
shanemc-goog: done!

browser-test/src/admin_application_statuses.test.ts
clouser2: Your argument is definitely valid. My main reaction was that I saw the two tests that had almost all of the same code and it wasn't immediately obvious to me what the difference between the two of them was until I read closer.

server/app/models/Application.java
clouser2: I'm indifferent. Just thought I'd point it out here as an area for optimization.

clouser2 (approved):

Issue #3029, opened 2022-08-02
CiviForm Admin UX: order question bank by recently updated


cassandralfernandes:
**Is your feature request related to a problem? Please describe.**
The CiviForm admin is overwhelmed by the long list of questions in the question bank in the questions tab

**Describe the solution you'd like**
The default list should be ordered by recently updated questions 

Add description/label saying that the list is sorted by most recently updated 

**Additional context**
From admin UXR feedback sheet

clouser2:
We only recently added timestamps for when a question was most recently updated. As such, this won't be an issue for new deployments, but may be something we need to deal with for existing deployments. What order should we use when the time a question was last updated can't be determined?

azizova-leyla:
@cassandralfernandes do you have a mock for the "sorted by most recently updated" string?

cassandralfernandes:
The default list should be ordered as "recently added" > "recently used" > all others

Mock [here](https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13ec4604693_0_147)

azizova-leyla:
Mock:

![Image](https://user-images.githubusercontent.com/1741747/185443714-1a00b436-1ede-4461-a3d6-ebb9b1432b62.png)

azizova-leyla:
@cassandralfernandes to clarify, do you want to reorder questions in both All questions view and the Question bank?

cassandralfernandes:
Yes, we should reorder in both

clouser2:
**Question list**

* @nbeloglazov will be adding the "last updated" timestamp to the list as part of #3033, so I'll start on this assuming it exists
* We sort programs by "last modified" time as well and do this client side since we eventually plan to support more filters. That's probably the case here as well per the mock, so I'll do client-side sorting similar to what was described in https://github.com/civiform/civiform/issues/1238#issuecomment-1127884289

**Question bank**

There's no set question order here as well. We already have a search box that filters the questions down. The work will mainly be to augment that sorting to use the same logic as the question list. Likely it will make the most sense to pull the relevant JS out into a helper used by the program list, question list, and question bank.

clouser2:
A barrier to sharing a helper utility is that we don't really have true javascript modules where we can change our TypeScript code to import a utility from another file. I'll explore if there's anything we can do here.

PR #3359, opened 2022-09-07
Sort question bank / question list questions by most recently modified

clouser2:
### Description

This makes use of the sorting mechanism added for the program list and applies it to both the question bank and question list. Additionally, a "sorted by" chip was added to the question / program list (screenshots below).

Sorting is done client-side per the rationale in https://github.com/civiform/civiform/issues/3029#issuecomment-1238371242. This approach was generalized and moved to `sorting.ts`. `sorting.ts` registers all of the elements on which it will sort rather than providing a general helper utility since we don't currently have the ability to use JavaScript modules (https://github.com/civiform/civiform/issues/3029#issuecomment-1238399530).

Finally, the "placeholder" mechanism was removed since it added complexity and wasn't actually working for the program list anyway.

### Screenshots

Question list before:
Screen Shot 2022-09-07 at 12 49 27 PM

Question list after:
Screen Shot 2022-09-07 at 12 46 17 PM

Program list before:
Screen Shot 2022-09-07 at 12 49 21 PM

Program list after:
Screen Shot 2022-09-07 at 12 46 12 PM

## Release notes:

Sorting the question list and questions shown in the program builder by the most recently updated question.

### Checklist

### Issue(s)

#3029


browser-test/src/civiform_admin_question_list.test.ts
```diff
@@ -0,0 +1,111 @@
+import {
+  AdminPrograms,
+  AdminQuestions,
+  createTestContext,
+  loginAsAdmin,
+} from './support'
+
+describe('Most recently updated question is at top of list.', () => {
```
clouser2, line 8: Note to the reviewer: This test was modeled off of the corresponding admin_program_list.test.ts (https://github.com/civiform/civiform/blob/08868c0bc54327e3ed4289c9827ba5c471f109e8/browser-test/src/admin_program_list.test.ts)

browser-test/src/civiform_admin_question_list.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    const {page, adminPrograms, adminQuestions} = ctx
+
+    await loginAsAdmin(page)
+
+    const questionOne = 'question list test question one'
+    const questionTwo = 'question list test question two'
+    await adminQuestions.addNameQuestion({questionName: questionOne})
+    await adminQuestions.addNameQuestion({questionName: questionTwo})
+
+    // Note: CI tests already have test questions
```
nbeloglazov, line 20: Is that true? Each test file starts with empty database. And if you have emptry database you can assume only your questions are present and simplify assertions below.

browser-test/src/support/admin_programs.ts
```diff
@@ -253,6 +253,14 @@ export class AdminPrograms {
     }
   }
 
+  async questionBankNames(programName: string) {
```
nbeloglazov, line 256: Add return type `string[]` given that it's public function.

browser-test/src/support/admin_questions.ts
```diff
@@ -244,6 +244,14 @@ export class AdminQuestions {
     )
   }
 
+  async questionNames() {
```
nbeloglazov, line 247: Same here.

server/app/assets/javascripts/sorting.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  // Question bank.
+  const questionBankParent = document.getElementById('question-bank-questions')
+  if (questionBankParent) {
+    ElementSorter.sortElementsOnLoad({
+      parent: questionBankParent,
+      compareFunc: lastUpdatedAndNameComparator,
+    })
+  }
```
nbeloglazov, line 89: Looks like all these 3 blocks are mostly the same except for the selector. Maybe do process them in a loop?

```
for (const selector of ['.cf-admin-question-list', ...]) {
  const el = document.querySelector(selector);
  if (!el) continue;
  ...
}
```

server/app/assets/javascripts/sorting.ts
```diff
@@ -0,0 +1,90 @@
...
+  }) {
+    const elements = Array.from(
+      parent.querySelectorAll(ElementSorter.sortableElementSelector),
+    )
+    try {
+      ElementSorter.sortCards(parent, elements, compareFunc)
+    } finally {
+      // Make sure to always show the element, even
+      // if there was an error while sorting.
+      parent.classList.remove('invisible')
```
nbeloglazov, line 22: That makes me a bit nervous that we rely on JS to make element visible. So if JS is slow to load user will see empty page. I think ideally we'd want to sort server-side and add ability to re-sort client-side if necessary. But it's more work to maintain both server-side and client-side sorting... Do we expect other client-side sorting options to be added soon?

nbeloglazov (changes requested):

clouser2:
This surfaced some existing issues where we weren't properly awaiting scripts after clicking a button which would trigger a navigation. While I rework things, I'll take it out of review.

clouser2 (commented):
Replying to comments for posterity. This PR is being dropped in favor of https://github.com/civiform/civiform/pull/3431 (where all feedback has been incorporated). We'll now do sorting server-side until we have a real need for client-side sorting.

PR #3364, opened 2022-09-07
fixing Create-new-question dropdown (as the scroll bar isn't working)

### LINKED ISSUE #3365: create-new-question is not scrollable
**Describe the bug**
create-new-question is not scrollable

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

swatkat1:
### Description

Dropdown for "create-new-question" doesn't have a scroll option. So Admins cannot choose other options.

## Release notes:

Dropdown for "create-new-question" doesn't have a scroll option. So Admins cannot choose other options.

### Checklist

### Issue(s)

Fixes #3365 ; 

Screenshots
Before:
image

After:
image

### DIFF
--- a/server/app/views/components/CreateQuestionButton.java
+++ b/server/app/views/components/CreateQuestionButton.java
@@ -35,7 +35,7 @@ public final class CreateQuestionButton {
                 Styles.BG_WHITE,
                 Styles.TEXT_GRAY_600,
                 Styles.SHADOW_LG,
-                Styles.OVERFLOW_AUTO,
+                Styles.ABSOLUTE,
                 Styles.ML_3,
                 Styles.MT_1,
                 Styles.HIDDEN);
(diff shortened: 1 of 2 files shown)


server/app/views/components/CreateQuestionButton.java
```diff
@@ -35,7 +35,7 @@ public static DivTag renderCreateQuestionButton(String questionCreateRedirectUrl
                 Styles.BG_WHITE,
                 Styles.TEXT_GRAY_600,
                 Styles.SHADOW_LG,
-                Styles.ABSOLUTE,
```
clouser2, line 38: I *believe* that the "absolute" class is required for positioning the dropdown appropriately next to the content.

I think the source of the problem is actually that we set 'overflow-hidden' in `QuestionBank.questionBankPanel`. Do you mind restoring the "absolute" tag here and removing the "overflow-hidden" tag in `QuestionBank.questionBankPanel`? If that works, there should be no need to set "overflow-auto".

clouser2 (changes requested):

server/app/views/components/CreateQuestionButton.java
swatkat1: that works too!

clouser2 (approved):

PR #3367, opened 2022-09-07
Docker files for civiform/civiform-cloud-deployment

azizova-leyla:
### Description

This creates Docker image for cloud/ directory + a command to run it.
Tested with AWS, but not Azure yet.

Client facing scripts will be wrapping bin/run-cloud most likely, currently for convenience of testing.

## Release notes:

### Checklist

### Issue(s)

Part of #3222


bin/run-cloud
```diff
@@ -0,0 +1,15 @@
...
+# bin/build-cloud
+# bin/run-cloud <command> <tag> <config_file>
+
+export COMMAND="${1:-deploy}"
+export IMAGE_TAG="${2:-latest}"
+export CONFIG="${3:-cloud/deploys/dev_aws/civiform_config.sh}"
+export AWS_ACCESS_KEY_ID=$(aws --profile default configure get aws_access_key_id)
+export AWS_SECRET_ACCESS_KEY=$(aws --profile default configure get aws_secret_access_key)
+echo "Running ${COMMAND} ${IMAGE_TAG} ${CONFIG}"
+docker compose -f cloud/docker-compose.yaml up
```
nbeloglazov, line 15: Will deploy scripts need to check out civiform repo at all? Or they can just run `docker run` and use the cloud image?

azizova-leyla: It will not need to checkout main repo, docker-compose file and scripts to run it will be in civiform-deploy repo itself. 
It's convenient to have it in the main repo for testing for now.

nbeloglazov (approved):

bin/build-cloud
```diff
@@ -0,0 +1,35 @@
+ #! /usr/bin/env bash
+
+ # DOC: Build civiform/cloud.
```
kmcnellis, line 3: Nit: there's an extra space before these lines at the top of the file.  I think we don't want that

bin/build-cloud
```diff
@@ -0,0 +1,35 @@
...
+
+IMAGE="cloud"
```
kmcnellis, line 5: I feel like "cloud" isn't the best name for this.  It's generic, and doesn't really describe what the image does (i.e. `docker run cloud` means almost nothing)

bin/build-cloud
```diff
@@ -0,0 +1,35 @@
...
+docker buildx build --load "${BUILD_ARGS[@]}"
+
+if [[ "${PUSH_IMAGE}" ]]; then
+  docker::do_dockerhub_login
+  # Push the image from the cache to dockerhub
+  echo "push ${IMAGE} build"
+  docker buildx build --push "${PLATFORM_ARG[@]}" "${BUILD_ARGS[@]}"
+fi
+
+docker tag "civiform/${IMAGE}:latest" "${IMAGE}:latest"
```
kmcnellis, line 35: If we do use something generic (i.e. cloud) as the name, we need to use the name `civiform-cloud` as the local name, so it doesn't conflict with any public packages

bin/run-cloud
```diff
@@ -0,0 +1,15 @@
+ #! /usr/bin/env bash
```
kmcnellis, line 1: Same here with the extra spaces

cloud/cloud.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+FROM eclipse-temurin:11.0.16_8-jre-alpine as amd64
+FROM bellsoft/liberica-openjre-alpine:11.0.16-8 as arm64
+
+FROM ${TARGETARCH}
+
+COPY --from=hashicorp/terraform:1.2.8 /bin/terraform /usr/local/bin/
+COPY --from=amazon/aws-cli:2.7.27 /usr/local /usr/local
+COPY --from=amazon/aws-cli:2.7.27 /aws /aws
+# COPY --from=mcr.microsoft.com/azure-cli:2.39.0 /usr/local /usr/local
```
kmcnellis, line 15: Why is this commented out?  Let's leave a TODO, or just remove the line

bin/run-cloud
kmcnellis, line 15: There's some arguments you can add to remove and shut down the container once the commands are done.  Might be a good idea here

cloud/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - AWS_REGION
+      - AZURE_CLIENT_ID
+      - AZURE_STORAGE_ACCOUNT_NAME
+      - AZURE_LOCAL_CONNECTION_STRING
+      - AZURE_STORAGE_ACCOUNT_CONTAINER
+      - CONFIG
+      - IMAGE_TAG
+      - COMMAND
+    volumes:
+      - ${PWD}/${CONFIG}:/${CONFIG}:ro
```
kmcnellis, line 17: "ro"? Let's add a comment for that

bin/run-cloud
kmcnellis: Add a TODO for that?

kmcnellis (changes requested):

bin/run-cloud
azizova-leyla: couldn't find arguments, but I addded rm command right after this.

azizova-leyla: I don't think todo is necessary here. That's a separate task I'm planning to do.

bin/build-cloud
azizova-leyla: renamed. if you have better suggestions - let me know, naming is hard.

bin/run-cloud-deployment
```diff
@@ -0,0 +1,16 @@
+#! /usr/bin/env bash
+
+# DOC: Run deploy scripts from Docker.
+
+# Usage:
+# bin/build-cloud
+# bin/run-cloud <command> <tag> <config_file>
```
bion, line 7: suggestion (blocking): update the comment here with the actual names of the script files

bin/run-cloud-deployment
```diff
@@ -0,0 +1,16 @@
...
+
+export COMMAND="${1:-deploy}"
+export IMAGE_TAG="${2:-latest}"
+export CONFIG="${3:-cloud/deploys/dev_aws/civiform_config.sh}"
+export AWS_ACCESS_KEY_ID=$(aws --profile default configure get aws_access_key_id)
+export AWS_SECRET_ACCESS_KEY=$(aws --profile default configure get aws_secret_access_key)
+echo "Running ${COMMAND} ${IMAGE_TAG} ${CONFIG}"
```
bion, line 14: suggestion (non-blocking): instead of echoing and then immediately executing, maybe confirm first?

bin/run-cloud-deployment
```diff
@@ -0,0 +1,16 @@
...
+docker compose -f cloud/docker-compose.yaml up
+docker compose -f cloud/docker-compose.yaml rm -fsv
```
bion, line 16: question (blocking): what do folks think of moving these scripts into `cloud/shared/bin`?

bion (changes requested):

cloud/shared/bin/run-cloud-deployment
```diff
@@ -0,0 +1,25 @@
...
+
+export COMMAND="${1:-deploy}"
+export IMAGE_TAG="${2:-latest}"
+export CONFIG="${3:-cloud/deploys/dev_aws/civiform_config.sh}"
+if [[ -z "${AWS_ACCESS_KEY_ID}" ]]; then
+  export AWS_ACCESS_KEY_ID=$(aws --profile default configure get aws_access_key_id)
+fi
+if [[ -z "${AWS_SECRET_ACCESS_KEY}" ]]; then
+  export AWS_SECRET_ACCESS_KEY=$(aws --profile default configure get aws_secret_access_key)
+fi
```
bion, line 17: suggestion (blocking): this only works with AWS, please check cloud provider first and branch from there

bion (changes requested):

cloud/shared/bin/run-cloud-deployment
azizova-leyla: we require AWS cli for both Azure and AWS setups. None of this works for Azure, that's coming separately.

bin/build-cloud-deployment
```diff
@@ -0,0 +1,35 @@
...
+  --cache-from "civiform/${IMAGE}"
+  --build-arg BUILDKIT_INLINE_CACHE=1
+  "${LOCATION}")
+
+PLATFORM_ARG=()
+if [[ "${PLATFORM}" ]]; then
+  PLATFORM_ARG=(--platform "${PLATFORM}")
+fi
+
+# Build the multi-platform image
```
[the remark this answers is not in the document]
azizova-leyla: I copied this from the way we build all of our docker images, @kmcnellis would know why exactly we're doing this.

bion (approved):

kmcnellis (approved):

Issue #3361, opened 2022-09-07
Enable type-based linting for TS code


nbeloglazov:
Currently eslint doesn't type-based checks so it's not fully enabled. Enabling it will allow us to catch errors like missing `await` on async functions.

PR #3368, opened 2022-09-07
Improve typing of TS browser tests

nbeloglazov:
### Description

Add type definition for the application JSON object that can be downloaded by program admins. It doesn't guarantee that type definition doesn't become stale but at least it provides TS code with better centralized type checking. Needed for #3361

### Checklist


clouser2 (approved):
Nice!

PR #3369, opened 2022-09-08
Rework eslint structure and enable type-based checks

nbeloglazov:
### Description

With the new approach each directory that contains TS code (`browser-test` or `server`) has its own `.eslintrc.json` file which extends base config in `formatter/eslint_base.json`. Because of that we need to run eslint twice: once for `browser-test` and once for `server`. 

I also re-did formatter docker image to pre-fetch nodjs modules for browser-test and server directories as they are needed for enabling type-based eslint checks. 

I removed eslint checks from JS files because configuring proper checks for projects that has mix of JS and TS is challenging and we don't have many JS files anyway, mostly configs. 

### Checklist

### Issue(s)

Issue #3361


formatter/eslintrc_base.json
```diff
@@ -0,0 +1,40 @@
...
+    "browser": true,
+    "commonjs": true
+  },
+  "plugins": ["@typescript-eslint"],
+  "rules": {
+    "require-jsdoc": "off",
+    "valid-jsdoc": "warn",
+    "@typescript-eslint/no-unused-vars": "off", // overlaps with standard eslint no-unused-vars.
+
+    // TODO(#3361): fix and reenable
```
kmcnellis, line 22: Which plugin is this TODO for?

formatter/fmt
```diff
@@ -47,9 +47,13 @@ npx prettier \
 echo 'Done formatting with prettier'
 
 echo 'Start formatting with eslint'
-npx eslint \
-  --fix \
-  {server/app/assets/javascripts,server/tailwind.config.js,test-support,browser-test}
+cd server
+npm install --silent
```
kmcnellis, line 51: Can we do also make sure to do these installs inside the dockerfile, so it's a one-time download (we would only need it here if the package.json is mounted at runtime and may change without a new docker build)

formatter/formatter.Dockerfile
```diff
@@ -16,15 +16,45 @@ RUN apk update && apk add --no-cache --update \
...
 
-COPY formatter/fmt /fmt
-VOLUME /code
+# Fetch node js dependencies for `browser-test` directory.
+ENV BROWSER_TEST_DIR /code/browser-test
+RUN mkdir -p $BROWSER_TEST_DIR
+COPY browser-test/package.json $BROWSER_TEST_DIR
+COPY browser-test/yarn.lock $BROWSER_TEST_DIR
+WORKDIR $BROWSER_TEST_DIR
+RUN npm install --global yarn
```
kmcnellis, line 42: This `npm install --global ` should go at the top, so it gets cached in a layer that doesn't change when the package.json files change

formatter/fmt
```diff
@@ -47,9 +47,13 @@ npx prettier \
...
+npx eslint --fix "app/assets/javascripts/**/*.ts"
+cd ../browser-test
+yarn install --silent
```
kmcnellis, line 54: [future improvement] I'd love to see us move to only use one-of yarn or npm

kmcnellis (approved):

formatter/eslintrc_base.json
nbeloglazov: All of them. I'm planning to reenable them one by one.

formatter/fmt
nbeloglazov: Yeah. It's annoying to have both.

formatter/formatter.Dockerfile
nbeloglazov: Done.

formatter/fmt
nbeloglazov: It also runs `npm install` in the dockerfile. So here it will most likely reuse already downloaded files. Running npm intall here is needed because lint checks need to have dependencies present in order to do type-based checks. So when engineer updates `package.json` locally - we'll need to run `npm install` during formatting for eslint to properly account for the new/changed dependencies.

PR #3377, opened 2022-09-08
Send status emails when a new status is set

shanemc-goog:
### Description

When a Program admin sets a status, sends the corresponding email if it exists and the admin requested it be sent.

Similar to ApplicantService, this sends emails to Staging lists when in staging mode, and sends TIs a different notice if present on the Application.

## Release notes:

Status Tracking: sends emails when a status is set on an Application.

### Checklist

### Issue(s)


server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -3,27 +3,58 @@
...
 
 /** The service responsible for mediating a program admin's access to the Application resource. */
 public final class ProgramAdminApplicationService {
+  static final String STATUS_UPDATE_EMAIL_SUBJECT_FORMAT = "An update on your application for: %s";
+
+  private final ApplicantService applicantService;
   private final ApplicationRepository applicationRepository;
   private final ApplicationEventRepository eventRepository;
+  private final SimpleEmail emailClient;
+  private final String baseUrl;
```
shanemc-goog, line 34: These 4 values are copied right out of ApplicantService

https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/services/applicant/ApplicantService.java?L62

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -53,16 +84,90 @@ public Optional<Application> getApplication(long applicationId, ProgramDefinitio
...
+      String programName, Applicant applicant, Status statusDef, Optional<String> applicantEmail) {
+    String civiformLink = baseUrl;
+    Locale locale = applicant.getApplicantData().preferredLocale();
+    String emailBody =
+        String.format(
+            "%s\n\nLog in to CiviForm at %s.",
+            statusDef.localizedEmailBodyText().get().getOrDefault(locale), civiformLink);
+    emailClient.send(
+        isStaging ? stagingApplicantNotificationMailingList : applicantEmail.get(),
+        String.format(STATUS_UPDATE_EMAIL_SUBJECT_FORMAT, programName),
```
clouser2, line 141: Blocking: The variable of `programName` looks to be correct here. However, it was initialized to `adminName`, which is the internal facing name of the program. I believe we should use `localizedName` for the current user's locale.

I think spotting these types of bugs might be clearer if the method took a `ProgramDefinition` as an argument and directly used `program.localizedName()...` since there's a single callsite. Passing a string feels a bit lossy to me (same commentary for the TI email below).

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -53,16 +84,90 @@ public Optional<Application> getApplication(long applicationId, ProgramDefinitio
...
+    String tiDashLink =
+        baseUrl
+            + controllers.ti.routes.TrustedIntermediaryController.dashboard(
+                    /* nameQuery= */ Optional.empty(),
+                    /* dateQuery= */ Optional.empty(),
+                    /* page= */ Optional.of(1))
+                .url();
+    String subject =
+        String.format(
+            "An update on the application for program %s on behalf of applicant %d",
```
clouser2, line 159: I'd either put the applicant name as the last argument or change the text a smidge since "on behalf of applicant 2" doesn't read very clearly to me.

server/app/services/applications/ProgramAdminApplicationService.java
clouser2, line 141: Can we create a bug for localizing the subject lines and leave a TODO here?

server/test/services/applications/ProgramAdminApplicationServiceTest.java
```diff
@@ -105,4 +164,192 @@ public void getNote_multipleNotes_findsLatest() throws Exception {
...
+    service =
+        new ProgramAdminApplicationService(
+            instanceOf(ApplicantService.class),
+            instanceOf(ApplicationRepository.class),
+            instanceOf(ApplicationEventRepository.class),
+            instanceOf(Config.class),
+            simpleEmail);
+
+    ProgramDefinition program =
+        ProgramBuilder.newActiveProgram("some-program")
```
clouser2, line 182: I'd update this test to use different names for the internal / external representation of the program name and assert that the subject contains the external representation (see comment in the corresponding service definition for the type of bug I'm thinking of).

server/test/services/applications/ProgramAdminApplicationServiceTest.java
```diff
@@ -105,4 +164,192 @@ public void getNote_multipleNotes_findsLatest() throws Exception {
...
+            .withStatusDefinitions(new StatusDefinitions(ORIGINAL_STATUSES))
+            .buildDefinition();
+    Account account = resourceCreator.insertAccount();
+    Applicant applicant = resourceCreator.insertApplicantWithAccount(Optional.of(userEmail));
+    Application application =
+        Application.create(applicant, program.toProgram(), LifecycleStage.ACTIVE)
+            .setSubmitTimeToNow();
+
+    StatusEvent event =
```
clouser2, line 191: Worth adding an additional assertion that the latest status event log matches this?

server/test/services/applications/ProgramAdminApplicationServiceTest.java
```diff
@@ -105,4 +164,192 @@ public void getNote_multipleNotes_findsLatest() throws Exception {
...
+        StatusEvent.builder().setEmailSent(adminRequestsEmail).setStatusText(status).build();
+
+    service.setStatus(application, event, account);
+
+    verify(simpleEmail, never()).send(anyString(), anyString(), anyString());
```
clouser2, line 353: Per my comment in the service class, can we add an assertion of the logged event, specifically for the (hopefully) degenerate cases of admin/user requesting an email is sent for either 1) an application with no submitter email (e.g. guest) and 2) no email is configured for the status definition.

In my opinion, situation #2 would be a runtime error and situation #1 would have a logged status event indicating no email was sent.

clouser2 (changes requested):

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -3,27 +3,58 @@
...
 import services.application.ApplicationEventDetails.NoteEvent;
 import services.application.ApplicationEventDetails.StatusEvent;
+import services.cloud.aws.SimpleEmail;
 import services.program.ProgramDefinition;
+import services.program.StatusDefinitions.Status;
+import services.program.StatusNotFoundException;
 
 /** The service responsible for mediating a program admin's access to the Application resource. */
 public final class ProgramAdminApplicationService {
+  static final String STATUS_UPDATE_EMAIL_SUBJECT_FORMAT = "An update on your application for: %s";
```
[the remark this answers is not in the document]
clouser2: Got it, thanks.

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -53,16 +84,90 @@ public Optional<Application> getApplication(long applicationId, ProgramDefinitio
    *
    * @param admin The Account that instigated the change.
    */
-  public void setStatus(Application application, StatusEvent newStatus, Account admin) {
+  public void setStatus(Application application, StatusEvent newStatusEvent, Account admin)
+      throws StatusNotFoundException {
+    Program program = application.getProgram();
```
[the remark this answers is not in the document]
clouser2: OK. This is likely one of those things where the boundaries between our different layers are pretty murky and there's no clear good answer.

server/app/services/applications/ProgramAdminApplicationService.java
clouser2: OK.

clouser2 (approved):

clouser2 (approved):

clouser2 (approved):

PR #3378, opened 2022-09-09
Redesign admin question list page

nbeloglazov:
### Description

1. Show question text and help text as main question info. Question name and description are now shown as "admin notes".
2. Add "draft" and "active" rows for questions. 
3. Render questions as a list rather than a table. 
4. Remove translated languages text.

## Release notes:

Redesign admin questions list page to display questions a list and provide clear active and draft versions for each question.

### Checklist

### Issue(s)

Issue #3033


server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -105,49 +111,31 @@ private DivTag renderSummary(ActiveAndDraftQuestions activeAndDraftQuestions) {
...
   }
 
-  /** Renders the full table. */
-  private Pair<TableTag, ImmutableList<Modal>> renderQuestionTable(
+  /**
+   * @return a pair. First element is list of DOM elements, each representing a row for a particular
+   *     question. Second element is collection of all modals that can be invoked from question
+   *     rows.
+   */
+  private Pair<ImmutableList<DomContent>, ImmutableList<Modal>> renderAllQuestionRows(
```
clouser2, line 119: As part of #3359, we'll want to attach a reference class to the container for all of the rows (https://github.com/civiform/civiform/pull/3359/files#diff-af6ae91491683e0fd29d6daa311d8748e93b651f7b67b4521e723b0090c4ca45R139). What do you think about having this return a DivTag rather than an ImmutableList<DomContent> in order to facilitate that? If you'd prefer to wait for the other PR to be merged, that works as well.

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -105,49 +111,31 @@ private DivTag renderSummary(ActiveAndDraftQuestions activeAndDraftQuestions) {
...
+    }
+    return Pair.of(rows.build(), modals.build());
   }
 
   /**
-   * Display this as a table row with all fields.
+   * Renders a row in the question list. The row contains question text, help text, active and draft
+   * versions, list of programs questions is used in and buttons to edit the question.
    *
    * <p>One of {@code activeDefinition} and {@code draftDefinition} must be specified.
```
clouser2, line 136: I think this comment is no longer relevant (and wasn't relevant prior to your PR either).

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -112,38 +123,26 @@ private ActiveAndDraftQuestions(Version active, Version draft, Version withDraft
    * Inspects the provided version and returns a map who's key is the question name and value is a
    * set of programs that reference the given question in this version.
    */
-  private static ImmutableMap<String, ImmutableSet<ProgramDefinition>> buildReferencingProgramsMap(
+  private static ImmutableMap<Long, ImmutableSet<ProgramDefinition>> buildReferencingProgramsMap(
       Version version) {
-    // Different versions of a question can have distinct IDs while still
```
clouser2, line 117: Can you help me better understand why we moved away from this model (e.g. was it giving incorrect results / counts)? I'm a bit wary since I'm not 100% sure when we would associate the new draft question with all the programs it's used in.

clouser2 (changes requested):

browser-test/src/application_review.test.ts
```diff
@@ -107,7 +107,6 @@ describe('Program admin review of submitted applications', () => {
     await adminQuestions.expectActiveQuestionExist('monthly-income-q')
```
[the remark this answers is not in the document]
nbeloglazov: I originally implemented the way you describe but Raha asked that admin notes will be under active/draft rows which adds that empty space.

browser-test/src/application_review.test.ts
```diff
@@ -107,7 +107,6 @@ describe('Program admin review of submitted applications', () => {
...
 
     await adminQuestions.goToViewQuestionPage('date-q')
-    await adminQuestions.expectViewOnlyQuestion('date-q')
```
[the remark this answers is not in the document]
nbeloglazov: There is no view-only mode. Clicking on row opens edit mode. The rationale is that they both provide same data. And it's unlikely that user by mistake will edit the question as they need to explicitly save it.

browser-test/src/support/admin_questions.ts
```diff
@@ -159,31 +150,29 @@ export class AdminQuestions {
...
-
-    expect(tableInnerText).toContain(questionName)
-    expect(tableInnerText).toContain(questionText)
-    expect(
-      await this.page.innerText(this.selectQuestionTableRow(questionName)),
-    ).toContain('Edit Draft')
+    const questionRowText = await this.page.innerText(
+      this.selectQuestionTableRow(questionName),
+    )
+    expect(questionRowText).toContain(questionText)
```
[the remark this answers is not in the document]
nbeloglazov: That assertion is implicit in `selectQuestionTableRow(...)` which selects only rows that contain questionName.

server/app/services/question/ActiveAndDraftQuestions.java
nbeloglazov: Originally I moved away from that model because I thought we need different counts for draft and active versions. Also mock had different counts. But now, after our discussion, I think it makes more sense to have the same count. Because conceptually whole question is assigned to the program. And you can't assign just draft or just active to a program. So not sure what exactly to do here. Maybe for now just keep same count and revert ActiveAndDraftQuestions changes?

browser-test/src/civiform_admin_question_references.test.ts
```diff
@@ -10,7 +10,7 @@ describe('view program references from question view', () => {
     await adminQuestions.addAddressQuestion({questionName})
     await adminQuestions.expectQuestionProgramReferencesText({
       questionName,
```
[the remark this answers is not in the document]
nbeloglazov: Done.

server/app/views/admin/questions/QuestionsListView.java
nbeloglazov: Done.

browser-test/src/support/admin_questions.ts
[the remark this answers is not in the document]
clouser2: Sounds goo.

browser-test/src/application_review.test.ts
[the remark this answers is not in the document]
clouser2: Gotcha, works for me.

browser-test/src/application_review.test.ts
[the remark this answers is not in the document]
clouser2: Gotcha.

server/app/services/question/ActiveAndDraftQuestions.java
clouser2: Per @jeanniefu 's feedback, I think we can address the "referencing programs" counts issue separately.

Given that, if it's not too much work, I'd vote for keeping the same count right now.

server/app/views/admin/questions/QuestionsListView.java
clouser2: Specifically, I mean the "One of activeDefinition and draftDefinition must be specified" line.

nbeloglazov: Done.

server/app/services/question/ActiveAndDraftQuestions.java
nbeloglazov: Reverted back to using only questionName instead of ids.

shanemc-goog:
Sorry I'm not going to have time to review this today, Status Tracking took my focus and I'm OOO mon/tue.  Bion can likely review the versioning model interactions if you want that.

nbeloglazov:
I reverted versioning changes so now it's only UI. Will be easier to get in.

server/app/services/question/ActiveAndDraftQuestions.java
```diff
@@ -68,6 +68,7 @@ private ActiveAndDraftQuestions(Version active, Version draft, Version withDraft
 
     versionedByName =
         Sets.union(activeNameToQuestion.keySet(), draftNameToQuestion.keySet()).stream()
+            .sorted()
```
clouser2, line 71: Was this call necessary? Does this mean there's now an implication of sort order that wasn't there before?

Mostly wondering since we're taking a stance in  https://github.com/civiform/civiform/issues/3029 that sorting would be done client side.

Additionally, the type we're assigning this to (e.g. `Map`) has no explicit guarantees about order. Given this, it feels like the call may be unnecessary.

clouser2 (approved):

server/app/services/question/ActiveAndDraftQuestions.java
nbeloglazov: ImmutableMap guaranteers stable order (insertion order). I added this as I saw some flakiness in the test where there were 2 questions. Sometimes their order changed. Given that you are adding sort-by-recent I decided that sorting here by name is ok and short-lived anyway.

clouser2: I guess my general comment is that a map is an unordered type (even if the concrete implementation preserves order). Given that this is short-lived, it sounds good to me. What do you think about addding a TODO against #3029 and we can remove it at that point?

Or maybe I'm just being overly pedantic about assuming order.

nbeloglazov: Done.

PR #3168, opened 2022-08-15
Add field requiredness indicators to program/question UIs.

clouser2:
### Description

While here, I put the program visibility mode / question export mode options inside a fieldset for better grouping (and a place to put field requiredness).

### Screenshots

Program creation:
Screen Shot 2022-08-15 at 1 44 17 PM

Question creation:
Screen Shot 2022-08-15 at 1 45 12 PM

## Release notes:

Add indicators for which fields are required when editing/creating questions and programs.

### Checklist

- [X] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)

Fixes #3023

PR #3381, opened 2022-09-09
Add requiredness indicators to question text

### LINKED ISSUE #2629: (WCAG 2.1) Required fields should be identified on first page load.

Severity (Critical/Serious/Moderate/Good to have): Serious

Required fields should be identified on first page load. We should identify required fields both visually and programmatically. 

Visually, we could: (1) put a "required" text label, or (2) mark it with a star. Programmatically, we should mark the html as required, either with (1) html required attribute, or (2) aria-required attribute. The html attribute is preferred, because it has baked in information.

arielwainer:
### Description

Indicate requiredness with a * and helper text based on similar change made in #3168

### Screenshots

Screen Shot 2022-09-09 at 11 45 04 AM

## Release notes:

Add indicators for which questions are required when filling out a form.

### Checklist

### Issue(s)

Fixes #2629;

### DIFF
--- a/server/app/views/applicant/ApplicantProgramBlockEditView.java
+++ b/server/app/views/applicant/ApplicantProgramBlockEditView.java
@@ -121,6 +121,7 @@ public final class ApplicantProgramBlockEditView extends ApplicationBaseView {
         .withAction(formAction)
         .withMethod(HttpVerbs.POST)
         .with(makeCsrfTokenInputTag(params.request()))
+        .with(requiredFieldsExplanationContent().withClasses(Styles.TEXT_SM, Styles.TEXT_GRAY_600, Styles.MB_2))
         .with(
             each(
                 params.block().getQuestions(),
--- a/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
+++ b/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
@@ -63,6 +63,8 @@ abstract class ApplicantQuestionRendererImpl implements ApplicantQuestionRendere
                         ApplicantStyles.QUESTION_HELP_TEXT)
                     .with(
                         TextFormatter.createLinksAndEscapeText(
+                          !question.isOptional() ?
+                          question.getQuestionHelpText() + "*" :
                             question.getQuestionHelpText(), TextFormatter.UrlOpenAction.NewTab)))
             .withClasses(Styles.MB_4);
 
--- a/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
+++ b/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
@@ -63,7 +63,7 @@ abstract class ApplicantQuestionRendererImpl implements ApplicantQuestionRendere
                         ApplicantStyles.QUESTION_HELP_TEXT)
                     .with(
                         TextFormatter.createLinksAndEscapeText(
-                          question.getQuestionHelpText(), TextFormatter.UrlOpenAction.NewTab)))
+                        question.getQuestionHelpText(), TextFormatter.UrlOpenAction.NewTab)))
             .withClasses(Styles.MB_4);
 
     ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> validationErrors;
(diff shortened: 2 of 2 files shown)


server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -99,7 +99,8 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
     ContainerTag questionTag;
     ImmutableList<DomContent> questionTextDoms =
         TextFormatter.createLinksAndEscapeText(
-            question.getQuestionText(), TextFormatter.UrlOpenAction.NewTab);
+            !question.isOptional() ? question.getQuestionText() + " *" : question.getQuestionText(),
```
jeanniefu, line 102: Oh actually can you remove the space before the asterisk? The screenshots in the admin PR weren't updated.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -99,7 +99,8 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
     ContainerTag questionTag;
     ImmutableList<DomContent> questionTextDoms =
         TextFormatter.createLinksAndEscapeText(
-            question.getQuestionText(), TextFormatter.UrlOpenAction.NewTab);
```
jeanniefu, line 102: Also the screenshots will need to be updated https://docs.civiform.us/contributor-guide/developer-guide/testing#screenshot-diffing

jeanniefu (changes requested):
Yay nice first PR! Btw, for branch names we typically prepend it with our usernames (e.g. "arielwainer/star"). That way there is less of a chance of branch naming conflicts, since when you push your branch it exists on the project level with everyone else's: https://github.com/civiform/civiform/branches. Don't change this PR though!

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
arielwainer: Done.

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
arielwainer: Done.

jeanniefu: Maybe it didn't push? the space is still there

arielwainer: Should be fixed now

browser-test/src/program_deep_link.test.ts
```diff
@@ -46,7 +46,7 @@ describe('navigating to a deep link', () => {
     // Assert
     await page.click('#continue-application-button')
     expect(await page.innerText('.cf-applicant-question-text')).toEqual(
```
jeanniefu, line 48: blocking: How about `toContain(questionText)`? so we can reuse the `questionText` variable

browser-test/src/program_deep_link.test.ts
```diff
@@ -60,7 +60,7 @@ describe('navigating to a deep link', () => {
...
-      questionText,
```
jeanniefu, line 63: Same comment here!

server/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -121,6 +121,9 @@ private ContainerTag<?> renderBlockWithSubmitForm(Params params) {
         .withAction(formAction)
         .withMethod(HttpVerbs.POST)
         .with(makeCsrfTokenInputTag(params.request()))
+        .with(
+            requiredFieldsExplanationContent()
```
jeanniefu, line 125: Sorry I missed this initially, but we'll need to use the messages API instead of using a hardcoded English string. This part: https://github.com/civiform/civiform/blob/2a3664c6477d0dca4996104a243b6bba8210206e/server/app/views/BaseHtmlView.java#L150-L153

should be updated. Here's an example PR of how to do that:
https://github.com/civiform/civiform/pull/3564/files

And can you actually split that out into a different PR? Based on the slack thread, we should submit it for translation first, then block this PR until it's been translated.

Our very official process for translations is to add the strings and then ping Bion :D

jeanniefu (changes requested):

browser-test/src/program_deep_link.test.ts
arielwainer: Done.

PR #3383, opened 2022-09-09
Clickable FileUploads for PDF export funtionality

### LINKED ISSUE #3351: PDF export should include link to any uploaded documents
Program admins find it simple and useful to view the submitted applications via the new PDF export. However, they cannot view any uploaded documents via the PDF export. It would be helpful if the PDF export included a link to any uploaded documents. It would be easiest if they could click on the link and have the document automatically download OR open in a new browser tab.

Include the URL for the residents' uploaded documents in the PDF export. It would be helpful it the Program Admin could click directly on the URL to download the document OR open in a new browser tab.

**Additional context**
Idea attached

**Done when**
Program admins can open residents' uploaded documents from the PDF export
[PDF with URLs.pdf](https://github.com/civiform/civiform/files/9499846/PDF.with.URLs.pdf)

swatkat1:
### Description

Added the FileUploads URL for the PDFs in the PDF export functionality.

## Release notes:

PDFs from the PDF Export will have clikable links to the FileUploads.

### Checklist

### Issue(s)

Fixes #3351


server/app/services/export/PdfExporter.java
```diff
@@ -19,9 +20,12 @@
...
 import javax.inject.Inject;
 import models.Application;
+import services.Path;
 import services.applicant.AnswerData;
 import services.applicant.ApplicantService;
 import services.applicant.ReadOnlyApplicantProgramService;
+import services.applicant.question.Scalar;
+import services.question.types.QuestionType;
 
 /** PdfExporter is meant to generate PDF files. The functionality is not fully implemented yet. */
```
clouser2, line 30: Mind removing the "is not fully implemented yet" part of the docstring? If there's still functionality left to be implemented, I'd consider explicitly listing it out (with a tracking bug).

server/app/services/export/PdfExporter.java
```diff
@@ -19,9 +20,12 @@
...
 public class PdfExporter {
```
clouser2, line 31: Not blocking: Would you mind marking this as `final` to make the decision to allow subclassing explicit rather than implicit?

server/app/services/export/PdfExporter.java
```diff
@@ -84,8 +88,19 @@ private byte[] buildPDF(
             new Paragraph(
                 answerData.questionDefinition().getName(),
                 FontFactory.getFont(FontFactory.HELVETICA_BOLD, 12));
-        Paragraph answer =
-            new Paragraph(answerData.answerText(), FontFactory.getFont(FontFactory.HELVETICA, 11));
+        Paragraph answer = null;
+        if (QuestionType.FILEUPLOAD == answerData.questionDefinition().getQuestionType()) {
+          Path contextPath = answerData.contextualizedPath().join(Scalar.FILE_KEY);
```
clouser2, line 93: Can we follow a similar pattern to what we do in JSONExporter ([code](https://github.com/civiform/civiform/blob/14b3f1519fdf2422063d97765212539eda9e3b17/server/app/services/export/JsonExporter.java#L107-L123)) of getting the concrete question type rather than duplicating the extraction code here?

I think it could be something along the lines of:
```
FileUploadQuestion fileUploadQuestion = answerData.applicantQuestion().createFileUploadQuestion();
Optional<String> maybeFileUrl = fileUploadQuestion.getFileKeyValue();
```

I think the current code may imply that we need to handle a missing file (perhaps due to the question being optional)?

server/test/services/export/PdfExporterTest.java
```diff
@@ -30,10 +35,23 @@ public void validatePDFExport() throws IOException, DocumentException {
     int pages = pdfReader.getNumberOfPages();
     for (int i = 1; i < pages; i++) {
       textFromPDF.append(PdfTextExtractor.getTextFromPage(pdfReader, i));
+      // Assertions to check if the URL is embedded for the FileUpload
```
clouser2, line 38: Does the fake data already contain an "optional" file upload question? Can we add some assertions for how the case where no file is provided (and the program defines it as optional) is handled?

server/test/services/export/PdfExporterTest.java
```diff
@@ -30,10 +35,23 @@ public void validatePDFExport() throws IOException, DocumentException {
...
+      assertThat(link.toString())
+          .isEqualTo(
+              "http://localhost:9000/admin/programs/"
+                  + applicationOne.getProgram().id
+                  + "/files/my-file-key");
     }
 
     pdfReader.close();
-
     assertThat(textFromPDF).isNotNull();
```
clouser2, line 55: Is there any browser test coverage of this? I understand that asserting PDF content in a browser test would likely not be worth our time. That said, we could have a test that clicks the download button and asserts that a non-empty file is downloaded.

As it stands, we have a bit of a gap in our coverage for a scenario where the PDFExporter code properly generates the data, but for some reason our controller doesn't send it back to the client.

clouser2 (changes requested):

server/test/services/export/PdfExporterTest.java
swatkat1: I don't think optional or not doesn't matter to the PDFExporter. The fake data has a file upload question and I am checking the PDFAnnots to make sure the URL is added.

server/app/services/export/PdfExporter.java
swatkat1: Good catch! Removed it.  Everything that was requested is implemented.

server/app/services/export/PdfExporter.java
swatkat1: Sure

server/app/services/export/PdfExporter.java
swatkat1: Very good observation! thank you!

server/test/services/export/PdfExporterTest.java
swatkat1: Added a browser tests

server/app/services/export/PdfExporter.java
```diff
@@ -84,8 +98,21 @@ private byte[] buildPDF(
             new Paragraph(
                 answerData.questionDefinition().getName(),
                 FontFactory.getFont(FontFactory.HELVETICA_BOLD, 12));
-        Paragraph answer =
-            new Paragraph(answerData.answerText(), FontFactory.getFont(FontFactory.HELVETICA, 11));
+        Paragraph answer = null;
+        if (answerData.fileKey().isPresent()) {
+          String encodedFileKey =
```
clouser2, line 103: This doesn't appear to be using the helper methods from `FileUploadQuestion`. Did that approach not work?

swatkat1: Which helper method? I don't see any method that gets a encodedFilekey

clouser2: I'd misunderstood. Not blocking this PR, but it might make sense to change `AnswerData.fileKey` to be `AnswerData.encodedFileKey()` and do the encoding in the place that sets it since all consumers of it turn right around and encode it.

clouser2 (approved):

PR #3384, opened 2022-09-09
Enable asset caching on browser tests

nbeloglazov:
### Description

This improves speed of browser tests. For `quesiton_lifecycle.test.ts` duration goes from 250s to 140s. 

### Checklist

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -613,3 +613,8 @@ application_status_tracking_enabled = ${?CIVIFORM_APPLICATION_STATUS_TRACKING_EN
 # work as expected in production if there are multiple servers.
 feature_flag_overrides_enabled = false
 feature_flag_overrides_enabled = ${?FEATURE_FLAG_OVERRIDES_ENABLED}
+
+# If enabled, all assets (JS, CSS files) will be returned with long caching.
+# This setting should be used only by browser tests. Note that it is used
+# in dev mode where normally files served with caching disabled.
+enable_asset_caching_for_browser_tests = false
--- a/server/conf/application.dev-browser-tests.conf
+++ b/server/conf/application.dev-browser-tests.conf
@@ -7,3 +7,6 @@ include "application.dev.conf"
 # they should enable them via the Feature Flags HTTP
 # handler as needed.
 application_status_tracking_enabled = false
+
+# Enable caching of JS/CSS resources to speed up tests.
+enable_asset_caching_for_browser_tests = true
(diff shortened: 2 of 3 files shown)


server/conf/application.dev-browser-tests.conf
```diff
@@ -7,3 +7,6 @@ include "application.dev.conf"
 # they should enable them via the Feature Flags HTTP
 # handler as needed.
 application_status_tracking_enabled = false
+
+# Enable caching of JS/CSS resources to speed up tests.
+enable_asset_caching_for_browser_tests = true
```
clouser2, line 12: I like the general approach. Is there any way to enable the "fingerprinting alluded to  in the omment in DisableCachingFilter.java?

Setting a long cache definitely makes sense to me for CI / run-browser-tests. Is it possible that run-browser-tests-local would end up keeping this cached between runs? The answer may be likely not, but I don't have a lot of context here since as I understand, the spawned browser in this case is local to a machine (semi-persistent) and not ephemeral like our Docker containers.

nbeloglazov: It kinda alludes to my question on the chat. I just tested and if we enable fingerprinting locally in dev mode - it automatically rebuilds/refingerprints JS upon changes. So we can potentially just enable it even in dev mode and enable full caching even on dev, which will help to avoid the need browser-test-specific logic.

@kmcnellis WDYT?

clouser2 (approved):

server/conf/application.conf
```diff
@@ -613,3 +613,8 @@ application_status_tracking_enabled = ${?CIVIFORM_APPLICATION_STATUS_TRACKING_EN
 # work as expected in production if there are multiple servers.
 feature_flag_overrides_enabled = false
 feature_flag_overrides_enabled = ${?FEATURE_FLAG_OVERRIDES_ENABLED}
+
+# If enabled, all assets (JS, CSS files) will be returned with long caching.
+# This setting should be used only by browser tests. Note that it is used
+# in dev mode where normally files served with caching disabled.
+enable_asset_caching_for_browser_tests = false
```
kmcnellis, line 620: I would probably prefer just name this "enable_dev_asset_caching", or something else that indicates that if this is true, it will enable it in all dev envs, not just in the test one.  Or, we could just have a flag that says "is_browser_test" and use that instead.

server/conf/application.dev-browser-tests.conf
kmcnellis: I would be OK with that, as long as it doesn't increase compile time

kmcnellis (changes requested):

server/conf/application.dev-browser-tests.conf
nbeloglazov: Opened another PR with the proposed approach: https://github.com/civiform/civiform/pull/3394/files It affects compile (recompile) time only when changing TS file and even than just 1-2 sec.

PR #3389, opened 2022-09-12
Generalizing approach of waiting for a script to be ready.

clouser2:
### Description

As part of #3029, we'll be adding client-side sorting of the question list. This sorting will occur after the 'load' event and affect the UI. As it stands, any code that merely waits for a page to complete loading might run into timing issues where the sorting script hasn't loaded.

#### Present state

We have precedent for browser test code waiting until page script execution has completed, but it's bespoke and hard-coded for `main.ts` / `modal.ts` ([code](https://github.com/civiform/civiform/blob/f220c1f004a12251e8cedb7c295f9c2ab0dfa581/server/app/assets/javascripts/main.ts#L586-L587)). Browser test code's `waitForPageJsLoad` explicitly waits for this attribute to be set ([code](https://github.com/civiform/civiform/blob/f220c1f004a12251e8cedb7c295f9c2ab0dfa581/browser-test/src/support/wait.ts#L16)). `waitForPageJsLoad` is called extensively from support code.

We have a variety of other scripts that we use which could have similar timing issues. Our present approach assumes that `main.ts` and `modal.ts` will be present on any page where `waitForPageJsLoad` is called. This translates to a limitation that similar waits can't be added for new scripts unless they're included in *all* page loads.

#### Post-PR state

This PR moves us to a more generalized approach where all CiviForm scripts have the `data-is-loaded` attribute explicitly set to false initially (see changes in `ViewUtils.java`):
```
<script src="/assets/admin_programs.ts" data-has-loaded="false" />
```

Then, each script is now responsible for setting the attribute to true once it's completed initialization (similar to the pattern already used in `main.ts`. This is achieved by making use of the `document.currentScript` global variable ([docs](http://go/mdn/API/Document/currentScript)), which points to the current script element being executed.

Finally, `waitForPageJsLoad` has now been modified to wait until all `script` elements with a `data-has-loaded` attribute resolve to true.

## Release notes:

N/A

### Checklist

### Issue(s)

#3029


clouser2:
FYI, the issue that led to this PR is no longer relevant since we don't do client-side sorting of the list. That said, this does allow us to improve our ability to "wait" for any script to be loaded on the current page.

I'm happy to close out if we don't see a need for it anymore.

bion:
We've switched to using webpack for serving javascript, removing the need to check if individual scripts have loaded.

PR #3394, opened 2022-09-12
Enable fingerpriting and caching for static files even on dev mode

nbeloglazov:
### Description

This improves speed of browser tests. For quesiton_lifecycle.test.ts duration goes from 250s to 140s. Fingerprinting guarantees that when while changes - a new version will be requested. The downside of that approach is that it adds an extra step (fingerprinting) into incremental rebuild of static assets (TS files). But that adds only around 1-2s so it's practically unnoticeable. It might become an issue in future if we have bigger static files, but with current guidelines (minimum JS) that unlikely to happen soon.

### Checklist

### DIFF
--- a/server/build.sbt
+++ b/server/build.sbt
@@ -120,8 +120,9 @@ lazy val root = (project in file("."))
     // https://github.com/sbt/zinc/issues/911
     incOptions := incOptions.value.withTransitiveStep(2),
     pipelineStages := Seq(digest, gzip), // plugins to use for assets
-    // Uncomment to test the sbt-web asset pipeline locally.
-    // Assets / pipelineStages  := Seq(digest, gzip), // Test the sbt-web pipeline locally.
+    // Enable digest for local dev so that files can be served çached improving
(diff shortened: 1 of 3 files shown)


server/build.sbt
```diff
@@ -120,8 +120,9 @@ lazy val root = (project in file("."))
     // https://github.com/sbt/zinc/issues/911
     incOptions := incOptions.value.withTransitiveStep(2),
     pipelineStages := Seq(digest, gzip), // plugins to use for assets
-    // Uncomment to test the sbt-web asset pipeline locally.
-    // Assets / pipelineStages  := Seq(digest, gzip), // Test the sbt-web pipeline locally.
+    // Enable digest for local dev so that files can be served çached improving
+    // page speed and also browser tests speed.
+    Assets / pipelineStages  := Seq(digest, gzip),
```
kmcnellis, line 125: The other big thing to check is if this decreases readability / debug-ability.

kmcnellis (approved):

Issue #3391, opened 2022-09-12
Programs / questions page inconsistency with the word "draft"


clouser2:
Thanks to @jeanniefu for bringing this up. @jeanniefu, @nbeloglazov, and I thought about this and are trying to summarize a bit of discussion. We'll also coordinate with @rahaghas to gather feedback on the proposal as well.

## Problem 
The questions page shows programs referencing a question to see where it's used from. The text shown in the list in the dialog references "active" and "draft" programs. The questions page also references "active" and "draft" questions. Showing this phrasing can be confusing.

If we can try and add clarity to how / when a question is used without directly using active/draft terminology, it might make the dialog easier to digest and hopefully make the feature more useful.

## Proposal

We can update both the summary line shown in the question list as well as the modal content.

### Referencing programs modal

Current UI: ![Screen Shot 2022-09-12 at 10 44 52 AM](https://user-images.githubusercontent.com/1870301/189721336-0bea89f1-450b-458e-942e-2372c0bb560d.png)

Rather than displaying "draft" and "active" programs referencing a question, we could instead have three distinct groups about how the question is referenced:

This question is used in:
  * Program A

This question is added to:
  * Program B
  * Program C

This question is removed from:
  * Program D

Note: If any section has no content, it will be hidden. Caveat: Unreferenced questions won't have a link shown in the question list, so we won't have an "empty dialog" state.

Additionally, we can change the title to "Programs referencing <questionname>".

### Question list

The question list UX is about to be redesigned using a card view (similar to the program list page) in #3378. In this, each "draft/active" row for a given question card would have a separate count for the number of programs referencing it (![Screen Shot 2022-09-12 at 10 42 38 AM](https://user-images.githubusercontent.com/1870301/189720941-3c181110-ff5d-4e70-9bdf-568d782f0c1e.png)
).

Creating a draft question doesn't actually have any tangible change on the number of programs referencing a question. That can only be changed from the program builder. Given this, it might make sense to only show the count once and somewhere not tied to the concept of the draft/active version of a question (perhaps on the left side of the card?).

The summary line itself can move away from the draft/active text and summarize the groups that would be shown in the dialog. Depending on how the question was changing in the draft program, any of the following summary lines could be possible (a link will be shown for all states except for a completely unreferenced question):

Used in 3 programs, added to 4 programs, and removed from 5 programs
Used in 3 programs.
Used in 3 programs and added to 4 programs.
Used in 3 programs and removed from 5 programs.
Added to 1 program.
[...]

PR #3405, opened 2022-09-12
Render "referenced programs" text outside of active/draft rows

nbeloglazov:
### Description

Relation between question and programs that reference it is orthogonal to the question being active or draft. Programs include question by name and it's not possible to include, for example, only draft version or only active version of a question. This PR makes it more clear by moving "referenced programs" text outside of active/draft rows. 

### Checklist

### Issue(s)

Issue #3391


nbeloglazov:
@rahaghas can you take a look at screenshots changes: https://github.com/civiform/civiform/pull/3405/files Should I align "refernced program" text vertically differently? Wasn't sure what would be the best, given that now this text spans potentially two rows.

rahaghas (approved):
Can we make it so each item is a new line? so instead of:

1 used, 3 added, 4 removed

it's:
1 used
3 added
4 removed

clouser2 (approved):

PR #3407, opened 2022-09-13
Support importing types in server TS code

nbeloglazov:
### Description

This PR adds example of exporting/importing types between two TS files used in the civiform app. Our current setup doesn't allow to export/import actual code as for that we need to bundle files together. But doing same with types is possible as types are removed during transpilation so at the end we get JS file with no real import statement. 

This allows us to properly import `Toast` controller which allows for better type checking and  makes code dependency more explicit. To make it work we need to include script tags as `<script src="..." type="module">`. That should be ok as modern browsers support it and we don't support IE11 and below.

### Checklist

### Issue(s)

Issue #3361


bion (changes requested):
please fix test failures

nbeloglazov:
Huh. Now I'm pretty confused. Tried running java unit tests. And it failed due to typescript compilation issue: it doesn't know `unknown` type and had troubles with `import type`. Both of those were added in Typescript 3+ while we are using plugin that uses Typescript 2. The odd thing is that when I run `bin/run-dev` - it works ok and happily compiles code. I don't know understand how it's doing that as it should be failing as well. 

I need to figure out that first before proceeding with this change.

nb1701:
Please reopen if we run into this issue again.

PR #3412, opened 2022-09-13
Render "draft" badge on program block edit page

nbeloglazov:
### Description

This PR has few changes:

1. Render "draft" badge instead of simple text on program blocks edit page. According to [slides](https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13e43d0b668_0_134). 
2. Reduce height of the badge from 48px to 40px which better aligns with slides.
3. Moves badge creation code into a shared function and removes duplication. 

### Checklist


server/app/views/components/ProgramCardFactory.java
```diff
@@ -134,23 +134,7 @@ private DivTag renderProgramRow(
...
-                    Styles.ML_2,
-                    StyleUtils.responsiveXLarge(Styles.ML_8),
-                    Styles.FONT_MEDIUM,
-                    Styles.ROUNDED_FULL,
-                    Styles.FLEX,
-                    Styles.FLEX_ROW,
-                    Styles.GAP_X_2,
-                    Styles.PLACE_ITEMS_CENTER,
-                    Styles.JUSTIFY_CENTER)
-                .withStyle("min-width:90px")
```
clouser2, line 149: Did the "min-width" get lost when pulling out into a common class? I think this was to prevent small screen sizes from collapsing the badge into a cramped state.

server/app/views/components/ProgramCardFactory.java
```diff
@@ -126,6 +121,11 @@ private DivTag renderProgramRow(
                 Styles.H_12,
                 programRow.extraRowActions().size() == 0 ? Styles.INVISIBLE : "");
 
+    PTag badge =
```
clouser2, line 124: The screenshots make it look like there's a fair amount of vertical spacing between the badge and the title. If that's what the mocks call for, then definitely ignore me.

clouser2 (approved):
Nice!

server/app/views/components/ProgramCardFactory.java
nbeloglazov: I replaced it with "width: 100px". Having fixed width is better as otherwise they have width few pixels off and you can see elements misaligned.

server/app/views/components/ProgramCardFactory.java
nbeloglazov: Yes, there is space between in the mocks.

PR #3415, opened 2022-09-14
Add aria error attributes to field upload client side validation

### LINKED ISSUE #1878: question validation error messages are not readable to a screen reader

jeanniefu:
### Description

Manually tested. Before this change, the error was not read out when the button was selected. After this change, it is read out.

## Release notes:

[a11y] Add aria error attributes to file upload questions, so screen readers read out the error.

### Checklist

### Issue(s)

Fixes #1878


server/app/assets/javascripts/file_upload.ts
```diff
@@ -26,12 +28,25 @@ function validateFileUploadQuestions(formEl: Element): boolean {
...
+    if (errorDiv && !isValid && !wasSetInvalid) {
+      // Add extra aria attributes to input if there is an error.
+      const errorId = errorDiv.getAttribute('id')
+      if (errorId) {
+        // Only allow this to be done once so we don't repeatedly append the error id.
+        wasSetInvalid = true
+        fileInput.setAttribute('aria-invalid', 'true')
+        const ariaDescribedBy = fileInput.getAttribute('aria-describedBy')
+        fileInput.setAttribute(
+          'aria-describedBy',
```
bion, line 45: question (blocking): does the b in "by" need to be lower case?

bion (changes requested):

server/app/assets/javascripts/file_upload.ts
```diff
@@ -26,12 +28,25 @@ function validateFileUploadQuestions(formEl: Element): boolean {
...
+          errorId + ' ' + ariaDescribedBy,
```
clouser2, line 46: Not blocking: This can be written as an interpolated string.

```
fileInput.setAttribute(
  'aria-describedBy',
  `${errorId} ${ariaDescribedBy}`,
)
```

clouser2 (approved):

server/app/assets/javascripts/file_upload.ts
jeanniefu: No, either is fine. Looking at the docs, it automatically all gets lowercased: 
https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#lower_casing
https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#parameters

And I see that the HTML that's rendered by j2html actually is all lowercase too. I'll go through and just change everything to `aria-describedby` since it's unnecessary that I capitalized the `B`. Will send that out separately.

server/app/assets/javascripts/file_upload.ts
jeanniefu: Nice, thanks!

bion (approved):

PR #3431, opened 2022-09-15
Sort question bank / question list questions by most recently modified

clouser2:
### Description

This sorts the question list and question bank server side by most recently modified with a fallback to the displayed name. Additionally, a "sorted by" chip was added to the question / program list (screenshots below).

This removes the client side sorting introduced for the program list per the rationale in https://github.com/civiform/civiform/issues/3029#issuecomment-1248285805. We can always reintroduce client-side sorting if it becomes necessary / helpful.

A previous iteration of this was sent out as PR #3359 (discarded) which performed the sorting client-side.

### Screenshots

Questions list before:
Screen Shot 2022-09-15 at 1 15 41 PM

Questions list after:
Screen Shot 2022-09-15 at 1 14 12 PM

Programs list before:
Screen Shot 2022-09-15 at 1 15 46 PM

Programs list after:
Screen Shot 2022-09-15 at 1 14 06 PM

## Release notes:

Sorting the question list and questions shown in the program builder by the most recently updated question.

### Checklist

### Issue(s)

#3029


browser-test/src/civiform_admin_question_list.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    ])
+    await expectQuestionBankTopElements(programName, adminPrograms, [
+      questionTwoPublishedText,
+      questionOnePublishedText,
+    ])
+
+    // Now create a draft version of the previously last question. After,
+    // it should be on top.
+    await adminQuestions.createNewVersion(questionOnePublishedText)
+    const questionOneDraftText = `${questionOnePublishedText} new version`
```
nbeloglazov, line 63: Here "new version" implicitly comes from the `createNewVersion` function?

browser-test/src/civiform_admin_question_list.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!expectedQuestions) {
+      throw new Error('expected at least one question')
+    }
+    const questionListNames = await adminQuestions.questionNames()
+    expect(questionListNames.length).toBeGreaterThanOrEqual(
+      expectedQuestions.length,
+    )
+    expect(questionListNames.slice(0, expectedQuestions.length)).toEqual(
+      expectedQuestions,
+    )
```
nbeloglazov, line 104: I think that's the question from the previous review, don't remember the answer. Why can't we assume that number of questions strictly matches expectedQuestions given that test wipes out DB at the beginning of each test file.

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -113,37 +115,82 @@ private DivTag renderSummary(ActiveAndDraftQuestions activeAndDraftQuestions) {
         .withClasses(Styles.FLOAT_RIGHT, Styles.TEXT_BASE, Styles.PX_4, Styles.MY_2);
   }
 
+  private static QuestionDefinition getDisplayQuestion(QuestionCardData cardData) {
+    if (cardData.draftQuestion().isPresent()) {
```
nbeloglazov, line 119: Elsewhere we've been using

return cardData.draftQuestion().orElseGet(cardData.activeQuestion()::get);

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -113,37 +115,82 @@ private DivTag renderSummary(ActiveAndDraftQuestions activeAndDraftQuestions) {
...
+  private Pair<DivTag, ImmutableList<Modal>> renderQuestionCard(
+      QuestionCardData cardData,
+      ActiveAndDraftQuestions activeAndDraftQuestions,
+      Http.Request request) {
+    if (cardData.draftQuestion().isEmpty() && cardData.activeQuestion().isEmpty()) {
       throw new IllegalArgumentException("Did not receive a valid question.");
     }
-    QuestionDefinition latestDefinition = draftDefinition.orElseGet(activeDefinition::get);
+    QuestionDefinition latestDefinition =
+        cardData.draftQuestion().orElseGet(() -> cardData.activeQuestion().get());
```
nbeloglazov, line 189: Hm. Java doesn't allow you to do `cardData.activeQuestion()::get` ?

server/app/views/components/ProgramCardFactory.java
```diff
@@ -185,25 +183,20 @@ private DivTag renderProgramRow(
...
-    if (cardData.draftProgram().isEmpty() && cardData.activeProgram().isEmpty()) {
-      throw new IllegalArgumentException("Program neither active nor draft.");
-    }
-
-    ProgramDefinition program =
-        cardData.draftProgram().isPresent()
-            ? cardData.draftProgram().get().program()
-            : cardData.activeProgram().get().program();
-    return program.lastModifiedTime().orElse(Instant.EPOCH);
+  public static Comparator<ProgramCardData> comparator() {
```
nbeloglazov, line 193: Can we clarify what fields comparator compares by? lastModifiedTimeThenNameComparator()? We are in java world after all and can use all the words.

nbeloglazov (changes requested):

browser-test/src/civiform_admin_question_list.test.ts
clouser2: I believe we likely can, but wanted to separate that from this change (since the pre-existing `admin_program_list.test.ts` follows the same pattern.

I left a TODO on line 29 to track cleaning this up in a follow-on PR if possible. At worst, I'm creating extra work for myself. Does that sound like an acceptable way forward to you?

server/app/views/admin/questions/QuestionsListView.java
clouser2: Done!

server/app/views/admin/questions/QuestionsListView.java
clouser2: Switched to that. I just hadn't thought of doing it that way :)

server/app/views/components/ProgramCardFactory.java
clouser2: Done.

clouser2 (commented):
Feedback incorporated.

browser-test/src/civiform_admin_question_list.test.ts
clouser2: That's correct. This was some surprising action-at-a-distance. I've now added an explicit comment around that.

nbeloglazov (approved):

Issue #1593, opened 2021-08-20
Login access as CiviForm Admin AND Program Admin


elisekalstad:
**Describe the bug**
I'm not sure if this is actually a bug--I would also understand if this is how the tool is intended to be in order to preserve additional layers of privacy. BUT

I'm looking for a good way to be able to see how many applications were submitted. As CiviForm Admin, I have no way (currently) of knowing if it is a test application or a real application. My work-around is to assign myself as a Program Administrator to get access to the applications as they come in. 

But I cannot login as a Program Administrator--my CiviForm Admin role takes precedence.

**To Reproduce**
1. Go to "Civiform.Seattle.gov"
2. Click on Admin Login'
3. Only able to view the CiviForm Admin view

Alternatively:
1. Open email when a new applicant submits an application
2. Click on the link to view the application
3. Get an error message, attached

**Expected behavior**
If the CiviForm Admin role is deliberately separate from the Program Admin role (which is very well might be!), then it's functioning as expected. It would be great to have a way to see how many applications were received that are REAL applications and not just test ones :)

**Screenshots**
[Program Admin vs. CiviForm Admin.docx](https://github.com/seattle-uat/civiform/files/7018387/Program.Admin.vs.CiviForm.Admin.docx)

elisekalstad:
I am unable to act as both a CiviForm Administrator AND Program Administrator. Without this access, I have a hard time troubleshooting issues that Program Admins face. I often cannot replicant their issue in Staging. It would also help to respond to resident questions about the status of their application, if I'm able to look them up in CiviForm. Is it possible to re-open this request? I'd like to better understand what it will take to grant CiviForm Admins (Matt and I) access as Program Admin)

azizova-leyla:
@cassandralfernandes need your input. 

It was an explicit decision to not give Civiform Admin access to Program Applications, to have better privacy model and limit number of people accessing PII. @elisekalstad if that's not a concern for you - we can change it.

Currently we have an explicit check to not allow Civiform Admin assign themselves as Program Admin. We can add a server config variable that will disable that check. So it will be up to civic entity to enable that or not.

That would behave similarly to demo server when using "Program and Civiform Admin" button. Applications will be under 3 dot menu: 
![image](https://user-images.githubusercontent.com/1741747/191316501-2e65b82f-adfc-410e-9dd1-d95264ca0068.png)

There two options for the user flow:
[...]

PR #3440, opened 2022-09-16
Add config variable for global admin to get access to programs

azizova-leyla:
### Description

Adding new config flag: allow_global_admin_acccess_programs
When set to true Global Admins can also be Program Admins.
It will enable Global Admin assign themselves as Program Admins, and get access to  all Applications for that program.

## Release notes:

### Checklist

### Issue(s)

Fixes https://github.com/civiform/civiform/issues/1593

### DIFF
--- a/server/app/services/role/RoleService.java
+++ b/server/app/services/role/RoleService.java
@@ -7,7 +7,6 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
 import com.typesafe.config.Config;
-
 import java.util.Optional;
 import javax.inject.Inject;
 import models.Account;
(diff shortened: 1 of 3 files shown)


server/app/services/role/RoleService.java
```diff
@@ -58,10 +64,12 @@ public Optional<CiviFormError> makeProgramAdmins(
     // Filter out CiviForm admins from the list of emails - a CiviForm admin cannot be a program
     // admin.
     ImmutableSet<String> globalAdminEmails =
-        getGlobalAdmins().stream()
```
clouser2, line 61: Not blocking: This implementation overrides the meaning of `globalAdminEmails` when the configuration value is set. I'd consider either 1) renaming it to something more generic or 2) Keeping the list the same here and updating the check on l77 to consult the variable instead.

clouser2 (approved):

PR #3449, opened 2022-09-17
Add "Edited on" text to the admin question list page.

### LINKED ISSUE #3073: CiviForm admin UX: allow user to navigate directly to add a question from the main program page
The CiviForm admin wants the ability to navigate directly to add a question from the program row on the main programs page

When a user clicks the entire program row on the main program page, it take the user directly to the page to add screens and questions to their program. Clicking a draft program takes the user to where they left off

Clicking the edit button also directly takes the user to the page to add screens and questions to their program

**Additional context**
From admin UXR feedback sheet

**screenshot**
Screen Shot 2022-08-05 at 1 44 31 PM

nbeloglazov:
### Description

Mocks tell to use "Published on" on active questions and "Edited on" on draft, but we don't have true publish date for questions, only lastModifiedTime. So for now it renders "Edited on" for both active and draft questions.

## Release notes:

Add "Edited on" text to the admin question list page.

### Checklist

### Issue(s)

Fixes #3073


clouser2 (changes requested):

browser-test/src/question_lifecycle.test.ts
```diff
@@ -155,12 +155,7 @@ describe('normal question lifecycle', () => {
     // Add a new valid dropdown question
     await adminQuestions.addDropdownQuestion({questionName, options})
     // Edit the newly created question
-    await page.click(
-      adminQuestions.selectWithinQuestionTableRow(
-        questionName,
-        ':text("Edit")',
-      ),
-    )
+    await adminQuestions.gotoQuestionEditPage(questionName)
```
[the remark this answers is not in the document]
nbeloglazov: Thanks for catching. Added logic that replaces dates with fixed dates.

browser-test/src/support/index.ts
```diff
@@ -439,6 +438,15 @@ export const validateScreenshot = async (
   if (DISABLE_SCREENSHOTS) {
     return
   }
+  // To make screenshots stable go through all date fields (elements that have cf-bt-date class) and replace date/time with fixed text.
+  await element.evaluate(() => {
+    for (const date of Array.from(document.querySelectorAll('.cf-bt-date'))) {
+      console.log('found date ' + date.textContent)
+      date.textContent = date
+        .textContent!.replace(/\d{4}\/\d{2}\/\d{2}/, '2030/01/01')
```
clouser2, line 446: What about putting only the date component in a `<span>` element and just replacing the text content with 2030/01/01?

Then if we wanted date time, we could add a cf-bt-datetime class and add more to it.

server/app/views/ViewUtils.java
```diff
@@ -124,4 +131,30 @@ public static PTag makeBadge(BadgeStatus status, String... extraClasses) {
...
+   * @return Tag containing rendered time.
+   */
+  public PTag renderEditOnText(String prefix, Optional<Instant> time) {
+    String formattedUpdateTime = time.map(dateConverter::renderDateTime).orElse("unknown");
+    String formattedUpdateDate = time.map(dateConverter::renderDate).orElse("unknown");
+    return p().with(
+            span(prefix),
+            span(formattedUpdateTime)
+                .withClasses(
+                    ReferenceClasses.DATE_FIELD_BROWSER_TEST,
```
clouser2, line 150: Up to you, but we tend to just name the reference class const the same as the class itself and don't distinguish as to why it's present in the const name.

clouser2 (approved):

browser-test/src/support/index.ts
nbeloglazov: Having regexp replace provides benefit that if underlying format changes the regex either will still work, replacing part of the text, or will break, causing screenshot change. Fully replacing text content has potential issue of hiding underlying changes in rendering. For example we decided to render full date like "September 20, 2022".

server/app/views/ViewUtils.java
nbeloglazov: Done.

browser-test/src/support/index.ts
clouser2: Works for me! I'd consider adding something along these lines as a comment in the code for future readers.

clouser2 (approved):

Issue #3436, opened 2022-09-15
Make Release Action work


azizova-leyla:
Lastest issue is it can't find email for the releaser.

PR #3468, opened 2022-09-20
set releaser email in the script. Remove check for civiform login name

azizova-leyla:
### Description

Latest fail was complaining about releaser email - setting it explicitly in the script.
Remove check for civiform login name that doesn't work locally depending on docker version.

## Release notes:

### Checklist

### Issue(s)

Should help with [#3436](https://github.com/civiform/civiform/issues/3436)

### DIFF
--- a/bin/create-release
+++ b/bin/create-release
@@ -59,8 +59,8 @@ def main():
     release_number = sys.argv[2]
     if 'GITHUB_ACTOR' in os.environ:
         shell_cmd(
-            f'git config --global user.email ' +
-            f'{os.environ["GITHUB_ACTOR"]}@users.noreply.github.com')
+            f'git config user.email ' +
+            f'"{os.environ["GITHUB_ACTOR"]}@users.noreply.github.com"')
     releaser_email = shell_cmd('git config --get user.email')
 
     ################################################################################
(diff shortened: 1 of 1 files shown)


bin/create-release
```diff
@@ -57,9 +57,11 @@ def main():
 
     commit_sha = sys.argv[1]
     release_number = sys.argv[2]
-    releaser_email = os.environ[
-        'GITHUB_ACTOR'] if 'GITHUB_ACTOR' in os.environ else shell_cmd(
-            'git config --get user.email')
+    if 'GITHUB_ACTOR' in os.environ:
+        shell_cmd(
+            f'git config --global user.email ' +
+            f'{os.environ["GITHUB_ACTOR"]}@users.noreply.github.com')
```
gwendolyngoetz, line 63: I think the email address should be surrounded by double quotes

bin/create-release
```diff
@@ -57,9 +57,11 @@ def main():
 
     commit_sha = sys.argv[1]
     release_number = sys.argv[2]
-    releaser_email = os.environ[
-        'GITHUB_ACTOR'] if 'GITHUB_ACTOR' in os.environ else shell_cmd(
-            'git config --get user.email')
+    if 'GITHUB_ACTOR' in os.environ:
+        shell_cmd(
+            f'git config --global user.email ' +
```
gwendolyngoetz, line 62: Does this need to be set as global?

azizova-leyla: probably not.

Issue #3427, opened 2022-09-15
Status Tracking: Add current status in CSV/etc and API export


shanemc-goog:
All of our export and API flows that provide Application data should include the current status

shanemc-goog:
Status text is done.

If we want the other data, it will require some performance work likely as we have to join and filter ApplicationEvents for every application.

PR #3469, opened 2022-09-20
Add Statuses to CSV, PDF, Json export

shanemc-goog:
### Description

Exports the Status Tracking status for applications.

## Release notes:

When Status Tracking is enabled a new `Status` column is present in all file export flows.

### Checklist

### Issue(s)
Addresses #3427
Fixes


server/app/featureflags/FeatureFlags.java
```diff
@@ -32,11 +32,20 @@ public boolean areOverridesEnabled() {
         && config.getBoolean(FEATURE_FLAG_OVERRIDES_ENABLED);
   }
 
-  /** If the Status Tracking feature is enabled. */
+  /**
+   * If the Status Tracking feature is enabled.
+   *
+   * <p>Allows for overrides set in {@code request}.
+   */
```
clouser2, line 39: High-level, can we add a browser smoke test for these or is that overkill? I mainly ask because we already have testing for program admin exports that the CSV content includes / excludes certain applications. Presumably we could add an additional assertion that the status is set appropriately in the response (e.g. admin_application_statuses.test.ts.

clouser2 (approved):

server/app/featureflags/FeatureFlags.java
shanemc-goog: Hmmmmm. I think so given that I recently added /dev/feature.  Currently it only prints the general configuration for overrides, but I was planning on adding the actual feature flags values too.

```
Overrides are allowed if all are true:
Server environment: true
Configuration: true
```


Are you OK with that in another PR?

clouser2: Works for me.

PR #3479, opened 2022-09-20
Publish all drafts modals now displays program visibility and link to edit

### LINKED ISSUE #2114: Add summary of what is being published when an Admin Publishes All Programs

clouser2:
### Description

Since we're now emphasing the public-facing question text / program name, I also updated the text shown to use those values rather than the internal-facing values.

This is a follow-on to #3343, where I forgot to include this information, even though it was in the original mocks.

## Release notes:

Update the publish drafts dialog shown to the CiviForm admin to indicate program visibility and add edit links for the draft questions and programs.

### Checklist

### Issue(s)

Fixes #2114


server/test/controllers/admin/AdminProgramControllerTest.java
```diff
@@ -266,10 +266,7 @@ public void update_overwritesExistingProgram() throws Exception {
     Result redirectResult = controller.index(addCSRFToken(Helpers.fakeRequest()).build());
     assertThat(contentAsString(redirectResult))
         .contains(
-            "Create new program",
```
clouser2, line 269: Note to the reviewer: This one wasn't obvious. But the reason the "Existing one" (e.g. internal admin program name) existed prior to this change was because it was shown in the modal. Now that we use the external-facing name everywhere in the program list view, it's no longer included.

nbeloglazov (approved):

PR #3480, opened 2022-09-21
Generalize handling of required question errors

### LINKED ISSUE #3393: Add required errors into the validation error set for questions
Currently we have a special case for error message handling for required fields: https://github.com/civiform/civiform/blob/main/server/app/views/questiontypes/ApplicantQuestionRendererImpl.java#L82-L97

We should add the required errors into the error map as well. Then we can cleanup some of the boolean piping (`hasQuestionErrors`) I added in #3354.

See https://github.com/civiform/civiform/pull/3354 discussion for context.

jeanniefu:
### Description

Add required errors into the error map. Remove special logic that handles rendering required errors differently, and use general error rendering instead.

Screenshot diffs are because we now unify the styling of required error messages and validation error messages.

## Release notes:

Internal refactor of required error handling to consolidate with other errors.

### Checklist

### Issue(s)

Fixes #3393


browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -281,12 +281,10 @@ describe('End to end enumerator test', () => {
     await waitForPageJsLoad(page)
 
     await applicantQuestions.deleteEnumeratorEntity('Bugs')
-    await applicantQuestions.deleteEnumeratorEntity('Daffy')
```
jeanniefu, line 284: I had to update this test because the behavior is slightly different for deleting required enumerators now.

Before:
- **required non-enumerators**: Answer required question and save. Go back and try to clear required question and continue. Fails with "This question is required". Original data is **still there** and can be seen in review page
- **required enumerators**: Add enumerators and save. Go back and try to clear out all enumerators and continue. Fails with "This question is required". Original data is **removed** and no longer seen on review page.

After:
- **required non-enumerators**: no change in behavior (same as above)
- **required enumerators**: behavior now matches non-enumerators. Original data is still there and can be seen in review page when trying to remove required enumerator answers.

This change seems fine to me since we're unifying the behavior.

-------

Note: I actually cannot figure out why this is happening though. I think there might be something in [here](https://sourcegraph.com/github.com/civiform/civiform@3314e0b8debb78803bd09ee7d75cdee1d5658790/-/blob/server/app/services/applicant/ApplicantService.java?L231-233) or [here](https://sourcegraph.com/github.com/civiform/civiform@3314e0b8debb78803bd09ee7d75cdee1d5658790/-/blob/server/app/services/applicant/ApplicantService.java?L243) that results in us treating enumerators differently?

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -89,7 +88,7 @@ protected DivTag renderTag(
                 validationErrors.getOrDefault(addressQuestion.getZipPath(), ImmutableSet.of()))
             .addReferenceClass(ReferenceClasses.ADDRESS_ZIP);
 
-    if (hasQuestionErrors) {
+    if (!validationErrors.isEmpty()) {
```
jeanniefu, line 91: I was hoping to push this change further into `FieldWithLabel`, so that we didn't have to manage `FieldWithLabel` 's `aria-invalid` at this level.

However, we currently only pass the field level errors to `FieldWithLabel` (and support 3 different types - [link](https://github.com/civiform/civiform/blob/9a0ddd0468c22e09eb98b637d77efeac2aa9c1ac/server/app/views/components/FieldWithLabel.java#L276-L301)), and it felt a bit messy to try to expand it to pass in the whole error map.

server/app/services/applicant/question/Question.java
```diff
@@ -54,29 +55,43 @@ public final ImmutableMap<Path, ImmutableSet<ValidationErrorMessage>> getValidat
...
+
+    // Add error for unanswered required question.
+    if (applicantQuestion.isRequiredButWasSkippedInCurrentProgram()) {
+      Path keyPath = applicantQuestion.getContextualizedPath();
+      var valueErrors = new ImmutableSet.Builder<ValidationErrorMessage>();
+      if (errorMap.containsKey(keyPath)) {
+        valueErrors.addAll(errorMap.get(keyPath));
+      }
+      valueErrors.add(ValidationErrorMessage.create(MessageKey.VALIDATION_REQUIRED));
+      errorMap.put(keyPath, valueErrors.build());
```
jeanniefu, line 89: I switched the implementation here to build a normal hash map first, and then convert it to ImmutableMap. I didn't see a way to modify an already existing entry with ImmutableMap.Builder. Looks like calling ImmutableMap.Builder.put() does not overwrite existing entries, and just errors:
https://guava.dev/releases/20.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html#put-K-V-

But hashmap put is fine: https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#put-K-V-

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -363,8 +363,7 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
     Block thisBlockUpdated = thisBlockUpdatedMaybe.get();
 
     // Validation errors: re-render this block with errors and previously entered data.
-    if (thisBlockUpdated.hasErrors()
-        || thisBlockUpdated.hasRequiredQuestionsThatAreSkippedInCurrentProgram()) {
```
clouser2, line 367: Yay!

server/test/services/applicant/BlockTest.java
```diff
@@ -468,12 +468,11 @@ public void hasRequiredQuestionsThatAreSkippedInCurrentProgram_returnsTrue() {
         .map(ApplicantQuestion::getContextualizedPath)
         .forEach(path -> QuestionAnswerer.addMetadata(applicantData, path, programId, 1L));
 
-    assertThat(block.hasRequiredQuestionsThatAreSkippedInCurrentProgram()).isTrue();
+    assertThat(block.hasErrors()).isTrue();
```
clouser2, line 471: Do we have the ability to assert that the "this question is required" error appears in there or is there not enough visibility for that assertion?

clouser2 (approved):
Great cleanup!

server/test/services/applicant/BlockTest.java
jeanniefu: Good idea! Done

PR #3484, opened 2022-09-21
Add browser tests asserting that the "notify" checkbox is shown and an email is sent to the applicant

### LINKED ISSUE #3297: Status Tracking: Test email checkbox shows in the confirmation dialog when changing statuses
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

clouser2:
### Description

For local browser tests, we run localstack, which has an endpoint including all emails sent since the process has started. When running in a non-prober environment, we can assert the email contents and ensure one is sent when a status has been changed.

## Release notes:

Adding automated testing that emails are appropriately sent when an application status has been changed.

### Checklist

### Issue(s)

Fixes #3297


browser-test/src/support/index.ts
```diff
@@ -472,3 +476,35 @@ export const validateScreenshot = async (
...
+  }
+  Source: string
+  Subject: string
+}
+
+export const extractEmailsForRecipient = async function (
+  page: Page,
+  recipientEmail: string,
+): Promise<LocalstackSesEmail[]> {
+  if (TEST_USER_AUTH_STRATEGY !== 'fake-oidc') {
```
shanemc-goog, line 496: use supportsEmailInspection() ?

browser-test/src/support/index.ts
```diff
@@ -472,3 +476,35 @@ export const validateScreenshot = async (
...
+    throw new Error('Unsupported call to extractEmailsForRecipient')
```
shanemc-goog, line 497: Doc this constraint on the method.

browser-test/src/support/index.ts
```diff
@@ -472,3 +476,35 @@ export const validateScreenshot = async (
...
+  }
+  const originalPageUrl = page.url()
+  await page.goto('http://localstack:4566/_localstack/ses')
```
shanemc-goog, line 500: having the port hard coded seems fragile, is that our only option?

shanemc-goog (changes requested):
nice!

bin/run-browser-test-env
```diff
@@ -131,6 +131,9 @@ if truth::is_enabled both_emulators; then
   else
     compose_browser_test "${use_dev}" --profile aws up --no-recreate -d localstack
   fi
+  # Wait for localstack to start. This script also has the side effect of verifying the email
```
clouser2, line 134: FYI, there's a large opportunity for cleanup here, but I think a targeted fix makes a lot more sense here. Filed #3497 as tech debt.

browser-test/src/support/index.ts
```diff
@@ -498,15 +499,21 @@ export type LocalstackSesEmail = {
   Subject: string
 }
 
+/**
+ * Queries the emails that have been sent for a given recipient. This method requires that tests
+ * run in an environment that uses localstack sinc it captures the emails sent using SES and makes
```
shanemc-goog, line 504:
```suggestion
 * run in an environment that uses localstack since it captures the emails sent using SES and makes
```

browser-test/src/support/index.ts
```diff
@@ -498,15 +499,21 @@ export type LocalstackSesEmail = {
...
+ * them available at a well-known endpoint). An error is thrown when the method is called from an
+ * environment that does not use localstack.
```
shanemc-goog, line 506: Add that `supportsEmailInspection` can be used to determine that.

Similar to a map.get() method saying you should do map.has() first etc

shanemc-goog (approved):

browser-test/src/support/index.ts
clouser2: Done.

browser-test/src/support/index.ts
clouser2: Done.

PR #3486, opened 2022-09-21
Setting tests to fail if screenshots have any differences rather than the previous value of 1%

clouser2:
### Description

While here, there were two additional fixes due to needing screenshot updates:
  * Add a `cf-application-id` class that can be used to replace the dynamic application ID shown when the applicant submits an application. This is replaced prior to a screenshot being taken
  * Fixing issue where the new date format regexp introduced by #3447 errantly matched the date format from the previous regexp

If we determine that screenshots are flaky with threshold=0, we can increase it again. Hopefully, we can add code to deflake rather than increasing the threshold. Additionally, another source of very small rendering differences could be when the underlying Playwright binary updates the set of browsers it uses. A new version of Chromium wouldn't produce a large difference in screenshots, but it could produce a few pixels of difference, especially around anti-aliasing.

Note: There was a single screenshot failure `enumerator-only-active-snap.png` that seemed to have only 4 pixels of difference near rounded corners for the program card and the "Create new question" button. I'm suspicious that this is the anti-aliasing issue described above.

## Release notes:

N/A

### Testing

Ran CI on this PR 5 times and confirmed that there weren't any failures due to screenshot differences.

Run 1: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5037261281
Run 2: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5037729580
Run 3: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5038019647
Run 4: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5038283397
Run 5: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5038557633 (failed with standard flake, rerun was successful: https://github.com/civiform/civiform/actions/runs/3108260833/jobs/5038869395)
### Checklist


browser-test/src/support/index.ts
```diff
@@ -449,9 +449,17 @@ export const validateScreenshot = async (
       // regexp will break and it will show up in screenshots.
       date.textContent = date
```
jeanniefu, line 450: cool!

browser-test/src/support/index.ts
```diff
@@ -449,9 +449,17 @@ export const validateScreenshot = async (
       // regexp will break and it will show up in screenshots.
```
jeanniefu, line 449: This comment is for line 442-443 (idk why I can't click there)

nit: Move to line 446

jeanniefu (approved):
Should we mark this as "ignore for release"?

nbeloglazov (approved):

browser-test/src/support/index.ts
clouser2: Done. Also added a comment for the application ID replacement.

PR #3487, opened 2022-09-22
Fix call for program index page to properly filter to the set associated with the applicant

clouser2:
### Description
This was a regression introduced by #2695.

## Release notes:

Fixes issue where the applicant-facing program index page errantly categorized programs as in-progress / submitted if any applicant had an in-progress / submitted application for that program.

In addition, the application submission date and current application status could be shown for these programs, but no other application data would have been visible.

### Checklist

- [X] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [X] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)

Fixes #<issue_number>; Fixes #<issue_number>; Fixes #<issue_number>...

PR #3488, opened 2022-09-22
Adding a browser-based regression test for #3487

clouser2:
### Description

This adds coverage for asserting the set of programs shown on the applicant-facing programs page. We weren't previously able to assert this state since we didn't have support for logging in as a test user in CI tests. Now that we have that ability (from #3326), we can assert that one applicant's state doesn't interfere with another applicant's.

## Release notes:

Adding browser-based regression test for the issue fixed in https://github.com/civiform/civiform/pull/3487.

### Checklist

### Issue(s)


browser-test/src/applicant_application_index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    await applicantQuestions.clickNext()
+    await applicantQuestions.submitFromReviewPage()
+    await applicantQuestions.returnToProgramsFromSubmissionPage()
+    await applicantQuestions.expectPrograms({
+      notStartedPrograms: [otherProgramName],
+      inProgressPrograms: [],
+      submittedPrograms: [primaryProgramName],
+    })
+
+    // Logout, then login as guest and confirm that everything appears unsubmitted. This is a
```
shanemc-goog, line 80: I'd remove the "regression" part.  It's a completely valid and good test regardless. The regression only makes sense during this review, and likely isn't helpful after.

browser-test/src/support/applicant_questions.ts
```diff
@@ -202,6 +203,52 @@ export class ApplicantQuestions {
...
+    const notStartedProgramNames = await this.programNamesForSection(
+      'Not started',
+    )
+    const inProgressProgramNames = await this.programNamesForSection(
+      'In progress',
+    )
+    const submittedProgramNames = await this.programNamesForSection('Submitted')
+
+    // Sort results before comparing since we don't care about order.
+    notStartedProgramNames.sort()
```
shanemc-goog, line 229: The naming is a little off here since both are "Names" but only one includes it.

Use want and got instead?

shanemc-goog (approved):

browser-test/src/applicant_application_index.test.ts
clouser2: Totally agree that this is a valid and good test. That said, I also want to specifically call out that it's preventing us from backsliding on a potentially bad issue for context. I modified it to take out the word "regression" and put the issue in a parenthetical.

browser-test/src/support/applicant_questions.ts
clouser2: Done.

browser-test/src/applicant_application_index.test.ts
shanemc-goog: That statement applies to most of our tests though which makes it noisy.    It's why we have browser tests for the most part :)

IMO making the intent of the test clear is how we address this concern. I'd say this is a great reason to have focused tests even if setup is redundant; and it makes it readable going forward.  In another year of development, no one is going to have context on this regression versus the tests of all the new features.

up to you though.

shanemc-goog (approved):

browser-test/src/applicant_application_index.test.ts
clouser2: K, I think I'm going to keep this one since I think the severity of a regression warrants easily finding the context for why it was added in the first place.

PR #3498, opened 2022-09-23
Surface user-facing question text in the question bank and program builder

clouser2:
### Description

## Release notes:

### Checklist

### Issue(s)

### DIFF
--- a/browser-test/src/support/admin_programs.ts
+++ b/browser-test/src/support/admin_programs.ts
@@ -94,7 +94,7 @@ export class AdminPrograms {
 
   // Question card within a program edit page
   questionCardSelectorInProgramEditor(questionName: string) {
-    return `.cf-program-question:has(:text("${questionName}"))`
+    return `.cf-program-question:has(:text("Admin ID: ${questionName}"))`
   }
 
   // Question card within a program edit page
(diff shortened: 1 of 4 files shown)


browser-test/src/support/admin_programs.ts
```diff
@@ -308,10 +308,23 @@ export class AdminPrograms {
...
     for (const questionName of questionNames) {
-      await this.page.click(`button >> text="${questionName}"`)
+      await this.addQuestionFromQuestionBank(questionName)
     }
   }
 
+  async addQuestionFromQuestionBank(questionName: string) {
+    const questionBankElementLocator = this.page.locator(
+      '.cf-question-bank-element',
+      {has: this.page.locator(`text="Admin ID: ${questionName}"`)},
```
nbeloglazov, line 318: `.cf-question-bank-element:has-text("Admin ID: ...")`

nbeloglazov (approved):

browser-test/src/support/admin_programs.ts
clouser2: This one's fun. The DOM used here is:

```
<div class="cf-question-bank-element">
  <p>User-facing display text</p>
  <p>User-facing help text</>
  <p>Admin ID: internal-question-identifier</p>
  <button type="submit" class="classes-that-cause-the-button-to-fill-the-whole-parent-container-and-have-a-higher-z-index-than-the-p-tags-above" />
</div>
```

Previously, we duplicated the `<p>` text to the button itself. The Playwright code was clicking on that element itself, which seemed like a bit of a lie. Now, by using the `has` qualifier, we select the `<div>` element that contains the `Admin ID` element and click that whole element. The `<button>` still traps the click, but we don't have to think about the details of how the DOM is organized. If we were to try and use the `has-text` approach above, it would match the `<p>` tag. Then, when Playwright tried to click, the button would intercept it and cause a test failure.

I tried to refrain from refactoring the question bank DOM in this PR since I wasn't sure why that approach was added initially.

I'm going to merge as-is and happy to address any follow-on questions / feedback in a separate PR.

clouser2: You are 100% correct, whoops!

Sent out #3517 to fix.

Issue #3075, opened 2022-08-05
CiviForm admin UX: Consolidate "admin name and description" and "name and description visible to applicants" 


cassandralfernandes:
**Is your feature request related to a problem? Please describe.**
The CiviForm admin does not see a need for "admin name and description" and "name and description visible to applicants" to exist separately. From the user "Admin name field is really confusing and not necessary. It’s not clear that the notes field is text for admin only."

**Describe the solution you'd like**
Consolidate "admin name and description" and "name and description visible to applicants" and only one field called "admin notes" 

With this change, the public facing question name should become the searchable name in the question bank 

**Additional context**
From admin UXR feedback sheet

**screenshots**
<img width="1550" alt="Screen Shot 2022-08-05 at 2 14 11 PM" src="https://user-images.githubusercontent.com/99513234/183196754-bf84effa-a2e1-4b22-ab9e-5aac7677b4cb.png">

[mock here](https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13ec4604693_0_105)

clouser2:
Our internal / external identifiers for programs and questions are baked pretty far into the system. This change unfortunately won't be straightforward since we require the internal name to not change over the course of question or program. Using the external value in the program / question UI could lead to confusion since the external name can change over time. Additionally, if there's both an active version and a draft version (each with different values for the display name), it will become difficult to choose what to show.

This one still needs a fair amount of scoping on the eng side. Perhaps a small approach might be to better group the fields like we do for program creation/editing.

cassandralfernandes:
After meeting with Brett, Leyla, Shane, Bion, and Raha (notes [here](https://docs.google.com/document/d/1peZ5tPhleq7Gdu7PoHJ_g9kUH9k5cBb9LXKMQYSvBcw/edit)) , the proposed approach is as follows: 

- For programs, change the way the name field is described to "Craft your program's permanent URL. This will end up looking like <insert example URL> to the applicant". No data cleanup right now. Gray out the field so it's more clear to users that they cannot edit it 
- On the questions page, remove the furthest left column that displays the admin question name
- For questions, change the name of the description field to “admin notes” 

Remind me where we landed for the internal name field for questions? @rahaghas @clouser2

cassandralfernandes:
Related to #3076

clouser2:
[...]

PR #3500, opened 2022-09-23
Showing user-facing text in the visibility condition builder

clouser2:
### Description

In addition to showing the user-facing text in the visibility condition builder, we'll also to show the internal-facing identifier within the "Admin ID:" field (similar to #3498) in order to allow the CiviForm admin to differentiate questions with the same user-facing text.

Additionally, we dynamically build a string representation of the visibility condition. Here, I thought the internal facing identifier made more sense to use since it's not necessarily a multi-sentence value.

### Screenshots

Note: Added screenshot tests for the show/hide visibility conditions. Leaving snapshots of the before/after here for posterity.

Before:
Screen Shot 2022-09-26 at 10 14 59 AM

After:
Screen Shot 2022-09-26 at 10 14 27 AM

## Release notes:

Surfacing the user-facing text within the CiviForm Admin's visibility condition builder UI.

### Checklist

### Issue(s)
#3075


browser-test/src/support/admin_predicates.ts
```diff
@@ -17,7 +17,10 @@ export class AdminPredicates {
     operator: string,
     value: string,
   ) {
-    await this.page.click(`text="${questionName}"`)
+    const addQuestionButtonLocator = this.page.locator('button', {
+      has: this.page.locator(`:text("Admin ID: ${questionName}")`),
```
nbeloglazov, line 21: Is it different from `this.page.locator('button:has-text("Admin ID...")')`?

nbeloglazov (approved):

browser-test/src/support/admin_predicates.ts
clouser2: Good call. Done.

PR #3502, opened 2022-09-24
use dev-oidc in oidc-compliant mode by default, allow logouts

kmcnellis:
### Description

- Change the CIVIFORM_APPLICANT_IDP in bin/run-dev to the generic-oidc implementation.  
- Sets up an additional dev-oidc client that follows the standard OIDC-compliant flow (vs the IDCS version).
- Allows the testing of logout and logging back in to the dev server
- Renames the fake client-id and client-secret 
- Delete the unused admin_idp conf variable

Tested by running bin/run-dev and performing login and logout

## Release notes:

n/a

### Checklist


docker-compose.yml
```diff
@@ -64,21 +64,21 @@ services:
...
       - APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM
-      - APPLICANT_OIDC_LOGOUT_CLIENT_PARAM
+      - APPLICANT_OIDC_LOGOUT_CLIENT_PARAM=client_id
+      - APPLICANT_OIDC_OVERRIDE_LOGOUT_URL
       - APPLICANT_OIDC_PROVIDER_NAME
-      - APPLICANT_OIDC_CLIENT_ID
-      - APPLICANT_OIDC_CLIENT_SECRET
-      - APPLICANT_OIDC_DISCOVERY_URI
+      - APPLICANT_OIDC_CLIENT_ID=standard
+      - APPLICANT_OIDC_CLIENT_SECRET=bar
```
clouser2, line 80: Not blocking: I'd consider a different value here such as `fake-oidc-client-secret` or something along those lines. That way if something ends up in logs, it's easier to find the reference than searching for `bar`.

server/conf/application.conf
```diff
@@ -71,8 +71,6 @@ auth {
   # Configure which IDP to use for applicant and admin authentication.
   applicant_idp = "idcs"
   applicant_idp = ${?CIVIFORM_APPLICANT_IDP}
-  admin_idp = "adfs"
-  admin_idp = ${?CIVIFORM_ADMIN_IDP}
```
clouser2, line 75: Why is it safe to remove this line?

docker-compose.yml
```diff
@@ -64,21 +64,21 @@ services:
       - LOGIN_RADIUS_PRIVATE_KEY_PASS
       - LOGIN_RADIUS_KEYSTORE_PASS
       - LOGIN_RADIUS_KEYSTORE_NAME
-      - CIVIFORM_APPLICANT_IDP
-      - CIVIFORM_ADMIN_IDP
+      - CIVIFORM_APPLICANT_IDP=generic-oidc
```
clouser2, line 67: Can you expand the PR description with some comments about what this unlocks? If possible, mind adding a quick "Tested" section that describes how this works in `bin/run-dev`?

clouser2 (approved):

docker-compose.yml
kmcnellis: Good idea!

server/conf/application.conf
kmcnellis: We don't actually use it anywhere.  We only have one option for admin identity providers (adfs).  It was confusing/misleading to provide it as an option

docker-compose.yml
kmcnellis: Done

PR #3532, opened 2022-09-27
Remove jest-playwright-preset since we don't make use of it

### LINKED ISSUE #3513: Consider moving away from jest-playwright-preset in favor of Playwright's builtin runner
Playwright is a library that allows interacting with a webpage through a headless browser. Jest is a test runner. `jest-playwright-preset` is a library that makes it possible to use the Jest test runner and interact with the browser using Playwright.

The jest-playwright-preset docs now suggest using Playwright's builtin test runner (https://www.npmjs.com/package/jest-playwright-preset), which presumably was recently built.

As part of https://github.com/civiform/civiform/issues/3302, we ran into issues where Jest 29 was updated, but `jest-playwright-preset` wasn't compatible with it immediately. It's unclear whether this will be resolved at some point. Either way, carrying this forward will always be a source of potential time sink. Given the warning on the `jest-playwright-preset`, it might make sense to try and use Playwright's native library instead.

I'll fill out a pointers as to what this migration might look like in future comments.

clouser2:
### Description

Context: https://github.com/civiform/civiform/issues/3513#issuecomment-1258671364 onwards.

## Release notes:

N/A

### Checklist

### Issue(s)

Fixes #3513


nbeloglazov (approved):
Interesting. Looks like transitively it removes ~12% of all dependencies

bion (approved):

PR #3541, opened 2022-09-28
Replace sbt-web with custom task

nbeloglazov:
### Description

[sbt-web](https://github.com/joost-de-vries/sbt-typescript) plugin we are using is no longer maintained and 5 years old. It uses Typescript 2 while current one is Typescript 4. Which means we can't use latest features and improvements. 

This PR removes sbt-web and instead adds custom TypescriptBuilder.scala file, similar to TailwindBuilder.scala. That file defines a task that calls `tsc` (typescript compiler) to compile our few TS files. Code is a bit complicated due to sbt APIs which aren't very straightforward. 

That task is hooked into sbt-web pipeline in `build.sbt` and runs before digest/gzip tasks. In dev mode, upon recompilation TS files will be recompiled only if some of them was changed. TS files are not recompiled on java file updates.

Additionally this PR updates TS version to 4.8. I couldn't find a way to split these two changes. 

### Checklist

### DIFF
--- a/server/project/TypescriptBuilder.scala
+++ b/server/project/TypescriptBuilder.scala
@@ -27,7 +27,7 @@ import scala.sys.process.{Process, ProcessLogger}
   * pipeline stage. Read about Asset Pipeline tasks
   * https://github.com/sbt/sbt-web#asset-pipeline The task implements Asset
   * Pipeline task and not Source File task because the latter doesn't allow us
-  * to keep ∂irectory structure: we want that file `assets/javascripts/foo.ts`
+  * to keep directory structure: we want that file `assets/javascripts/foo.ts`
   * be compiled to `assets/javascripts/foo.js`. So we are using Pipeline.Stage
   * for that.
   *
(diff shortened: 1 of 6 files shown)


bion (changes requested):
@nbeloglazov please re-request review when presubmit checks are passing

nbeloglazov:
Sorry, forgot to remove comma from JSON file. Passing now

server/project/TypescriptBuilder.scala
```diff
@@ -0,0 +1,114 @@
...
+  */
+object TypescriptBuilder extends AutoPlugin {
+  override def trigger = allRequirements
+
+  object autoImport {
+    val compileTypescript = taskKey[Pipeline.Stage](
+      "Compiles typescript from assets/javascript folder"
+    )
+  }
+  import autoImport._
```
bion, line 45: suggestion (blocking): most people reading this are going to be pretty unfamiliar with sbt and scala, a little more commenting may be helpful. Maybe mention where the entrypoint is for code execution in the plugin, for example

bion (changes requested):

bion (approved):
Thanks for the big improvement in documenting this!

server/project/TypescriptBuilder.scala
nbeloglazov: Added more comments. Agree that scala and sbt are pretty unfriendly to noobies like me due the amount of DSL they rely.

Issue #3548, opened 2022-09-28
Title for login page is not localized


clouser2:
**Describe the bug**
We hardcode a value of "Login" for the page title [here](https://github.com/civiform/civiform/blob/d818b469c9030f1a63a03f640c672a8e0337ad26/server/app/views/LoginForm.java#L73).

The assumption may have been that this page would only be shown prior to language selection. However, an applicant can:
 1) Continue as guest
 2) Choose another language
 3) Navigate back to the front page (maybe by logging out).
 4) The login page is rendered with localized text

We should likely localize this text, similar to the other applicant-facing page titles.

Noticed as part of #3545.

jeanniefu:
Currently the front page is translated, and the language selection page is not.

Should we even translate the front page? In my opinion, it makes sense for the front page + language selection page to both be in English. My expectation is that when I click logout, the website is returned to its default state (English).

nbeloglazov:
It think we can best-guess user language even before they select language based on `accept-language` header. It provides languages user likely uses. For example here mine: `accept-language: be,en-GB;q=0.9,en;q=0.8,en-US;q=0.7,ru;q=0.6,la;q=0.5` Which states that my 2 preferred languages are Belarusian and English GB (no idea why GB over US but ok). If do that than localizing first screen makes more sense.

MichaelZetune:
https://user-images.githubusercontent.com/30369272/220974946-f3a5d781-cd58-41c4-b559-6968c318d7e9.mp4

It looks like we already pre-select the correct language for the user based on the `accept-language` header. This video demonstrates that with the ModHeader chrome extension.

Changing the content of the log in and language page based on this header (even before the language is confirmed) seems reasonable to me. Will have to check with PMs on whether we should proceed. #4269 just adds the log in title translation for now.

PR #3549, opened 2022-09-28
Fix login / select language page titles to be more accessible

### LINKED ISSUE #3545: (WCAG 2.1) Login and language selection page titles are not well defined

**WCAG SC**: [2.4.2. Page Titled](https://www.w3.org/TR/WCAG21/#page-titled)

**Observed results**: The Login (Login) and Select language (Application Information: CiviForm) pages do not have proper page titles with well-defined context.

**How to fix**: Currently the titles read "Login" and "Application Information: CiviForm". We should update the text to "Login - CiviForm" and "Select Language - CiviForm".

clouser2:
#### Login page
We had logic to suffix the page title with text depending on the layout (e.g. `- CiviForm` for applicants and `- CiviForm Admin Console` for program/CiviForm admins).

Our login page wasn't make use of either of those specific layouts. This pushes the "suffix the page title" logic down to the `BaseHtmlLayout` class with a default value of "CiviForm".

Note: While here, I also filed #3548 to track the fact that the "Login" text is not localized.

#### Select language
This had an incorrect set of text. Since it wasn't localized, I also removed it from the translations files and inlined it directly with a comment. Finally, I moved the `<h1>` SR-only tag above the content that it was describing for clarity.

## Release notes:

Updating the titles for the login page and language selection page to be more accessible.

### Checklist

### Issue(s)

Fixes #3545

### DIFF
--- a/server/app/views/applicant/ApplicantInformationView.java
+++ b/server/app/views/applicant/ApplicantInformationView.java
@@ -73,15 +73,13 @@ public class ApplicantInformationView extends BaseHtmlView {
         submitButton(submitText).withClasses(ApplicantStyles.BUTTON_SELECT_LANGUAGE);
     formContent.with(formSubmit);
 
+    String title = "Select lanaguage";
     HtmlBundle bundle =
         layout
             .getBundle()
-            .setTitle(messages.at(MessageKey.CONTENT_APPLICANT_INFORMATION.getKeyName()))
+            .setTitle(title)
             .addMainStyles(ApplicantStyles.MAIN_APPLICANT_INFO)
-            .addMainContent(formContent);
-    bundle.addMainContent(
-        h1(messages.at(MessageKey.CONTENT_APPLICANT_INFORMATION.getKeyName()))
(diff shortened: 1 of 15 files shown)


server/app/views/applicant/ApplicantInformationView.java
```diff
@@ -73,15 +73,15 @@ public Content render(
         submitButton(submitText).withClasses(ApplicantStyles.BUTTON_SELECT_LANGUAGE);
     formContent.with(formSubmit);
 
+    // No translation needed since this appears before applicants select their preferred language,
+    // so we always use the default.
+    String title = "Select lanaguage";
```
jeanniefu, line 78: nit: typo in language

jeanniefu (approved):
Thanks for picking this up and fixing!!

server/app/views/applicant/ApplicantInformationView.java
clouser2: Good find!

PR #3551, opened 2022-09-28
Fail on new images being added

nbeloglazov:
### Description

jest-image-snapshot automatically adds image snapshot if it's missing during test run. That's good for local development but for CI it's undesired: we want test to fail in cases of missing snapshots. This PR adds extra step to github action workflow which ensures that no screenshots being added after test run.

Here is how failure looks:

image

### Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -71,7 +71,7 @@ jobs:
         run: bin/run-browser-test-env --aws --ci
       - name: Run browser tests with Localstack
         run: bin/run-browser-tests-ci-batch ${{ matrix.batch_number }}
-      - name: Check no new image snapshots
+      - name: Verify no new image snapshots added
         run: git add browser-test/image_snapshots ; git diff --compact-summary --exit-code HEAD
       - name: Upload image diff outputs
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


clouser2 (approved):

.github/workflows/tests.yaml
```diff
@@ -71,6 +71,8 @@ jobs:
         run: bin/run-browser-test-env --aws --ci
       - name: Run browser tests with Localstack
         run: bin/run-browser-tests-ci-batch ${{ matrix.batch_number }}
+      - name: Check no new image snapshots
+        run: git add .; git diff --exit-code HEAD
```
gwendolyngoetz, line 75: How does this know to fail just on the snapshot images?

nbeloglazov: I doesn't know that. And it fails on any diff. Given that we don't expect diffs it should be reasonable? Though I can add only `browser-test/image_snapshots` folder. More problematic though that it fails with not very readable message. Like if I saw that error I wouldn't know why it failed. Need to see if we can prettify it somehow.

nbeloglazov: Ok, I think now it's better. As you suggested it checks only image snapshots and also uses compact diff format.

gwendolyngoetz (approved):

PR #3595, opened 2022-09-30
Add Status management image diffs

shanemc-goog:
### Description

Add image diffs for status management page.

## Release notes:

NA - testing

### Checklist

### Issue(s)


browser-test/src/civiform_admin_program_statuses.test.ts
```diff
@@ -38,6 +41,15 @@ describe('modify program statuses', () => {
       await ctx.adminPrograms.gotoDraftProgramManageStatusesPage(programName)
     })
 
+    it('renders create new status model', async () => {
```
nbeloglazov, line 44: nit: modAl

nbeloglazov (approved):

browser-test/src/civiform_admin_program_statuses.test.ts
shanemc-goog: done!

shanemc-goog:
PTAL, I accidentally muddied the PR with similar changes for another file and can't undo it.

nbeloglazov:
I see that we are are capturing applicant ID in screenshots which is not stable. In cases like these we mark such DOM elements with class like [`.cf-applicants-id`](https://github.com/civiform/civiform/blob/db5a39a18392beb1b8d1f72dd3fa5dd0bc324f98/server/app/views/style/ReferenceClasses.java#L134) which later [replaced with hardcoded value](https://github.com/civiform/civiform/blob/main/browser-test/src/support/index.ts#L458) when screenshot is taken. I think you'll have to do the same thing for applicant id in your screensshots.

nbeloglazov (approved):

shanemc-goog:
> I see that we are are capturing applicant ID in screenshots which is not stable. In cases like these we mark such DOM elements with class like [`.cf-applicants-id`](https://github.com/civiform/civiform/blob/db5a39a18392beb1b8d1f72dd3fa5dd0bc324f98/server/app/views/style/ReferenceClasses.java#L134) which later [replaced with hardcoded value](https://github.com/civiform/civiform/blob/main/browser-test/src/support/index.ts#L458) when screenshot is taken. I think you'll have to do the same thing for applicant id in your screensshots.

I think you saw my update already, but I had to make the normalizing code also process sub frames, which is why the screenshots weren't processing the application view subframe.

PR #3615, opened 2022-10-05
Add padding to organization's logo on applicants page

### LINKED ISSUE #3613: Logo with text should have padding
**Describe the bug**
Logo is rendered in the top left corner with no padding. If logo image itself doesn't have any padding - it doesn't look good: 
image

Instead we should add some padding by default. And support images of different sizes. Currently image size hardcoded to 175x70px. 

Proposed rendering with padding:
image

nbeloglazov:
### Description

That way logo is not expected to be padded and looks reasonable without padding. Organizations can add extra padding if they want.

Before:
image

After:
image

### Checklist

### Issue(s)

Fixes #3613


bion:
@nbeloglazov  Have you checked in with Bloomington about this change?

nbeloglazov:
I did not. Here is how it will look for Bloomington: 
image
becomes
image

I think Bloomington should update their logo for that to have blue background (and maybe white flower and text). But overall they have the same issue (logo sits in the top-left corner). I'm going to reach out to them on chat.

nbeloglazov:
Asked Bloomington. They also use logos fully stripped with no paddings.

bion (approved):

PR #3622, opened 2022-10-06
fix style conflict in admin application cards

syntapy:
### Description

Fix another style conflict that breaking a style in https://github.com/civiform/civiform/pull/2630

### DIFF
--- a/server/app/assets/javascripts/admin_applications.ts
+++ b/server/app/assets/javascripts/admin_applications.ts
@@ -64,14 +64,13 @@ class AdminApplications {
   viewApplication(selectedCard: Element, applicationUrlPath: string) {
     // Remove selection style from previously selected card.
     this.cards.forEach((card) => {
-        this._assertNotNull(card.children[0], 'card inner div').classList.remove(
-          AdminApplications.BACKGROUND_GRAY_CLASS,
-        )
-        this._assertNotNull(card.children[0], 'card inner div').classList.add(
-          AdminApplications.BACKGROUND_WHITE_CLASS,
-        )
-      }
-    )
+      this._assertNotNull(card.children[0], 'card inner div').classList.remove(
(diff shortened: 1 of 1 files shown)


server/app/assets/javascripts/admin_applications.ts
```diff
@@ -68,12 +69,23 @@ class AdminApplications {
       ),
     )
 
+    this.cards.forEach((card) =>
+      this._assertNotNull(card.children[0], 'card inner div').classList.add(
+        AdminApplications.BACKGROUND_WHITE_CLASS,
```
nbeloglazov, line 74: Let's add it in function on line 67:

```
const child = this._assertNotNull(card.children[0], 'card inner div');
child.classList.remove(AdminApplications.BACKGROUND_GRAY_CLASS);
child.classList.add(AdminApplications.BACKGROUND_WHITE_CLASS);
```

nbeloglazov (changes requested):

server/app/assets/javascripts/admin_applications.ts
syntapy: done

nbeloglazov: You could still create local `child` variable to avoid unnecessary code duplication, but it's not that important.

syntapy: done

nbeloglazov (approved):

server/app/assets/javascripts/admin_applications.ts
nbeloglazov: You still call assertNonNull twice on the same element. That's reduntant and calling it once is enough to narrow down type to be non-nullable: `const child = this._assertNotNull(card.children[0], 'card inner div');` here `child` is a non-nullable and you can call methods on it without Typescript complaining that you element might be null

PR #3625, opened 2022-10-07
shrink and compress Chief Seattle image

syntapy:
### Description

Reduces size of chief seattle image from 204k to 45k by shrinking from 3300x3300 pixels to 500x500 pixels using GIMP and LoHalo interpolation

This reduces time to interactive from average 5.4s to 5.1s over 3 lighthouse runs

## Release notes:

Reduce site time to interactive by shrinking chief seattle image size


server/app/views/ViewUtils.java
```diff
@@ -75,10 +75,14 @@ LinkTag makeLocalCssTag(String filename) {
         .withRel("stylesheet");
   }
 
-  public ImgTag makeLocalImageTag(String filename) {
+  public ImgTag makeLocalImageTagPNG(String filename) {
     return img().withSrc(assetsFinder.path("Images/" + filename + ".png"));
```
nbeloglazov, line 79: I would actually keep `makeLocalImageTag` name and instead passed filename with extension:

.makeLocalIImageTag("ChiefSeattle_Blue.png")

I don't think that passing name without extension saves much space.

nbeloglazov:
I agree that having 3000x3000 logo is overkill. I think we can bring it down to 500x500 even. Not sure about webp though. Is webp smaller significantly smaller compared to png of the same size? Working with webp might be harder (fewer image editors support AFAIK).

syntapy:
Shrinking from 3300x3300 to 500x500 and keeping as png reduces size from 204K to 45K

After that, converting to webp further reduces size down to 22K

From my last test, reducing from 204K to 35K (169K difference) elliminated an average of 1s from time to interactive, so going from 45K to 22K would probably shave 0.13s  off time to interactive

So it doesn't seem too much a difference, but is nicer to have a smaller image size nonetheless

I don't know anything about the image editors or how this would effect UI/UX team, so please let me know whether to use webp or png

server/app/views/ViewUtils.java
syntapy: This is a non-issue now that we're not using webpack

nbeloglazov (approved):

nbeloglazov:
Looks like you need to update some screenshots.

syntapy:
For some reason shrinking the image to 500x500 causes the final rendered image to shrink. But shrinking to 700x700 causes the final image to shrink less. I will need to trouble shoot that

landing-page-diff

PR #3639, opened 2022-10-12
Adding default state to the Address Question

### LINKED ISSUE #3630: Default address does not display
**Describe the bug**
As a CiviForm Admin, I'd like to be able to set "Washington" as the default state for all address questions. This would prevent erroneous address entries (WA, Wash, Washington, Wa$hingt0n) and ensure a consistent address export. The default, however, does not display within the forms. It also appears the Address question type does not save any changes to the default state.

**To Reproduce**

_As a CiviForm Admin_
1. Go to an address question type (i.e. Mailing Address)
2. Click on 'Default State'
3. Select 'Washington'
4. Save
5. Republish

_As a resident_
6. Try to complete the form as a Resident and see that the default state does not show.

_As CiviForm Admin_
7. See that the address question type did not save the selected default state.

**Expected behavior**
I would expect the selected default state to show on any forms that use an address question type. I would expect that the selected default state would save on the question.
[Default Address not visible.pdf](https://github.com/civiform/civiform/files/9758806/Default.Address.not.visible.pdf)

**Screenshots**
Attached

**Additional context**
The Default State would only show for future forms. For example, we won't be back and fix any inconsistent data prior to the default state being set. But any forms published or republished with a default state address question component should show going forward.

swatkat1:
### Description
When the address question is configured and a default state is chosen, then the applicants must see the default state entered in their application. 

## Release notes:

When a default state is selected for an addressed and published as part of a program, than the default state value (eg.Washington) must appear when an applicant fills the address question.

### Checklist

### Issue(s)

Fixes #3630 ; 

### Screenshots
image


bion (changes requested):
Please re-request review when tests are passing. Please update PR title to be more descriptive.

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -74,8 +74,11 @@ protected DivTag renderInputTags(
         FieldWithLabel.input()
             .setFieldName(addressQuestion.getStatePath().toString())
             .setLabelText(messages.at(MessageKey.ADDRESS_LABEL_STATE.getKeyName()))
+            .setValue(
+                addressQuestion
+                    .getDefaultState()
+                    .orElse(addressQuestion.getStateValue().orElse("")))
```
nbeloglazov, line 80: Shouldn't be the other way around? 

`addressQuestion.getStateValue().orElse(adressQuestion.getDefaultState().orElse(""))`

server/app/views/admin/questions/QuestionConfig.java
```diff
@@ -293,7 +294,10 @@ public DivTag getContainer() {
    */
   private static ImmutableList<SelectWithLabel.OptionValue> stateOptions() {
     return ImmutableList.of(
-        SelectWithLabel.OptionValue.builder().setLabel("-- Leave blank --").setValue("-").build(),
-        SelectWithLabel.OptionValue.builder().setLabel("Washington").setValue("WA").build());
+        SelectWithLabel.OptionValue.builder().setLabel("-- Leave blank --").setValue("").build(),
+        SelectWithLabel.OptionValue.builder()
+            .setLabel("Washington")
+            .setValue("Washington")
```
nbeloglazov, line 300: Why did you change value from "WA" to "Washington"? Are you we using full state names in address question?

server/app/services/question/types/AddressQuestionDefinition.java
```diff
@@ -72,16 +72,24 @@ public static AddressValidationPredicates parse(String jsonString) {
...
+    public static AddressValidationPredicates create(
+        boolean disallowPoBox, Optional<String> defaultState) {
+      return builder().setDisallowPoBox(disallowPoBox).setDefaultState(defaultState).build();
     }
 
     @JsonProperty("disallowPoBox")
     public abstract Optional<Boolean> disallowPoBox();
 
+    @JsonProperty("defaultState")
+    public abstract Optional<String> defaultState();
```
nbeloglazov, line 91: Comment for @bion `defaultState` doesn't feel like it should be part of `AddressValidationPredicates` as it's not a validation. Should this field be set as part of different class?

server/app/services/applicant/question/AddressQuestion.java
```diff
@@ -212,4 +212,9 @@ public ImmutableList<Path> getAllPaths() {
     return ImmutableList.of(
         getStreetPath(), getLine2Path(), getCityPath(), getStatePath(), getZipPath());
   }
+
+  public Optional<String> getDefaultState() {
+    AddressQuestionDefinition definition = getQuestionDefinition();
+    return definition.getDefaultState();
```
nbeloglazov, line 218: Personally I don't think that method is useful. Callers can always get question definition themselves and retrieve default state from there. And default state is semantically part of QuestionDefinition and not Question itself (values that user input).

server/app/assets/javascripts/preview.ts
```diff
@@ -84,6 +86,19 @@ class PreviewController {
...
+    const defaultState = document.getElementById(
+      PreviewController.QUESTION_ADDRESS_DEFAULT_STATE,
+    ) as HTMLInputElement | null
+    if (defaultState) {
+      defaultState.addEventListener(
+        'input',
+        (ev) => {
+          PreviewController.updateDefaultStateText(defaultState.value)
+        },
+        false,
```
nbeloglazov, line 98: Don't need to pass `false` here. This field ([useCapture](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)) defaults to false anyway.

browser-test/src/support/applicant_questions.ts
```diff
@@ -47,6 +47,39 @@ export class ApplicantQuestions {
...
+    await this.page.waitForSelector('.cf-address-zip input')
+
+    // Check values are equal to expected
+    await this.validateInputValue(street, '.cf-address-street-1 input')
+    await this.validateInputValue(line2, '.cf-address-street-2 input')
+    await this.validateInputValue(city, '.cf-address-city input')
+    await this.validateInputValue('Washington', '.cf-address-state input')
+    await this.validateInputValue(zip, '.cf-address-zip input')
+  }
+  async answerDefaultStateAddressQuestion(
```
nbeloglazov, line 70: nit: you can update `answerAddressQuestion` above to call this new method and set only "state" as extra field there.

browser-test/src/support/applicant_questions.ts
```diff
@@ -47,6 +47,39 @@ export class ApplicantQuestions {
     await this.validateInputValue(zip, '.cf-address-zip input')
   }
 
+  async checkDefaultAddressQuestionValue(
```
nbeloglazov, line 50: To be more precise this method should say somehting like "checkAddressQuestionValueWithDefaultStateWashington" because this method checks that state is set to Washington. But I don't think that having it as a separate method warranted. You can use "checkAddressQeustionValue" and pass state as "Washington" to it and add a comment there saying that it verifires that default state is shown. Like

```

      await applicantQuestions.checkDefaultAddressQuestionValue(
        '1234 St',
        'Unit B',
        'Sim',
        // Should be set to default state
        'Washington', 
        '54321',
      )
```

nbeloglazov (changes requested):

server/app/views/questiontypes/AddressQuestionRenderer.java
swatkat1: No we should use Default state first. The reason being applicants are using- WA, Wash, Washington, Wa$hingt0n many cretive ways to fill the form. We should first check of the admin has set a default value, if yes, use it.. or not, use the value in the system (which is usually empty)

server/app/views/admin/questions/QuestionConfig.java
swatkat1: requirement as per- https://github.com/civiform/civiform/issues/3630

browser-test/src/support/applicant_questions.ts
swatkat1: oh nice call! will change!

server/app/views/questiontypes/AddressQuestionRenderer.java
nbeloglazov: Doesn't it mean that if I (applicant) selected "California", saved the answer and then came back to it - wouldn't I see "Washington" instead of "California"? Essentially your change will *always* show default state even if applicant answered a different state?

swatkat1: Yes that is exactly the expected behavior. If the admin has set that this program is only for Washington residents and wants to address to be in-state, than, the the state must by Washington. They cannot be out state addresses. This will also help reduce noise in the admin side.

nbeloglazov: That's not how I thought "default state" works at all. I thought that "default state" is just that - default. With an option to change it. If we want to set state to be unchangeable (always be "Washington") - then we should make the State field uneditable. So that user can't change it at all. That should be quite simple change - set it to readonly: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

Additionally we should rename the field "Default state" in admin question page to something like "Force state" or something similar.

server/app/services/question/types/AddressQuestionDefinition.java
bion: Agreed, the default state is unrelated to validation predicates and should not be stored here.

bion: Not quite, when fields are added on QuestionDefinition, Question needs to be updated to know how to persist them: https://github.com/civiform/civiform/blob/main/server/app/models/Question.java#L241

Rather than adding a field that's specific to one question type (`AddressQuestionDefinition`, in this case), a more general approach could be to add a new inner class on `QuestionDefinition` similar to `ValidationPredicates` that also serializes to JSON, and then inherit from it in an inner class in `AddressQuestionDefinition`, similar to how `AddressValidationPredicates` inherits from `ValidationPredicates`.

swatkat1: @bion -  i can do it but it will increase the size and scope of PR a lot. I will be again stuck with a huge PR and unable to checkin. Is there a way to break it up?

nb1701:
@swatkat1 Is this still in progress?

PR #3648, opened 2022-10-14
use dev-oidc in oidc-compliant mode by default, allow logouts

kmcnellis:
### Description

- Change the CIVIFORM_APPLICANT_IDP in bin/run-dev to the generic-oidc implementation.  
- Sets up an additional dev-oidc client that follows the standard OIDC-compliant flow (vs the IDCS version).
- Allows the testing of logout and logging back in to the dev server
- Renames the fake client-id and client-secret 
- Delete the unused admin_idp conf variable

Tested by running bin/run-dev and performing login and logout

## Release notes:

n/a

### Checklist

### Issue(s)

Moved from  #3502


nbeloglazov (approved):

browser-test/src/landing_page.test.ts
```diff
@@ -1,9 +1,10 @@
-import {createTestContext} from './support'
+import {createTestContext, validateScreenshot} from './support'
 
 describe('the landing page', () => {
   const ctx = createTestContext()
 
   it('it has login options', async () => {
     expect(await ctx.page.textContent('html')).toContain('Continue as guest')
+    await validateScreenshot(ctx.page, 'landing_page')
```
nbeloglazov, line 8: For consistency use "-" between words in screenshot name. Also we have validation in validateScreenshot for that. I'm not sure why it didn't fail for you.

browser-test/src/support/index.ts
```diff
@@ -231,7 +231,17 @@ export const gotoEndpoint = async (page: Page, endpoint: string) => {
...
   await page.click('text=Logout')
-  await page.click('button[name="logout"]') // OIDC confirmation page.
+  try {
+    const pageContent = await page.textContent('html', {timeout: 100})
+    if (pageContent!.includes('Do you want to sign-out from')) {
+      // OIDC central provider confirmation page
+      await page.click('button:has-text("Yes")')
+    }
+  }
+  catch (e) {
```
nbeloglazov: Is error here expected/normal?

kmcnellis: No, but the innertext selector could theoretically return null/reject the promise. (aka it never loads).  
We do the same check in https://github.com/civiform/civiform/blob/5d7c2d5e09b699258994606620023ed9c235208c/browser-test/src/support/admin_api_keys.ts

It doesn't really matter, though.  The follow-up check "page.waitForURL" would catch the failure to logout

browser-test/src/landing_page.test.ts
kmcnellis: The regex only was testing for presence of those characters, not that the entire string matched it.  Fixed the regex and the name

nbeloglazov: Ugh, thanks. I thought I tested it when this check was added originally. But probably not well enough.

PR #3656, opened 2022-10-15
Deduplicate browser test scripts

nbeloglazov:
### Description

Today we have 2 scripts that run browser tests: `bin/run-browser-tests-local` and `wait_for_server_start_and_run_tests.sh`. They do basically the same thing with the only difference in server urls they use to call CiviForm. This PR updates `bin/run-browser-tests-local` to call `wait_for_server_start_and_run_tests.sh` and have the latter the meaty logic of running tests. 

Additionally removes --debug option from `wait_for_server_start_and_run_tests.sh`. That option outputs extra API logs which I don't believe we use. We have different debugging approach described in https://docs.civiform.us/contributor-guide/developer-guide/testing#debugging-browser-tests that allows to run playwright inspector and step-through the code.

If some engineer will need to go deep into playwright debugging they can read guides like this: https://www.browserstack.com/guide/playwright-debugging and figure out which knobs need to be set.

### Checklist

### DIFF
--- a/bin/run-browser-tests-local
+++ b/bin/run-browser-tests-local
@@ -3,34 +3,11 @@
 # DOC: Run the browser tests locally. Requires browser test env already running.
 
 source bin/lib.sh
-export BASE_URL=http://localhost:9999
+# Disable screenshots for local tests because it will be use browser on current
+# machine which can be different from the browser used by browser-test docker
+# image and produce different screenshots.
 export DISABLE_SCREENSHOTS=true
-export TEST_USER_AUTH_STRATEGY=fake-oidc
-export TEST_USER_LOGIN=testuser
-export TEST_USER_PASSWORD=anotsecretpassword
-# The display name returned by test_oidc_provider.js is <username>@example.com.
-export TEST_USER_DISPLAY_NAME=testuser@example.com
+export BASE_URL=http://localhost:9999
 
-START_TIME=$(date +%s)
-DEADLINE=$(($START_TIME + 500))
-
-if ! output="$(node -v)"; then
-  echo output
-  echo "You must have node installed locally to run this command. Go to https://nodejs.org/en/download/package-manager/ for installation instructions."
-  exit 1
-fi
-
-npm install --force --prefix browser-test --quiet
-
-echo "Polling to check server start"
-
-until $(curl --output /dev/null --silent --head --fail --max-time 2 "${BASE_URL}"); do
-  if (($(date +%s) > "${DEADLINE}")); then
-    echo "Deadline exceeded waiting for server start"
-    exit 1
-  fi
-done
-
-echo "Detected server start"
-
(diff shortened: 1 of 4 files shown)


bin/run-browser-tests-local
```diff
@@ -3,34 +3,11 @@
...
-export TEST_USER_AUTH_STRATEGY=fake-oidc
-export TEST_USER_LOGIN=testuser
-export TEST_USER_PASSWORD=anotsecretpassword
-# The display name returned by test_oidc_provider.js is <username>@example.com.
-export TEST_USER_DISPLAY_NAME=testuser@example.com
-
-START_TIME=$(date +%s)
-DEADLINE=$(($START_TIME + 500))
-
-if ! output="$(node -v)"; then
```
shanemc-goog, line 17: Can we keep the node requirement just in the place that requires it?  It's not common to run -local so as is we'll require people to install it when they may never need it.

browser-test/bin/wait_for_server_start_and_run_tests.sh
```diff
@@ -4,15 +4,37 @@ set -euo pipefail
...
+export TEST_USER_DISPLAY_NAME=testuser@example.com
+
+if ! output="$(node -v)"; then
+  echo output
+  echo "You must have node installed locally to run this command. Go to https://nodejs.org/en/download/package-manager/ for installation instructions."
+  exit 1
+fi
+
+# Sanity check that this script was called from within browser-test directory.
+if [ ! -d "image_snapshots" ]; then
```
shanemc-goog, line 25: is image_snapshots correct? the comment and error messages say browser-test

shanemc-goog (changes requested):

browser-test/bin/wait_for_server_start_and_run_tests.sh
nbeloglazov: Clarified comment. I'm checking that image_snapshots directory exists as a proxy for checking from which directory the script is called. I can't check name of the current directory as in some docker runs we map browser-test to a differently named directory.

bin/run-browser-tests-local
nbeloglazov: This is the place that requires it. When we call npm few lines down - it requires node present. When engineers run this script from run-browser-test - the file is executed from within docker and docker environment has node. So engineers don't need to have node unless they run -local.

nbeloglazov: Those two scripts also require and use node. Because they use npm. Previously they didn't do any checks for node presence and assumed that node/npm was always present.

nbeloglazov: The browser-test image (which is based on [playwright image](https://github.com/civiform/civiform/blob/main/browser-test/playwright.Dockerfile#L2)) comes with node preinstalled.

shanemc-goog (approved):

Issue #3659, opened 2022-10-18
Standardize button UI across all pages


nbeloglazov:
CiviForm has 2 types of buttons. Old rectangular ones and new rounded buttons. The new buttons were added after the recent admin UX redesign. Still there is a number of places where the old-style buttons are used. We should revisit them and update. It's mostly mechanical work of finding calls of `BaseHtmlView.submitButton` and updating them to use `BaseHtmlView.makeSvgTextButton` where appropriate (where we can add an icon) or just add `AdminStyles.PRIMARY_BUTTON_STYLES` or `AdminStyles.SECONDARY_BUTTON_STYLES`. 

It would be great if we had utility functions like `createPrimaryButton` and `createSecondaryButton` but it's challenging to do with j2html as the library prevents adding extra classes to the already created element. 

Example of old buttons:
![image](https://user-images.githubusercontent.com/252053/196308760-54b42dea-76e1-46ed-802e-cc18810ad6f2.png)

Example of new buttons:  
![image](https://user-images.githubusercontent.com/252053/196308777-23ce2537-2895-4cf5-830b-cf3cc5721217.png)

lwsm99:
I would like to work on this issue

nbeloglazov:
Hi @lwsm99. 

Feel free to join our [slack channel](https://docs.civiform.us/governance-and-management/governance/communication#slack-channels) where we discuss project-related things and generally sync. Also it's a good place to ask questions if you have some issues with development setup. If you feel like it - please introduce yourself on #introductions channel. 

As for working on this issue - feel free to open PRs. Currently no one is working on it and there is a bunch of place where we use old-style buttons. So it's quite easy to parallelize.

PR #3660, opened 2022-10-18
Update style of buttons on program edit page

nbeloglazov:
### Description

This is a first step towards updating program edit page for the new design. Old rectangular buttons replaced with the new rounded ones with icons where reasonable. Also makes all buttons as secondary as according to UX guidance there should be only one primary button on the page. In later updates some of those buttons will be hidden under triple-dot menus. 

Update text of some buttons from camel-case or ALL_CAPS to regular case.

Additionally update `npm test` call to pass parameters using `--` so that hey are passed all the way to the jest and not processed by npm itself. Found this issue as passing `-u` didn't work properly.

## Release notes:

Update buttons on program edit page to use modern rounded shape with icons.

### Checklist

### Issue(s)

Issue #3659


shanemc-goog (approved):
FWIW In the future it'd be nice to have the shell change in a different PR as it's functionally unrelated to this PR.  If we needed to roll one or the other back it gets murkier than rolling back the entire PR

nbeloglazov:
Ugh, accidentally made "Create new question" to white in Question list page too. Need to fix it.

PR #3671, opened 2022-10-18
Improve logout handling in browser tests

### LINKED ISSUE #3669: Browser tests are flaky: possible logout interaction
We're noticing that browser tests are a lot more flakey lately.

One hypothesis is that it's related to the new logout logic.

nbeloglazov:
### Description

During logout we need to have extra logic only when user was properly logged through OIDC. In those cases we need to click on "Yes" button on OIDC logout page. In other cases we should just continue.

Also removes timeout from `textContent` call. That timeout caused errors sometimes that could explain flakiness. 

Ran github actions 3 times and no failures in browser tests
 
### Checklist

### Issue(s)

Fixes #3669

### DIFF
--- a/browser-test/src/support/index.ts
+++ b/browser-test/src/support/index.ts
@@ -231,14 +231,13 @@ export const gotoEndpoint = async (page: Page, endpoint: string) => {
 
 export const logout = async (page: Page) => {
   await page.click('text=Logout')
-  try {
-    const pageContent = await page.textContent('html', {timeout: 100})
+  console.log(page.url())
+  if (page.url().match('fake-idcs.*/session/end')) {
+    const pageContent = await page.textContent('html')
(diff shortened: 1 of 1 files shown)


browser-test/src/support/index.ts
```diff
@@ -231,14 +231,15 @@ export const gotoEndpoint = async (page: Page, endpoint: string) => {
 
 export const logout = async (page: Page) => {
   await page.click('text=Logout')
-  try {
-    const pageContent = await page.textContent('html', {timeout: 100})
+  // If user logged through OIDC previous - during logout page they are
```
shanemc-goog, line 234:
```suggestion
  // If the user logged in through OIDC previously - during logout they are
```

shanemc-goog (approved):

nbeloglazov:
Sorry, will apply proposed changes in follow up PR.

Issue #3678, opened 2022-10-19
Remove "legacy-peer-deps" and "force" from npm install once jest updated


nbeloglazov:
In browser tests at the moment we are using version of jest-image-snapshot that doesn't support jest 29 while we are using jest 29. It breaks normal `npm install` as it detects requirement mismatch. To workaround we suppress that error by adding `npm install --force` or `legacy-peer-deps` in `.npmrc`. This is temporary workaround until https://github.com/americanexpress/jest-image-snapshot/pull/309 is submitted and new version jest-image-snapshot is released.

PR #3679, opened 2022-10-19
Add option to ignore npm version mismatch

nbeloglazov:
### Description

See https://github.com/civiform/civiform/issues/3678 for more details. This PR adds `.npmrc` with option that ignores version mismatch. This is required for rennovate bot to be happy. Currently it fails to update some npm dependencies because it trips over that warning. 

### Checklist

### DIFF
--- a/browser-test/package-lock.json
+++ b/browser-test/package-lock.json
@@ -6583,8 +6583,7 @@
       "version": "5.3.2",
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-5.3.2.tgz",
       "integrity": "sha512-rq9s+JNhf0IChjtDXxllJ7g41oZk5SlXtp0LHwyA5cejwn7vKmKp4pPri6YEePv2PU65sAsegbXtIinmDFDXgQ==",
-      "dev": true,
-      "requires": {}
+      "dev": true
     },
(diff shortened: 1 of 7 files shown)


browser-test/package-lock.json
```diff
@@ -6583,8 +6583,7 @@
       "version": "5.3.2",
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-5.3.2.tgz",
       "integrity": "sha512-rq9s+JNhf0IChjtDXxllJ7g41oZk5SlXtp0LHwyA5cejwn7vKmKp4pPri6YEePv2PU65sAsegbXtIinmDFDXgQ==",
-      "dev": true,
-      "requires": {}
```
shanemc-goog, line 6587: do we need to put this back after the issue is resolved too?  if so add comment.

shanemc-goog (approved):

browser-test/package-lock.json
nbeloglazov: I think it's some odd side-effect of adding legacy-peer-deps=true. This file, package-lock.json is auto-generated during `npm install`. And reading description of `requires` field - it sounds like empty object or missing are the same. So I'm just going to ignore it given it's auto-generate.

PR #3706, opened 2022-10-21
Add screen reader text for errors for translation

jeanniefu:
### Description

This will be used to announce that errors are present for screen reader users. I checked with UX that this text sounds good.

## Release notes:

N/A

### Checklist

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

### Issue(s)

#3563

PR #3708, opened 2022-10-21
Announce that form errors are present for screen reader users

### LINKED ISSUE #3563: (WCAG 2.1) Announce when errors are present to screen reader users

**WCAG SC**:  [4.1.3. Status messages](https://www.w3.org/TR/WCAG21/#status-messages)

**Observed results**: There is no announcement by screen reader that errors are present.

**How to fix**: 
From a11y team:
Assign appropriate ARIA roles or aria-live properties to status messages so that the new content is spoken to users of screen readers. The element which acts as the live region needs to be present in the document object model (DOM) prior to the content being inserted to make sure the change is detected by some screen readers. Status message can be visible or not. If visible, these should be at the top of the form.

Examples:
```
<div role=""status"">Fix errors in the form before you can proceed</div> OR
<div class=""sr-only"" aria-live=""polite"">Address fields in error before proceeding</div>
```

From me:
This needs additional scoping for the proper solution. I think file upload errors can use `aria-live` (since errors are handled client side). All other error messages actually are rendered server-side, so we'll need to look more into whether using `aria-live` works here. From a quick skim of the w3.org docs, one approach that works is "to include an error notification in the page title (title element)" ([source](https://www.w3.org/WAI/WCAG21/Techniques/general/G83#description)).

jeanniefu:
### Description

For file upload, we use client side validation so the error message appears without reloading the page. We indicate the presence of that new error with `aria-live`. 

For all other questions, we use server side validation. We add sr-only text for the user to indicate there were errors. 

Tested manually with a screen reader to ensure this works.

SR text was added in #3706, and has not been translated yet but I think this is still worth adding in now.

## Release notes:

Announce that form errors are present for screen reader users.

### Checklist

### Issue(s)

Fixes #3563


server/app/views/applicant/ApplicantProgramBlockEditView.java
```diff
@@ -98,6 +99,20 @@ private ContainerTag<?> renderBlockWithSubmitForm(Params params) {
     if (params.block().isFileUpload()) {
       return fileUploadStrategy.renderFileUploadBlock(params, applicantQuestionRendererFactory);
     }
+
+    FormTag form = form();
+
+    if (params.block().hasErrors()
+        && params.errorDisplayMode()
+            == ApplicantQuestionRendererParams.ErrorDisplayMode.DISPLAY_ERRORS) {
```
bion, line 107: suggestion (blocking): use `.equals` for comparing enums
```suggestion
        && ApplicantQuestionRendererParams.ErrorDisplayMode.DISPLAY_ERRORS.equals(params.errorDisplayMode())) {
```

bion (changes requested):

bion (approved):

Issue #3419, opened 2022-09-14
Implement the program edit view V2


cassandralfernandes:
The Civiform Admin didn’t know where to start on the program edit view. Ensure the content and visual design is focused on the essentials. The goal is to make the page easier to parse and reduce confusion on landing by having less elements on the page  

[Mock here](https://docs.google.com/presentation/d/15fcq3dYFgy1Dfy2UjZybmHD9KIW5cz8Q1BjwgOpLJ7g/edit#slide=id.g13e43d0b668_0_134): 
![Screen Shot 2022-09-14 at 12 44 09 PM](https://user-images.githubusercontent.com/99513234/190247897-706379a4-1e6c-4e91-8cec-b2c5df94ecd1.png)

nbeloglazov:
I'm starting working on it. Some notes about implementation. I'm going to use term "screen" instead of "page" which is used in mocks. This is to avoid confusion between the web page where civiform renders admin UI and "page" as in separate part of a civiform program where applicant fills out questions. 

1. The edit page will contain all screens and questions at once. Clicking "Add question" should open question bank only with the questions relevant to the corresponding screen. Opening question bank is client-side action so we'll need a logic that filters out questions for the selected screen. The plan is to render the question bank with all questions (question that can be added at least to one of the screens). Each question will contain `data-screen-X` attribute indicating that question can be added to screen X. When user clicks "Add question" on screen Y we loop through questions with `data-screen-Y` and show only them. 

2. After any modification to a page (question add, delete, mark optional, move page up/down) the whole page is refreshed.  Which leads to scroll position jumping all the way to the top. We already heard complaints about it. But now it will be more important given that height of the page will be bigger as we are going to render all screens and questions at once. The plan is after each page update redirect user to the edit page to the particular screen that they modified. We are going to use hash in URL that tells browser to scroll to the element with the corresponding id. For example after user added question to 5th screen we redirect them back to `/admin/programs/1/editBlocks#screen-5`. This will still introduce some "jumping" but it will be less annoying.

3. Need to decide how to handle enumerator screens. Whether they should be rendered the same as regular screens or maybe visually different. For example children screens of an enumerator screens can have slight padding to indicate that they are sub-screens.

nb1701:
If we pull this back in, https://github.com/civiform/civiform/pull/3732 is a step in this direction.

PR #3732, opened 2022-10-28
Refactor question bank to support multiple blocks

nbeloglazov:
### Description

As preparation for #3419 this PR refactors question bank component to support displaying questions for multiple blocks on the same page. We are planning to refactor program edit page to contain all blocks and questions on the same page and admin can add question to any block from that page. 

Previously question bank was used to display questions only for a specific block. We filtered questions relevant to that block server side during rendering. Now question bank contains all questions that can be added to at least 1 block in a given program. All questions rendered hidden initially but their DOM elements contain additional attributes: `data-block-1`, `data-block-4`, ... indicating to which blocks given question can be added. 

Upon admin clicking "Add question" we show the question bank and using JS filter out questions which are not eligible for the given block. The block id is retrieved from the "Add question" button DOM attribute. So now page contain multiple "Add question" button and as long as they have correct DOM attributes with block id the question bank will show correct questions.

Additionally we now use `sqb` URL param to indicate two things:
1. Question bank is in open state.
2. Block id that the question bank should show eligible questions for.

The URL param allows us to keep question bank open state between page refreshes and redirects improving user experience.

This PR has no effect on admin experience. 

## Release notes:

Refactor question bank in program edit page to prepare for the upcoming program edit redesign. This change has no impact on UI.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3419

### DIFF
--- a/server/app/views/components/QuestionBank.java
+++ b/server/app/views/components/QuestionBank.java
@@ -11,6 +11,7 @@ import static j2html.TagCreator.p;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import j2html.tags.specialized.ButtonTag;
 import j2html.tags.specialized.DivTag;
 import j2html.tags.specialized.FormTag;
@@ -20,8 +21,6 @@ import java.net.URISyntaxException;
 import java.time.Instant;
 import java.util.Comparator;
 import java.util.Optional;
-import java.util.function.Predicate;
-import java.util.stream.Stream;
 import org.apache.http.client.utils.URIBuilder;
 import play.mvc.Http;
 import play.mvc.Http.HttpVerbs;
@@ -34,15 +33,23 @@ import views.style.AdminStyles;
 import views.style.ReferenceClasses;
 import views.style.StyleUtils;
 
-/** Contains methods for rendering question bank for an admin to add questions to a program. */
+/**
+ * Contains methods for rendering question bank for an admin to add questions to a program. Question
+ * bank contains questions for all blocks in a given program while admin interacts with the question
+ * bank within context of a specific block. Questions for the block filtered client-side that allows
+ * to switch between blocks client-side without reloading page.
+ */
 public final class QuestionBank {
 
   // Url parameter used to force question bank open upon initial rendering
   // of program edit page.
   private static final String SHOW_QUESTION_BANK_PARAM = "sqb";
 
+  // Attribute set on buttons that trigger display of question bank. See
+  // usage of that attribute in questionBank.ts.
+  private static final String CURRENT_BLOCK_ID_ATTRIBUTE = "data-block-id";
+
   private final QuestionBankParams params;
-  private Optional<Long> enumeratorQuestionId;
 
   /**
    * Possible states of question bank upon rendering. Normally it starts hidden and triggered by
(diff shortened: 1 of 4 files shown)


shanemc-goog:
Can you link mocks that this is implementing?

nbeloglazov:
Here is the mock: 
image

Admin can click "Add a question" for different blocks and question bank need to show corresponding questions.

server/app/views/components/QuestionBank.java
```diff
@@ -219,50 +232,65 @@ private DivTag renderQuestionDefinition(QuestionDefinition definition) {
...
+   * data attribute on the button that triggered showing the bank. This method sets that attribute.
+   */
+  public static void setCurrentBlockAttribute(ButtonTag element, long blockId) {
+    element.attr(
+        QuestionBank.CURRENT_BLOCK_ID_ATTRIBUTE,
+        QuestionBank.getQuestionEligibleForBlockAttribute(blockId));
+  }
+
+  /**
+   * Used to check whether given question can be added to the given block.
```
shanemc-goog, line 254: Summary fragments should be written as if they started with "This method" and should say what the method does. This content should come second.

https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

server/app/views/components/QuestionBank.java
```diff
@@ -219,50 +232,65 @@ private DivTag renderQuestionDefinition(QuestionDefinition definition) {
...
+   *   <li>If there is at least one question in the block, all single-block questions are filtered.
+   *   <li>If there is a single block question in the block, all questions are filtered.
    *   <li>If this is a repeated block, only the appropriate repeated questions are showed.
    *   <li>Questions already in the program are filtered.
    * </ul>
    */
-  private Stream<QuestionDefinition> filterQuestions() {
-    if (containsSingleBlockQuestion()) {
-      return Stream.empty();
+  private boolean canAddQuestionToBlock(QuestionDefinition question, BlockDefinition block) {
```
shanemc-goog, line 265: Consider renaming to `canQuestionBeAddedToBlock` or `isQuestionValidForBlock` so it has more sentence like phrasing.  Can add Question doesn't flow well.

server/app/views/components/QuestionBank.java
```diff
@@ -219,50 +232,65 @@ private DivTag renderQuestionDefinition(QuestionDefinition definition) {
...
+        .collect(ImmutableSet.toImmutableSet());
   }
 
-  /**
-   * A non-empty block cannot add single-block questions, in addition to {@link
-   * QuestionBank#questionFilter}.
-   */
-  private boolean nonEmptyBlockFilter(QuestionDefinition questionDefinition) {
-    return !singleBlockQuestion(questionDefinition) && questionFilter(questionDefinition);
+  /** If a block already contains a single-block question, no more questions can be added. */
```
shanemc-goog, line 291: The comment is more relevant to the caller not this method which doesn't enforce the comment's statement.

server/app/views/components/QuestionBank.java
```diff
@@ -219,50 +232,65 @@ private DivTag renderQuestionDefinition(QuestionDefinition definition) {
...
+    }
+    // if question is not related to an enumerator - its enumerator id will be
+    // empty Optional, same as getEnumberatorQuestionId for block.
+    return question.getEnumeratorId().equals(getEnumeratorQuestionId(block));
   }
 
   /**
-   * An block can add questions with the appropriate enumerator id that aren't already in this
-   * program.
+   * Get list of blocks to which the given question can be added. See {@code
```
shanemc-goog, line 281:
```suggestion
   * Returns the list of blocks to which the given question can be added. See {@code
```

server/app/views/components/QuestionBank.java
```diff
@@ -219,50 +232,65 @@ private DivTag renderQuestionDefinition(QuestionDefinition definition) {
...
+   * canAddQuestionToBlock()} for conditions.
    */
-  private boolean questionFilter(QuestionDefinition questionDefinition) {
-    return questionDefinition.getEnumeratorId().equals(getEnumeratorQuestionId())
-        && !params.program().hasQuestion(questionDefinition);
+  private ImmutableSet<Long> getBlocksThatQuestionEligibleFor(QuestionDefinition question) {
+    return params.program().blockDefinitions().stream()
+        .filter(b -> this.canAddQuestionToBlock(question, b))
```
shanemc-goog, line 286:
```suggestion
        .filter(block -> this.canAddQuestionToBlock(question, block))
```

shanemc-goog: Java style requires readable vars, even in lambdas afaik.

shanemc-goog (changes requested):
Mostly doc suggestions

nbeloglazov:
I'm going to put this PR on hold. Learned that there is another version of Program edit page UI, in Northstar:
image

That version uses tabs to navigate between screens. It's close to what we have today and doesn't require changes in question bank.

PR #3736, opened 2022-10-29
Extract block question validation logic into separate utility class

nbeloglazov:
### Description

`ProgramBlockValidation` contains logic that checks whether a question can be added to a block. The class is introduces mainly for the purpose of sharing between two callsites:
1. During question bank rendering to ensure that only eligible questions are rendered for a given block.
2. When processing request for adding question to a block to guard against bad request or bugs in our "client-side" checks at step 1.

Additionally PR replaces `DuplicateQuestionException` with a more generic `CantAddQuestionToBlockException` that can provide few different reasons why question cant be added.

## Release notes:

Enhance logic that validates question addition to program screens. 

#### General

#### User visible changes

#### New Features


server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+import services.question.types.EnumeratorQuestionDefinition;
+import services.question.types.QuestionDefinition;
+
+/** Helper class for performing validation related to creating or modifying program blocks. */
+public final class ProgramBlockValidation {
+
+  private ProgramBlockValidation() {}
+
+  /**
+   * Result of checking whether a question can be added to a specific blocks. Only ELIGIBLE means
```
shanemc-goog, line 16:
```suggestion
   * Result of checking whether a question can be added to a specific block. Only ELIGIBLE means
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+   * that question can be adde. All other states indicate that question is not eligible for the
```
shanemc-goog, line 17:
```suggestion
   * that question can be added. All other states indicate that question is not eligible for the
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+    // Cannot add question to the block because the block contains a question
+    // of special type (enumerator or file upload) that prohibits having
+    // other questions in the block.
+    BLOCK_IS_SINGLE_QUESTION,
+
+    // Cannot add special question type (enumerator or file upload) to
+    // non-empty block. These questions must be an only question in a block.
+    CANT_ADD_SINGLE_BLOCK_QUESTION_TO_NON_EMPTY_BLOCK,
+
+    // Cannot add question to the block because either block is a child of enumerator
```
shanemc-goog, line 36:
```suggestion
    // Cannot add question to the block because either block is a child of an enumerator
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+    // block and can contain only questions that themselves children of the enumerator and
```
shanemc-goog, line 37:
```suggestion
    // block and can contain only questions that themselves are children of the enumerator and
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+    // provided question is not. Or provided question is a child of enumerator question while
```
shanemc-goog, line 38:
```suggestion
    // provided question is not; or the provided question is a child of an enumerator question while
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+    // the block is a regular block.
+    ENUMERATOR_MISMATCH
+  }
+
+  /**
+   * Check whether given question can be added to the provided block. This should be checked both
```
shanemc-goog, line 44:
```suggestion
   * Check whether given question can be added to the provided block.
   *
   * <pr>This should be checked both
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -0,0 +1,100 @@
...
+        .getEnumeratorId()
+        .equals(ProgramBlockValidation.getEnumeratorQuestionId(program, block))) {
+      return AddQuestionResult.ENUMERATOR_MISMATCH;
+    }
+    return AddQuestionResult.ELIGIBLE;
+  }
+
+  private static boolean isSingleBlockQuestion(QuestionDefinition question) {
+    switch (question.getQuestionType()) {
+      case ENUMERATOR:
```
shanemc-goog, line 71: Your other PR changes the construction of this check, how are you reconciling the two?

server/test/services/ProgramBlockValidationTest.java
```diff
@@ -0,0 +1,97 @@
...
+            .withRequiredQuestionDefinition(nameQuestion)
+            .buildDefinition();
+    assertThat(
+            ProgramBlockValidation.canAddQuestion(
+                program, program.getLastBlockDefinition(), fileQuestion))
+        .isEqualTo(AddQuestionResult.CANT_ADD_SINGLE_BLOCK_QUESTION_TO_NON_EMPTY_BLOCK);
+  }
+
+  @Test
+  public void canAddQuestion_cantAddEmuratorQuestionToNonEnumeratorBlock() throws Exception {
```
shanemc-goog, line 70: Please add a positive assertion test for enumerator question added to enumerator block

shanemc-goog:
I don't believe this should be ignore-for-release as you're changing how validation works which could cause user visible issues (so we should indicate it in the release notes) as well you're changing the error messaging format.  I'd mark it as Feature.

shanemc-goog (changes requested):

shanemc-goog:
Mark this as a "Feature", you're changing how validation works which could cause user visible issues, so we should indicate it in the release notes, as well you're changing the user visible error messaging format.

server/app/services/ProgramBlockValidation.java
nbeloglazov: I'll merge this PR first. And then figure out the other PR if I end up merging it.

server/app/services/program/CantAddQuestionToBlockException.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class CantAddQuestionToBlockException extends Exception {
+  public CantAddQuestionToBlockException(
+      ProgramDefinition program,
+      BlockDefinition block,
+      QuestionDefinition question,
+      ProgramBlockValidation.AddQuestionResult reason) {
+    super(
+        String.format(
+            "Can't add question to the block. Error: %s. Check comments in AddQuestionResult enum."
```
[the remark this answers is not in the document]
shanemc-goog: ?

shanemc-goog (changes requested):

server/app/services/program/CantAddQuestionToBlockException.java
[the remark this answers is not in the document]
nbeloglazov: Done.
Though the error is not meant to be understandable or actionable by CiviForm admins. They are not expected to see it but if they do - they should report it to eng and we'll try to understand what went wrong. We consider adding text something like "Please report this error to ...".

As for message being atypical, I find it nice when error tries to provide more info on how to debug it. Though here I'm guessing engineers will try and find the enum value anyway.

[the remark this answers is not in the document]
shanemc-goog: The comment was because we will print this in a notFound HTTP response, as I see in the PR, which the site user will see; a CiviForm Admin.  If we want someone for the IT manager then we can log that

PR #3739, opened 2022-10-31
Add bin/docker-reset

bion:
### Description

Add a new `bin/` script for destroying all docker resources. 

#### General

#### User visible changes

#### New Features

### DIFF
--- a/bin/docker-reset
+++ b/bin/docker-reset
@@ -5,13 +5,13 @@
 source bin/lib.sh
 
 read -p "Are you sure you want to destroy ALL containers, volumes, and networks? [y/n] " -r
-if ! [[ $REPLY =~ ^[Yy]$ ]]; then
+if ! [[ "${REPLY}" =~ ^[Yy]$ ]]; then
   exit 0
 fi
 
 if [[ $(docker container ls -q | wc -c) -ne 0 ]]; then
   echo "Killing running containers"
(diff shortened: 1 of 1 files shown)


bin/docker-reset
```diff
@@ -0,0 +1,29 @@
+#! /usr/bin/env bash
+
+# DOC: Stop ALL docker containers and destroy ALL containers, volumes, networks.
+
+source bin/lib.sh
```
gwendolyngoetz, line 5: NB: Do we need to source this for this script?

gwendolyngoetz (approved):

bin/docker-reset
```diff
@@ -0,0 +1,29 @@
...
+
+read -p "Are you sure you want to destroy ALL containers, volumes, and networks? [y/n] " -r
+if ! [[ $REPLY =~ ^[Yy]$ ]]; then
```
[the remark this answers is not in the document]
bion: Thanks! I forgot

bin/docker-reset
bion: It would probably work fine without it, but even if we don't use the functions defined in `bin/lib.sh` it does set some good default bash options e.g. `set -e` which we generally want to be consistent across all the scripts

nbeloglazov (approved):

PR #3751, opened 2022-11-02
Surface how many questions will be auto-filled on the applicant-facing program listing page

seanmalbert:
### Description

This adds the autofill count to a the relevant programs for an applicant.

## Release notes:

TODO

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


seanmalbert:
@bion, I still have to write the browser tests and update to not show when zero questions are autofilled, but I want to make sure this is on the right track before continuing.

server/app/services/applicant/ApplicantService.java
```diff
@@ -665,6 +674,22 @@ private void logDuplicateDrafts(ImmutableSet<Application> applications) {
     }
   }
 
+  private int getAutoFillCount(ReadOnlyApplicantProgramService roApplicantService) {
+    return roApplicantService.getSummaryData().stream().filter(data -> data.isPreviousResponse()).collect(ImmutableList.toImmutableList()).size();
+  }
+
+  private int getAutoFillCount(Application application) {
+    ReadOnlyApplicantProgramService roApplicantService = getReadOnlyApplicantProgramService(application).toCompletableFuture().join();
```
bion, line 682: suggestion (blocking): avoid calling `join` here because it blocks the HTTP thread on database IO, which we want to avoid doing for any applicant-facing routes.

Two possible approaches for improving this:
1. do the IO here asynchronously and make this method and its caller asynchronous (change their return type to `CompletionStage<T>`)
1. collect the data needed for this method to do its work somewhere else asynchronously so that this code accesses local data only and stays synchronous

seanmalbert: @bion ,
Got back to this and I ended going with the first approach, since it introduced the least amount of complexity. I'm sure we can call `join` less than what I have here, but that requires more overhaul, so let me know what you think.

bion: > I'm sure we can call join less than what I have here, but that requires more overhaul, so let me know what you think.

Unfortunately that's what's required to maintain [the standard of not blocking HTTP threads in applicant-facing routes](https://docs.civiform.us/contributor-guide/developer-guide/development-standards#async-request-handling).

server/app/services/applicant/ApplicantService.java
```diff
@@ -608,18 +614,21 @@ private ApplicationPrograms relevantProgramsForApplicant(
         Sets.difference(activeProgramNames.keySet(), programNamesWithApplications);
     unappliedActivePrograms.forEach(
         programName -> {
+          OptionalInt autoFillCount = getAutoFillCount(applicantId, activeProgramNames.get(programName).id()).toCompletableFuture().join();
```
bion, line 617: suggestion (blocking): this is moving in the right direction, but the `join` calls here and  on L579 block the HTTP thread.

bion (changes requested):

nb1701:
We now have "you may be eligible" badge which supercedes this work.

PR #3753, opened 2022-11-02
Closes #816: Add maxlength to all input fields.

### LINKED ISSUE #816: If an applicant enters a very long string in a text field, they get a 400
We should guard against this by putting some reasonable max length on strings (10,000 characters?) and returning a helpful validation message if they enter something longer.

ckirksey3:
### Description

I add a maxlength HTML attribute to all LabelTags with a set length of 10,000 characters. This should reduce the likelihood of users submitting responses so large that the form post fails.

The attribute will exist on tag types for which it doesn't make sense (like checkboxes), but this seemed the cleanest way to achieve the protection given the current layout of the code and I don't believe the superfluous attribute will have any effect.

## Release notes:

Enforce max length of 10,000 characters on input fields.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #816;


nbeloglazov:
You want to try and add browser test that submits looooong value? Also how does it look visually when you are submitting long value?

ckirksey3:
> Also how does it look visually when you are submitting long value?

If you try to type another character after reaching 10k, it just doesn't type. It's still possible for someone to inject a string larger than 10k into the dom and submit, but this prevents accidental failures.

browser-test/src/text.test.ts
```diff
@@ -114,6 +114,52 @@ describe('Text question for applicant flow', () => {
...
...456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890' +
```
nbeloglazov, line 146: I suggest generating 10k string programmattically:

```
let str = '';
for (let i = 0; i < 1000; i++) {
  str += '123456789';
}
```

nbeloglazov (changes requested):

browser-test/src/text.test.ts
ckirksey3: Good call. Updated!

nbeloglazov (approved):

Issue #3727, opened 2022-10-27
Allow to run all development tasks from single dev environment


nbeloglazov:
The goal is to allow developers to run dev tasks (running server, running unit tests, running browser tests) from the same environment removing the need to switch between environments that leads to long waiting times (full recompilation). Additionally it will help ensure that environments don't diverge too much, making debugging tests easier. 

Doc: https://docs.google.com/document/d/1ckZmKo_7IXf3W4YgbG465OMsv6iCaGirf_zDP_aU7gg/edit

PR #3755, opened 2022-11-02
Add two sbt commands to run civiform and dev and browser test modes

nbeloglazov:
### Description

This makes it possible to switch between servers from sbt shell. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3727

### DIFF
--- a/browser-test/browser-test-compose.dev.yml
+++ b/browser-test/browser-test-compose.dev.yml
@@ -13,7 +13,7 @@ services:
     tty: true # docker run -t
     ports:
       - 9457:9457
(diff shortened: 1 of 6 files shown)


bion (approved):

bion:
Please update the docs repo as well

nbeloglazov:
We don't have anything about running server from sbt shell in docs yet. For that we first need to make more changes into our scripts. So I don't think there are any docs to update. I was planning to make sure that we have a way to launch sbt shell in a way that allows to call `runDevServer` from it and work with it. At the moment running `bin/sbt` launches sbt shell in a different container that doesn't expose port 9000 so you can't run server from it.

PR #3756, opened 2022-11-03
Remove noisy " Application {} does not include an applicant name" warning

nbeloglazov:
### Description

During browser test run there is a lot of warnings in console like

```
[warn] s.a.ApplicantData - Application {} does not include an applicant name. This is expected for guest users.
```

They are misleading as they don't indicate that something is wrong or should be fixed. Additionally it uses incorrect template with missing application id. Reading the code there is no application id attached to ApplicantData as it is expected to contain answers to all questions. That's why I decided to remove this warning message altogether as it doesn't give any useful information.

#### General

#### User visible changes

#### New Features

### DIFF
--- a/server/app/services/applicant/ApplicantData.java
+++ b/server/app/services/applicant/ApplicantData.java
@@ -7,7 +7,6 @@ import com.google.common.collect.ImmutableMap;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.Locale;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


shanemc-goog:
Is there a way to know if it's not a Guest and only log then?


I'm less familiar with the Applicant-seed questions so @bion .

nbeloglazov:
Even for logged-in user we might not have a name if it wasn't returned by auth provider or not answered in any question. For example login.gov doesn't return it. So I don't know what case that warning was expected to catch. Decided to send PR to remove it and see if someone thinks of a reason to keep it.

bion (approved):

bion:
Thanks for cleaning this up!

shanemc-goog (approved):

PR #3757, opened 2022-11-03
Show public program name and description on block edit page

### LINKED ISSUE #3741: Show applicant-facing program name and description on program edit page
We should be showing applicant-facing program name and description across all pages (both admin and applicant). Here is example of what we show in program list page for admin:

image

However, on the program edit page we still who admin id and description: 
image

We should be showing applicant-facing there as well.

nbeloglazov:
### Description

Recently we switched to showing public name and description instead of admin name/description on program index page and on program description page. Block edit page is still showing admin name/description which is confusing: from program index page admin clicks "edit" on program named "Official looking program name" and on block edit page we show "shorter-url-compatible-name" as program name. 

At the moment browser tests didn't catch any real changes because we use the same string for admin name and public name. I'll update browser tests to use different names to make changes like these more obvious.

## Release notes:

Show public program name and description on block edit page. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3741


server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -13,22 +14,30 @@
 
 abstract class ProgramBlockView extends BaseHtmlView {
   /** Renders a div with internal/admin program information. */
-  protected final DivTag renderProgramInfo(ProgramDefinition programDefinition) {
-    DivTag programTitle =
-        div(programDefinition.adminName()).withId("program-title").withClasses("text-3xl", "pb-3");
-    DivTag programDescription = div(programDefinition.adminDescription()).withClasses("text-sm");
+  protected final DivTag renderProgramInfo(ProgramDefinition program) {
```
shanemc-goog, line 17: There is a "program" concept also; suggestion to make the name shorter with `programDef`

server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -13,22 +14,30 @@
...
+    DivTag title =
+        div(program.localizedName().getDefault())
```
shanemc-goog, line 19: None of the tests exercise this change, can you update the tests so they do?  Presumably they are setting the visible and admin name to the same hyphenated thing?

shanemc-goog (changes requested):

server/app/views/admin/programs/ProgramBlockView.java
nbeloglazov: Updated unit tests. For browser tests it's going to be a bit more work so I decided to make it into a follow up PR. Currently browser tests always use the same hyphenated string for public and admin names.

shanemc-goog (approved):

PR #3796, opened 2022-11-11
if register_uri is set show create account button

azizova-leyla:
### Description

Simplifying logic for "Create Account" button. If APPLICANT_REGISTER_URI (preferred) or IDCS_REGISTER_URI (currently used in Seattle) variables are set - show the button. Otherwise hide it.
It is recommended to set it for better user experience.

Should be no-op for current deployments.

Comment about IDCS needing this extra flow is no longer relevant, this can be used with any identity provider that allows new account creation.

## Release notes:

Added new config value: APPLICANT_REGISTER_URI. When set to a uri for new account registration - "Create Account" button will be shown on the login page. 
Screen Shot 2022-11-10 at 10 17 42 AM

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


server/conf/application.conf
```diff
@@ -78,6 +78,10 @@ auth {
   oidc_override_logout_url = ${?APPLICANT_OIDC_OVERRIDE_LOGOUT_URL}
   oidc_post_logout_param = ${?APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM}
   oidc_logout_client_id_param = ${?APPLICANT_OIDC_LOGOUT_CLIENT_PARAM}
+
+  # Direct URI to create new account.
+  register_uri = ${?APPLICANT_REGISTER_URI}
+  register_uri = ${?IDCS_REGISTER_URI}
```
gwendolyngoetz, line 84: NB: I'm slightly concern having two different environment variables setting the same value could make troubleshooting difficult. Is there anything we can do to log an error if both APPLICANT_REGISTER_URI and IDCS_REGISTER_URI are set? If both are set is that such an exceptional event we should prevent the application from starting?

shanemc-goog: As I understand it we would want to have Seattle move to the new general APPLICANT one then remove the IDCS one completely.

Is that correctly @azizova-leyla ?

gwendolyngoetz (approved):

server/conf/application.conf
gwendolyngoetz: I switched Seattle over

azizova-leyla: that was fast =) https://github.com/civiform/civiform/pull/3800 has cleanup.

gwendolyngoetz: Just wanna get it done

Issue #3566, opened 2022-09-29
(WCAG 2.1) Set focus after removing an enumerator


jeanniefu:
Required to meet WCAG 2.1 guidelines. Identified during an accessibility audit against WCAG 2.1.

**WCAG SC**: [2.4.3. Focus Order](https://www.w3.org/TR/WCAG21/#focus-order)

**Observed results**: Focus is not explicitly set after removal of an enumerator ("household member" in this example).

**How to fix**: With respect to the enumerator in the screenshot - After removing a household member, set focus on the last "Remove household member" button. And if there are none, set focus to '+Add household member" button.

![image](https://user-images.githubusercontent.com/9094240/192911742-202770b4-ef8e-4b29-8a7d-8b36562767b7.png)

PR #3798, opened 2022-11-11
Refactor enumerator js code into its own file

jeanniefu:
### Description

Refactors the enumerator js code out of `main.js` into its own `enumerator.js` file. The new `enumerator.js` file is only added for enumerator pages.

I'll be adding more enumerator js code as part of #3566, and didn't want to make main.js bigger.

## Release notes:

N/A - internal refactor

#### General

#### User visible changes

#### New Features

### Issue(s) this completes
Refactoring prep for #3566.


server/app/assets/javascripts/enumerator.ts
```diff
@@ -0,0 +1,174 @@
...
+  // Configure the button on the enumerator question form to add more enumerator field options
+  const enumeratorOptionButton = document.getElementById(
+    'enumerator-field-add-button',
+  )
+  if (enumeratorOptionButton) {
+    enumeratorOptionButton.addEventListener('click', addNewEnumeratorField)
+  }
+
+  // Configure existing enumerator entity remove buttons
+  addEventListenerToElements(
```
nbeloglazov, line 13: This function is defined in main.ts as global function and used in this file. That's not great as function should be properly imported. But currently our TS is not setup in a way that supports it. So I guess that's fine. But can you add a comment here explaining that? Because it's pretty unusual and engineers might be confused on why this function works.

nbeloglazov (changes requested):
I don't think loading enumerator file just on enumerator pages instead of bundling it with main helps with performance as main.js is permanently cached. In future we should start bundling and minifying JS anyway. But this refactor definitely helps with code health. Thanks! 

Left minor request to add a comment to avoid potential WTF from developers working with TS code.

server/app/assets/javascripts/enumerator.ts
jeanniefu: Ah I just blindly copied and pasted things into this file. I didn't even realize this was coming from main.js. Thanks I added a comment!

jeanniefu:
> I don't think loading enumerator file just on enumerator pages instead of bundling it with main helps with performance as main.js is permanently cached. In future we should start bundling and minifying JS anyway. But this refactor definitely helps with code health. Thanks!
> 
> Left minor request to add a comment to avoid potential WTF from developers working with TS code.

Oh good to know. I forgot to mention that this change is because I wanted to add more js logic for enumerators (just updated the PR description). How is main.js permanently cached? I see it added here https://github.com/civiform/civiform/blob/main/server/app/views/BaseHtmlLayout.java#L99.

nbeloglazov:
> How is main.js permanently cached?

We use sbt-web plugin `digest` ([code](https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/build.sbt?L127)) which adds hash to all JS filenames served by CiviForm. And these "digested" files have long TTL.

nbeloglazov (approved):

jeanniefu:
> > How is main.js permanently cached?
> 
> We use sbt-web plugin `digest` ([code](https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/build.sbt?L127)) which adds hash to all JS filenames served by CiviForm. And these "digested" files have long TTL.

Thanks!

PR #3809, opened 2022-11-15
Refactor run-dev and sbt commands

nbeloglazov:
### Description

Changes docker files so that civiform container doesn't launch server upon start. Instead civiform server is launched by running `docker exec` on the running container. This allows to run different commands on the same environment. For example we can launch sbt shell and then launch server or run unit tests from the shell. 

This change should have no affect on current developer workflows. Running `bin/run-dev` will still launch civiform server as it does today. 

This PR moves `entrypoint.sh` from root directory to `server` as that file makes sense within the context of server. Additionally it's no longer used as default entry point for the image. Instead we call it directly in `docker exec`. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3727

### DIFF
--- a/bin/sbt
+++ b/bin/sbt
@@ -8,5 +8,5 @@ docker::set_project_name_dev
 bin/pull-image
 
 docker::compose_dev up --wait -d
-docker exec -it "${COMPOSE_PROJECT_NAME}-civiform-1" ./entrypoint.sh -jvm-debug "*:8457" -Dsbt.offline
+docker::dev_server_sbt_command
 docker::compose_dev stop civiform
--- a/bin/sbt
+++ b/bin/sbt
@@ -7,6 +7,6 @@ docker::set_project_name_dev
 
 bin/pull-image
 
-docker::run_shell_container
-docker::run_shell_command sbt "$@"
-docker::stop_shell_container
+docker::compose_dev up --wait -d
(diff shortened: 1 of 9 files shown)


server/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+# This is file is used with `docker exec` in bin scripts by civiform engineers.
+# It's not used in by prod civiform.
```
bion, line 4:
```suggestion
# It's not used by prod civiform.
```

bin/sbt
```diff
@@ -7,6 +7,6 @@ docker::set_project_name_dev
 
 bin/pull-image
 
-docker::run_shell_container
-docker::run_shell_command sbt "$@"
-docker::stop_shell_container
+docker::compose_dev up --wait -d
+docker exec -it "${COMPOSE_PROJECT_NAME}-civiform-1" ./entrypoint.sh -jvm-debug "*:8457" -Dsbt.offline
```
bion, line 11: suggestion (non-blocking): consider pulling `docker exec -it "${COMPOSE_PROJECT_NAME}-civiform-1" ./entrypoint.sh -jvm-debug "*:8457" -Dsbt.offline` into a bash function `docker::dev_server_sbt_command`

bion (changes requested):

bin/sbt
nbeloglazov: Thanks, it's indeed cleaner that way. Done.

bion (approved):

PR #3815, opened 2022-11-15
Fix program card external link rendering

nbeloglazov:
### Description

It got broken by https://github.com/civiform/civiform/pull/3805 The reason is that link produced by LinkElement now always use `display: inline-flex` so that element inside the link (text and icon) are aligned properly. But it means that link is always inline. If callers want to make it block - they need to wrap it with div() as callers can't override LinkElement class.

## Release notes:

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


browser-test/src/application_navigation.test.ts
```diff
@@ -133,12 +134,24 @@ describe('Applicant navigation flow', () => {
     })
 
     it('verify program list page', async () => {
-      const {page} = ctx
+      const {page, adminPrograms} = ctx
+      await loginAsAdmin(page)
+      const programWithExternalLink = 'Program with external link'
+      await adminPrograms.addProgram(
+        programWithExternalLink,
```
azizova-leyla, line 141: to be thorough - can you add two case with and without external site.

nbeloglazov: The case without a site already present. (created in beforeAll). Here I'm adding a second program, with external site. So we are testing both here. Or your meant something else?

azizova-leyla (approved):

PR #3820, opened 2022-11-16
Jessbumblebee/dateconversion3

### LINKED ISSUE #3607: Datetime Values in Data Exports
**Describe the bug**
Data exports display datetimes in UTC while the UI shows them localized. This causes confusion for end users.

**To Reproduce**
View an application in the admin pages `/admin/programs/:programId/applications` and note the date shows as PDT/PST.

image

Do a demographics export or a program export (csv or json). Look at the dates in the export file. They are UTC.

**Expected behavior**
Have the export localize the dates for a consistent end user experience.

jessbumblebee:
### Description

Change the format of creation and submission times in the data exports to be in local time
## Release notes:

Change data exports (CSV and Json) so that creation and submission time are recorded in local time rather than UST for consistency with the date and time displayed in the admin UI.

#### General

#### User visible changes

N/A

#### New Features

N/A

### Issue(s) this completes

There is an additional change required to fix the issue (changing the time in the dropdown selector to PST/PDT)


server/app/services/export/CsvExporter.java
```diff
@@ -70,13 +74,14 @@ public void exportRecord(
...
         case SUBMIT_TIME:
-          printer.print(
-              application.getSubmitTime() != null
-                  ? application.getSubmitTime().toString()
-                  : EMPTY_VALUE);
+          if (application.getSubmitTime() == null) {
+            printer.print(EMPTY_VALUE);
+            break;
+          }
+          printer.print(dateConverter.renderDateTimeDataOnly(application.getSubmitTime()));
```
gwendolyngoetz, line 84: Can you move this into a private method of it's own? It's a little hard to read having multiple breaks in the same case.

server/app/services/DateConverter.java
```diff
@@ -61,10 +61,19 @@ public Instant parseIso8601DateToStartOfDateInstant(String dateString) {
...
+    return dateTime.format(DateTimeFormatter.ofPattern("yyyy/MM/dd 'at' h:m a z"));
+  }
+
+  /**
+   * Formats an {@link Instant} to a date and time in the local time zone for the purpose of data
+   * exports (with no filler words).
+   */
+  public String renderDateTimeDataOnly(Instant time) {
+    ZonedDateTime dateTime = time.atZone(zoneId);
+    return dateTime.format(DateTimeFormatter.ofPattern("yyyy/MM/dd h:m:s a z"));
```
gwendolyngoetz, line 76: I think the minute and second need to be mm and ss so they have the 0 prepended for single digits.

server/app/services/export/CsvExporter.java
jessbumblebee: Good point. Actually, I've changed it to an if...else, which looks more readable to me and keeps the logic in place.
Let me know in case it still looks a bit cryptic.

server/app/services/DateConverter.java
jessbumblebee: Good catch

gwendolyngoetz (changes requested):
LGTM. Just update the `Release notes` description and I'll approve.

jessbumblebee:
Done

gwendolyngoetz (approved):

Issue #1920, opened 2022-02-11
Program application review page changes


glorialiou:
There are two "Continue" buttons on the program application review page in the applicant experience. Both buttons are problematic: https://docs.google.com/presentation/d/1goxiepPXPrW3wK8wUSNoCpcf7iXIohXfj96iTYiQxo0/edit#slide=id.gfa7bf315fa_0_1862

There is a smaller "Continue" button that takes the applicant to that particular question (it's the last question). This should be replaced by "Edit" to match the other questions.

Then there is the larger blue "Continue" button at the bottom of the page that currently takes the applicant to the start of the application, which is unexpected. This button should take the applicant to where they were previously -- the last page of the application where they can submit it.

kristinyim:
Rochester UXR confirms the need for this. Research participants were thrown off by the highlighted rows and unclear CTAs ("Continue" vs "Edit")

kristinyim:
Another suggestion from UXR is to add edit icons (i.e. a pencil)

kristinyim:
To be clear the changes are:

- Remove highlighting of the skipped questions and replace it with an alert icon to indicate it needs extra attention.
- Larger Continue button at the bottom should take applicant to where they were previously
- Replace smaller continue to edit to match other questions (so it always says edit regardless of whether question was skipped or not)
- Add edit icons (pencils) next to the edit button.

azizova-leyla:
@kristinyim @rahaghas can you link me to the icons that we should use?

rahaghas:
I suggest changing "Continue" to "Submit application" and throw an error if that button is pressed before all required items are completed (as opposed to disabling it, which can often cause user confusion). I also suggest changing the "Continue" used currently for individual unanswered questions with "Answer" and add a trailing "arrow forward" icon.  
Here's a mock:
<img width="529" alt="Screen Shot 2022-04-22 at 5 30 50 PM" src="https://user-images.githubusercontent.com/103459221/164798875-1c7479e9-110d-4e1a-951a-3db3198ef7c2.png">

Here are the necessary icons:
Edit: https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Aedit%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048

Arrow forward: https://fonts.google.com/icons?icon.query=arrow

azizova-leyla:
I have to pause my work on this issue for the time being to focus on higher priority items.
[...]

PR #3827, opened 2022-11-17
Allow users to answer questions out of order

azizova-leyla:
### Description

DO NOT SUBMIT until translations are ready.

Changes based on https://docs.google.com/presentation/d/1aRaYEP4UBzO-ubS9bqPX1bhoSJlfpH59eRv1X6X9pwI/edit#slide=id.g11d69b7a937_0_176

Added "Answer->" link next to every unanswered question.
Renamed Program application review/preview to Program application summary and cleaned up the code so there's no difference between them.

## Release notes:

Added "Answer->" link next to every unanswered question.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Addresses most of the suggestions from #1920.


server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -175,49 +163,44 @@ private DivTag renderQuestionSummary(
...
-                  .url()
-              : routes.ApplicantProgramBlocksController.review(
+                  .url())
+          .setText(messages.at(MessageKey.LINK_EDIT.getKeyName()))
+          .setIcon(Icons.EDIT)
+          .setIconPosition(LinkElement.IconPosition.START);
+    } else {
+      editElement
+          .setHref(
+              routes.ApplicantProgramBlocksController.edit(
```
nbeloglazov, line 181: Is there a difference between ApplicantProgramBlocksController.edit and ApplicantProgramBlocksController.review? I just tried on staging and html looked the same.

server/app/views/components/LinkElement.java
```diff
@@ -119,10 +127,20 @@ public LinkElement setIcon(Icons icon) {
     return this;
   }
 
+  public LinkElement setIconPosition(IconPosition position) {
```
nbeloglazov, line 130: How about adding icon position as second optional param of `setIcon`? To make sure that they both used together. Also

server/app/views/components/LinkElement.java
```diff
@@ -84,6 +84,14 @@ public final class LinkElement {
   private boolean doesOpenInNewTab = false;
   private Optional<Icons> icon = Optional.empty();
 
+  public enum IconPosition {
```
nbeloglazov, line 87: Given that you added IconPosition - can you update this code to use it as well: https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/views/applicant/ProgramIndexView.java?L322 ?

server/app/views/components/LinkElement.java
```diff
@@ -84,6 +84,14 @@ public final class LinkElement {
...
+    NONE,
```
nbeloglazov, line 88: Why do we need NONE value?

nbeloglazov (changes requested):

server/app/views/applicant/ApplicantProgramSummaryView.java
azizova-leyla: So glad you asked... The difference is what is the next page it will go when the user clicks "save and next"
https://github.com/civiform/civiform/blob/b1157e5bea524ec28db1dec4450e2c1013b3cf6a/server/app/controllers/applicant/ApplicantProgramBlocksController.java#L92-L121

I've looked at that code, and I'm not convinced it actually makes sense to have that distinction, and some of the transitions seem confusing there (e.g. it could navigate backwards). 
It's also not very well tested.
I even started to simplify that logic https://github.com/civiform/civiform/pull/3823. Definitely out of scope for this PR, happy to talk separately.

server/app/views/components/LinkElement.java
azizova-leyla: great idea. done.

server/app/views/components/LinkElement.java
azizova-leyla: It's a good practice to have unset state for enums. Renamed to UNSET to not be confused with visibility none.

nbeloglazov: I think you are referring to recommendations for enums in proto files? And those are dictated by backward compatibility and over-wire transmission (basically that 0 should be set to UNSET/UNSPECIFIED/NONE). I don't think it applies here where there is no mapping to actual number and this enum never leaves java code. If it ever needs to be transmitted we'll have to consider anyway how to serialize properly.

nbeloglazov (approved):

server/app/views/components/LinkElement.java
```diff
@@ -114,15 +122,21 @@ public LinkElement opensInNewTab() {
...
     this.icon = Optional.of(checkNotNull(icon));
+    this.iconPosition = position;
     return this;
   }
 
   public ATag asAnchorText() {
     ATag tag = a();
-    this.icon.ifPresent(icon -> tag.with(Icons.svg(icon).withClasses("mr-2", "w-5", "h-5")));
-    tag.withText(this.text);
+    if (this.iconPosition == IconPosition.END) {
```
bion, line 133: suggestion (blocking): per java style guide
```suggestion
    if (IconPosition.END.equals(this.iconPosition)) {
```

bion (approved):

shanemc-goog:
I see "DO NOT SUBMIT until translations are ready." in the desc.

SHould this have been submitted?

azizova-leyla:
> I see "DO NOT SUBMIT until translations are ready." in the desc.
> 
> @azizova-leyla Should this have been submitted or does the desc need updating?

I added translations, but forgot to update the description.

PR #3834, opened 2022-11-19
Improve custom logout url support

### LINKED ISSUE #3621: Enable OIDC logout during user logout
Currently applicant logs out from CiviForm - we only log them out from CiviForm session. They are still logged in on the corresponding auth provider. Kiana implemented this functionality in https://github.com/civiform/civiform/pull/3171 but it's not enabled by default. And not enabled in any of the deployments. We should enable it by default and work with existing deployments to make sure it works there.

nbeloglazov:
### Description

For some environments the standard OIDC logout flow is not working well. For them we need to use custom logout url. For example for Seattle we need to use the following URL (with no extra URL params): `https://login.seattle.gov/#/logout?appName=CIVIFORM`. This PR improves CiviForm logic to support this use case. In particular it does:

1. Allow custom logout URL to have fragment `#`. Default pac4j logic removes everything after the fragment. Unclear whether it's intentional or it just was implemented that way, before client-side rendering/routing became popular. For context here is pac4j logic: [link](https://sourcegraph.com/maven/com.nimbusds/oauth2-oidc-sdk/-/blob/com/nimbusds/openid/connect/sdk/LogoutRequest.java?L266). 

2. Handle empty `auth.oidc_post_logout_param` config setting as signal to omit logout redirect param from logout request. 

3. Remove unused constructors.

In parallel Seatlle is talking to Oracle about IDCS and whether we maybe using standard logout flow incorrectly. If it gets resolved - then we might not need these changes. But meanwhile I wanted to add support given that it's not many changes.

## Release notes:

Improve support for using custom logout URL.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3621

### DIFF
--- a/server/app/auth/oidc/CustomOidcLogoutRequest.java
+++ b/server/app/auth/oidc/CustomOidcLogoutRequest.java
@@ -4,10 +4,12 @@ import com.google.common.collect.ImmutableMap;
 import com.nimbusds.oauth2.sdk.id.State;
 import com.nimbusds.openid.connect.sdk.LogoutRequest;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import org.pac4j.core.exception.TechnicalException;
 
 /**
  * Custom Logout Request that allows for divergence from the [oidc
@@ -85,8 +87,8 @@ public final class CustomOidcLogoutRequest extends LogoutRequest {
     if (getEndpointURI().getRawFragment() != null) {
       try {
         return new URI(uri + "#" + getEndpointURI().getRawFragment());
-      } catch (Exception e) {
-        throw new RuntimeException(e);
+      } catch (URISyntaxException e) {
+        throw new TechnicalException(e);
       }
     }
     return uri;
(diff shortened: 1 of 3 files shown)


server/app/auth/oidc/CustomOidcLogoutRequest.java
```diff
@@ -110,4 +74,21 @@ public Map<String, List<String>> toParameters() {
...
+  public URI toURI() {
+    URI uri = super.toURI();
+    // default behavior of LogoutRequest.toURI() removes fragment from the URI.
+    // For some usecases (e.g. IDCS on Seattle) they use logout URI that contains
+    // fragment and read it client-side. Here we add fragment back if it was
+    // present in the original logout URI.
+    if (getEndpointURI().getRawFragment() != null) {
+      try {
+        return new URI(uri + "#" + getEndpointURI().getRawFragment());
+      } catch (Exception e) {
```
bion, line 88: suggestion (blocking): please handle a more specific exception or remove the `catch`

bion (changes requested):

server/app/auth/oidc/CustomOidcLogoutRequest.java
nbeloglazov: Oops, thanks. I forgot I used Exception here when was just trying to make it work.

bion (approved):

Issue #3864, opened 2022-11-30
Support JS bundling in CiviForm


nbeloglazov:
The goal is to use JS bundling for serving JS files. The main benefit is that it will allow us to properly reuse code and use third-party libraries. Here is the design doc: https://docs.google.com/document/d/1EmvjzyB2qwkFm1xNcLlGsH33dPHnOZCx648CPVjQNu8/edit?ouid=104182791098567727887&usp=docs_home&ths=true

nbeloglazov:
JS bundles enabled and the old code is cleaned up. This issue is completed. There are some refactorings left in TS code itself, adding unit tests, extracting common functionality into util files. But those are separate small tasks, not part of JS bundling.

PR #3869, opened 2022-12-01
Refactor JS files to use consistent initialization pattern

nbeloglazov:
### Description

Most JS files today use initialization pattern where they check whether certain DOM element is present on the page and if so - initialize themselves. This PR updates all files to use the same pattern. This is required for upcoming JS bundling change where we'll compile all files together for admins and for applicants. So each JS file must handle the case where it's loaded on a page that doesn't actually "need" that file.

## Release notes

No-op refactoring of JS files for upcoming JS bundling chnage.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3864

### DIFF
--- a/server/app/assets/javascripts/azure_delete.ts
+++ b/server/app/assets/javascripts/azure_delete.ts
@@ -3,8 +3,12 @@
  */
 class AzureDeleteController {
   private static FILEUPLOAD_DELETE_ID = 'fileupload-delete-button'
+  private static AZURE_UPLOAD = 'azure-upload'
 
   constructor() {
+    if (document.querySelector(AzureDeleteController.AZURE_UPLOAD) == null) {
+      return
+    }
     const deleteContainer = document.getElementById(
       AzureDeleteController.FILEUPLOAD_DELETE_ID,
     )
--- a/server/app/assets/javascripts/azure_delete.ts
+++ b/server/app/assets/javascripts/azure_delete.ts
@@ -3,10 +3,13 @@
  */
 class AzureDeleteController {
   private static FILEUPLOAD_DELETE_ID = 'fileupload-delete-button'
-  private static AZURE_UPLOAD = 'azure-upload'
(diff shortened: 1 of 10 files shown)


server/app/assets/javascripts/azure_delete.ts
```diff
@@ -3,8 +3,12 @@
  */
 class AzureDeleteController {
   private static FILEUPLOAD_DELETE_ID = 'fileupload-delete-button'
+  private static AZURE_UPLOAD = 'azure-upload'
```
bion, line 6: suggestion (blocking): this value is being used as a query selector, but is missing a `.` at the beginning (assuming this is querying for an HTML class). Please add a test that fails when this selector is incorrect.

bion: suggestion (blocking): please rename to `AZURE_UPLOAD_SELECTOR`

server/app/assets/javascripts/azure_upload.ts
```diff
@@ -3,9 +3,13 @@
  */
 
 class AzureUploadController {
-  static FILEUPLOAD_FORM_ID = 'cf-block-form'
+  private static FILEUPLOAD_FORM_ID = 'cf-block-form'
+  private static AZURE_UPLOAD = '.azure-upload'
```
bion, line 7: suggestion (blocking): please rename to `AZURE_UPLOAD_SELECTOR`

bion (changes requested):

server/app/assets/javascripts/azure_upload.ts
nbeloglazov: Done.

server/app/assets/javascripts/azure_delete.ts
nbeloglazov: I can't add tests at the moment as we don't have infra for TS tests in server folder. I'll need to implement it first. Will do it in parallel.

bion (approved):

PR #3875, opened 2022-12-02
Fail browser tests if detecting JS or network error within browser

nbeloglazov:
### Description

This PR adds a listener to browser JS and network errors. At the end of the test, if the listener detected any errors - it will fail the test providing all caught errors. This is important for e2e tests where we some issues might not be reflected in screenshots or by our normal DOM checks. For example this PR helped to uncover three different issues:

1. Error in QuestionEditView/preview.ts where for "static text" questions we hide "question help text" field. But the element is still present in the DOM and preview.ts thinks that it normal visible element and processes it, causing JS error (example from the detector):

```
Detected 1 unhandled JS exceptions in browser during test run:

Page: http://civiform:9000/admin/questions/new?type=static&redirectUrl=%2Fadmin%2Fquestions
Error: TypeError: Cannot set properties of null (setting 'innerHTML')

  at PreviewController.setTextAndHighlightEnumeratorReferences (http:/civiform:9000/assets/javascripts/4f08e7e0dbbabbed9b34e26f3425033f-preview.js:155:30)
  at PreviewController.updateFromNewQuestionHelpText (http:/civiform:9000/assets/javascripts/4f08e7e0dbbabbed9b34e26f3425033f-preview.js:129:27)
  at new PreviewController (http:/civiform:9000/assets/javascripts/4f08e7e0dbbabbed9b34e26f3425033f-preview.js:16:31)
  at http:/civiform:9000/assets/javascripts/4f08e7e0dbbabbed9b34e26f3425033f-preview.js:308:25
  at BrowserErrorWatcher.failIfContainsErrors (src/support/browser_error_watcher.ts:22:11)
  at resetContext (src/support/index.ts:191:31)
  at Object.<anonymous> (src/support/index.ts:235:11)
```

This error is being fixed in this PR.

2. We have overlapping clickable elements. To prevent click on the inner element triggering the listener on the outer element - we use `Event.preventDefault`. But we haven't added that logic for buttons with `form` attributes. This causing transient failing GET requests:

image

3. Detected that we use Google Analytics even in browser tests which lead to realizing that we we shouldn't have default analytics id set: https://github.com/civiform/civiform/pull/3876

4. Detected JS error in #3896

Lastly fixes linter invocation in `formatter/fmt` script to pass list of files to lint. Currently it's not linting at all.

## Release notes

Minor bug fixes to remove transient failed GET requests during admin question editing. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3864


server/app/assets/javascripts/main.ts
```diff
@@ -363,6 +363,18 @@ function attachRedirectToPageListeners() {
   })
 }
 
+/**
+ * Adds listeners to buttons with 'form' attributes. These buttons trigger form
+ * submission without JS. But we still need to stop propagation of events
+ * because it's possible that some other button up-stream in ancestor chain
```
shanemc-goog, line 369:
```suggestion
 * because it's possible that some other button up-stream in the ancestor chain
```

server/app/assets/javascripts/main.ts
```diff
@@ -363,6 +363,18 @@ function attachRedirectToPageListeners() {
...
+ * contains click listener as we have nested clickable elements.
+ */
+function attachListenersOnFormButtons() {
```
shanemc-goog, line 372: Rename this for what the attached listener does, the naming is too vague as anything could be happening in the listener, without clarity in the name.

EG: attachStopPropogationListenerOnFormButtons.

formatter/fmt
```diff
@@ -65,7 +65,7 @@ else
   npx eslint --fix "app/assets/javascripts/**/*.ts"
   cd ../browser-test
   npm install --silent
-  npx eslint --fix
+  npx eslint --fix "src/**/*.ts"
```
shanemc-goog, line 68: This seem unrelated to this PR, should it be here?

browser-test/src/support/error_pages.ts
```diff
@@ -1,11 +1,13 @@
...
 
 export class NotFoundPage {
-  public page!: Page
+  public ctx: TestContext
 
-  constructor(page: Page) {
-    this.page = page
+  constructor(ctx: TestContext) {
+    this.ctx = ctx
+    ctx.browserErrorWatcher.ignoreErrorsFromUrl(/.*\/ezbezzdebashiboozook/)
```
shanemc-goog, line 10: Can you make ezbezzdebashiboozook a const and reuse it?

browser-test/src/support/browser_error_watcher.ts
```diff
@@ -0,0 +1,75 @@
+import {Page} from 'playwright'
+
+interface ErrorOnPage {
+  message: string
+  url: string
+}
+
+/**
+ * Class that watches for various errors that can happen within browser
```
shanemc-goog, line 9:
```suggestion
 * Class that watches for various errors that can happen within a browser
```

browser-test/src/support/browser_error_watcher.ts
```diff
@@ -0,0 +1,75 @@
...
+ * controlled by Playwright. During test run this class accumulates errors
```
shanemc-goog, line 10:
```suggestion
 * controlled by Playwright. During a test run this class accumulates errors
```

browser-test/src/security.test.ts
```diff
@@ -10,13 +10,17 @@ describe('applicant security', () => {
 
   it('applicant cannot access another applicant data', async () => {
     const {page} = ctx
+    ctx.browserErrorWatcher.ignoreErrorsFromUrl(/.*applicants\/1234\/programs/)
```
shanemc-goog, line 13: It looks like you're trying to piggy back your new test on an existing one.  We should have clear focused testing whenever viable, and not use side effects of other tests which are hard to read and understand later.

Seemingly you should have a browser_error_watcher.test.ts.  It can duplicate a test like this but should otherwise be focused on the error watcher not something else.

shanemc-goog (changes requested):
FYI One of the tests is failing even after I reran it.

browser-test/src/security.test.ts
nbeloglazov: It's not trying to piggyback. This test checks that request returns 401 when user has no access. 401 is considered an error from point of view of BrowserErrorWatcher. So we need to exclude that page from being watched.

formatter/fmt
nbeloglazov: My first PR attempt failed due to lint error and exposed the fact that formatter/format currently doesn't run linter properly on browser-test. Here I'm fixing it. I decided not to pull it out to separate PR as the change is tiny and overhead of creating PR is bigger.

shanemc-goog: Please include in the description then as otherwise it looks like an error.

browser-test/src/security.test.ts
shanemc-goog: Ah, Please add a comment to this setup then that this is knowingly ignoring an error that we are generating.

browser-test/src/security.test.ts
```diff
@@ -10,13 +10,17 @@ describe('applicant security', () => {
...
     await loginAsGuest(page)
     const response = await gotoEndpoint(page, '/applicants/1234/programs')
     expect(response!.status()).toBe(401)
   })
 
   it('admin cannot access applicant pages', async () => {
     const {page} = ctx
+    ctx.browserErrorWatcher.ignoreErrorsFromUrl(
```
shanemc-goog, line 21: Add the comment here too.

shanemc-goog (approved):

PR #3876, opened 2022-12-02
Remove default hardcoded google analytics id and allow it to be empty

nbeloglazov:
### Description

By default we should not be including Google Analytics and only deployments that want it need to enable it. 

#### General

#### User visible changes

#### New Features

### DIFF
--- a/server/app/views/BaseHtmlLayout.java
+++ b/server/app/views/BaseHtmlLayout.java
@@ -10,6 +10,7 @@ import com.google.common.collect.ImmutableList;
 import com.typesafe.config.Config;
 import j2html.tags.specialized.ScriptTag;
 import java.net.URI;
+import java.util.Optional;
 import javax.inject.Inject;
 import play.twirl.api.Content;
 import views.components.ToastMessage;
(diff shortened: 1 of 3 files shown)


server/app/views/BaseHtmlLayout.java
```diff
@@ -93,7 +97,10 @@ public HtmlBundle getBundle(HtmlBundle bundle) {
     bundle.addStylesheets(viewUtils.makeLocalCssTag(TAILWIND_COMPILED_FILENAME));
 
     // Add Google analytics scripts.
-    bundle.addFooterScripts(getAnalyticsScripts(measurementId).toArray(new ScriptTag[0]));
+    measurementId.ifPresent(
```
shanemc-goog, line 100: optional: you don't need () and {} for one liner lambdas.

id -> bundle.addFooterScripts(getAnalyticsScripts(id).toArray(new ScriptTag[0]))

or 

```
measurementId
  .map(id -> getAnalyticsScripts(id).toArray(new ScriptTag[0]))
  .ifPresent(bundle::addFooterScripts);
```

shanemc-goog (approved):

shanemc-goog:
Please coordinate with Seattle to ensure they set this in their config before the next deployment

@gwendolyngoetz @swatkat1

server/app/views/BaseHtmlLayout.java
nbeloglazov: Done.

Issue #3889, opened 2022-12-07
Submitting from review page does not update Application data from the Applicant's answers


shanemc-goog:
Background: In an application flow, when a screen is submitted we update the Applicant to contain that data then seemingly copy the Applicant data into the Application.  (This happens emperially, need to find the code reference)

Steps:
1. Create a Question A
2. Create two programs P1 and P2 and contain Question A.
3. As an Applicant, fill out P1 so that QA = 20  but don't submit.
4. Fill out P2 so that QA = 100
5. View the review page for P1 which shows QA = 100 (from being updates in P2).
6. Click Submit
7. View the Application in the Database or through the Program Admin interface and see that the value is actually 20
 
The value should be the Applicants latest value of 100.

The Submit action only changes the Draft state to Active, it doesn't copy forward the Applicants updated answers into the Application.

shanemc-goog:
@bion

PR #3891, opened 2022-12-07
Ensure application data consistency at submit time

bion:
### Description

When an applicant answers their first question on a program application, CiviForm creates an application record for them associating them with the specific version of the program they're applying to. This application record would also get updated with the contents of the applicant's `ApplicantData` each time they answered a question. When the applicant submitted the application, CiviForm would change its status from `DRAFT` to `ACTIVE`.

This caused a bug where if Program A and Program B shared Question C and an applicant answered Question C in Program A but did not submit the application, then went to program B and changed their answer to Question C, then finally went back to Program A and submitted their application, their application to Program A would contain their original answer to Question C from being answered in Program A despite having shown the updated answer on the block and review pages.

This PR fixes the bug by changing the time that the application's `ApplicantData` gets set from when the applicant answers a question, to when they submit the application.

## Release notes

Fixes a subtle bug where stale question data could be submitted for applications containing shared questions that had been answered in a specific sequence between programs.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/3889

### DIFF
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -105,6 +105,7 @@ public final class ApplicationRepository {
           drafts.isEmpty()
               ? new Application(applicant, program, LifecycleStage.ACTIVE)
               : drafts.get(0);
+      application.setApplicantData(applicant.getApplicantData());
       application.setLifecycleStage(LifecycleStage.ACTIVE);
       application.setSubmitTimeToNow();
       if (tiSubmitterEmail.isPresent()) {
@@ -198,8 +199,7 @@ public final class ApplicationRepository {
               .eq("lifecycle_stage", LifecycleStage.DRAFT)
               .findOneOrEmpty();
       Application application =
-          existingDraft.orElse(new Application(applicant, program, LifecycleStage.DRAFT));
-      application.setApplicantData(applicant.getApplicantData());
+          existingDraft.orElseGet(() -> new Application(applicant, program, LifecycleStage.DRAFT));
       application.save();
       database.commitTransaction();
       return application;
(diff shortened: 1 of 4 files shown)


server/app/repository/ApplicationRepository.java
```diff
@@ -198,8 +199,7 @@ private Application createOrUpdateDraftApplicationInternal(Applicant applicant,
               .eq("lifecycle_stage", LifecycleStage.DRAFT)
               .findOneOrEmpty();
       Application application =
-          existingDraft.orElse(new Application(applicant, program, LifecycleStage.DRAFT));
-      application.setApplicantData(applicant.getApplicantData());
+          existingDraft.orElseGet(() -> new Application(applicant, program, LifecycleStage.DRAFT));
```
bion, line 202: Not part of the bugfix, but no need to instantiate an `Application` if it won't be used

server/test/services/export/AbstractExporterTest.java
```diff
@@ -400,12 +404,14 @@ protected void createFakeProgramWithEnumerator() {
...
     applicationTwo =
         new Application(applicantTwo, fakeProgramWithEnumerator, LifecycleStage.ACTIVE);
+    applicationTwo.setApplicantData(applicantTwo.getApplicantData());
     CfTestHelpers.withMockedInstantNow(
         "2022-02-01T00:00:00Z", () -> applicationTwo.setSubmitTimeToNow());
     applicationTwo.save();
 
     applicationThree =
         new Application(applicantTwo, fakeProgramWithEnumerator, LifecycleStage.OBSOLETE);
+    applicationThree.setApplicantData(applicantTwo.getApplicantData());
```
shanemc-goog, line 414: Does an Obsolete application necessarily come from an ACTIVE one? If it can come from a draft maybe this shouldn't have applicant Data

shanemc-goog (approved):

server/test/services/export/AbstractExporterTest.java
bion: > Does an Obsolete application necessarily come from an ACTIVE one?
I believe so, yes

bion: > Does an Obsolete application necessarily come from an ACTIVE one?
I believe so, yes

PR #3892, opened 2022-12-07
Use "localhoost:<port number>" for test and development environments to fix security vulnerability

jessbumblebee:
### Description
Changes the port definition used for debugging the dev environment and unit tests from *:<portnumber> to <ip-address>:portnumber to avoid a security vulnerability. Servers that expose debug protocol ports are by design vulnerable to remote code execution. By limiting the option to execute code remotely to the specific IP, the setup is less vulnerable because only the docker container with the specific ip can execute on the host machine.

I tested this with bin/run-dev bin/run-tests and can connect the debugger successfully. For browser tests we decided in the eng weekly that debugging the server while running browser tests isn't a common enough use case to justify the additional work due to a more complex setup. I'm therefor removing the option and will document how to put the code back in for single use cases.

## Release notes
N/A

#### General

N/A

#### User visible changes
N/A

#### New Features
N/A

### Issue(s) this completes

N/A


nbeloglazov:
Have you tried connecting to the running server using java debugger? You should be able to do it using intellij remote debugger. Otherwise looks good.

nbeloglazov (approved):

nbeloglazov (approved):

PR #3909, opened 2022-12-09
Add `use_js_bundles` feature flag

nbeloglazov:
### Description

When the flag is enabled JS bundles will be served instead of individual JS files. The plan is to add the flag, flip the default value to true, ensuring that all tests pass, and wait for a release or two. After that I'll clean up the flag and the old approach.

## Release notes

Introduce `USE_JS_BUNDLES ` flag for controlling JS serving mode. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3864

### DIFF
--- a/server/app/views/JsBundle.java
+++ b/server/app/views/JsBundle.java
@@ -1,0 +1,16 @@
+package views;
+
+public enum JsBundle {
+  ADMIN("admin.bundle"),
+  APPLICANT("applicant.bundle");
+
+  private final String jsPath;
+
+  JsBundle(String jsPath) {
+    this.jsPath = jsPath;
+  }
+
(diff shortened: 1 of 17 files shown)


server/app/views/JsBundle.java
```diff
@@ -0,0 +1,16 @@
+package views;
+
+public enum JsBundle {
```
bion, line 3: suggestion (blocking): please add java doc for this that mentions the difference between different bundles

bion (approved):

server/app/views/JsBundle.java
nbeloglazov: Done.

PR #3913, opened 2022-12-10
Make body scrollable on admin pages

nbeloglazov:
### Description

Non-scrollable body makes it impossible for playwright to take full screenshots. And overall non-scrollable body doesn't seem to be necessary here. 

I had to make changes to question bank JS as making body scrollable affected it. 

Additionally cleaned some unused styles from AdminStyles and inlined constants that don't add any context.

#### General

#### User visible changes

#### New Features


server/app/assets/javascripts/toast.ts
```diff
@@ -35,6 +35,7 @@ class ToastController {
       'top-0',
       'transform',
       '-translate-x-1/2',
+      'z-20',
```
shanemc-goog, line 38: Why add z-20 in the screen height PR?

server/app/assets/javascripts/questionBank.ts
```diff
@@ -88,6 +96,18 @@ class QuestionBankController {
     window.history.replaceState({}, '', url.toString())
   }
 
+  static makeBodyNonScrollable() {
+    // When question bank is visible, only the bank should be scrollable. Body
```
shanemc-goog, line 100:
```suggestion
    // When the question bank is visible, only the bank should be scrollable. Body
```

server/app/assets/javascripts/questionBank.ts
```diff
@@ -88,6 +96,18 @@ class QuestionBankController {
...
+    // and all other elements on the page should be non-scrollable.
+    // Using https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
+    // doesn't work as body still scrollable when scrolling over glasspane.
```
shanemc-goog, line 103:
```suggestion
    // doesn't work as body is still scrollable when scrolling over glasspane.
```

shanemc-goog (changes requested):
looks good.

A number of the images appear identical like [question_lifecycle_test/enumerator-only-active.png](https://github.com/civiform/civiform/pull/3913/files#diff-d14c73e38b88be10c9b77cb3cdb54b02fd37fc720536d04cf76d5dabe7b15ace) do you know what's actually different?

shanemc-goog:
Please update the title to specify it's for tests and not, say, the user facing application.

Also the release note should be more appropriate for Civic entities to read, can you update that?  If there are no actual user visible effects then I believe we ignore it for release.

server/app/assets/javascripts/toast.ts
nbeloglazov: I had to add z indices to make things work. When I made body scrollable - text of some elements in body started to be on top of the header during scrolling. To fix it I made z-index of the header to be 10. But then I need to make toast messages to be on top of the header so adding z-20 here.

nbeloglazov:
Originally this change had some minor UI changes related to modals (header became visible). But I updated it to remove that change so modals now cover full screen as they used to and no changes to UI.

nbeloglazov:
Apparently there are some tiny practically invisible changes. For example in this picture apparently 1200 pixels different: 

`Expected image to match or be a close match to snapshot but was 0.138671875% different from snapshot (1278 differing pixels).`

image

nbeloglazov:
Hm, seems like height of the image slightly changes

shanemc-goog (approved):

shanemc-goog:
> Hm, seems like height of the image slightly changes

Ok thanks for looking, doesn't seem like a real user impact and I guess some artifact of the other changes

PR #3914, opened 2022-12-10
Ignore sporadic failures coming from browser error watcher

nbeloglazov:
### Description

Looks like we can't reliably get response from request objects on playwright's `requestfinished` event as those request can belong to sessions that has been closed. It's not a big deal and just means that we might not detect some corner-case failures that don't affect test success anyway. 

I noticed this error while running browser tests locally.

#### General

#### User visible changes

#### New Features


YazeedLoonat (approved):
LGTM

Issue #3859, opened 2022-11-29
Predicate: "earlier than" date evaluates as true for the same date


shanemc-goog:
A date Predicate of "earlier than" 2021-01-01 will evaluate that 2021-01-01 meets the condition.

PR #3915, opened 2022-12-12
update date predicates to not evaluate same date as true

YazeedLoonat:
### Description
Updates the date predicate so that selecting a date that would land on the cut off date, the condition would fail. For example, selecting "earlier than" 2022-12-01, then selecting 2022-12-01 would mean the condition is not met.

## Release notes
Updates the date predicate so that selecting a date that would land on the cut off date, the condition would fail. For example, selecting "earlier than" 2022-12-01, then selecting 2022-12-01 would mean the condition is not met.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/3859


server/app/services/program/predicate/Operator.java
```diff
@@ -16,8 +16,8 @@ public enum Operator {
       "is greater than or equal to",
       ImmutableSet.of(ScalarType.CURRENCY_CENTS, ScalarType.LONG)),
   IN("in", "is one of", ImmutableSet.of(ScalarType.STRING, ScalarType.LONG)),
-  IS_AFTER(">=", "is later than", ImmutableSet.of(ScalarType.DATE)),
-  IS_BEFORE("<=", "is earlier than", ImmutableSet.of(ScalarType.DATE)),
+  IS_AFTER(">", "is later than", ImmutableSet.of(ScalarType.DATE)),
```
shanemc-goog, line 19: I'm thinking IS_AFTER maybe should still be >=  as conceptually date comparisons probably want "on or after" and "before".  or maybe we need a third one to handle that?

Might we worth asking @bion  if there's backing UX research on this

browser-test/src/admin_predicates.test.ts
```diff
@@ -371,13 +371,15 @@ describe('create and edit predicates', () => {
...
     // Earlier than 2021-01-01 is allowed
-    // TODO(#3859): 2021-01-01 evaluates as earlier, but it shouldn't.
     await applicantQuestions.answerDateQuestion('2021-01-02')
     await applicantQuestions.clickNext()
     await applicantQuestions.expectReviewPage()
     await page.goBack()
     await applicantQuestions.answerDateQuestion('2020-12-31')
     await applicantQuestions.clickNext()
+    await page.goBack()
+    await applicantQuestions.answerDateQuestion('2020-01-01')
```
shanemc-goog, line 381: This is in effect double testing the condition which we aren't making a point of doing in this test.  I tink you want to remove this and change the 2021-01-02 above (the false evaluation) to 2021-01-01

shanemc-goog (changes requested):

YazeedLoonat:
@shanemc-goog I think I got this updated per our convo lemme know what ya think!

browser-test/src/admin_predicates.test.ts
```diff
@@ -371,12 +387,15 @@ describe('create and edit predicates', () => {
...
 
     // Earlier than 2021-01-01 is allowed
-    // TODO(#3859): 2021-01-01 evaluates as earlier, but it shouldn't.
-    await applicantQuestions.answerDateQuestion('2021-01-02')
+    await applicantQuestions.answerDateQuestion('2021-01-01')
     await applicantQuestions.clickNext()
     await applicantQuestions.expectReviewPage()
     await page.goBack()
-    await applicantQuestions.answerDateQuestion('2020-12-31')
+    await applicantQuestions.answerDateQuestion('2020-01-01')
```
shanemc-goog, line 394: Suggest using the first date that is valid (2020-12-31). As is it could be almost a year off and we wouldn't detect it.

browser-test/src/admin_predicates.test.ts
```diff
@@ -371,12 +387,15 @@ describe('create and edit predicates', () => {
...
+    await applicantQuestions.clickNext()
+
+    // on or earlier than 2020-01-01 is allowed
```
shanemc-goog, line 397:
```suggestion
    // On or earlier than 2020-01-01 is allowed
```

browser-test/src/admin_predicates.test.ts
```diff
@@ -221,6 +221,9 @@ describe('create and edit predicates', () => {
     await adminQuestions.addTextQuestion({
       questionName: 'depends on previous',
     })
+    await adminQuestions.addTextQuestion({
```
shanemc-goog, line 224: I think you meant for this to be a DateQuestion, I'd move it up with the existing one too.

shanemc-goog: Also update the name to be more readable "on or after"

browser-test/src/admin_predicates.test.ts
```diff
@@ -295,8 +301,18 @@ describe('create and edit predicates', () => {
       '100.01',
     )
 
-    // Date predicate
+    // Date predicate is on or before
     await adminPrograms.goToEditBlockPredicatePage(programName, 'Screen 7')
+    await adminPredicates.addPredicate(
+      'predicate-date',
```
shanemc-goog, line 307: This need to be the new question name you're making.

Currently you're using the same question in both predicates.

browser-test/src/admin_predicates.test.ts
```diff
@@ -371,12 +387,15 @@ describe('create and edit predicates', () => {
...
+    await applicantQuestions.answerDateQuestion('2021-01-01')
     await applicantQuestions.clickNext()
     await applicantQuestions.expectReviewPage()
     await page.goBack()
-    await applicantQuestions.answerDateQuestion('2020-12-31')
+    await applicantQuestions.answerDateQuestion('2020-01-01')
+    await applicantQuestions.clickNext()
+
+    // on or earlier than 2020-01-01 is allowed
+    await applicantQuestions.answerTextQuestion('date-predicate')
```
shanemc-goog, line 398: This should be answering a DateQuestion that the new predicate will evaluate.

shanemc-goog (changes requested):

server/app/services/program/predicate/Operator.java
bion: +1 to Shane's interpretation

browser-test/src/admin_predicates.test.ts
```diff
@@ -214,6 +214,7 @@ describe('create and edit predicates', () => {
       questionName: 'predicate-currency',
     })
     await adminQuestions.addDateQuestion({questionName: 'predicate-date'})
+    await adminQuestions.addDateQuestion({questionName: 'on-or-after-date-question'})
```
shanemc-goog, line 217: Make this and the previous predicate-date name follow a similar naming scheme

browser-test/src/admin_predicates.test.ts
```diff
@@ -295,7 +297,7 @@ describe('create and edit predicates', () => {
       '100.01',
     )
 
-    // Date predicate
+    // Date predicate is on or before
```
shanemc-goog, line 300: this is just "before" and not "on" I believe

server/app/services/program/predicate/Operator.java
shanemc-goog: Let's make sure to sync with Elise and Bloomington folks on the removal of = from IS_AFTER

This will change any predicate evaluation they have on dates, though they may not have any.

shanemc-goog (changes requested):
looking good, need to sync the branch

YazeedLoonat:
Hey @shanemc-goog @bion I think I got this updated with everything, if you both want to take another peek please

bion (approved):

shanemc-goog (approved):

browser-test/src/admin_predicates.test.ts
shanemc-goog: Still need to do this. Use the predicate-date prefix and update the existing to be specific given the new one is now there too.

shanemc-goog (approved):

YazeedLoonat:
Hey @shanemc-goog I got this updated, and also updated some screenshots to line up with the changes, can you take a quick peek and let me know if did that correctly?

shanemc-goog:
> Hey @shanemc-goog I got this updated, and also updated some screenshots to line up with the changes, can you take a quick peek and let me know if did that correctly?

Will do.  Also please click the circle-arrows thing in the reviewer section to help reviewers not lose track.

shanemc-goog:
@YazeedLoonat the screenshots won't load for me, any idea why?

Are they just updating because they use the test names in their display?

shanemc-goog (approved):
I

PR #3926, opened 2022-12-13
Refactor evaluation of Visibility questions and add getting of Eligibility questions

shanemc-goog:
### Description

Adds lookup for which questions are available for eligibility conditions for a given block.

refactors the logic used to make the same determination for visibility so that the main method effectively finds all visible questions up to and including a given block.  Visibility then looks for the questions available through the block before it, and eligibility through its block.

## Release notes

No-op Refactor of logic to determine which questions are available to Visibility conditions to support the in-development eligibility feature.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

### DIFF
--- a/server/app/services/program/ProgramDefinition.java
+++ b/server/app/services/program/ProgramDefinition.java
@@ -488,7 +488,7 @@ public abstract class ProgramDefinition {
   }
 
   /**
-   * Returns a list of the question definitions that may be used to define show/hide predicates on
+   * Returns a list of the question definitions that may be used to define visibility predicates on
    * the block definition with the id {@code blockId}.
    *
    * <p>The available question definitions for predicates satisfy ALL of the following:
@@ -502,6 +502,8 @@ public abstract class ProgramDefinition {
    */
   public ImmutableList<QuestionDefinition> getAvailableVisibilityPredicateQuestionDefinitions(
       long blockId) throws ProgramBlockDefinitionNotFoundException {
+    // Questions through the block before this one are available for this block's visibility
+    // conditions.
     return getAvailablePredicateQuestionDefinitions(blockId - 1);
   }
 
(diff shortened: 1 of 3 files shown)


server/app/services/program/ProgramDefinition.java
```diff
@@ -487,9 +487,42 @@ public ImmutableList<BlockDefinition> getNonRepeatedBlockDefinitions() {
...
+   * Returns a list of the question definitions that may be used to define visibility predicates on
+   * the block definition with the id {@code blockId}.
+   *
+   * <p>The available question definitions for predicates satisfy ALL of the following:
+   *
+   * <ul>
+   *   <li>In a block definition that comes sequentially before the given block definition.
+   *   <li>In a block definition that either has the same enumerator ID as the given block
+   *       definition, or has the same enumerator ID as some "parent" of the given block definition.
+   *   <li>Is not an enumerator.
```
bion, line 500: suggestion (non-blocking): if it's not too much trouble, could you add and implement "Is not a static text question"?

server/app/services/program/ProgramDefinition.java
```diff
@@ -500,10 +533,13 @@ public ImmutableList<BlockDefinition> getNonRepeatedBlockDefinitions() {
    *   <li>Is not an enumerator.
    * </ul>
    */
-  public ImmutableList<QuestionDefinition> getAvailablePredicateQuestionDefinitions(long blockId)
+  private ImmutableList<QuestionDefinition> getAvailablePredicateQuestionDefinitions(long blockId)
```
bion, line 536: suggestion (blocking): looks like the javadoc for this isn't quite accurate after this change --> "In a block definition that comes sequentially before the given block definition."

bion (changes requested):

server/app/services/program/ProgramDefinition.java
shanemc-goog: That should be separate from this refactor as it's unrelated to it.

server/app/services/program/ProgramDefinition.java
shanemc-goog: thanks, c&p error.

I thought it better to make this private which made the doc not technically reference-able in the other methods docs

bion (approved):

PR #3936, opened 2022-12-15
Migrate quesiton edit, API key and translation pages to use standard buttons

nbeloglazov:
### Description

Additionally made view changes:

1. Switched view code from using `LinkButton` to use `ViewUtils.makeSvgButton` and removed code that no longer used.
2. Replaced "Remove" with "Delete" to be consistent with other pages in admin.
3. Updated JS to use `event.currentTarget` instead `event.target` to get element that event listener was registered on instead of element that received the event. That's important as now buttons can have nested elements (svg, span) and event listeners normally care only about button.

## Release notes

Admin question edit page, API key page and translation page now use standard buttons. 

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Issue #3659


mferhut:
Hi Mikita! The button changes look great. I think the only thing is the buttons with an icon look like the padding to the left of the icon within the button looks a bit larger than the padding on the right side. Thank you for making these changes!

https://screenshot.googleplex.com/Krutc3s6ytsouGj

nbeloglazov:
Filed https://github.com/civiform/civiform/issues/3938 to align buttons better

jessbumblebee (approved):

PR #3960, opened 2022-12-17
Update AWS java deps and change S3 setup for localstack

nbeloglazov:
### Description

AWS S3 supports two styles of addressing buckets ([docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html)). For production we use hosted-styled (recommended). For tests we've been using path-styled (not recommended). With the awssdk deps 2.18.0 update they updated underlying endpoint providers that changed our localstack setup to use hosted-styled URLs:

Instead of `http://localstack:4566/civiform-local-s3` 
It now uses `http://civiform-local-s3.localstack:4566/`

It broke tests as that new domain was not setup in Docker. This PR updates docket network and uses `civiform-local-s3.localhost.localstack.cloud` domain for localstack as it explicitly wants `localhost.localstack.cloud` ([docs](https://docs.localstack.cloud/user-guide/aws/s3/)) in order to read Host header. 

This PR should have no effect on production. Though will test on staging once its submitted. 
#### General

#### User visible changes

#### New Features

### DIFF
--- a/server/app/services/cloud/aws/SimpleStorage.java
+++ b/server/app/services/cloud/aws/SimpleStorage.java
@@ -11,6 +11,7 @@ import java.net.URL;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.mockito.Mockito;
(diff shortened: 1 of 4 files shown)


server/app/services/cloud/aws/SimpleStorage.java
```diff
@@ -207,7 +209,16 @@ public S3Presigner getPresigner() {
 
     @Override
     public String bucketAddress() {
-      return String.join("/", localEndpoint, bucket);
+      try {
+        return S3EndpointProvider.defaultProvider()
```
nbeloglazov, line 213: I did this update to be consistent with some other URL which is generated on line 202 `presigner = S3Presigner.builder().endpointOverride(localUri).region(region).build();` That other url uses hosted-style so I updated bucketAddress here to use it as well.

bion (approved):

PR #3969, opened 2022-12-20
Always close BrowserContext after test finishes 

nbeloglazov:
At the moment if test finished but there is an error detected then BrowserContext isn't closed properly. It leads to video not being saved or potentially other inconveniences.

### DIFF
--- a/browser-test/src/support/index.ts
+++ b/browser-test/src/support/index.ts
@@ -194,9 +194,10 @@ export const createTestContext = (clearDb = true): TestContext => {
           ctx.browserErrorWatcher.failIfContainsErrors()
         }
       } finally {
-        // browserErrorWatcher might throw error that should be bubble up all
-        // the way to user. But even if error is thrown we need to close the
-        // browser context, for example to save videos.
+        // browserErrorWatcher might throw an error that should bubble up all
+        // the way to the developer. Regardless whether the error is thrown or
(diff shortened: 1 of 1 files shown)


browser-test/src/support/index.ts
```diff
@@ -189,10 +189,16 @@ export const createTestContext = (clearDb = true): TestContext => {
...
-      if (!DISABLE_BROWSER_ERROR_WATCHER) {
-        ctx.browserErrorWatcher.failIfContainsErrors()
+      try {
+        if (!DISABLE_BROWSER_ERROR_WATCHER) {
+          ctx.browserErrorWatcher.failIfContainsErrors()
+        }
+      } finally {
+        // browserErrorWatcher might throw error that should be bubble up all
+        // the way to user. But even if error is thrown we need to close the
+        // browser context, for example to save videos.
```
jessbumblebee, line 199: Nit: some articles and a "d" are missing

browserErrorWatcher might throw an error that should be bubbled up all the way to the user. However even if an error is thrown...

jessbumblebee (approved):
Please fix the comment before submitting.

browser-test/src/support/index.ts
nbeloglazov: Thanks. Done.

PR #3976, opened 2022-12-21
Refactor TS code to use explicit initialization pattern 

nbeloglazov:
### Description

This makes code more predictable and explicit. Additionally it helps with testability as now modules can be imported in tests and individual functions tested.

#### General

#### User visible changes

#### New Features


server/app/assets/javascripts/accordion.ts
```diff
@@ -32,4 +32,6 @@ class AccordionController {
   }
 }
 
-new AccordionController()
+export function init() {
+  new AccordionController()
+}
```
bion, line 37: suggestion (non-blocking): the pattern I'd prefer to see would be to move more toward OO encapsulation. Each of these files would export the class definition by default so imports would be e.g. `import Accordion from ./accordion`. We'd move event registration code out of the constructors into `registerEventListeners()` methods. Entry point files would look more like:

```typescript
window.addEventListener('load', () => {
  new Accordion().registerEventListeners()
})
```

The overall goal just being to move toward more structure and less free-floating JS code.

Feel free to merge without doing this refactor.

bion (approved):

server/app/assets/javascripts/accordion.ts
nbeloglazov: I'd like to discuss that before merging. What is the benefit of encapsulating function within classes? For majority cases we don't have any state, properties on class instances. Most functions/methods are stateless. Module encapsulation provides same benefits as class encapsulation. So would like to better understand your preference for wrapping code in classes in TS.

The approach leaning towards is essentially this lint rule: https://typescript-eslint.io/rules/no-extraneous-class/

bion: Let's merge as-is for now then

PR #3980, opened 2022-12-21
Fix bin/sbt

nbeloglazov:
It got broken by https://github.com/civiform/civiform/pull/3892


jessbumblebee (approved):
Thanks, good to know that presubmit won't catch all breakages in scripts

PR #3996, opened 2022-12-28
Reenable TS eslint checks

nbeloglazov:
### Description

Reenables a bunch of TS checks so to improve code quality. To reenable the checks had to fix code here and there. No major changes. One of the checks helped to catch non-working test method in ti_dashboard.ts that did nothing while it claimed to assert DOM. 

#### General

#### User visible changes

#### New Features


server/app/assets/javascripts/admin_programs.ts
```diff
@@ -1,3 +1,8 @@
+// !!! DO NOT SUBMIT ME
```
shanemc-goog: ping

shanemc-goog (changes requested):

browser-test/src/support/ti_dashboard.ts
```diff
@@ -33,19 +33,20 @@ export class TIDashboard {
...
     expect(rowText).toContain(client.lastName)
+    // date of birth rendered as <input> rather than plain text.
+    expect(await row.locator('input[name="dob"]').inputValue()).toEqual(
+      client.dobDate,
+    )
   }
 
   async expectDashboardNotContainClient(client: ClientInformation) {
     const tableInnerText = await this.page.innerText('table')
-    expect(tableInnerText).not.toContain(this.convertToMMDDYYYY(client.dobDate))
```
shanemc-goog, line 48: Can this check not be maintained?

shanemc-goog (changes requested):

browser-test/src/support/ti_dashboard.ts
nbeloglazov: Not easily. We need to check that table has no <input> fields with value equal to the date of birth. It gets very specific and britle. Checking that table doesn't contain email and name should be sufficient to ensure that client doesn't is not rendered.

server/app/assets/javascripts/admin_programs.ts
nbeloglazov: Fixed in latest merge.

shanemc-goog (approved):

PR #4000, opened 2022-12-30
Add feature flag

bion:
Add `PREDICATES_MULTIPLE_QUESTIONS_ENABLED` feature flag for admin multi-question predicate UI, default to `false`.

### DIFF
--- a/server/app/featureflags/FeatureFlags.java
+++ b/server/app/featureflags/FeatureFlags.java
@@ -32,7 +32,7 @@ public final class FeatureFlags {
       "program_eligibility_conditions_enabled";
   public static final String PROGRAM_READ_ONLY_VIEW_ENABLED = "program_read_only_view_enabled";
   public static final String PREDICATES_MULTIPLE_QUESTIONS_ENABLED =
-    "predicates_multiple_questions_enabled";
+      "predicates_multiple_questions_enabled";
 
   private final Config config;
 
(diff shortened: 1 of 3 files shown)


server/app/featureflags/FeatureFlags.java
```diff
@@ -58,6 +60,15 @@ public boolean isProgramEligibilityConditionsEnabled() {
     return config.getBoolean(PROGRAM_ELIGIBILITY_CONDITIONS_ENABLED);
   }
 
+  /**
+   * If specifying multiple questions in a predicate is enabled.
+   *
+   * <p>Allows for overrides set in {@code request}.
+   */
+  public boolean isPredicatesMultipleQuestionsEnabled(Request request) {
```
shanemc-goog, line 68: There's an easy-ish test to update for this too. 

https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/test/featureflags/FeatureFlagsTest.java?subtree=true

Just add it to the first couple that check these conditions together for all flags.  

the program read only one could be folded in to it too but wasn't.

shanemc-goog (changes requested):

server/app/featureflags/FeatureFlags.java
bion: done

shanemc-goog (approved):

PR #4011, opened 2023-01-04
Refactor to support a read only version to view a Program

jessbumblebee:
### Description

This change allows setting the ProgramBlockEditView to "DRAFT" or "ACTIVE" at compile time. The version for an ACTIVE program is a a read only version while the one for a DRAFT adds the UI elements tha allow changing the program.

This is the basis for providing a way to view a program without editing it as described in the prd for improving program publishing (https://docs.google.com/document/d/1oD8EaFejs6L7_I-LKxt5ixQRFi5gtcTeL6NOwPMahf0/edit?resourcekey=0-76DVYMd4iDDbpA243q6l_Q#heading=h.atpwj1mzv74n)
This change does not do any user visible changes yet, it only provides the functionality.
It replaces https://github.com/civiform/civiform/pull/3885 based on some engineering discussions that happened partly in the rquest and partly on slack.

Observation: This implementation is not consistent with the implementation of questions. However currently the functionality between Programs and Questions is inconsistent too and based on some conversations with Cassandra we decided to move forward as is and look into consistency across Questions and Programs as a separate topic as there is significant work to be done in that area.

## Release notes

N/A no user visible changes

#### General

- [ x ] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
N/A : Created tests which fail without the change (if possible)
N/A :  Extended the README / documentation, if necessary

#### User visible changes

N/A

#### New Features

N/A

### Issue(s) this completes


server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -28,18 +28,24 @@ protected final DivTag renderProgramInfo(ProgramDefinition programDefinition) {
...
   }
+
+  /** Defines the String that will be shown on the Edit button */
+  protected abstract String getEditButtonText();
+
+  /** Defines the navigation destination for the Edit button */
+  protected abstract String getEditButtonUrl(ProgramDefinition programDefinition);
+
+  /**
+   * Returns the badge which represents the status of the program displayed within this
```
shanemc-goog, line 47:
```suggestion
   * Returns the {@link ProgramDisplayType} which represents the view of the program being displayed.
```

server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -28,18 +28,24 @@ protected final DivTag renderProgramInfo(ProgramDefinition programDefinition) {
...
+   * ProgramBlockView. It will be shown at the top of the page.
```
shanemc-goog, line 48:
```suggestion
   * It will be shown at the top of the page.
```

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -49,45 +53,61 @@
...
+    @Assisted ProgramDisplayType programViewType,
+    AdminLayoutFactory layoutFactory,
+    Config config,
+    FeatureFlags featureFlags) {
     this.layout = checkNotNull(layoutFactory).getLayout(NavPage.PROGRAMS);
     this.featureFlags = checkNotNull(featureFlags);
     this.featureFlagOptionalQuestions = checkNotNull(config).hasPath("cf.optional_questions");
+    this.programDisplayType = programViewType;
+
+    if (!(programDisplayType == DRAFT)) {
```
shanemc-goog, line 98:
```suggestion
    if (!programDisplayType.equals(DRAFT)) {
```

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -128,29 +148,34 @@ public Content render(
...
-                questionBankPanel(
-                    questions,
-                    programDefinition,
-                    blockDefinition,
-                    csrfTag,
-                    QuestionBank.shouldShowQuestionBank(request)))
-            .addModals(blockDescriptionEditModal);
+                                    featureFlags.isProgramEligibilityConditionsEnabled(request)))));
+
+      // Add top level UI  that is only visible in the editable version.
```
shanemc-goog, line 166:
```suggestion
      // Add top level UI that is only visible in the editable version.
```

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -199,38 +224,34 @@ private DivTag addFormEndpoints(InputTag csrfTag, long programId, long blockId)
...
+          .withClasses(AdminStyles.SECONDARY_BUTTON_STYLES, "m-4")
+          .withType("submit")
+          .withId("add-block-button")
+          .withForm(CREATE_BLOCK_FORM_ID));
+    }
     return ret;
   }
 
-  private DivTag renderBlockList(
+  private DivTag blockList(
```
shanemc-goog, line 241: method names should typically start with a verb like render which was removed.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -243,19 +264,27 @@ private DivTag renderBlockList(
...
                   StyleUtils.hover("border-gray-300"),
-                  selectedClasses)
-              .with(
-                  a().withClasses("flex-grow", "overflow-hidden")
-                      .withHref(editBlockLink)
-                      .with(p(blockName), p(questionCountText).withClasses("text-sm")))
-              .with(moveButtons);
+                  selectedClasses);
 
+        String switchBlockLink =
```
shanemc-goog, line 269: suggest to use edit instead of switch. switch implies a transition where this is just the link to the edit route.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -243,19 +264,27 @@ private DivTag renderBlockList(
...
+          controllers.admin.routes.AdminProgramBlocksController.edit(
```
shanemc-goog, line 270: same as my other comment, edit() will throw an exception if loaded on an Active program which I believe this is rendered for.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -397,21 +394,67 @@ private DivTag blockEditPanel(
...
 
-    return div.with(programQuestions, addQuestion);
-  }
+    // Add buttons to change the block.
+    DivTag buttons = div().withClasses("flex", "flex-row", "gap-4");
+
+    // Buttons are only needed when the view is used for editing
+      buttons.with(blockDescriptionModalButton);
+      if (blockDefinitionIsEnumerator) {
+        buttons.with(
```
shanemc-goog, line 431: You can bake the if into this condWith

buttons.condWith(blockDefinitionIsEnumerator, button(".......

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -49,45 +53,61 @@
...
   private static final String DELETE_BLOCK_FORM_ID = "block-delete-form";
+  private static final String NOT_YET_IMPLEMENTED_ERROR =
+    " The read only version of ProgramBlockView is not fully implemented. It should only be "
+      + "used once issue #3162 is closed.";
 
   @Inject
-  public ProgramBlockEditView(
-      AdminLayoutFactory layoutFactory, Config config, FeatureFlags featureFlags) {
+  public ProgramBlockAllAspectsView(
+    @Assisted ProgramDisplayType programViewType,
```
shanemc-goog, line 89: I'm a little wary this is going to WAI with Play's injection model, but let me look into it more.

shanemc-goog (changes requested):

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -49,45 +53,61 @@
...
  *   <li>Edit the name and description
  *   <li>View, add, delete and reorder questions
  *   <li>View and navigate to the visibility criteria
  * </ul>
+ *
+ * For Active program viewing it contains the same elements, but UI elements
+ * that can be used for editing.
  */
-public final class ProgramBlockEditView extends ProgramBlockView {
+public final class ProgramBlockAllAspectsView extends ProgramBlockView {
```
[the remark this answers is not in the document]
jessbumblebee: Sounds good and yes I was trying to find a name that would work and not clash. AbstractProgramBlockView was another option I considered for the parent, but "Base" sounds better. I'm creating an issue that tracks the cleanups (there were a few others that I also wanted to track)

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
jessbumblebee: I generally agree. The reason I changed this is because the entire rest of the class sticks to the pattern of not using a verb. There are some other bits which aren't great, for example, the signatures are much more complicated than they need to be.

In the initial change that used inheritance I included some cleanups that simplify the code and intermitendly also started renaming the methods, but it increased the size of the change quite a bit.Considering how long this change has been dangling, I decided not to include the cleanups here and track them instead.

Basically the idea was: step 1, consistency in the file, step 2 rename all methods to stick to conventions.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
jessbumblebee: I renamed this to switch, because the effect of clicking it is that the view switches the block you are viewing or editing. In the case of a read only version of the view, no editing happens or is allowed and therefore "edit" is misleading.

I think there is a more general Problem with the naming of the class unrelated to this change because the view does not really display only one block, it displays all blocks and you pick one to see the details for. (goes beyond this change though)

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -747,4 +804,44 @@ private Modal blockDescriptionModal(
...
+    }
+  }
+
+  @Override
+  protected String getEditButtonUrl(ProgramDefinition programDefinition) {
+    if (viewAllowsEditingProgram()) {
+      return routes.AdminProgramController.edit(programDefinition.id()).url();
+    } else {
+      try {
+        // TODO(#3162): Manually test all navigation paths once the full flow
```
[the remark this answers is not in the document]
jessbumblebee: I've removed it as I'll do the testing anyways. I only put it there because I thought you'd likely have concerns about this navigation not working fully yet. I've removed it completely and return the not yet implemented error now similar to your suggestion of returning a url that is clearly wrong in the other change.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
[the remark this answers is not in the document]
shanemc-goog: Unresolving so we remember to do it in this PR

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
shanemc-goog: Let's stick-with/go with the verb route as it's is the correct one and this was pre-existing too; IE we shouldn't correct inconsistency towards the incorrect route.

Generally we normalize to inconsistent style when we've affirmatively decided it's better, but that's not the case here.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
shanemc-goog: I don't see that this was applied, am I missing that?

shanemc-goog (changes requested):
just a few replies.  Also if you don't mind, let me close the discussions I start so I can affirm they're resolved.

shanemc-goog:
> Hi Shane
> 
> I have replied to all of your comments and applied the changes that I agreed with. In response to your higher level points:
> 
> 1.) I'll wait for Bion's change to go in, merging with it doesn't look too bad. Let's hope that there won't be too many other merges after my vacation.
> 
> 2.) Yes, the implementation switches to the "DRAFT" version and has no visible changes. The code for the "ACTIVE" version is never excercised. The UI for it is fully implemented, but anything to do with navigation is not yet fully functional (There is some more information in some of my replies to your comments) I left todos in the incomplete parts.
> 
> 3.) From the research I did, what assisted@ is supposed to do, is to support building objects with different values for one of the method parameters. The reason I chose it , is that this happens at compile time, which hopefully eliviates your concern about our injections not being dynamic. As I tested it with "DRAFT", worst case I'd discover any issues with multiple values when I finish the implementation for "ACTIVE", but I'd be surprised if it causes problems. If that's not what you meant, let me know.


#2. I don't think we should be checking in code that isn't working or tested.

It'd be more proper to throw an UnsupportedOperationException in those paths than submit code we believe won't execute yet, especially knowing it does not work.   I understand having it there aids your future development, but that's something we need to manage locally in our dev process and not through the code base.

The problem with doing so is it 1. commits forever orphaned code paths that we may not actually come back to un-orphaned 2. We don't actually know they're orphaned as there's no testing as such.  The intent is that they are, but it's hard to prove a negative.

So please either:
1.  remove the "ACTIVE" paths and throw a java.lang.UnsupportedOperationException, alternatively there's a NetYetImplementedException
2. Finish Implementing the Active path, put it into use and add the relevant testing.



#3 interesting, let me learn more about that.

jessbumblebee:
#2 I am throwing an unsupported operations exception in the constructor which means that none of the other paths can be reached.  I considered adding more exceptions in each method before sending the change to you, but to be honest, I think it just adds more work without actually adding more value. If we don't finish this feature, there will be orphaned code either way and a cleanup will be required with the exception as an entry point. If your main concern are the routes that use edit instead of the not yet provided view I can remove those (but think the value add is small due to the exception at creation time which ensures that this is not reached)

As for the other option you mention (finishing the feature), I prefer to break changes down into logical units  and believe it is a cleaner approach than implementing a whole feature in one change to avoid intermediate steps.

#3 sure. Let me know in case you have any new insights I could learn from.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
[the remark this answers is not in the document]
jessbumblebee: Ok, sorry, I misunderstood you, I thought you meant that we merge and submit and rename as a separate change.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
jessbumblebee: done

shanemc-goog:
> #2 I am throwing an unsupported operations exception in the constructor which means that none of the other paths can be reached. I considered adding more exceptions in each method before sending the change to you, but to be honest, I think it just adds more work without actually adding more value. If we don't finish this feature, there will be orphaned code either way and a cleanup will be required with the exception as an entry point. If your main concern are the routes that use edit instead of the not yet provided view I can remove those (but think the value add is small due to the exception at creation time which ensures that this is not reached)
> 
> As for the other option you mention (finishing the feature), I prefer to break changes down into logical units and believe it is a cleaner approach than implementing a whole feature in one change to avoid intermediate steps.
> 
> #3 sure. Let me know in case you have any new insights I could learn from.

#2  As a principle we shouldn't be checking in incorrect code, code that is untested or written to be untestable.

I think removing the ACTIVE only paths is a small change and doesn't really add work to re-apply later given one path is incorrect as is.

There's only two else branches in the View that trigger for the ACTIVE only path, and the getEditButtonUrl one is incorrect code so should not be submitted as well is unlikely to be the final form of that else block.  So I think those serve little to negative value to keep especially since they're untested code too.

server/app/modules/ProgramCreationModule.java
```diff
@@ -0,0 +1,18 @@
...
+/*
+ * This class is a Guice module which handles bindings and dependency injections
+ * related to the program creation and editing flow for civiform admins. The Guice
+ * module is created when the Play application starts.
+ **/
+public class ProgramCreationModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    install(new FactoryModuleBuilder().build(ProgramBlockAllAspectsView.Factory.class));
```
shanemc-goog, line 16: This makes it to we don't have to write the Factory as `ApplicantQuestionRendererFactory` does correct?

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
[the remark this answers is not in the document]
shanemc-goog: Let's do it in this PR rather than having churn in the permanent repo.

By squared away, I meant resolving the discussions and have approval otherwise so that the incremental change history wasn't lost.

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -243,13 +265,23 @@ private DivTag renderBlockList(
...
-              .with(
-                  a().withClasses("flex-grow", "overflow-hidden")
-                      .withHref(editBlockLink)
-                      .with(p(blockName), p(questionCountText).withClasses("text-sm")))
-              .with(moveButtons);
+                  selectedClasses);
 
+      // TODO(#3162) use a different route when viewAllowsEditingProgram is
+      //  false once a read only navigation option is available.
+      String switchBlockLink =
```
shanemc-goog, line 272: Should this be in the `if viewAllowsEditingProgram()` check since it's the DRAFT route?

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
```diff
@@ -225,7 +225,8 @@ private DivTag addFormEndpoints(InputTag csrfTag, long programId, long blockId)
   private DivTag blockOrderPanel(Request request, ProgramDefinition program, long focusedBlockId) {
     DivTag ret = div().withClasses("shadow-lg", "pt-6", "w-2/12", "border-r", "border-gray-200");
     ret.with(
-        blockList(request, program, program.getNonRepeatedBlockDefinitions(), focusedBlockId, 0));
+        renderBlockList(
+            request, program, program.getNonRepeatedBlockDefinitions(), focusedBlockId, 0));
```
shanemc-goog, line 229:
```suggestion
            request, program, program.getNonRepeatedBlockDefinitions(), focusedBlockId, /* level= */ 0));
```

shanemc-goog: not sure if you added or moved this, but ambiguous params should be documented inline.  the `=` syntax will trigger a linter if the param name changes.

shanemc-goog (changes requested):

shanemc-goog:
I see there's image.tar in this PR, please remove that

shanemc-goog:
> #2 As a principle we shouldn't be checking in incorrect code, code that is untested or written to be untestable.
> 
> I think removing the ACTIVE only paths is a small change and doesn't really add work to re-apply later given one path is incorrect as is.
> 
> There's only two else branches in the View that trigger for the ACTIVE only path, and the getEditButtonUrl one is incorrect code so should not be submitted as well it is unlikely to be the final form of that else block. So I think those serve little to negative value to keep especially since they're untested code too.
> 
> If you did want to preserve it for those later changes, you could copy the file to the side, or; remove the lines commit the change. Under the removal, move your edits into a new branch made off of this PRs one and commit them there for later. But the two elses are small enough you could also just stick them in a file.


What was the resolution here?

shanemc-goog:
FYI: In general we don't remove the description checkboxes if they're not applicable we just leave them unchecked.  Leaving them helps the reviewer also check that the valid ones have been addressed.

server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -1,98 +1,844 @@
...
+                  "border",
+                  "border-white",
+                  StyleUtils.hover("border-gray-300"),
+                  selectedClasses);
+      String editUrl =
+          controllers.admin.routes.AdminProgramBlocksController.edit(
+                  programDefinition.id(), blockDefinition.id())
+              .url();
+      String switchBlockLink =
+          viewAllowsEditingProgram() ? editUrl : NOT_YET_IMPLEMENTED_ERROR_MESSAGE;
```
shanemc-goog, line 277: **issue (blocking):** Throw an exception here rather than rendering an invalid link.  Though I think you can just assume everything is for editing in this PR given the constructor requires it, rather than doing this handling through the method.   It's fine to keep the `viewAllowsEditingProgram` checks elsewhere

server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -1,98 +1,844 @@
...
-    container.with(
-        text(blockName + " is " + predicateDefinition.action().toDisplayString() + " any of:"));
-    UlTag conditionList = ul().withClasses("list-disc", "ml-4", "mb-4");
+  @Override
+  protected String getEditButtonUrl(ProgramDefinition programDefinition) {
+    if (viewAllowsEditingProgram()) {
+      return routes.AdminProgramController.edit(programDefinition.id()).url();
+    }
+    // TODO(#3162) add the route once a read only navigation option is available
+    return NOT_YET_IMPLEMENTED_ERROR_MESSAGE;
```
shanemc-goog, line 833: **issue (blocking):** Same as above, throw an exception rather than create a silent error.

server/app/views/admin/programs/ProgramBlockBaseView.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /** Returns the string that will be shown on the Edit button */
+  protected abstract String getEditButtonText();
+
+  /** Returns the navigation destination for the Edit button */
+  protected abstract String getEditButtonUrl(ProgramDefinition programDefinition);
+
+  /**
+   * Returns the badge which represents the status of the program. It will be shown at the top of
```
shanemc-goog, line 104: **nitpick (blocking):** Update to rephrase "badge" since you renamed it.

server/app/views/admin/programs/ProgramBlockBaseView.java
```diff
@@ -0,0 +1,108 @@
...
+import services.program.predicate.PredicateDefinition;
+import services.program.predicate.PredicateExpressionNode;
+import services.question.types.QuestionDefinition;
+import views.BaseHtmlView;
+import views.ViewUtils;
+import views.ViewUtils.ProgramDisplayType;
+import views.components.Icons;
+import views.style.AdminStyles;
+
+abstract class ProgramBlockBaseView extends BaseHtmlView {
```
shanemc-goog, line 23: **nitpick (blocking):** Can you add a doc for this?  Something about providing utility methods to render the program UI boilerplate, etc

server/app/views/admin/programs/ProgramBlockBaseView.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /** Renders a div with internal/admin program information. */
```
shanemc-goog, line 25: **nitpick (blocking):** Would you mind fleshing out what is rendered more?  Our View javadocs are fairly lacking imo so it's hard to know what is visually being created in methods.

So I think here it'd include rendering the banner with program details

shanemc-goog (changes requested):
Looks good!  Just the small things

server/app/views/admin/programs/ProgramBlockBaseView.java
jessbumblebee: Hi Shane, 

I agree that better documentation would be good. However as this pull request has been hanging for a long time and my second pull request depends on it too, I'd rather leave additional code cleanups of existing code for a separate change. I already created #4019 for cleanups in this area a while back to track some easy cleanups which are nice fillers for when I can't focus on more complex stuff.

server/app/views/admin/programs/ProgramBlockBaseView.java
jessbumblebee: Please see other comment about cleanups tracked in #4019

server/app/views/admin/programs/ProgramBlockBaseView.java
jessbumblebee: Done

server/app/views/admin/programs/ProgramBlockView.java
jessbumblebee: I believe that this is good enough because the class itself throws an exception when it is created "in the wrong state". My other change is ready to replace this (pending review).
It also matches your original suggestion of returning an obviously wrong value rather than using the edit version.

server/app/views/admin/programs/ProgramBlockView.java
jessbumblebee: See other comment. I think this is good enough as my other Pr is ready.

server/app/views/admin/programs/ProgramBlockView.java
```diff
@@ -1,98 +1,844 @@
...
+  private final boolean featureFlagOptionalQuestions;
+  private final ProgramDisplayType programDisplayType;
+
+  public static final String ENUMERATOR_ID_FORM_FIELD = "enumeratorId";
+  public static final String MOVE_QUESTION_POSITION_FIELD = "position";
+  private static final String CREATE_BLOCK_FORM_ID = "block-create-form";
+  private static final String CREATE_REPEATED_BLOCK_FORM_ID = "repeated-block-create-form";
+  private static final String DELETE_BLOCK_FORM_ID = "block-delete-form";
+  private static final String NOT_YET_IMPLEMENTED_ERROR_MESSAGE =
+      " The read only version of ProgramBlockView is not fully implemented. It should only be "
```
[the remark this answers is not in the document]
jessbumblebee: Done

server/app/views/admin/programs/ProgramBlockAllAspectsView.java
jessbumblebee: Can this comment be closed?

shanemc-goog (approved):
Woohoo!  Thanks for this! and the follow on PR!

PR #4012, opened 2023-01-04
Add validation check upon application submit

### LINKED ISSUE #3890: Upon submit, reviewed question data may have changed unknowingly
1. Create a Question A used in Program P1
2. As an Applicant, fill out P1 so that QA = 20 but don't submit, instead stay on the review page indicating QA = 20
3. In another window update P1 so that QA = 100
4. Return to the window in 3 and click submit.

The UI indicated 20 however the Application is saved with 100.

We should detect that the data changed and let the user know to re-review the data before submit.

shanemc-goog:
### Description

Upon submit verify that all questions in the application have been answered. If any have not, show the review page with an appropriate toast message.

This situation will occur if:

1. The application has been filled out and the user is on the page with the Submit button
2. In another tab they change the answer to a visibility question in a way that should show more screens in 1
3. Return to the tab and click submit.

## Release notes

Verify applications are complete when submitted.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3890


shanemc-goog (commented):
baseline

server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -59,7 +59,7 @@ public interface ReadOnlyApplicantProgramService {
    * Get the program block with the lowest index that has missing answer data if there is one.
    * Static questions are marked as incomplete.
    */
-  Optional<Block> getFirstIncompleteBlock();
```
shanemc-goog, line 62: I renamed this as it seemed like for callers the implication was subtle but very important to know in doing the right thing.

browser-test/src/application_navigation.test.ts
```diff
@@ -268,6 +268,31 @@ describe('Applicant navigation flow', () => {
...
+
+      // The UI correctly won't let us submit, so add a submit button and click it.
+      // Note the form already triggers for the submit action.
+      await page.evaluate(() => {
+        const buttonEl = document.createElement('button')
+        buttonEl.id = 'test-form-submit'
+        buttonEl.type = 'SUBMIT'
+        const formEl = document.querySelector('.cf-debounced-form')!
+        formEl.appendChild(buttonEl)
+      })
```
bion, line 286: golf clap.jpg

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -106,12 +117,41 @@ public CompletionStage<Result> submit(Request request, long applicantId, long pr
...
+   *
+   * <ul>
+   *   <li>The application view with submit button contains stale data that has changed visibility
+   *       or eligibility conditions that result in the application being incomplete or ineligible.
+   * </ul>
+   *
+   * @return a {@link ApplicationOutOfDateException} wrapped in a failed future with a user visible
+   *     message for the issue.
+   */
+  private CompletableFuture<Void> validateApplication(
```
bion, line 144: suggestion (blocking): given the validation is specific to submitting, recommend renaming to `validateApplicationForSubmission`

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -106,12 +117,41 @@ public CompletionStage<Result> submit(Request request, long applicantId, long pr
...
                 throw new RuntimeException(cause);
               }
               throw new RuntimeException(ex);
             });
   }
 
+  /**
+   * Validates that the application is complete and correct to submit.
+   *
+   * <p>An application may be submitted but incomplete for a variety of reason:
```
bion, line 134: suggestion (blocking): The comment says "a variety of reasons" but then only lists one. Suggest combining into a single paragraph without the ul

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -106,12 +117,41 @@ public CompletionStage<Result> submit(Request request, long applicantId, long pr
...
+   */
+  private CompletableFuture<Void> validateApplication(
+      ReadOnlyApplicantProgramService roApplicantProgramService, Messages messages) {
+    // Check that all blocks have been answered.
+    if (!roApplicantProgramService.getFirstIncompleteBlockExcludingStatic().isEmpty()) {
+      return CompletableFuture.failedFuture(
+          new ApplicationOutOfDateException(
+              messages.at(MessageKey.TOAST_APPLICATION_OUT_OF_DATE.getKeyName())));
+    }
+    return CompletableFuture.completedFuture(null);
```
bion, line 152: suggestion (blocking): this logic should be encapsulated in a service rather than a controller (remember the litmus test: the same logic should be reusable if we added another endpoint to do this via API). Suggest moving this logic into either the implementation of `applicantService.submitApplication` or as a new method on `AppplicantService`

server/app/services/applicant/ReadOnlyApplicantProgramService.java
bion: ack

server/conf/messages
```diff
@@ -185,6 +185,9 @@ button.submit=Submit
 # Continue application button
 button.continue=Continue
 
+# A toast message shown after submitting an application when it's been determined the application is incomplete typically due to external changes.
+toast.applicationOutOfDate=There''s been an update to the application, please review and answer questions to proceed
+
```
bion, line 190: make a new tab to this sheet and add a row for this message key: https://docs.google.com/spreadsheets/d/1x9xq-X61Gev2c470dPYe69tzI4azLTerZu-jfzFHmTE/edit

bion (changes requested):

server/conf/messages
shanemc-goog: done

server/app/controllers/applicant/ApplicantProgramReviewController.java
shanemc-goog: done

server/app/controllers/applicant/ApplicantProgramReviewController.java
shanemc-goog: done

server/app/controllers/applicant/ApplicantProgramReviewController.java
shanemc-goog: done

browser-test/src/application_navigation.test.ts
```diff
@@ -269,6 +269,31 @@ describe('Applicant navigation flow', () => {
       await validateAccessibility(page)
       await validateScreenshot(page, 'program-submission')
     })
+
+    it('shows error with incomplete submission', async () => {
+      const {page, applicantQuestions} = ctx
+      await loginAsGuest(page)
+      await selectApplicantLanguage(page, 'English')
+      await applicantQuestions.clickApplyProgramButton(programName)
```
dkatzz, line 277: At this point, hasn't the user never applied to the program (since in the beforeAll the admin has just created questions) and they click into the application and they see the summary of what's to come then they click continue to proceed to the first question. I don't see where there's been an update to the application since the user started filling it out in this flow? So confused why they'd see the error at this point? Could be missing something though.

shanemc-goog: The setup for this issue isn't doable in our test framework as it would require 2 browser contexts.

So per the comment below, I'm manipulating the page to let us submit an incomplete application by changing the existing button in the page to do a submit rather than an edit action.  That submit will then result in submitting an incomplete application and the banner.

dkatzz: Ohhh okay that makes sense. Maybe add "The UI correctly won't let us submit, since responses haven't been filled out" or something to the comment below?

dkatzz (approved):

browser-test/src/application_navigation.test.ts
shanemc-goog: done

PR #4016, opened 2023-01-05
Separate enumerator values by semi-colon

### LINKED ISSUE #3934: Enumerator Names difficult to decipher
**Describe the bug**
As a Program Admin when reviewing applications within the browser (that is, NOT in the CSV export), the names in the enumerator are difficult to decipher. There is no comma or semi color or line return to show that there are multiple people. It is a string of names on one line. 

**To Reproduce**

1. Login as a Program Admin
2. Click on an application that has multiple household members (i.e. there are multiple names entered in the enumerator question)
3. See that the names appear in one line, difficult to decipher

**Expected behavior**
I would expect there to be some kind of "break" between names; such as a comma, semi colon or line return.

**Screenshots**
Included

image

**Additional context**

swatkat1:
### Description

Adding list separator i.e. semi-colon between values in an Enum question answer.

## Release notes

Adding list separator i.e. semi-colon between values in an Enum question answer.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3934
### Screenshot
image

### DIFF
--- a/server/test/services/export/PdfExporterTest.java
+++ b/server/test/services/export/PdfExporterTest.java
@@ -227,7 +227,7 @@ public class PdfExporterTest extends AbstractExporterTest {
           + "-- my-file-key UPLOADED (click to download) --\n"
           + "Answered on : 1969-12-31\n"
           + "applicant household members\n"
-          + "item1\n"
+          + "item1;\n"
           + "item2\n"
           + "Answered on : 1969-12-31\n"
           + "applicant ice cream\n"
--- a/server/test/services/export/PdfExporterTest.java
+++ b/server/test/services/export/PdfExporterTest.java
@@ -227,7 +227,7 @@ public class PdfExporterTest extends AbstractExporterTest {
           + "-- my-file-key UPLOADED (click to download) --\n"
           + "Answered on : 1969-12-31\n"
           + "applicant household members\n"
-          + "item1;\n"
+          + "item1\n"
           + "item2\n"
           + "Answered on : 1969-12-31\n"
           + "applicant ice cream\n"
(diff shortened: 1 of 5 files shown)


shanemc-goog:
@swatkat1 is there a reason semicolon was chosen over comma?  I would think comma is a more common expectation.

@mferhut thoughts?

mferhut:
Both can work. Commas usually used for smaller separations and semicolons to show bigger separations.

eg.   Spring, Winter    vs.      Spring; Winter

shanemc-goog:
Would you mind updating the description to be more a little more specific like "Separate enumerator values by semi-colon"

server/test/services/export/PdfExporterTest.java
```diff
@@ -227,7 +227,7 @@ public void statusTrackingDisabled() throws IOException, DocumentException {
           + "-- my-file-key UPLOADED (click to download) --\n"
           + "Answered on : 1969-12-31\n"
           + "applicant household members\n"
-          + "item1\n"
+          + "item1;\n"
```
shanemc-goog, line 230: I don't think we want ; here as there's visual newline separators already

shanemc-goog:
@swatkat1 I think we need to take a different approach rather than changing the value you currently are as it has a lot of downstream implications including I imagine CSV files.

FWIW the Applicant view changes the display to have \<BR\>line breaks, I think the Admin view should be able to do the same thing even if it's using ; instead of \<BR\>

server/test/services/export/PdfExporterTest.java
swatkat1: \n isn't a newline in HTML, it's just a space. Any series of whitespace characters in HTML is treated as one space. So Admins don't know when one name ends and another name begins.

shanemc-goog: This is the PDF test though.  I assume in the PDF the newlines are rendered as newlines; such that now they will have semicolons after them

swatkat1: have changed only the UI now. Replaced "\n" with "; "

shanemc-goog (approved):

Issue #3379, opened 2022-09-09
"Previously answered on YYYY-MM-DD" text on application summary page has wrong date


jeanniefu:
I answered this question on 2022-09-08, but the text displays the wrong day("09" instead of "08"). My guess is that we are not using the correct timezone (but I haven't dug into this yet).

<img width="1757" alt="Screen Shot 2022-09-08 at 5 50 55 PM" src="https://user-images.githubusercontent.com/9094240/189249970-4a1b11c4-9640-4107-ac57-06cb2ed8d0c6.png">

shanemc-goog:
Note: until recently we stored pre-submit data in the application but largely read it from the applicant stored data too.  It seems likely that is having an effect here.

PR #4037, opened 2023-01-11
Update timezone of previously answered on text

dkatzz:
### Description

Using ZoneId.systemDefault() will show the timezone as GMT. By initializing the View with ZoneId and then calling this.zoneId, the zone is "America/Los_Angeles". This also seems to be how this is done in other files, so this seems consistent with other places where we localize a timestamp or date. 

It doesn't seem like there is currently a browser test that checks whether "previously answered state" is set correctly. I was thinking of filing a separate bug for that, but let me know if this seems in scope for this bug. 

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

#3379

### DIFF
--- a/server/app/views/applicant/ApplicantProgramSummaryView.java
+++ b/server/app/views/applicant/ApplicantProgramSummaryView.java
@@ -23,6 +23,7 @@ import java.util.Optional;
 import play.i18n.Messages;
 import play.mvc.Http;
 import play.twirl.api.Content;
+import services.DateConverter;
 import services.MessageKey;
 import services.applicant.AnswerData;
 import services.applicant.RepeatedEntity;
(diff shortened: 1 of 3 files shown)


bion (approved):

bion:
> It doesn't seem like there is currently a browser test that checks whether "previously answered state" is set correctly. I was thinking of filing a separate bug for that, but let me know if this seems in scope for this bug.

best to handle than in a separate PR

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -161,8 +163,7 @@ private DivTag renderQuestionSummary(AnswerData data, Messages messages, long ap
     DivTag actionAndTimestampDiv = div().withClasses("pr-2", "flex", "flex-col", "text-right");
     // Show timestamp if answered elsewhere.
     if (data.isPreviousResponse()) {
-      LocalDate date =
-          Instant.ofEpochMilli(data.timestamp()).atZone(ZoneId.systemDefault()).toLocalDate();
+      LocalDate date = Instant.ofEpochMilli(data.timestamp()).atZone(this.zoneId).toLocalDate();
```
bion, line 166: suggestion (blocking): use [DateConverter](https://github.com/civiform/civiform/blob/main/server/app/services/DateConverter.java)

bion (changes requested):

server/app/views/applicant/ApplicantProgramSummaryView.java
dkatzz: Updated!

bion (approved):

shanemc-goog (approved):

PR #4038, opened 2023-01-11
Fix editing published program with screen ID 1 removed

### LINKED ISSUE #4029: Edit link for an active program without drafts causes error every time after the first ever created screen is removed from the program
**Describe the bug**
When you edit a program, remove the screen that was created first, publish all programs and then edit the program again, an error message is shown complaining that the first screen does not exist. 

I investigated and this is caused bye the value "1" being hardcoded AdminProgramController.newVersionFrom()

https://github.com/civiform/civiform/blob/1dd788827166557ab4c87d2786976617ce6ea06c/server/app/controllers/admin/AdminProgramController.java#L144

**To Reproduce**

1. Log into civiform as a program admin
2. On the program dashboard, choose any program and click edit (either on the draft or the active version)
3. On the left hand side you can see all screens of the program, choose the one that was created first. You can find it by clicking on each of them on the left until you find the one where the url reads something like https://staging-azure.civiform.dev/admin/programs/22/blocks/1/edit. The "1" tells you that this is the oldest screen
4. Delete the screen by clicking "delete screen"
5. Click back
6. On the program Dashboard click "publish all programs", you should see only active programs now, no drafts
7. Next to the program you just edited, click Edit

**Expected behavior**
The edit view should open

**Actual behavior**
An error is shown:
services.program.ProgramBlockDefinitionNotFoundException: Block not found in Program (ID 22) for block definition ID 1

Severity:
Not a total show stopper, but we should fix this fairly soon. While it won't happen all the time, once a program has been affected by it, the bug occurs after each program publishing for as long as the program exists.

seanmalbert:
### Description

This fixes a bug with removing a screen with id 1 and editing the publish draft.

Browser tests are still needed.

## Release notes

This fixes a bug with published programs where the first screen was deleted.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #<issue_number>; Fixes #4029

### DIFF
--- a/server/app/controllers/admin/AdminProgramController.java
+++ b/server/app/controllers/admin/AdminProgramController.java
@@ -141,7 +141,7 @@ public final class AdminProgramController extends CiviFormController {
         // Make a new draft from the provided id.
         idToEdit = programService.newDraftOf(id).id();
       }
-      return redirect(controllers.admin.routes.AdminProgramBlocksController.edit(idToEdit, 1));
+      return redirect(controllers.admin.routes.AdminProgramBlocksController.index(idToEdit).url());
     } catch (ProgramNotFoundException e) {
       return notFound(e.toString());
     } catch (Exception e) {


jessbumblebee:
Hi Sean,

While writing browswer tests myself, I was reminded of this bug.

After my pull request is submitted (https://github.com/civiform/civiform/pull/4069/files) you can add very easily test this by adding the following line at the end of "delete first block and edit" in admin_program_creation.test.ts

    await adminPrograms.gotoViewActiveProgramPage(programName)

That would trigger the bug I believe and thus provide easy testing for you.

seanmalbert:
@jessbumblebee ,

I remembered to check on this again and updated the test to use `await adminPrograms.gotoViewActiveProgramPageAndStartEditing(programName)`. Thank you for the update, that made it super easy.

browser-test/src/admin_program_creation.test.ts
```diff
@@ -405,12 +405,16 @@ describe('program creation', () => {
...
 
     await loginAsAdmin(page)
+    // needed for gotoViewActiveProgramPageAndStartEditing
+    await enableFeatureFlag(page, 'program_read_only_view_enabled')
 
     const programName = 'Test program 5'
     await adminPrograms.addProgram(programName)
     await adminPrograms.addProgramBlock(programName)
     await adminPrograms.removeProgramBlock(programName, 'Screen 1')
     await adminPrograms.gotoEditDraftProgramPage(programName)
```
jessbumblebee, line 415: suggestion (none blocking): Considering that there were two different manifestations of the bugs, it would be nice to add a comment to this line and to line 417 explaining why we call both of them (the different ways in which the bug manifested)

jessbumblebee (approved):

browser-test/src/admin_program_creation.test.ts
seanmalbert: added comments

Issue #4047, opened 2023-01-12
Add check to Block for address validation predicate


seanmalbert:
CF needs the ability to check if a program has an address validation predicate, which uses LeafAddressServiceAreaExpressionNode. The check can live in services/applicant/Block to see if the leaf node is an address service area validation predicate.

Issue #4045, opened 2023-01-12
Create new predicate leaf node type for address service area validation


seanmalbert:
A new leaf node (LeafAddressServiceAreaExpressionNode) is required to specify if a program should validate that a user's address is within a service area.

A case for this new node should also be added to PredicateExpressionNode.

PR #4053, opened 2023-01-12
Add LeafAddressServiceAreaExpressionNode

bion:
### Description

Adds `LeafAddressServiceAreaExpressionNode` along with `BlockDefinition` methods for retrieving them.

#### General

### Issue(s) this completes

Closes https://github.com/civiform/civiform/issues/4047 https://github.com/civiform/civiform/issues/4045

### DIFF
--- a/server/app/services/program/predicate/LeafAddressServiceAreaExpressionNode.java
+++ b/server/app/services/program/predicate/LeafAddressServiceAreaExpressionNode.java
@@ -26,7 +26,7 @@ public abstract class LeafAddressServiceAreaExpressionNode
   @JsonProperty("questionId")
   public abstract long questionId();
 
-  /** The string label of the service area the address should be checked for inclusion in. */
+  /** The string ID of the service area the address should be checked for inclusion in. */
   @JsonProperty("serviceAreaId")
   public abstract String serviceAreaId();
 
@@ -44,7 +44,7 @@ public abstract class LeafAddressServiceAreaExpressionNode
 
   /**
    * Displays a human-readable representation of the assertion in the format "[question name] is in
-   * service area [service area label]".
+   * service area [service area ID]".
    */
   @Override
   public String toDisplayString(ImmutableList<QuestionDefinition> questions) {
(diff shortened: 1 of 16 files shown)


server/app/services/program/predicate/LeafAddressServiceAreaExpressionNode.java
```diff
@@ -0,0 +1,78 @@
...
+/** Represents an assertion that an address question is within a given service area. */
+@JsonTypeName("leafAddressServiceArea")
+@AutoValue
+public abstract class LeafAddressServiceAreaExpressionNode
+    implements ConcretePredicateExpressionNode {
+
+  @JsonCreator
+  public static LeafAddressServiceAreaExpressionNode create(
+      @JsonProperty("questionId") long questionId,
+      @JsonProperty("serviceAreaLabel") String serviceAreaLabel) {
```
seanmalbert, line 21: I'm a little concerned about using the service area label instead of the value. The service area value isn't going to change, but what if the label is changed after an address predicate is added to a program?

bion: Renamed to `serviceAreaId`

server/app/services/program/predicate/LeafAddressServiceAreaExpressionNode.java
```diff
@@ -0,0 +1,78 @@
...
+      @JsonProperty("questionId") long questionId,
+      @JsonProperty("serviceAreaId") String serviceAreaId) {
+    return builder().setQuestionId(questionId).setServiceAreaId(serviceAreaId).build();
+  }
+
+  /** The ID of the address {@link services.question.types.QuestionDefinition} this node checks. */
+  @JsonProperty("questionId")
+  public abstract long questionId();
+
+  /** The string label of the service area the address should be checked for inclusion in. */
```
seanmalbert, line 29: suggestion (non-blocking): cleanup references to label in this file.

seanmalbert (approved):
Just one non blocking suggestion.

Thanks getting to this so quickly!

server/app/services/program/predicate/LeafAddressServiceAreaExpressionNode.java
bion: oh woops, thanks!

PR #4054, opened 2023-01-12
Replace abandoned file-changes-action with one that is still maintained

### LINKED ISSUE #3810: The file-changes-action uses deprecated node version
**Notes on the dependency**

We need to go through and scrub the use of the trilom/file-changes-action. It hasn't been updated in three years and is still using node12. This is a problem because [github actions have deprecated use of node12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12) and will soon be removed.

If we don't replace this our builds will start failing once Github removes node12.

I tried to fork it and upgrade it, but it's looking to be more work to get it updated than I want to deal with. 

Additionally that action [makes use of set-output which is also deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands).

I haven't looked to see if there is an alternative action out there.

gwendolyngoetz:
### Description

The [trilom/file-changes-action](https://github.com/trilom/file-changes-action) hasn't been updated in years. It now uses deprecated github features that will be sunset later this spring. This replaces it with [tj-actions/changed-files](https://github.com/tj-actions/changed-files) which has a thriving community around it and is being maintained.

The only difference I see in output is that each file is surrounded by double-quotes in the new action. 

**Examples**
Before with _Trilom_ `[bin/bin0341_1.txt,bin/bin0341_2.txt,bin/bin1.txt]`
After with _tj-actions_ `["bin/bin0341_1.txt","bin/bin0341_2.txt","bin/bin1.txt"]`

I've tried simple examples in an empty repository similar to how we use the list of changed files. This looks like it should do the job, but it needs to be tested in our full actions to be certain at this point.

When format.yaml looks for files changes one of the triggers is changes to itself. Recommend reviewing the action results post merge with main with a commit that doesn't affect format.yaml.

Finally, push_build_ci only triggers when main is pushed so precheck tests won't be able to check it.

#### General

### Issue(s) this completes

Fixes #3810

### DIFF
--- a/.github/workflows/build_push_ci.yaml
+++ b/.github/workflows/build_push_ci.yaml
@@ -86,18 +86,22 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - id: file_changes
-        uses: trilom/file-changes-action@v1.2.4
+        uses: tj-actions/changed-files@v35
+        with:
+          json: 'true'
         if: ${{ github.event_name != 'workflow_dispatch' }}
 
       - name: Check build & push
         id: check_build
         if: >
           ${{ (github.event_name == 'workflow_dispatch') ||
-            contains(toJSON(steps.file_changes.outputs.files), matrix.file_pattern) ||
-            contains(toJSON(steps.file_changes.outputs.files), matrix.script) ||
-            contains(toJSON(steps.file_changes.outputs.files), '.github') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_push_ci.yaml
```diff
@@ -86,18 +86,22 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
bion, line 90: good improvement!

bion (approved):

PR #4061, opened 2023-01-13
Add browser test for "previously answered on" display

### LINKED ISSUE #3379: "Previously answered on YYYY-MM-DD" text on application summary page has wrong date

dkatzz:
### Description

Adding a browser test screenshot to ensure that previously answered on shows when clicking into a program that has a question that has already been answered

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3379


shanemc-goog:
In the checklist you checked that you added the correct label but I don't see any of them applied

browser-test/src/applicant_application_index.test.ts
```diff
@@ -76,6 +80,10 @@ describe('applicant program index page', () => {
       wantSubmittedPrograms: [primaryProgramName],
     })
 
+    // Check that the question repeated in the second program shows previously answered.
+    await applicantQuestions.clickApplyProgramButton(otherProgramName)
+    await validateScreenshot(page, 'question-shows-previously-answered')
```
shanemc-goog, line 85: **issue (blocking):** This should be in its own test, not the "categorizes programs for draft and applied applications" one

While some of our tests are monolithic they typically are so with a common theme to reduce redundancy or to reduce redundant setup, which isn't a concern here.

shanemc-goog (changes requested):

browser-test/src/applicant_application_index.test.ts
dkatzz: Sounds good - updated to separate into separate test.

dkatzz:
> In the checklist you checked that you added the correct label but I don't see any of them applied

I labeled as dow-swe and testing since this particular PR didn't seem to cover the other labels in the checklist. Should I uncheck that box? Or should this be labeled as a feature request since its paired with a feature change?

shanemc-goog:
> * the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >

You need to add one and only one of these.

`the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >`

browser-test/src/applicant_application_index.test.ts
shanemc-goog: It's ok to keep it in the same file and describe and that lets you re-use the beforeAll setup which is good. 

It just should be it's own `it` test.

shanemc-goog:
Your PR description will go into the release notes since you didn't specify the separate Release Notes

Suggest you update the description to something like : Add browser test for "previously answered on" display

browser-test/src/applicant_application_index.test.ts
shanemc-goog: It's ok to keep it in the same file and describe and that lets you re-use the beforeAll setup which is good. 

It just should be its own `it` test.

shanemc-goog (changes requested):

browser-test/src/applicant_application_index.test.ts
dkatzz: Ohh makes sense! Updated!

browser-test/src/applicant_application_index.test.ts
```diff
@@ -13,7 +14,7 @@ describe('applicant program index page', () => {
   const primaryProgramName = 'Application index primary program'
   const otherProgramName = 'Application index other program'
 
-  beforeAll(async () => {
+  beforeEach(async () => {
```
shanemc-goog, line 17: **suggestion:** You should be able to keep this as Before All if you make the `ctx` creation on L12 to 

` const ctx = createTestContext(/* clearDb= */ false)`

browser-test/src/applicant_application_index.test.ts
```diff
@@ -90,4 +94,56 @@ describe('applicant program index page', () => {
...
+    // Fill out second question.
+    await applicantQuestions.clickContinue()
+    await applicantQuestions.answerTextQuestion('second answer')
+    await applicantQuestions.clickNext()
+    await applicantQuestions.submitFromReviewPage()
+
+    // Check that the original program doesn't show previously answered.
+    await applicantQuestions.returnToProgramsFromSubmissionPage()
+    await applicantQuestions.clickApplyProgramButton(otherProgramName)
+    await validateScreenshot(page, 'question-does-not-show-previously-answered')
```
shanemc-goog, line 133: **issue (blocking):** We should only rely on screenshots for rendering validation and not correctness.  Changes the visual styling can easily be noise to hide correctness bugs.

For correctness we inspect the dom to ensure that it is constructed as expected.  

You'll see examples in other tests where we question the dom and assert it's state.

browser-test/src/applicant_application_index.test.ts
```diff
@@ -90,4 +94,56 @@ describe('applicant program index page', () => {
...
+    await applicantQuestions.clickApplyProgramButton(primaryProgramName)
+    await applicantQuestions.clickEdit()
+    await applicantQuestions.answerTextQuestion('first answer 2')
+    await applicantQuestions.clickNext()
+    await applicantQuestions.submitFromReviewPage()
+
+    // Check that the first program now shows the previously answered text.
+    await applicantQuestions.returnToProgramsFromSubmissionPage()
+    await applicantQuestions.clickApplyProgramButton(otherProgramName)
+    await validateScreenshot(page, 'first-program-shows-previously-answered')
```
shanemc-goog, line 147: **issue (blocking):** "first" won't make sense among all the screenshots, and isn't readable for what it's indicating, plus as far as a screenshot it's not important to its use, you could likely use it in different situations where it isn't the first.  It is a little tricky since there is some context in the test that usually matters but not always for the suite.

Suggestion: "other-program-shows-previously-answered-on"

shanemc-goog (changes requested):

browser-test/src/applicant_application_index.test.ts
dkatzz: Oh good call - updated!

browser-test/src/applicant_application_index.test.ts
dkatzz: Makes sense! Updated to add a dom assertion instead.

browser-test/src/applicant_application_index.test.ts
dkatzz: Yeah, makes sense! Updated.

shanemc-goog:
This should have the ignore not the feature label BTW. We use feature for customer facing features that aren't flag guarded.

browser-test/src/applicant_application_index.test.ts
```diff
@@ -141,9 +142,10 @@ describe('applicant program index page', () => {
     await applicantQuestions.clickNext()
     await applicantQuestions.submitFromReviewPage()
 
-    // Check that the first program now shows the previously answered text.
+    // Check that the other program now shows the previously answered text.
     await applicantQuestions.returnToProgramsFromSubmissionPage()
     await applicantQuestions.clickApplyProgramButton(otherProgramName)
-    await validateScreenshot(page, 'first-program-shows-previously-answered')
+    await applicantQuestions.validatePreviouslyAnsweredText()
```
shanemc-goog, line 148: **suggestion (non-blocking):** It would be nice if we could verify that the correct question has the text instead of just any question has it. 

The `other` program only has 1 question, You could reverse how the programs are used so you're checking the program with 2 questions.

To verify you could follow this pattern.  Note that it uses a [ReferenceClass added to the dom element](https://sourcegraph.com/search?q=APPLICANT_NOT_ELIGIBLE_TEXT+repo%3Aciviform%2Fciviform%24&groupBy=repo) and [used in the .ts helper](https://sourcegraph.com/search?q=cf-applicant-not-eligible-text+repo%3Aciviform%2Fciviform%24&groupBy=repo)

https://sourcegraph.com/github.com/civiform/civiform@main/-/blob/browser-test/src/application_navigation.test.ts?L363

shanemc-goog (approved):

browser-test/src/applicant_application_index.test.ts
dkatzz: Makes sense! Thanks for that example. I realized there is a bt-date attribute on the previously answered on element, so I used that and added those checks. Let me know what you think.

shanemc-goog: Thank you!!

We should add a different class though instead of relying on that just happening to be there as it's effectively different from this need.

FWIW the .bt- classes are specifically added to indicate which items should be scrubbed to make screenshot validation stable, so dates, times, ids, etc.   bt stands for Browser Test

shanemc-goog (changes requested):

shanemc-goog (approved):

Issue #4003, opened 2022-12-30
"Previously answered on" is not translated


shanemc-goog:
https://sourcegraph.com/github.com/civiform/civiform@shanemc-eligibility-review-page/-/blob/server/app/views/applicant/ApplicantProgramSummaryView.java?L186

PR #4063, opened 2023-01-13
Add message key for previously answered on

dkatzz:
### Description

Added message key and english translation for "Previously translated on"

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

#4003


shanemc-goog:
Make sure to go through the checklist, this is likely a "feature" since it's user visible.

server/conf/messages.en-US
```diff
@@ -189,6 +189,9 @@ content.notEligible=Not Eligible
 link.edit=Edit
 link.answer=Answer
 
+# The text that appears next to a question that was answered in another program to let the user know the date it was previously answered on.
+content.previouslyAnsweredOn=Previously answered on
```
shanemc-goog, line 193: **issue (blocking):** I believe you should include the date substitution here so it can be properly placed in the translated content.  See the other examples with `{0}`

shanemc-goog (changes requested):

dkatzz:
> Make sure to go through the checklist, this is likely a "feature" since it's user visible.

Sounds good! Didn't think I need a feature flag - can just submit this when the translations are complete? But let me know if I'm wrong? I was adding the browser test in a separate CL, so I didn't check those off

server/conf/messages.en-US
shanemc-goog: I'm not sure what you mean by resubmit, but yeah it's a different message since it includes the substitution.

shanemc-goog:
> > Make sure to go through the checklist, this is likely a "feature" since it's user visible.
> 
> Sounds good! Didn't think I need a feature flag - can just submit this when the translations are complete? But let me know if I'm wrong? I was adding the browser test in a separate CL, so I didn't check those off

Feature Flag: yeah don't need that.
Browser test: Given this is existing code and the change is reasonably a no-op doing that separately is fine.

The checklist is meant as a reminder of things to consider, not that everything needs to be checked.  But for the label one you checked it but didn't apply one of the labels.

server/conf/messages.en-US
dkatzz: Sorry I meant create a new localization request with the {0}.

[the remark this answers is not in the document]
dkatzz: Ah thanks, updated!

shanemc-goog (approved):

PR #4076, opened 2023-01-18
Fix formatting on "Static Question" in admin preview page

dkatzz:
### Description

In the admin preview page, the check to format a static question wasn't set up correctly, so these questions weren't getting formatted. Updated the check to use the right name for a static question and added a browser test to ensure the formatting was working as expected. 

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/assets/javascripts/preview.ts
```diff
@@ -34,6 +34,8 @@ class PreviewController {
   private static readonly DEFAULT_ENTITY_TYPE = 'Sample repeated entity type'
   private static readonly DEFAULT_OPTION_TEXT = 'Sample question option'
 
+  private static readonly STATIC_QUESTION_TYPE = 'Static Text'
```
shanemc-goog, line 37: **issue (blocking):** TYPE should be TEXT here I believe right?

shanemc-goog, line 37: **suggestion (non-blocking):** Can you add a comment that this is from `{@link QuestionType}` ?

shanemc-goog, line 37: **suggestion (non-blocking):** Related can you comment above the previous three on L33 that they come from `{@link ApplicantQuestionRendererFactory}`

shanemc-goog (approved):

server/app/assets/javascripts/preview.ts
dkatzz: I put type because it was coming from the question types, but updated to text since I think that makes sense too.

dkatzz: Done

dkatzz: Done

shanemc-goog: Text/Type looks here to indicate what data item is being matched and not the broader source/file of that info.

PR #4107, opened 2023-01-25
Fix displaying CSV values in new (feature flagged) admin predicate editor

### LINKED ISSUE #4064: Predicate editor v2: Editing deserialization of multi values doesn't remove brackets/quotes
For scalars operations that allow for multiple values, when you edit them brackets are added in the value erroneously.  The brackets are likely the json serialization being written back directly to the UI rather than being scrubbed for the UI.  In the v1 UI we only ever display the value, without re-edit, and that likewise includes the brackets.

It's likely the quotes should be omitted too.

Saving and re-editing continously wraps the value in quoted brackets
"20008"", ""20001" ->
["20008"", ""20001"] ->
["["20008"", ""20001"]"]

image

bion:
### Description

- Fixes a bug where CSV values were displayed incorrectly in the predicate editor.
- Fixes a bug where attempting to create a new predicate when a new one already exists, with different questions that the original predicate, would cause a `NullPointerException`

#### General

#### User visible changes

- [x] Wrote browser tests 

### Issue(s) this completes

Fixes #4064


browser-test/src/admin_predicates.test.ts
```diff
@@ -355,6 +357,18 @@ describe('create and edit predicates', () => {
           operator: 'is less than',
           values: ['10', '20'],
         },
+        {
+          questionName: 'list of longs',
+          scalar: 'number',
+          operator: 'is one of',
```
dkatzz, line 363: **suggestion:** Consider changing one of these to "is any of"? Since I think that would allow you to select multiple and it would be displayed differently?

dkatzz (approved):

browser-test/src/admin_predicates.test.ts
bion: This specific bug isn't for multi-value questions, but rather for single value questions where the predicate comparison value is using CSV format to offer multiple multiple values.

bion: Done and updated the comparison values used in the test to make that more clear.

PR #4109, opened 2023-01-26
Promote status tracking to a default feature.

### LINKED ISSUE #3942: Promote Status Tracking to a default feature
We've heard from Bloomington and Seattle that the MVP feature is working as expected.

The code should be updated to so the feature is a default part of the system with no flag guard.

Specifically remove the [Feature flag](https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/featureflags/FeatureFlags.java?L29&subtree=true), and all code that conditions the feature off based on it.

shanemc-goog:
### Description

Makes status tracking an always on feature and removes the control to enable/disable it.

The application.conf values are maintained though have no effect, as removal will cause errors in deployments that have them configured.

## Release notes

Promotes the Status Tracking feature to be always available.  Exported data will now include "Status" fields if the feature had not been enabled already.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #3942

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -655,6 +655,7 @@ allow_civiform_admin_access_programs = ${?ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS}
 # Preview features.
 
 # In development features.
+# Deprecated Jan 2023, this is a default feature.
 application_status_tracking_enabled = false
 application_status_tracking_enabled = ${?CIVIFORM_APPLICATION_STATUS_TRACKING_ENABLED}
 program_read_only_view_enabled = false
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -655,8 +655,8 @@ allow_civiform_admin_access_programs = ${?ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS}
 # Preview features.
 
 # In development features.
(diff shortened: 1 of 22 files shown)


server/conf/application.conf
```diff
@@ -655,7 +655,8 @@ allow_civiform_admin_access_programs = ${?ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS}
 # Preview features.
 
 # In development features.
-application_status_tracking_enabled = false
+# Status Tracking is a default feature as of Jan 2023, this control is deprecated.
+application_status_tracking_enabled = true
 application_status_tracking_enabled = ${?CIVIFORM_APPLICATION_STATUS_TRACKING_ENABLED}
```
bion, line 660: question (non-blocking): why not remove this config value altogether?

bion (approved):

server/conf/application.conf
shanemc-goog: I seem to recall that removing this or maybe CIVIFORM_APPLICATION_STATUS_TRACKING_ENABLED , if deployment scripts still use them can cause issues.  Basically that all references need to be removed first.

Does that sound right?

bion: Not that I can remember, doesn't hurt to take this approach though

Issue #4071, opened 2023-01-17
"Error saving application" banner is not translated


shanemc-goog:
https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/controllers/applicant/ApplicantProgramReviewController.java?L155:64&subtree=true

PR #4114, opened 2023-01-26
i18n for 'Error saving application' banner

nb1701:
### Description

This adds i18n support for the 'Error saving application' banner, which shows when an exception occurs while saving the application.

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

https://github.com/civiform/civiform/issues/4071

### DIFF
--- a/server/conf/messages
+++ b/server/conf/messages
@@ -277,6 +277,9 @@ toast.mayNotQualify=Based on your responses, you may not qualify for the {0}. If
 # Toast message that shows that a client may not be eligible for a program, when someone else is filling out the application on a client's behalf.
 toast.mayNotQualifyTi=Based on your responses, your client may not qualify for the {0}. If your client's information has changed you can edit your answers and then proceed with the application.
 
+# Error when there was an exception while submitting the application
+banner.errorSavingApplication=Error saving application
+
 #---------------------------------------------------------------------------------------------#
 # APPLICANT APPLICATION CONFIRMATION PAGE - text for a page confirming application submission #
 #---------------------------------------------------------------------------------------------#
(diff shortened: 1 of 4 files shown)


nb1701:
I wasn't quite sure what to make the message key name be.  There are no other `banner` ones, but I wasn't sure if `content` was appropriate?

nb1701:
Tested by adding `return CompletableFuture.failedFuture(new ApplicationSubmissionException(applicantId, programId));` to https://github.com/civiform/civiform/blob/442d80a7f556cda5ce7d4c87717617d0225766f2/server/app/services/applicant/ApplicantService.java#L320, commenting out the rest of the function, and turning off -Werror.  Probably a better way, but verified changes to the `messages` file are reflected in the banner.

server/conf/messages
```diff
@@ -213,6 +213,9 @@ ariaLabel.edit=Edit {0}
 # Program Summary Title
 title.programSummary=Program application summary
 
+# Error when there was an exception while submitting the application
+banner.errorSavingApplication=Error saving application
+
```
bion, line 218: **suggestion (blocking):** also add to messages.en-US

bion (approved):
`banner.errorSavingApplication` looks like a fine message key to me!

server/conf/messages
nb1701: @bion Do we need to have text in the other localized files as well?  We don't have translations yet, but I'm guessing it will throw an error if it tries to find the key and it doesn't exist.  Do we just put the English text in there for now?

nb1701: Answered in channel

PR #4120, opened 2023-01-27
Program Card changes- Reassigning External Link as ProgramDetails and existing ProgramDetails as backup.

swatkat1:
### Description

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

### DIFF
--- a/server/app/views/applicant/ProgramIndexView.java
+++ b/server/app/views/applicant/ProgramIndexView.java
@@ -286,41 +286,41 @@ public final class ProgramIndexView extends BaseHtmlView {
               preferredLocale, cardData.latestSubmittedApplicationStatus().get()));
     }
     programData.with(title, description);
-
-    // Add info link.
-    String infoUrl =
-        controllers.applicant.routes.ApplicantProgramsController.view(applicantId, program.id())
-            .url();
-    ATag infoLink =
-        new LinkElement()
-            .setId(baseId + "-info-link")
-            .setStyles("mb-2", "text-sm", "underline")
-            .setText(messages.at(MessageKey.LINK_PROGRAM_DETAILS.getKeyName()))
-            .setHref(infoUrl)
-            .asAnchorText()
-            .attr(
-                "aria-label",
-                messages.at(
-                    MessageKey.LINK_PROGRAM_DETAILS_SR.getKeyName(),
-                    program.localizedName().getOrDefault(preferredLocale)));
-    programData.with(div(infoLink));
-
     // Add external link if it is set.
     if (!program.externalLink().isEmpty()) {
       ATag externalLink =
           new LinkElement()
               .setId(baseId + "-external-link")
               .setStyles("mb-2", "text-sm", "underline")
-              .setText(messages.at(MessageKey.EXTERNAL_LINK.getKeyName()))
+              .setText(messages.at(MessageKey.LINK_PROGRAM_DETAILS.getKeyName()))
               .setHref(program.externalLink())
               .opensInNewTab()
               .setIcon(Icons.OPEN_IN_NEW, LinkElement.IconPosition.END)
               .asAnchorText()
               .attr(
                   "aria-label",
-                  messages.at(MessageKey.EXTERNAL_LINK_OPENS_IN_NEW_TAB.getKeyName()));
+                  messages.at(MessageKey.LINK_PROGRAM_DETAILS_SR.getKeyName()));
 
       programData.with(div(externalLink));
+    }
+    else {
+      // Add info link.
+      String infoUrl =
+        controllers.applicant.routes.ApplicantProgramsController.view(applicantId, program.id())
+          .url();
+      ATag infoLink =
+        new LinkElement()
+          .setId(baseId + "-info-link")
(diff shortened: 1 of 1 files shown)


server/app/views/applicant/ProgramIndexView.java
```diff
@@ -286,41 +286,37 @@ private LiTag programCard(
...
-              .setText(messages.at(MessageKey.EXTERNAL_LINK.getKeyName()))
+              .setText(messages.at(MessageKey.LINK_PROGRAM_DETAILS.getKeyName()))
               .setHref(program.externalLink())
               .opensInNewTab()
               .setIcon(Icons.OPEN_IN_NEW, LinkElement.IconPosition.END)
               .asAnchorText()
-              .attr(
-                  "aria-label",
-                  messages.at(MessageKey.EXTERNAL_LINK_OPENS_IN_NEW_TAB.getKeyName()));
+              .attr("aria-label", messages.at(MessageKey.LINK_PROGRAM_DETAILS_SR.getKeyName()));
```
dkatzz, line 300: **suggestion (blocking):** This message string includes a [message](https://sourcegraph.com/github.com/civiform/civiform@5de389d/-/blob/server/conf/messages?L152) about what program it's for, so change to `messages.at(MessageKey.LINK_PROGRAM_DETAILS_SR.getKeyName(), program.localizedName().getOrDefault(preferredLocale))` or change to LINK_PROGRAM_DETAILS.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -286,41 +286,37 @@ private LiTag programCard(
...
+    // Add external link if it is set else info link.
     if (!program.externalLink().isEmpty()) {
       ATag externalLink =
           new LinkElement()
               .setId(baseId + "-external-link")
               .setStyles("mb-2", "text-sm", "underline")
-              .setText(messages.at(MessageKey.EXTERNAL_LINK.getKeyName()))
+              .setText(messages.at(MessageKey.LINK_PROGRAM_DETAILS.getKeyName()))
               .setHref(program.externalLink())
               .opensInNewTab()
```
dkatzz, line 297: **suggestion:** Seems a little funny to me that this would open in a new tab and the other wouldn't, since now they have the same text. Thoughts on using the conditional to set the href value, but have the externalLink definition be the same in either case to ensure the same look and functionality?

swatkat1: Actually it is a good call to use a conditional statement outside to assign the href and use the same formatting. I had thought about it but later forgot. Have made the changes and used the most of programDetails formatting and added .opensInNewTab() and .SetIcon() from externalLink..

server/app/views/applicant/ProgramIndexView.java
swatkat1: Got rid of this block. We don't need external link anymore. Confuses the residents.

browser-test/src/application_navigation.test.ts
```diff
@@ -154,23 +154,18 @@ describe('Applicant navigation flow', () => {
...
-
-    it('verify program details page', async () => {
-      const {page, applicantQuestions} = ctx
-      await loginAsGuest(page)
-      await selectApplicantLanguage(page, 'English')
-      await applicantQuestions.clickProgramDetails(programName)
 
-      // Verify we are on program details page. Url should end in "/programs/{program ID}"
-      expect(page.url()).toMatch(/\/programs\/[0-9]+$/)
+      const cardHtml = await page.innerHTML(
```
dkatzz, line 158: **suggestion:** Here we verify that the program with the external link has the correct URL. Could we also do a validation that the first program (with program details link) has the correct URL? Maybe by leaving in the clickProgramDetails() and asserting that way?

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -286,17 +286,21 @@ private LiTag programCard(
               preferredLocale, cardData.latestSubmittedApplicationStatus().get()));
     }
     programData.with(title, description);
-
-    // Add info link.
-    String infoUrl =
-        controllers.applicant.routes.ApplicantProgramsController.view(applicantId, program.id())
-            .url();
+    // Add external link if it is present else use the default Program details page
```
dkatzz, line 289: **suggestion:** Consider updating to something like: "Navigate to the external link if it is set; otherwise, navigate to the default program details page"

swatkat1: Navigation happens after the button click, we are just using this here.

browser-test/src/application_navigation.test.ts
swatkat1: We cannot directly test it as it opens in a new tab. I have added the verify_program_details test and managed to test the URL using the popup functionality of TS.

dkatzz: Sounds good, thank you for adding that!!

server/app/views/applicant/ProgramIndexView.java
dkatzz: Perfect, thanks!

dkatzz (approved):

PR #4130, opened 2023-01-30
Removing Seattle Staging Hostname

gwendolyngoetz:
### Description

This is now configured in Seattle's repository. Removing the hard-coded valued of Seattle's staging site in order to clean up the config file.

Without default staging_hostname to "" the application breaks on localhost.

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -184,7 +184,7 @@ adfs.ad_groups_attribute_name = ${?AD_GROUPS_ATTRIBUTE_NAME}
 base_url = "http://localhost:9000"
 base_url = ${?BASE_URL}
 # Staging hostname should not start with https(s)://.
-staging_hostname = "staging.seattle.civiform.com"
+staging_hostname = ""
 staging_hostname = ${?STAGING_HOSTNAME}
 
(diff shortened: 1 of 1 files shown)


server/conf/application.conf
```diff
@@ -184,7 +184,7 @@ adfs.ad_groups_attribute_name = ${?AD_GROUPS_ATTRIBUTE_NAME}
 base_url = "http://localhost:9000"
 base_url = ${?BASE_URL}
 # Staging hostname should not start with https(s)://.
-staging_hostname = "staging.seattle.civiform.com"
+staging_hostname = ""
```
bion, line 187: **issue (blocking):** CiviForm uses the staging hostname to detect whether not it's in staging mode, this value should be set to whatever the new domain is: https://github.com/civiform/civiform/blob/main/server/app/controllers/dev/DevController.java#L33

bion (changes requested):

server/conf/application.conf
bion: Oh I just realized this is the main repo and not Seattle-specific. So long as this value is correctly set in the Seattle staging deployment code then you're good to go!

bion (approved):

PR #4138, opened 2023-01-31
Switching from using full state name to state abbreviation

gwendolyngoetz:
### Description
Our address control uses a dropdown for state. The dropdown is configured solely as two character state abbreviations. Currently the ESRI client is returning the full name of the state. This switches it from using the field `Region` to `RegionAbbr` in the ESRI reponse. 

So going from using `Washington` to just `WA`.

#### General


seanmalbert (approved):
Looks good, @gwendolyngoetz !

PR #4150, opened 2023-02-02
Fix vscode-setup

nb1701:
### Description

Previously, bin/sbt was not taking any additional arguments and passing them along to docker::dev_and_test_server_sbt_command. This meant that when vscode-setup tried running 'bin/sbt makePom', it wouldn't actually do that and it would just sit in the shell. Also, for some reason, docker compose can't seem to figure out what the 'civiform-shell' service is, even though it's defined in docker-compose.dev.yml.  However, we're just using the civiform service anyway, so this fixes the cp command to use the civiform service to copy the target directory to the local server directory so we can munge the pom file appropriately.

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

### DIFF
--- a/bin/sbt
+++ b/bin/sbt
@@ -8,5 +8,5 @@ docker::set_project_name_dev
 bin/pull-image
 
 docker::compose_dev up --wait -d
-docker::dev_and_test_server_sbt_command 8457
+docker::dev_and_test_server_sbt_command 8457 $1
 docker::compose_dev stop civiform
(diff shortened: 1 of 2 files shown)


nb1701:
Thanks @rockycodes for noticing we weren't passing arguments into the docker command, so `makePom` wasn't doing anything!

bion (approved):

PR #4155, opened 2023-02-02
Fix missing bold fonts in screenshot tests (1/2)

### LINKED ISSUE #3225: Fix text bolding in browser screenshot tests
When running browser tests with our default command (bin/run-browser-tests), the screenshots that are taken do not appear to show bolded text. We should investigate why this is, and fix it so that the screenshots are more representative of what is shown in prod.

From some initial investigation:
* We currently use Chromium for our browser tests. If we use e.g. webkit or firefox, then the bolding works.
* When we run the test locally (e.g. not through docker), the bolding also works.

One hypothesis: 
* I believe we are using the default font, which is specified by a combination of OS and browser. Perhaps when running through docker, we use a stripped down version of fonts where the bolded versions are not included (?). We could try specifying our own default font that CiviForm should use (should check with UX).

avaleske:
### Description

Add a fonts.conf to the browser-tests docker container. Part 1/2. 

Details:
- Part 2 is #4239 and updates the goldens.
- This is split into two parts so that the presubmits don't break, since there's an issue where they're using the old docker container for the presubmit run.
- The font used by Chromium in the browser-tests docker image did not have a full range of bolds, so bold fonts were rendered as normal weights.
- To fix this, we override the font picked by the OS for the `system-ui` font-family to be `Ubuntu`, which has a full range of bold weights.
- Additional debugging and context for why this was the best choice for the fix are in #3225.

## Release notes

n/a

#### General

- [n/a] Created tests which fail without the change (if possible)

#### User visible changes

- [n/a] Manually tested at 200% size
- [n/a] Manually evaluated tab order

#### New Features

- [n/a] Add new FeatureFlag env vars to `server/conf/application.conf`
- [n/a] Conditioned new functionality on a [FeatureFlag](https://docs.civiform.us/contributor-guide/developer-guide/feature-flags)
- [n/a] Wrote browser tests with the feature flag off and on, etc.

### DIFF
--- a/browser-test/fonts.conf
+++ b/browser-test/fonts.conf
@@ -1,0 +1,12 @@
+<?xml version='1.0'?>
+<!DOCTYPE fontconfig SYSTEM 'fonts.dtd'>
+<!--
+  This is a partial solution to https://github.com/civiform/civiform/issues/3225.
+  It forces fc-match to pick a font that contains a range of bold styles
+-->
+<fontconfig>
+  <alias>
+    <family>sans-serif</family>
+    <prefer><family>Ubuntu</family></prefer>
+  </alias>
+</fontconfig>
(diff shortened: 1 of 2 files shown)


bion (approved):
Great job tracking this down and delivering a fix. It took quite a bit of tenacity!

dkatzz (approved):
Awesome job!! It looks like there may still be some snapshots to update, but otherwise LGTM!

Issue #4124, opened 2023-01-30
Program summary shows "Not eligible" before question answered


dkatzz:
**Describe the bug**
When eligibility is set on a question and the applicant navigates to the summary page. They'll see "Not eligible" before answering the question. The "not eligible" text should only appear if the question has been answered and the answer is not eligible. 

**To Reproduce**

1. Go to admin page and create a program with a date question that has eligibility criteria
2. Login as applicant and apply to program
3. See not eligible text next to date question

**Expected behavior**
"Not eligible" text should only appear after question is answered and applicant isn't eligible

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #4159, opened 2023-02-03
 Show not eligible banner on pre-view page

### LINKED ISSUE #4141: Show not eligible banner on pre-view page
**Description**
On the application preview and review page show a "Not Eligible" banner at the top of the page when there are questions the applicant has answered that don't meet the eligibility criteria of the program.

[Mocks](https://docs.google.com/presentation/d/1sNk1JSumJl8VbxKfukCkugIIm2zVDeNFWjJJuSWvflQ/edit#slide=id.g1b79a1aa5c0_0_33)

**Additional context**
This is part of the eligibility work (#3744)

**Done when**
Banner appears at the top of preview and review pages when there are any questions where the user is "not eligible"

dkatzz:
### Description

Show the not eligible toast when the application is not eligible, even if the application is in progress. 

This also makes toasts more accessible, by adding role and aria-live attributes.

See mocks [here](https://docs.google.com/presentation/d/1sNk1JSumJl8VbxKfukCkugIIm2zVDeNFWjJJuSWvflQ/edit#slide=id.g1b79a1aa5c0_0_33)

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #4141


server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -85,6 +85,9 @@ public interface ReadOnlyApplicantProgramService {
   /** Returns if the Program eligibility criteria are met. */
   boolean isApplicationEligible();
 
+  /** Returns if any of the answered questions in the program are not eligible. */
+  boolean isInProgressApplicationNotEligible();
```
bion, line 89: suggestion (blocking): Just reading the javadoc on this method and comparing to the method above on `isApplicationEligible`, I might guess that this method is just `!isApplicationEligible()`. Please add more detail to the javadoc above and to this one so it's clear how they implement different behavior beyond just inverting the same condition and that it is possible and expected for there to be cases when both return `false`.

bion, line 89: suggestion (non-blocking): consider renaming to `isApplicationNotEligible()`

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -98,6 +98,12 @@ public boolean isApplicationEligible() {
     return getAllActiveBlocks().stream().allMatch(block -> isBlockEligible(block.getId()));
   }
 
+  @Override
+  public boolean isInProgressApplicationNotEligible() {
+    return getSummaryData().stream()
+        .anyMatch(question -> !question.isEligible() && question.isAnswered());
```
bion, line 104: question (blocking): why does this use the summary data whereas `isApplicationEligible` uses the active blocks? Could this not also use the active blocks?

server/app/assets/javascripts/toast.ts
```diff
@@ -51,6 +51,8 @@ export class ToastController {
     const toastMessage = document.createElement('div')
     toastMessage.setAttribute('id', message.id)
     toastMessage.setAttribute('ignorable', String(message.canIgnore))
+    toastMessage.setAttribute('role', 'alert')
+    toastMessage.setAttribute('aria-live', 'polite')
```
bion, line 55: praise: thanks for catching this!

bion (changes requested):

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
dkatzz: I originally went with active blocks, but with that it won't show the not eligible message toast if the applicant has two questions in a block and one of them is answered and not eligible and the other hasn't been answered. It would only show the not eligible message if all of the questions have been answered and any are not eligible. The reason we have to check for isAnswered is because of #4124. By using summary data we can check each question on its own, so we don't have to worry about having some questions on the block answered and some unanswered. Does that make sense? Is there a way of doing this I might be missing?

server/app/services/applicant/ReadOnlyApplicantProgramService.java
dkatzz: Updated - let me know if it is more clear. Did a test and saw that applicationIsEligible returns false when a summary page is loaded where the eligibility question hasn't been answered yet and isApplicationNotEligible will return false in that case too.

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
dkatzz: That's a good point, thanks! Updated.

bion (approved):
Big improvement with adding `isQuestionEligibleInBlock`!

PR #4160, opened 2023-02-03
Store address geo correction status and location fields on applicant

### LINKED ISSUE #4043: Store address geo correction status and location fields on applicant
CF needs the ability to store the coordinates and the wellKnownId (used for spatial reference) for a corrected address on the applicant.

services/applicant/question/AddressQuestion.java and views/questiontypes/AddressQuestionRenderer.java should be updated to include fields for:
- corrected
- longitude
- latitude
- wellKnownId

New scalars should be added to services/applicant/question/Scalar.java accommodate this:

- ADDRESS_CORRECTED( "corrected", ScalarType.STRING)
- ADDRESS_LONGITUDE("longitude",ScalarType.DOUBLE)
- ADDRESS_LATITUDE("latitude",ScalarType.DOUBLE)
- ADDRESS_WELLKNOWNID("wellKnownId",ScalarType.LONG)

And these should be added to ADDRESS_SCALARS in the same file.

Double is a new scalar type that should be added to services/question/types/ScalarType.java

services/applicant/ApplicantService.java stageNormalUpdates has a switch case that should be updated to handle DOUBLE. There is already an applicantData.putDouble you can call.

See the [storing](https://docs.google.com/document/d/1Hm-LIra58dGPP1iAj6Vci62_4wZg6wl-kkR8ERYfMxY/edit#heading=h.mcve4bqfxroa) and [reading](https://docs.google.com/document/d/1Hm-LIra58dGPP1iAj6Vci62_4wZg6wl-kkR8ERYfMxY/edit#heading=h.hrk7b0vf5wi1) sections for more details.

gwendolyngoetz:
### Description

Store address geo correction status and location fields on applicant.

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Partial work for #4043


server/app/services/applicant/question/Scalar.java
```diff
@@ -34,6 +37,7 @@ public enum Scalar {
   STATE("state", ScalarType.STRING),
   STREET("street", ScalarType.STRING),
   TEXT("text", ScalarType.STRING),
+  WELLKNOWNID("wellknownid", ScalarType.LONG),
```
bion, line 40: **nitpick:** 
```suggestion
  WELL_KNOWN_ID("well known id", ScalarType.LONG),
```

server/app/services/applicant/question/Scalar.java
```diff
@@ -18,14 +18,17 @@
  */
 public enum Scalar {
   CITY("city", ScalarType.STRING),
+  CORRECTED("corrected", ScalarType.STRING),
```
bion, line 21: **suggestion (blocking):** this looks like it should be a `ScalarType.BOOLEAN`

server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -109,6 +110,30 @@ protected DivTag renderInputTags(
...
+            .withType("hidden")
+            .withName(addressQuestion.getLongitudePath().toString())
+            .withValue(addressQuestion.getLongitudeValue().orElse(0.0).toString());
+
+    InputTag wellKnownIdField =
+        j2html.TagCreator.input()
+            .withType("hidden")
+            .withName(addressQuestion.getWellKnownIdPath().toString())
+            .withValue(addressQuestion.getWellKnownIdValue().orElse(0L).toString());
+
```
bion, line 136: **suggestion (blocking):** If we're going to take the approach of setting these values by storing them in the DOM, we need to protect against a malicious request where the user has altered the values. One way to do that could be to use [`CryptographicUtils.sign`](https://github.com/civiform/civiform/blob/main/server/app/services/CryptographicUtils.java#L12) to generated an HMAC for either each of these values or all of them combined and validate it on the server when the form is submitted.

The other option is to store these values server-side instead of passing them through the DOM. The HMAC approach seems likely to be less work.

bion (changes requested):

server/app/views/questiontypes/AddressQuestionRenderer.java
gwendolyngoetz: @bion Would much rather to it server side. I haven't done anything with storing session data before. Can you point me at an existing example in the app to follow?

server/app/services/applicant/question/Scalar.java
gwendolyngoetz: @seanmalbert Are there any other states you think this could be? Otherwise I'll switch this to a bool.

server/app/views/questiontypes/AddressQuestionRenderer.java
bion: https://www.playframework.com/documentation/2.8.x/JavaSessionFlash

Please check that the session is encrypted automatically by Play. If it isn't, please use an HMAC to ensure data integrity before saving the values to the DB.

server/app/services/applicant/question/Scalar.java
seanmalbert: @gwendolyngoetz and @bion, 
It may be useful to capture a failed state like we do with service area validation, so CF would know if address correction was attempted but the external service returned a non 200 status. If this fails, we cannot perform service area validation for an address.

server/app/views/questiontypes/AddressQuestionRenderer.java
gwendolyngoetz: I'm removing the view change part from this PR. These changes will solely cover the internals of allowing it to be saved. The view changes will be done in a separate PR.

bion (approved):

Issue #4048, opened 2023-01-12
Store address service area validation on applicant


seanmalbert:
The result of an address' service area validation needs to be stored on the applicant.

To do this, a new scalar should be added to accommodate this. One possibility is that ADDRESS_SERVICE_AREAS can be a list of strings represented like ["seattle_valid", "tucson_invalid"] or ["seattle_1", "tucson_0"].

See [the TDD for more details](https://docs.google.com/document/d/1Hm-LIra58dGPP1iAj6Vci62_4wZg6wl-kkR8ERYfMxY/edit#heading=h.p7udt32obj7r)

PR #4161, opened 2023-02-03
Adds getServiceAreaInclusionGroup and isServiceAreaOptionInInclusionGroup to EsriClient

seanmalbert:
### Description

Adds two new methods, getServiceAreaInclusionGroup and isServiceAreaOptionInInclusionGroup  to EsriClient and removes isAddressLocationInServiceArea.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Adds method and class needed for #4048


server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,36 +269,91 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
+              if (optionMap.isEmpty()) {
+                logger.error(
+                    "Error calling EsriClient.getServiceAreaInclusionGroups. Error:"
+                        + " EsriServiceAreaValidationConfig.getImmutableMap() returned empty.");
+                esriServiceAreaInclusionBuilder
+                    .setArea(esriServiceAreaValidationOption.getId())
+                    .setState(EsriServiceAreaState.FAILED)
+                    .setTimeStamp(Instant.now());
+                inclusionListBuilder.add(esriServiceAreaInclusionBuilder.build());
+                return inclusionListBuilder.build();
```
bion, line 312: **suggestion (blocking):** this should happen before the call to ESRI

server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,36 +269,91 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
                         + " EsriServiceAreaValidationOption = {}. AddressLocation = {}",
                     esriServiceAreaValidationOption,
                     location);
-                return Optional.empty();
+                inclusionListBuilder.add(
+                    esriServiceAreaInclusionBuilder
+                        .setArea(esriServiceAreaValidationOption.getId())
+                        .setState(EsriServiceAreaState.FAILED)
+                        .setTimeStamp(Instant.now())
+                        .build());
```
bion, line 296: **question:** should we store a failure entry for each option in the config?

server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,36 +269,91 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
                   ctx.read(
                       "features[*].attributes." + esriServiceAreaValidationOption.getAttribute());
-              Optional<String> feature =
-                  features.stream()
-                      .filter(val -> esriServiceAreaValidationOption.getId().equals(val))
-                      .findFirst();
-              return Optional.of(feature.isPresent());
+
+              for (EsriServiceAreaValidationOption option : optionMap.get().values()) {
+                if (features.contains(option.getId())) {
```
bion, line 324: **suggestion (blocking):** it's a bit confusing that each option has its own `attribute` and `url` when the code here and in the call to ESRI assumes that the attribute and URl will be the same for all of them.

Please restructure the code or add comments explaining the assumptions we make around this.

bion (changes requested):

server/app/services/geo/esri/EsriClient.java
seanmalbert: That makes sense to me.

server/app/services/geo/esri/EsriClient.java
seanmalbert: Thanks for catching this. This is updated so that it adds the service areas that have the same URL specified in the config to the inclusion group.

This does mean that CF will have to call this method again if a service area validation predicate is selected that hasn't been checked against (e.i. the service area ID isn't in the inclusion group).

The other option would be to fetch service area results for each unique URL specified. However, in most cases it seems the URL will be the same and we'll have to accommodate adding new service area options to the config and allow for subsequent calls to fetch previously unchecked service areas anyway.

server/app/services/geo/esri/EsriClient.java
seanmalbert: Moved to before the call to fetchServiceAreaFeatures.

server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,36 +269,91 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
+   * @return boolean.
+   */
+  public Boolean isServiceAreaOptionInInclusionGroup(
+      EsriServiceAreaValidationOption esriServiceAreaValidationOption,
+      ImmutableList<EsriServiceAreaInclusion> inclusionGroup) {
+    return inclusionGroup.stream()
+        .map(EsriServiceAreaInclusion::getArea)
+        .collect(ImmutableList.toImmutableList())
+        .contains(esriServiceAreaValidationOption.getId());
+  }
```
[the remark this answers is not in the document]
seanmalbert: Nice suggestion. Moved.

server/app/services/geo/esri/EsriServiceAreaInclusion.java
```diff
@@ -0,0 +1,32 @@
+package services.geo.esri;
+
+import com.google.auto.value.AutoValue;
+import java.time.Instant;
+
+/** Represents a service area and its inclusion state for an address location. */
```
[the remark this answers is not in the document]
seanmalbert: committed suggestion

server/app/services/geo/esri/EsriServiceAreaInclusion.java
```diff
@@ -0,0 +1,32 @@
...
+@AutoValue
+public abstract class EsriServiceAreaInclusion {
+  public static Builder builder() {
+    return new AutoValue_EsriServiceAreaInclusion.Builder();
+  }
+
+  /** returns the service area that was checked for inclusion. */
+  public abstract String getArea();
```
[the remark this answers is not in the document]
seanmalbert: Updated get/setArea to get/setServiceAreaId

server/app/services/geo/esri/EsriServiceAreaInclusion.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /** returns the inclusion state of the service area. */
+  public abstract EsriServiceAreaState getState();
+
+  /** return the timestamp for when the inclusion check was made. */
+  public abstract Instant getTimeStamp();
```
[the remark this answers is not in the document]
seanmalbert: Makes sense. Updated.

server/app/services/geo/esri/EsriServiceAreaState.java
```diff
@@ -0,0 +1,8 @@
+package services.geo.esri;
+
+/** Defines the states of inclusion for a service area */
```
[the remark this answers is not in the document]
seanmalbert: Added comment explaining the state.

server/app/services/geo/esri/EsriServiceAreaState.java
```diff
@@ -0,0 +1,8 @@
...
+public enum EsriServiceAreaState {
+  INAREA,
+  NOTINAREA,
+  FAILED
```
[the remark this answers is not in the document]
seanmalbert: Updated to use SCREAMING_SNAKE_CASE

server/app/services/geo/esri/EsriServiceAreaValidationConfig.java
```diff
@@ -123,7 +124,13 @@ public Optional<ImmutableMap<String, EsriServiceAreaValidationOption>> getImmuta
...
       mapBuilder.put(values.get(i), option);
     }
-    return Optional.of(mapBuilder.build());
+
+    Optional<ImmutableMap<String, EsriServiceAreaValidationOption>> optionalMap =
+        Optional.of(mapBuilder.build());
+
+    esriServiceAreaValidationMap = optionalMap;
+
+    return optionalMap;
```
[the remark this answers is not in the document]
seanmalbert: Updated.

server/app/services/geo/esri/EsriServiceAreaState.java
```diff
@@ -0,0 +1,11 @@
...
+
+/** Defines the states of inclusion for an address in a service area. */
+public enum EsriServiceAreaState {
+  IN_AREA,
+  NOT_IN_AREA,
+  // FAILED indicates that the service area validation check failed.
+  // This can happen if the service area valiation option config does not load properly,
+  // or if the external Esri service returns a non 200 status
+  FAILED
+}
```
bion, line 11: **suggestion:** since the serialization format is critical to the feature working correctly let's make it explicit. Also a nit is that service area validation is not necessarily coupled to ESRI if we integrate with some other GIS system.

```suggestion
/** The result of checking an address for membership in a service area. */
public enum ServiceAreaState {
  // The address is in the service area.
  IN_AREA("InArea"),
  // The address is not in the service area.
  NOT_IN_AREA("NotInArea"),
  // The check failed for technical reasons.
  FAILED("Failed");

  private final String serializationFormat;

  ServiceAreaState(String serializationFormat) {
    this.serializationFormat = checkNotNull(serializationFormat);
  }

  /** The string representation of this state for storage in the database. */
  public String getSerializationFormat() {
    return this.serializationFormat;
  }
}
```

bion (changes requested):
About to go into a meeting but here is a partial review

server/app/services/geo/esri/EsriClient.java
bion: Looks like it's still storing just the one

server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,13 +270,41 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
+              + " EsriServiceAreaValidationConfig.getImmutableMap() returned empty.");
+      return supplyAsync(
+          () -> {
+            esriServiceAreaInclusionBuilder
+                .setServiceAreaId(esriServiceAreaValidationOption.getId())
+                .setState(EsriServiceAreaState.FAILED)
+                .setTimeStamp(Instant.now());
+            inclusionListBuilder.add(esriServiceAreaInclusionBuilder.build());
+            return inclusionListBuilder.build();
+          });
```
bion, line 298: **suggestion (blocking):** I'm not seeing a reason to run this in a separate threadpool
```suggestion
      inclusionListBuilder.add(esriServiceAreaInclusionBuilder
                .setServiceAreaId(esriServiceAreaValidationOption.getId())
                .setState(EsriServiceAreaState.FAILED)
                .setTimeStamp(Instant.now())
                .build());
      return CompletableFuture.completedFuture(inclusionListBuilder.build());
```

bion: In another comment thread we discussed returning `FAILED` for all options as well

server/app/services/geo/esri/EsriClient.java
```diff
@@ -262,13 +270,41 @@ CompletionStage<Optional<JsonNode>> fetchServiceAreaFeatures(
...
+    }
+
+    // Create a list of options with the same URL as the passed in EsriServiceAreaValidationOption.
+    // In most cases the same URL should provide all of the service areas.
+    ImmutableList<EsriServiceAreaValidationOption> optionList =
+        maybeOptionMap.get().values().stream()
+            .filter(option -> option.getUrl().equals(esriServiceAreaValidationOption.getUrl()))
+            .collect(ImmutableList.toImmutableList());
```
bion, line 306: **suggestion (blocking):** recommend moving this to `EsriServiceAreaValidationConfig`

bion (changes requested):

server/app/services/geo/esri/EsriServiceAreaState.java
seanmalbert: Good call. Moved this and ServiceAreaInclusion up one level out of /esri.

server/app/services/geo/esri/EsriClient.java
seanmalbert: Moved. Very open to suggestions on the new method name `mapToListWithSameServiceAreaOptionUrl`.

server/app/services/geo/esri/EsriClient.java
seanmalbert: That is done if the map isn't empty and the call to `fetchServiceAreaFeatures` is made and returns empty. Here `maybeOptionMap` would be empty so we can't get at the other options.

bion: ahh, thanks

server/app/services/geo/esri/EsriServiceAreaValidationConfig.java
```diff
@@ -132,12 +132,41 @@ public Optional<ImmutableMap<String, EsriServiceAreaValidationOption>> getImmuta
...
-        .filter(option -> option.getUrl().equals(serviceAreaOption.getUrl()))
-        .collect(ImmutableList.toImmutableList());
+  public Optional<ImmutableList<EsriServiceAreaValidationOption>> getOptionsWithSharedBackend(
+      String serviceAreaUrl) {
+    if (!isConfigurationValid()) {
+      logger.error(
+          "Error calling EsriServiceAreaValidationConfig.getOptionsWithSharedBackend. Error: Esri"
+              + " Address Service Area Config is missing settings.");
+      return Optional.empty();
+    }
```
bion, line 142: **suggestion (blocking)** the code that consumes this calls `get()` immediately so will throw a `NullPointerException` if `Optional.empty()` is returned. Recommend throwing an exception here instead and returning `ImmutableList<EsriServiceAreaValidationOption>`.

In general would prefer to avoid using `Optional` for indicating errors in favor of throwing checked exceptions.

bion (changes requested):

server/app/services/geo/esri/EsriServiceAreaValidationConfig.java
seanmalbert: That's true, though the code above where this is consumed first does a check with `!esriServiceAreaValidationConfig.isConfigurationValid()` and returns empty if not valid, so it's technically guarded. That said, working with all of these optionals is a pain and it would be nicer to throw an exception when `EsriServiceAreaValidationConfig` is initialized and the config is not valid.

seanmalbert: Added `InvalidEsriServiceAreaValidationConfigException` and updated `getOptionsWithSharedBackend` and `getImmutableMap` to throw if the config is invalid.

bion: Ah right on, sorry I missed the guard there. Thanks for the change!

bion (approved):

PR #4170, opened 2023-02-07
Only allow one address question per block to enable address correction

nb1701:
This prevents the user from enabling address correction for question blocks that contain more than one address question. It should be disallowed by the UI, but in case it gets through, it is also disallowed by the controller and returns a 406 if this is encountered.

## Release notes

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/controllers/admin/AdminProgramBlockQuestionsController.java
```diff
@@ -20,6 +20,7 @@
 import services.program.CantAddQuestionToBlockException;
 import services.program.IllegalPredicateOrderingException;
 import services.program.InvalidQuestionPositionException;
+import services.program.ProgramBlockDefinitionContainsInvalidQuestionException;
```
nb1701, line 23: This is a verrrrrry long name.  Open to suggestions to make this more concise.  Could be used in the future for other invalid questions in a program block.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -788,6 +789,18 @@ public ProgramDefinition setProgramQuestionDefinitionAddressCorrectionEnabled(
           "Unexpected error: updating a non address question with address correction enabled");
     }
 
+    List<ProgramQuestionDefinition> pqdWithAddressCorrectionEnabled =
```
nb1701, line 792: There's perhaps a more efficient way to do this?

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
seanmalbert: It is informative, but could be `ProgramQuestionDefinitionInvalidException`.

server/app/services/program/ProgramServiceImpl.java
seanmalbert: Suggestion: 
```
if (blockDefinition.programQuestionDefinitions().stream()
        .filter(
            question ->
                question.addressCorrectionEnabled() && question.id() != questionDefinitionId)
        .findAny()
        .isPresent()) {
      throw new ProgramBlockDefinitionContainsInvalidQuestionException(
          programId, blockDefinitionId, questionDefinitionId);
    }
```

Or use what you have in ProgramBlockEditView. Recommend creating a method in BlockDefinition, similar to `hasAddress` to make the check reusable.

seanmalbert (changes requested):

server/app/services/program/ProgramServiceImpl.java
nb1701: Actually, I don't think I actually need to collect it to a list anymore.  I think i can use the same logic as in the other place.  This was an experiment.

nb1701: Sorry, didn't reload the page and see your comment.  I'll take a look.

server/app/controllers/admin/AdminProgramBlockQuestionsController.java
nb1701: I like that better 👍

server/app/services/program/BlockDefinition.java
```diff
@@ -128,6 +128,19 @@ public boolean hasAddress() {
         .anyMatch(questionType -> questionType.equals(QuestionType.ADDRESS));
   }
 
+  /**
+   * Check if block already contains a question with address correction enabled, ignoring
+   * the given question ID that you are trying to enable it on.
+   */
+  @JsonIgnore
+  public boolean hasAddressCorrectionEnabledOnDifferentQuestion(long questionDefinitionId) {
```
nb1701, line 136: This is, once again, quite wordy.  Open to suggestions to making it more concise.

server/app/services/program/ProgramServiceImpl.java
nb1701: Great idea, moved it into BlockDefinition.

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -788,6 +788,11 @@ public ProgramDefinition setProgramQuestionDefinitionAddressCorrectionEnabled(
           "Unexpected error: updating a non address question with address correction enabled");
     }
 
+    if (blockDefinition.hasAddressCorrectionEnabledOnDifferentQuestion(questionDefinitionId)) {
+      throw new ProgramQuestionDefinitionInvalidException(
```
nb1701, line 792: @bion I noticed there was a recent changed that uses `BadRequestException` for something like this.  Do you think I should just use that here rather than create a new exception?

bion: If there's a chance that the exceptional situation should be handled some other way than returning a 400 then go ahead and make a specific checked exception type for it, otherwise feel free to throw the `BadRequestException`.

A good thing to know about with error handling is this thing: https://github.com/civiform/civiform/blob/main/server/app/controllers/ErrorHandler.java

nb1701: I forgot I'm catching this specifically in AdminProgramBlockQuestionsController so maybe on second thought, I'll leave it.

nb1701:
Added a browser test and a unit test for the BlockDefinition method I added.  Could possibly use unit tests for the ProgramServiceImpl and AdminProgramBlockQuestionsController to ensure it is throwing the right exception?  Not sure how important that is.

server/app/services/program/BlockDefinition.java
bion: lgtm

bion (approved):

nb1701:
Added another test

seanmalbert (approved):

PR #4171, opened 2023-02-07
Adds CiviForm version to login page.

avrittrohwer:
### Description

Adds the CiviForm version to the login footer.  Feature is enabled by setting SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE=true in the civiform server environment variables.

#### General

#### User visible changes

#### New Features


bion:
@avrittrohwer this will look pretty ugly in prod with something like `CiviForm version: SNAPSHOT-k4jh3k-1231235` showing up on the page. Also we're already injecting this into a meta tag.

avrittrohwer:
@bion the plan is to followup with an additional PR in civiform/cloud-deploy-infra that sets CIVIFORM_IMAGE_TAG to the version in the civiform_config.sh file.  So if someone is deploying v1.13.0, both this new element and the meta tag will have v1.13.0 instead of the commit tag

browser-test/src/landing_page.test.ts
```diff
@@ -1,23 +1,65 @@
...
+    await validateScreenshot(ctx.page, screenshotName)
   })
 
   it('has no accessibility violations', async () => {
     await validateAccessibility(ctx.page)
   })
+}
+
+describe('the landing page', () => {
+  const ctx = createTestContext()
```
bion, line 30: **suggestion:** 
```suggestion
  const {page} = createTestContext()
```

avrittrohwer: Reverted the test

bion (approved):

Issue #4028, opened 2023-01-09
Address verification predicate logic


shanemc-goog:
Need to connect the new address verification feature with the predicate logic system.

PR #4172, opened 2023-02-07
Evaluate address service area predicates

bion:
### Description

Evaluates `LeafAddressServiceAreaExpressionNode`s using the JsonPath library.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4028

### DIFF
--- a/server/app/services/applicant/predicate/JsonPathPredicateGenerator.java
+++ b/server/app/services/applicant/predicate/JsonPathPredicateGenerator.java
@@ -11,6 +11,7 @@ import services.applicant.RepeatedEntity;
 import services.applicant.exception.InvalidPredicateException;
 import services.applicant.question.ApplicantQuestion;
 import services.applicant.question.Scalar;
+import services.geo.ServiceAreaState;
 import services.program.predicate.LeafAddressServiceAreaExpressionNode;
 import services.program.predicate.LeafExpressionNode;
 import services.program.predicate.LeafOperationExpressionNode;
@@ -62,6 +63,16 @@ public final class JsonPathPredicateGenerator {
    */
   public JsonPathPredicate fromLeafAddressServiceAreaNode(LeafAddressServiceAreaExpressionNode node)
       throws InvalidPredicateException {
+    if (!LeafAddressServiceAreaExpressionNode.SERVICE_AREA_ID_PATTERN
+        .matcher(node.serviceAreaId())
+        .matches()) {
+      throw new InvalidPredicateException(
+          String.format(
+              "Service area ID invalid for LeafAddressServiceAreaExpressionNode. question ID: %d,"
+                  + " service area ID: %s",
+              node.questionId(), node.serviceAreaId()));
+    }
+
     return JsonPathPredicate.create(
         String.format(
             "%s[?(@.%s %s %s)]",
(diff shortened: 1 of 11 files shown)


server/app/services/applicant/predicate/JsonPathPredicateGenerator.java
```diff
@@ -37,10 +41,39 @@ public JsonPathPredicateGenerator(
...
+      throws InvalidPredicateException {
+    return JsonPathPredicate.create(
+        String.format(
+            "%s[?(@.%s %s %s)]",
+            getPath(node).predicateFormat(),
+            Scalar.SERVICE_AREA.name().toLowerCase(),
+            Operator.IN_SERVICE_AREA.toJsonPathOperator(),
+            String.format(
+                "/([a-zA-Z\\-]+_[a-zA-Z]+_\\d+,)*%s_InArea_\\d+(,[a-zA-Z\\-]+_[a-zA-Z]+_\\d+)*/",
+                node.serviceAreaId())));
```
bion, line 73: I don't love this regex-based approach, but there are tradeoffs either way.

Pros:
- this code stays consistent for how it generally works between the two leaf node types
- this was really quick to implement!

Cons:
- regexes tend to be are write-only (illegible)
- does not differentiate between `NotInArea` and `Failed`

seanmalbert: If we add validation to the service area IDs, keeping the regex makes sense to me. And passing EsriServiceAreaState can make it more flexible with something like:
`String.format("/%1$s_(%2$s|%3$s)_(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)/", node.serviceAreaId, EsriServiceAreaState.IN_AREA, EsriServiceAreaState.FAILED)`


For reference, tested `([a-zA-Z]+)_(InArea|Failed)_(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+Z)` against:
- Seattle_InArea_2023-02-07T11:49:22.456Z
- Bellevue_Failed_2023-02-07T11:49:22.4556666Z

bion: Cool, and good call on checking for `Failed` as well.

For the timestamps I recommend we use an integer unix timestamp since that's what we're doing elsewhere in the codebase.

bion:
@seanmalbert ready for another round of review. I'll wait until your PR is in before merging so I can make sure everything is good with the shared pieces (e.g. `server/app/services/geo/ServiceAreaState.java`)

server/app/services/applicant/predicate/PredicateEvaluator.java
```diff
@@ -50,6 +54,27 @@ private boolean evaluateLeafNode(LeafOperationExpressionNode node) {
...
+          node,
+          e.getMessage());
+      return false;
+    }
+  }
+
+  /**
+   * Returns true if and only if the answer for the address question referenced by the {@link
+   * LeafAddressServiceAreaExpressionNode} has an in-area service area in {@link ApplicantData}.
+   */
```
seanmalbert, line 68:
```suggestion
  /**
   * Returns true if and only if the answer for the address question referenced by the {@link
   * LeafAddressServiceAreaExpressionNode} has an in-area or failed service area in {@link ApplicantData}.
   */
```

seanmalbert (approved):

Issue #3227, opened 2022-08-22
Program App List Display Error If Submit Time is Unset


gwendolyngoetz:
I've been seeing the error "Application ### submitted without submission time marked" getting logged. See [code](https://github.com/civiform/civiform/blob/ccf88d6a9284c8eafd06c3d825078727f7244551/server/app/views/admin/programs/ProgramApplicationListView.java#L281-L288)

This looks to be happening on the Admin Program Application List (/admin/programs/41814/applications) when an Application has this state:
* lifecycle_stage = 'obsolete'
* submit_time = null

We either need to filter the list to active applications only (which I believe is not the desired solution) or not log this as an error.

cassandralfernandes:
To confirm: Does this bug impact users, or is it just in logs?

gwendolyngoetz:
This would affect admin users

nb1701:
Needs some investigation as to how to submit_time is being set to null.  Is it because the application was never submitted? Or was the value cleared out at some point incorrectly?

bion:
@gwendolyngoetz it'd be helpful to know the submission time of the most recent application that is either `ACTIVE` or `OBSOLETE` that does not have a submission time.

PR #4179, opened 2023-02-08
Ensure applicant uses existing application

bion:
### Description

This PR addresses this bug: https://github.com/civiform/civiform/issues/3227

The program admin UI shows all non-draft, non-deleted applications for a given program ([query](https://github.com/civiform/civiform/blob/bionj-missing-submission-time-bug/server/app/repository/ProgramRepository.java#L224-L233)), meaning all `ACTIVE` and `OBSOLETE` applications are considered submitted. This is a reasonable assumption given submitting an application updates it from `DRAFT` to `ACTIVE` and marks all other applications to the same program as `OBSOLETE`, throwing an exception if more than one `DRAFT` is found for the program ([that happens here](https://github.com/civiform/civiform/blob/main/server/app/repository/ApplicationRepository.java#L86-L123)).

However, the logic to throw an exception if multiple drafts are found is more recent than the rest of the code, so it's possible multiple drafts were found in the past and marked `OBSOLETE` without setting a submission time. This PR sets the submission time on those and logs an error so we know it's happening if it continues to.

Also, when the applicant successfully (no validation errors) submits their first block update to a program, CiviForm creates a DRAFT application for them for that particular program version. It does this by checking if there is an existing draft yet ([that happens here](https://github.com/civiform/civiform/blob/main/server/app/repository/ApplicationRepository.java#L190-L209)). However the previous version of the code only checked for drafts to that particular version of the program, instead of all versions of the program. This PR fixes that.

This PR also adds more logging in general to the area around submitting applications in hopes we can better understand this bug if it is in fact an ongoing an issue and not an artifact of the older server code's behavior.

#### General

### Issue(s) this completes

Fixes [#3227](https://github.com/civiform/civiform/issues/3227)

### DIFF
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -3,6 +3,7 @@ package repository;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.ebean.DB;
@@ -32,7 +33,7 @@ public final class ApplicationRepository {
   private final UserRepository userRepository;
   private final Database database;
   private final DatabaseExecutionContext executionContext;
-  private static final Logger logger = LoggerFactory.getLogger(ApplicationRepository.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(ApplicationRepository.class);
 
   @Inject
   public ApplicationRepository(
@@ -58,11 +59,7 @@ public final class ApplicationRepository {
         .findEach(fn);
   }
 
-  /**
-   * Submit an application, which will delete any in-progress drafts, obsolete any submitted
-   * applications to a program with the same name (to include past versions of the same program),
-   * and create a new application in the active state.
-   */
+  @VisibleForTesting
   public CompletionStage<Application> submitApplication(
       Applicant applicant, Program program, Optional<String> tiSubmitterEmail) {
     return supplyAsync(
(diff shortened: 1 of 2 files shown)


server/app/repository/ApplicationRepository.java
```diff
@@ -58,18 +59,19 @@ public void forEachSubmittedApplication(Consumer<Application> fn) {
         .findEach(fn);
   }
 
-  /**
-   * Submit an application, which will delete any in-progress drafts, obsolete any submitted
-   * applications to a program with the same name (to include past versions of the same program),
-   * and create a new application in the active state.
-   */
+  @VisibleForTesting
```
gwendolyngoetz, line 62: Is this needed given submitApplication is public?

server/app/repository/ApplicationRepository.java
```diff
@@ -90,21 +92,28 @@ private Application submitApplicationInternal(
...
+
+      final Application application;
+      if (drafts.size() == 1) {
+        application = drafts.get(0);
+      } else if (drafts.isEmpty()) {
+        LOGGER.error(
+            "No DRAFT applications found when submitting for applicant {} program {}",
+            applicant.id,
+            program.id);
+        application = new Application(applicant, program, LifecycleStage.ACTIVE);
```
gwendolyngoetz, line 109: (NB) Since this is handling creating an application object I'd consider making the log level warning and not an error.

server/app/repository/ApplicationRepository.java
```diff
@@ -118,6 +127,15 @@ private Application submitApplicationInternal(
             || app.getLifecycleStage().equals(LifecycleStage.OBSOLETE)) {
           continue;
         }
+        LOGGER.error(
+            "Multiple applications found at submit time for applicant {} to program {} {}:"
+                + " application {}",
+            applicant.id,
+            program.id,
+            program.getProgramDefinition().adminName(),
+            app.id);
```
gwendolyngoetz, line 136: (NB) I'd also consider making this error a warning since it doesn't seem to create a hard stop for the applicant.

server/app/repository/ApplicationRepository.java
bion: The annotation indicates why it is public and doesn't influence access level

server/app/repository/ApplicationRepository.java
bion: done

server/app/repository/ApplicationRepository.java
bion: done

gwendolyngoetz (approved):

PR #4187, opened 2023-02-08
Add a new ProgramType enum to Programs.

reemaka:
### Description

This is part of the Common Intake Form work and will be used to differentiate the Common Intake Form from regular programs. Note that this is not gated on a feature flag yet because it just adds the enum; being able to set the value to COMMON_INTAKE_FORM will be gated by a feature flag when that capability is added.

#### General


server/conf/evolutions/default/47.sql
```diff
@@ -0,0 +1,8 @@
+# --- !Ups
+
+alter table programs add column program_type varchar;
+update programs set program_type = 'DEFAULT' where program_type is null;
```
bion, line 4: **question (blocking):** `'DEFAULT'` is all-caps here but the enum `dbValue` is lowercase. Is that intentional?

bion, line 4: **question (blocking):** why two statements rather than `alter table programs add column program_type varchar default 'DEFAULT';`?

bion (changes requested):

server/conf/evolutions/default/47.sql
reemaka: Oops nope, I just switched from all caps to lower case in the enum and forgot to make the switch here as well. thanks!

reemaka: switched to one statement.

bion (approved):

PR #4189, opened 2023-02-09
Fix multiple service area values bug

bion:
### Description

Fixes a bug in the multi-value predicate configure view (enabled by `PREDICATES_MULTIPLE_QUESTIONS_ENABLED` feature flag). Address service area values are specified using a `select` element, whereas all other scalar values use an `input`. Since the JS code that creates new value rows only looks for `input` elements, it didn't update the `name` attribute of the `select` for address service areas.

#### General

### DIFF
--- a/server/app/assets/javascripts/admin_predicate_configuration.ts
+++ b/server/app/assets/javascripts/admin_predicate_configuration.ts
@@ -250,6 +250,20 @@ class AdminPredicateConfiguration {
       el.closest('label')?.setAttribute('for', newId)
     })
 
+    newRow.querySelectorAll('select').forEach((el: HTMLSelectElement) => {
+      const groupNumString = assertNotNull(el.name.match(/group-(\d+)/))[1]
+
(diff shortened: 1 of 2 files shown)


seanmalbert (approved):

server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -250,6 +250,20 @@ class AdminPredicateConfiguration {
       el.closest('label')?.setAttribute('for', newId)
     })
 
+    newRow.querySelectorAll('select').forEach((el: HTMLSelectElement) => {
+      const groupNumString = assertNotNull(el.name.match(/group-(\d+)/))[1]
```
seanmalbert, line 254: Suggestion (non-blocking): there is logic above this for inputs that is identical, consider creating a shared function.

bion: Done

PR #4197, opened 2023-02-10
Add delete screen confirmation modal

### LINKED ISSUE #3766: Add confirmation prompt when deleting a screen
We should consider adding a confirmation prompt when a program admin clicks the 'delete screen' button on the edit program page. Since there is no undo option they could lose a lot of work if the button is accidentally clicked.

I don't see the same need for the delete button for individual questions since those can be re-added easily from the question bank.

rockycodes:
### Description

Added a modal that prompts the user to confirm they want to delete the screen they are on. The modal text changes depending on what content the screen contains (questions, eligibility conditions, and/or visibility conditions).

#### General

#### User visible changes

- [ ] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method (I tried to do this but the Modal component exists outside of the page landmarks which is not something I can control in this PR)
- [x] Manually evaluated tab order (I did this but it's not ideal. May want to revisit this in future work.)

### Issue(s) this completes

Fixes #3766

https://user-images.githubusercontent.com/24319951/217995870-115864fe-9b51-465e-8ed5-0a663485ad14.mov


server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -878,6 +881,82 @@ private DivTag questionBankPanel(
...
+      // print the appropriate message.
+      String listItemsInBlock = "";
+      if (itemsInBlock.size() == 1) {
+        listItemsInBlock = itemsInBlock.get(0);
+      } else if (itemsInBlock.size() == 2) {
+        listItemsInBlock = itemsInBlock.get(0) + " and " + itemsInBlock.get(1);
+      } else {
+        // there will only ever be 1, 2, or 3 items
+        listItemsInBlock =
+            itemsInBlock.get(0) + ", " + itemsInBlock.get(1) + " and " + itemsInBlock.get(2);
```
bion, line 932: **thought:** this makes me miss rails 😆 
https://apidock.com/rails/ActiveSupport/CoreExtensions/Array/Conversions/to_sentence

bion (approved):
Looks great!

server/app/views/admin/programs/ProgramBlockEditView.java
rockycodes: 😱 wow this would be soooo nice

Issue #3899, opened 2022-12-08
Durable job scheduler


bion:
See [design doc](https://docs.google.com/document/d/1pSKbJXrQz-qtq-ZNkS4x1Zbwa8Y3pkRU6pBDkA-lv7s/edit)

- [x] persistence layer
- [x] job registry
- [x] job runner
- [x] recurring job scheduling
- [x] error notifications

PR #4210, opened 2023-02-14
Durable job runner

bion:
### Description

See [design doc](https://docs.google.com/document/d/1pSKbJXrQz-qtq-ZNkS4x1Zbwa8Y3pkRU6pBDkA-lv7s/edit#heading=h.fknq0a9xnthm) for context. This PR implements the durable job runner.

I will hold off on adding to the docs repo until the system is ready for use.

#### General

### Issue(s) this completes

Contributes to https://github.com/civiform/civiform/issues/3899


server/app/durablejobs/DurableJobRunner.java
```diff
@@ -0,0 +1,178 @@
...
+      ZoneId zoneId) {
+    this.durableJobExecutionContext = Preconditions.checkNotNull(durableJobExecutionContext);
+    this.durableJobRegistry = Preconditions.checkNotNull(durableJobRegistry);
+    this.jobTimeoutMinutes = config.getInt("durable_jobs.job_timeout_minutes");
+    this.persistedDurableJobRepository = Preconditions.checkNotNull(persistedDurableJobRepository);
+    this.runnerLifespanSeconds = config.getInt("durable_jobs.poll_interval_seconds");
+    this.nowProvider = Preconditions.checkNotNull(nowProvider);
+    this.zoneOffset = zoneId.getRules().getOffset(nowProvider.get());
+  }
+
```
dkatzz, line 63: **suggestion:** Add javadoc for this method and maybe others?

server/app/repository/PersistedDurableJobRepository.java
```diff
@@ -1,17 +1,52 @@
...
 
 /** Implements queries related to {@link PersistedDurableJob}. */
 public final class PersistedDurableJobRepository {
 
   private final Database database;
+  private final Provider<LocalDateTime> nowProvider;
 
-  public PersistedDurableJobRepository() {
+  @Inject
+  public PersistedDurableJobRepository(
```
dkatzz, line 21: **suggestion:** Add javadoc comment here too?

server/app/repository/PersistedDurableJobRepository.java
```diff
@@ -1,17 +1,52 @@
...
+      @BindingAnnotations.Now Provider<LocalDateTime> nowProvider) {
     this.database = DB.getDefault();
+    this.nowProvider = Preconditions.checkNotNull(nowProvider);
+  }
+
+  /**
+   * Gets a job that is ready to be executed or empty if none are available.
+   *
+   * <p>A job is ready to be executed if:
```
dkatzz, line 30: **nitpick:** change to "if it"?

server/test/durablejobs/DurableJobRunnerTest.java
```diff
@@ -0,0 +1,167 @@
...
+            instanceOf(PersistedDurableJobRepository.class),
+            () ->
+                instanceOf(
+                    new BindingKey<>(LocalDateTime.class)
+                        .qualifiedWith(BindingAnnotations.Now.class)),
+            instanceOf(ZoneId.class));
+  }
+
+  @Test
+  public void runJobs_timesOut() {
```
dkatzz, line 50: **suggestion:** Add another test where the timeout isn't 0?

server/app/durablejobs/DurableJobRunner.java
```diff
@@ -0,0 +1,178 @@
...
+              e.getMessage(),
+              ExceptionUtils.getStackTrace(e));
+      LOGGER.error(msg);
+      persistedDurableJob.appendErrorMessage(msg).save();
+    }
+  }
+
+  private LocalDateTime resolveStopTime() {
+    // We set poll interval to 0 in test
+    if (runnerLifespanSeconds == 0) {
```
nb1701, line 152: For this and the `jobTimeoutMinutes` below, would it make sense to set the timeout in tests, rather than putting this in the code to cover tests?  I just wonder if we should throw an exception if the value given is <= 0, rather than effectively making 0 == 5.  Or perhaps disable the timeout entirely if it's 0?

server/app/modules/DurableJobModule.java
```diff
@@ -17,6 +25,36 @@
...
+        ExecutionContext executionContext,
+        Provider<DurableJobRunner> durableJobRunnerProvider) {
+      int pollIntervalSeconds = config.getInt("durable_jobs.poll_interval_seconds");
+
+      actorSystem
+          .scheduler()
+          .scheduleAtFixedRate(
+              // Wait a random amount of time to decrease likelihood of synchronized
+              // polling with another server.
+              /* initialDelay= */ Duration.ofSeconds(new Random().nextInt(/* bound= */ 30)),
```
nb1701, line 51: Do we have an idea about how long we expect jobs utilizing this system to take?  I wonder if for some jobs, waiting up to 30 seconds might be overkill and cause a normally short-running job to take much longer.

server/test/durablejobs/DurableJobRunnerTest.java
bion: The timeout is 0 specifically for test mode so the times can be much shorter

server/app/modules/DurableJobModule.java
bion: This is the delay from when the server starts to when it starts polling the database for jobs to run.

server/app/durablejobs/DurableJobRunner.java
bion: It's useful to test that the timeout works, so I'd rather not disable timeout entirely.

The tricky bit is that we have to specify chrono units with the time, making passing much shorter times in for testing awkward

server/app/durablejobs/DurableJobRunner.java
bion: Done

server/app/repository/PersistedDurableJobRepository.java
bion: We don't usually document c'tors, was there something specific here you'd like documented?

server/app/repository/PersistedDurableJobRepository.java
bion: Done

server/app/repository/PersistedDurableJobRepository.java
dkatzz: Oops yes makes sense!

dkatzz (approved):

server/app/durablejobs/DurableJobRunner.java
bion: I'd also prefer not to have the special case in the code but I'm not seeing a way to do it that does all of:
- gives us a short timeout for testing
- doesn't require passing in an awkward number to the server config e.g. `600` for specifying a 10 minute timeout

nb1701 (approved):

PR #4195, opened 2023-02-10
Add env-var-docs implementation 

avrittrohwer:
### Description

Implementation of https://docs.google.com/document/d/1jYNeMqa9qK40JhPrtvNehzl-ooBeqQlnbZif2HvhuKw/edit#.

### Checklist

- [ ] complete env-var-docs.json
- [x] add env-var-docs-checker action on PRs
- [ ] run generate_markdown.py in release action
- [x] make sure python tests, formatter, and mypy are run on PRs editing python files
- [x] support updating existing file in generate_markdown.py

#### General

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [x] Created tests which fail without the change (if possible)
- [x] Extended the README / documentation, if necessary

PR #4225, opened 2023-02-16
env-var-docs: add PR action and env_var_docs helper package

avrittrohwer:
### Description

Adds a parsing-package helper package that supports validating and processing environment variable documentation files. 
 Also adds a github action that runs python type checking and unit tests for the env-var-docs directory.  

This PR is broken off of https://github.com/civiform/civiform/pull/4195/files#diff-7ba49bdd2226c2f4528596d2a18ca32e84c56f34e5e5d7891533c6e02047b7c5 to make review easier.

parser.py is made a package so that it can be installed by python code in the cloud-deploy-infra repository using pip: https://pip.pypa.io/en/stable/topics/vcs-support/#git.  The structure of parsing-package was influenced by https://packaging.python.org/en/latest/tutorials/packaging-projects/ and https://docs.pytest.org/en/latest/explanation/goodpractices.html#test-package-name.

#### General


avrittrohwer (commented):
BLOCKING: add unit tests for validator.py and visitor.py

Done.

env-var-docs/README.md
```diff
@@ -0,0 +1,91 @@
...
+```sh
+$ pip install ./env_var_docs
+$ pip install -r requirements.txt
+```
+
+To exit the virtual environment, run:
+
+```sh
+$ deactivate
+```
```
bion, line 73: **suggestion (blocking):** instead of having a human read through and execute these steps please write a bin script that takes care of this

env-var-docs/env_var_docs/schema/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The env-var-docs.json file contains key-value pairs. Each key is either a group
+of environment variables or an environment variable referenced in
+application.conf.
+
+Groups have the following fields:
+
+- `group-description`: A human-readable description of what the group
+  configures.
+- `members`: An object containing groups or variables.
```
bion, line 14: **question (blocking):** thoughts on renaming `members` -> `variables`?

env-var-docs/env_var_docs/schema/README.md
```diff
@@ -0,0 +1,97 @@
...
+The `group-description` and `members` fields must be set for each group.
+
+Variables have the following fields:
+
+- `description`: A human-readable description of what the environment variable
+  configures.
+- `type`: The value type expected. Can be 'string', 'int', 'bool', or
+  'index-list'.
+- `required`: If the environment variable is required to be set. If the
+  `required` field is not set, the environment variable is not required.
```
bion, line 25: **question (blocking):** is it just checking if it's set or can it be `"required": false`?

env-var-docs/env_var_docs/schema/README.md
```diff
@@ -0,0 +1,97 @@
...
+- `values`: If `type` is string, a list of valid strings can be provided. If
+  `values` is defined, `regex` and `regex_tests` can not be defined.
+- `regex`: If `type` is string, a regular expression using [python re
+  syntax](https://docs.python.org/3/library/re.html#regular-expression-syntax)
+  can be provided. This regular expression defines the set of valid values.
+  Because the regex is in a JSON string, any " or \ characters in it need to be
+  escaped like \" or \\. If `regex` is defined, `values` can not be defined.
+- `regex_tests`: A list of objects. Each object contains a `val` field with a
+  string value and a `shouldMatch` field with a boolean value. If `regex` is
+  defined, `regex_tests` must also be defined.
```
bion, line 35: **praise:** this is such a good idea

env-var-docs/env_var_docs/schema/README.md
```diff
@@ -0,0 +1,97 @@
...
+    "CLOUD_PROVIDER": {
+        "description": "What cloud services to connect to.",
+        "type": "string",
+        "values": ["aws", "azure", "gcp"]
+    }
+    "LANGUAGES": {
+        "description": "Supported languages.",
+        "type": "index-list"
+    }
+}
```
bion, line 85: **suggestion (blocking):** include an example of a group with members/variables in it

bion (changes requested):

env-var-docs/README.md
```diff
@@ -0,0 +1,91 @@
+# Environment Variable Documentation
+
+The main configuration file for the CiviForm server is in
+[application.conf](../server/conf/application.conf). Configuration values can
```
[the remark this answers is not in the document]
avrittrohwer: Done.

env-var-docs/README.md
```diff
@@ -0,0 +1,91 @@
...
+be read at server start time from environment variables. This is done using the
+`${?SOME_ENV_VAR}` substitution syntax.
+
+To aid in deploying the CiviForm server, we ensure every environment variable
+referenced in the configuration file has appropriate documentation. This
+documentation lives in [env-var-docs.json](../server/conf/env-var-docs.json).
+See [./env_var_docs/schema/README.md](./env_var_docs/schema/README.md) for
+documentation on the format. As part of a CiviForm release, the documentation
+is rendered as markdown and added to our
+[documentation](https://docs.civiform.us/it-manual/sre-playbook/server-environment-variables).
```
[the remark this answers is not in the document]
avrittrohwer: updated

env-var-docs/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+[check_vars_documentation.py](./check_vars_documented.py) and
+[run_regex_tests.py](./run_regex_tests.py): runs on any PR editing
+application.conf or env-var-docs.json. They ensure the documentation is
+complete and that any provided regular expression tests pass.
+
+[generate_markdown.py](./generate_markdown.py): runs during our release
+automation. Generates markdown from env-var-docs.json and submits it to our
+[documentation
+repository](https://github.com/civiform/docs/tree/main/docs/it-manual/sre-playbook/server-environment-variables).
```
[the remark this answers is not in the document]
avrittrohwer: Done.

env-var-docs/env_var_docs/schema/README.md
avrittrohwer: Because a group can contain another group, I think members is a better name otherwise 'variable' would have two meanings

env-var-docs/env_var_docs/schema/README.md
avrittrohwer: required can be set to false.  If the required key is not set, validation code should treat the variable as not required.

env-var-docs/env_var_docs/schema/README.md
avrittrohwer: Updated.

env-var-docs/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Developer setup
+
+Ensure you have a system installation of python version 3.9 or greater:
+
+```sh
+$ python3 --version
+```
```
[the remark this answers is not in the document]
avrittrohwer: It is only required for people who want to make changes to the python code.  I added "To contribute to any of the python files in this directory," to this section

env-var-docs/README.md
avrittrohwer: I am hesitant adding a layer of indirection for the venv stuff, in both cases someone would have to run the activate and deactivate commands.  I think not wrapping these steps with bin scripts would make them easier to understand

env-var-docs/env_var_docs/schema/README.md
bion: gotcha, sounds good

env-var-docs/README.md
bion: Could we compromise and add the bin script but leave the docs explaining how to do it without? I'm a big believer in speedy dev tooling

avrittrohwer: Added a script that checks python 3.10+ is installed and creates a venv directory if one does not exist.  The activate and deactivate commands must be called in the interactive shell and do not work if called in a script that the shell calls.  The new bin script echos out the commands to run to make it easier tho

bion: perfect, thank you!

env-var-docs/parser-package/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Developer Setup
+
+See [../README.md#developer-setup](../README.md#developer-setup).
+
+### Environment variable documentation file format
+
+An environment variable documentation file contains key-value pairs. Each key
+is either a group of environment variables or an environment variable
+referenced in application.conf.
```
bion, line 14: **nitpick:** 
```suggestion
An environment variable documentation file contains key-value pairs. Each key
references either a group of environment variables or an environment variable
referenced in application.conf.
```

env-var-docs/parser-package/README.md
```diff
@@ -0,0 +1,109 @@
...
+Here is an example .env file that meets the above documentation validation rules:
+
+```
+TITLE="The title"
+SOME_NUMBER=42
+LOGO_URL="https://my-cool-logo.png"
+CLOUD_PROVIDER="aws"
+LANGUAGES.0="en-US"
+LANGUAGES.1="en-PI"
+```
```
bion, line 109: **praise:** great doc page

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,522 @@
+"""Provides functionality for parsing and visiting an environment variable
+documentation file.
+"""
```
bion, line 3: **suggestion (blocking):** mention that the parser implements validation as well

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,522 @@
...
+
+    group_errors: ParseErrors
+    """Errors from attempting to parse as a Group."""
+
+    variable_errors: ParseErrors
+    """Errors from attempting to parse as a Variable."""
+
+
+@dataclasses.dataclass
+class NodeInfo:
```
bion, line 64: **suggestion (non-blocking):** why not `Node` instead of `NodeInfo`?

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,522 @@
...
+    # Is this a variable?
+    var, var_errors = _parse_variable(path, value)
+    if len(var_errors) == 0:
+        assert var is not None  # appease mypy
+        nodes_to_visit.append(
+            NodeInfo(level=level, json_path=parent_path, name=key, details=var))
+        return
+
+    # details is neither a group nor a variable.
+    parsing_errors.append(NodeParseError(path, group_errors, var_errors))
```
bion, line 169: **thought (non-blocking):** looks like if a node is malformed we present the errors for it not being a var, ignoring that it could be a malformed group. That seems fine but want to double check that's the intention.

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,522 @@
...
+                obj=test)
+            errors.extend(errs)
+            _, errs = _parse_field(
+                parent_path=index_path,
+                key="should_match",
+                json_type=bool,
+                required=True,
+                obj=test)
+            errors.extend(errs)
+    return errors
```
bion, line 522: **praise:** great work on this

formatter/fmt
```diff
@@ -74,3 +74,6 @@ echo 'Start format shell'
 shfmt -bn -ci -i 2 -w -l \
   $(shfmt -f . | grep -v -e /node_modules)
 echo 'Done formatting shell'
+
+echo 'Start format env-var-docs python'
+yapf --verbose --style='{based_on_style: google, SPLIT_BEFORE_FIRST_ARGUMENT:true}' --in-place --recursive --exclude env-var-docs/venv env-var-docs
```
bion, line 79: **nitpick:** one line per flag please

bion (changes requested):
Some small things and some questions but overall this is looking great!

bin/env-var-docs-create-venv
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env bash
+
+# DOC: Checks that python 3.10+ is install and a venv directory exists in env-var-docs.
+
+source bin/lib.sh
+
+python_path="$(which python3)"
+
+if [[ "${python_path}" =~ "env-var-docs/venv/bin/python3" ]]; then
+  echo "ERROR: venv is currently activated, first run 'deactivate' in your shell then re-run this script."
```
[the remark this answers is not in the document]
avrittrohwer: That is correct, the venv/bin/activate script registers the deactivate function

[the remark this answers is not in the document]
bion: got it, thanks!

env-var-docs/parser-package/src/env_var_docs/parser.py
avrittrohwer: The `NodeParseError` type wraps the `ParseError` type so that callers of `visit` can separate the parse errors resulting from attempting to parse the object as a `Group` and the errors resulting from attempting to parse as a `Variable`.

The one place this abstraction breaks down is at the beginning of `visit` where it checks if the file is valid JSON and that it contains a single object.  In these cases I return a `NodeParseError` with the file-level error reported in the `group_errors` field.  This is not strictly correct because the file root is neither a `Group` nor a `Variable` but it seemed good enough rather than further complicating `NodeParseError` with a `file_errors` field.

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+- [check_vars_documentation.py](./check_vars_documented.py): Ensures all
+  environment variables referenced in application.conf are defined in
+  env-var-docs.json and that env-var-docs.json does not document any variables
+  referenced in application.conf. Runs on any PR changing application.conf or
+  env-var-docs.json.
+
+- [run_regex_tests.py](./run_regex_tests.py): Ensures that all regular
+  expression validation rules in env-var-docs.json compile and that all
+  provided regualar expressions tests pass. Runs on any PR changing
```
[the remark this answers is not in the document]
avrittrohwer: Fixed

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
...
+documentation on the format. As part of a CiviForm release, the documentation
+is rendered as markdown and added to our [documentation
+website](https://docs.civiform.us/it-manual/sre-playbook/server-environment-variables).
+
+## GitHub actions automations
+
+- [check_vars_documentation.py](./check_vars_documented.py): Ensures all
+  environment variables referenced in application.conf are defined in
+  env-var-docs.json and that env-var-docs.json does not document any variables
+  referenced in application.conf. Runs on any PR changing application.conf or
```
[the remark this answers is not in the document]
avrittrohwer: Good catch, changed to 

Ensures all environment variables referenced in application.conf are documented in env-var-docs.json and that env-var-docs.json does not document any variables not referenced in application.conf (ensures a 1:1 mapping).

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```sh
+$ python3 --version
+```
+
+We use [virtual python
+environments](https://docs.python.org/3/library/venv.html) to manage python
+dependencies. To set up a virtual environment with the required dependencies,
+run `bin/env-var-docs-create-venv` from the repository root. To activate the
+virual environment in your shell, run `source env-var-docs/venv/bin/activate`.
```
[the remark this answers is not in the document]
avrittrohwer: Fixed

env-var-docs/parser-package/README.md
avrittrohwer: Changed

env-var-docs/parser-package/README.md
avrittrohwer: Thank you!

env-var-docs/parser-package/src/env_var_docs/parser.py
avrittrohwer: Updated to:

Provides a 'visit' function that parses an environment variable
documentation file and returns any parsing errors due to unexpected structure.
If no parsing errors are found, 'visit' runs a user-provided function on each
node in the file.

See parser-package/README.md for a description of the expected structure.

env-var-docs/parser-package/src/env_var_docs/parser.py
avrittrohwer: No reason.  Changed to Node.

env-var-docs/parser-package/src/env_var_docs/parser.py
avrittrohwer: Thank you!

formatter/fmt
avrittrohwer: Done.

avrittrohwer:
Note for reviewers on why I structured the parsing-package directory the way I did:  I was primarily influenced by https://docs.pytest.org/en/latest/explanation/goodpractices.html#test-package-name and https://packaging.python.org/en/latest/tutorials/packaging-projects/

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+    return ".".join(args)
+
+
+def _ensure_no_extra_fields(
+        parent_path: str, obj: UnparsedJSON,
+        valid_fields: list[str]) -> ParseErrors:
+    """Returns ParseErrors for any keys in obj that are not in valid_fields."""
+    errors = []
+    for key in list(obj):
+        if key not in valid_fields:
```
jessbumblebee, line 411: Is there any case when we also have to check that there are no duplicates or does json handle that internally?

env-var-docs/parser-package/tests/readme_test.py
```diff
@@ -0,0 +1,38 @@
+import io
+import os
+import env_var_docs.parser
+import unittest
+
+
+class TestReadmeExample(unittest.TestCase):
```
jessbumblebee, line 7: Considering that there isn't a direct match for the file name, could you add some comments here to explain what this is testing? Changing the name of the test method below could also help make it more clear.

jessbumblebee (commented):
Nice change. I learned a lot about Python and json parsing too.
Most comments are readability suggestsions and smaller nits.

env-var-docs/parser-package/src/env_var_docs/parser.py
bion: looks great

bin/env-var-docs-create-venv
```diff
@@ -0,0 +1,44 @@
+#! /usr/bin/env bash
+
+# DOC: Creates a python virtual environment for the env-var-docs python files
```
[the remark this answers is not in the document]
avrittrohwer: Is your first comment is missing some suggested text?

bin/env-var-docs-create-venv
```diff
@@ -0,0 +1,44 @@
...
+# DOC: and installs dependencies. Will replace an existing venv directory if
+# DOC called multiple times (a good way to get back to a known good state).
+
+source bin/lib.sh
+
+python_path="$(which python3)"
+
+if [[ "${python_path}" =~ "env-var-docs/venv/bin/python3" ]]; then
+  echo "ERROR: venv is currently activated, first run 'deactivate' in your shell then re-run this script."
```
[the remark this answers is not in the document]
avrittrohwer: 'deactivate' is a function or script added by the venv python module, so I don't control the naming of it

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
...
+environments](https://docs.python.org/3/library/venv.html) to manage
+dependencies. To set up a virtual environment with the required dependencies,
+run `bin/env-var-docs-create-venv` from the repository root. To activate the
+virtual environment in your shell, run `source env-var-docs/venv/bin/activate`.
+Re-running `bin/env-var-docs-create-venv` will delete and re-create the virtual
+environment. Doing so is a good way to get back to a known good state.
+
+### Tests
+
+We run [mypy](https://mypy-lang.org/) to check any [type
```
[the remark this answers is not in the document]
avrittrohwer: Type hints are a way to specify expected types of function parameters and return values.  Do you think I should elaborate more in the README over the linked documentation (https://docs.python.org/3/library/typing.html)

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+
+def _recursively_parse(
+        level: int, parent_path: str, key: str, value: UnparsedJSON,
+        nodes_to_visit: list[Node], parsing_errors: list[NodeParseError]):
+    """Recursively visits a node and its children, in preorder traversal order (visit root then children)."""
+    path = _path(parent_path, key)
+
+    # Is this a group?
+    group, group_errors = _parse_group(path, value)
+    if len(group_errors) == 0:
```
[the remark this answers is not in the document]
avrittrohwer: The environment variable documentation file will be parsed in GitHub actions whenever the application.conf or env-var-docs.json files are changed.  If the file has any parsing errors, the action will fail and the editor will need to push new commits to fix the issues.

To make this process as easy as possible for devs, I choose to implement a parser that attempts to parse the whole document even if errors are occurred.  This way the dev knows about all errors after the first action run.  If we raised an exception, the dev would only know about the first error in the file.

env-var-docs/parser-package/src/env_var_docs/parser.py
bion: Sounds good, thanks for the explanation

bion (approved):

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+
+
+def _recursively_parse(
+        level: int, parent_path: str, key: str, value: UnparsedJSON,
+        nodes_to_visit: list[Node], parsing_errors: list[NodeParseError]):
+    """Recursively visits a node and its children, in preorder traversal order (visit root then children)."""
+    path = _path(parent_path, key)
+
+    # Is this a group?
+    group, group_errors = _parse_group(path, value)
```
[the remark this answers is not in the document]
avrittrohwer: The readability hit it due to wanting to return very detailed parsing errors.  If parsing fails, I want it be be obvious what field is missing/has a typo in the name.  So, we attempt to parse as a Group and get any errors from doing so.  If there were no errors, it is a Group and we process it.  Otherwise, we try to parse as a Variable.  If there were no errors, we process it.  If there were errors, then we could not parse the object as a Group nor a Variable.  We return the parsing errors from both attempts.  This makes it easier for callers of visit to print out detailed error messages.

For example, if we tried to parse:

```
{
  "My group": {
    "group_description": "A group",
    "members": {
       "MY_VAR": {
        "description": "A var"
        "typo": "string"
      }
    }
  }
}
```

```
Error at 'file.my group.MY_VAR':
  Could not parse as Group:
    'group_description' is a required field.
    'members' is a required field.
  Could not parse as a Variable:
    'type' is a required field.
    'typo' is not a valid field, valid fields are [...]
```

Hopefully from this error message it is easy to tell that the 'typo' field needs to change to 'type'

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+
+
+@dataclasses.dataclass
+class Group:
+    group_description: str
+    members: dict[str, UnparsedJSON]
+
+
+@dataclasses.dataclass
+class RegexTest:
```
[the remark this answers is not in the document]
avrittrohwer: It is a test case of a Variable's regex field.  In the env-var-docs.json file, if someone adds a regex validation check, they also need to add regex_tests.  This makes it much less likely that regex typos or errors are submitted to env-var-docs.json.

I agree it could be a bit weird to see a Test class in a non-test python file but I think RegexTest is the most clear name for the struct

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
...
+system installation of python version 3.10 or greater:
+
+```sh
+$ python3 --version
+```
+
+We use [virtual python
+environments](https://docs.python.org/3/library/venv.html) to manage
+dependencies. To set up a virtual environment with the required dependencies,
+run `bin/env-var-docs-create-venv` from the repository root. To activate the
```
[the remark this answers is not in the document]
avrittrohwer: Because the python stdlib module is named venv (https://docs.python.org/3/library/venv.html) I think it is more appropriate to reuse venv in names rather than expand it to virtual_env.

I have changed the README to "We use virtual python environments (venv) ..." to help make this association clearer.

env-var-docs/parser-package/README.md
```diff
@@ -0,0 +1,109 @@
...
+- `values`: If `type` is string, a list of valid strings can be provided. If
+  `values` is defined, `regex` and `regex_tests` can not be defined.
+- `regex`: If `type` is string, a regular expression using [python re
+  syntax](https://docs.python.org/3/library/re.html#regular-expression-syntax)
+  can be provided. This regular expression defines the set of valid values.
+  Because the regex is in a JSON string, any " or \ characters in it need to be
+  escaped like \\" or \\\. If `regex` is defined, `values` can not be defined.
+  If `regex` is defined, `regex_tests` must also be defined.
+- `regex_tests`: A list of objects. Each object contains a `val` field with a
+  string value and a `should_match` field with a boolean value.
```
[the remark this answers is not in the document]
avrittrohwer: Added.

.github/workflows/pr_env_var_docs.yaml
```diff
@@ -0,0 +1,42 @@
+name: pr_env_var_docs
+
+# cancels in-progress jobs on this pull request
```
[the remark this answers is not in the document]
avrittrohwer: Changed.

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+"""
+
+import dataclasses
+import typing
+import json
+
+UnparsedJSON = dict[str, typing.Any]
+
+
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
avrittrohwer: Added.

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+            if len(errors) == 0:
+                value = obj[key]
+                if extract_fn is not None:
+                    value = extract_fn(obj)
+
+    return value, errors
+
+
+def _path(*args: str) -> str:
+    """Creates a path from segments. Segments can not be the empty string."""
```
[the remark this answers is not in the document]
avrittrohwer: Changed to:

Creates a JSON key path from keys.

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,524 @@
...
+
+    if key not in obj:
+        if required:
+            errors.append(
+                ParseError(parent_path, f"'{key}' is a required field"))
+        else:
+            value = default
+    else:
+        if not isinstance(obj[key], json_type):
+            # Printing json_type directly produces strings like "<class'str'>".
```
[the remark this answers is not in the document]
avrittrohwer: :)

env-var-docs/parser-package/src/env_var_docs/parser.py
avrittrohwer: I did not think about this.  If there are duplicate keys in an object, the json package uses the last-specified key: https://docs.python.org/3/library/json.html#repeated-names-within-an-object.

To make it easier for us to catch bugs due to this behavior, I added a object_pairs_hook parameter to json.load that raises an exception if it is passed an object with duplicate keys.

env-var-docs/parser-package/tests/readme_test.py
avrittrohwer: Added.

env-var-docs/README.md
```diff
@@ -0,0 +1,69 @@
+# Environment Variable Documentation
+
+The main configuration file for the CiviForm server is in
+[application.conf](../server/conf/application.conf). The file is in [HOCON
```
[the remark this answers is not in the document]
avrittrohwer: I agree some documentation on the layers is worth adding.  Because the env-var-docs is the inner-most layer, I don't think it should have specific details about civiform_config.sh.  I added the following paragraph:

The documentation includes support for adding value validation rules. Tools
that deploy the CiviForm server
([civiform/cloud-deploy-infra](https://github.com/civiform/cloud-deploy-infra),
for example) should use these rules to validate user-provided values before the
server is deployed.

env-var-docs/parser-package/src/env_var_docs/parser.py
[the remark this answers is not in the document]
jessbumblebee: Oops, sorry, I thought I had removed this comment because I realised the error collection as I read on. my bad.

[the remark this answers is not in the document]
avrittrohwer: Gotcha, no worries

env-var-docs/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+```sh
+$ python3 --version
+```
+
+We use [virtual python
+environments](https://docs.python.org/3/library/venv.html) (venv) to manage
+dependencies. A virtual environment isolates the available python packages to
+just those explicitly installed inside the virtual environment. Using a virtual
+environment decreases the chances that the env-var-docs python scripts rely on
```
jessbumblebee, line 54: What risk would relying on an undeclared dependency actually cause? 
The reason I ask is that we are adding complexity with venv and I'm trying to assess if it's worth the extra complexity.

env-var-docs/parser-package/src/env_var_docs/parser.py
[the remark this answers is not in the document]
jessbumblebee: PS: You could put those fields inside each of the classes that the list represents (group, variable, regex test etc)

env-var-docs/parser-package/src/env_var_docs/parser.py
```diff
@@ -0,0 +1,549 @@
...
+def _parse_group(parent_path: str,
+                 obj: UnparsedJSON) -> tuple[Group | None, ParseErrors]:
+    """Parses a Group from obj. If ParseErrors are encountered, None is returned."""
+
+    errors = []
+    errors.extend(
+        _ensure_no_extra_fields(
+            parent_path, obj, ["group_description", "members"]))
+
+    group_description, errs = _parse_field(
```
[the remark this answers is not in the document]
avrittrohwer: Because the schema of the env-var-docs.json file allows for Groups and Variables to be in the same place (top-level fields in the file as well as fields in a Group's members), there is no getting around attempting to parse as both a Variable and a Group.  _recursively_parse handles this by bubbling up ParseErrors only when both _parse_group and _parse_errors return errors

env-var-docs/parser-package/src/env_var_docs/parser.py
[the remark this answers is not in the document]
avrittrohwer: Added.

bin/env-var-docs-create-venv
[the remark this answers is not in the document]
avrittrohwer: Added.

env-var-docs/parser-package/src/env_var_docs/parser.py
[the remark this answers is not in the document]
avrittrohwer: Do you think it would be easier to understand if the ensure_no_extra_fields call occurred after all the expected fields were parsed?

env-var-docs/README.md
avrittrohwer: Relying on an undeclared dependency introduces developer pain and confusion when the tests and coverage works on their local machine but fail in the github action.  Using a venv makes it easier to get a consistent python environment across dev machines and github action runners.  From my research venvs are a python industry standard

jessbumblebee (approved):
From our sync: Feel free to apply the changes we discussed and submit so that you are unblocked. I'm happy to have another look at the comments etc later if you'd like me to.

PR #4226, opened 2023-02-16
Add nongated eligibility setting for programs.

reemaka:
### Description

Add a program-level setting for whether or not that program's eligibility criteria is gating. Also fixes a typo in the feature flag confs.

## Release notes

#### General

#### User visible changes

#### New Features


avaleske (changes requested):
LGTM!

Slight nit: I find `isEligibilityGating` kind of hard to parse, but that might just be me. What about something like `gatesEligibility`, or similar, if that still matches the dev standards?

reemaka:
Yeah struggled a bit with the best name for it. I think `gatesEligibility` might be confusing because it's unclear what the subject of "gates" is (i.e. what exactly gates eligibility?) but I could do `gatedEligibility` if that's clearer. Also thought of doing something like nongatingEligibility at first since by default it's gated, but then thought maybe it's confusing to have a boolean phrased negatively like that.

avaleske:
Mmm I agree that a boolean phrased negatively is also confusing.

Something more explicit like `programGatesEligibility` could work? or `programIsEligibilityGating` so that it's clear what `is` refers to? But then that's a little repetitive when using it like `if(snapBenefitsProgram.programIsEligibilityGating()) { ... }`

It definitely might be that `isEligibilityGating` is the least confusing of the options! But I wanted to see what else we had considered.

reemaka:
The thing with `programGatesEligibility` and `programIsEligibilityGating` is that the it's not exactly true that the program is gating eligibility. This field is saying that if this program has any eligibility questions set, _those questions_ block (or don't block) application submission. (Sorry to be so pedantic.)

I think approaching it the way you've described -- thinking about what's the clearest when accessing the field on a program -- is helpful! Maybe `program.eligibilityIsGating()` is clearer? Or abandoning the "gating" wording and using something like "blocking" or "enforced".

avaleske:
Oh I like something in the direction of `program.eligibilityIsBlocking()` or `program.eligibilityIsGating()`! That reads more naturally and makes it clear that it's a type of eligibility (gating vs non-gating) that is being selected.

reemaka:
Cool, switched to eligibilityIsGating

avaleske (approved):

PR #4243, opened 2023-02-17
ProgramDetails link in Applicant's Ineligible page should point to the program's external link

### LINKED ISSUE #4240: Program Details in Not Eligible Message not linked to program website
Program cards were updated to remove External Website reference and change Program Details to link to the program website configured in the program details. However, the message that applicants receive if they do not qualify includes a Program Details link  that still links to the description.  This needs to be updated to send the applicant to the Program Website.

swatkat1:
### Description

ProgramDetails link in Applicant's Ineligible page should point to the program's external link.

## Release notes

ProgramDetails link in Applicant's Ineligible page should point to the program's external link.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #4240

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramBlocksController.java
+++ b/server/app/controllers/applicant/ApplicantProgramBlocksController.java
@@ -38,6 +38,7 @@ import services.applicant.exception.ProgramBlockNotFoundException;
 import services.applicant.question.FileUploadQuestion;
 import services.cloud.StorageClient;
 import services.program.PathNotInBlockException;
+import services.program.ProgramDefinition;
 import services.program.ProgramNotFoundException;
 import services.program.ProgramService;
 import services.question.exceptions.UnsupportedScalarTypeException;
@@ -408,7 +409,7 @@ public final class ApplicantProgramBlocksController extends CiviFormController {
         && !roApplicantProgramService.isBlockEligible(blockId)) {
       CiviFormProfile submittingProfile = profileUtils.currentUserProfile(request).orElseThrow();
       try {
-        String externalLink = programService.getProgramDefinition(programId).externalLink();
+        ProgramDefinition programDefinition = programService.getProgramDefinition(programId);
 
         return supplyAsync(
             () ->
@@ -420,7 +421,7 @@ public final class ApplicantProgramBlocksController extends CiviFormController {
                         applicantName,
                         messagesApi.preferred(request),
                         applicantId,
-                        externalLink)));
+                        programDefinition)));
       } catch (ProgramNotFoundException e) {
         notFound(e.toString());
       }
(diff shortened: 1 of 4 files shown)


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -403,16 +407,23 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
...
-              ok(
-                  ineligibleBlockView.render(
-                      request,
-                      submittingProfile,
-                      roApplicantProgramService,
-                      applicantName,
-                      messagesApi.preferred(request),
-                      applicantId)));
+      try {
+        String externalLink = programService.getProgramDefinition(programId).externalLink();
```
dkatzz, line 411: **suggestion:** Shouldn't this only be this link if it's set and if not it should be the program details (here and below)?

dkatzz (commented):
Thanks for updating this! I think there should be a fall back if the program details link is unset though, right?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
swatkat1: That part is handled in the UI.. I didn't want to call ProgramService calls in UI hence it is done in controller.

server/app/views/applicant/IneligibleBlockView.java
```diff
@@ -38,14 +38,20 @@ public Content render(
...
-      long applicantId) {
+      long applicantId,
+      String externalLink) {
     long programId = roApplicantProgramService.getProgramId();
     boolean isTrustedIntermediary = submittingProfile.isTrustedIntermediary();
+    // Use external link if it is present else use the default Program details page
+    String programDetailsLink =
+        externalLink.isEmpty()
+            ? routes.ApplicantProgramsController.view(applicantId, programId).url()
+            : externalLink;
```
swatkat1, line 49: Here is the fall back if exterlink is empty

server/app/controllers/applicant/ApplicantProgramBlocksController.java
dkatzz, line 411: **suggestion:** Thoughts on passing the full ProgramDefinition here and below instead of the externalLink, so we're not passing a primitive and the view can benefit from having the program definition if needed for other data in the future?

server/app/views/applicant/IneligibleBlockView.java
dkatzz: Thanks!

server/app/controllers/applicant/ApplicantProgramBlocksController.java
swatkat1: I did think about it, but I stuck with less is more. I am ok with either approach.

swatkat1: @dkatzz  - I have changed per your feedback

dkatzz (approved):

server/app/controllers/applicant/ApplicantProgramBlocksController.java
dkatzz: Thanks!

dkatzz:
We can add the ignore for release label as well, since this is behind a feature flag. Sorry for not catching before.

Issue #4019, opened 2023-01-06
Cleanups for ProgramBlockView and functionality


jessbumblebee:
A selection of suggested improvements for the ProgramBlockViews and their functionality

1) Cleanup methods and names (they take more arguments than required and they don't use standard conventions (starting with a verb)
2) Rename ProgramBlockAllAspectsView and ProgramBlockView to ProgramBlockView and ProgramBlockBaseView

jessbumblebee:
Would also be good to improve the comments in the ProgramBlockViews 
Examples: specify better what each bit is rendering to make it easier to find
Ensure each class has good top level docs

Main place Shane and I noticed this in: ProgramBlockBaseView

jessbumblebee:
We could also consider improving tests by introducing BeforeAll and possibly several describes in one file https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/civiform_admin_program_statuses.test.ts?subtree=true

I should check with Bion first what specifics actually make a difference for performance though

jessbumblebee:
Also rename ProgramEditView to ProgramDetailsEditView because it is the view that gets shown when you click on "Edit Program details" and it is easy to confuse with ProgramBlockEditView.

jessbumblebee:
Other things that could be improved:
- There are various todos saying that "Block" should be renamed to Screen
- The classes that are named ProgramBlockXXXview are not limited to one ProgramBlock. From their functionality they are more of a ProgramXXXView. Renaming would improve readability, but would require a bit of research into what names are in use already

I'm using this bug as an umbrella bug for various ideas and observations, which could be split out later on.

jessbumblebee:
Suggestions for improving  the admin_program_viewing.test.ts:

The questions themselves in the block view have some edit controls e.g. the optional toggle and soon the address verification toggle. Need to account for those when rendering the read only version.

PR #4247, opened 2023-02-20
Add support for address verification and optional toggle in Program read only view

jessbumblebee:
### Description

Improve the read only version for programs and tests by 
- handling the new address verification properly
- handling the optional toggle better
- adding accessibility check  in tests

There are no mocks for this and I went with what I thought made sense. I'd appreciate a closer look at the screenshots by the reviewers to check if they aggree that having text labels here makes sense and that the text I chose represents the state well. For example I was depating if it should be "required question" or just "required etc. Feedback welcome.

Trackes as one of various improvement suggestions in #4019

## Release notes

N/A hidden behind flag, not user visible

#### General

- [ n/a ] Extended the README / documentation, if necessary

#### User visible changes

#### New Features

### Issue(s) this completes

n/a


server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -616,6 +616,19 @@ private DivTag renderQuestion(
               blockDefinition.id(),
               questionDefinition,
               canRemove));
+    } else {
+      // For each toggle, use a label instead in the read only view
+      if (maybeAddressCorrectionEnabledToggle.isPresent()) {
+        String label =
+            addressCorrectionEnabled
+                ? "Address correction: enabled"
```
dkatzz, line 624: **suggestion (non-blocking):** Optionally can pull this common part out and just use the ternary for the enabled / disabled piece.

dkatzz (approved):
Changes LGTM - thanks!

bion (approved):
The UI looks good to me. "Required question" is very clear which IMO is good.

server/app/views/admin/programs/ProgramBlockEditView.java
jessbumblebee: Done

Issue #3744, opened 2022-11-01
Tracking: Eligibility conditions for programs


shanemc-goog:
Tracking issue for adding Eligibility checking to programs.

The basic approach will be to mimic the Show/Hide system and re-use the Predicate logic backend flows and data storage.

This does not include the feature flag being removed (that is tracked [here](https://github.com/civiform/civiform/issues/4605)).

[Design Doc](https://docs.google.com/document/d/1A7ftdnAOghd2BCohNRtppyRpB4oKkik_0zAal4yUcuk/edit#heading=h.fvf1dku1q5pk) 

## Predicate logic system
 
- [x] Support and/or operations #3743 
- [x] Storage for Eligibility Predicates #3922
- [x] Feature Flag #3923

## Admin Flows
- [x] -- Evaluate eligibility questions #3926
- [x] -- Add/Remove eligibility #3932 
- [x] -- Restrict Eligibility available questions to the block #4025 
- [x] -- Show eligibility in Application view #4079

## Applicant Flows 
- [x] -- Evaluate eligibility upon submit #4080
- [x] -- Cleanup IneligibleBlockView #4083
- [x] -- Show status in preview/review page #4002 
- [x] -- Show not eligible banner on pre-view page #4141
- [x] -- Show status on program list for in-progress application #4142

## Trusted Intermediary Flows 
- [x] -- Ensure correct text shows for trusted intermediaries #4184

## Custom logic
- [x] Fix Currency Predicates #3854

## Other
- [x] Add translations #4127
- [x] #4315

bion:
@dkatzz can this issue be closed?

dkatzz:
> @dkatzz can this issue be closed?

The only open item is enabling the flag, but everything else has been completed. Happy to separate that out since its separate from the work?

PR #4261, opened 2023-02-21
Make applicant code async when showing the IneligiblePage from a submit exception

dkatzz:
### Description

Moved the call to get the ReadOnlyApplicantProgramService towards the beginning, so its available to us later on. 

I also realized I could make the call to get applicantName in the inEligibleBlockView instead of passing it, so I did that as well.

By doing this, we're better following the development standards to implement applicant code asynchronously - https://docs.civiform.us/contributor-guide/developer-guide/development-standards#async-request-handling

#### General

### Issue(s) this completes

#3744

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramReviewController.java
+++ b/server/app/controllers/applicant/ApplicantProgramReviewController.java
@@ -9,6 +9,7 @@ import com.google.common.collect.ImmutableList;
 import controllers.CiviFormController;
 import featureflags.FeatureFlags;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -169,7 +170,12 @@ private CompletionStage<Result> submitInternal(
...
             submittingProfile,
             featureFlags.isProgramEligibilityConditionsEnabled(request));
-    return submitApp
+    CompletableFuture<ReadOnlyApplicantProgramService> readOnlyApplicantProgramServiceFuture =
+        applicantService
+            .getReadOnlyApplicantProgramService(applicantId, programId)
+            .toCompletableFuture();
+    return readOnlyApplicantProgramServiceFuture
+        .thenComposeAsync(v -> submitApp.toCompletableFuture())
         .thenApplyAsync(
```
bion, line 179: **thought (non-blocking):** this does the right thing I think. AFAICT the tricky thing here is having multiple futures we want to resolve before continuing, another way to do that could be to use `CompleteableFuture.allOf`, which may make that intention a bit more clear (I always forget what `thenCompose` does other than it's usually not what I expect or want).

I don't know if using `allOf` is actually an improvement for readability though so up to you.

bion (approved):
@dkatzz please update the PR title to be more specific

dkatzz:
> @dkatzz please update the PR title to be more specific

Updated - let me know how that looks.

server/app/controllers/applicant/ApplicantProgramReviewController.java
dkatzz: Good call - its somewhat similar for readability I think. I used allOf instead of what I had before, since these don't depend on one another - let me know if you have a preference for the other method though.

bion: LGTM!

bion (approved):

PR #4273, opened 2023-02-23
Add settings page where eligibility can be set to non-gating

reemaka:
### Description

Adds a new program settings page where admins can control whether or not eligibility criteria is gating or non-gating for a specific program.

#### General

#### New Features


server/app/views/admin/programs/ProgramSettingsEditView.java
```diff
@@ -0,0 +1,109 @@
...
+                            .withClasses(
+                                "absolute",
+                                "bg-white",
+                                eligibilityIsGating ? "left-1" : "right-1",
+                                "right-1",
+                                "top-1",
+                                "w-4",
+                                "h-4",
+                                "rounded-full")))
+            .with(p("Non-gating e;ligibility").withClasses("hover-group:text-white", "ml-1"));
```
avaleske, line 83: Extra `;`

Also, can we check with Miriam about this copy? Something like `Eligibility criteria does not block submission` or `Allow applicants to submit applications even if eligibility criteria are not met` or `Allow submission of ineligible applications` fits our natural language standards better.

server/app/views/admin/programs/ProgramSettingsEditView.java
```diff
@@ -0,0 +1,109 @@
...
+import views.admin.AdminLayout;
+import views.admin.AdminLayout.NavPage;
+import views.admin.AdminLayoutFactory;
+import views.style.StyleUtils;
+
+/** Renders a page for editing program-level settings. */
+public final class ProgramSettingsEditView extends BaseHtmlView {
+  private static final String ELIGIBILITY_TOGGLE_ID = "eligibility-toggle";
+  private static final String ELIGIBILITY_IS_GATING_DESCRIPTION =
+      "When using eligibility conditions, you can set the experience to gating or non-gating. The"
```
avaleske, line 29: Same as below, but can we check with Miriam about how final this copy is? Something like:


> When enabled, applicants can submit applications even if the eligibility criteria are not met. When disabled, applications must meet all eligibility criteria in order to submit an application.


has slightly more natural language.

avaleske (approved):
LGTM! with a few nits and a question about the copy on the settings page.

server/app/views/admin/programs/ProgramSettingsEditView.java
reemaka: The copy isn't final since Miriam is still working on the copy doc; meant to mention that in the PR description but forgot, sorry. I like your suggestions though, and switched to using those for now. We can run it by Miriam later and update it if needed.

Issue #4174, opened 2023-02-07
Allow an admin to set a default application status for a program 


isabelcshaw:
**Is your feature request related to a problem? Please describe.**
This feature request aims to address the following problems: 
- Reduce the number of emails and calls to program admins asking about application status by improving communication (both content and speed) with applicants around the status of their applications 
- Reduce the time Program Administrators take to manually update application status - with the number of applications expected for Fresh Bucks, the time required will be not feasible 

[More detailed description can be found in the [**Application Status Management PRD**](https://docs.google.com/document/d/1EbWixfzrWdgDvCBBFW_Mw3KKm8JfXVNg6JlYKGrhMic/edit#)]

**Describe the solution you'd like**
- Using the existing application status feature, the admin can add a new status or select an existing status 
- The admin can set that status as a default 
- The admin can see which status is set to default when they look at their list of statuses 
- (A program is NOT required to have a default application status) 
- When a program has a “default status” added, all new applications to that program are automatically assigned the custom “default status” and the Program’s “custom email” is sent
- Programs with default statuses do not send out the standard “application received” email
- CiviForm applicants with accounts will see the “default status” on their programs view if they have applied (existing behavior) 

**Additional context**
[Detailed mocks](https://www.figma.com/file/4MirFQhWIUeeqkCWkQdjfk/CiviForm---Status-management-features---Public-file?node-id=404%3A27316)

This feature is needed in order to launch Fresh Bucks in Seattle.

seanmalbert:
@isabelcshaw ,

To clarify this point:
> Programs with default statuses do not send out the standard “application received” email

The email sent should use the `Email the applicant about the status change` text instead?

PR #4275, opened 2023-02-23
Add ability to set default status for programs, apply status to new applications, and i18n for emails

nb1701:
### Description

This adds the concept of a "default status" to a particular program.  When set, new applications to the program are automatically set to this status, and the corresponding email (if set) for that status is sent to the applicant rather than the default "application received" email.  

Because current statuses will not have the defaultStatus field in them, this is marked as Optional and is false if the field is not present.

Setting the default status on new applications is done via creating a new ApplicationEvent without a creator.  The creator parameter is now an Optional to account for this, which is then treated as a null in the database column.  We should only provide an empty creator for programmatic actions like this, and never on actions derived from user input.

This adds a bit of client-side javascript for confirming that a user wants to change or add a default status.  While this is different from the original mocks, it turned out that triggering a modal from another modal would have been pretty cumbersome, so this is the first pass at this functionality.

This also adds a helper script for checking any emails that have been sent by the dev environment to LocalStack, so we can verify what emails are actually being sent.

Adds i18n to email subjects and body boilerplate sent by ApplicantService and ProgramAdminApplicationService.

## Release notes

This adds the concept of a "default status" to a particular program.  When set, new applications to the program are automatically set to this status, and the corresponding email (if set) for that status is sent to the applicant rather than the default "application received" email.  

Additionally, emails sent upon application submission and status updates to an applicant and Trusted Intermediary are now localized.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

https://github.com/civiform/civiform/issues/4174


server/app/services/applicant/ApplicantService.java
```diff
@@ -320,6 +325,68 @@ public CompletionStage<Application> submitApplication(
...
+
+  /**
+   * Email the applicant with the status' defined email, if the applicant's email address exists and
+   * the status has an email associated with it.
+   *
+   * @param applicantId the ID of the applicant
+   * @param programDef the ProgramDefinition that the applicant applied for
+   * @param status the status from which to get the email body to send
+   */
+  private void maybeNotifyApplicantWithStatus(
```
nb1701, line 363: I don't love that I'm basically copying the code from ProgramAdminApplicationService here.  Maybe there's another place we can put this?

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -334,8 +348,38 @@ private Modal makeStatusUpdateModal(
...
+                        BaseStyles.CHECKBOX,
+                        "default-checkbox-" + formData.getConfiguredStatusText()),
+                span("Set as default status"));
+    Optional<StatusDefinitions.Status> currentDefaultStatus =
+        program.toProgram().getDefaultStatus();
+    String defaultStatusText =
+        currentDefaultStatus.isPresent() ? currentDefaultStatus.get().statusText() : "";
+    String messagePart =
+        currentDefaultStatus.isPresent() ? "from " + defaultStatusText + " to " : "to ";
+    // I'm so sorry
```
nb1701, line 369: Heh, I should probably remove this comment.  I really don't love this at all.  But it seemed like all the other options required twisting the code in knots to show a nice confirmation modal.  If anyone has ideas on how to do this better, please let me know!

server/app/views/components/Icons.java
```diff
@@ -39,6 +39,7 @@ public enum Icons {
           + " 6.375t-6.35 4.3Q28.15 44 24 44Zm0-3q7.1 0 12.05-4.975Q41 31.05 41"
           + " 24q0-7.1-4.95-12.05Q31.1 7 24 7q-7.05 0-12.025 4.95Q7 16.9 7 24q0 7.05 4.975"
           + " 12.025Q16.95 41 24 41Zm0-17Z"),
+  CHECK(960, "M378 810 154 586l43-43 181 181 384-384 43 43-427 427Z"),
```
nb1701, line 42: Need to double check why this would only show with 960 and not 48.

nb1701:
Still need to clean this up, modify a toast message on default status change, and add tests, but should mostly be ready for initial review.

nb1701:
Does this need a feature flag?

server/app/services/applicant/ApplicantService.java
```diff
@@ -320,6 +325,68 @@ public CompletionStage<Application> submitApplication(
...
+      long applicantId, ProgramDefinition programDef, StatusDefinitions.Status status) {
+    Optional<LocalizedStrings> emailBody = status.localizedEmailBodyText();
+    Optional<String> applicantEmail = getEmail(applicantId).toCompletableFuture().join();
+    if (emailBody.isPresent() && applicantEmail.isPresent()) {
+      Locale locale =
+          getPreferredLocale(applicantId)
+              .toCompletableFuture()
+              .join()
```
bion, line 371: **suggestion (blocking):** this is on an HTTP thread for the applicant so needs to be async

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -559,11 +559,16 @@ public ErrorAnd<ProgramDefinition, CiviFormError> appendStatus(
         .isPresent()) {
       throw new DuplicateStatusException(status.statusText());
     }
+    List<StatusDefinitions.Status> statusesCopy =
+        Lists.newArrayList(program.statusDefinitions().getStatuses());
+    if (status.defaultStatus().isPresent() && status.defaultStatus().get().equals(true)) {
```
bion, line 564: **nitpick:** 
```suggestion
    if (status.defaultStatus().orElse(false)) {
```

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -574,6 +579,26 @@ public ErrorAnd<ProgramDefinition, CiviFormError> appendStatus(
             .join());
   }
 
+  private List<StatusDefinitions.Status> unsetDefaultStatus(
+      List<StatusDefinitions.Status> statuses, Optional<String> exceptStatusName) {
+    return statuses.stream()
+        .<StatusDefinitions.Status>map(
+            status -> {
+              if (exceptStatusName.isPresent()
+                  && status.statusText().equals(exceptStatusName.get())) {
```
bion, line 588: **nitpick:** 
```suggestion
              if (exceptStatusName.map(name -> name.equals(status.statusText()).orElse(false)) {
```

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -600,7 +625,13 @@ public ErrorAnd<ProgramDefinition, CiviFormError> editStatus(
         && statusNameToIndex.containsKey(editedStatus.statusText())) {
       throw new DuplicateStatusException(editedStatus.statusText());
     }
+    if (editedStatus.defaultStatus().isPresent()
+        && editedStatus.defaultStatus().get().equals(true)) {
```
bion, line 629: **nitpick:** same as above, could use `map` to avoid the `&&`

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -371,8 +371,12 @@ private PTag programCardApplicationStatus(
       Locale preferredLocale, StatusDefinitions.Status status) {
     return p().withClasses("border", "rounded-lg", "px-2", "py-1", "mb-4", "bg-blue-100")
         .with(
-            span(status.localizedStatusText().getOrDefault(preferredLocale))
-                .withClasses("text-xs", "font-medium"));
+            Icons.svg(Icons.INFO)
+                .withClasses("inline-block")
+                .withStyle("width: 18px; height: 18px;"),
```
bion, line 376: **suggestion (blocking):** recommend using tailwind classes rather than inline styles. If you need values that aren't provided out of the box you can use https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values

server/app/views/admin/programs/ProgramStatusesView.java
```diff
@@ -334,8 +348,38 @@ private Modal makeStatusUpdateModal(
...
+    String confirmJs =
+        "var dontShow="
+            + isCurrentDefault.toString()
+            + ";var statusValue=document.querySelector('.input-status-text-"
+            + formData.getConfiguredStatusText()
+            + " .flex input').value;if(!dontShow && document.querySelector('.default-checkbox-"
+            + formData.getConfiguredStatusText()
+            + "').checked){return confirm(\"The default status will be updated "
+            + messagePart
+            + "\"+statusValue+\". Are you sure?\")}else{return true}";
```
bion, line 379: **suggestion (blocking):** I'd recommend storing whatever data you need from the server as data attributes on the form and then accessing them in JS.

server side:

```java
form()
  .withClasses("status-delete-form")
  .withData("statusText", formData.getConfiguredStatusText())
  .withData("dontShow", isCurrentDefault.toString())
  .withData("messagePart", messagePart)
  // etc
```

client side:

```typescript
import {addEventListenerToElements, assertNotNull} from './util'

addEventListenerToElements(".status-delete-form", "submit", confirmStatusDelete);

function confirmStatusDelete(event: Event) {
  const deleteForm = event.target as HTMLFormElement

  const statusText = assertNotNull(event.dataset.statusText)
  const dontShow = assertNotNull(event.dataset.dontShow) === "true"
  const messagePart = assertNotNull(event.dataset.messagePart)
  const checked = assertNotNull(document.querySelector(`.default-checkbox-${statusText}`) as HTMLInputElement).checked

  if (!dontShow && checked) {
    event.preventDefault();
    if (confirm(confirmationMessage)) {
      deleteForm.submit()
    }
  }
}
```

bion (changes requested):

server/app/services/applicant/ApplicantService.java
nb1701: Cool, I wasn't sure if it was too weird to call an admin service from the applicant service, but that'd be better than copying the code.

server/app/services/applicant/ApplicantService.java
```diff
@@ -320,6 +325,68 @@ public CompletionStage<Application> submitApplication(
...
+   * the status has an email associated with it.
+   *
+   * @param applicantId the ID of the applicant
+   * @param programDef the ProgramDefinition that the applicant applied for
+   * @param status the status from which to get the email body to send
+   */
+  private void maybeNotifyApplicantWithStatus(
+      long applicantId, ProgramDefinition programDef, StatusDefinitions.Status status) {
+    Optional<LocalizedStrings> emailBody = status.localizedEmailBodyText();
+    Optional<String> applicantEmail = getEmail(applicantId).toCompletableFuture().join();
```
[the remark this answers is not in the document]
nb1701: I think in the other place we are sending the application email (the regular one without a status applied), we do this synchronously.  Should we change it there too? https://github.com/civiform/civiform/blob/893c3c8aecf4f3e47b060211c22b73259c7b273a/server/app/services/applicant/ApplicantService.java#L454

server/app/views/admin/programs/ProgramStatusesView.java
nb1701: Thanks, this is exactly what I was looking for, much better!

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
bion: 🤦🏻‍♂️ 

Yes, the standard is for all I/O on the applicant side of things to be done off the HTTP threadpool, though there are places where it looks like we've deviated from that.

It'd be nice if we had a linter that could catch these but it'd be a bit of a pain to write.

server/app/services/program/ProgramServiceImpl.java
nb1701: `editedStatus.defaultStatus().orElse(false)` is maybe even shorter?

bion: LGTM

server/app/services/applicant/ApplicantService.java
nb1701: This is probably sufficiently different now that it is async to leave it as is.

server/app/views/applicant/ProgramIndexView.java
nb1701: This is what we're doing in other places like https://github.com/civiform/civiform/blob/7b89d2c75152a745fb9a63c075a3d7453df7b4c8/server/app/views/ViewUtils.java#L88-L89 and https://github.com/civiform/civiform/blob/7b89d2c75152a745fb9a63c075a3d7453df7b4c8/server/app/views/admin/programs/ProgramStatusesView.java#L267-L269.  If it seems okay, I can change those as well.

bion: bummer. well maybe we can figure out why at somepoint, feel free to ignore for now

nb1701: Actually, what we can do is add this to our tailwind.config.js
```
theme: {
  extend: {
    spacing: {
      4.5: '18px'
    }
  }
}
```
And then use `h-4.5` and `w-4.5`.  Works like a charm.

bion: nice!

server/app/assets/javascripts/admin_program_statuses_view.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  private confirmStatusDefaultChange(event: Event) {
+    const statusForm = event.target as HTMLFormElement
+
+    const statusText = assertNotNull(statusForm.dataset.statustext)
+    const dontShow = assertNotNull(statusForm.dataset.dontshow) === 'true'
+    const messagePart = assertNotNull(statusForm.dataset.messagepart)
+    const checked = assertNotNull(
+      document.querySelector(
+        `.default-checkbox-${statusText}`,
```
nb1701, line 26: Need to rethink the strategy I used here, since statuses with spaces are not going to work.

nb1701:
Still working on testing, but the rest of it should be complete, I think.

server/app/services/applicant/ApplicantService.java
```diff
@@ -382,6 +397,34 @@ public CompletionStage<Application> submitApplication(
...
+            .setEmailSent(true)
+            .build();
+    ApplicationEventDetails details =
+        ApplicationEventDetails.builder()
+            .setEventType(ApplicationEventDetails.Type.STATUS_CHANGE)
+            .setStatusEvent(statusEvent)
+            .build();
+    // Because we are doing this automatically, set the Account to empty.
+    return applicationEventRepository.insertAsync(
+        new ApplicationEvent(application, Optional.empty(), details));
```
bion, line 425: **nitpick:** 
```suggestion
        new ApplicationEvent(application, /* creator= */ Optional.empty(), details));
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -495,41 +564,124 @@ private void notifyTiSubmitter(
...
+    CompletableFuture<Optional<String>> emailFuture = getEmail(applicantId).toCompletableFuture();
+    CompletableFuture<Optional<Locale>> localeFuture =
+        getPreferredLocale(applicantId).toCompletableFuture();
+    return CompletableFuture.allOf(emailFuture, localeFuture)
+        .thenRunAsync(
+            () -> {
+              // Not blocking since these are already completed
+              Optional<String> email = emailFuture.join();
+              Locale locale = localeFuture.join().orElse(LocalizedStrings.DEFAULT_LOCALE);
+              if (!email.isEmpty()) {
```
bion, line 611: **nitpick:** invert the condition and return early instead to avoid indenting the code currently in the `if` body

bion (approved):
Overall looks good! A bunch of small things mostly idiosyncratic to our style guide. Since I'm going to be out I'm handing off the follow up reviews to @dkatzz.

server/app/services/applicant/ApplicantService.java
```diff
@@ -398,13 +441,39 @@ CompletionStage<Application> submitApplication(
...
+                            .toCompletableFuture());
+              } else {
+                maybeUpdateStatusAndEmailApplicantFuture =
+                    maybeNotifyApplicant(applicantId, application.id, programName)
+                        .toCompletableFuture();
+              }
+              return CompletableFuture.allOf(
+                      maybeUpdateStatusAndEmailApplicantFuture,
+                      maybeNotifyTiSubmitterFuture,
+                      updateStoredFileAclsForSubmit(applicantId, programId).toCompletableFuture())
```
bion, line 476: @dkatzz you were pretty successful in wrangling a bunch of java futures into a better state recently so I'm curious if you have any ideas for organizing the async work in this file a bit better. No worries if not, there's only so much that can be done given what the language provides us to work this.

server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -357,10 +368,7 @@ private DivTag renderApplicationListItem(
             .condWith(
                 displayStatus,
                 p().withClasses("text-sm", "text-gray-700")
-                    .with(
-                        span("Status: "),
-                        span(application.getLatestStatus().orElse("None"))
-                            .withClass("font-semibold")))
+                    .with(span("Status: "), span(statusString).withClass("font-semibold")))
```
dkatzz, line 371: **suggestion:** I changed the formatting for eligibility status to truncate the tag after the word as well as show green in the case that they may qualify. I think these statuses should follow a similar pattern (with green for when approved)?https://github.com/civiform/civiform/pull/4283/files

server/app/services/applicant/ApplicantService.java
```diff
@@ -398,13 +441,39 @@ CompletionStage<Application> submitApplication(
               String programName = application.getProgram().getProgramDefinition().adminName();
               notifyProgramAdmins(applicantId, programId, application.id, programName);
 
+              CompletableFuture<Void> maybeNotifyTiSubmitterFuture;
```
dkatzz, line 444: **suggestion:** Maybe you could do something like this here and below to clean it up a bit:

```
CompletableFuture<Void> maybeNotifyTiSubmitterFuture = tiSubmitterEmail.isPresent() ? maybeNotifyTiSubmitter(
                            tiSubmitterEmail.get(), applicantId, application.id, programName)
                        .toCompletableFuture() : CompletableFuture.completedFuture(null);
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -398,13 +441,39 @@ CompletionStage<Application> submitApplication(
...
+                  applicationProgram.getDefaultStatus();
+              CompletableFuture<Void> maybeUpdateStatusAndEmailApplicantFuture;
+              if (maybeDefaultStatus.isPresent()) {
+                StatusDefinitions.Status defaultStatus = maybeDefaultStatus.get();
+                maybeUpdateStatusAndEmailApplicantFuture =
+                    CompletableFuture.allOf(
+                        setApplicationStatus(application, defaultStatus).toCompletableFuture(),
+                        maybeNotifyApplicantWithStatus(
+                                applicantId,
+                                application.getProgram().getProgramDefinition(),
```
dkatzz, line 465: **suggestion:** Can swap for applicationProgram.getProgramDefinition(), since application.getProgram() is defined in that variable above.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -86,9 +88,27 @@ public Result createOrUpdate(Http.Request request, long programId)
...
       return ok(statusesView.render(request, program, Optional.of(form)));
     }
+    // Because we need to look directly at the value sent (or absent) in the form data
+    // itself to determine if a checkbox is checked, we need to calculate this here
+    // before passing the value into createOrEditStatusFromFormData. If the checkbox
+    // is not checked, no value is sent. If it is checked, "on" is sent for the value.
+    Optional<String> maybeSetDefault =
+        Optional.ofNullable(form.rawData().get(ProgramStatusesForm.DEFAULT_CHECKBOX_NAME));
+    final boolean setDefault;
+    if (!maybeSetDefault.isPresent() || maybeSetDefault.get().isBlank()) {
```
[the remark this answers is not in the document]
nb1701: At this point, we want to see what the state of the checkbox is.  `form.get().getDefaultStatus()` will still return `false` when we've got the checkbox set, since we haven't actually updated the status to make it default yet.  This is determining if we should be setting the status default to true when we call `createOrEditStatusFromFormData`. 

That said, I've got a bit of tunnel vision on this now.  I can't remember exactly why `defaultStatus` on the form isn't set to the state of the checkbox on submission.  Something about not being able to tell a checkbox input to set it's value on form submission I think.

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -103,11 +123,17 @@ public Result createOrUpdate(Http.Request request, long programId)
     }
     boolean isUpdate =
         previousStatusCount == mutateResult.getResult().statusDefinitions().getStatuses().size();
-    return redirect(indexUrl).flashing("success", isUpdate ? "Status updated" : "Status created");
+    String toastMessage = isUpdate ? "Status updated" : "Status created";
+    if (setDefault
+        && previousDefaultStatus.isPresent()
+        && !previousDefaultStatus.get().statusText().equals(form.get().getConfiguredStatusText())) {
```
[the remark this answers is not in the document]
nb1701: Changing this a little bit, because it complains about `form.get().getConfiguredStatusText()` not being final when used inside a lambda, and because I think I actually want it to show the "updated" message when you are setting a default when there was no default previously.

```
    final String configuredStatusText = form.get().getConfiguredStatusText();
    if (setDefault && previousDefaultStatus.map(
        status -> 
          !status.statusText().equals(configuredStatusText)).orElse(true)) {
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -586,6 +738,39 @@ public CompletionStage<Optional<String>> getEmail(long applicantId) {
...
+  public CompletionStage<Optional<Locale>> getPreferredLocale(long applicantId) {
+    return userRepository
+        .lookupApplicant(applicantId)
+        .thenApplyAsync(
+            applicant -> {
+              if (applicant.isEmpty()) {
+                return Optional.empty();
+              }
+              return Optional.of(applicant.get().getApplicantData().preferredLocale());
+            },
```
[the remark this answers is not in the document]
nb1701: Not sure if we can use `flatMap` because `preferredLocale` returns `Locale`, not `Optional<Locale>`?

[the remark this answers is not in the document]
nb1701: @dkatzz What do you think?

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
nb1701: @dkatzz Does that look right to you?

server/app/repository/UserRepository.java
```diff
@@ -63,6 +64,16 @@ public Optional<Account> lookupAccountByEmail(String emailAddress) {
     return database.find(Account.class).where().eq("email_address", emailAddress).findOneOrEmpty();
   }
 
+  public CompletionStage<Optional<Account>> lookupAccountByEmailAsync(String emailAddress) {
+    if (emailAddress == null || emailAddress.isEmpty()) {
+      return CompletableFuture.completedStage(Optional.empty());
```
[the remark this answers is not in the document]
nb1701: I going off of what the synchronous version of this method was doing, and the database allows for a null string, so I'm wondering if we have a code path that allows for `Account` to have an empty email?

server/app/views/admin/programs/ProgramApplicationListView.java
nb1701: Oh right!  I meant to change that after you made that change and completely forgot.  Thanks for the reminder!

server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -331,14 +334,22 @@ private Modal renderDownloadApplicationsModal(
...
     String applicantNameWithApplicationId =
         String.format(
             "%s (%d)",
             applicantUtils.getApplicantNameEnUs(application.getApplicantData().getApplicantName()),
             application.id);
     String viewLinkText = "View →";
 
+    String statusString = application.getLatestStatus().orElse("None");
+    statusString +=
+        defaultStatus.isPresent() && statusString.equals(defaultStatus.get().statusText())
```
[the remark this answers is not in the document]
nb1701: Definitely like the `matches` idea, but it doesn't like that `statusString` is non-final and used inside a lambda.  I also realized this may do something funky if you actually have a status named "None", which is perhaps not outside the realm of possibility.

This is not terribly pretty, but I think it ends up being more correct.
```
Optional<String> appStatusString = application.getLatestStatus();
    String statusString;
    if (appStatusString.isPresent()) {
      statusString =
          appStatusString.get()
              + (defaultStatus
                      .map(
                          defaultStatusString -> defaultStatusString.matches(appStatusString.get()))
                      .orElse(false)
                  ? " (default)"
                  : "");
    } else {
      statusString = "None";
    }
```

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -575,6 +580,25 @@ public ErrorAnd<ProgramDefinition, CiviFormError> appendStatus(
             .join());
   }
 
+  private List<StatusDefinitions.Status> unsetDefaultStatus(
+      List<StatusDefinitions.Status> statuses, Optional<String> exceptStatusName) {
+    return statuses.stream()
+        .<StatusDefinitions.Status>map(
+            status -> {
+              if (exceptStatusName.map(name -> name.equals(status.statusText())).orElse(false)) {
```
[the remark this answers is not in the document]
nb1701: Cleaned this up even a little more:
```java
private ImmutableList<StatusDefinitions.Status> unsetDefaultStatus(
      List<StatusDefinitions.Status> statuses, Optional<String> exceptStatusName) {
    return statuses.stream()
        .<StatusDefinitions.Status>map(
            status ->
                exceptStatusName.map(name -> status.matches(name)).orElse(false)
                    ? status
                    : status.toBuilder().setDefaultStatus(Optional.of(false)).build())
  }
```

server/app/views/admin/programs/ProgramApplicationListView.java
nb1701: I'm not sure we can actually set green for approved, because it will be up to the admin to create an "approved" status to set it to, so we'd have to have a different mechanism to say that a particular status is the "approved" status.

dkatzz (commented):
It may be worth adding a couple of browser tests as well in https://github.com/civiform/civiform/blob/main/browser-test/src/admin_application_statuses.test.ts and https://github.com/civiform/civiform/blob/main/browser-test/src/civiform_admin_program_statuses.test.ts

nb1701:
> It may be worth adding a couple of browser tests as well in https://github.com/civiform/civiform/blob/main/browser-test/src/admin_application_statuses.test.ts and https://github.com/civiform/civiform/blob/main/browser-test/src/civiform_admin_program_statuses.test.ts

Oh yeah, definitely!  Sorry, forgot to mention, still working on those :)

server/app/controllers/admin/AdminProgramStatusesController.java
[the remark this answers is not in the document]
dkatzz: LGTM, thanks!

nb1701:
@dkatzz Had a convo with @mferhut, looks like the text on the badges should be blue/green as well according to https://www.figma.com/file/xOimWyFjr4haCJuOFcAifh/CiviForm---Common-intake-(aka-Pre-screener)-feature---Public-file?node-id=939%3A15891&t=zFrZ5rngEtrN7FVd-0.  Mind if I add that to the eligibility badges too in this PR?

dkatzz:
> @dkatzz Had a convo with @mferhut, looks like the text on the badges should be blue/green as well according to https://www.figma.com/file/xOimWyFjr4haCJuOFcAifh/CiviForm---Common-intake-(aka-Pre-screener)-feature---Public-file?node-id=939%3A15891&t=zFrZ5rngEtrN7FVd-0. Mind if I add that to the eligibility badges too in this PR?

The eligibility tags should already be green - https://github.com/civiform/civiform/blob/main/browser-test/image_snapshots/application_navigation_test/eligible-home-page-program-tag.png, but if there is anything to change yeah, feel free to add as part of this!

nb1701:
> > @dkatzz Had a convo with @mferhut, looks like the text on the badges should be blue/green as well according to https://www.figma.com/file/xOimWyFjr4haCJuOFcAifh/CiviForm---Common-intake-(aka-Pre-screener)-feature---Public-file?node-id=939%3A15891&t=zFrZ5rngEtrN7FVd-0. Mind if I add that to the eligibility badges too in this PR?
> 
> The eligibility tags should already be green - https://github.com/civiform/civiform/blob/main/browser-test/image_snapshots/application_navigation_test/eligible-home-page-program-tag.png, but if there is anything to change yeah, feel free to add as part of this!

I think the change would be to make the text a dark green as well, instead of black (and dark blue for the blue tag), but @mferhut correct me if I'm wrong!  I can add that to this PR.

server/app/services/applicant/ApplicantService.java
```diff
@@ -519,41 +585,125 @@ private void notifyTiSubmitter(
...
+        .thenRunAsync(
+            () -> {
+              // Not blocking since they've already completed
+              Optional<Locale> maybeLocale = getLocaleFuture.join();
+              Optional<String> maybeEmail = getEmailFuture.join();
+              Locale locale = maybeLocale.orElse(LocalizedStrings.DEFAULT_LOCALE);
+              if (emailBody.isPresent() && maybeEmail.isPresent()) {
+                String programNameForEmail = programDef.localizedName().getOrDefault(locale);
+                // TODO(#3377): Similar to the code in ProgramAdminApplicationService, we are mixing
+                // english
```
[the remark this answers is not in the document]
nb1701: Thanks for catching this.  I've noticed the formatter does this sometimes.  I wonder if we can make it smarter somehow.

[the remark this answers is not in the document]
nb1701: Also, this comment is no longer necessary anyway :)

mferhut:
That's correct. Please update text to dark green. Please reference: https://www.figma.com/file/xOimWyFjr4haCJuOFcAifh/CiviForm---Common-intake-(aka-Pre-screener)-feature---Public-file?node-id=939%3A15891&t=EF66RGACTa7vUoCt-1

Thanks!

server/app/views/admin/programs/ProgramApplicationListView.java
nb1701: Yeah, I think we'll just leave it blue and maybe later on we can think about if we want to create some default "Approved, Rejected, Waiting" statuses or something.

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
nb1701: @dkatzz As it turns out, I missed that ProgramAdminApplicationService is also sending out emails, which I think handles sending emails when the admin goes and changes the status on an application.  Is it fine to do things synchronously in there, since it's an admin-facing thing, or do we need to do things asynchronously there as well?  Looks like it's all pretty much sync already.

dkatzz:
> > > @dkatzz Had a convo with @mferhut, looks like the text on the badges should be blue/green as well according to https://www.figma.com/file/xOimWyFjr4haCJuOFcAifh/CiviForm---Common-intake-(aka-Pre-screener)-feature---Public-file?node-id=939%3A15891&t=zFrZ5rngEtrN7FVd-0. Mind if I add that to the eligibility badges too in this PR?
> > 
> > 
> > The eligibility tags should already be green - https://github.com/civiform/civiform/blob/main/browser-test/image_snapshots/application_navigation_test/eligible-home-page-program-tag.png, but if there is anything to change yeah, feel free to add as part of this!
> 
> I think the change would be to make the text a dark green as well, instead of black (and dark blue for the blue tag), but @mferhut correct me if I'm wrong! I can add that to this PR.

Ah sorry I missed the word "text" :) sounds good - thanks for making that change in both places!

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
dkatzz: Hmm based on [this](https://docs.civiform.us/contributor-guide/developer-guide/development-standards#async-request-handling) in the docs, it seems okay if things are synchronous on the admin side. We may want to consider refactoring that code in the future to be asynchronous though.

nb1701:
Text colors and i18n for ProgramAdminApplicationService added

nb1701:
Docs update https://github.com/civiform/docs/pull/263

browser-test/image_snapshots/applicant_application_statuses_test/program-list-with-status.png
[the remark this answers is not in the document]
dkatzz: Got it, that makes sense!

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -348,6 +348,100 @@ describe('view program statuses', () => {
...
+
+    it('when a default status is set, applications with that status show (default)', async () => {
+      const {page, adminPrograms} = ctx
+      expect(
+        await page.innerText(
+          adminPrograms.selectApplicationCardForApplicant('Guest'),
+        ),
+      ).toContain(`Status: ${waitingStatus} (default)`)
+      expect(
+        await page.innerText(
```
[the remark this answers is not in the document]
nb1701: Good idea!

browser-test/src/civiform_admin_program_statuses.test.ts
```diff
@@ -71,6 +72,159 @@ describe('modify program statuses', () => {
...
+      })
+      await confirmHandle.click()
+
+      await waitForPageJsLoad(page)
+      await adminProgramStatuses.expectProgramManageStatusesPage(programName)
+      await validateToastMessage(
+        page,
+        'Default Status 1 has been updated to the default status',
+      )
+      await validateScreenshot(page, 'status-list-with-default-status')
```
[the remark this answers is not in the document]
nb1701: It wasn't clear to me when we actually programmatically look for things being in the right place on the page vs. when we just let the screenshot validate that.  Seems like double coverage is good though :)

server/conf/messages.am
```diff
@@ -436,3 +436,16 @@ adminValidation.multiOptionEmpty=ባለብዙ አማራጭ ጥያቄዎች ባ
 error.notFoundTitle=ለመጎብኘት የሞከሩትን ገጽ ማግኘት አልቻልንም
 error.notFoundDescription=«እባክዎ ወደኋላ ይሂዱ ወይም ይመለሱ ወደ »
 error.notFoundDescriptionLink=መነሻ ገጽ
+
+#-------------------------------------------------------------#
+# EMAILS - boilerplate text contained in emails sent to users #
+#-------------------------------------------------------------#
+email.applicationReceivedBody=የ{0} ፕሮግራም ማመልከቻዎ ደርሷል። የአመልካች መታወቂያዎ {1} ነው፣ እና የማመልከቻ መታወቂያው {2} ነው
+email.applicationReceivedSubject=የ{0} ፕሮግራም ማመልከቻዎ ደርሷል
+email.applicationUpdateSubject=An update on your application {0}
```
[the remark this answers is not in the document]
nb1701: That makes sense.  We can hang onto this PR until we get the next batch of translations, or I could split out the i18n stuff into a separate PR.  That kind of scope creeped into this PR.

browser-test/src/admin_application_statuses.test.ts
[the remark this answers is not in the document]
nb1701: Going to refactor a few other things as well.  Too much copy/paste.

nb1701:
Alright, I think this has everything addressed now, once we get the rest of the translations.

browser-test/src/support/admin_programs.ts
```diff
@@ -66,6 +66,28 @@ export class AdminPrograms {
...
+    expect(
+      await this.page.innerText(
+        this.selectApplicationCardForApplicant(applicant),
+      ),
+    ).toContain(`Status: ${statusString}`)
+  }
+
+  async expectApplicationStatusDoesntContain(
+    applicant: string,
+    doesntContain: string,
```
dkatzz, line 82: **nitpick:** maybe call this statusString too?

server/conf/messages.am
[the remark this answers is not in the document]
dkatzz: Sounds good! Unfortunately it can take up to 2 weeks for new strings to come back as a heads up, but if that's okay, keeping this until those come back sound fine to me!

browser-test/src/support/admin_program_statuses.ts
```diff
@@ -36,6 +36,22 @@ export class AdminProgramStatuses {
     }
   }
 
+  async expectStatusIsDefault(statusName: string) {
```
dkatzz, line 39: **praise:** Thanks for splitting all these out!

dkatzz (approved):
LGTM once those translations are added! Thanks for all your work on this!

server/conf/messages.am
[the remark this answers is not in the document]
nb1701: I'll check with Isabel.  I think there's no rush to get this in, but if we want to just close it out, I'll remove these.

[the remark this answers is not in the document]
nb1701: I just took those three out of the non-English translations, so I think we can merge this whenever.

PR #4279, opened 2023-02-24
Rename auth classes

MichaelZetune:
### Description

Many classes in the `auth` package are confusingly named. This PR renames many of them to make it easier to read and understand code in the future.

For example, our `OidcProfileAdapter` extends pac4j's `OidcProfileCreator`, which mainly has a `Optional<UserProfile> create()` method. We should be calling our class `CiviformOidcProfileCreator` to (1) differentiate it from pac4j's and (2) more importantly, continue suggesting that it does in fact create, not adapt, profiles.

Also, implement some trivial fixes in the auth package that IntelliJ suggests

### Issue(s)

Steps towards #4280


bion (approved):
This is a great improvement, thanks for taking some time to make the code base more maintainable!

PR #4284, opened 2023-02-24
Add an indicator showing which program is the common intake.

reemaka:
### Description

Add an indicator showing which program is the common intake form (aka pre-screener) in the admin program list page.

#### General

#### New Features


server/app/views/components/ProgramCardFactory.java
```diff
@@ -65,9 +70,19 @@ public DivTag renderCard(ProgramCardData cardData) {
...
                                 "text-black",
                                 "font-bold",
-                                "text-xl"),
+                                "text-xl"))
+                    .with(
                         p(programDescriptionText)
-                            .withClasses("line-clamp-2", "text-gray-700", "text-base")),
+                            .withClasses("line-clamp-2", "text-gray-700", "text-base"))
+                    .condWith(
+                        featureFlags.isIntakeFormEnabled(request)
```
avaleske, line 78: Can we pull this check out into its own method, like this:
https://github.com/civiform/civiform/blob/5fbdecbab2ec80ff185857ecdd985eabc65bc1bc/server/app/views/components/QuestionBank.java#L258-L262

I think that will make this a little easier to scan quickly

server/app/views/components/ProgramCardFactory.java
```diff
@@ -165,11 +180,16 @@ private static ProgramDefinition getDisplayProgram(ProgramCardData cardData) {
 
   public static Comparator<ProgramCardData> lastModifiedTimeThenNameComparator() {
```
avaleske, line 181: Can we update the name of this comparator so it reflects the additional behavior of sorting the common intake form above everything else?

PR #4287, opened 2023-02-27
Fix intake form uniqueness logic to look at active and draft programs.

reemaka:
### Description

Previously, the uniqueness validation logic for the common intake form looked at all programs. This included obsolete programs and active programs for which there was a draft, however, which meant that you couldn't actually change which program was the common intake form. This change fixes it by only checking for the common intake form among draft programs and active programs that don't have a draft. 

#### General

### DIFF
--- a/server/app/services/program/ProgramServiceImpl.java
+++ b/server/app/services/program/ProgramServiceImpl.java
@@ -214,8 +214,7 @@ public final class ProgramServiceImpl implements ProgramService {
       errorsBuilder.add(CiviFormError.of("A program link must begin with 'http://' or 'https://'"));
     }
     if (isIntakeFormFeatureEnabled) {
-      if (programType == ProgramType.COMMON_INTAKE_FORM
-          && programRepository.commonIntakeFormExists()) {
+      if (programType.equals(ProgramType.COMMON_INTAKE_FORM) && getCommonIntakeForm().isPresent()) {
         errorsBuilder.add(CiviFormError.of(DUPLICATE_INTAKE_FORM_MSG));
       }
     } else {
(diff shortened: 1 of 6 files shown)


bion (approved):

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -1166,4 +1167,20 @@ private ProgramQuestionDefinition syncProgramQuestionDefinition(
...
     return pqd.loadCompletely(programDefinitionId, questionDefinition);
   }
+
+  /*
+   * Looks at the most recent version of each program and returns the program marked as the
+   * common intake form if it exists. The most recent version may be in the draft or active stage.
+   */
+  private Optional<ProgramDefinition> getCommonIntakeForm() {
+    ActiveAndDraftPrograms activeAndDraftPrograms = getActiveAndDraftPrograms();
+    for (String name : activeAndDraftPrograms.getProgramNames()) {
```
avaleske, line 1177: idk if this suggestion makes sense, but:
What if, instead of getting a list of active and draft program names from `ActiveAndDraftPrograms` and then asking `ActiveAndDraftPrograms` if each of them is the most recent version, we added a `getMostRecentVersionsOfActiveAndDraftPrograms()` method to `ActiveAndDraftPrograms` that did that work for us?

Then this could look like
```
return getActiveAndDraftPrograms().getMostRecentVersionsOfActiveAndDraftPrograms().stream().findFirst(p -> p.isCommonIntakeForm());
```
since `findFirst()` returns an optional already.

avaleske (changes requested):
I hadn't even thought of draft programs when reviewing the other PR so good catch!

I had a question about how we're organizing some of the logic to find the most recent programs, but otherwise lgtm

server/app/services/program/ProgramServiceImpl.java
avaleske: And I think that `getMostRecentProgramDefintions()` function could look like 
```java
  public ImmutableList<ProgramDefinition> getMostRecentProgramDefinitions() {
    return getProgramNames()
      .stream()
      .map(name -> getDraftProgramDefinition(name).orElseGet(getActiveProgramDefinition(name)::get))
  }
```

reemaka: Honestly I think they're pretty equivalent. Not doing that is slightly more efficient because it exits early without getting the definition for every name, but they have the same big O time so it's negligible overall. Changed it since you have a preference readability-wise.

avaleske (approved):

server/app/services/program/ProgramServiceImpl.java
avaleske: Yeah that's fair. It wasn't a strongly held preference, fwiw, but thanks for changing it 🙂

PR #4320, opened 2023-03-03
Adding new container for mock web services

gwendolyngoetz:
### Description

Adding new container for development time use to provide mock web service endpoints, such as for ESRI. It's able to deal with other possible endpoints in the future so we don't have to make multiple different containers.

It's a very light Flask app that just serves up specific json responses based on a handful of known inputs. I built it to mimic that FakeEsriService.java file so they can be used interchangeably. Everything is plumbed into the build with a container already.

#### General

### DIFF
--- a/bin/build-mock-web-services
+++ b/bin/build-mock-web-services
@@ -1,0 +1,41 @@
+#! /usr/bin/env bash
+
+# DOC: Build a new docker image for running the mock web services container and optionally push to Docker Hub if PUSH_IMAGE is set to anything. Builds for "linux/amd64" unless PLATFORM is set
+
+source bin/lib.sh
+
+IMAGE="mock-web-services"
+LOCATION="mock-web-services/"
+DOCKERFILE="mock-web-services/mock-web-services.Dockerfile"
+
+BUILD_ARGS="-f ${DOCKERFILE}
+  -t civiform/${IMAGE}:latest
+  --cache-from civiform/${IMAGE}
+  --build-arg BUILDKIT_INLINE_CACHE=1
+  --build-context server=server/
+  ${LOCATION}"
+
+PLATFORM_ARG=()
+if [[ "${PLATFORM}" ]]; then
+  PLATFORM_ARG=(--platform "${PLATFORM}")
+fi
+
+# Build the multi-platform image
+echo "start ${IMAGE} build"
+cmd="docker buildx build ${PLATFORM_ARG[@]} ${BUILD_ARGS[@]}"
+$cmd
+
+# Load the image from the cache
+echo "load ${IMAGE} build"
+cmd="docker buildx build --load ${BUILD_ARGS}"
+$cmd
+
+if [[ "${PUSH_IMAGE}" ]]; then
+  docker::do_dockerhub_login
+  # Push the image from the cache to dockerhub
+  echo "push ${IMAGE} build"
+  cmd="docker buildx build --push ${PLATFORM_ARG[@]} ${BUILD_ARGS}"
+  $cmd
+fi
+
+docker tag "civiform/${IMAGE}:latest" "civiform-${IMAGE}:latest"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,9 @@ services:
       # Use an explicit port to not conflict with other test instances.
       - OIDC_PORT=3390
 
+  mock-web-services:
+    image: civiform/mock-web-services
+
   db:
     image: postgres:12.14
     restart: always
@@ -53,6 +56,7 @@ services:
     links:
       - 'db:database'
       - 'dev-oidc'
+      - 'mock-web-services'
     expose:
       - '9000'
       - '8457'
@@ -109,4 +113,8 @@ services:
       - WHITELABEL_CIVIC_ENTITY_SHORT_NAME
       - WHITELABEL_CIVIC_ENTITY_FULL_NAME
       - FAVICON_URL
+      - ESRI_ADDRESS_CORRECTION_ENABLED=true
+      - ESRI_FIND_ADDRESS_CANDIDATES_URL=http://mock-web-services:8000/esri/findAddressCandidates
+      - ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED=true
+      - ESRI_ADDRESS_SERVICE_AREA_VALIDATION_URLS.0=http://mock-web-services:8000/esri/serviceAreaFeatures
     entrypoint: /bin/bash
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -113,8 +113,4 @@ services:
       - WHITELABEL_CIVIC_ENTITY_SHORT_NAME
       - WHITELABEL_CIVIC_ENTITY_FULL_NAME
       - FAVICON_URL
-      - ESRI_ADDRESS_CORRECTION_ENABLED=true
-      - ESRI_FIND_ADDRESS_CANDIDATES_URL=http://mock-web-services:8000/esri/findAddressCandidates
-      - ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED=true
-      - ESRI_ADDRESS_SERVICE_AREA_VALIDATION_URLS.0=http://mock-web-services:8000/esri/serviceAreaFeatures
     entrypoint: /bin/bash
--- a/mock-web-services/app.py
+++ b/mock-web-services/app.py
@@ -2,28 +2,46 @@ from flask import Flask, Response, request, json
 from esri import EsriMockWebService
 
 app = Flask(__name__)
-esriWS = EsriMockWebService(app)
+esri_ws = EsriMockWebService(app)
 
 
 @app.route("/esri/findAddressCandidates")
-def findAddressCandidates():
+def find_address_candidates():
+    """
+    Endpoint to mock Esri findAddressCandidates. We use the address as a key in
+    determining which type of response this will return.
+
+    Valid options are:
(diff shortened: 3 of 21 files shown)


avrittrohwer (commented):
To make it easier to write tests against the mock server, could you add a esri-esri-mock-service/README.md that describes the available API paths and the data they return?

esri-mock-service/app.py
```diff
@@ -0,0 +1,164 @@
+from flask import Flask
+
+app = Flask(__name__)
+
+
+@app.route("/findAddressCandidates")
+def findAddressCandidates():
+    jsonstr = """{
```
avrittrohwer, line 8: Do you think it would be more maintainable if we stored these responses as json files, then read them into memory at server start?

avrittrohwer:
Alternative approach we may want to consider (that would also work for the staging deployments): add a FakeESRIClient akin to https://github.com/civiform/civiform/blob/main/server/app/auth/FakeAdminClient.java

esri-mock-service/app.py
seanmalbert: We do have these stored in `test/resources/esri` used by the unit tests, so hopefully we can use the same ones.

seanmalbert:
> Alternative approach we may want to consider: add a FakeESRIClient akin to https://github.com/civiform/civiform/blob/main/server/app/auth/FakeAdminClient.java
> 
> Would be a lot simpler to integrate (don't need to touch any of the docker stuff) and would work on the staging deployments because the fake client would be included in the main civiform server docker image

Thanks for sharing this, @avrittrohwer. This seems like a great alternative and should be simple enough to inject into ApplicantService and set as the esriClient if `FakeEsriClient.canEnable(host)` is true.

gwendolyngoetz:
Oh, I like FakeAdminClient idea much better. Good catch

nb1701:
@gwendolyngoetz Okay to close this since we have the fake esri client now?

gwendolyngoetz:
No, Sean's document planned on keeping it. Needs to be updated.

gwendolyngoetz:
There doesn't appear to be any docs on usage of the FakeEsriClient so I'll write something up for both.

bin/build-mock-web-services
```diff
@@ -0,0 +1,41 @@
...
+
+BUILD_ARGS="-f ${DOCKERFILE}
+  -t civiform/${IMAGE}:latest
+  --cache-from civiform/${IMAGE}
+  --build-arg BUILDKIT_INLINE_CACHE=1
+  --build-context server=server/
+  ${LOCATION}"
+
+PLATFORM_ARG=()
+if [[ "${PLATFORM}" ]]; then
```
bion, line 19: **nitpick:** https://google.github.io/styleguide/shellguide.html#testing-strings
```suggestion
if [[ -n "${PLATFORM}" ]]; then
```

bin/build-mock-web-services
```diff
@@ -0,0 +1,41 @@
...
+  PLATFORM_ARG=(--platform "${PLATFORM}")
+fi
+
+# Build the multi-platform image
+echo "start ${IMAGE} build"
+cmd="docker buildx build ${PLATFORM_ARG[@]} ${BUILD_ARGS[@]}"
+$cmd
```
bion, line 26: **question (blocking):** why assign the command to a variable first instead of `docker buildx build "${PLATFORM_ARG[@]}" "${BUILD_ARGS[@]}"`?

mock-web-services/app.py
```diff
@@ -0,0 +1,31 @@
+from flask import Flask, Response, request, json
+from esri import EsriMockWebService
+
+app = Flask(__name__)
+esriWS = EsriMockWebService(app)
+
+
+@app.route("/esri/findAddressCandidates")
```
bion, line 8: **suggestion (blocking):** Please add a doc comment for this file and a README.md for this new directory

mock-web-services/app.py
```diff
@@ -0,0 +1,31 @@
+from flask import Flask, Response, request, json
+from esri import EsriMockWebService
+
+app = Flask(__name__)
+esriWS = EsriMockWebService(app)
```
bion, line 5: **nitpick (blocking):** snake_case for python
```suggestion
esri_ws = EsriMockWebService(app)
```

mock-web-services/app.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@app.route("/esri/findAddressCandidates")
+def findAddressCandidates():
+    try:
+        address = request.args.get("address")
+        return esriWS.findAddressCandidates(address)
+    except Exception as e:
+        print(e)
+        return Response("Bad request", status=400)
```
bion, line 15: **suggestion (blocking):** include the exception in the response for ease of reference

mock-web-services/app.py
```diff
@@ -0,0 +1,31 @@
...
+
+@app.route("/esri/serviceAreaFeatures")
+def serviceAreaFeatures():
+    try:
+        geometry = json.loads(request.args.get("geometry").replace("'", '"'))
+        latitude = geometry.get("y")
+        return esriWS.serviceAreaFeatures(latitude)
+    except Exception as e:
+        print(e)
+        return Response("Bad request", status=400)
```
bion, line 26: **suggestion (blocking):** include the exception in the response for ease of reference

mock-web-services/esri/__init__.py
```diff
@@ -0,0 +1,32 @@
+from webservice import WebService
+
+
+class EsriMockWebService(WebService):
```
bion, line 4: **suggestion (blocking):** please add a doc string for this file

mock-web-services/esri/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, app):
+        WebService.__init__(self, app)
+        self.app = app
+        self.file_root = "../server/test/resources/esri/"
```
bion, line 9: **suggestion (blocking):** please mention in the doc string and also maybe the README that this loads/shares its data fixtures from the server junit test subdir

mock-web-services/app.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        address = request.args.get("address")
+        return esriWS.findAddressCandidates(address)
+    except Exception as e:
+        print(e)
+        return Response("Bad request", status=400)
+
+
+@app.route("/esri/serviceAreaFeatures")
+def serviceAreaFeatures():
```
bion, line 19: **nitpick:** snake_case for method names in this file

mock-web-services/esri/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        if address == "Address In Area" or address == "Legit Address":
+            return self.returnJsonResponseFromFile(
+                self.file_root + "findAddressCandidates.json")
+        elif address == "Bogus Address":
+            return self.returnJsonResponseFromFile(
+                self.file_root + "findAddressCandidatesNoCandidates.json")
+        else:
+            raise Exception("Invalid mock request")
+
+    def serviceAreaFeatures(self, latitude):
```
bion, line 21: **nitpick:** snake_case for method names in this file

mock-web-services/webservice/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+        with open(filename, "r") as file:
+            jsonstr = file.read()
+
+        return jsonstr
+
+    def returnJsonResponse(self, jsonstr):
+        return self.app.response_class(
+            response=jsonstr, mimetype="application/json")
+
+    def returnJsonResponseFromFile(self, filename):
```
bion, line 16: **nitpick:** snake_case for method names in this file

bion (changes requested):
This is a huge improvement for the reliability of the ESRI-integration!

bin/build-mock-web-services
gwendolyngoetz: I'd have to look. All the existing ones are this way and I just copied it over. I'd rather they maintain the same format so if I start changing one I'd rather clean them all up in a separate PR.

bin/build-mock-web-services
gwendolyngoetz: Same as the comment below. If I'm going to clean up these files I'd rather do them all together in a separate PR.

docker-compose.yml
```diff
@@ -109,4 +113,8 @@ services:
       - WHITELABEL_CIVIC_ENTITY_SHORT_NAME
       - WHITELABEL_CIVIC_ENTITY_FULL_NAME
       - FAVICON_URL
+      - ESRI_ADDRESS_CORRECTION_ENABLED=true
```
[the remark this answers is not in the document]
gwendolyngoetz: I want to have dev/browser-tets and the end to end runs in CI using the service. Unit tests should still use the FakeEsriClient in java. There are not set in the application.test.conf. Let me double check the order the config files layer things.

I don't think we need to remove the feature flags or enable it for prod. If a Civic Entity doesn't have or want to use it there's no need to enable it at all.

mock-web-services/app.py
gwendolyngoetz: Thanks. It gets confusing going back and forth sometimes between java and python.

bion (approved):

docker-compose.yml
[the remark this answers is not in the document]
gwendolyngoetz: I moved the settings from the docker-compose file and set them in the application.dev.conf file. The browser tests inherit from the dev.conf file. I also set the test.conf to not use a url which will use the FakeEsriClient class. This should all work how I want.

bion (approved):

PR #4321, opened 2023-03-03
Add enumerator indexes to labels and buttons

### LINKED ISSUE #2624: (WCAG 2.1) Add indexes to enumerator labels

Severity (Critical/Serious/Moderate/Good to have): Good to have

In the below example, the placeholder text "child name" could be changed to "child # 2 name".

Screen Shot 2022-06-01 at 6 20 13 PM

Related issue:
* https://github.com/seattle-uat/civiform/issues/2617

rockycodes:
### Description

This PR add logic to add indexes to enumerator labels and "Remove" buttons. The indexes are re-ordered when an element is removed from the page. 

This PR also updates the "Preview" behavior for enumerator questions -- adding indexes to labels and buttons and fixing an issue where clicking the "Add" button did not actually result in a new field rendering.

This PR also fixes a bug in Firefox where the default button type was causing the page to re-load when clicked.

https://user-images.githubusercontent.com/24319951/222842446-0828e984-66dc-44f3-a044-5f09438a9bde.mov

#### General

#### User visible changes

### Issue(s) this completes

Fixes #2624;


server/app/assets/javascripts/admin_entry_point.ts
```diff
@@ -31,4 +32,5 @@ window.addEventListener('load', () => {
   modal.init()
   questionBank.init()
   preview.init()
+  enumerator.init()
```
rockycodes, line 35: fixes bug where enumerator fields weren't being added in admin preview

server/app/assets/javascripts/enumerator.ts
```diff
@@ -174,14 +187,18 @@ function addEnumeratorListeners() {
...
       for (const newNode of Array.from(record.addedNodes)) {
-        const newInputs = Array.from(
-          (<Element>newNode).querySelectorAll('input'),
-        )
-        newInputs.forEach((newInput) => {
-          newInput.addEventListener('input', () => {
-            maybeToggleEnumeratorAddButton(enumeratorQuestion)
+        // changes to the label and button texts trigger the mutationObserver which results in an error
+        // this if statement protects against that case
+        if (typeof (<Element>newNode).querySelectorAll !== 'undefined') {
```
swatkat1, line 192: wondering if this explicit call to not equal is needed. I have seen if(element) --> automatically checks if the element is not null. So something like if(<Element>newNode).querySelectorAll)) might work here

rockycodes: good callout! i thought i had tried that and it didn't work, but i just tried again and it works fine so i updated it.

swatkat1 (approved):

PR #4342, opened 2023-03-08
Always allow submitting an application if eligibility is non-gating.

reemaka:
### Description

Updates the application logic for non-gating eligibility so that application submission is never blocked and text about the applicant not qualifying is removed.

#### General

#### New Features

### DIFF
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -507,6 +507,40 @@ describe('Applicant navigation flow', () => {
       await applicantQuestions.expectIneligiblePage()
     })
 
+    it('shows not eligible upon submit with ineligible answer with gating eligibility', async () => {
+      const {page, applicantQuestions} = ctx
+      await loginAsGuest(page)
+      await selectApplicantLanguage(page, 'English')
+      await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
+      await enableFeatureFlag(page, 'nongated_eligibility_enabled')
+      await applicantQuestions.applyProgram(fullProgramName)
+
+      // Fill out application and submit.
+      await applicantQuestions.answerNumberQuestion('1')
+      await applicantQuestions.clickNext()
+      await applicantQuestions.expectIneligiblePage()
+
+      // Verify the question is marked ineligible.
+      await applicantQuestions.gotoApplicantHomePage()
+      await applicantQuestions.seeEligibilityTag(
+        fullProgramName,
+        /* isEligible= */ false,
+      )
+      await applicantQuestions.clickApplyProgramButton(fullProgramName)
+      await applicantQuestions.expectQuestionIsNotEligible(
+        AdminQuestions.NUMBER_QUESTION_TEXT,
+      )
+
+      // Answer the other question.
+      await applicantQuestions.clickContinue()
+      await applicantQuestions.answerEmailQuestion('email@email.com')
+
+      // Submit and expect to be told it's ineligible.
+      await applicantQuestions.clickNext()
+      await applicantQuestions.clickSubmit()
+      await applicantQuestions.expectIneligiblePage()
+    })
+
     it('shows may be eligible with nongating eligibility', async () => {
       const {page, adminPrograms, applicantQuestions} = ctx
       await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -370,6 +370,7 @@ describe('Applicant navigation flow', () => {
       await loginAsGuest(page)
       await selectApplicantLanguage(page, 'English')
       await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
+      await disableFeatureFlag(page, 'nongated_eligibility_enabled')
       await applicantQuestions.applyProgram(fullProgramName)
 
       // Fill out application and submit.
@@ -398,6 +399,7 @@ describe('Applicant navigation flow', () => {
       await loginAsGuest(page)
       await selectApplicantLanguage(page, 'English')
       await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
+      await disableFeatureFlag(page, 'nongated_eligibility_enabled')
       await applicantQuestions.applyProgram(fullProgramName)
 
       // Fill out application and without submitting.
@@ -434,6 +436,7 @@ describe('Applicant navigation flow', () => {
       // Add the partial program.
       await loginAsAdmin(page)
       await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
+      await disableFeatureFlag(page, 'nongated_eligibility_enabled')
       await adminPrograms.addProgram(overlappingOneQProgramName)
       await adminPrograms.editProgramBlock(
         overlappingOneQProgramName,
@@ -475,6 +478,7 @@ describe('Applicant navigation flow', () => {
       await loginAsGuest(page)
       await selectApplicantLanguage(page, 'English')
       await enableFeatureFlag(page, 'program_eligibility_conditions_enabled')
+      await disableFeatureFlag(page, 'nongated_eligibility_enabled')
       await applicantQuestions.applyProgram(fullProgramName)
 
       // Fill out application and submit.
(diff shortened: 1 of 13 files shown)


dkatzz:
More of a high level question, but I feel like hiding the "May not qualify" tag / text is almost separate from the gating feature. I would think that a program would want to show the "May not qualify" text, so users know they may not qualify before submitting an application, but still allow the user to submit it or at least have this as another program option like "show an applicant when they may not qualify". I guess more of a UX question, but just wanted to see if this was discussed or considered? If we think we may add this functionality in the future, it may be worth having another setting on the program for showIfIneligble or something and set it to !isGatingOn for now. So basically the combinations of program settings could be:
- showIfIneligble = true && gateSubmission = true -> applicant sees all the tags about whether or not they're eligible, but they can submit the application
- showIfIneligble = false && gateSubmission = true -> this would be the default setting for now if gating is on since these are the requirements for common intake and would be that the applicant could see if they're eligible, but wouldn't know if they weren't eligible and could submit any application
- showIfIneligble = false && gateSubmission = false -> the applicant could see if they weren't eligible and wouldn't be allowed to submit an application if they're not eligible

cc @mferhut

mferhut:
We won't need it for Arkansas as they want to avoid negative phrasing. However, I do agree we should offer that option when an entity requests this scenario, where they would like to still message the applicant that they 'may not qualify' regardless of gated or non-gated. 

In terms of where to place that setting, for gated and non-gated initially we heard having this setting at the program level makes sense to quickly set eligibility in one setting however Arkansas has recently mentioned that there could be future programs that need control at a more granular level thus it may make sense to move the gated and non-gated eligibility setting to each eligibility condition. If that's the case then the future setting for still showing 'may not qualify' message regardless of gated or non-gated will also be shown at each eligibility condition as well.

dkatzz:
Sounds good, thanks both!! @reemaka I think you're pretty far along at having the gating / non-gating setting at the program level, right? If that's the case, I'm thinking it makes sense to move forward with the changes and then as a follow-up we can think through how to add these settings on the eligibility condition level. And if it isn't more difficult to add a setting later to show / hide the "may not qualify" messages as well, I'm cool with moving forward with this implementation for common intake for and rethinking this and adding it later.

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -569,6 +568,21 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
                     applicantId, programId, nextBlockIdMaybe.get())));
   }
 
+  private boolean shouldRenderIneligibleBlockView(
```
dkatzz, line 571: **suggestion:** Would you mind adding a comment here too similar to below explaining the cases where we should / shouldn't render this view?

mferhut:
I spoke with Reema, we should proceed with what is built which is eligibility at the program level and when we receive a future request from entity to move to a granular level vs program level eligibility control we can look at making updates at that point and what that can look like.

server/app/services/applicant/ApplicantService.java
```diff
@@ -416,19 +423,36 @@ CompletionStage<Application> submitApplication(
...
+      boolean eligibilityFeatureEnabled,
+      boolean nonGatedEligibilityFeatureEnabled,
+      long programId) {
     // Check that all blocks have been answered.
     if (!roApplicantProgramService.getFirstIncompleteBlockExcludingStatic().isEmpty()) {
       throw new ApplicationOutOfDateException();
     }
-    if (eligibilityFeatureEnabled && !roApplicantProgramService.isApplicationEligible()) {
+
+    if (!eligibilityFeatureEnabled) {
```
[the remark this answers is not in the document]
reemaka: The non-gated eligibility feature flag should only have any effect if the eligibility flag is also enabled. If I moved this then you'd be able to have a situation where turning on the non-gated flag while leaving the eligibility flag off changes the behavior, which we don't want.

browser-test/src/application_navigation.test.ts
```diff
@@ -502,6 +506,109 @@ describe('Applicant navigation flow', () => {
       await applicantQuestions.clickSubmit()
       await applicantQuestions.expectIneligiblePage()
     })
+
+    it('shows not eligible upon submit with ineligible answer with gating eligibility', async () => {
```
dkatzz, line 510: **suggestion:** Maybe say something like "with gating eligibility available, but not set" to make this a bit more clear?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -569,6 +568,22 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
                     applicantId, programId, nextBlockIdMaybe.get())));
   }
 
+  /** Returns true if eligibility is gating and the application is ineligible, false otherwise. */
```
dkatzz, line 571: **nitpick:** "block" instead of "application"

server/app/services/applicant/AnswerData.java
```diff
@@ -59,6 +59,12 @@ public static Builder builder() {
    */
   public abstract boolean isEligible();
 
+  /**
+   * True if the program's eligibility is set to gating. If false, tags that note that the applicant
```
dkatzz, line 63: **suggestion:** If "true", right?

dkatzz (approved):

server/app/services/applicant/AnswerData.java
reemaka: No, if eligibilityIsGating == true, eligibility gates submission and all tags should be shown. If eligibilityIsGating == false, eligibility does NOT gate submission and 'not eligibile' tags should be hidden.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -367,6 +366,21 @@ private LiTag programCard(
...
+   * If eligibility is gating, the eligibility tag should always show when present. If eligibility
+   * is non-gating, the eligibility tag should only show if the user may be eligible.
+   */
+  private boolean shouldShowEligibilityTag(
+      Http.Request request, ApplicantService.ApplicantProgramData cardData) {
+    if (!featureFlags.isProgramEligibilityConditionsEnabled(request)
+        || !cardData.isProgramMaybeEligible().isPresent()) {
+      return false;
+    }
+    return !featureFlags.isNongatedEligibilityEnabled(request)
```
avaleske, line 379: making sure I'm understanding:
Regardless of the value of `isNongatedEligibilityEnabled()`, we want to show the tag if the program has gated eligibility, even if they're not eligible?

server/app/services/applicant/AnswerData.java
dkatzz: Lol sorry - Friday brain. Yes!

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -367,6 +366,21 @@ private LiTag programCard(
         .with(actionDiv);
   }
 
+  /**
+   * If eligibility is gating, the eligibility tag should always show when present. If eligibility
+   * is non-gating, the eligibility tag should only show if the user may be eligible.
+   */
+  private boolean shouldShowEligibilityTag(
+      Http.Request request, ApplicantService.ApplicantProgramData cardData) {
+    if (!featureFlags.isProgramEligibilityConditionsEnabled(request)
```
avaleske, line 375: nit: These conditions don't feel related or dependent on each other. Can we break them apart into

```java
if (!featureFlags.isProgramEligibilityConditionsEnabled(request) {
    return false;
}

if(!cardData.isProgramMaybeEligible().isPresent()) {
    return false;
}
```
so it's a little easier to parse?

browser-test/src/application_navigation.test.ts
reemaka: To clarify, eligibilityIsGating is by default true and means that eligibility does gate submission. If it's set to false, then it does NOT gate submission. So, I don't think your suggestion is accurate but I could change it to 'shows not eligible upon submit with ineligible answer with non-gating eligibility available, but not set' if you think that's clearer, or maybe just 'with eligibility set to gating' is clearer?

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -951,13 +957,52 @@ public void submitApplication_failsWithApplicationNotEligibleException() {
                         applicant.id,
                         programDefinition.id(),
                         trustedIntermediaryProfile,
-                        /* eligibilityFeatureEnabled= */ true)
+                        /* eligibilityFeatureEnabled= */ true,
```
avaleske, line 960: Should we have test cases where one is true and the other is false? Or is that not a possible condition?

avaleske (approved):
LGTM! I just had some clarifying questions in a few spots.

server/app/views/applicant/ProgramIndexView.java
reemaka: Lol that's how I had it originally but then thought maybe that made it more confusing. Helpful feedback that that is in fact easier to parse :)

server/app/views/applicant/ProgramIndexView.java
reemaka: Yes, when eligibility is gating, we always show all tags.

server/test/services/applicant/ApplicantServiceTest.java
reemaka: There were no existing tests for the eligibility flag, but sure added a couple of tests to make sure both flags are being respected.

Issue #4280, opened 2023-02-24
Tracking: Refactor auth package


michaelzetune:
General tracking issue for refactoring the `app/auth` package. Improvements include but are not limited to:

* Renaming classes and other symbols to better represent their use case
* Restructuring the inheritance hierarchy and using other architectural constructs to more clearly represent the relationships and reuse of logic between classes.
* Factoring out common, redundant, or misplaced logic across files.

MichaelZetune:
Added back the tech-debt tag because I think the current state of the auth package is convoluted and hard to understand, which is something I consider to be tech debt. Nick, feel free to reply if you disagree.

michaelzetune:
Closing this as the issue is relatively subjective, several PRs have been put out for it, and it's been a while since it's been touched.

PR #4343, opened 2023-03-08
Remove state generator param from CiviformOidcLogoutActionBuilder

MichaelZetune:
### Description

Remove state generator param from `CiviformOidcLogoutActionBuilder`. We already have it in the `OidcConfiguration`.

### Issue(s) this completes

Progress towards #4280

### DIFF
--- a/server/app/auth/oidc/applicant/LoginGovClientProvider.java
+++ b/server/app/auth/oidc/applicant/LoginGovClientProvider.java
@@ -99,7 +99,6 @@ public final class LoginGovClientProvider extends GenericOidcClientProvider {
 
     // Apply logout settings specific to login.gov
     ((CiviformOidcLogoutActionBuilder) client.getLogoutActionBuilder())
-        .enableStateGenerator()
         // See https://developers.login.gov/oidc/#logout-request
         .setExtraParams(ImmutableMap.of("client_id", getClientID()));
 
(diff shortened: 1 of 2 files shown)


server/app/auth/oidc/applicant/LoginGovClientProvider.java
```diff
@@ -99,7 +99,7 @@ public OidcClient get() {
```
MichaelZetune, line 99: We set a state generator on line 86. This was originally duplicately set on 102.

avrittrohwer: I see that setStateGenerator() is called on the config, but this seems to be called on the client.  How does the config and the client interact?

avrittrohwer: Oh I see, the config is used in civiformOidcLogoutActionBuilder.java:106

MichaelZetune: `LoginGovClientProvider` indirectly extends `OidcClientProvider`, which has the base `OidcClient get()` method.

That base method calls `getConfig()` to get the `OidcConfiguration`, which should drop down to the lowest implementation available, here.

Let me know if you disagree.

avrittrohwer (approved):
Looks good to me, only request is to see if you can break any existing tests by intentionally messing this PR up and if not, add a test that breaks if not too much effort

PR #4346, opened 2023-03-08
updates docker hub username

seanmalbert:
### Description

Moves docker hub user name into a secret after converting civiform to an organization.


avaleske (approved):

gwendolyngoetz (commented):
@seanmalbert  I'd recommend moving the username into the secrets.

gwendolyngoetz (approved):

PR #4349, opened 2023-03-09
Show message and exit if trying to build prod image on Mac

gwendolyngoetz:
### Description

Our prod image currently doesn't support arm. This causes confusion if you don't know that is the case. Adding a check and message to alert if trying to build an unsupported platform. We can remove this later when we get arm support added.

### DIFF
--- a/bin/build-prod
+++ b/bin/build-prod
@@ -15,14 +15,14 @@ readonly DOCKERFILE="prod.Dockerfile"
 readonly HOST_CPU_ARCH="$(docker info --format='{{.Architecture}}')"
 
 if [[ "${HOST_CPU_ARCH}" != "x86_64" ]]; then
-    echo "----------------------------------------------------------------"
-    echo " WARNING"
-    echo "----------------------------------------------------------------"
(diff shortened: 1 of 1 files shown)


bin/build-prod
```diff
@@ -12,6 +12,18 @@ readonly SNAPSHOT_TAG="SNAPSHOT-${SHORT_SHA}-${DATE_IN_UNIX_SECONDS}"
...
+
+if [[ "${HOST_CPU_ARCH}" != "x86_64" ]]; then
+    echo "----------------------------------------------------------------"
+    echo " WARNING"
+    echo "----------------------------------------------------------------"
+    echo " Detected ${HOST_CPU_ARCH} as host system cpu architecture."
+    echo " Currently only x86_64 is supported by our prod.Dockerfile."
+    echo "----------------------------------------------------------------"
+    echo ""
+    exit
```
bion, line 25: **suggestion:** use a non-zero exit code

bion (changes requested):

bin/build-prod
gwendolyngoetz: Good catch. Thanks!

bion (approved):

PR #4352, opened 2023-03-09
Remove use of `APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM`.

michaelzetune:
### Description

Neither Seattle nor Bloomington use `APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM` - we can remove it and have the default configuration be for `post_logout_redirect_uri`.

In general, the need for this URI to be overriden is a consequence of the identity provider, not of the civic entity's preferences, so we should do it in the auth provider code.

The only auth provider that needs to override it right now is Auth0, which overrides it to `returnTo`.

Relates to #4280.

### DIFF
--- a/server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
+++ b/server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java
@@ -46,14 +46,6 @@ public final class CiviformOidcLogoutActionBuilder extends OidcLogoutActionBuild
     this.clientId = clientId;
   }
 
-  /** Helper function for retriving values from the application.conf, */
-  private static Optional<String> getConfigurationValue(Config civiformConfiguration, String name) {
-    if (civiformConfiguration.hasPath(name)) {
-      return Optional.ofNullable(civiformConfiguration.getString(name));
-    }
-    return Optional.empty();
-  }
-
   /**
    * Sets param that contains uri that user will be redirected to after they are logged out from the
    * auth provider. In OIDC spec it should be `post_logout_redirect_uri` but some providers override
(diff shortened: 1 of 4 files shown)


MichaelZetune:
Todo(me): remove this from seattle deploy as well:

https://github.com/seattle-civiform/civiform-deploy/blob/cbb15be58faf9be4fa2e0f3fc31fd25e2af9da48/infra/containers.yaml#L233

MichaelZetune:
Obstacles to continuing:

- As mentioned above, Bloomington does not use the `APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM`.
- Seattle *does* set the `APPLICANT_OIDC_POST_LOGOUT_REDIRECT_PARAM` in their config ([source](https://github.com/seattle-civiform/civiform-deploy/blob/cbb15be58faf9be4fa2e0f3fc31fd25e2af9da48/infra/containers.yaml#L233)), but it is set to an empty string. I contacted @nbeloglazov about this and he said the following:

> Ok. So Seattle uses this url for logout: https://github.com/seattle-civiform/civiform-deploy/blob/main/infra/containers.yaml#L38. If you add post_logout_redirect_uri to it - it will break. Because JS or server-side doesn't like that extra param. So we removed it. They generally have odd setup because their logout url contains hash and pac4j doesn't like urls with hashes. I had to override some methods to bring back hash. But it feels like we are going against standard libraries when using logout url with hash. We talked with some guy, Vinay, from Seattle about their IDCS implementation because it seemed not work properly wrt to logout url. But it hasn't been resolved so I made those workarounds. Also I might be rusty on details. As auth stuff has lot's of edge cases and I probably forgot specifics. And we should have documented that in Seattle deploy yaml file

So it seems like we need to keep some version of the option to modify the post-logout redirect param **key** (not value) to something other than `post_logout_redirect_uri`. Since Seattle is the only case, it would be great if we could only expose an option to remove it, rather than an option to arbitrary change it.

That being said, this seems like an issue with IDCS, not Seattle specifically, so maybe making this modification in the `IdcsClientProvider` class instead of in the application config makes more sense, so (1) future IDCS users will get the benefit and (2) there are fewer knobs exposed to civic entities.

PR #4363, opened 2023-03-13
Adds route to get the status of a feature flag

seanmalbert:
### Description

This adds the ability to get the status of a feature flag. This will be useful for browser tests, so a test can check if a feature is enabled before running.

#### General


avrittrohwer:
@seanmalbert should we mark this as ignore-for-release?

avrittrohwer (approved):

Issue #2985, opened 2022-07-26
In bin/fmt have flags to only format subset of languages


azizova-leyla:
I frequently only want Java. (there are no changes in any of the other files). And having to wait for the rest is a time sink.

nb1701:
We do change detection and skip formatters for files that haven't changed.  Done.

reemaka:
It appears that what's happening is that the `git diff --diff-filter=d --name-only origin/main` command that `bin/fmt` uses diffs the local branch against the remote version of main. So, if you haven't pulled and rebased in a while, you get a lot of noise.

Some options:

- Change the diffbase to `main` instead of `origin/main` so that the diffbase is the local main branch.
  - This works if the main branch hasn't been pulled since the feature branch was created, but it will have noise if it's been pulled but the feature branch hasn't been rebased. 
- Allow passing in a custom value for the diffbase.
  - The user could set it to a branch, like `main`, or a commit, which would allow them to run `bin/fmt` only on, for example, the changes since their last commit.
- Do the original plan of adding flags to the formatter allowing you to pass in languages.
  - Specifying a language reduces the amount of work, but can still run into the problem of running the formatter on extra files within that language if you're not up-to-date with `origin/main`.
- Add an argument to allow specifying which files to run the formatter on.  
  - For Java and Typescript, we can then detect which formatter to run based on the file extension. As @gwendolyngoetz pointed out, we can't do this for Python and shell scripts, however, as they don't always have a file extension. In that case we could detect that there's no extension and run both formatters.

A caveat to some of these is that the formatter is used in the [format Github Action](https://sourcegraph.com/github.com/civiform/civiform/-/blob/.github/workflows/format.yaml) so we need to make sure that modifying this doesn't break the action.

A workaround is to just make sure you pull and rebase before running `bin/fmt`.

PR #4388, opened 2023-03-14
Only format python files if they have changes

avrittrohwer:
### Description

Updates formatter script to only format python files if there are changes to them.

Part of https://github.com/civiform/civiform/issues/2985.

Tested: build the formatter image with this change, ran with no python changes.  "No modified python files found" was printed.  I changed a python file and re-ran, python formatting was triggered.


nb1701 (approved):

nb1701:
Thanks for this!

Issue #4367, opened 2023-03-13
Application Confirmation Page Changes


MichaelZetune:
[Figma link](https://www.figma.com/file/TkgERgy6MlyXBmebkTVIxS/CiviForm---Login%2C-Language-and-Landing-page-improvements?node-id=1508%3A10861&t=R48xS58zBKfUdnPl-0)

Guest Mode updates include: 
* Update copy for ‘Create an account or ’ section to help an applicant understand the benefits of choosing those options. 
* Update button label from ‘all caps’ to ‘sentence case’ aka less shouting.
   * Miriam and I agreed to apply this change to all buttons on CiviForm, as there is global styling of buttons and this is actually the simpler approach anyways.
* Include login/prompt if applicant continues to apply for another program without an account

Logged in updates include: 
* Remove the ‘I’m done’ CTA.

*(double check for Figma updates before closing this issue)*

Related PRs:
* #4488
* #4391
* #4409
* #4500

Part of #4360.

![image](https://user-images.githubusercontent.com/30369272/227967464-afb65f69-a437-40de-b845-8093cc616f4b.png)

![image](https://user-images.githubusercontent.com/30369272/227967520-808d5942-9b88-441a-b2dc-7d8a23353eff.png)

![image](https://user-images.githubusercontent.com/30369272/227967620-b5ccf00f-d757-4c28-a131-028319cdcca9.png)

PR #4391, opened 2023-03-14
Update copy on application confirmation screen

MichaelZetune:
### Description

Updates the copy for application confirmation as per the mocks in #4367.

This PR updates the English, and we put a request for the other languages to be updated in the next translations batch. The other languages are there now, just with a tiny bit different phrasing.

#### General

- [x] Update English in `messages` file.
- [x] Update screenshots

### Issue(s) this completes

Progress on #4367.


server/conf/messages
```diff
@@ -328,7 +328,7 @@ content.confirmed=Thank you for applying to {0}.  Your application has been rece
 title.createAnAccount=Create an account or sign in
 
 # A message urging users to create an account to save their data.
-content.pleaseCreateAccount=If you sign into a city account now, your information will be saved and you''ll be able to return at any time to complete future applications.  If you do not already have an account, the login page will allow you to register.  You won''t lose any data you''ve already entered.
+content.pleaseCreateAccount=Log in to your City of Seattle account or create  an account to have your information saved. You'll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won''t lose any data you?ve already entered.
```
jessbumblebee, line 331: question(blocking): Is it intentional that this text is Seattle specific? I'd have expected a text that inserts the civic entity name here.

jessbumblebee: Also, should it be "you've already entered" instead of "you?ve already entered"?

server/conf/messages.en-US
```diff
@@ -61,7 +61,7 @@ content.mobileFileUploadHelp=On your phone? “Choose File” also allows you to
 content.noValidAddress=We could not verify the address that you provided. Please edit your address to continue.
 content.notEligible=Not Eligible
 content.or=or
-content.pleaseCreateAccount=If you sign into a city account now, your information will be saved and you''ll be able to return at any time to complete future applications.  If you do not already have an account, the login page will allow you to register.  You won''t lose any data you''ve already entered.
+content.pleaseCreateAccount=Log in to your City of Seattle account or create  an account to have your information saved. You'll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won''t lose any data you've already entered.
```
jessbumblebee, line 64: question(blocking): Same as above, should this be none Seattle specific?

MichaelZetune: Fixed.

server/conf/messages
```diff
@@ -322,7 +322,7 @@ content.confirmed=Thank you for applying to {0}.  Your application has been rece
 title.createAnAccount=Create an account or sign in
 
 # A message urging users to create an account to save their data.
```
jessbumblebee, line 324: Suggestion:blocking I would recommend finding out a bit more about the translation process to understand what the person who translates text gets to see (this file?). In that place I would document well what the value that is passed in represents and how it fits into the context of the sentence. The goal would be that the person who does the translations fully understands what they are translating. This is important for example in languages where the word order is different because translating in sections can result in a sentence with a completely different meaning

suggestion (optional)
Big kudos if you also want to update our process a bit to make this the standard and prevent future translation errors (could be done separately, but would be quick while you are here)

I'm thinking along the lines of adding a comment such as:

This files contains the strings used in applicant UI as they are seen by translators. To make correct translations as easy as possible, please include a description that allows someone without knowledge about where this string is being used to understand it fully. For example explain where the string is shown and if strings are deployment specific (e.g. they use the civic entities name), specify what each passed through value represents in the following format...

And then updating the comments of the  few strings where the {0} notation is already used

MichaelZetune: Confirmed with Austin that the translators do see the description. Added a sentence describing what {0} represents, and added a paragraph similar to your suggestion to the top of the file.

server/conf/messages
```diff
@@ -318,8 +322,8 @@ content.confirmed=Thank you for applying to {0}.  Your application has been rece
 # Title (not a main page title) on section prompting an applicant to create an account or sign in to save their data.
 title.createAnAccount=Create an account or sign in
 
-# A message urging users to create an account to save their data.
-content.pleaseCreateAccount=If you sign into a city account now, your information will be saved and you''ll be able to return at any time to complete future applications.  If you do not already have an account, the login page will allow you to register.  You won''t lose any data you''ve already entered.
+# Shown to applicants to encourage login; {0} represents the civic entity's name, in reference to their login portal.
+content.pleaseCreateAccount=Log in to your {0} account or create an account to have your information saved. You''ll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won''t lose any data you''ve already entered.
```
jessbumblebee, line 326:
```suggestion
content.pleaseCreateAccount=Log in to your {0} account or create an account to have your information saved. You'll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won't lose any data you've already entered.
```

server/conf/messages.en-US
```diff
@@ -59,7 +59,7 @@ content.mobileFileUploadHelp=On your phone? “Choose File” also allows you to
 content.noValidAddress=We could not verify the address that you provided. Please edit your address to continue.
 content.notEligible=Not Eligible
 content.or=or
-content.pleaseCreateAccount=If you sign into a city account now, your information will be saved and you''ll be able to return at any time to complete future applications.  If you do not already have an account, the login page will allow you to register.  You won''t lose any data you''ve already entered.
+content.pleaseCreateAccount=Log in to your {0} account or create an account to have your information saved. You''ll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won''t lose any data you''ve already entered.
```
jessbumblebee, line 62:
```suggestion
content.pleaseCreateAccount=Log in to your {0} account or create an account to have your information saved. You'll be able to return at any time to complete future applications. If you do not already have an account, the login page will allow you to sign up and you won't lose any data you've already entered.
```

jessbumblebee (approved):

server/conf/messages
MichaelZetune: We need the double apostrophes:
https://github.com/civiform/civiform/blob/8b443012c970ab8c46816548906f81b39869155b/server/conf/messages#L5

server/conf/messages.en-US
MichaelZetune: We need the double apostrophes:
https://github.com/civiform/civiform/blob/8b443012c970ab8c46816548906f81b39869155b/server/conf/messages#L5

PR #4396, opened 2023-03-14
Fixes for bin/run-browser-tests-local

nb1701:
This fixes several issues found with running `bin/run-browser-tests-local`.

1. Adds a `--local` flag to `bin/run-browser-test-env`, which sets `BASE_URL` to `http://localhost:9999`, which the local browser tests expect. Otherwise, upon logout, it will attempt to go to `civiform:9000` instead and fail. It also includes an additional compose file that exposes port 3390, as local tests will need to be able to talk to dev-oidc directly for redirects. Because this will conflict with `bin/run-dev`, this is not enabled by default.

2. Sets `LOCALSTACK_URL` to point to port 6645 in `bin/run-browser-tests-local`, which is what is specified in `browser-test-compose.yml`.

3. Adds the flag for starting the JVM debugger on port 9457 to allow connecting the debugger during browser tests.

4. Updates the `test_oidc_provider.js` file to add the paths for `localhost:9999` as allowed redirect URIs.

#### General

### DIFF
--- a/bin/run-browser-test-env
+++ b/bin/run-browser-test-env
@@ -13,6 +13,11 @@ truth::declare_var_false continuous_integration
 # tests also require localstack for SES email sending.
 truth::declare_var_false both_emulators
 
+# Default to not running browser tests in local mode, which
+# exposes port 3390 of the OIDC container, as this may conflict
+# with other concurrently running stacks.
+truth::declare_var_false local
+
 # Default to using Localstack emulator.
 emulators::set_localstack_emulator_vars
 
@@ -29,6 +34,8 @@ emulators::set_localstack_emulator_vars
 #   1: "$@" - full args array for the script
 #######################################
 function set_args() {
+  # Default for browser tests running in Docker
+  export BASE_URL="http://civiform:9000"
   while [ "${1:-}" != "" ]; do
     case "$1" in
       "--help")
@@ -47,6 +54,9 @@ Options:
 
 --azure Use the Azure emulator (Azurite) for simulating cloud storage for file
         upload and download. Also use localstack, since we need that for SES.
+
+--local Set BASE_URL to http://localhost:9999 to enable running
+        bin/run-browser-tests-local
 EOF
         exit 0
         ;;
@@ -69,6 +79,11 @@ EOF
         emulators::ensure_only_one_cloud_provider_flag aws
         # Already defaulted to AWS.
         ;;
+
+      "--local")
+        export BASE_URL="http://localhost:9999"
+        truth::enable local
+        ;;
     esac
 
     shift
@@ -85,8 +100,13 @@ function compose_browser_test() {
   local use_dev_env="$1"
   shift 1
   if truth::is_enabled use_dev_env; then
-    echo "docker::compose_browser_test_dev" "$@"
-    docker::compose_browser_test_dev "$@"
+    if truth::is_enabled local; then
+      echo "docker::compose_browser_test_dev_local"
+      docker::compose_browser_test_dev_local "$@"
+    else
+      echo "docker::compose_browser_test_dev" "$@"
+      docker::compose_browser_test_dev "$@"
+    fi
   else
     echo "docker::compose_browser_test" "$@"
     docker::compose_browser_test "$@"
--- a/browser-test/browser-test-compose.yml
+++ b/browser-test/browser-test-compose.yml
@@ -25,9 +25,9 @@ services:
       - IDCS_DISCOVERY_URI=http://dev-oidc:3390/.well-known/openid-configuration
       - APPLICANT_REGISTER_URI=http://dev-oidc:3390/
       - DB_JDBC_STRING=jdbc:postgresql://database:5432/postgres
-      - BASE_URL=http://civiform:9000
+      - BASE_URL=${BASE_URL:-http://civiform:9000}
       - LOCALSTACK_URL=http://localhost.localstack.cloud:4566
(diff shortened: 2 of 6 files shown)


nb1701:
Still need to update docs

bin/run-browser-test-env
```diff
@@ -47,6 +54,9 @@ Options:
 
 --azure Use the Azure emulator (Azurite) for simulating cloud storage for file
         upload and download. Also use localstack, since we need that for SES.
+
+--local Set BASE_URL to http://localhost:9999 to enable running
```
MichaelZetune, line 58: `local_mode`?

MichaelZetune, line 58: Is it possible for `--local` to be `--local_mode` to match? Or do they necessarily have to be different?

bin/run-browser-test-env
```diff
@@ -85,8 +100,13 @@ function compose_browser_test() {
   local use_dev_env="$1"
   shift 1
   if truth::is_enabled use_dev_env; then
-    echo "docker::compose_browser_test_dev" "$@"
-    docker::compose_browser_test_dev "$@"
+    if truth::is_enabled local_mode; then
+      echo "docker::compose_browser_test_dev_local"
```
MichaelZetune, line 104: Add the `"$@"` here like in line 107

bin/run-browser-test-env
nb1701: No, they could be the same.  I'm not sure why I chose `local_mode` for the var name elsewhere :)

bin/run-browser-test-env
nb1701: Oops, good catch!

bin/run-browser-test-env
```diff
@@ -47,6 +54,9 @@ Options:
 
 --azure Use the Azure emulator (Azurite) for simulating cloud storage for file
         upload and download. Also use localstack, since we need that for SES.
+
+--local Set BASE_URL to http://localhost:9999 to enable running
+        bin/run-browser-tests-local
```
MichaelZetune, line 59: Suggestion: Since we have `bin/run-browser-tests` and bin/run-browser-tests-local`, can we have a `bin/run-browser-test-env-local` as well? Then we can remove the `--local` arg, set `BASE_URL` in `bin/run-browser-test-env-local`, and call `bin/run-browser-test-env` from that new file. What do you think?

bin/run-browser-test-env
MichaelZetune: Would you make that change? Unless you agree with my suggestion below.

bin/run-browser-test-env
nb1701: We could, but I think we'd end up duplicating a lot of the code in here?

nb1701: Sorry, didn't fully grok what you said at first.  Yeah, that's a good point, that might be cleaner.

browser-test/browser-test-compose.yml
```diff
@@ -25,9 +25,9 @@ services:
       - IDCS_DISCOVERY_URI=http://dev-oidc:3390/.well-known/openid-configuration
       - APPLICANT_REGISTER_URI=http://dev-oidc:3390/
       - DB_JDBC_STRING=jdbc:postgresql://database:5432/postgres
-      - BASE_URL=http://civiform:9000
+      - BASE_URL=${BASE_URL:-http://civiform:9000}
```
MichaelZetune, line 28: For my own understanding, why does `BASE_URL` get declared twice here?

nb1701: This lets us override the `BASE_URL` env var in the container.  If it's not set, then we default it to `http://civiform:9000`.  This lets us set BASE_URL to something else (`http://localhost:9999`) for the local browser tests.

nb1701: I think to split it out into a separate file, we will either need to pass a flag into `run-browser-test-env` anyway to tell it to run in local mode or add something to `truth.sh` to detect when a variable is set and not exit, as we'll be wanting to set a variable in `run-browser-test-env-local`.  What do you think?

bin/run-browser-test-env
nb1701: I think to split it out into a separate file, we will either need to pass a flag into run-browser-test-env anyway to tell it to run in local mode or add something to truth.sh to detect when a variable is set and not exit, as we'll be wanting to set a variable in run-browser-test-env-local. What do you think?

MichaelZetune: I see - if we would need the `local_mode` flag regardless, then I'll leave it up to you. I guess it's not so different to run `bin/run-browser-test-env --local` vs `bin/run-browser-test-env-local`

MichaelZetune:
Approved pending the unresolved comments and a quick test to make sure it's working after those changes. Thanks Nick!

MichaelZetune (approved):
Approved pending the unresolved comments and a quick test to make sure it's working after those changes. Thanks Nick!

bin/run-browser-test-env
nb1701: Alright.  I think for now I'll just leave it.  Suppose we could always have `bin/run-browser-test-env-local` that calls `bin/run-browser-test-env --local` :)

nb1701:
Tested manually (with a locally rebuilt dev-oidc container) with the latest changes, everything still works 🎉

nb1701:
Docs update: https://github.com/civiform/docs/pull/262

Issue #4270, opened 2023-02-22
Specific CiviForm version number not shown on login page when using 'latest' version


rasmi:
#4171 enables showing the active CiviForm version on the login page. When deploying with tag `'latest'` and `SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE=true`, CiviForm displays the verstion as "latest" rathter than the version number. 

![Screenshot 2023-02-22 16 59 07](https://user-images.githubusercontent.com/2267370/220769951-fc9225a8-d507-49e5-b842-df53cf26d1a7.png)

avrittrohwer:
@rasmi what is the desired behavior?  When CIVIFORM_VERSION is set to 'latest', the docker image tagged 'latest' is pulled.  https://hub.docker.com/layers/civiform/civiform/latest/images/sha256-180920fd9e59644bf81817ae8fcb75d35ff79e21054382cef50aa51be76f4c6b?context=explore for example.  Is having "CiviForm version: SNAPSHOT-fd0aaa0-1677605688" the desired behavior?

rasmi:
I see, I did not realize every commit is pushed to docker and that `latest` pointed to non-release version. 

I think if the code coincides with a release/version number, that version number should be shown. Otherwise, commit tag (if there is no clear release/version number) would be helpful. Ideally this happens even if someone uses `latest` in the deployment script. 

Maybe a longer-term improvement is to change how we use tags in Docker so that `latest` always pulls the most recent release version, and we use some other tag for 'give me the most recent snapshot of `main`'. For me, it is convenient to always use 'latest' because I can ideally just re-run the script with no changes to update my release -- but with the expectation that I'm getting the latest official release.

This is not important right now IMO, we should just re-emphasize with users that the recommended way to deploy is by specifying a specific version number (and maybe warn/ ask for confirmation if they don't use a specific version).

avrittrohwer:
Implementing the above comment is a small change, got it working in my dev deployment: 
![image](https://user-images.githubusercontent.com/17995083/225446998-c7966fbf-3b9b-42d1-aa38-887174453797.png)

rasmi:
That looks great! My impression is that no one except developers/IT staff will want to see the snapshot on the login page, so maybe we can only display that on admin page footers instead. WDYT? 

[...]

PR #4405, opened 2023-03-15
Show civiform image tag if CIVIFORM_VERSION is set to "latest"

avrittrohwer:
### Description

Deployments set CIVIFORM_VERSION in their civiform_config.sh file.  The value can be any valid docker tag, like "latest", "v1.18.0", or "SNAPSHOT-3af89974-1678907834".  Currently, if SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE is true, the CIVIFORM_VERSION value is directly show.  This change makes it so that if the value is "latest", the snapshot version like "SNAPSHOT-3af89974-1678907834" is shown.

#### General

#### User visible changes

### Issue(s) this completes

Part of https://github.com/civiform/civiform/issues/4270

### DIFF
--- a/server/app/views/LoginForm.java
+++ b/server/app/views/LoginForm.java
@@ -32,6 +32,7 @@ public class LoginForm extends BaseHtmlView {
 
   private final BaseHtmlLayout layout;
   private final String civiformImageTag;
+  private final String civiformVersion;
   private final boolean isDevOrStaging;
   private final boolean disableDemoModeLogins;
   private final boolean disableApplicantGuestLogin;
@@ -53,6 +54,7 @@ public class LoginForm extends BaseHtmlView {
     checkNotNull(deploymentType);
 
     this.civiformImageTag = config.getString("civiform_image_tag");
+    this.civiformVersion = config.getString("civiform_version");
     this.isDevOrStaging = deploymentType.isDevOrStaging();
     this.disableDemoModeLogins =
         this.isDevOrStaging && config.getBoolean("staging_disable_demo_mode_logins");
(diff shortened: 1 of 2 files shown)


server/conf/application.conf
```diff
@@ -480,9 +480,14 @@ ebean.default = "models.*"
 civiform.time.zoneid="America/Los_Angeles"
 civiform.time.zoneid=${?CIVIFORM_TIME_ZONE_ID}
 
+# The value of SNAPSHOT_TAG, set in bin/build-prod. For example: SNAPSHOT-3af8997-1678895722.
 civiform_image_tag = "dev"
 civiform_image_tag = ${?CIVIFORM_IMAGE_TAG}
 
+# The civiform version being deployed, can be "latest" or a version like "v1.18.0".
+civiform_version = "dev"
```
rockycodes, line 488: what does `dev` mean in this context?

rockycodes (approved):
Looks great! Thanks for the well formed comments in `LoginForm.java`. I just have one clarifying question for my own understanding.

server/conf/application.conf
avrittrohwer: Updated

server/app/views/LoginForm.java
```diff
@@ -168,7 +170,14 @@ private DivTag mainContent(Http.Request request, Messages messages) {
                 "text-base")
             .with(p(adminPrompt).with(text(" ")).with(adminLink(messages)));
     if (featureFlags.showCiviformImageTagOnLandingPage(request)) {
```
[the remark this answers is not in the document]
avrittrohwer: This feature is guarded by the SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE which is default false.  We do have a test case with the feature enabled: https://github.com/civiform/civiform/blob/main/browser-test/src/landing_page.test.ts#L49

PR #4409, opened 2023-03-16
Remove shouting text in buttons

MichaelZetune:
### Description

In new mocks, we remove shouting buttons (buttons with ALL CAPS text) on some pages. To accomplish this, we can simply remove the `uppercase` attribute from the global styling for buttons. See #4367.

## Release notes

Replace shouting text (ALL CAPS) with normal text.

#### General

- [x] Remove the `uppercase` attribute from the global styling for buttons
- [x] Update screenshot goldens

### Issue(s) this completes

Relates to #4367.


rockycodes (approved):
looks great! much nicer without all the shouting 🙂

PR #4411, opened 2023-03-16
Adds fake Esri client

seanmalbert:
### Description

Per the [Address Correction and Service Area Validation Testing Strategy](https://docs.google.com/document/d/1_rXidQeZ2bL0rg5hDEPJmFazVsdQjsuPnXZ1xINXDAU/edit#) doc, this adds a fake esri client to use for dev and testing. This does not cover adding a mock service to simulate external IO.

To test in development you will need to enable the address correction and service area validation features. In admin, you will also have to enable address correction for an address question and add service area eligiblity on the block.

When you get to the applicant address question, there are four options you can enter for the address value. The other field values do not matter.
1. `Legit Address` - this will provide three suggestions for tesing service area validation:
    1. `Address In Area`
    2. `Address With No Service Area Features`
    3. `Address Not In Area`
2. `Bogus Address` - Use this for testing address that won't return any suggestions. This is likely not to happen in the real world, but you'll see that this isn't handled properly at the moment (and won't be handled in this PR).
3. `Error Address` - Use this to test what would happen if the external Esri service were to return a non 200 status. The end result is the same as `Bogus Address`.
4. Any other address will have the same effect as Bogus and Error.


server/app/services/geo/esri/FakeEsriClient.java
avrittrohwer, line 1: Could you add the known address input values in the comments here?

server/app/services/applicant/ApplicantService.java
```diff
@@ -107,7 +109,8 @@ public ApplicantService(
       HttpExecutionContext httpExecutionContext,
       DeploymentType deploymentType,
       ServiceAreaUpdateResolver serviceAreaUpdateResolver,
-      EsriClient esriClient) {
+      EsriClient esriClient,
```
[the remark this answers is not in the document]
avrittrohwer: I did some looking into this and I think this can be achieved by adding a new Esri module: https://www.playframework.com/documentation/2.8.x/JavaDependencyInjection#Programmatic-bindings

For example, the file storage provider strategy implementation is provided based on a value in the config: https://github.com/civiform/civiform/blob/main/server/app/modules/CloudStorageModule.java.  We could provider the FakeEsriClient if DeploymentType.isDevOrStaging() is true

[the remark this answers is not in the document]
gwendolyngoetz: Building on what Avritt said, rather than inject both EsriClient and FakeEsriClient I think we should do one of the following:
* Create a [custom Guice module](https://www.playframework.com/documentation/2.8.x/JavaDependencyInjection#Programmatic-bindings) to allow the dependency injection framework to inject the appropriate type (recommended). This will allow just having to inject the one interface.
* Create a factory that can determine the appropriate object to use, and inject the factory.

gwendolyngoetz (changes requested):

server/app/services/applicant/ApplicantService.java
```diff
@@ -126,7 +129,11 @@ public ApplicantService(
         checkNotNull(configuration).getString("staging_ti_notification_mailing_list");
     this.stagingApplicantNotificationMailingList =
         checkNotNull(configuration).getString("staging_applicant_notification_mailing_list");
-    this.esriClient = checkNotNull(esriClient);
+
+    this.esriClient =
```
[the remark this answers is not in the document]
seanmalbert: I didn't end up using DeploymentType, since we want to be able to use the real esri client in staging.

server/app/services/geo/esri/FakeEsriClient.java
seanmalbert: added options in comment for `fetchAddressSuggestions`.

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
seanmalbert: Thanks for the great suggestion @avrittrohwer and @gwendolyngoetz! Added an EsriModule and EsriClient is now an abstract class. Went with a class instead of an interface for EsriClient, since `getAddressSuggestions` and `getServiceAreaInclusionGroup` methods are the same between RealEsriClient and FakeEsriClient.

seanmalbert:
Latest main has bugs, so waiting until after that's fixed to resolve merge conflict.

server/app/services/applicant/ApplicantService.java
```diff
@@ -1260,7 +1260,9 @@ public ApplicantQuestion getFirstAddressCorrectionEnabledApplicantQuestion(Block
   public CompletionStage<AddressSuggestionGroup> getAddressSuggestionGroup(Block block) {
     ApplicantQuestion applicantQuestion = getFirstAddressCorrectionEnabledApplicantQuestion(block);
     AddressQuestion addressQuestion = applicantQuestion.createAddressQuestion();
-
+    // TODO:Issue #4441 - this shouldn't throw if no suggestions are empty and CF should let the
+    // user continue on
+    // with the application
```
nb1701, line 1265: I think the formatter mangled this comment

server/app/services/geo/esri/EsriClient.java
```diff
@@ -180,9 +75,9 @@ public CompletionStage<Optional<AddressSuggestionGroup>> getAddressSuggestions(A
         .thenApply(
             (maybeJson) -> {
               if (maybeJson.isEmpty()) {
-                logger.error(
-                    "EsriClient.fetchAddressSuggestions JSON response is empty. Called by"
-                        + " EsriClient.getAddressSuggestions. Address = {}",
+                this.logger.error(
```
gwendolyngoetz, line 78: Nitpick: Remove `this.`

server/app/services/geo/esri/EsriClient.java
```diff
@@ -180,9 +75,9 @@ public CompletionStage<Optional<AddressSuggestionGroup>> getAddressSuggestions(A
...
+                    "RealEsriClient.fetchAddressSuggestions JSON response is empty. Called by"
+                        + " RealEsriClient.getAddressSuggestions. Address = {}",
```
gwendolyngoetz, line 80: I think the logged message should still use `EsriClient` and not `RealEsriClient` since it's within the EsriClient. If this is specific to the Real client we should make it abstract and move this into there.

server/app/services/geo/esri/EsriClient.java
```diff
@@ -309,8 +177,8 @@ public CompletionStage<ImmutableList<ServiceAreaInclusion>> getServiceAreaInclus
             (maybeJson) -> {
               if (maybeJson.isEmpty()) {
                 logger.error(
-                    "EsriClient.fetchServiceAreaFeatures JSON response is empty. Called by"
-                        + " EsriClient.isAddressLocationInServiceArea."
+                    "RealEsriClient.fetchServiceAreaFeatures JSON response is empty. Called by"
+                        + " RealEsriClient.isAddressLocationInServiceArea."
```
gwendolyngoetz, line 181: Same as above comments on the logger error message about `RealEsriClient`

server/app/services/geo/esri/EsriClient.java
seanmalbert: updated

server/app/services/geo/esri/EsriClient.java
seanmalbert: Thanks for catching this. I think I had an iteration where this was just in the Real one and didn't change this back.

server/app/services/geo/esri/EsriClient.java
seanmalbert: updated

gwendolyngoetz (approved):

server/app/modules/EsriModule.java
```diff
@@ -0,0 +1,42 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.inject.AbstractModule;
+import com.typesafe.config.Config;
+import java.net.URI;
+import play.Environment;
+import services.geo.esri.EsriClient;
+
+public class EsriModule extends AbstractModule {
```
bion, line 12: **suggestion (blocking):** please add javadoc

server/app/modules/EsriModule.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public EsriModule(Environment environment, Config config) {
+    this.environment = checkNotNull(environment);
+    this.config = checkNotNull(config);
+  }
+
+  @Override
+  protected void configure() {
+    String host = URI.create(config.getString("base_url")).getHost();
+    ImmutableSet<String> acceptedHosts = ImmutableSet.of("localhost", "civiform");
```
bion, line 27: **suggestion (blocking):** extract to constant

server/app/modules/EsriModule.java
```diff
@@ -0,0 +1,42 @@
...
+    boolean useFakeEsriClient =
+        acceptedHosts.stream()
+            .anyMatch(
+                acceptedHost -> host.equals(acceptedHost) || host.startsWith(acceptedHost + ":"));
+    String className =
+        useFakeEsriClient ? FAKE_ESRI_CLIENT_CLASS_NAME : REAL_ESRI_CLIENT_CLASS_NAME;
```
bion, line 33: **suggestion (blocking):** log which one is being used

server/app/modules/EsriModule.java
```diff
@@ -0,0 +1,42 @@
...
+    try {
+      Class<? extends EsriClient> bindingClass =
+          environment.classLoader().loadClass(className).asSubclass(EsriClient.class);
+      bind(EsriClient.class).to(bindingClass);
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException(String.format("Failed to load esri client class: %s", className));
```
bion, line 39: **suggestion (blocking):** 
```suggestion
      throw new RuntimeException(String.format("Failed to load esri client class: %s", className), e);
```

server/app/services/geo/esri/FakeEsriClient.java
```diff
@@ -0,0 +1,155 @@
...
+import org.slf4j.LoggerFactory;
+import play.libs.Json;
+import services.AddressField;
+import services.geo.AddressLocation;
+
+/**
+ * Provides overide methods for {@link EsriClient} to illiminate the need for calling the external
+ * Esri services for dev and testing.
+ */
+public class FakeEsriClient extends EsriClient {
```
bion, line 23: **nitpick:** 
```suggestion
public final class FakeEsriClient extends EsriClient {
```

server/app/services/geo/esri/RealEsriClient.java
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * <p>@see <a
+ * href="https://developers.arcgis.com/rest/geocode/api-reference/geocoding-find-address-candidates.htm">Find
+ * Address Candidates</a>
+ *
+ * @see <a
+ *     href="https://developers.arcgis.com/rest/services-reference/enterprise/query-feature-service-layer-.htm">Query
+ *     (Map Service/Layer)</a>
+ */
+public class RealEsriClient extends EsriClient implements WSBodyReadables, WSBodyWritables {
```
bion, line 37: **nitpick:** 
```suggestion
public final class RealEsriClient extends EsriClient implements WSBodyReadables, WSBodyWritables {
```

bion (changes requested):

server/app/services/geo/esri/FakeEsriClient.java
```diff
@@ -0,0 +1,155 @@
...
+ * Provides overide methods for {@link EsriClient} to illiminate the need for calling the external
+ * Esri services for dev and testing.
+ */
+public final class FakeEsriClient extends EsriClient {
+  private String jsonResources = System.getProperty("user.dir") + "/test/resources/esri/";
+  private JsonNode addressCandidates;
+  private JsonNode noAddressCandidates;
+  private JsonNode serviceAreaFeatures;
+  private JsonNode serviceAreaNoFeatures;
+  private JsonNode serviceAreaNotInArea;
```
bion, line 29: **issue (blocking):** Ensure these members won't be reused between requests.

By default, Play uses dependency injection to instantiate an instance of the controller needed to handle a request, and then reuses that instance for subsequent requests. Dependencies of that controller (and therefore their dependencies) will similarly be reused. There are two options for changing this behavior, one is to inject a `Provider<YouClass>` and call `get` on it to force creation of a new instance. The other is to prefix the route with an `@` in the routes file, forcing Play to inject a `Provider<YourController>` in its generated router class.

More info here: https://www.playframework.com/documentation/2.8.x/JavaDependencyInjection#Dependency-injecting-controllers

server/app/services/geo/esri/FakeEsriClient.java
```diff
@@ -0,0 +1,155 @@
...
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
bion, line 30: **nitpick:** 
```suggestion
  private final Logger LOGGER = LoggerFactory.getLogger(this.getClass());
```

server/app/services/geo/esri/FakeEsriClient.java
```diff
@@ -0,0 +1,155 @@
...
+          resource = new File(jsonResources + "findAddressCandidatesNoCandidates.json");
+          inputStream = new FileInputStream(resource);
+          noAddressCandidates = Json.parse(inputStream);
+          maybeJson = Optional.of(noAddressCandidates);
+        } catch (FileNotFoundException e) {
+          logger.error("fakeEsriClient fetchAddressSuggestions file not found: {}", e);
+        }
+        break;
+      case "Error Address":
+      default:
```
bion, line 92: **suggestion (blocking):** at least log an error here, consider throwing an exception as well to signal the error state to the caller

bion (changes requested):

server/app/services/geo/esri/FakeEsriClient.java
nb1701: If I'm reading this correctly (and I may not be), I think it's okay for these to be reused.  They will contain the contents of the json files that are used for the mock data, and it's just a caching mechanism for that.  None of that data will change.

bion (approved):

Issue #4366, opened 2023-03-13
Login Page Changes (LLL Phase 1)


MichaelZetune:
[Figma link](https://www.figma.com/file/TkgERgy6MlyXBmebkTVIxS/CiviForm---Login%2C-Language-and-Landing-page-improvements?node-id=1507%3A9757&t=R48xS58zBKfUdnPl-0)

**UPDATE: these changes will not be made directly on the current login page anymore; the log in page as we know it is going away. However, we will use a prompt in #4369 that will either be simple (more likely) or a port of this one. See the screenshots on that issue. For now, this issue is no longer part of Phase 1**

Seattle has agreed on v2 of the options on Figma. For mobile, try adding tooltip if feasible; we can add the text directly if need be instead.

Updates include: 
* Adding copy below ‘[City] CiviForm’ to help ground an applicant as to what CiviForm is and reassure them that they’re in the correct place. 
* Include copy for ‘Log in’ and ‘Create an account’ sections to help an applicant understand the benefits of choosing those options. 
* Update admin login copy to explain who should be using that link.
* Update button label from ‘all caps’ to ‘sentence case’ aka less shouting.
* Change guest button to link to underemphasize ‘guest’ which will highlight ‘login’ and ‘create an account’ options
* Add divider between ‘login’ and ‘create an account’ sections

Related PRs:

* #4425
* #4435
* #4434

Part of #4360.

*(double check for Figma updates before starting and before closing this issue)*

![image](https://user-images.githubusercontent.com/30369272/225749852-6602d2b6-9e91-4aa3-a737-9cae3702a737.png)

PR #4425, opened 2023-03-17
Refactor `LoginForm.java`

MichaelZetune:
### Description

This PR just cleans up `LoginForm.java` in preparation for larger frontend changes to this file coming up in #4366. There are no visual changes, as proven by the fact that `src/landing_page.test.ts` still passes.

### Issue(s) this completes

Relates to #4366.

### DIFF
--- a/server/app/views/LoginForm.java
+++ b/server/app/views/LoginForm.java
@@ -19,6 +19,7 @@ import featureflags.FeatureFlags;
 import j2html.tags.specialized.ATag;
 import j2html.tags.specialized.ButtonTag;
 import j2html.tags.specialized.DivTag;
+import j2html.tags.specialized.ImgTag;
 import java.util.Optional;
 import play.i18n.Messages;
 import play.mvc.Http;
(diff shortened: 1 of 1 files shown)


server/app/views/LoginForm.java
```diff
@@ -75,7 +76,22 @@ public Content render(Http.Request request, Messages messages, Optional<String>
     String title = messages.at(MessageKey.TITLE_LOGIN.getKeyName());
 
     HtmlBundle htmlBundle = this.layout.getBundle().setTitle(title);
-    htmlBundle.addMainContent(mainContent(request, messages));
+
+    DivTag content = div().withClasses(BaseStyles.LOGIN_PAGE);
+
+    content.with(logo());
+
```
swatkat1, line 83: Can these changes be tested with a browser tests? Even a screenshot tests would be helpful.

swatkat1: Just read the description again, so no visual changes. In that case, I think it's alright to check-in.

swatkat1 (approved):

server/app/views/LoginForm.java
MichaelZetune: Will do

PR #4428, opened 2023-03-17
Add age-related operators for date question predicates

### LINKED ISSUE #4329: Age-related operators for date question predicates
## Overview

Admins need to be able to specify predicates that operate on date questions as ages. For example, "applicant must be younger than 18". The immediate use case for this is WIOA applications at Arkansas that require applicants to some programs be between 14 (inclusive) and 25 (exlusive) years old.

## Done when

- [ ] admin can configure date question predicates where the date is "younger than" a given number of years
- [ ] admin can configure date question predicates where the date is "older than" a given number of years
- [ ] (maybe? PM discuss) admin can configure date question predicates where the date is "between" a given number of years old

## Implementation details

Recommendation from @bion:

1. Add new [Operator](https://github.com/civiform/civiform/blob/main/server/app/services/program/predicate/Operator.java) instances for `AGE_YOUNGER_THAN`, `AGE_OLDER_THAN`, `AGE_BETWEEN`.
1. Update the admin predicate configuration view to use a `number`  input for value when one of those above operators is selected
1. Add custom implementations of `com.jayway.jsonpath.Predicate` that read the `LocalDate` and perform the comparison ([see docs](https://github.com/json-path/JsonPath/))
1. Add an optional field for a custom predicate to [JsonPathPredicate](https://github.com/civiform/civiform/blob/main/server/app/services/applicant/predicate/JsonPathPredicate.java)
1. Populate that field with an instance of the custom predicate in [JsonPathPredicateGenerator](https://github.com/civiform/civiform/blob/main/server/app/services/applicant/predicate/JsonPathPredicateGenerator.java)
1. Update [CfJsonDocumentContext.evalPredicate](https://github.com/civiform/civiform/blob/main/server/app/services/CfJsonDocumentContext.java#L529) to use the custom jsonpath predicate if it is present

dkatzz:
### Description

Add more age related operators for date question predicates so that:
- admin can configure date question predicates where the date is "younger than" a given number of years
- admin can configure date question predicates where the date is "older than" a given number of years
- admin can configure date question predicates where the date is "between" a given number of years old

This is somewhat different than the implementation described [here](https://github.com/civiform/civiform/issues/4329), but I had trouble setting up a new predicate type and this seems like it works.

- [x] use app zone ID instead of system default
- [x] handle bad input

## Release notes

Add "age older than", "age younger than" and "age between" options for eligibility predicates on date questions.

#### General

### Issue(s) this completes

Fixes #4329


server/app/services/DateConverter.java
```diff
@@ -101,4 +104,9 @@ public String renderAsTwoDigitMonthAndYear(Timestamp timestamp) {
 
     return (calendar.get(Calendar.MONTH) + 1) + "/" + calendar.get(Calendar.YEAR);
   }
+
+  /** Gets the {@link Long} timestamp from an age, by subtracting the age from today's date. */
+  public long getDateTimestampFromAge(Long age) {
+    return LocalDate.now(clock).minusYears(age).atStartOfDay(zoneId).toInstant().toEpochMilli();
```
avaleske, line 110: tk what does this do on your birthday?

server/app/views/admin/programs/ProgramBlockPredicateConfigureView.java
```diff
@@ -745,7 +745,17 @@ private DivTag createValueField(
                     "hidden",
                     "text-xs",
                     BaseStyles.FORM_LABEL_TEXT_COLOR)
-                .withText("Enter a list of comma-separated values. For example, \"v1,v2,v3\"."));
+                .withText("Enter a list of comma-separated values. For example, \"v1,v2,v3\"."))
```
avaleske, line 748: _very_ nit here and 756: This file seems to _mostly_ use `div("text").withClasses()` instead of `div().withClasses().withText()`

server/app/views/admin/programs/ProgramBlockPredicateConfigureView.java
```diff
@@ -745,7 +745,17 @@ private DivTag createValueField(
...
+        .with(
+            div()
+                .withClasses(
+                    ReferenceClasses.PREDICATE_VALUE_BETWEEN_HELP_TEXT,
+                    "hidden",
+                    "text-xs",
+                    BaseStyles.FORM_LABEL_TEXT_COLOR)
+                .withText(
+                    "Enter two comma-separated or dash-separated values. For example \"18,30\" or"
```
avaleske, line 757: Can we describe the inclusive/exclusive logic in help text somewhere? It may not be clear to admins that `BETWEEN` is not inclusive on the upper-bound, or that `AGE_GREATER_THAN` is inclusive.

avaleske (approved):
This took me like an hour and a half to review cause it's my first time reading the predicate logic 😅 

Re the tests:
- Can we add some tests that check birthdays right on the boundaries? Like if you're filling this out in the morning of your birthday?
- The presubmits say we're missing a test to cover line 303 of the `PredicateGenerator`

Some other questions:
- What is the expected behavior if someone is filling this out on their birthday in a timezone different than the app is in? Should we give more helpful error text, like "try filling this out tomorrow?"
- I asked on `DateConverter` about the the inclusive/exclusive comparison logic for each operator.

bion:
> What is the expected behavior if someone is filling this out on their birthday in a timezone different than the app is in? Should we give more helpful error text, like "try filling this out tomorrow?"

They'll get the logic for the civic entity's timezone, so it may be that there's a discrepancy for up to a day.

We don't have any predicate-specific error message at the moment, we just show a page that says "you are are ineligible because [question text]".

bion:
@avaleske 

> Can we add some tests that check birthdays right on the boundaries? Like if you're filling this out in the morning of your birthday?

Thanks for this comment, it revealed a bug.

PR #4430, opened 2023-03-17
Evaluate eligibility based on application rather than latest applicant data

### LINKED ISSUE #4414: Eligibility evaluation upon application re-submission uses latest application
**Describe the bug**
If an applicant submits an application to a program, then edits their answers and submits another application, the meets/does not meet eligibility designation in the admin UI seems to use the most recent application for all versions of the application.

**To Reproduce**

1. Publish a program with eligibility criteria. Set criteria to non-gating.
2. Log in as guest
3. Fill out the application with eligible answers and submit.
4. Click apply to more programs
5. Edit the application for the same program, updating an answer to an ineligible answer.
6. Log out and log back in as civiform and program admin.
7. View applications for the program.
8. Note that there are 2 applications, and both say "ineligible" even though the answers for the first one correctly show as set to something that should be eligible.

**Expected behavior**
One application should say eligible and the other should say ineligible.

**Screenshots**
image
Here, spring should be an eligible multiple choice answer, but because they changed to winter in their next application, they are not listed as eligible.

**Additional context**
I tried this for both date and multiple choice questions and the behavior was the same for both.

reemaka:
### Description

See #4414 for a description of the bug. The code was using applicant.applicantData() instead of application.applicantData() in some places, causing the applicant's latest answers to be evaluated for eligibility rather than their answers on the specific application.

#### General

### Issue(s) this completes

Fixes #4414

### DIFF
--- a/server/app/services/applicant/ApplicantService.java
+++ b/server/app/services/applicant/ApplicantService.java
@@ -842,10 +842,6 @@ public final class ApplicantService {
       ApplicantData applicantData, ProgramDefinition programDefinition) {
     ReadOnlyApplicantProgramService roAppProgramService =
         getReadOnlyApplicantProgramService(applicantData, programDefinition);
-    System.out.println("*** reemax");
-    System.out.println(roAppProgramService.isApplicationEligible());
-    System.out.println(roAppProgramService.isApplicationNotEligible());
-    System.out.println(applicantData);
     return programDefinition.hasEligibilityEnabled()
         ? Optional.of(!roAppProgramService.isApplicationNotEligible())
         : Optional.empty();
--- a/server/app/services/applicant/ApplicantService.java
+++ b/server/app/services/applicant/ApplicantService.java
@@ -835,13 +835,31 @@ public final class ApplicantService {
   }
 
   /**
-   * Returns whether an application is maybe eligible for a program, and empty if there are no
-   * eligibility conditions for the program.
+   * Returns whether an applicant is maybe eligible for a program based on their latest answers, and
+   * empty if there are no eligibility conditions for the program.
    */
-  public Optional<Boolean> getOptionalEligibilityStatus(
+  public Optional<Boolean> getApplicantMayBeEligibleStatus(
       ApplicantData applicantData, ProgramDefinition programDefinition) {
     ReadOnlyApplicantProgramService roAppProgramService =
         getReadOnlyApplicantProgramService(applicantData, programDefinition);
+    System.out.println("*** reemax");
+    System.out.println(roAppProgramService.isApplicationEligible());
+    System.out.println(roAppProgramService.isApplicationNotEligible());
+    System.out.println(applicantData);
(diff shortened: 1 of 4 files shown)


server/app/services/applicant/ApplicantService.java
```diff
@@ -835,13 +835,31 @@ public CompletionStage<ApplicationPrograms> relevantProgramsForApplicant(long ap
...
-   * eligibility conditions for the program.
+   * Returns whether an applicant is maybe eligible for a program based on their latest answers, and
+   * empty if there are no eligibility conditions for the program.
    */
-  public Optional<Boolean> getOptionalEligibilityStatus(
+  public Optional<Boolean> getApplicantMayBeEligibleStatus(
       ApplicantData applicantData, ProgramDefinition programDefinition) {
     ReadOnlyApplicantProgramService roAppProgramService =
         getReadOnlyApplicantProgramService(applicantData, programDefinition);
+    System.out.println("*** reemax");
```
dkatzz, line 845: **suggestion (blocking):** Reminder to remove the print statements

server/app/services/applicant/ApplicantService.java
```diff
@@ -835,13 +835,31 @@ public CompletionStage<ApplicationPrograms> relevantProgramsForApplicant(long ap
   }
 
   /**
-   * Returns whether an application is maybe eligible for a program, and empty if there are no
-   * eligibility conditions for the program.
+   * Returns whether an applicant is maybe eligible for a program based on their latest answers, and
```
dkatzz, line 838: **nitpick:** This is not necessarily whether an `applicant` may be eligible because the applicantData of the application could be passed through as well. You could have this function accept an "applicant" instead of applicantData and then do getApplicantData on the applicant to ensure this is working as expected? Thoughts?

reemaka: Good point! I think that's a good idea; one question though -- this is called in `relevantProgramsForApplicant()` with `draftApp.getApplicant().getApplicantData()`. Any thoughts on whether or not for draft applications we should use the application.applicantData or the applicant.applicantData?

dkatzz: I _think_ we should use the draftApp.getApplicant().getApplicantData(). That's because if the admin gates on eligibility < 30 and I go in as an applicant fill out `application 1` that my age is 29 then I abandon it because I'm not eligible and I go to fill out `application 2` a few months later and fill in that my age is 30, I'd like to go to the home page and know that I'm now eligible for `application 1` too, but if we look at the application.getApplicantData the application that hasn't been submitted won't be looking at my latest answer for eligibility. But I think if I opened that application it would sync with what I answered most recently (in application 2). So basically for unsubmitted applications in the applicant UI, I think it should look at the most recent applicant data, but on the admin side it should be looking at the application data. Does that make sense?

dkatzz (approved):

PR #4447, opened 2023-03-20
Remove FeatureFlag per-flag methods

MichaelZetune:
### Description

A follow-up to #4437. To further simplify the feature flag codebase and remove duplication, we remove all `getXEnabled()` style methods here.

Note we leave `isAdminReportingUiEnabled()` for a future PR as this is slightly more complex.

PR #4452, opened 2023-03-20
Remove `isAdminReportingUiEnabled()`

MichaelZetune:
### Description

A follow up to #4447 where we remove the final helper method for flags. Since we don't want to check the request for session overrides for Admin UI, we allow the request to be null in `getFlagEnabled()`.

### DIFF
--- a/server/app/featureflags/FeatureFlags.java
+++ b/server/app/featureflags/FeatureFlags.java
@@ -1,7 +1,6 @@
 package featureflags;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static featureflags.FeatureFlag.ADMIN_REPORTING_UI_ENABLED;
 
 import com.google.common.collect.ImmutableSortedMap;
 import com.typesafe.config.Config;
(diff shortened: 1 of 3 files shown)


server/app/featureflags/FeatureFlags.java
```diff
@@ -62,7 +56,7 @@ public boolean getFlagEnabled(Request request, FeatureFlag flag) {
     }
     Boolean configValue = maybeConfigValue.get();
 
-    if (!areOverridesEnabled()) {
+    if (!overridesEnabled() || request == null) {
```
bion, line 59: **suggestion (blocking):** please use `Optional` rather than null: https://docs.civiform.us/contributor-guide/developer-guide/development-standards#prefer-optional-instead-of-null

bion (changes requested):

server/app/featureflags/FeatureFlags.java
MichaelZetune: Done. Also clearly delineated the two methods while we're at it.

bion (approved):

PR #4454, opened 2023-03-20
Remove `main` branch from pr_server.yaml

MichaelZetune:
### Description

Currently we don't run CI on PRs that have a base branch. We can fix this by letting GitHub run tests on all branches, not just `main`. This is helpful for stacked changes, where you would otherwise have to wait for the first change to be merged before you get CI feedback from GitHub.

image (1)

### DIFF
--- a/.github/workflows/pr_server.yaml
+++ b/.github/workflows/pr_server.yaml
@@ -8,8 +8,6 @@ concurrency:
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


MichaelZetune:
Closing this out after discussion in sync. I misunderstood what merging the PR in the screenshot would actually do; it would only merge into the base branch, not main.

Issue #4442, opened 2023-03-20
Address Correction - Back button should go back to address question


seanmalbert:
**Describe the bug**
Clicking the back button should take me to the last page which shows the address fields. Instead it takes me back to the first page in the application which is incorrect. See [Applicant UX: Address correction dialog with 1 suggestion](https://docs.google.com/document/d/14zPxWA-TrjCmyxEgSkawYYUfDYxQPgY9DKQVQBRgfhc/edit?resourcekey=0-_QBYpeXQPsCN_ZyX_SuumA)

PR #4464, opened 2023-03-22
Address correction - previous button returns to address entry screen

rockycodes:
### Description

Added an override method for `renderPreviousButton` so that when the user clicks "Previous" it takes them back to the address entry page, and not to the previous question.

Before:

https://user-images.githubusercontent.com/24319951/226769050-a304315b-791c-4243-8b70-b2d3361286f5.mov

Now:

https://user-images.githubusercontent.com/24319951/226769068-497216d0-4237-4948-a81b-5153cb625c25.mov

#### General

### Issue(s) this completes

Fixes [#4442](https://github.com/civiform/civiform/issues/4442)

### DIFF
--- a/server/app/views/applicant/AddressCorrectionBlockView.java
+++ b/server/app/views/applicant/AddressCorrectionBlockView.java
@@ -1,12 +1,12 @@
 package views.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static j2html.TagCreator.a;
 import static j2html.TagCreator.div;
 import static j2html.TagCreator.form;
 import static j2html.TagCreator.h2;
 import static j2html.TagCreator.h3;
 import static j2html.TagCreator.label;
-import static j2html.TagCreator.a;
 
 import com.google.common.collect.ImmutableList;
 import controllers.applicant.routes;
@@ -35,7 +35,9 @@ import views.components.LinkElement;
 import views.style.ApplicantStyles;
 import views.style.StyleUtils;
 
-/** Renders a page asking the applicant to confirm their address from a list of corrected addresses. */ 
+/**
+ * Renders a page asking the applicant to confirm their address from a list of corrected addresses.
+ */
 public final class AddressCorrectionBlockView extends ApplicationBaseView {
   private static final String BLOCK_FORM_ID = "cf-block-form";
   private static final int MAX_SUGGESTIONS_TO_DISPLAY = 3;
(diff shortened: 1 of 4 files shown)


nb1701 (approved):

server/app/views/applicant/AddressCorrectionBlockView.java
```diff
@@ -221,4 +224,17 @@ private ButtonTag renderNextButton(Params params) {
         .withClasses(ApplicantStyles.BUTTON_BLOCK_NEXT)
         .withId("cf-block-submit");
   }
+
+  @Override
+  protected ATag renderPreviousButton(Params params) {
```
MichaelZetune, line 229: How hard would it be to make the address correction stage be its own block, so that we don't need this override to begin with? It seems like a code smell to me to be working around this.

If that is 100% not possible for some reason, can we do this, to avoid the duplication introduced here?

```java
  @Override
  protected ATag renderPreviousButton(Params params) {
    // Set the block index to the next block, so that the renderPreviousButton
    // method will render the correct block.
    Params newParams =
        params.toBuilder().setBlockIndex(params.blockIndex()+1).build();
    return super.renderPreviousButton(newParams);
  }
```

(which would require adding a `toBuilder()` method ([directions](https://github.com/google/auto/blob/main/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance)).

This is not tested by me, so double check my code please.

MichaelZetune (changes requested):

server/app/views/applicant/AddressCorrectionBlockView.java
rockycodes: Refactoring this to be a block is more of a lift than it's worth at this point I think (the original designs called for this to be a popup which I think is still the v2 goal), but I get your point about not wanting to repeat code so I'll try it out with calling `super`. Good suggestion, thank you!

MichaelZetune: Gotcha, would you make an issue with the tech-debt tag to look into it in the future, and tag it in the comment?

rockycodes: Will do. I'm guessing this will be part of a larger track of v2 work for address correction, so I'm waiting to hear back on that but I will update once I have that.

MichaelZetune: Cool, no need to block this PR on that.

MichaelZetune (approved):
Approved assuming you were able to run a successful test (replicating the second video in your description)

server/app/views/applicant/AddressCorrectionBlockView.java
rockycodes: It's still TBD what exactly v2 work will entail so I made a placeholder issue for this in the meantime: https://github.com/civiform/civiform/issues/4650

PR #4465, opened 2023-03-22
Add generate_markdown.py script to release automation

avrittrohwer:
### Description

Adds a python script that generates markdown from the server/config/env-var-docs.json file.

#### General

- [ ] Create a civiform-github-automation github user account with a personal access token that has write access to civiform/docs.

### DIFF
--- a/bin/env-var-docs-generate-markdown
+++ b/bin/env-var-docs-generate-markdown
@@ -1,0 +1,23 @@
+#! /usr/bin/env bash
+
+# DOC: Runs env-var-docs/generate_markdown.py.
+# DOC:
+# DOC: Writes generated markdown to stdout by default. To enable submitting to
+# DOC: the docs repository, set LOCAL_OUTPUT=false and provide values for
+# DOC: RELEASE_VERSION, GITHUB_ACCESS_TOKEN, TARGET_REPO, and TARGET_PATH.
+
+source bin/lib.sh
+
+if [[ ! -d env-var-docs/venv ]]; then
+  echo "ERROR: venv directory does not exist. First run 'bin/env-var-docs-create-venv' then run this script."
+  exit 1
+fi
+
+source env-var-docs/venv/bin/activate
+
+local_output=${LOCAL_OUTPUT:-true}
+export LOCAL_OUTPUT=${local_output}
+export ENV_VAR_DOCS_PATH=server/conf/env-var-docs.json
+
+echo "Running env-var-docs/generate_markdown.py with LOCAL_OUTPUT=${LOCAL_OUTPUT}..."
+python env-var-docs/generate_markdown.py
--- a/env-var-docs/README.md
+++ b/env-var-docs/README.md
@@ -40,11 +40,13 @@ server is deployed.
 
 ### Running locally
 
-To run check_vars_documented.py and run_regex_tests.py locally:
+Before running any scripts, a python virtual environment (venv) must be
+created: `bin/env-var-docs-create-venv`.
 
-1. First ensure a python virtual environment (venv) is created to run the
-   scripts in: `bin/env-var-docs-create-venv`.
-1. Run `bin/env-var-docs-check-vars`.
+To run check_vars_documented.py and run_regex_tests.py:
+`bin/env-var-docs-check-vars`.
+
+To run generate_markdown.py locally: `bin/env-var-docs-generate-markdown`.
 
 ## Developer setup
 
--- a/env-var-docs/generate_markdown.py
+++ b/env-var-docs/generate_markdown.py
@@ -1,0 +1,155 @@
+"""Generates markdown from an environment variable documentation file and write it to a github repository.
+
+Requires the following variables to be present in the environment:
+    ENV_VAR_DOCS_PATH: the path to env-var-docs.json.
+
+If LOCAL_OUTPUT is set, markdown will be written to stdout instead of GitHub.
+
+If LOCAL_OUTPUT is not set, the following variables must be set:
+    RELEASE_VERSION: CiviForm version being released. 
+    GITHUB_ACCESS_TOKEN: personal access token to call github API with.
+    TARGET_REPO: the target github repository in 'owner/repo-name' format.
+    TARGET_PATH: the relative path within TARGET_REPO to write the file to.
+"""
+
+import dataclasses
+import env_var_docs.parser
+import env_var_docs.errors_formatter
+import github
+import os
+import sys
+import typing
+
+
+def errorexit(msg):
+    """Logs a message and exits"""
+    sys.stderr.write(msg + "\n")
+    exit(1)
+
+
+@dataclasses.dataclass
+class Config:
+    docs_path: str
+    version: str
+    local_output: bool
+    access_token: str
+    repo: str
+    repo_path: str
+
+
+def make_config() -> Config:
+    """Returns a Config by reading variables set in the environment. See docs
+    at the beginning of this file for required and optional variables.
+
+    Exits if any there are any validation errors.
+    """
+    if "ENV_VAR_DOCS_PATH" not in os.environ:
+        errorexit(
+            "ENV_VAR_DOCS_PATH must be present in the environment variables")
+
+    docs = os.environ["ENV_VAR_DOCS_PATH"]
+    if not os.path.isfile(docs):
+        errorexit(f"'{docs}' does not point to a file")
+
+    local = (os.environ.get("LOCAL_OUTPUT", "false") == "true")
+    if not local:
+        try:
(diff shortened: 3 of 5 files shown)


bin/env-var-docs-generate-markdown
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+source env-var-docs/venv/bin/activate
+
+local_output=${LOCAL_OUTPUT:-true}
+export LOCAL_OUTPUT=${local_output}
+export ENV_VAR_DOCS_PATH=server/conf/env-var-docs.json
+
+echo "Running env-var-docs/generate_markdown.py with LOCAL_OUTPUT=${LOCAL_OUTPUT}..."
+python env-var-docs/generate_markdown.py
```
bion, line 23: **question (blocking):** should this be `python3`? i don't have a `python` command on glinux or gmac

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+import dataclasses
+import env_var_docs.parser
+import env_var_docs.errors_formatter
+import github
+import os
+import sys
+import typing
+
+
+def errorexit(msg):
```
bion, line 24: **nitpick:** 
```suggestion
def error_exit(msg):
```

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+    if "ENV_VAR_DOCS_PATH" not in os.environ:
+        errorexit(
+            "ENV_VAR_DOCS_PATH must be present in the environment variables")
+
+    docs = os.environ["ENV_VAR_DOCS_PATH"]
+    if not os.path.isfile(docs):
+        errorexit(f"'{docs}' does not point to a file")
+
+    local = (os.environ.get("LOCAL_OUTPUT", "false") == "true")
+    if not local:
```
bion, line 55: **nitpick:** remove `not` and switch the cases since there's an `else` branch

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+
+def generate_markdown(
+    docs_file: typing.TextIO
+) -> tuple[str | None, list[env_var_docs.parser.NodeParseError]]:
+    out = ""
+
+    def output(node: env_var_docs.parser.Node):
+        nonlocal out  # Need to declare out nonlocal otherwise it gets shadowed.
```
bion, line 113: **suggestion (non-blocking):** move this method to the top level instead

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+        f"https://github.com/blob/main/{config.repo}/{path} updated in commit {res['commit']}"
+    )
+
+
+def generate_markdown(
+    docs_file: typing.TextIO
+) -> tuple[str | None, list[env_var_docs.parser.NodeParseError]]:
+    out = ""
+
+    def output(node: env_var_docs.parser.Node):
```
bion, line 112: **nitpick (blocking):** please give this method a more descriptive name that uses a verb

bion (changes requested):

bin/env-var-docs-generate-markdown
avrittrohwer: `python` inside the venv points to the venv python3 binary

jessbumblebee (commented):
Nice change

env-var-docs/generate_markdown.py
avrittrohwer: For consistency with the other scripts I will leave as errorexit

env-var-docs/generate_markdown.py
avrittrohwer: `output` closes over the local variable `out` so unless I made `out` a global `output` needs to be at the function level

env-var-docs/generate_markdown.py
avrittrohwer: Renamed to `append_node_to_out`

env-var-docs/generate_markdown.py
avrittrohwer: Done

env-var-docs/generate_markdown.py
bion: Why not pass it as a variable? I think that'd be more quickly understood by most readers

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+            "ENV_VAR_DOCS_PATH must be present in the environment variables")
+
+    docs = os.environ["ENV_VAR_DOCS_PATH"]
+    if not os.path.isfile(docs):
+        errorexit(f"'{docs}' does not point to a file")
+
+    local = (os.environ.get("LOCAL_OUTPUT", "false") == "true")
+    if not local:
+        try:
+            version = os.environ["RELEASE_VERSION"]
```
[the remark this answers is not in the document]
avrittrohwer: I do not expect this use case.  If someone needs to re-run it, I expect they re-run the step in the GitHub release action workflow

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+    repo_path: str
+
+
+def make_config() -> Config:
+    """Returns a Config by reading variables set in the environment. See docs
+    at the beginning of this file for required and optional variables.
+
+    Exits if any there are any validation errors.
+    """
+    if "ENV_VAR_DOCS_PATH" not in os.environ:
```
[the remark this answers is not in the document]
avrittrohwer: It is possible, it is just simpler code to look at environment variables vs parsing options from an args string.

env-var-docs/generate_markdown.py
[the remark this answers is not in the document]
avrittrohwer: `out` is the markdown document that we are building up.  Because `output` (renamed to `append_node_to_out`) is a callback passed to the parsing package `visit` function, it needs to capture `out` so that it can write to it.  We need to declare `out` as nonlocal for the capture to work

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
...
+
+    print(
+        f"https://github.com/blob/main/{config.repo}/{path} updated in commit {res['commit']}"
+    )
+
+
+def generate_markdown(
+    docs_file: typing.TextIO
+) -> tuple[str | None, list[env_var_docs.parser.NodeParseError]]:
+    out = ""
```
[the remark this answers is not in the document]
avrittrohwer: Yes, I am planning on having a top-level page for the generated docs that I will submit separately.  These generated pages will be separate files under the top-level page.

env-var-docs/generate_markdown.py
```diff
@@ -0,0 +1,155 @@
+"""Generates markdown from an environment variable documentation file and write it to a github repository.
```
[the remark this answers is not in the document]
avrittrohwer: Done.

env-var-docs/README.md
```diff
@@ -40,11 +40,13 @@ server is deployed.
 
 ### Running locally
 
-To run check_vars_documented.py and run_regex_tests.py locally:
+Before running any scripts, a python virtual environment (venv) must be
```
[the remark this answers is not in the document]
avrittrohwer: Done

bion (approved):

env-var-docs/generate_markdown.py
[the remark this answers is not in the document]
jessbumblebee: I disagree in this instance.
I'd consider it good practice to keep values/vars accessible in the scope that uses it only.
Env variables are kind of the opposite as they are accessible even outside the program. Or someone could refactor the file and the definition and reading end up in different files and make it hard to figure out where the value is coming from. 

Method arguments have the advantage of more clearly specifying the values a method uses to achieve its goal 

I wouldn't block the review on it, but if you dislike the idea of passing an argument, I'd favour a global variable scoped to this class over an env variable.

jessbumblebee (approved):

env-var-docs/generate_markdown.py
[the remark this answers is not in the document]
avrittrohwer: I guess I don't see a difference between getting the config values from the environment vs from a `--docs_path=... --version=...` in terms of scope accessibility.  The environment variables are accessibly anywhere in the python code with `os.environ` and the command line are are also accessible anywhere in the python code with `sys.argv`: https://docs.python.org/3/library/sys.html#sys.argv

I contain all the reading from the environment to the `make_config()` function which returns an object that the rest of the python code interacts with.  The python code does not write any new variables to the environment

Issue #4444, opened 2023-03-20
Address Correction - Admin tooltip copy


seanmalbert:
**Describe the bug**
The copy is incorrect for the tooltip after address correction is enabled on an address question. Additionaly, change `?` icon to `i` icon. See [Admin UX: New Address Correction toggle will be shown only on address type questions](https://docs.google.com/document/d/14zPxWA-TrjCmyxEgSkawYYUfDYxQPgY9DKQVQBRgfhc/edit?resourcekey=0-_QBYpeXQPsCN_ZyX_SuumA)

PR #4479, opened 2023-03-23
Change address correction tooltip copy and fix tooltip z

nb1701:
This changes the address correction tooltip copy. It also fixes an issue where the tooltip would render underneath other question elements.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4444


server/app/views/ViewUtils.java
MichaelZetune, line 1: Update screenshot tests that are failing?

server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -825,8 +825,18 @@ private Optional<FormTag> addressCorrectionEnabledToggle(
     boolean questionIsUsedInPredicate =
         programDefinition.isQuestionUsedInPredicate(questionDefinition.getId());
 
+    boolean addressCorrectionEnabledQuestionAlreadyExists =
+        blockDefinition.hasAddressCorrectionEnabledOnDifferentQuestion(questionDefinition.getId());
+
     String toolTipText =
-        "Enabling address correction will check the resident's address to ensure it is accurate.";
+        "Enabling 'address correction' will check the resident's address to ensure it is accurate.";
```
MichaelZetune, line 832: why add single quotes here?

MichaelZetune (changes requested):

server/app/views/admin/programs/ProgramBlockEditView.java
nb1701: This was what was in the copy, and seemed to be the only piece that was missing before.

PR #4498, opened 2023-03-27
Fix logged in application confirmation button

MichaelZetune:
### Description

The application confirmation button for logged in users has broken CSS for the button (below).

This PR fixes the tailwind code and adds a screenshot test for this case.

### Before:

image

### After:

image


rockycodes (approved):
lgtm!

PR #4503, opened 2023-03-28
Remove file ACL migration code

bion:
### Description

In 6/2022 we rewrote our file permissions logic to rely on per-file ACLs rather than URL-based system. This required migrating existing files into the new system. That migration has completed and the code for it is no longer needed.


server/app/controllers/FileController.java
```diff
@@ -82,7 +76,7 @@ public CompletionStage<Result> show(Request request, long applicantId, String fi
    */
```
MichaelZetune, line 76: Nit: change time -> team while we're here.

MichaelZetune (approved):

server/app/controllers/FileController.java
MichaelZetune: Actually, is this comment referring to the old code we deleted here? If so, let's remove?

bion: It does refer to the old code, but we need to keep it around to continue supporting the old URL pattern, even though the auth logic is no longer divergent.

PR #4504, opened 2023-03-28
Use a stepped builder for `Modal`

MichaelZetune:
### Description

Use a stepped builder in the `Modal` `AutoValue` class as per

https://github.com/google/auto/blob/main/value/userguide/builders-howto.md#-create-a-step-builder

This allows us to force clients to set a Modal ID and Content, without needing a dedicated `builder()` method for it.

### Checklist

- [x] Add `RequiredModalId` and `RequiredContent` interfaces to force clients to set these fields
- [x] Update usages


avaleske (approved):
I hadn't seen stepped builders before, they seem pretty useful. Thanks!

PR #4508, opened 2023-03-28
Hotfix: moved test up to be nested under dev check

rockycodes:
### Description

Prober tests are failing on AWS staging in part because we do not have an ESRI client and this test expects one. Moving this test up under the `isDevelopment` check makes it so that this test will not run on staging.


rockycodes:
resolved by https://github.com/civiform/civiform/pull/4411

Issue #4520, opened 2023-03-29
Write tests for `messages.*` files


MichaelZetune:
We should write tests to enforce invariants in the messages files and ensure they do not get out of sync.

PR #4516, opened 2023-03-28
Add tests for language files

MichaelZetune:
### Description

This PR adds some tests to enforce the following invariants on the language files:
* All keys in `messages` must be present in all other messages files
* An `IGNORE_LIST` is defined to enumerate exceptions to this rule. If a key is on the ignore list, (1) it must be in `messages` and (2) *may or may not* not be in one or more other messages files.

In a future PR:
* We will enforce that the keys in `messages` and other messages files must be the **exact same** (except for the `IGNORE_LIST`), as opposed to the current state where the other language files just need to *have* those same keys. Once this is complete, if a key is on `IGNORE_LIST` it **may not** be in any language file other than `messages`.
   * Currently, we cannot do this because there are keys that are present in some but not all non-primary language files.

### Checklist

- [x] Created `test/conf/MessagesTest.java` as per description above.
- [x] Delete `bin/list-incomplete-messages` which does the same thing as this test (but doesn't run on CI)

## Issues

Progress towards #4520.

PR #4530, opened 2023-03-29
Test full key equality in `MessagesTest.java`

MichaelZetune:
### Description

Previously in #4516, we only tested that all keys in `messages` appeared in other messages files. Now, we test that all keys are exactly the same in all files, except for the `IGNORE_LIST`.

The new error message looks something like this:

`[error] Test conf.MessagesTest.messages_keysInPrimaryFileInAllOtherFiles(conf/messages.so) [4] failed: org.junit.ComparisonFailure: [content.submittedDate] found in primary language file but not in conf/messages.so. Add these keys to conf/messages.so or to the ignore list in conf.MessagesTest to resolve this issue.`

`[error] [content.requiredFieldsAnnotation] found in conf/messages.so file but not in primary language file. Add these keys to primary language file or to the ignore list in conf.MessagesTest to resolve this issue.`

#### General

- [x] Create `errorMessage()` method to encapsulate logic to create an error string that describes extra keys in primary language file or extra keys in foreign language files
- [x] Change `assertThat()` call to be simpler and just check that the key set in each pair of files is equal.

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4520

### DIFF
--- a/server/test/conf/MessagesTest.java
+++ b/server/test/conf/MessagesTest.java
@@ -55,28 +55,9 @@ public class MessagesTest {
 
     // Checks that the language file contains exactly the same message keys as the
     // primary language file.
-    //
-    // TODO(#4520): Modify this to ensure the keys are the same, not just that all
-    //  keys in primary file exist in foreign language files. Currently this is not possible,
-    //  even with IGNORE_LIST, because there are keys that are present in some but not all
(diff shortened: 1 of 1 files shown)


rockycodes (approved):
looks great! i tested it out locally and everything works as expected.

PR #4531, opened 2023-03-29
Unify Modal Styling

MichaelZetune:
### Description

Currently our modals look roughly like the following:

image

We have mocks that use a non-existent modal like so:

image
There's no reason to have and maintain different styling between applicant and admin modals, so this PR changes the `Modal` class (the only one) to match the new mocks.

---

**NOTE**: the buttons in the modal are not centralized in the Modal class and associated styles (maybe we can do so in the future. The client provides the button content. As a result, you can see in the screenshots that the old button styles remain.

## Release notes

Modal design has been modernized.

#### General

- [x] Change some styling in `BaseStyles.java`.
- [x] Change some styling in clients of `Modal.java` as they need to account for the new container margin etc.
- [x] Update browser test screenshots.

### Issue(s) this completes

Relates to #4367.


server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -991,13 +991,12 @@ private Modal blockDeleteModal(
       deleteBlockForm
           .withId("block-delete-form")
           .with(
-              div(h1("Are you sure you want to delete this screen?")
-                      .withClasses("text-base", "mb-2"))
-                  .withClasses("mx-4"),
+              div(
+                  h1("Are you sure you want to delete this screen?")
+                      .withClasses("text-base", "mb-4")),
```
swatkat1, line 996: is mb-2 to mb-4 part of the new design?

server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -1015,15 +1014,13 @@ private Modal blockDeleteModal(
...
               div(
-                      h1(join(blockDefinition.name(), " includes ", b(listItemsInBlock + ".")))
-                          .withClasses("text-base", "mb-2"),
-                      h1("Are you sure you want to delete this screen?")
-                          .withClasses("text-base", "mb-2"))
-                  .withClasses("mx-4"),
+                  h1(join(blockDefinition.name(), " includes ", b(listItemsInBlock + ".")))
+                      .withClasses("text-base", "mb-2"),
+                  h1("Are you sure you want to delete this screen?")
+                      .withClasses("text-base", "mb-4")),
```
swatkat1, line 1020: So we are getting rid of mb-2 replacing it with mb-4? just checking if we still want to keep line 1018 as mb-2

server/app/views/admin/programs/ProgramBlockEditView.java
MichaelZetune: No quite; because we are changing the full modal container, we needed to make adjustments to several clients of the Modal class so that the padding wasn't too much, otherwise they are indented too far to the right/there is too much space below the modal content.

In principle, the same thing was done with the original Modal container; clients had to adjust their margin/padding to make it look good.

server/app/views/admin/programs/ProgramBlockEditView.java
MichaelZetune: Yep, we want to keep mb-2 on the "includes" line. This is it:

image

swatkat1 (approved):

PR #4535, opened 2023-03-30
Remove eligibility criteria for the common intake form.

### LINKED ISSUE #4512: Eligibility conditions should be removed when a program becomes the CIF
The Common Intake Form is not intended to have eligibility conditions so we've hidden the eligibility modification section from its program blocks edit view. However, if an admin creates a regular program with eligibility conditions and later sets it as the common intake form, the eligibility conditions would persist on the backend and applicants will still see the experience. Probably the worst part of this is that since the eligibility conditions section is hidden from the admin at that point, they won't be able to remove the conditions.

**To Reproduce**

1. Create a program, don't mark it as the pre-screener.
2. Add eligibility conditions to the program
3. Publish the program
4. Edit the program and mark it as the pre-screener.
5. Observe that the eligibility conditions section has been hidden.
6. Publish the program.
7. Login as an applicant and apply to the program.
8. Observe that the eligibility conditions are still in place.

reemaka:
### Description

Fix for #4512.

1. Clears eligibility conditions when a program is set as the common intake form.
2. Adds some safeguards when adding eligibility conditions to make sure that they're not being set on the CIF -- there's no UI for this so it shouldn't ever happen, but just to be safe.
3. Adds a test for adding eligibility conditions in general since there was none.

#### General

### Issue(s) this completes

Fixes #4512

### DIFF
--- a/server/app/services/program/EligibilityNotValidForProgramTypeException.java
+++ b/server/app/services/program/EligibilityNotValidForProgramTypeException.java
@@ -1,0 +1,12 @@
+package services.program;
+
+/**
+ * EligibilityNotValidForProgramType is thrown when eligibility predicates are attempted to be added
+ * to a Program that has a ProgramType that cannot have eligibility conditions.
+ */
+public class EligibilityNotValidForProgramTypeException extends Exception {
+
+  public EligibilityNotValidForProgramTypeException() {
+    super("Eligibility conditions cannot be set for this ProgramType.");
+  }
+}
--- a/server/app/services/program/EligibilityNotValidForProgramTypeException.java
+++ b/server/app/services/program/EligibilityNotValidForProgramTypeException.java
@@ -6,7 +6,9 @@ package services.program;
  */
 public class EligibilityNotValidForProgramTypeException extends Exception {
 
-  public EligibilityNotValidForProgramTypeException() {
-    super("Eligibility conditions cannot be set for this ProgramType.");
+  public EligibilityNotValidForProgramTypeException(ProgramType programType) {
+    super(
+        String.format(
+            "Eligibility conditions cannot be set for ProgramType %s", programType.toString()));
   }
 }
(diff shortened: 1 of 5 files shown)


server/app/services/program/EligibilityNotValidForProgramTypeException.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * EligibilityNotValidForProgramType is thrown when eligibility predicates are attempted to be added
+ * to a Program that has a ProgramType that cannot have eligibility conditions.
+ */
+public class EligibilityNotValidForProgramTypeException extends Exception {
+
+  public EligibilityNotValidForProgramTypeException() {
+    super("Eligibility conditions cannot be set for this ProgramType.");
+  }
```
bion, line 11: **nitpick (non-blocking, if-minor):** include the program type in the error message (on the theory that it could be something other than common intake in the future)

server/app/services/program/ProgramServiceImpl.java
```diff
@@ -862,6 +872,10 @@ public ProgramDefinition removeBlockEligibilityPredicate(long programId, long bl
     } catch (IllegalPredicateOrderingException e) {
       // Removing a predicate should never invalidate another.
       throw new RuntimeException("Unexpected error: removing this predicate invalidates another");
+    } catch (EligibilityNotValidForProgramTypeException e) {
+      // Removing eligibility predicates should always be valid.
+      throw new RuntimeException(
+          "Unexpected error: removing this predicate is not allowed for this ProgramType");
```
bion, line 878: **nitpick (blocking):** include the original exception so we can see it in the stack trace
```suggestion
      throw new RuntimeException(
          "Unexpected error: removing this predicate is not allowed for this ProgramType", e);
```

bion (approved):
couple minor things but looks good!

PR #4500, opened 2023-03-27
Prompt guests to log in with modal on application confirmation screen

MichaelZetune:
### Description

Adds a new modal to prompt guests to log in before applying to more programs. This modal pops up when guests click "Apply to another program". 

## Release notes

Adds a new modal to prompt guests to log in before applying to more programs.

#### General

- [x] Replace "All Done" and "Create account or sign in" with "Log in" and "Create Account"
- [x] Make "Apply to another program lead to a login prompt instead of straight back to the programs page
- [x] Add screenshot test for new modal
- [x] Add translations

image

PR #4542, opened 2023-03-30
Add `flex` tags to applicant styles

MichaelZetune:
### Description

In the admin styles, we set `flex` on the `body` tag to layout all elements in the DOM:

https://github.com/civiform/civiform/blob/11cfe44b129325ecda6a137b1ef5fffb315df1c8/server/app/views/style/AdminStyles.java#L54-L55

We don't do the same in the applicant styles. Incidentally, all uses of `Modal`s are on the admin site.

It turns out that using a `Modal` in our codebase requires that the `body` has a `flex` tag. So we need to add it to `ApplicantStyles.java`. This will unblock the child PR #4500.

Unfortunately this trivially changes just about every screenshot in our tests. It's visually hardly visible, but that's why this PR is large.

#### General

- [x] Add `flex` and `flex-col` to `ApplicantStyles.java`.
- [x] Regenerate screenshots

### Issue(s) this completes

Indirectly helps with #4367 by unblocking #4500.


avaleske (approved):
Thanks for dubugging this!

Issue #4217, opened 2023-02-15
Browser tests are flaky


nb1701:
Ideas:

- [ ] run the find flaky test script on a cron, report results
- [x] look at Brett's browser test PR: https://github.com/civiform/civiform/pull/3389
- [x] figure out what's taking the popup so long

Flaky tests:

- [x] [src/application_navigation.test.ts:149-150](https://github.com/civiform/civiform/actions/runs/4557075631/jobs/8038245538)
   - Fixed in #4539
- [ ] [src/admin_predicates.test.ts:270](https://github.com/civiform/civiform/actions/runs/4547433374/jobs/8017303164)
   - Maybe fixed in #4550. Hard to tell as local runs have been consistently passing.
- [ ] [src/support/admin_programs.ts:774](https://github.com/civiform/civiform/actions/runs/4558582078/jobs/8041625390)
   - May be related to #2018
   - Cannot reproduce flakiness with `bin/find-flakey-browser-tests -n 5 -t application_review.test.ts` and `fit()` on the affected test (`src/application_review.test.ts` uses `admin_programs.ts`)
 - [x] [src/file.test.ts](https://github.com/civiform/civiform/actions/runs/4852129375/jobs/8646731307)
    - `net::ERR_ABORTED` error appears
    - Bion suggests swallowing the error when it happens as it wouldn't interfere with the test's assertions
    - Implemented in #4873

nb1701:
From @gwendolyngoetz : This should help diagnose flaky tests https://github.com/civiform/civiform/blob/main/bin/find-flakey-browser-tests

shanemc-goog:
Fyi: for interacting with the question bank I reduced the animation wait from 1s to 250ms, which saves at least 4 minutes in the test runs. #4131 

If flakyness is around the question bank then that might need upping again, or for the QB interaction detection to be better

nb1701:
Investigate if https://github.com/civiform/civiform/pull/3389 makes the tests less flaky.

MichaelZetune:
Idea from Gwen: consider lowering hardware specs on Docker container to see if that makes the flakiness more apparent locally. See if we can match the GitHub actions specs.

gwendolyngoetz:
Related #4111

michaelzetune:
Closing as lately we haven't been seeing much flakiness in browser tests.

PR #4550, opened 2023-03-31
Fix flakiness in `admin_predicates_test.ts` by waiting for the popup before clicking a link

MichaelZetune:
### Description

Based on #4217, `admin_predicates.test.ts` has been known to be flaky. Unfortunately, running `bin/find-flakey-browser-tests -n 5 -t src/admin_predicates.test.ts` does not expose the flakiness on my local machine.

However, this is the same problem we observed in `application_navigation.test.ts`, so apply the same exact fix as in #4539; hopefully it will do the trick.

## Issues

Relates to #4217

### DIFF
--- a/browser-test/src/admin_predicates.test.ts
+++ b/browser-test/src/admin_predicates.test.ts
@@ -266,8 +266,10 @@ describe('create and edit predicates', () => {
     await applicantQuestions.expectIneligiblePage()
     await validateScreenshot(page, 'ineligible')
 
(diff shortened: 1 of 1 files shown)


avrittrohwer:
I tested locally with 20 runs both with and without this change.  All 40 tests passed.  Given that we saw a similar issue with another test, it LGTM to change this test as well

avrittrohwer (approved):

PR #4559, opened 2023-04-03
Re-trigger address validation upon change to address

### LINKED ISSUE #4310: Changes to Address not Triggering Correction a Second Time

gwendolyngoetz:
### Description

Allows re-triggering address validation if the user makes a change to their address. The corrected, lat, long, and wellKnownId values are reset, which clears the flow to be picked up as normal by the address correction process.

#### General

### Issue(s) this completes

Fixes #4310

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramBlocksController.java
+++ b/server/app/controllers/applicant/ApplicantProgramBlocksController.java
@@ -444,15 +444,19 @@ public final class ApplicantProgramBlocksController extends CiviFormController {
             v -> {
               DynamicForm form = formFactory.form().bindFromRequest(request);
               ImmutableMap<String, String> formData = cleanForm(form.rawData());
-
-              return applicantService.stageAndUpdateIfValid(
-                  applicantId,
-                  programId,
-                  blockId,
-                  formData,
-                  featureFlags.getFlagEnabled(
-                      request, ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED));
+              return applicantService.resetAddressCorrectionWhenAddressChanged(
+                  applicantId, programId, blockId, formData);
             },
+            httpExecutionContext.current())
+        .thenComposeAsync(
+            formData ->
+                applicantService.stageAndUpdateIfValid(
+                    applicantId,
+                    programId,
+                    blockId,
+                    formData,
+                    featureFlags.getFlagEnabled(
+                        request, ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED)),
             httpExecutionContext.current())
         .thenComposeAsync(
             roApplicantProgramService ->
--- a/server/app/services/applicant/ApplicantService.java
+++ b/server/app/services/applicant/ApplicantService.java
@@ -1519,4 +1519,48 @@ public final class ApplicantService {
               return suggestionsMaybe.get();
             });
   }
+
+  /**
+   * Checks for an {@link AddressQuestion} that has address correction enabled. If found and the
+   * data in the database differs from the submitted form data. Set the corrected, latitude, and
+   * longitude, and wellKnownId values to empty. This is done to allow triggering the address
+   * correction process again, but only when there have been changes.
+   *
+   * <p>Otherwise if no {@link AddressQuestion} or no changes to the address we return the untouched
+   * formdata
+   */
+  public CompletionStage<ImmutableMap<String, String>> resetAddressCorrectionWhenAddressChanged(
+      long applicantId, long programId, String blockId, ImmutableMap<String, String> formData) {
+    return getReadOnlyApplicantProgramService(applicantId, programId)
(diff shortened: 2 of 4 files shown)


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -444,16 +444,20 @@ public CompletionStage<Result> update(
...
               ImmutableMap<String, String> formData = cleanForm(form.rawData());
-
-              return applicantService.stageAndUpdateIfValid(
-                  applicantId,
-                  programId,
-                  blockId,
-                  formData,
-                  featureFlags.getFlagEnabled(
-                      request, ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED));
+              return applicantService.resetAddressCorrectionWhenAddressChanged(
```
nb1701, line 447: **thought (non-blocking):** Would it makes sense to do this inside `stageAndUpdateIfValid`?  Those functions are a bit hairy, but we might be able to reuse the roApplicantProgramService and grabbing the block.

server/app/services/applicant/ApplicantService.java
```diff
@@ -1519,4 +1519,48 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestionGroup(Block b
...
+
+              if (blockMaybe.isEmpty()) {
+                return CompletableFuture.failedFuture(
+                    new ProgramBlockNotFoundException(programId, blockId));
+              }
+
+              Optional<ApplicantQuestion> addressQuestionMaybe =
+                  blockMaybe.get().getAddressQuestionWithCorrectionEnabled();
+
+              if (addressQuestionMaybe.isPresent()) {
```
nb1701, line 1547: **nitpick:** Perhaps return early if empty to reduce the nesting.

server/app/services/applicant/ApplicantService.java
```diff
@@ -1519,4 +1519,48 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestionGroup(Block b
...
+                AddressQuestion addressQuestion =
+                    addressQuestionMaybe.get().createAddressQuestion();
+
+                if (addressQuestion.hasChanges(formData)) {
```
nb1701, line 1551: **nitpick:** 
```java
return CompletableFuture.completedFuture(
  addressQuestion.hasChanges(formData)
  ? new ImmutableMap.Builder<String, String>()
       .putAll(formData)
       .put(addressQuestion.getCorrectedPath().toString(), "")
       .put(addressQuestion.getLatitudePath().toString(), "")
       .put(addressQuestion.getLongitudePath().toString(), "")
       .put(addressQuestion.getWellKnownIdPath().toString(), "")
       .build());
  : formData);
```

server/app/services/applicant/question/AddressQuestion.java
```diff
@@ -321,4 +323,36 @@ public ImmutableList<Path> getAllPaths() {
...
+            && formData.containsKey(getZipPath().toString());
+
+    if (!hasAllRequiredPaths) {
+      return false;
+    }
+
+    BiFunction<Path, Optional<String>, Boolean> checkIfChanged =
+        (path, value) -> !Objects.equals(formData.get(path.toString()), value.orElse(""));
+
+    boolean hasChangesStreet = checkIfChanged.apply(getStreetPath(), getStreetValue());
```
nb1701, line 346: **nitpick:** 
```java
return checkIfChanged.apply(getStreetPath(), getStreetValue())
  || checkIfChanged.apply(getLine2Path(), getLine2Value())
  || checkIfChanged.apply(getCityPath(), getCityValue())
  || checkIfChanged.apply(getStatePath(), getStateValue())
  || checkIfChanged.apply(getZipPath(), getZipValue());
```

nb1701 (changes requested):

server/app/services/applicant/ApplicantService.java
gwendolyngoetz: done

server/app/services/applicant/ApplicantService.java
gwendolyngoetz: I think going the ternary route in this instance negatively affects the readability more than it helps. I want it to remain very clear the default it returns is the original formData and the length with the Builder makes that.

server/app/controllers/applicant/ApplicantProgramBlocksController.java
gwendolyngoetz: Unfortunately the call to reset the previously corrected address can't happen inside `stageAndUpdateIfValid`. The `stageAndUpdateIfValid` method is also called to update the address after the user selects a value from the list of address correction options, but since it is now different it would trigger the reset again and get stuck in a loop. The only context of which place the user did something is coming from knowing the controller action.

nb1701 (approved):

PR #4567, opened 2023-04-03
Add back button to program settings.

### LINKED ISSUE #4494: Add a back button on the program settings page
Add a back button on the program settings page.

From Miriam: "Can the back link be dynamic and either show '← Back to all programs' or '← Back to [Program name] program edit page'"

reemaka:
### Description

Add a back button to the program settings page. The target of the back button dynamically changes depending on whether the admin is coming from the program index view or the program blocks view.

#### General

### Issue(s) this completes

Fixes #4494


server/app/views/admin/programs/ProgramSettingsEditView.java
```diff
@@ -101,9 +110,28 @@ public Content render(Request request, ProgramDefinition program) {
...
+            .with(div().withClasses("mt-4").with(h1(title)))
             .with(div(formTag));
 
     return layout.renderCentered(layout.getBundle().setTitle(title).addMainContent(contentDiv));
   }
+
+  private ATag getBackButton(Request request, ProgramDefinition program) {
+    String backTarget = controllers.admin.routes.AdminProgramController.index().url();
+    String buttonText = "Back to all programs";
+    Optional<String> navSource = request.session().get(NAVIGATION_SOURCE_SESSION_KEY);
```
gwendolyngoetz, line 123: Rather than rely on the session, you can probably just pull the referrer from the http headers to see which page you came from. It'll only be populated if the user got to this page by clicking a link instead of typing in a url, but that's probably pretty unlikely for this page and it's essentially the same problem if you came here without having set the session value you have.

If that solves the problem I'd recommend going that route, otherwise we've been asked to make sure that values we put in the session are encrypted because they are stored in the user's cookie.

Here's an example where that was done to serialize json.
https://github.com/civiform/civiform/blob/56b2253ce22d4657b98556cee29a13cd0bbdfc10/server/app/controllers/geo/AddressSuggestionJsonSerializer.java

Usage of AddressSuggestionJsonSerializer
https://github.com/civiform/civiform/blob/5fd6e5f09957d60baaba39ac424830eeb0315447/server/app/controllers/applicant/ApplicantProgramBlocksController.java#L165-L167

gwendolyngoetz (changes requested):

server/app/views/admin/programs/ProgramSettingsEditView.java
reemaka: This didn't originally work because of the page reload that happened after clicking the toggle. I just fixed that issue in #4598, however, so I've now updated this PR to use the referrer. Thanks for the suggestion!

gwendolyngoetz (approved):

Issue #4443, opened 2023-03-20
Address Correction - Suggestions with only City/State shouldn't show


seanmalbert:
**Describe the bug**
CiviForm should not show suggestions with only the City/State or without a street number: [Applicant UX: Address correction dialog with 1 suggestion](https://docs.google.com/document/d/14zPxWA-TrjCmyxEgSkawYYUfDYxQPgY9DKQVQBRgfhc/edit?resourcekey=0-_QBYpeXQPsCN_ZyX_SuumA)

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #4570, opened 2023-04-03
Ignore incomplete address suggestions

nb1701:
Sometimes, the Esri service will give back suggestions that do not include a number in the street address, or it omits a street address entirely. This ignores these suggestions.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4443


server/app/services/geo/esri/EsriClient.java
```diff
@@ -105,6 +105,15 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
                         .setState(attributes.get("RegionAbbr").asText())
                         .setZip(attributes.get("Postal").asText())
                         .build();
+                // Suggestion must be a fully formed address, including a number.
+                // Sometimes the esri service returns only a partially formed address.
+                if (candidateAddress.getStreet().isEmpty()
+                    || !candidateAddress.getStreet().matches("^[0-9]+.*")
```
gwendolyngoetz, line 111: Just checking that the street value is not empty should suffice. We risk losing potential valid answers with the regex check.

gwendolyngoetz (changes requested):

server/app/services/geo/esri/EsriClient.java
nb1701: According to https://github.com/civiform/civiform/issues/4443, we also wanted to get rid of suggestions without a number.  I think when you provide a fairly well formed address, sometimes the esri service offers up suggestions that omit the number for some reason.  What kind of addresses might we want to include without a number?  I feel like there are some, I just can't picture it right now. :)

nb1701: I guess P.O. boxes?

gwendolyngoetz: There could be PO Boxes

nb1701: Hmm....gotta think about how we should handle that.  Maybe it's not so terrible to have a suggestion that doesn't have a number when it should.  Users can just ignore it.

nb1701: Suppose we could try to get fancy and do something like "if a user provided a street address number, remove suggestions that don't have one too" but I fear it might be creating edge cases for ourselves.

gwendolyngoetz: Yeah I don't like going with a complex plan unless it is causing real problems. I think as long as all the fields have a value it should be fine.

nb1701: 👍 Updated

gwendolyngoetz (approved):

Issue #4473, opened 2023-03-22
Address Correction - add "Learn More" link to tooltip


nb1701:
https://docs.google.com/document/d/14zPxWA-TrjCmyxEgSkawYYUfDYxQPgY9DKQVQBRgfhc/edit?resourcekey=0-_QBYpeXQPsCN_ZyX_SuumA#

The address correction tooltip, when the feature is disabled, should point the user to a CiviForm docs page giving instructions for IT managers to set up the GIS service.

https://docs.civiform.us/it-manual/sre-playbook/configure-gis-service

PR #4582, opened 2023-04-05
Add "Learn more" link to address correction tooltip

nb1701:
This adds a "Learn more" link to the address correction tooltip when AC is not enabled, pointing to the docs page that describes how to enable it. It also reduces the top margin to zero in this case so that the tooltip doesn't disappear when a user mouses away from the tooltip icon to go to click the link.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4473

### DIFF
--- a/server/app/views/admin/programs/ProgramBlockEditView.java
+++ b/server/app/views/admin/programs/ProgramBlockEditView.java
@@ -827,9 +827,10 @@ public final class ProgramBlockEditView extends ProgramBlockBaseView {
 
     boolean questionIsUsedInPredicate =
         programDefinition.isQuestionUsedInPredicate(questionDefinition.getId());
-
     boolean addressCorrectionEnabledQuestionAlreadyExists =
         blockDefinition.hasAddressCorrectionEnabledOnDifferentQuestion(questionDefinition.getId());
+    boolean addressCorrectionDisabled =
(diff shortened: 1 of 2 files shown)


server/app/views/admin/programs/ProgramBlockEditView.java
```diff
@@ -840,16 +843,35 @@ private Optional<FormTag> addressCorrectionEnabledToggle(
...
+      // Leave the space at the end, because we will add a "Learn more" link. This
+      // should always be the last string added to toolTipText for this reason.
+      toolTipText +=
+          " To use this feature, you will need to have your IT manager configure the GIS service. ";
+      // We must set mt-0, overriding mt-2.5 that comes before it in the classes string, or else
+      // a user will be unable to move their mouse inside the tooltip to be able to click on the
+      // link without the tooltip disappearing.
+      toolTip =
+          ViewUtils.makeSvgToolTip(
+              toolTipText, Icons.INFO, Optional.of("right-1/2 mt-0"), learnMoreLink);
```
nb1701, line 870: Not sure the `mt-0` is actually working reliably.  Hmm...

nb1701: Probably a bad idea to include both `mt-2.5` and `mt-0` anyway.  I think I'll just have to create a new function in ViewUtils instead.

nb1701:
image

reemaka (approved):

PR #4583, opened 2023-04-05
Skip validation screen if user enters valid address

### LINKED ISSUE #4471: Don't show suggestions when suggestions contains the exact address entered
Currently, the address suggestions are always shown for address questions with address correction enabled, even if the suggestion as received from the esri service exactly matches the address given.

image

Instead, we should accept the address as entered and not require the user to choose a suggested address that exactly matches.

rockycodes:
### Description

Adds logic to skip the address validation screen if a user enters a valid address (one of the returned suggestions exactly matches the entered address) and send the validated address to the db.

For tests, this required adding a case to `FakeEsriClient` and adding a method that allows clearing `applicantData` off the `applicant` object before calling `getApplicantData` so that a fresh `ApplicantData` object is generated.

#### General
- [x] Created tests which fail without the change

### Issue(s) this completes

Fixes #4471;


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -166,6 +165,18 @@ public CompletionStage<Result> confirmAddress(
         addressSuggestionJsonSerializer.deserialize(
             maybeAddressJson.orElseThrow(() -> new RuntimeException("Address JSON missing")));
 
+    return confirmAddressWithSuggestions(
```
nb1701, line 168: **suggestion:** Maybe we don't need the extra function here?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -519,6 +530,33 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
                 addressSuggestionGroup -> {
                   ImmutableList<AddressSuggestion> suggestions =
                       addressSuggestionGroup.getAddressSuggestions();
+
+                  /**
+                   * Check for an address suggestion that exactly matches that address entered by
+                   * the applicant. If a match is found, save the matched suggestion to the db and
+                   * skip the address validation screen.
+                   */
+                  for (int i = 0; i < suggestions.size(); i++) {
```
nb1701, line 539: **suggestion:** Might be able to do something like this:
```java
if (suggestions.stream().anyMatch(
  suggestion -> suggestion.getAddress().equals(addressQuestion.getAddress())) {
    confirmAddressWithSuggestions...
}
```

nb1701 (changes requested):

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -165,7 +166,18 @@ public CompletionStage<Result> confirmAddress(
...
+  }
 
+  private CompletionStage<Result> confirmAddressWithSuggestions(
+      Request request,
+      long applicantId,
+      long programId,
+      String blockId,
+      boolean inReview,
+      String selectedAddress,
+      ImmutableList<AddressSuggestion> suggestions) {
```
rockycodes, line 180: Note that the way github is collapsing this is confusing. `confirmAddressWithSuggestions` just contains logic I factored out of `confirmAddress`. The logic below is added to the `renderErrorOrRedirectToNextBlock` method (expand the hidden lines to see the full method).

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -513,18 +525,37 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
...
-                addressSuggestionGroup -> {
-                  ImmutableList<AddressSuggestion> suggestions =
-                      addressSuggestionGroup.getAddressSuggestions();
-                  String json = addressSuggestionJsonSerializer.serialize(suggestions);
-
-                  Boolean isEligibilityEnabledOnThisBlock =
-                      thisBlockUpdated.getLeafAddressNodeServiceAreaIds().isPresent();
-
-                  return ok(addressCorrectionBlockView.render(
+        AddressSuggestionGroup addressSuggestionGroup =
```
rockycodes, line 528: I'm making the call to get the address suggestions synchronous here. I feel like that isn't ideal (??) but I couldn't figure out how else to have the user paths divide. 

The key here is that if there is a matching suggestion, we want to save it to the backend and then move on to the next screen without showing the address correction screen (calling `confirmAddressWithSuggestions` accomplishes this because it invokes `renderErrorOrRedirectToNextBlock` again, this time with a corrected address). If we don't find a matching suggestion, we want to show the user the address correction screen.

Advice welcome!

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -513,18 +525,37 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
...
+        ImmutableList<AddressSuggestion> suggestions =
+            addressSuggestionGroup.getAddressSuggestions();
+
+        AddressSuggestion[] suggestionMatch =
+            suggestions.stream()
+                .filter(suggestion -> suggestion.getAddress().equals(addressQuestion.getAddress()))
+                .toArray(AddressSuggestion[]::new);
+
+        if (suggestionMatch.length > 0) {
+          return confirmAddressWithSuggestions(
```
rockycodes, line 542: The key here is that if there is a matching suggestion, we want to save it to the backend and then move on to the next screen without showing the address correction screen (calling `confirmAddressWithSuggestions` accomplishes this because it invokes `renderErrorOrRedirectToNextBlock` again, this time with a corrected address).

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -513,18 +525,37 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
...
+              inReview,
+              suggestionMatch[0].getSingleLineAddress(),
+              suggestions);
+        } else {
+          String json = addressSuggestionJsonSerializer.serialize(suggestions);
+
+          Boolean isEligibilityEnabledOnThisBlock =
+              thisBlockUpdated.getLeafAddressNodeServiceAreaIds().isPresent();
+
+          return supplyAsync(
```
rockycodes, line 556: If we don't find a matching suggestion, we want to show the user the address correction screen.

server/app/controllers/applicant/ApplicantProgramBlocksController.java
bion: To make this asynchronous, return the completion stage returned by `applicantService.getAddressSuggestionGroup` move the remainder of the method into a callback on it:

```java
return applicantService.getAddressSuggestionGroup(thisBlockUpdated)
  .thenComposeAsync((AddressSuggestionGroup addressSuggestionGroup) -> {
        ImmutableList<AddressSuggestion> suggestions =
            addressSuggestionGroup.getAddressSuggestions();
        // etc
      },

```

rockycodes: @bion that's how it is [currently](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/ApplicantProgramBlocksController.java#L516) and the issue I ran into there is the branching code paths. returning `return ok(addressCorrectionBlockView.render(...` is all well and good, but if we don't want to render the correction screen, we need to call `confirmAddressWithSuggestions` which has a `CompletionStage<Result>` type which `thenComposeAsync` doesn't like. another thought i had was to refactor `confirmAddressWithSuggestions` so it returns a `Result` but I couldn't figure out how to do that with the code contained in that method.

server/app/models/Applicant.java
```diff
@@ -51,6 +51,16 @@ public Applicant() {
     super();
   }
 
+  /**
+   * Clears applicantData so that a new ApplicantData object will be returned from getApplicantData.
+   * Useful when applicantData has been stored as an empty object and needs to be udpated with the
+   * data stored in object
+   */
+  public ApplicantData getApplicantData(boolean expireCache) {
```
bion, line 59: **suggestion (blocking):** Generally speaking, booleans should be avoided in method parameters because they:
- give no suggestion to what they're for at the call site without an inline comment
- need to be replaced with enums when more than two states become necessary

In this case I'd recommend removing the param, changing the method name to `getApplicantDataSkipCache` (or similar), removing `public` and adding the `@VisibleForTesting` annotation

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -513,32 +529,22 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
...
-                              inReview,
-                              roApplicantProgramService,
-                              thisBlockUpdated,
-                              applicantName,
-                              ApplicantQuestionRendererParams.ErrorDisplayMode.DISPLAY_ERRORS),
-                          messagesApi.preferred(request),
-                          addressSuggestionGroup,
-                          isEligibilityEnabledOnThisBlock))
-                      .addingToSession(request, ADDRESS_JSON_SESSION_KEY, json);
+                  return maybeRenderAddressCorrectionScreen(
```
nb1701, line 537: I think you might not need the `{}` and `return` here if it's only one statement?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -165,7 +166,22 @@ public CompletionStage<Result> confirmAddress(
...
         addressSuggestionJsonSerializer.deserialize(
             maybeAddressJson.orElseThrow(() -> new RuntimeException("Address JSON missing")));
+    return confirmAddressWithSuggestions(
+        request, applicantId, programId, blockId, inReview, selectedAddress, suggestions);
+  }
 
+  /**
+   * This method saves the selected corrected address to the db and redirects the user to the next
+   * screen
+   */
```
bion, line 176: **nitpick:** update comment wording and formatting to match style guide: https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -606,6 +611,64 @@ private CompletionStage<Result> renderErrorOrRedirectToNextBlock(
                 .flashing(flashingMap));
   }
 
+  /**
+   * This method determines if the address entered by the user matches one of the suggestions
+   * returned by the Esri service. If a matching suggestion is found, it is saved to the db and the
+   * user is sent on to the next question. Otherwise, the user is directed to a screen where they
+   * can pick from the corrected address suggestions returned by the Esri service.
```
bion, line 618: **nitpick:** javadoc style here too

server/app/models/Applicant.java
```diff
@@ -107,4 +108,14 @@ public Applicant saveAndReturn() {
     save();
     return this;
   }
+
+  /**
+   * Clears applicantData so that a new ApplicantData object will be returned from getApplicantData.
+   * Useful in tests when applicantData has been stored as an empty object and needs to be udpated
+   * with the data stored in object
```
bion, line 115: **nitpick:** comment style, also please use `{@link ApplicantData}` and `{@code getApplicantData}` where applicable for easier javadoc navigation

bion (approved):
Looks good, couple nitpicks around javadoc style

nb1701 (approved):

PR #4586, opened 2023-04-05
Have child action inherit secrets

gwendolyngoetz:
### Description

Makes the call to the generate docs action inherit the secrets from the release action context. This should allow the generate docs action to run standalone and as a child action

#### General


gwendolyngoetz:
Covered by #4587

PR #4587, opened 2023-04-05
Avritt/md

avrittrohwer:
### Description

release.yaml run https://github.com/civiform/civiform/actions/runs/4621475391 failed because it was referencing the wrong secret name.  This change fixes that.


gwendolyngoetz (approved):
I was just about to send a PR of my own out to fix this same issue. Will close mine :+1:

PR #4589, opened 2023-04-05
Limit number of calls on program admin index page

### LINKED ISSUE #4608: Admin index page takes a long time to load
**Description**
Recently, Seattle has mentioned latency on the admin program index page (~6 seconds load time) 

This is largely due to loading active and draft programs, as well as loading the questions within it.

dkatzz:
### Description

Create a version of ActiveAndDraftPrograms that doesn't load the question data. This is called to load the CiviForm admin program index page, but we don't need the questions when we load the programs there.

This makes the page load ~150 ms, which is over a 10X improvement.

This change reduces the number of database calls for an index page with 31 programs from [671](https://docs.google.com/document/d/13FqPvA1yUoEuCw8DNAMEalM2wq8eCUU1DsSFOUBbRlM/edit) to [14](https://docs.google.com/document/d/1JHvo17lyPmRRQ_EyYsqKdNgUsfoJ0-IiPmy2A11R6Jo/edit).

## Release notes

Improve load time of the CiviForm admin index page

#### General

### Issue(s) this completes

Fixes #4608

### DIFF
--- a/server/app/services/program/ActiveAndDraftPrograms.java
+++ b/server/app/services/program/ActiveAndDraftPrograms.java
@@ -30,7 +30,7 @@ public final class ActiveAndDraftPrograms {
    * state. Since a ProgramService argument is included, we will get the full program definition,
    * which includes the question definitions.
    */
-  public static ActiveAndDraftPrograms buildFromCurrentVersionsWithProgramServiceQuestionLoad(
+  public static ActiveAndDraftPrograms buildFromCurrentVersionsSynced(
       ProgramService service, VersionRepository repository) {
     return new ActiveAndDraftPrograms(
         repository.getActiveVersion(), repository.getDraftVersion(), Optional.of(service));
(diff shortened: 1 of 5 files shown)


server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -27,27 +27,46 @@ public final class ActiveAndDraftPrograms {
...
    * Queries the existing active and draft versions and builds a snapshotted view of the program
-   * state.
+   * state. Since a ProgramService argument is included, we will get the full program definition,
+   * which includes the question definitions.
    */
   public static ActiveAndDraftPrograms buildFromCurrentVersions(
       ProgramService service, VersionRepository repository) {
     return new ActiveAndDraftPrograms(
-        service, repository.getActiveVersion(), repository.getDraftVersion());
+        repository.getActiveVersion(), repository.getDraftVersion(), Optional.of(service));
```
bion, line 36: **suggestion (blocking):** for both public static methods, rename them so the method name is explicit around the behavior with the `ProgramService` and syncing questions. That way callers will know what's happening without having to read the comments.

bion (changes requested):

server/app/services/program/ActiveAndDraftPrograms.java
dkatzz: Updated to `buildFromCurrentVersionsWithoutQuestionLoad` and `buildFromCurrentVersionsWithProgramServiceQuestionLoad`. Those names are pretty long though so open to other name ideas. I was thinking buildFromCurrentVersions Shallow and Deep could work as well. Let me know what you think.

bion: synced / unsyced? that would match the verb used inside program service `syncProgram` etc

dkatzz: Sounds good, updated!

bion (approved):

PR #4591, opened 2023-04-05
Evenly distribute tests in `bin/run-browser-tests-ci-batch`

MichaelZetune:
### Description

This cleans up the `bin/run-browser-tests-ci-batch` script to:

* Modularize logic in functions
* Ensure that each GitHub runner participates in running tests.

Prior to this PR, if there were 45 tests and 10 runners, we would calculate the tests per runner as `batch_size = 45 // 10 = 5`. The first 9 runners would take 5 tests each, and the **10th runner would get none**.

Now, we assign a test to a given runner `N` if the index of the test index `i` satisfies the condition `i % BATCH_COUNT == N - 1`. For example if there are 45 tests and 10 runners,

* The "0th" runner will get tests 0, 10, 20, 30, and 40
* The "1st" runner will get tests 1, 11, 21, 31, 41
* ...
* The "8th" runner will get tests 8, 18, 28, 38
* The "9th" runner will get tests 9, 19, 29, 39

[Paste of all tests per batch after this PR](https://pastebin.com/BNX0VT7z)

In this case the first 5 runners get 5 tests each, and the last 5 runners get 4 tests each. This is a better distribution than before.

### DIFF
--- a/bin/run-browser-tests-ci-batch
+++ b/bin/run-browser-tests-ci-batch
@@ -4,43 +4,61 @@
 # DOC: 1-indexed batch number.
 
 import glob
-import math
 import subprocess
 import sys
 
 BATCH_COUNT = 10
 
 
-def bad_batch_number_input(batch_num_input):
+def handle_bad_input(batch_num_input):
     print(f"Invalid batch number: {batch_num_input}", file=sys.stderr)
     exit(1)
 
 
-if len(sys.argv) < 2:
-    bad_batch_number_input("none")
+# Gets the batch number from the command line arguments. The batch number is *one-indexed*.
+def get_batch_number():
+    if len(sys.argv) < 2:
+        handle_bad_input("none")
+    try:
+        batch_number = int(sys.argv[1])
+    except ValueError:
+        handle_bad_input(sys.argv[1])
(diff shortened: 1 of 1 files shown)


bin/run-browser-tests-ci-batch
```diff
@@ -4,43 +4,61 @@
...
-test_files.sort()
-batch_size = math.ceil(len(test_files) / BATCH_COUNT)
 
-batch_contents = test_files[(batch_number - 1) * batch_size:batch_number *
-                            batch_size]
+# Returns a subset of the test files, based on the batch number. This ensures we
+# are evenly distributing the test across all available GitHub Actions runners.
+def get_batch():
+    all_test_files = get_all_test_files()
+    batch_files = []
```
nb1701, line 43: **suggestion:** If I'm remembering my python correctly, I think you could do something like `all_test_files[get_batch_number()-1::BATCH_COUNT]` to get every nth index.

nb1701 (approved):

bin/run-browser-tests-ci-batch
MichaelZetune: Nice catch!

PR #4594, opened 2023-04-05
Lowering token permissions in actions

gwendolyngoetz:
### Description

Setting the token permissions to a lower level at the behest of the code scanning alert 

- https://github.com/civiform/civiform/security/code-scanning/346
- https://github.com/civiform/civiform/security/code-scanning/361

#### General


bion (approved):

bion (changes requested):
@avrittrohwer do these actions need write permissions?

avrittrohwer:
They just need to be able to read the secret so I think read-only will work

PR #4602, opened 2023-04-06
Add `MessagesTest` test to enforce `messages` and `messages.en-US` match

MichaelZetune:
### Description

Another invariant we should enforce in our language files is that `messages` and `messages.en-US should be exactly the same. This PR adds a test that checks for this.

### Checklist

- [x] Add `messages_valuesEqualMessagesEnUsValues()` test
- [x] Fix lines that are breaking this invariant.


server/conf/messages.en-US
```diff
@@ -46,7 +46,7 @@ content.addressEnteredWithQuestionName={0} entered:
 content.adminLoginPrompt=Looking for something else?
 content.alternativeLoginPrompt=Don''t have an account?
 content.benefits=Get benefits
-content.cannotCheckEligibility=We’re working to fix this issue quickly. Try again at a later time.
```
avaleske, line 49: Should we add a test for curly quotes?

avaleske (approved):

server/conf/messages.en-US
MichaelZetune: Not a bad idea - I wrote it down to consider for a future PR

PR #4623, opened 2023-04-10
Make a Button AutoValue class to simplify button creation

MichaelZetune:
### Description

This is the first PR towards consolidating all button-related logic into a new `Button.java` class.

#### What are we accomplishing?

Currently, buttons are created in a very ad-hoc way, with little consistency. For example, the `ApplicantStyles.java` class has many style variables to represent the same style in many places. In this PR and future ones related to #4657, we consolidate all this logic into a new `buttons` module with a few Java classes.

### Checklist

- [x] Create `Button` AutoValue class that consolidates creation and management of buttons.
- [x] Create `ButtonStyle` enum and `ButtonStyles` class to centralize styling of buttons.
- [x] Create `ButtonAction` `AutoOneOf` to represent the different actions that a client may have for a button; currently these are redirect, submit, and submit with a form ID, which represent the existing uses of buttons.
- [x] Migrate many **but not all** buttons (yet) to the new classes.
- [x] Add TODOs for remaining cleanup to perform for this refactor. The full refactor is tracked in #4657.

#### Future changes

* Resolve all TODOs, including converting **all** buttons to the new `Button` class.
* Remove all uses of `ButtonStyles.java`, which includes the Tailwind style sets, to `ButtonStyle.java` alone, so all clients **must** use the new button library to style buttons instead of using `ButtonsStyles` like they are now.
* Remove all button helper methods like `submitButton()` and `button()`.

I'll prioritize finishing these asap as I don't want to leave two ways of creating buttons in the codebase for long.

#### User visible changes

None; this should be a code-only change.

### Issue(s) this completes

Progress towards #4657.


server/app/views/components/buttons/ButtonAction.java
```diff
@@ -0,0 +1,36 @@
...
+    SUBMIT_WITHOUT_FORM_ID
+  }
+
+  public abstract Action action();
+
+  public static ButtonAction ofRedirect(String redirectUrl) {
+    return AutoOneOf_ButtonAction.redirect(redirectUrl);
+  }
+
+  public static ButtonAction ofSubmitWithFormId(String formId) {
```
reemaka, line 23: Could you instead overload a method name for the versions with and without form ID, e.g. ofSubmit() and ofSubmit(String formId)?

server/app/views/components/buttons/ButtonAction.java
```diff
@@ -0,0 +1,36 @@
...
+public abstract class ButtonAction {
+  public enum Action {
+    REDIRECT,
+    SUBMIT_WITH_FORM_ID,
+    SUBMIT_WITHOUT_FORM_ID
+  }
+
+  public abstract Action action();
+
+  public static ButtonAction ofRedirect(String redirectUrl) {
```
reemaka, line 19: Is the "of" prefix standard? Feels more readable to me to omit it

reemaka (changes requested):
I see the value in consolidating the ButtonStyles, but as is I'm wondering what the value is in the Button AutoValue class itself is. It seems like the same amount of work and just as standardized to write:
```
button().withId(id).withClasses(ButtonStyle.SOLID_BLUE).withText(text).withType("submit").withForm(form);
```
As it is to write:
```
Button.setId(id).setStyle(ButtonStyle.SOLID_BLUE).setText(text).setButtonAction(ButtonAction.ofSubmitWithFormId(form);
```

I can see value in consolidating repeated logic into Button.java, e.g. if I'm able to set the button type to submit and then the default styling for all submit buttons get applied, but right now you still have to do that manually everywhere.

Let me know if I'm missing something about the long-term vision for this and what problem it's intended to solve.

MichaelZetune:
> I see the value in consolidating the ButtonStyles, but as is I'm wondering what the value is in the Button AutoValue class itself is. It seems like the same amount of work and just as standardized to write:
> 
> ```
> button().withId(id).withClasses(ButtonStyle.SOLID_BLUE).withText(text).withType("submit").withForm(form);
> ```
> 
> As it is to write:
> 
> ```
> Button.setId(id).setStyle(ButtonStyle.SOLID_BLUE).setText(text).setButtonAction(ButtonAction.ofSubmitWithFormId(form);
> ```
> 
> I can see value in consolidating repeated logic into Button.java, e.g. if I'm able to set the button type to submit and then the default styling for all submit buttons get applied, but right now you still have to do that manually everywhere.
> 
> Let me know if I'm missing something about the long-term vision for this and what problem it's intended to solve.


Understandable point. Here's my take on the value of this extra layer of abstraction:

- We can enforce required pieces of every button, such as making sure every client has a standardized style applied. With the first approach you mentioned, clients are still free to just style buttons as they like, and I'm pretty sure we will backslide back to it sooner or later.
- We can enforce a small set of actions. IIUC there are many ways to set actions for buttons - URL redirect, multiple types of submits, no action in some cases (like closing a modal), etc. With this approach we can clearly document the subset of possible button actions we are using in CiviForm and have static references to other buttons that are doing the same thing
- We can get rid of the utility classes like [`submitButton()`](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/BaseHtmlView.java#L61) and [`redirectButton()`](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/BaseHtmlView.java#L69), which **technically** reduce duplication and centralize logic, but do so in a messy way right now IMO - [example](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/admin/questions/QuestionEditView.java#L269).
- Any logic we want to consolidate into the Button itself in the future can be done there, especially conditional logic that can't be done with Tailwind classes.

server/app/views/components/buttons/Button.java
```diff
@@ -0,0 +1,92 @@
...
+
+  // Required. What to do when the button is clicked.
+  public abstract ButtonAction buttonAction();
+
+  // Custom classes to be added to the button. Optional. Should be used sparingly;
+  // significant changes to CiviForm buttons should be made to ButtonStyles.java instead.
+  public abstract Optional<String[]> customClasses();
+
+  // Forces us to set an ID at compile-time.
+  public interface RequiredId {
```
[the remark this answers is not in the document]
MichaelZetune: It enforced that clients set the fields at compile time. See https://github.com/google/auto/blob/main/value/userguide/builders-howto.md#-create-a-step-builder

server/app/views/components/buttons/ButtonAction.java
MichaelZetune: It's recommended by the how-to guide:
https://github.com/google/auto/blob/main/value/userguide/howto.md#-make-a-class-where-only-one-of-its-properties-is-ever-set

Open to changing if you think it's better.

server/app/views/components/buttons/ButtonAction.java
MichaelZetune: The method names need to match the enum above. But I can change submitWithoutFormId to just submit. Done.

server/app/views/components/buttons/Button.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.google.auto.value.AutoValue;
+import j2html.tags.specialized.ButtonTag;
+import java.util.Optional;
+import views.style.StyleUtils;
+
+@AutoValue
+public abstract class Button {
+
+  // Required. The id of the button.
```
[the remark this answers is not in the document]
MichaelZetune: Done.

server/app/views/BaseHtmlView.java
```diff
@@ -77,12 +82,14 @@ protected static ButtonTag redirectButton(String id, String text, String redirec
    *
    * @return The element itself.
    */
+  // TODO(#4657): remove this method.
   protected static <T extends Tag> T asRedirectElement(T element, String redirectUrl) {
```
reemaka, line 86: Heads up this is also called on a non-button object [here](https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/views/admin/questions/QuestionsListView.java?L327)

reemaka (approved):
Thanks for expanding. I do still worry about it being overly restrictive, e.g. you'd need quite a few changes to this to get [this button](https://sourcegraph.com/github.com/civiform/civiform/-/blob/server/app/views/admin/programs/ProgramSettingsEditView.java?L50) working with the Button class, but I'm fine with migrating to this class if you feel strongly about it.

> We can get rid of the utility classes like [submitButton()](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/BaseHtmlView.java#L61) and [redirectButton()](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/BaseHtmlView.java#L69), which technically reduce duplication and centralize logic, but do so in a messy way right now IMO - [example](https://github.com/civiform/civiform/blob/0ed40707f7e79645c1cea546ebf5463f57e15940/server/app/views/admin/questions/QuestionEditView.java#L269).

nit: I think that example could actually be cleaned up by using redirectButton(). Right now it uses asRedirectElement.

PR #4653, opened 2023-04-18
Set aria-required for required questions.

### LINKED ISSUE #3955: Mark required fields with proper html/aria attributes for screen readers
Splitting off from https://github.com/civiform/civiform/issues/2629:

Programmatically, we should mark the html as required, either with (1) html required attribute, or (2) aria-required attribute. The html attribute is preferred, because it has baked in information.

reemaka:
### Description

Set aria-required for required questions to help screen reader users be able to tell which questions are required. The file upload question type is an exception since it doesn't have any non-hidden input tags to be able to set the aria-required attribute.

The html required attribute was proposed as another solution here, but that causes the browser to show a built-in "Please fill out this field" tooltip on form submit. We already do our own validation, so I'm using aria-required to avoid side effects on the UI.

#### General

### Issue(s) this completes

Fixes #3955

### DIFF
--- a/server/app/views/admin/programs/ProgramFormBuilder.java
+++ b/server/app/views/admin/programs/ProgramFormBuilder.java
@@ -131,6 +131,7 @@ abstract class ProgramFormBuilder extends BaseHtmlView {
                 FieldWithLabel.radio()
                     .setId("program-display-mode-public")
                     .setFieldName("displayMode")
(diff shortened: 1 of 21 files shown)


dkatzz:
Could you tag this with bug or feature? Otherwise LGTM!

PR #4656, opened 2023-04-18
Fix nav links for CF admins on applications and reporting pages.

### LINKED ISSUE #4648: Navigation bar in application review page has incorrect links
**Describe the bug**
If you're logged in as a CiviForm admin, the Programs link in the nav bar usually leads to the program dasboard page. However, if you view the applications for a program, the Programs link in the nav bar instead links to a readonly view (maybe the Program Admins' view). Also, some of the links are missing.

**To Reproduce**

1. Log in as CiviForm and Program Admin
2. Note that you're on the Program dashboard page with the ability to edit programs and there are 5 links in the nav bar.
3. Click on "Applications" for a program from the dropdown.
4. Note that the nav bar now only has 2 links, Programs and Reporting.
5. Click on Programs.
6. Note that you're on a readonly view of the Programs page.

**Expected behavior**
For CiviForm admins, the Programs page link should always lead to the Program Dashboard. Also, all of the nav links should always be present.

reemaka:
### Description

The application and reporting pages were showing the Program Admin view for Program and CiviForm Admins. This fixes it by checking if the user is also a CF admin before setting it to use the Program Admin view.

#### General

### Issue(s) this completes

Fixes #4648


dkatzz:
Can we add some tests for this? Maybe a browser test asserting the right thing appears in the various situations?

server/app/controllers/admin/AdminApplicationController.java
```diff
@@ -508,6 +509,7 @@ public Result index(
     return ok(
         applicationListView.render(
             request,
+            getCiviFormProfile(request),
```
dkatzz, line 512: **question:** What happens if you set this before this return statement, like:

`CiviFormProfile profile = profileUtils.currentUserProfile(request).orElseThrow();..`

And then used that variable here?

Not sure why that would make a difference, but just curious if it does.

reemaka: Browser tests still fail with that change.

Still haven't tracked down the exact issue. I'm thinking it might have something to do with AdminLayout being a Singleton. Now that I'm thinking about it, I think it's kind of brittle to rely on the callers to call setOnlyProgramAdminType() anyway. Thinking through what a better option would be. I would want to make the render() classes take in the CiviForm profile so that the AdminLayout can check the profile itself, but those are overridden methods from the base class BaseHtmlLayout. Thinking of instead at least changing setOnlyProgramAdminType to setCiviFormProfile and then checking the profile status before rendering.

dkatzz: Yeah that makes sense! Does getting the profile within the layout work as opposed to passing it through?

reemaka: No unfortunately bc it needs the request to get the profile

dkatzz (approved):
Awesome, thanks for fixing this!!

PR #4671, opened 2023-04-24
Create `enforce-labels.yml`

michaelzetune:
### Description

Enforces we don't forget to add labels to PRs. Adds https://github.com/marketplace/actions/enforce-pr-labels to the CIviForm repository to ensure that all PRs have any label.

Here's what the error looks like when we forget:

image

#### General

#### User visible changes

None.

#### New Features

- [x] Enforce adding labels to GitHub PRs as required by project policy

### Issue(s) this completes

None.

### DIFF
--- a/.github/workflows/enforce-labels.yml
+++ b/.github/workflows/enforce-labels.yml
@@ -1,0 +1,27 @@
+name: Label Checker
+on:
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    - reopened
+    - labeled
+    - unlabeled
+
+jobs:
+
+  check_labels:
+    name: Check labels
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Check for label
+      run: |
+        if [ ${{ github.event.pull_request.labels[@] }} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


reemaka:
Is there a way to have a label be required but not to limit it to the set you've specified? I'm imagining PRs for documentation or tech debt for example that fall outside of those

MichaelZetune:
> Is there a way to have a label be required but not to limit it to the set you've specified? I'm imagining PRs for documentation or tech debt for example that fall outside of those

I believe we can do that by simply removing the set of labels from the yml file. But then we will allow ex. "Resident feedback" tag to satisfy the check on a PR, which might be a fine tradeoff to your point. I'm ok either way - I was just echoing the default issue text into the acceptable labels list.

reemaka:
I'm not sure if that list was supposed to mean your label had to be one of those or if it was just listing examples. So maybe it's worth checking with others to see what the intention was. Personally I would vote for it to be unrestricted and have this just serve as a reminder to set it to something without being too limited, but I'm fine with whatever.

reemaka (approved):

MichaelZetune:
Looks like this is failing on formatting, and `bin/fmt` doesn't touch `.yml` files. Will look into fixing CI so that this can be submitted, soon.

PR #4678, opened 2023-04-25
Add required indicator to first and last name fields.

### LINKED ISSUE #4673: Add asterisk to required first and last name
The "name" question type does not have use asterisks to show the first and last name are required.

If a name is required, show an asterisk next to First and Last Name fields. (middle is not required)

**Additional context**
Something like this:

image

**Done when**
User can see that the first and last name is required.

reemaka:
### Description

Add required indicator to first and last name fields.

#### General

### Issue(s) this completes

Fixes #4673


browser-test/image_snapshots/admin_program_translation_test/applicant-toast-error.png
jessbumblebee, line 1: It may be unrelated to this change (haven't looked at the rest yet): I noticed that the yellow message at the top says "Lo sentimos, este programa ne esta traducido por completo al ingles"

The translation seems wrong
It says that the page wasn't completely translated to English, when it should probably read "to Spanish" or "from English

Would you mind checking if it is the test or the actual translation that is wrong and filing a bug?

server/app/views/components/FieldWithLabel.java
```diff
@@ -280,12 +280,12 @@ public FieldWithLabel setDisabled(boolean disabled) {
...
   /**
-   * Set this field to required. Causes a required question indicator to be displayed when this
-   * field is rendered.
+   * Set this field to required or optional. If required, a required question indicator will be
+   * displayed when this field is rendered.
    */
-  public FieldWithLabel isRequired() {
-    this.required = true;
-    setAriaRequired(true);
+  public FieldWithLabel setRequired(boolean isRequired) {
```
jessbumblebee, line 286: Nice rename.

suggestion(none-blocking as not a a regression):
Is there a reason why we need to methods (setRequired and setAriaRequired)?
If the answer is yes, could you expand the comments to make that more clear?

jessbumblebee (approved):

browser-test/image_snapshots/admin_program_translation_test/applicant-toast-error.png
reemaka: Good catch, looks like it is wrong in the actual translation. Filing a bug.

server/app/views/components/FieldWithLabel.java
reemaka: Yes, the setRequired one changes the UI and the setAriaRequired doesn't change the UI but just adds the aria label for screen readers. This is useful sometimes when we use multiple fields for the same question and we don't need the asterisk to show up on each and every field.

PR #4690, opened 2023-04-25
Allow reordering answer options for multiple choice questions

### LINKED ISSUE #4558: Reorder values within questions
As a CiviForm Admin, when I add a new value to a question with lists (drop down, radio, check box), I don't have a way to rearrange the order. The current workaround is to delete the existing list and rebuild parts or all of it to accommodate the new value. 

A way to move values within a question either up or down, similar to the reordering function of questions within programs. 

**Additional context**
Screenshot attached

Ideally, as a CiviForm Admin, I would want to be able to add new values to questions and then use arrows to move the value to the correct order, such as alphabetical

**Done when**
I can move values either up or down a list.

reemaka:
### Description

Makes it easy to change the order of answer options for multiple choice question types (dropdown, checkbox, radio) by adding up and down buttons to move the option. For code simplicity, there is still an up button on the first option and a button option on the last option, but they don't do anything. Re-ordering the HTML elements causes Play Java to bind the form with the fields in the correct order and for the configuration to be stored correctly.

## Release notes

#### General

### Issue(s) this completes

Fixes #4558


avaleske (approved):
LGTM!

If we're interested, we could probably use the `::first-of-type` and `:last-of-type` psuedo-classes to hide the up and down arrows from the first and last buttons.

I'm not sure exactly how to do it in tailwind, since the buttons one level down from the button container, but I think it's using the `group-` styles:
https://tailwindcss.com/docs/hover-focus-and-other-states

reemaka:
Hmm yeah I can't seem to get that working correctly with Tailwind.

PR #4701, opened 2023-04-26
Add `auth.applicant_auth_provider_name` to allow civic entities to customize the name of their auth provider

MichaelZetune:
### Description

For some civic entities, it doesn't make sense for us to tell users to log in to the "City of X" portal. Instead, we will allow this name to be customizable.

Example:

image

## Release notes

Allow civic entities to customize the authentication provider name displayed when prompting users to log in.

#### General

#### User visible changes

#### New Features

- [n/a] Conditioned new functionality on a [FeatureFlag](https://docs.civiform.us/contributor-guide/developer-guide/feature-flags)
- [n/a] Wrote browser tests with the feature flag off and on, etc.
- [x] Update browser tests

### Issue(s) this completes

None.


server/conf/env-var-docs.json
```diff
@@ -46,6 +46,10 @@
             "description": "URI to create a new account in the applicant identity provider.",
             "type": "string"
           },
+          "APPLICANT_PORTAL_NAME": {
+            "description": "The name of the portal that applicants log into.",
```
avaleske, line 50: Would it be helpful to have the an example of how this will be used in a sentence? eg "The name of the portal that applicants log into. Used in sentences like 'Log into your APPLICANT_PORTAL_NAME account.'"?

avaleske (approved):

server/conf/env-var-docs.json
MichaelZetune: Done.

PR #4686, opened 2023-04-25
Add login prompt when guests click Apply

MichaelZetune:
### Description

As per #4365, we want to prompt guests to log in when they start to apply for a program. In this PR, we check if an applicant is a guest and add the modal before taking them to the application.

Additionally, we refactor the common logic in `createLoginPromptModal()` so into its own class, as it is being used in three places and needs to be a bit more generalizable.

apply-program-login-prompt

## Release notes

Adds a login upsell prompt when to encourage guests to log in at the time they click "Apply" on a program.

#### General

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary
   - Not neccessary

#### User visible changes

- [x] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method
- [x] Manually tested at 200% size
- [x] Manually evaluated tab order

### Issue(s) this completes

Fixes #4365

PR #4704, opened 2023-04-27
Remove Login Form in CiviForm

michaelzetune:
### Description

As per Language, Login, Landing improvements, we are no longer going to have a login form in CiviForm. This means that users will be directed to the programs page directly, and will be logged in as a guest.

**In this PR, we only remove the login form itself.** A future PR will remove the language form, leaving users to select a language from the header.

**[Relevant Figma Mocks](https://www.figma.com/file/tUV2i9QdFU2zQqZWg7Xl5g/CiviForm---Login%2C-Language-and-Landing-page-improvements-(public-link)?node-id=215-1600&t=VO87CqGBYavjSUzu-0)**

Planning to wait until #4686 is merged before merging this one.

## Outline

* Introduce `BYPASS_LOGIN_LANGUAGE_SCREENS` feature flag to allow civic entities to flip these changes on when they're ready. 
   * Enable for dev CiviForm and browser tests.
   * We will not encourage entities to flip the flag until the language screen is also removed, as implied by the name of the flag.
* Extract `debugContent()` method into its own file so it can be used in both the login form (for instances where the flag is off) and the language selector form (when the flag is on)
* In `HomeController`, the controller responsible for redirecting the user when they navigate to the root (`/`) part of the site:
  * Continue redirecting to the login form if the new feature flag is off
  * Redirect to the programs page if the flag is on
* All browser tests and Java tests assume that the flag is on. I considered parameterized testing for both cases for everything, but this was a substantial amount of work and the login page is unlikely to change between now and when we fully delete it.

The debug content that was originally on the login page is now on the language selection page. Once the language selection page is removed, it'll be on its own dedicated page.

image

## Release notes

Remove the log in form and direct users to the language page directly, when new `BYPASS_LOGIN_LANGUAGE_SCREENS` flag is on.

#### General

- [x] Verify the login tests work on staging and Seattle staging environments before merging, if possible
   - Confirmed: https://pastebin.com/c9f9EyNi

#### User visible changes

- [x] ~Wrote~ **Removed login** browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method
- [n/a] Manually tested at 200% size
- [n/a] Manually evaluated tab order

#### New Features

None.

### Issue(s) this completes

Relates to #4705


browser-test/src/landing_page.test.ts
```diff
@@ -1,22 +1,13 @@
...
   createTestContext,
   enableFeatureFlag,
   disableFeatureFlag,
   validateAccessibility,
   validateScreenshot,
   TestContext,
 } from './support'
-import {TEST_USER_AUTH_STRATEGY} from './support/config'
 
 function sharedTests(ctx: TestContext, screenshotName: string) {
```
MichaelZetune, line 10: The landing page (which is language selection, as of this PR) no longer has login options directly on the page; they are in the header.

browser-test/src/support/index.ts
```diff
@@ -319,10 +322,8 @@ export const setLangEsUS = async (page: Page) => {
  *     login can be initiated from different pages, for example after program
  *     submission.
  */
```
MichaelZetune, line 324: There are no usages of `loginButton` below.

We also change it from `button` to `a` so that the test code always clicks on the "Log In" link on the header, rather than the "Log In" button on the programs page which is sometimes but not always available:

image

server/app/controllers/HomeController.java
```diff
@@ -51,11 +47,13 @@ public HomeController(
   public CompletionStage<Result> index(Http.Request request) {
     Optional<CiviFormProfile> maybeProfile = profileUtils.currentUserProfile(request);
```
MichaelZetune, line 49: The `loginForm` endpoint below is gone. We replace it with the endpoint that the login form directs users to when they used to click "Continue as Guest".

server/app/controllers/applicant/RedirectController.java
```diff
@@ -72,13 +70,6 @@ public RedirectController(
   public CompletionStage<Result> programBySlug(Http.Request request, String programSlug) {
     Optional<CiviFormProfile> profile = profileUtils.currentUserProfile(request);
 
-    if (profile.isEmpty()) {
```
MichaelZetune, line 75: We currently ask users to log in whenever they go to a program slug. For example, open this in incognito:

https://civiform.seattle.gov/applicants/25936/programs/1384

Now, we just assume the user is a guest and take them to the program.

server/app/featureflags/FeatureFlag.java
```diff
@@ -31,10 +31,7 @@ public enum FeatureFlag {
   NONGATED_ELIGIBILITY_ENABLED,
 
   // Phone number question type.
-  PHONE_QUESTION_TYPE_ENABLED,
-
-  // New login form
-  NEW_LOGIN_FORM_ENABLED;
```
MichaelZetune, line 37: The "new login form" never came to fruition because we are getting rid of the form altogether; removing the feature flag here.

server/app/views/applicant/ApplicantInformationView.java
```diff
@@ -93,7 +117,65 @@ public Content render(
             .addMainStyles(ApplicantStyles.MAIN_APPLICANT_INFO)
             .addMainContent(h1(title).withClasses("sr-only"), formContent);
```
MichaelZetune, line 118: Below, we add all the dev information that we previously had on the login page to the language selection page. See the PR description for a screenshot.

server/app/views/LoginForm.java
```diff
@@ -248,43 +233,4 @@ private ATag adminLink(Messages messages) {
         .withHref(routes.LoginController.adminLogin().url())
         .withClasses(BaseStyles.ADMIN_LOGIN);
   }
```
MichaelZetune, line 235: Moved the below to `DebugContent.java`

server/app/views/LoginForm.java
```diff
@@ -75,21 +75,6 @@ public LoginForm(
...
 
   public Content render(Http.Request request, Messages messages, Optional<String> message) {
```
MichaelZetune, line 77: The "new login page" never came to fruition, so remove the branched logic along with the `NEW_LOGIN_FORM_ENABLED` flag while we're here.

server/conf/helper/feature-flags.conf
```diff
@@ -44,6 +44,7 @@ nongated_eligibility_enabled = ${?NONGATED_ELIGIBILITY_ENABLED}
 phone_question_type_enabled = false
 phone_question_type_enabled = ${?PHONE_QUESTION_TYPE_ENABLED}
```
MichaelZetune, line 46: We'll make this off by default and ask entities to turn it on one at a time.

server/app/controllers/HomeController.java
```diff
@@ -46,16 +51,25 @@ public HomeController(
     this.languageUtils = checkNotNull(languageUtils);
     this.faviconURL =
         Optional.ofNullable(Strings.emptyToNull(configuration.getString("whitelabel.favicon_url")));
+    this.featureFlags = featureFlags;
```
bion, line 54: **nitpick:** 
```suggestion
    this.featureFlags = checkNotNull(featureFlags);
```

server/app/controllers/HomeController.java
```diff
@@ -46,16 +51,25 @@ public HomeController(
...
+    boolean bypassLogin =
+        featureFlags.getFlagEnabled(request, FeatureFlag.BYPASS_LOGIN_LANGUAGE_SCREENS);
+
+    // If the user isn't already logged in within their browser session, consider them a guest.
     if (maybeProfile.isEmpty()) {
-      return CompletableFuture.completedFuture(
-          redirect(controllers.routes.HomeController.loginForm(Optional.empty())));
+      return bypassLogin
+          ? CompletableFuture.completedFuture(
+              redirect(routes.CallbackController.callback(GuestClient.CLIENT_NAME).url()))
```
bion, line 67: **issue (blocking):** this will create a new `Account` record every time someone visits the landing page, regardless of whether they continue as a guest or login to their account. Can we delay doing that until they proceed to apply for something as a guest?

server/app/controllers/applicant/RedirectController.java
bion: question (blocking): Does the design not take into consideration users with accounts who are not logged in who land on a program slug?

server/app/views/dev/DebugContent.java
```diff
@@ -0,0 +1,62 @@
+package views.dev;
+
+import static j2html.TagCreator.div;
+import static j2html.TagCreator.p;
+
+import auth.FakeAdminClient;
+import j2html.tags.specialized.DivTag;
+import views.BaseHtmlView;
+
+public class DebugContent extends BaseHtmlView {
```
bion, line 10: **suggestion (blocking):** add javadoc for this class

bion, line 10: **nitpick:** 
```suggestion
public final class DebugContent extends BaseHtmlView {
```

server/app/views/dev/DebugContent.java
```diff
@@ -0,0 +1,62 @@
...
+                "feature-flags",
+                "Feature Flags",
+                controllers.dev.routes.FeatureFlagOverrideController.index().url()),
+            redirectButton(
+                "database-seed",
+                "Seed Database",
+                controllers.dev.routes.DatabaseSeedController.index().url()));
+  }
+
+  public static DivTag civiformVersionDiv(String civiformVersion, String civiformImageTag) {
```
bion, line 50: **suggestion (blocking):** make this class initializable and inject its dependencies rather than forcing the caller to inject and then pass them

bion (changes requested):

server/app/controllers/HomeController.java
MichaelZetune: Done.

server/app/controllers/HomeController.java
MichaelZetune: Edit: I wrote this comment before I introduced the feature flag. The `loginForm` endpoint **will** be done once this change is finalized and the flag is removed.

server/app/views/dev/DebugContent.java
MichaelZetune: Done.

server/app/views/dev/DebugContent.java
MichaelZetune: Done.

server/app/controllers/HomeController.java
MichaelZetune: To clarify, do you mean a new `Applicant` record? Since guests haven't created an `Account` yet?

I need to think about this problem a bit more. It seems to me that having an additional user mode other than "logged in" and "guest" would add a significant amount of complexity to the rest of the application. For example, in many places we check whether a user "is a guest" and assume that if not, they aren't logged in (header bar is an instance of this).

While it's true that visiting the landing page as a new user creates a guest record for them, refreshing the page or revisiting it later doesn't (IIUC). Also, I anecdotally suspect that most guests will click "Apply" on **something**, if not out of sheer curiosity to see what happens, and that would render this optimization useless.

How significant of a problem is this in your opinion to warrant adding this new user state? Am I missing something?

server/app/controllers/applicant/RedirectController.java
bion: That sounds good to me but we should should confirm with @isabelcshaw and @sijiayam

server/app/controllers/HomeController.java
bion: > Also, I anecdotally suspect that most guests will click "Apply" on something, if not out of sheer curiosity to see what happens, and that would render this optimization useless.

@gwendolyngoetz @elisekalstad would it be possible to check Seattle GA to get a sense of unique monthly visitors to the landing page vs unique monthly visitors to the sum of all the `/programs/<programs-slug>` routes?

bion: discussed out-of-band in a meeting, good to proceed

server/app/controllers/applicant/RedirectController.java
bion: Resolved in an out-of-band discussion

bion (approved):

server/app/controllers/applicant/RedirectController.java
MichaelZetune: Follow up PR is tracked as part of https://github.com/civiform/civiform/issues/4705.

PR #4712, opened 2023-05-01
Check for slanted quotes in `MessagesTest`, fix one instance

MichaelZetune:
### Description

Slanted quotes like `“` instead of `"` show up often in messages when they shouldn't. This can lead to bad output when this text is displayed. For example:

```
content.mobileFileUploadHelp=Ginagamit ang iyong telepono? Binibigyang-daan ka rin ng “Pumili ng File” na gamitin ang camera ng iyong telepono para mag-upload ng mga dokumento.
```

Shows up as:

```
Ginagamit ang iyong telepono? Binibigyang-daan ka rin ng âPumili ng Fileâ na gamitin ang camera ng iyong telepono para mag-upload ng mga dokumento.
```

Fix these instances and write a test to check for this. Notably, in Java we must check for `â` not `“` for this to work. I'm not sure why, but I've checked that it works as intended.

## Release notes

None.

#### General

   - Not necessary

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

Relates to #4520.


server/test/conf/MessagesTest.java
```diff
@@ -92,6 +92,32 @@ public void messages_valuesEqualMessagesEnUsValues() throws Exception {
         .containsAllEntriesOf(entriesInEnUsLanguageFile);
   }
```
MichaelZetune, line 94: Notably, below we check the primary language file many times due to parameterization. I realize we could extract this into its own test, but I don't see the need. Feel free to push back.

avaleske: Mmm I don't feel terribly strongly but since you suggested it, having it be its own test might be slightly cleaner.

Also it would be good to test for smart/curly single quotes as well, and it might be more reliable to search for the specific unicode codes:

Character | Unicode
-- | --
“ | \u201C
” | \u201D
‘ | \u2018
’ | \u2019

avaleske (approved):
Some thoughts about unicode characters but otherwise lgtm

server/test/conf/MessagesTest.java
MichaelZetune: Refactored out the primary language check into two methods.

About the unicocde codes, that's an excellent idea, but because of the way the `messages` files are being read in I don't seem to be able to check for them. They get converted somehow into the weird `â` character (I guess in a similar way as how `''` is changed to `'`.

That being said, all four cases are caught by this weird character, so I guess it's working.

Issue #4705, opened 2023-04-27
Remove Login Page and Language Selection Page


michaelzetune:
As per Language, Login, Landing improvements, we are no longer going to have a login form in CiviForm. This means that users will be directed to the programs page directly, and will be logged in as a guest.

[Relevant Figma Mocks](https://www.figma.com/file/tUV2i9QdFU2zQqZWg7Xl5g/CiviForm---Login%2C-Language-and-Landing-page-improvements-(public-link)?node-id=215-1600&t=VO87CqGBYavjSUzu-0)

PRs:
- #4686
- #4704
- #4720
- #4771
- #4780
- #4801

Still TODO:

- [x] Add a login prompt modal to the program slug page so that users are prompted to login in case they have an account, if they are taken to the application directly by a link.
   - Done in #4686
- [x] Make the login prompt modal only appear once per guest session.
   - #4771
- [x] Make a toast appear when End Session is clicked to show the user their session is finished.
   - Originally, we were going to close the tab. As it turns out, this is [prohibited in most modern web browsers](https://www.computerhope.com/issues/ch000178.htm).
   - #4780
- [x] Show a login prompt on the program slug page if the user is a guest
   - #4801
- [x] Fix all instances of `TODO(#4705)`, clean up code
   - #4928
   - #4929
   - #4935
- [x] Remove the `STAGING_DISABLE_APPLICANT_GUEST_LOGIN` env var. This is because when this var is enabled, all that happens is we remove the guest button from the login form. Since the login form won't be available anymore, it doesn't make sense to keep this. This can be done at the same time we delete `LoginForm.java`.
   - Also done in #4928

Let Seattle know when this issue is done for #4729.

PR #4720, opened 2023-05-01
Remove language form when going to CiviForm

michaelzetune:
### Description

This PR removes the language form when opening CiviForm. Users are already logged in as a guest by default; now they will bypass choosing a language. Instead, the default language from their browser will be used (much like we preselected the browser's language in the language selection radio buttons) and they will continue to be able to change the language in the header bar as needed.

## Release notes

Remove language form from CiviForm when `BYPASS_LOGIN_LANGUAGE_SCREENS` is enabled.

## Outline

* Bypass `ApplicantInformationView`, which encapsulated the language selection page, when `BYPASS_LOGIN_LANGUAGE_SCREENS` is enabled
* Move debug tools from the language selection page to the home (index page), when `BYPASS_LOGIN_LANGUAGE_SCREENS` is enabled
* Use the browser's preferred language to automatically get the user's language preference and set it for CiviForm. Previously, we would use the browser's preferred language only to pre-select the radio button on the language selection form.
* Update `support/index.ts` `selectApplicantLanguage()` to select from the dropdown at the top instead of the language selection page. Remove language selection page-specific tests. Remove bespoke `setLangEsUS()` method.
* Change `landing_page.test.ts` to `footer.test.ts`. The was originally testing whether the version tag was appearing on the footer of the language selection screen. Now, we just test that it appears on the footer in general, and update the test name to reflect that.

### New Dev Tools

See "DevTools" link in header:

dev-tools-in-header

And modal that appears on click:

dev-tools-modal

#### General

   - Not necessary, only removing things

#### User visible changes

#### New Features

   - `BYPASS_LOGIN_LANGUAGE_SCREENS` is already there.
- [x] Wrote browser tests with the feature flag ~off and~ on, etc.

### Issue(s) this completes

Relates to #4705


server/app/controllers/applicant/ApplicantInformationController.java
```diff
@@ -52,18 +57,90 @@ public ApplicantInformationController(
...
     this.httpExecutionContext = httpExecutionContext;
     this.messagesApi = messagesApi;
     this.informationView = informationView;
     this.repository = repository;
     this.formFactory = formFactory;
     this.profileUtils = profileUtils;
     this.applicantService = applicantService;
+    this.layout = layout;
+    this.featureFlags = featureFlags;
   }
```
MichaelZetune, line 73: Below, `legacyEdit()` continues to implement the old logic, and that code path is executed when `BYPASS_LOGIN_LANGUAGE_SCREENS` is not yet enabled.

When it is enabled, we do things slightly differently:

* Instead of using the user's selected language from a form, we immediately store the user's preferred language in the database
* We pull the rest of the redirect logic from `ApplicantInformationView.java` (which represents the language selection form view) into here, so that that file can be deleted in a later PR.

server/app/views/LoginForm.java
```diff
@@ -98,7 +98,8 @@ public Content render(Http.Request request, Messages messages, Optional<String>
         div().withClasses("fixed", "w-screen", "h-screen", "bg-gray-200").with(content));
```
MichaelZetune, line 99: A small mistake from earlier: the dev tools aren't appearing on the login form. This is irrelevant when the bypass flag is on, but came up when testing Seattle staging before the flag was flipped. This file will be deleted soon.

server/app/controllers/applicant/ApplicantInformationController.java
MichaelZetune, line 1: TODO(me): fix `src/program_deep_link.test.ts`, which these changes broke.

browser-test/src/security.test.ts
```diff
@@ -39,14 +39,6 @@ describe('applicant security', () => {
     expect(response!.status()).toBe(403)
   })
 
-  it('redirects to language selection form when not logged in', async () => {
```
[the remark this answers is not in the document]
MichaelZetune: Sure! Done.

michaelzetune:
TODO:
* Check on halfway completed application case
* Login modal shows only once per user

Issue #4775, opened 2023-05-04
Use X-Forwarded-For for IP checking on AWS deployments


bion:
We currently use [`WebContext.getRemoteAddr`](https://github.com/civiform/civiform/blob/main/server/app/auth/ApiAuthenticator.java#L117) to determine the remote IP address when authenticating API requests. When CiviForm is deployed on AWS, this value will refer to internal AWS resources (such as load balancers) rather than the actual API client IP address.

When CiviForm is deployed in AWS it should instead refer to the [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) header for retrieving the client's IP address.

avaleske:
We need to split `X-Forwarded-For` on `,` and take the first one:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax

```
X-Forwarded-For: <client>, <proxy1>, <proxy2>
```

avaleske:
Confirming that this fix works against staging!

Repro:
- Fetch your IP with `curl ifconfig.me`
- Use that IP to generate an API key
- Use that API key to curl against https://staging-aws.civiform.dev/api/v1/checkAuth
- See a 200!

PR #4777, opened 2023-05-05
Parse X-Forwarded-For header

bion:
### Description

Fixes https://github.com/civiform/civiform/issues/4775, see last comment


avaleske (approved):
LGTM with the note that the Mozilla docs say that some poorly behaved proxies may have multiple `X-Forwarded-For` headers, so it may need additional adjustments depending on AWS's behavior.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#parsing

PR #4780, opened 2023-05-05
Session Ended Toast

michaelzetune:
### Description

When ending a session in CiviForm, we no longer go to the login page nor language page, which makes it less clear that the session ended. As part of the project, we were going to close the browser tab upon clicking "End Session" in order to signal that they were finished with their session.

As it turns out, this is [prohibited in most modern web browsers](https://www.computerhope.com/issues/ch000178.htm). The original UX idea before closing the tab was to show a toast, so we are going to revert to that:

image

## Release notes

Displays a toast when user ends their session and `BYPASS_LOGIN_LANGUAGE_SCREENS` is on.

#### General

   - Validates the toast appears every time we call the `logout()` helper method

#### User visible changes

#### New Features

No new feature flags.

### Issue(s) this completes

Progress towards #4705


browser-test/image_snapshots/auth_test/user-just-ended-session.png
swatkat1, line 1: This is the test for a logged in user, but it says here "You're a guest user". Is this is a screen shot after the user has clicked logout?

swatkat1 (approved):

PR #4801, opened 2023-05-08
Show a login prompt on the program slug redirect page

michaelzetune:
## Description

Before Login, Language, Landing changes, users that go to a program slug (like `https://civiform.something.gov/programs/someslug`) would be first prompted to log in if they were a guest, by being redirected to the login page.

With the login page gone, we want to prompt users to log in with the standard log in modal we have elsewhere. We want to show this modal if (1) a user was redirected by a program slug (i.e. `https://civiform.something.gov/programs/someslug` to `https://civiform.something.gov/applicants/123/programs/456`) and (2) they are a guest.

This also fixes up a test I had modified earlier to get a LLL PR out, `program_deep_link.test.ts`.

login-prompt-for-guest-users-using-program-slug

## Outline

* Simplify `AuthenticateUpsellCreator` to return a `Modal.Builder` and allow clients to modify it further, rather than relying on helper methods. Update usages.
* Add a `Optional<Modal> loginPromptModal` to `ApplicantProgramSummaryView.Params`, which governs whether to show a modal on that page when users arrive at it.
* Update `RedirectController` to add a flashing string called `"redirected-from-program-slug"` when users are sent from a program slug to an application.
* Use the flashing string in `ApplicantProgramReviewController` to add a login prompt modal when appropriate.
* Revert the changes to `program_deep_link.test.ts` made [here](https://github.com/civiform/civiform/pull/4720/files#diff-c9763a45398797693b2c3cd69956e1595245c8983e4e87d15c0484e52cf542a1). This was because logging in with the Log In button on the header can't redirect to the page it came from, but logging in with the newly added modal can.

## Release notes

Show a login prompt when guests go to the program slug page.

#### General

#### User visible changes

#### New Features

No new feature flags necessary

### Issue(s) this completes

Relates to #4705


browser-test/src/program_deep_link.test.ts
```diff
@@ -36,11 +40,40 @@ describe('navigating to a deep link', () => {
...
+    const {page} = ctx
+
+    await gotoEndpoint(page, '/programs/test-deep-link')
+    await loginAsTestUser(page, 'button:has-text("Log in")')
+    expect(await page.innerText('html')).not.toContain(
+      'Create an account or sign in',
+    )
+  })
+
+  it('as a guest user or registered user', async () => {
```
rockycodes, line 69: nitpick (non-blocking): add a more descriptive test name

rockycodes (approved):
looks great!

Issue #4734, opened 2023-05-03
Email Alerts -- auto messages missing for TIs and Clients


elisekalstad:
**Describe the bug**
When a TI applies on behalf of their client, there are two types of communications that are instantly triggered:
1. initial email (this one just says "you submitted an application to X program.."
2. default status (this one is customized for the program and gives follow-up instructions "your application to Fresh Bucks may..."

The two types of communications are not sent to both parties (the TI AND their client). Each person receives a different message.

**To Reproduce**

1. _(Make sure a program is set up with a default status)_
2. Log in as a TI
3. Apply to the program on behalf of a client 
4. See that the TI receives the **initial email** but not the default status
5. See that the client receives the **default status** but not the initial email

**Expected behavior**
Would expect the TI AND their client to receive both communications: 1) the initial email and 2) the default status message.

**Screenshots**
![image](https://user-images.githubusercontent.com/86082659/235811921-433a1d25-aaa0-49d0-ba01-dc2bff631e06.png)

![image](https://user-images.githubusercontent.com/86082659/235811948-6068204f-1062-483a-ac73-2d7aa5a82cdb.png)

![image](https://user-images.githubusercontent.com/86082659/235811974-28d4fda2-cd4f-46e7-a2a7-fcf022afcc55.png)

elisekalstad:
**REVISION**
Issue can be resolved if TI experience matches the current Client experience; i.e. TI should just get the Default Status message. 

If there is a default status, no one needs to get the initial "application received" email.

msprenke:
@nb1701 Nick can you confirm that BOTH client and TI receive additional status updates beyond the default?  example, default is waitlist then program admin changes to APPROVED.  Do they both get it?  Thanks.

PR #4805, opened 2023-05-08
Refactor email notification functions and change TI email behavior

nb1701:
### Description

This refactors the 'maybeNotify' functions that email the applicants and TIs. This combines each into one function, rather than separate functions depending on if a default status exists. Additionally, this sends the status email body to the TI as well as the applicant, rather than the default 'application received' email to the TI. This also modifies the emailing of admins to be asynchronous, to match the rest of the functions here since this is client-facing behavior.

Finally, this modifies the get-emails script so it doesn't show the progress bar when fetching emails from localstack, and adds a delete-emails script to clear out the emails stored by localstack.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4734

### DIFF
--- a/server/app/services/applicant/ApplicantService.java
+++ b/server/app/services/applicant/ApplicantService.java
@@ -652,12 +652,10 @@ public final class ApplicantService {
     CompletableFuture<Optional<Locale>> localeFuture =
         getPreferredTiLocale(tiEmail).toCompletableFuture();
     return localeFuture
-        .thenRunAsync(
-            () -> {
+        .thenAcceptAsync(
+            (localeMaybe) -> {
               // Not blocking since it already completed
-              Locale locale = localeFuture.join().orElse(LocalizedStrings.DEFAULT_LOCALE);
-              boolean useStatusMessage =
-                  status.map(s -> s.localizedEmailBodyText().isPresent()).orElse(false);
+              Locale locale = localeMaybe.orElse(LocalizedStrings.DEFAULT_LOCALE);
               Messages messages =
                   messagesApi.preferred(ImmutableSet.of(Lang.forCode(locale.toLanguageTag())));
               String subject =
@@ -665,6 +663,8 @@ public final class ApplicantService {
                       MessageKey.EMAIL_TI_APPLICATION_SUBMITTED_SUBJECT.getKeyName(),
                       programName,
                       applicantId);
+              boolean useStatusMessage =
+                  status.map(s -> s.localizedEmailBodyText().isPresent()).orElse(false);
               String message =
                   String.format(
                       "%s\n%s",
@@ -705,11 +705,9 @@ public final class ApplicantService {
       Optional<StatusDefinitions.Status> status) {
     CompletableFuture<Optional<Locale>> localeFuture =
         getPreferredLocale(applicantId).toCompletableFuture();
-    return localeFuture.thenRunAsync(
-        () -> {
-          // Java seems not to like using thenApplyAsync here for some reason I don't understand,
-          // so using thenRunAsync and joining the future.
-          Locale locale = localeFuture.join().orElse(LocalizedStrings.DEFAULT_LOCALE);
+    return localeFuture.thenAcceptAsync(
+        (localeMaybe) -> {
+          Locale locale = localeMaybe.orElse(LocalizedStrings.DEFAULT_LOCALE);
           boolean useStatusMessage =
               status.map(s -> s.localizedEmailBodyText().isPresent()).orElse(false);
           Messages messages =


server/app/services/applicant/ApplicantService.java
```diff
@@ -635,35 +647,51 @@ private CompletionStage<Void> maybeNotifyTiSubmitter(
...
     CompletableFuture<Optional<String>> emailFuture = getEmail(applicantId).toCompletableFuture();
     CompletableFuture<Optional<Locale>> localeFuture =
         getPreferredLocale(applicantId).toCompletableFuture();
     return CompletableFuture.allOf(emailFuture, localeFuture)
         .thenRunAsync(
             () -> {
-              // Not blocking since these are already completed
-              Optional<String> email = emailFuture.join();
               Locale locale = localeFuture.join().orElse(LocalizedStrings.DEFAULT_LOCALE);
+              Optional<String> email = emailFuture.join();
```
michaelzetune, line 672: Could we just get the email before this async block then return early before the existing return statement? Then we will have a `String` ready to go for the async block

server/app/services/applicant/ApplicantService.java
```diff
@@ -594,7 +594,15 @@ private void notifyProgramAdmins(
   }
 
   private CompletionStage<Void> maybeNotifyTiSubmitter(
-      String tiEmail, long applicantId, long applicationId, String programName) {
+      Optional<String> maybeTiEmail,
+      long applicantId,
+      long applicationId,
+      String programName,
+      Optional<StatusDefinitions.Status> status) {
+    if (!maybeTiEmail.isPresent()) {
```
michaelzetune, line 602: Personally not a big fan of having an Optional as a param just for the purpose of making the method a no-op if the param is empty. How about we keep this a String, change the method to `notifyTiSubmitter`, and don't call it if the email is empty?

server/app/services/applicant/ApplicantService.java
```diff
@@ -465,34 +471,47 @@ CompletionStage<Application> submitApplication(
...
-                        setApplicationStatus(application, defaultStatus).toCompletableFuture(),
-                        maybeNotifyApplicantWithStatus(
-                                applicantId, programDefinition, defaultStatus)
-                            .toCompletableFuture());
-              } else {
-                maybeUpdateStatusAndEmailApplicantFuture =
-                    maybeNotifyApplicant(applicantId, application.id, programName)
-                        .toCompletableFuture();
-              }
+              Optional<String> applicantEmail = applicantEmailFuture.join();
```
michaelzetune, line 476: Move this line down closer to where `applicantEmail` is actually used.

server/app/services/applicant/ApplicantService.java
```diff
@@ -593,8 +612,12 @@ private void notifyProgramAdmins(
     }
   }
 
-  private CompletionStage<Void> maybeNotifyTiSubmitter(
-      String tiEmail, long applicantId, long applicationId, String programName) {
+  private CompletionStage<Void> notifyTiSubmitter(
```
michaelzetune, line 615: Javadoc for this please

server/app/services/applicant/ApplicantService.java
```diff
@@ -465,34 +471,47 @@ CompletionStage<Application> submitApplication(
...
+              CompletableFuture<ApplicationEvent> maybeUpdateStatusFuture =
+                  maybeDefaultStatus
+                      .map(
+                          status -> setApplicationStatus(application, status).toCompletableFuture())
+                      .orElse(CompletableFuture.completedFuture(null));
+              CompletableFuture<Void> notifyProgramAdminsFuture =
+                  CompletableFuture.runAsync(
+                      () ->
+                          notifyProgramAdmins(applicantId, programId, application.id, programName));
+              CompletableFuture<Void> maybeNotifyTiSubmitterFuture =
```
michaelzetune, line 487: Remove the `maybe` here and below? Now that the method names don't have them

server/app/services/applicant/ApplicantService.java
```diff
@@ -465,34 +471,47 @@ CompletionStage<Application> submitApplication(
...
+                  tiSubmitterEmail
+                      .map(
```
michaelzetune, line 489: I think `isPresent()` may be more readable:

```java
CompletableFuture<Void> maybeNotifyTiSubmitterFuture =
    tiSubmitterEmail.isPresent() ? 
            notifyTiSubmitter(
                    tiSubmitterEmail.get(),
                    applicantId,
                    application.id,
                    programName,
                    maybeDefaultStatus)
                .toCompletableFuture()
        : CompletableFuture.completedFuture(null);
```

Happy with either though. What do you think?

server/app/services/applicant/ApplicantService.java
michaelzetune: I find it more readable, but if you disagree feel free to roll it back; I won't block on it (hehe, `CompletableFuture` pun)

server/test/services/applicant/ApplicantServiceTest.java
michaelzetune, line 1: > Additionally, this sends the status email body to the TI as well as the applicant, rather than the default 'application received' email to the TI.

Is it possible to test this behavior in this file?

michaelzetune (changes requested):

server/app/services/applicant/ApplicantService.java
michaelzetune: SG, I'm happy to keep it as-is.

server/app/services/applicant/ApplicantService.java
```diff
@@ -635,87 +662,54 @@ private CompletionStage<Void> maybeNotifyTiSubmitter(
...
-                  amazonSESClient.send(stagingApplicantNotificationMailingList, subject, message);
-                } else {
-                  amazonSESClient.send(maybeEmail.get(), subject, message);
-                }
-              }
-            },
-            httpExecutionContext.current());
+    return localeFuture.thenRunAsync(
+        () -> {
+          // Java seems not to like using thenApplyAsync here for some reason I don't understand,
```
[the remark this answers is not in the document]
michaelzetune:
```java
  private CompletionStage<Void> notifyApplicant(
      long applicantId,
      long applicationId,
      String applicantEmail,
      ProgramDefinition programDef,
      Optional<StatusDefinitions.Status> status) {
    CompletableFuture<Optional<Locale>> localeFuture =
        getPreferredLocale(applicantId).toCompletableFuture();
    return localeFuture.thenApplyAsync(
      (Optional<Locale> localeOpt) -> {
        Locale locale = localeOpt.orElse(LocalizedStrings.DEFAULT_LOCALE);
        boolean useStatusMessage =
          status.map(s -> s.localizedEmailBodyText().isPresent()).orElse(false);
          Messages messages =
              messagesApi.preferred(ImmutableSet.of(Lang.forCode(locale.toLanguageTag())));
          String programName = programDef.localizedName().getOrDefault(locale);
          String subject =
              messages.at(MessageKey.EMAIL_APPLICATION_RECEIVED_SUBJECT.getKeyName(), programName);
          String message =
              String.format(
                  "%s\n%s",
                  useStatusMessage
                      ? status.get().localizedEmailBodyText().get().getOrDefault(locale)
                      : messages.at(
                          MessageKey.EMAIL_APPLICATION_RECEIVED_BODY.getKeyName(),
                          programName,
                          applicantId,
                          applicationId),
                  messages.at(MessageKey.EMAIL_LOGIN_TO_CIVIFORM.getKeyName(), baseUrl));
          if (isStaging) {
            amazonSESClient.send(stagingApplicantNotificationMailingList, subject, message);
          } else {
            amazonSESClient.send(applicantEmail, subject, message);
          }
          return null;
        },
  }
```

I think I figured it out - you have to return `null` for it to accept that the `thenReturnAsync` is returning a `CompletionStage<Void>`. Compiles for me.

If you run into more trouble you can just go with the `thenRunAsync` version.

server/test/services/applicant/ApplicantServiceTest.java
michaelzetune: Ah, I see now. Thanks

server/app/services/applicant/ApplicantService.java
```diff
@@ -609,6 +656,8 @@ private CompletionStage<Void> maybeNotifyTiSubmitter(
             () -> {
               // Not blocking since it already completed
               Locale locale = localeFuture.join().orElse(LocalizedStrings.DEFAULT_LOCALE);
+              boolean useStatusMessage =
```
michaelzetune, line 659: nit: move this down closer to where the var is used.

michaelzetune (approved):

PR #4807, opened 2023-05-09
add label to EMAIL_SEND_COUNT inc

### LINKED ISSUE #4768: Investigate email sends from prober tests in AWS staging
In AWS staging we are limited to 200 email sends per day. We are currently hitting that limit because of prober tests that run when staging is deployed which send real emails. This means that we can't send adhoc test emails from staging and sometimes are blocked from doing things like adding a verified email address (which requires sending a verification email).

One option is to request production access for SES (Amazon Simple Email Service) which would up our quota to 50k sends per day.

To avoid extra charges, however, it would be great to leave it in Sandbox mode and figure out a way to run the tests without sending real emails. Some options for doing this might include:
- Updating tests to use localstack
- Adding a flag in tests to prevent emails from actually being sent
- Using a non-verified email address in tests (AWS will only send from a verified email address)

rockycodes:
### Description

Application submissions on staging are failing with a null pointer exception pointing to the `noLabelsChild.inc(amt);` line in
```
  /**
   * Increment the counter with no labels by the given amount.
   * @throws IllegalArgumentException If amt is negative.
   */
  public void inc(double amt) {
    noLabelsChild.inc(amt);
  }
```

 in `io.prometheus.client.Counter`.

I think this is because we're calling `inc()` [here](https://github.com/civiform/civiform/blob/aaec4b78138b0b74aaebc7238f7760d64d93e8d9/server/app/services/cloud/aws/SimpleEmail.java#L111) without adding a label to it ([docs](https://prometheus.github.io/client_java/io/prometheus/client/Counter.html)).

This pr adds a label to the counter and also calls `inc()` on errors so we will count those according to their statuses.

## Release notes

### Issue(s) this completes

Fixes #4768;

### DIFF
--- a/server/app/services/cloud/aws/SimpleEmail.java
+++ b/server/app/services/cloud/aws/SimpleEmail.java
@@ -16,6 +16,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.Environment;
 import play.inject.ApplicationLifecycle;
+import software.amazon.awssdk.http.HttpStatusCode;
 import software.amazon.awssdk.services.ses.SesClient;
 import software.amazon.awssdk.services.ses.model.Body;
 import software.amazon.awssdk.services.ses.model.Content;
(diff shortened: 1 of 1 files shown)


dkatzz (approved):
Thank you for fixing this!!

PR #4817, opened 2023-05-09
Fix guest merge logic

michaelzetune:
### Description

This PR avoids merging a guest with a user logging in if the guest has no submitted applications. This resolves a bug where if `BYPASS_LOGIN_LANGUAGE_SCREENS` is set, a TI creates an applicant, and the applicant later logs in to CiviForm with the same email, then a duplicate applicant will be created.

Before the login changes, this wasn't a possibility because users would see the login screen first, and directly log in with their email that matches the one the TI used. Now, since they start as a guest upon loading CiviForm, this issue in the merging logic is manifesting.

## Release notes

None.

#### General

   - `CiviFormProfileMergerTest.java` is updated to account for that fact that we should not merge guests with no applications. 
   - There are browser tests are in [mzetune/duplicant-applicant-test](https://github.com/civiform/civiform/commit/2cf5040636971b6185892f8304a3aae1da8e575f). I've asserted that they fail before this change and pass afterwards, Unfortunately I don't think it's feasible to include these tests in the PR because they will rely on the name of the test user the TI creates exactly matching the name of the test user that the user logs in with, which couples this test with the test environment.

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

None.

### DIFF
--- a/server/test/auth/CiviFormProfileMergerTest.java
+++ b/server/test/auth/CiviFormProfileMergerTest.java
@@ -7,10 +7,12 @@ import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import com.google.common.collect.ImmutableList;
 import java.util.Collections;
 import java.util.Optional;
 import models.Account;
 import models.Applicant;
+import models.Application;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


server/test/auth/CiviFormProfileMergerTest.java
michaelzetune, line 1: Here we make `Applicant applicant` a mock so that we can make it `thenReturn` the singleton list of `Application`. 

By default, we return this list to exercise the general case of a guest with applications being merged in with the database applicant. In the new unit test we override this to an empty list to test the new behavior.

michaelzetune: Please resolve this once you read it so auto-merge goes through :)

rockycodes (approved):

PR #4821, opened 2023-05-10
Take TIs directly to their dashboard instead of the programs index page

### LINKED ISSUE #4816: TI flow incorrect
**Describe the bug**
As a Trusted Intermediary, the current flow is:
1. Login in as a TI
2. Land on the Program Cards page
3. Click on "Add a Client" 
4. Land on Client Roster page

Landing directly on the Program Cards page is confusing and has led to TIs applying for programs for themselves (not on behalf of clients). To avoid confusion, TIs should land directly on the Client Roster page.

**Screenshots**
Current State
image

Desired Flow (TIs should land on the Client Roster directly after logging in)
image

michaelzetune:
## Description

Takes TIs directly to their dashboard instead of the programs index page. Previously, we would take TIs to the programs index page, and this causes confusion as they often apply for programs under their own account instead of their managed users. See #4816

## Release notes

Takes TIs directly to their dashboard instead of the programs index page.

## Test strategy

1. Create a new user `myuser` in the test environment.
2. Set `member_of_group_id` with `UPDATE accounts SET member_of_group_id = 1 WHERE id = 123;`
3. Log in as `myuser` and expect to be taken directly to TI dashboard.

### General

   - Not possible to unit test this behavior directly, as our test helper `loginAsTrustedIntermediary` uses the fake admin login, which is not the same as a true TI. The other conditional routes in `HomeController.index()` aren't unit tested for this same reason.
   - Not necessary

### User visible changes

   - Same issue as earlier, `loginAsTrustedIntermediary` depends on fake admin so hard to test this directly.

### New Features

No new feature flags

### Issue(s) this completes

Fixes #4816

### DIFF
--- a/server/app/controllers/HomeController.java
+++ b/server/app/controllers/HomeController.java
@@ -78,6 +78,13 @@ public class HomeController extends Controller {
     } else if (profile.isProgramAdmin()) {
       return CompletableFuture.completedFuture(
           redirect(controllers.admin.routes.ProgramAdminController.index()));
+    } else if (profile.isTrustedIntermediary()) {
+      return CompletableFuture.completedFuture(
+          redirect(
+              controllers.ti.routes.TrustedIntermediaryController.dashboard(
+                  /* nameQuery= */ Optional.empty(),
+                  /* dateQuery= */ Optional.empty(),
+                  /* page= */ Optional.empty())));
     } else {
       return profile
           .getApplicant()


dkatzz:
We have browser tests for trusted intermediaries [here](https://github.com/civiform/civiform/blob/main/browser-test/src/trusted_intermediary.test.ts). Can we add a test for this there?

dkatzz (approved):

PR #4829, opened 2023-05-10
Make bin/pull-image better handle USE_LOCAL_CIVIFORM

gwendolyngoetz:
### Description

This was skipping if you had USE_LOCAL_CIVIFORM set at all making overriding back to 0 impossible. Now it will only skip if the env variable is set to 1.

#### General

### DIFF
--- a/bin/pull-image
+++ b/bin/pull-image
@@ -8,7 +8,7 @@
 
 source bin/lib.sh
 
-if [[ "${USE_LOCAL_CIVIFORM}" == 1 ]]; then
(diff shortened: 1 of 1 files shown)


bin/pull-image
```diff
@@ -8,7 +8,7 @@
 
 source bin/lib.sh
 
-if [[ -n "${USE_LOCAL_CIVIFORM}" ]]; then
+if [[ "${USE_LOCAL_CIVIFORM}" == 1 ]]; then
```
bion, line 11: **suggestion (non-blocking):** maybe also?
```suggestion
if [[ "${USE_LOCAL_CIVIFORM}" == 1 || "${USE_LOCAL_CIVIFORM}" == "true" ]]; then
```

avaleske (approved):

PR #4836, opened 2023-05-11
Add validation for predicate selections

### LINKED ISSUE #4383: Saving eligibility condition without operator throws an error
**Describe the bug**
When adding an eligibility condition from the admin UI, if none of the [contains any of, is none of, is a subset of] options are selected, an error is thrown.

**Expected behavior**
Form should not be allowed to be submitted and should instead show an error message to the admin that they need to set the value.

dkatzz:
### Description

Add validation to check whether the predicate form is filled out before triggering the submit. Previously, if there was an issue with the inputs, the user would still be navigated to the next page, but that page would just show an error. By adding error messaging, the admin can fix the errors and not lose their work if they forget a field. We can make improvements to this in the future to make the error messages more specific (checking with Sijia on the behavior), but for now this is catches all cases where a scalar, operator, or value is missing for any fields. 

## Release notes

Add error validation to predicate selection page

#### General

### Issue(s) this completes

Fixes #4383


server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -22,6 +23,10 @@ class AdminPredicateConfiguration {
       return
     }
 
+    addEventListenerToElements('button[type=submit]', 'click', (event: Event) =>
```
reemaka, line 26: **optional:** might be better to use an ID to identify the button in case the UI changes in the future

reemaka (approved):

server/app/assets/javascripts/admin_predicate_configuration.ts
dkatzz: Good call - updated!

server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -52,6 +59,69 @@ class AdminPredicateConfiguration {
...
+      ?.querySelector('#predicate-config-value-row-container')
+      ?.querySelectorAll('input')
+      .forEach((input) => {
+        if (input.value == '') {
+          hasValueMissing = true
+        }
+      })
+
+    // If there are issues with any of the fields, we show a toast and prevent submit.
+    if (hasValueMissing || hasSelectionMissing) {
```
reemaka, line 97: Is there a way to return early instead of wrapping all the toast logic in this if?

server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -52,6 +59,69 @@ class AdminPredicateConfiguration {
...
+      event.preventDefault()
+      // Scroll to the top of the page to ensure the user sees the error message.
+      window.scrollTo(0, 0)
+      let errorMessage
+      if (hasSelectionMissing && hasValueMissing) {
+        errorMessage =
+          'One or more form fields or drop-down is missing an entry. Please fill out all form fields and dropdowns before saving.'
```
reemaka, line 104: nit: here we use "drop-down" and later "dropdowns". I think this one should be plural and we should be consistent between the two with dash or no dash.

reemaka (approved):

avaleske (approved):

PR #4878, opened 2023-05-17
Create restrictions on number of questions in predicate configuration

### LINKED ISSUE #4703: Misaligned input fields and overflow of background box
**Describe the bug**
When creating an eligibility condition that contains 5+ conditions, the conditions overflow the right side of the background grey box.  Additionally, the input fields don't line up with the selectors.

image

**To Reproduce**

1. Start a dev instance
2. Seed the database with the default questions and program
3. Create an eligibility condition on block 2 that uses all questions in that block.

**Expected behavior**
Input fields line up with the selection fields, and the 5+ conditions either wrap or the other elements expand to fit.

**Screenshots**

dkatzz:
### Description

There is strange formatting when selecting more than 4 questions to configure a predicate. We're restricting the number of questions allowed to 4 in order to avoid these errors, but can revisit this if CIs need to add more questions to a predicate block. We also throw an error when no questions are selected.

#### General

### Issue(s) this completes

Fixes #4703


server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -145,6 +142,38 @@ class AdminPredicateConfiguration {
...
+        '.cf-predicate-question-options',
+      ),
+    ).filter((el) => el.checked).length
+    if (numChecked != 0 && numChecked <= 4) {
+      return
+    }
+    event.preventDefault()
+    const errorMessage =
+      numChecked > 4
+        ? 'Please select less than 5 questions.'
```
reemaka, line 157: nit: Should say "fewer" not "less"

dkatzz: Ah thanks updated!

server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -145,6 +142,38 @@ class AdminPredicateConfiguration {
...
+      document.querySelectorAll<HTMLInputElement>(
+        '.cf-predicate-question-options',
+      ),
+    ).filter((el) => el.checked).length
+    if (numChecked != 0 && numChecked <= 4) {
+      return
+    }
+    event.preventDefault()
+    const errorMessage =
+      numChecked > 4
```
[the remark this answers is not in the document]
dkatzz: Done!

Issue #4421, opened 2023-03-16
Check if email address can ever be empty in an Account


nb1701:
In https://github.com/civiform/civiform/blob/075768e1a845f65e296082ca31ef966146e6d925/server/app/repository/UserRepository.java#L59-L64 and also the related async version of this function, we return `Optional.empty()` if the email address string passed in is null or empty.  The email address field is nullable in the database, but we may not ever actually allow for saving a null or empty string.  We should investigate in the app if we prevent this field from ever being empty, and if so, make throw an exception instead of returning `Optional.empty()`.  Additionally, if this field is indeed never empty, we should probably require it not to be null in the database.

nb1701:
*Probably* only doing this to look up admins or TIs, and you're doing this to look up an account, so almost definitely needs to throw an exception instead of swallow an empty input.

PR #4880, opened 2023-05-17
Throw exception on null or empty argument in UserRepository account lookup functions

nb1701:
These three functions used for looking up an account should never be passed a null or empty string. Rather than passing back an Optional.empty(), this throws the appropriate exception.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4421


bion:
@nb1701 please use a more descriptive PR title or add the `ignore-for-release` label

bion (approved):

Issue #1589, opened 2021-08-18
Allow only one program's draft to be published


lindsayzad:
**Is your feature request related to a problem? Please describe.**
An admin may be making edits to multiple programs at the same time.  One program may be ready to publish so that the edits can be tested and reviewed by others, while the other program's edits are still in draft and it isn't ready to be published yet.  

**Describe the solution you'd like**
In the program admin home page, add a publish draft option button next to each individual program, rather than just one common 'Publish all drafts' button.  Ideally, get this UX reviewed/designed, and build the ability on the BE

**Additional context**
This is a real pain point for users, as well as a point of frustration.  As cities scale civiform and roll it out to more and more programs, the need for this feature will increase.

**Done when**
1. BE ability build
2. UX design 
3. FE button build out and tested with full browser tests.  (bug bash?)

reemaka:
PRD started here: https://docs.google.com/document/d/1kDWF94YMDE_SOCYrhayLs4qFtT0_hoLsUZ8XtSXo0d8/edit
TDD started here: https://docs.google.com/document/d/1c9hbptNFeD_107E_YQaflKVDTNXyszwWWWLXFPHaW60/edit

PR #4882, opened 2023-05-17
Add a confirmation dialog before publishing a program.

reemaka:
### Description

Add a confirmation dialog that is displayed when a single program is being published.

Using the Modal class would be a bit complicated in this case because the target of the confirm button is dependent on which program's publish button was clicked. For simplicity, I'm using the browser's built-in confirmation instead.

#### General

### Issue(s) this completes

Part of #1589


swatkat1 (approved):
It would be nice to have a testcase for it.

swatkat1 (approved):
Thank you!

Issue #4891, opened 2023-05-18
Migrate translation process to Transifex


michaelzetune:
This issue tracks migrating our translation process from manual changes to `messages.xx` files to changes made in the Transifex dashboard.

This [design document](https://docs.google.com/document/d/1yxIifVl0B91KVTGMwFk55J0OAqWlkBI2rJpgLHrMCtE/edit?resourcekey=0--cedXnckaVGf0B-0He3oVA#heading=h.70uyn9ev3cfk) describes the entire migration.

Remaining TODOs as of 05/19/2023:

- [x] Prepare Google.org and Exygy to handle Transifex translations (medium-term and long-term, respectively)
   - For google.org:
      - [x] Figure out how to download all untranslated strings to upload to Google localization tool. Discuss with @dkatzz.
      - [x] Update google.org [g3doc](https://g3doc.corp.google.com/company/teams/dotorg-works/civic/civiform-eng.md?cl=head#translations) (Google internal) with instructions.
         - Done in [cl/534186819](https://critique.corp.google.com/cl/534186819) (Google internal)
    - For Exygy using a Transifex vendor:
       - Tracked in #5063
- Coordinate with Arkansas on a Marshallese translator. Give them access to Transifex as Translator (lowest access level possible)
   - In contact with Jake Walker from AR to coordinate this. They need to set up a contract with someone. All the infrastructure is set up; I (Michael) will train the translator on using Transifex when they are ready.
   - Tracked in #5064
- [x] [Not feasible] Enable the Slack integration (having trouble with it as of 5/19/23)
   - We might be restricted based on using the open-source plan, look into this. Contacted Transifex and waiting for a response.
   - Transifex responded that Slack is not available for open source teams. Marking this as finished.
- [x] Remove the comments from `messages` and `messages.xx` files and put them into the Transifex portal as a [context string](https://help.transifex.com/en/articles/6244125-the-complete-guide-to-adding-context-to-strings#h_2808bf112c). Create a process or GitHub check to ensure that comments are always added to Transifex as soon as source strings are created
   - #4924
   - GitHub check doesn't seem to be possible
- [x] (Optional) If possible, write a unit test that checks to make sure KVs are removed from `messages` when no longer used in CiviForm. Possibly using Java reflection.
   - #4919

michaelzetune:
Some remaining todos are tracked in new issues, linked in the description of this one.

PR #4900, opened 2023-05-18
Allow keys in `messages` to not be present in foreign language `messages.xx` files

michaelzetune:
### Description

The current `MessagesTest` checks to make sure that all keys in `messages` are present in all foreign language files.

We can't keep this check in as-is, because in the new [internationalization process](https://docs.civiform.us/contributor-guide/developer-guide/internationalization-i18n#internationalization-for-application-strings), we allow developers to insert strings into `messages` and wait for Transifex to update `messages.xx` files itself. That means `MessagesTest` would break until Transifex adds the translations.

Instead, we check that every key in the *foreign language file* is in the primary language file. While a subtle change, it still helps us check that Transifex is not making any errors in its pull requests.

## Release notes

None.

#### General

-  Extended the README / documentation, if necessary
   - https://docs.civiform.us/contributor-guide/developer-guide/internationalization-i18n#internationalization-for-application-strings

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

Relates to #4891

### DIFF
--- a/server/test/conf/MessagesTest.java
+++ b/server/test/conf/MessagesTest.java
@@ -142,8 +142,8 @@ public class MessagesTest {
     if (!keysInForeignLangFileCopy.isEmpty()) {
       return String.format(
           "%s found in %s file but not in primary language file. Add these keys to primary"
-              + " language file or to the ignore list in %s to resolve this issue.",
-          keysInForeignLangFileCopy, foreignLanguageFile, getClass().getName());
+              + " language file to resolve this issue.",
+          keysInForeignLangFileCopy, foreignLanguageFile);
     }
 
     return "No fail message available";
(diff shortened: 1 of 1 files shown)


dkatzz (approved):

server/test/conf/MessagesTest.java
```diff
@@ -132,31 +133,19 @@ private static String[] otherLanguageFiles() throws Exception {
...
+      TreeSet<String> foreignLangKeys,
+      String foreignLanguageFile) {
 
     TreeSet<String> keysInForeignLangFileCopy = new TreeSet<>(foreignLangKeys);
     keysInForeignLangFileCopy.removeAll(primaryLangKeys);
     if (!keysInForeignLangFileCopy.isEmpty()) {
-      stringBuilder.append(
-          String.format(
-              "%s found in %s file but not in primary language file. Add these keys to primary"
-                  + " language file or to the ignore list in %s to resolve this issue.",
```
dkatzz, line 156: **question:** There isn't an ignore list anymore, right?

dkatzz (approved):

server/test/conf/MessagesTest.java
michaelzetune: Good catch thanks!

PR #4907, opened 2023-05-19
Show archive status for questions.

### LINKED ISSUE #1319: On Civiform admin's Questions page, separate archived and unarchived questions
In the Civiform admin's Questions page, they see the list of all questions in CiviForm. Archived and regular (non-archived) questions are intermingled here. To make the page easier to skim, we should either:

- Style archived questions differently so the admin can tell what's active or not; or, 
- Separate archived questions, perhaps by putting them at the very bottom of the screen in a separate section.

Here's how the page looks right now, btw:

image

### LINKED ISSUE #4344: Indicate question archival status on admin question index page
## Background

The admin can archive questions (i.e. remove them from the admin index page) if they are not used in any current draft or active programs. To archive a question, they mark it for archival and then publish the new version.

## Problem

When a question is marked for archival, there is no visual indication on the index page. The only way to know that it is marked is by clicking the triple dot button to see the archival button.

## Done when

Questions listed in the admin index give some visual indication that they are marked for archival.

reemaka:
### Description

Improves the experience of archiving questions. Before this change, there was no visual indication that a question was marked for archival besides the button in the dropdown changing to "Restore archived".

1. When archiving a question, creates a draft of the question first if it does not exist. Previously, an active question could be tombstoned without needing a draft. This was confusing as you expect a question to have a draft row if it's been modified in any way. It also meant that the modified timestamp would not indicate the time at which it was marked for archival.
2. Moves questions that are marked for archival to a new section at the bottom.
3. Sets the badge to a new "ARCHIVED" badge for questions being archived. This involves changing the size of all badges to fit the longer text while keeping them consistent.

## Release notes

#### General

### Issue(s) this completes

Fixes #1319, fixes #4344


server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -158,9 +160,21 @@ private static QuestionDefinition getDisplayQuestion(QuestionCardData cardData)
     return cardData.draftQuestion().orElseGet(cardData.activeQuestion()::get);
   }
 
+  private static boolean questionIsPendingDeletion(
+      QuestionCardData card, ActiveAndDraftQuestions activeAndDraftQuestions) {
+    return questionIsPendingDeletion(getDisplayQuestion(card), activeAndDraftQuestions);
+  }
+
+  private static boolean questionIsPendingDeletion(
```
swatkat1, line 168: Nit- May be rename this to isQuestionPendingDeletion() as we are returning a boolean?

reemaka: Done

reemaka:
> So question in active state cannot be archived, just "marked for archiving" , Am I correct in my understanding?
Yeah, it was already the case that if you marked an active question as archived, it wouldn't actually be archived until the next time you publish. IMO, making it into a draft first makes it clearer that the changes will happen when you publish.

> Also, What happens if we bring back an archived question to active state?
Before you publish, you can click the "Restore archived" button to unarchive it. It'll still be a draft, but it won't be in the "Marked for archival section" and it'll have the purple "DRAFT" badge. When you publish, it'll be a regular active question. We don't currently have a way to recover archived questions after publishing.

swatkat1 (approved):
Thanks!

PR #4909, opened 2023-05-19
Allow publishing questions without programs.

reemaka:
### Description

Follow up to #4898. I missed this edge case that happens when no programs have ever been created and you try to publish questions.

#### General

### DIFF
--- a/server/test/repository/VersionRepositoryTest.java
+++ b/server/test/repository/VersionRepositoryTest.java
@@ -252,14 +252,12 @@ public class VersionRepositoryTest extends ResetPostgres {
     assertThat(versionRepository.getDraftVersion().getPrograms()).isEmpty();
     assertThat(versionRepository.getDraftVersion().getQuestions()).isEmpty();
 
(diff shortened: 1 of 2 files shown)


server/test/repository/VersionRepositoryTest.java
```diff
@@ -219,6 +219,46 @@ public void testPublishWithOnlyQuestionsModified() {
     assertThat(oldActive.getLifecycleStage()).isEqualTo(LifecycleStage.OBSOLETE);
   }
 
+  @Test
+  public void testPublishWithOnlyQuestions() {
```
dkatzz, line 223: **suggestion:** Could you name this a bit more descriptively? The one before is called `testPublishWithOnlyQuestionsModified` and it's hard to tell what's different here without looking more closely.

reemaka: Done, hope that's clearer

dkatzz (approved):

PR #4928, opened 2023-05-22
Clean up Login/Languages Code

michaelzetune:
### Description

This cleans up all code related to the Login, Languages, Landing project.

Specifically:
- Removes all instances of `TODO(#4705)`
- Removes `loginAsGuest`, which is now redundant since we login as a guest by default
- Removes the usages of `selectApplicantLanguage` when it selects English, since this is now redundant.
- Removes the `LoginForm` view and the `ApplicantInformationView`, which was just the language selection screen.

Also:
- Changes a button in `EnumeratorQuestionRenderer` to use `Icons.PLUS` instead of a literal "+" sign. This is better practice
   - This is a strange PR to include this change in, but something in the way the tests are structured and changed in this PR has set off some kind of non-determinism in the button's spacing. This non-determinism is solved by using the standard icons for this type of button, so we do so here.
enumerator-errors-diff

## Release notes

None.

#### General

   - Not necessary; code cleanup
   - Not necessary; code cleanup

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

Relates to #4705.


browser-test/image_snapshots/mobile_views_test/mobile-initial-load.png
michaelzetune, line 1: The only thing that changed here is the font changed from gray to black. I don't know why, but it likely has to do with #3225

rockycodes (approved):
looks good! exciting to see all this going away 😄

michaelzetune:
TODO: remove instances of the string `loginForm`

PR #4936, opened 2023-05-23
Fixing code scanning alert on permissions

gwendolyngoetz:
### Description

This fixes a 'high' code scanning alert. Setting default permissions to read. Also adding a comment on the token being used.

#### General

### DIFF
--- a/.github/workflows/auto_request_review.yml
+++ b/.github/workflows/auto_request_review.yml
@@ -1,5 +1,7 @@
 name: Auto Request Review
 
+permissions: read-all
+
 on:
   pull_request:
     types: [opened, ready_for_review, reopened, synchronize]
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_request_review.yml
```diff
@@ -1,5 +1,7 @@
 name: Auto Request Review
 
+permissions: read-all
+
```
michaelzetune, line 4: It looks like the `read-all` permission is specific to `GITHUB_TOKEN`:
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

Better to remove it here so there's no confusion?

michaelzetune (approved):

.github/workflows/auto_request_review.yml
gwendolyngoetz: We've typically set it to read-all as the default at the top level of the action. Without the system generates high token-permission warnings in the security tab. If we add anything else to this that doesn't need the custom token we'll already be set and can override at the job level.

michaelzetune: SGTM

PR #4937, opened 2023-05-23
Increase contrast for Archived badge.

reemaka:
### Description

I ran the colors through an a11y checker and they weren't contrasting enough.

#### General


avaleske (approved):
Good call!

Issue #4529, opened 2023-03-29
Clean up fix for program index page


nb1701:
https://github.com/civiform/civiform/pull/4528 fixes the issue found in https://github.com/civiform/civiform/issues/4507, but in a synchronous way that incurs a performance hit (though somewhat minimal since we are only acting on visible programs).

Done when:

- Browser test coverage for the original bug
- Code is asynchronous for the applicant
- Loading the applicant index view has fewer database queries (may be addressed by fixing https://github.com/civiform/civiform/issues/3609)

PR #4940, opened 2023-05-23
syncProgramDefinitionQuestions permanent fix

nb1701:
### Description

In https://github.com/civiform/civiform/pull/4528, we added a bug fix for an issue that had escaped and was showing up in production. This fix was synchronous, and since it's in the code path for user-facing functionality we need to fix it to be asynchronous.

First, this removes the "asynchronous" version of `getReadOnlyQuestionService`.  What this function was doing wasn't actually asynchronous, but simply wrapping a `CompletableFuture.completedStage` around a synchronous function.  It was being used to chain other async blocks to it, but this can be done with `CompletableFuture.supplyAsync`.  

Next, this puts the bug fix inside a `CompletableFuture.supplyAsync` in order to make it asynchronous.

A browser test is added to ensure we don't hit this bug again.

Finally, a small change to `bin/run-browser-tests` is made to remove the `diff_output` directory before each test run, so as not to collect diffs from multiple test run invocations.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4529

### DIFF
--- a/bin/run-browser-tests
+++ b/bin/run-browser-tests
@@ -6,9 +6,10 @@ source bin/lib.sh
 docker::set_project_name_browser_tests
 bin/pull-image
 
-# Remove the diff_output directory, since it may contain diffs from
+# Remove these directories since they may contain output from
 # previous test runs.
 rm -rf browser-test/diff_output
+rm -rf browser-test/updated_snapshots
 
 # Run browser tests from within the civiform-browser-test container. We map
 # full browser-test local directory to the container so that it uses local changes.


server/app/services/program/ProgramServiceImpl.java
```diff
@@ -131,31 +131,32 @@ public CompletionStage<ProgramDefinition> getActiveProgramDefinitionAsync(String
...
+                      return syncProgramDefinitionQuestions(
+                          programDef, questionService.getReadOnlyVersionedQuestionService(v));
+                    } catch (QuestionNotFoundException e) {
+                      throw new RuntimeException(
+                          String.format("Question not found for Program %s", programDef.id()), e);
+                    }
+                  })
+              .collect(ImmutableList.toImmutableList());
+        },
+        httpExecutionContext.current());
```
bion, line 159: **issue (blocking):** wrapping the contents of this method in `CompletableFuture.supplyAsync` and passing `httpExecutionContext.current` does not change its threading behavior from the previous implementation: an HTTP thread will still be blocked waiting for database IO.

bion (changes requested):

Issue #4948, opened 2023-05-24
Show read-only settings panel in CiviForm Admin UI


bion:
Phase 1 of [admin settings panel](https://docs.google.com/document/d/1FK5YLJd6jVu1O1uuu2OOR2mXTs3yVEXkzyfjVwnkz8s/edit)

- [x] update env-var-docs.json with `mode` attr for each var, update python code to check for `mode` on PRs that update the file https://github.com/civiform/civiform/pull/4958
- [x] fix inconsistencies between server-internal HOCON names and env var names https://github.com/civiform/civiform/pull/4982
- [x]  write script for generating `SettingsManifest.java` https://github.com/civiform/civiform/pull/4987
- [x]  add `Settings` page for CF admin, release behind feature flag https://github.com/civiform/civiform/pull/5013
- [ ]  create mocks to display read only settings page @sijiayam 
- [ ] replace uses of `FeatureFlags` with `SettingsManifest` https://github.com/civiform/civiform/pull/5070

Feature flag: https://github.com/civiform/civiform/issues/5014 `ADMIN_SETTINGS_PANEL_ENABLED`

PR #4958, opened 2023-05-24
Add support for mode prop to env variables

bion:
### Description

Per [Admin Settings Panel TDD](https://docs.google.com/document/d/1FK5YLJd6jVu1O1uuu2OOR2mXTs3yVEXkzyfjVwnkz8s/edit), this PR adds a `mode` attribute to the `Variable` type in the env var docs system.

For now it only supports `HIDDEN` and `ADMIN_READABLE` modes, since the first phase of this feature is read-only.

#### General

### Issue(s) this completes

Contributes to https://github.com/civiform/civiform/issues/4948

### DIFF
--- a/env-var-docs/parser-package/README.md
+++ b/env-var-docs/parser-package/README.md
@@ -93,7 +93,7 @@ Here is an example environment variable documentation file (validated in
             "TITLE": {
                 "description": "Sets the CiviForm title.",
                 "type": "string",
-                "mode": "ADMIN_WRITEABLE"
+                "mode": "ADMIN_READABLE"
             },
             "LOGO_URL": {
                 "description": "URL of the logo.",
--- a/env-var-docs/parser-package/README.md
+++ b/env-var-docs/parser-package/README.md
@@ -92,7 +92,8 @@ Here is an example environment variable documentation file (validated in
         "members": {
             "TITLE": {
                 "description": "Sets the CiviForm title.",
-                "type": "string"
+                "type": "string",
+                "mode": "ADMIN_WRITEABLE"
             },
             "LOGO_URL": {
                 "description": "URL of the logo.",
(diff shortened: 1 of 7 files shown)


env-var-docs/parser-package/README.md
```diff
@@ -92,7 +92,8 @@ Here is an example environment variable documentation file (validated in
         "members": {
             "TITLE": {
                 "description": "Sets the CiviForm title.",
-                "type": "string"
+                "type": "string",
```
rockycodes, line 95: suggestion (blocking): please add `mode` + an explanation of the field to the "Variables have the following fields:" section above. Since mode is a required field, please also add it to this sentence: "The `description` and `type` fields must be defined for each variable." 

I think you should also add an "is not None" check for Mode in `parser.py` (beneath the "# appease mypy" comment), although I'm not positive because I'm not totally clear on why we need that.

rockycodes (changes requested):
a couple of documentation requests, but overall looks good!

server/conf/env-var-docs.json
```diff
@@ -453,43 +539,53 @@
...
         "type": "bool"
       },
       "NONGATED_ELIGIBILITY_ENABLED": {
+        "mode": "ADMIN_READABLE",
         "description": "Enables the feature that allows setting eligibility criteria to non-gating.",
         "type": "bool"
       },
 
       "STAGING_ADD_NOINDEX_META_TAG": {
+        "mode": "ADMIN_READABLE",
```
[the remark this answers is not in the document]
bion: Good catch! I just mindlessly set everything under "Feature Flags" to be admin readable but this probably shouldn't be. Changed this and `STAGING_ADD_NOINDEX_META_TAG` to `HIDDEN`.

env-var-docs/parser-package/README.md
bion: Done

rockycodes (approved):

PR #4967, opened 2023-05-25
Parameterize `createXQuestion()`

michaelzetune:
### Description

Quick refactor of `createXQuestion()` with parameterization

Instead of:

https://github.com/civiform/civiform/blob/52b543260f9e0557039e93e4de870a3065cc70f7/server/app/services/applicant/question/ApplicantQuestion.java#L214-L276

We just use:

```java
  public <T extends Question> T createQuestion(Class<T> questionClass) {
    try {
      // Get the constructor for T that has a single parameter of type ApplicantQuestion.
      Constructor<T> constructor = questionClass.getDeclaredConstructor(this.getClass());
      return constructor.newInstance(this);
    } catch (InstantiationException
        | IllegalAccessException
        | NoSuchMethodException
        | InvocationTargetException e) {
      throw new RuntimeException(
          "Could not find appropriate constructor for " + questionClass.getName(), e);
    }
  }
```

## Release notes

None.

#### General

   - Not necessary
   - Not necessary

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

None.

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramBlocksController.java
+++ b/server/app/controllers/applicant/ApplicantProgramBlocksController.java
@@ -43,6 +43,7 @@ import services.applicant.ReadOnlyApplicantProgramService;
 import services.applicant.exception.ApplicantNotFoundException;
 import services.applicant.exception.ProgramBlockNotFoundException;
 import services.applicant.question.AddressQuestion;
+import services.applicant.question.ApplicantQuestion;
 import services.applicant.question.FileUploadQuestion;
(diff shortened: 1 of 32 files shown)


avaleske:
**Question:** This opinion isn't strongly held but I'd like to here what @bion or someone else thinks.
I think reducing code duplication and boilerplate in `ApplicantQuestion` is a good goal, but here it makes the API less convenient, since it's slightly less readable and breaks the IDE's tab completion.

Issue #4178, opened 2023-02-07
Creating questions with distinct names but only varying by non-alphabetic characters results in a conflict


nb1701:
**Describe the bug**
When creating two questions whose name varies only by non-alphabetic characters, such as `screen1address` and `screen1address2`, the second question will fail to be created with an error banner such as `Question 'screenaddress' conflicts with question id: ##`.

This is likely due to https://github.com/civiform/civiform/blob/1fc5c822e4d4a98bcc4bef0715bfe4b80aeb5b97/server/app/services/question/types/QuestionDefinition.java#L119 where we are clearing out all non-alphabetic and non-whitespace characters from the name.  See https://github.com/civiform/civiform/issues/783 for more details on why we're doing this.

**To Reproduce**

1. Go to the admin interface and edit a screen.
2. Click on 'Add a question'
3. Click on 'Create new question' (type shouldn't matter)
4. Create a question with name like "question1"
5. Repeat step 4 with a name like "question2"
6. See error

**Expected behavior**
User is able to create questions with distinct names that differ only by non-alphabetic and non-whitespace characters.  At least numbers should be allowed, if not the full charset.

**Screenshots**
![image](https://user-images.githubusercontent.com/122842472/217388739-6d652f52-c3e7-44e2-a277-f9c0ec035e31.png)

**Additional context**
https://github.com/civiform/civiform/issues/783

reemaka:
I looked into this a bit and it seems like changing the path generation logic might have consequences like any existing ApplicantData not having the correct question paths. That might be fine for numbers if no CEs are currently using numbers in the admin IDs, but might be bad for characters like "-" which are likely being used in admin IDs more often. We should check with CEs about what kinds of characters they've been using in question admin IDs before making this change.

PR #4991, opened 2023-05-31
Display clearer error message when question name would create path conflicts.

reemaka:
### Description

Temporary mitigation for #4178. Characters that are not alphabetic or spaces get stripped out of the name for the question path, creating confusing conflicts with existing questions. For example, "question1" and "question2" would have the same path, so the admin would see an error message like "Question 'question' conflicts with question id: 123" when attempting to create "question2". This improves the experience in the meantime by rewording the error.

#### General

### Issue(s) this completes

Part of #4178

### DIFF
--- a/server/app/services/question/types/QuestionDefinition.java
+++ b/server/app/services/question/types/QuestionDefinition.java
@@ -237,10 +237,7 @@ public abstract class QuestionDefinition {
     if (name.isBlank()) {
       errors.add(CiviFormError.of("Administrative identifier cannot be blank"));
     }
-    // Check if any characters were stripped out, which only happens if they are not alphabetic or
-    // spaces.
-    System.out.println(name);
-    System.out.println(getQuestionPathSegment());
+    // Question names should not contain characters we can't handle in question paths.
     if (!name.matches("^[a-zA-Z ]*$")) {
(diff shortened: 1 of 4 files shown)


server/app/services/question/types/QuestionDefinition.java
```diff
@@ -237,6 +237,10 @@ public final ImmutableSet<CiviFormError> validate() {
     if (name.isBlank()) {
       errors.add(CiviFormError.of("Administrative identifier cannot be blank"));
     }
+    // Question names should not contain characters we can't handle in question paths.
+    if (!name.matches("^[a-zA-Z ]*$")) {
```
gwendolyngoetz, line 241: Can you put the regex string in a constant called VALID_QUESTION_NAME_CHARACTERS or something of that ilk?

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

gwendolyngoetz (approved):

gwendolyngoetz (approved):

reemaka:
Scaled this back to just making the error message clearer. Validating the question name was causing all of the browser tests that set question names to break. I thought about allowing characters like hyphens and underscores instead, but if we update the path creation logic, it will affect how things are stored in the database so I didn't want to do that.

PR #4994, opened 2023-05-31
Fix login in tests on AWS staging

nb1701:
Auth0 added a hidden Continue button in addition to the actual, visible Continue button for some reason. This changes the selector to only click the visible button.


bion:
image
image

PR #5032, opened 2023-06-08
Allow TIs to authenticate with ADFS, in addition to the applicant identity provider

### LINKED ISSUE #5017: Allow Trusted Intermediaries to authenticate with either applicant or staff identity backend
# Overview

Trusted Intermediaries currently can only authenticate with the [applicant identity provider](https://github.com/civiform/civiform/blob/bionj-admins-can-see-settings/server/conf/helper/auth.conf#L3). This prevents civic entity staff from assuming the Trusted Intermediary role using their staff accounts.

A workaround is for staff to create TI accounts using the applicant identity provider, but this is not acceptable for civic entities that want the security guarantees of their staff authentication backend.

# UX considerations

Currently, when any staff-authenticated account logs in they are presented with the program index view in the admin layout. If they have the OIDC claim for CiviForm Admin, they are presented with the CiviForm Admin layout. Otherwise they are assumed to be a Program Admin and are shown a list of whatever programs they have been assigned as a Program Admin.

When a staff account logs that is designated as a TI, what should they see? What if they are designated as both a TI and a Program Admin? Also a CiviForm Admin?

# Relevant code paths

- Determining landing page based on user role: https://github.com/civiform/civiform/blob/main/server/app/controllers/HomeController.java#L59-L72
- Adding a new TI (creates a user account with the provided email if one doesn't exist): https://github.com/civiform/civiform/blob/main/server/app/repository/UserRepository.java#L172-L194
- Role creation for ADFS (staff/admin) profiles: https://github.com/civiform/civiform/blob/main/server/app/auth/oidc/admin/AdfsProfileCreator.java#L40-L47
- Role creation for non-staff profiles:
https://github.com/civiform/civiform/blob/main/server/app/auth/oidc/applicant/ApplicantProfileCreator.java#L85-L91

# Updates

* Based on conversation with Arkansas ([notes](https://docs.google.com/document/d/1gBsLGLlskSoAuMaFWLnDL0SyxPX_JLNx-Fw6unZY6ys/edit?resourcekey=0-v__fP947SnRcAsatHfo3eQ#bookmark=id.m7ieta7hoi3h)), we can assume that users will continue to be in exactly one role. This significantly reduces the implementation complexity since we don't need to address the case where a user logs in and needs to make a decision between going to the TI portal or [program|CiviForm] admin portal.

michaelzetune:
### Description

This PR opens the door to trusted intermediaries to log in with ADFS in addition to the applicant backend. Based on the requirements discussed, we do not need to account for users that are both TIs and CiviForm/Program admins. Therefore, we implement a rejection of adding TIs if they are already in one of these roles.

## Release notes

Trusted Intermediaries can now use ADFS to log in to CiviForm.

#### General

   - Since we don't have a fake ADFS the same way we have fake OIDC, I don't know how we could test that logging in as a TI with ADFS works correctly.
   - Not necessary

#### User visible changes

   - Not possible to authenticate with ADFS in a browser test so we cannot directly test logging in
   - Not possible to test the rejection of new TIs if they are already a program/civiform admin, because we need their email address (like `fake-local-admin-123@example.com`) and this is generated on-the-fly with `loginAsAdmin()` and we cannot retrieve this email in the test.
   - Not possible to authenticate with ADFS in dev mode

#### New Features

   - Not necessary
   - Not necessary
   - No flag

### Issue(s) this completes

Fixes #5017

### DIFF
--- a/server/app/auth/oidc/CiviformOidcProfileCreator.java
+++ b/server/app/auth/oidc/CiviformOidcProfileCreator.java
@@ -191,4 +191,8 @@ public abstract class CiviformOidcProfileCreator extends OidcProfileCreator {
         .toCompletableFuture()
         .join();
   }
+
+  protected final boolean isTrustedIntermediary(CiviFormProfile profile) {
+    return profile.getAccount().join().getMemberOfGroup().isPresent();
+  }
 }
--- a/server/app/auth/oidc/CiviformOidcProfileCreator.java
+++ b/server/app/auth/oidc/CiviformOidcProfileCreator.java
@@ -109,6 +109,13 @@ public abstract class CiviformOidcProfileCreator extends OidcProfileCreator {
   /** Merge the two provided profiles into a new CiviFormProfileData. */
   protected CiviFormProfileData mergeCiviFormProfile(
       CiviFormProfile civiformProfile, OidcProfile oidcProfile) {
+
+    // If the civiformProfile is a trusted intermediary, bypass all merging because
+    // we don't want to actually merge the guest profile into theirs.
+    if (isTrustedIntermediary(civiformProfile)) {
+      return civiformProfile.getProfileData();
+    }
+
     String emailAddress =
         getEmail(oidcProfile)
             .orElseThrow(
--- a/server/app/auth/oidc/CiviformOidcProfileCreator.java
+++ b/server/app/auth/oidc/CiviformOidcProfileCreator.java
@@ -110,7 +110,14 @@ public abstract class CiviformOidcProfileCreator extends OidcProfileCreator {
   protected CiviFormProfileData mergeCiviFormProfile(
       CiviFormProfile civiformProfile, OidcProfile oidcProfile) {
 
-    // If the civiformProfile is a trusted intermediary, bypass all merging because
+    // Meaning: whatever you signed in with most recently is the role you have.
+    ImmutableSet<Role> roles = roles(civiformProfile, oidcProfile);
+    roles.stream()
+        .map(Role::toString)
+        .forEach(role -> civiformProfile.getProfileData().addRole(role));
+    adaptForRole(civiformProfile, roles);
+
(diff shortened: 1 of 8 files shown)


server/app/auth/oidc/CiviformOidcProfileCreator.java
```diff
@@ -191,4 +198,8 @@ public final Optional<Applicant> getExistingApplicant(OidcProfile profile) {
         .toCompletableFuture()
         .join();
   }
+
+  protected final boolean isTrustedIntermediary(CiviFormProfile profile) {
```
michaelzetune, line 202: This was moved into this base class, from `ApplicantProfileCreator.java`.

michaelzetune, line 202: This was moved into this base class, from `ApplicantProfileCreator.java`.

michaelzetune, line 202: This was moved into this base class, from `ApplicantProfileCreator.java`.

server/app/repository/UserRepository.java
```diff
@@ -189,6 +190,11 @@ public void addTrustedIntermediaryToGroup(long id, String emailAddress)
               a.save();
               return a;
             });
+
+    if (account.getGlobalAdmin() || !account.getAdministeredProgramNames().isEmpty()) {
```
bion, line 194: **question (non-blocking):** Is there a technical reason for preventing this?

bion (approved):

server/app/repository/UserRepository.java
bion: That's helpful context, thanks!

michaelzetune: Actually there is the UX reason of "which portal do we redirect an admin+TI user upon login" that we would have to resolve here if we allowed it - forgot about that.

bion: Aha, thanks for mentioning

PR #5051, opened 2023-06-12
Fix for SELECT_TI Visibility Browser tests failures

### LINKED ISSUE #5050: ProgramVisibility Browser tests fails on AWS staging and Seattle staging.
**Describe the bug**
ProgramVisibility Browser tests fails on AWS staging and Seattle staging. 

**Deatils**
The testuser is hard-coded in the browser test which causes this failure. This should be fixed.

swatkat1:
### Description

Fix for SELECT_TI Visibility Browser tests failures.

### Bug details
Previously to create a logged in TI experience, the TI Client's email 'testuser@example.com' was hard coded in the browser tests which ended up breaking the test runs in staging environments. Ideally it should be pulled from the environment variables (part of  Github secrets). This fix, instead of having a hard-coded value, will use the export value from config, so that the browser test runs for all environment pass successfully. 

Fixes #5050

### DIFF
--- a/browser-test/src/admin_program_visibility.test.ts
+++ b/browser-test/src/admin_program_visibility.test.ts
@@ -125,7 +125,7 @@ describe('Validate program visibility is correct for applicants and TIs', () => 
     await adminTiGroups.addGroupMember('groupOne@bar.com')
     await adminTiGroups.gotoAdminTIPage()
     await adminTiGroups.editGroup('groupTwo')
-    // GroupTwo will have the TestUser as the TI, so it will be visible to this TI only
+    // Add the TEST_USER_DISPLAY_NAME user to GroupTwo, which is the user that gets logged into upon calling loginAsTestUser()
     await adminTiGroups.addGroupMember(TEST_USER_DISPLAY_NAME)
     await logout(page)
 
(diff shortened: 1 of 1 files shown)


browser-test/src/admin_program_visibility.test.ts
```diff
@@ -125,7 +126,7 @@ describe('Validate program visibility is correct for applicants and TIs', () =>
     await adminTiGroups.gotoAdminTIPage()
     await adminTiGroups.editGroup('groupTwo')
     // GroupTwo will have the TestUser as the TI, so it will be visible to this TI only
```
michaelzetune, line 128: Can we be a little more specific in this comment so future readers understand why we needed to get this from the config? How about "Add the TEST_USER_DISPLAY_NAME user to GroupTwo, which is the user that gets logged into upon calling loginAsTestUser()" ? Here and in the line below as well.

michaelzetune (approved):

dkatzz (approved):

browser-test/src/admin_program_visibility.test.ts
swatkat1: Done

PR #5052, opened 2023-06-12
Fix header bugs when logged in as a TI

### LINKED ISSUE #4976: Don't show the "Applying as x" message when the TI hasn't gone to the applicant dashboard
**Describe the bug**
When a TI signs in initially and goes to the client dashboard, "applying as: Guest" appears:

This is because there isn't a profile set yet, so we assume the "user" is a guest, even though this doesn't make sense. We should hide this text whenever they haven't gone to the applicant dashboard. Notably, TIs may not apply on behalf of their clients as a guest, anyways.

**To Reproduce**

1. Go to TI dashboard on dev mode
2. See the text above

**Expected behavior**
There should be no "Applying as ..." text until the TI selects a client to apply on behalf of

**Screenshots**

image

- OS: macOS
- Browser Brave
- Version 1.51.110

**Additional context**
Related: #4977

See branches:
* https://github.com/civiform/civiform/compare/mzetune/applicant-display-label...mzetune/fake-ti-user
* https://github.com/civiform/civiform/compare/main...mzetune/fake-ti-user-2

### LINKED ISSUE #4977: Fix the Logged In As Section for Fake TIs
**Describe the bug**
For Fake TIs, we currently show "You're a guest user" on the header. This is because we're taking the user attributes from their applicant, instead of themselves.

**To Reproduce**

1. Go to TI dashboard on dev mode
2. See the text above

**Expected behavior**
Show the auth status for the TI, not the applicant.

**Screenshots**
image

- OS: macOS
- Browser Brave
- Version 1.51.110

**Additional context**
Related: #4976

See branches:
* https://github.com/civiform/civiform/compare/mzetune/applicant-display-label...mzetune/fake-ti-user
* https://github.com/civiform/civiform/compare/main...mzetune/fake-ti-user-2

michaelzetune:
### Description

There are a couple bugs on the header when logging in as a TI:

For one, we show `(applying as: <name>)` even when looking on the TI dashboard. We're actually showing this based on the TI's profile, which doesn't make sense as TIs shouldn't apply for themselves:

image
We can simply remove this when we're on the TI dashboard.

Second, we show the phrase "You are a Guest" to TIs even though they've been logged in:

image

This is actually because the TI logic was determining whether someone was a guest based on whether they had a name in the database. Instead, we now always assume that a TI is logged in. Additionally, we seed fake TIs with a name, "Fake_TI", so that we don't run into an error that they're not populated.

## Release notes

Fixes bugs related to the TI header.

#### General

   - Updated screenshots
   - Not necessary

#### User visible changes

   - Updated screenshots
   - TI dashboard is not optimized for mobile view

#### New Features

None.

### Issue(s) this completes

Fixes #4976, fixes #4977

## Meta

There's one more issue that this PR does not address, which is that when the TI goes into the applicant dashboard, the top right header says "Logged in as <client name>" - this is because the TI is impersonating the applicant. However this is confusing because we also say "applying as:" which implies they mean different things. Opened #5054 to track this.

image


server/app/views/applicant/ApplicantLayout.java
```diff
@@ -315,7 +316,8 @@ private DivTag maybeRenderTiButton(
                       "opacity-75",
                       StyleUtils.hover("opacity-100"),
                       ButtonStyles.SOLID_BLUE_TEXT_XL))
-          .with(
+          .condWith(
+              !request.uri().contains("admin/tiDash"),
```
michaelzetune, line 320: I recognize that using the string literal `admin/tiDash` here isn't ideal. Calling `controllers.ti.routes.TrustedIntermediaryController.dashboard(...).url()` isn't possible because it adds some query params to the URL which cannot be compared to for equality with `request.uri()`.

Open to alternate suggestions.

avaleske: Can we strip off the query params before making the comparison? Maybe by dropping it in a `Uri` and pulling out the path to compare?

avaleske (approved):

server/app/views/applicant/ApplicantLayout.java
michaelzetune: Done!

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -463,4 +471,28 @@ private int getPercentComplete(int blockIndex, int totalBlockCount, boolean forS
 
     return (int) (numerator / denominator * 100.0);
   }
+
+  /**
+   * Returns true is the request object points to a URI that is the Trusted Intermediary Dashboard.
```
avaleske, line 476: spelling: if

avaleske (approved):

PR #5007, opened 2023-06-02
Fix issue that allowed applicants to access draft programs.

reemaka:
### Description

If you knew the id of the draft program, you were able to access it by putting its ID into the URL. The method to get the active program was not actually checking if the program was active.

#### General

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release >
- [x] Created tests which fail without the change (if possible)

Issue #240, opened 2021-02-25
Admins can preview program forms in sandbox


neel-a-mehta:
Can "fill in" a program form and "submit it" without anything going to the database. Instead, they can see CSVs/PDFs of what they _would_ have generated.

glorialiou:
Support a “Preview” mode (like Google Forms) for admins for quick validation that content looks as expected.

Currently, there is no quick way to review the user-facing program application before/without publishing. This has led to many versions of CiviForm forms being published during the testing and review process between CF Admin and Program admins, even for minor updates. Frustrating admin experience.

kristinyim:
want a link to preview that admins can send to co-workers for feedback

seanmalbert:
@isabelcshaw and @cassandralfernandes,

This request came from Arkansas recently:
> Is it possible to add a “Preview Form” capability within the CiviForm Admin which allows paging through the screens to review what a form will look like in its entirety. The preview state would not allow any answers to be submitted and would only be used to review each screen for accuracy and useability.

nbeloglazov:
We also have #3080 which I think is talking about the same thing (preview how question/screen looks during building it). Maybe merge them?

reemaka:
Design doc here: https://docs.google.com/document/d/1ACKnO4Z3dU9mAPEx5BnflKqcJ9IeZbbCFCGz2TMVyhE/edit#heading=h.bknox3ib8wpb

PR #5068, opened 2023-06-13
Fix incorrect method name and clean up tests.

reemaka:
### Description

The getActiveProgramDefinitionAsync method name is incorrect since it gets the ProgramDefinition regardless of what version it is. This PR renames the method and related method getProgramForSlug() for clarity.

As part of upcoming feature #240, we'll be adding authorization checks to stop applicants from accessing draft programs. To prepare for that change and as an overall clarity improvement, this also updates the tests to use active programs rather than drafts wherever they logically should be used.

In the future PR that adds the checks, tests will also be added to make sure applicants can't access drafts and that they can access all other version types.

This is a partial roll forward of #5007 which was reverted in #5053.

#### General

#### Fixes

Pre-work for #240


server/app/services/program/ProgramService.java
```diff
@@ -63,14 +63,15 @@ CompletionStage<ImmutableList<ProgramDefinition>> syncQuestionsToProgramDefiniti
       ImmutableList<ProgramDefinition> programDefinitions);
 
   /**
-   * Get the definition of a given program asynchronously.
+   * Get the definition of a given program asynchronously. The returned program may correspond to
```
michaelzetune, line 66: Do you mean that the id may correspond to any version? 

The returned program must be the one with this id (IIUC, one program ID corresponds to one version, let me know if that's off)

michaelzetune (approved):

server/app/services/program/ProgramService.java
reemaka: Yeah that's a better way to phrase it, wasn't sure how to make that clear. Changed it to use your wording.

PR #5091, opened 2023-06-15
Remove Seattle from messages file and disallow other civic entities from appearing

michaelzetune:
### Description

Arkansas pointed out that we have Seattle-specific phrasing in CiviForm - my mistake:

https://github.com/civiform/civiform/blob/d133a18a80ebd08c18115abc95709cebbe3e1643/server/conf/i18n/messages#L139

This PR replaces "City of Seattle" with `{0}`, which is populated with the var `applicant_portal_name`. In browser tests this is "TestPortal".

## Release notes

Fixes Seattle-specific phrasing in login prompt.

#### General

   - Not necessary

#### User visible changes

   - Updated screenshot

#### New Features

None.

### Issue(s) this completes

None.


server/test/conf/MessagesTest.java
```diff
@@ -42,6 +42,18 @@ public class MessagesTest {
...
+  private static final Set<String> PROHIBITED_WORDS =
+      Set.of(
+          "Seattle",
+          "Washington",
+          "Bloomington",
+          "Indiana",
+          "Arkansas",
+          "Charlotte",
+          "North Carolina");
+
```
bion, line 56: 👍🏻👍🏻👍🏻

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -134,7 +134,11 @@ public CompletionStage<Result> review(Request request, long applicantId, long pr
                                 .programBySlug(
                                     request.flash().get("redirected-from-program-slug").get())
                                 .url(),
-                            MessageKey.INITIAL_LOGIN_MODAL_PROMPT,
+                            messages.at(
+                                MessageKey.INITIAL_LOGIN_MODAL_PROMPT.getKeyName(),
+                                // The applicant portal name should always be set (there is a
+                                // default setting as well).
+                                settingsManifest.getApplicantPortalName().orElse("PortalMissing")),
```
bion, line 141: **suggestion (blocking):** a more graceful degradation if the config value is missing would be to use an empty string so the UI reads "log in to your account" rather than "log in to your PortalMissing account"

server/app/views/applicant/ApplicantCommonIntakeUpsellCreateAccountView.java
```diff
@@ -67,7 +67,7 @@ public Content render(
         createLoginPromptModal(
                 messages,
                 redirectTo,
-                /*description=*/ MessageKey.GENERAL_LOGIN_MODAL_PROMPT,
+                /*description=*/ messages.at(MessageKey.GENERAL_LOGIN_MODAL_PROMPT.getKeyName()),
```
bion, line 70: **nitpick:** 
```suggestion
                /* description= */ messages.at(MessageKey.GENERAL_LOGIN_MODAL_PROMPT.getKeyName()),
```

server/test/conf/MessagesTest.java
```diff
@@ -42,6 +42,18 @@ public class MessagesTest {
...
```
bion, line 56: **praise:** thanks for adding this!

bion (changes requested):

server/app/controllers/applicant/ApplicantProgramReviewController.java
michaelzetune: Went ahead and committed the change, but I'm not sure this makes sense in other languages where the subject-object-verb syntax doesn't follow English?

bion (approved):

PR #5108, opened 2023-06-16
Stop non-CiviForm Admin users from applying to draft programs.

reemaka:
### Description

Check that the user is a CiviForm Admin before allowing access to a draft program's application pages.

This is a bug fix but is also set-up work for #240.

#### General

### DIFF
--- a/server/app/controllers/CiviFormController.java
+++ b/server/app/controllers/CiviFormController.java
@@ -1,9 +1,13 @@
 package controllers;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import auth.ProfileUtils;
 import com.google.common.collect.ImmutableSet;
 import java.util.StringJoiner;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import javax.inject.Inject;
 import play.mvc.Controller;
 import play.mvc.Http;
 import repository.VersionRepository;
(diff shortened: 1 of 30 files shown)


server/app/controllers/CiviFormController.java
```diff
@@ -39,4 +40,19 @@ protected CompletableFuture<Void> checkProgramAdminAuthorization(
         .orElseThrow()
         .checkProgramAuthorization(programName, request);
   }
+
+  /** Checks that the profile is authorized to access the specified program. */
+  protected CompletableFuture<Void> checkProgramAuthorization(
+      ProfileUtils profileUtils,
+      VersionRepository versionRepository,
```
bion, line 47: **suggestion (blocking):** Make `profileUtils` and `versionRepository` members of `CivFormController` and pass them via `super(profileUtils, versionRepository)` in the child class c'tors instead of passing them to each method call.

server/app/controllers/CiviFormController.java
```diff
@@ -39,4 +40,19 @@ protected CompletableFuture<Void> checkProgramAdminAuthorization(
...
+      Http.Request request,
+      Long programId) {
+    return CompletableFuture.runAsync(
+        () -> {
+          if (versionRepository.isDraftProgram(programId)
+              && !profileUtils.currentUserProfile(request).orElseThrow().isCiviFormAdmin()) {
+            throw new SecurityException();
+          }
+        });
```
bion, line 56: **suggestion (blocking):** This blocks the a thread from the default thread pool on `versionRepository.isDraftProgram(programId)` to avoid this, add a new method to `versionRepository` that performs the IO using the [database connection threadpool](https://github.com/civiform/civiform/blob/main/server/app/repository/StoredFileRepository.java#L26):

```java
public CompletionStage<Boolean> isDraftProgramAsync(programId) {
  return supplyAsync(() -> isDraftProgram(programId), databaseExecutionContext);
}
```

The code here then consumes the new method and avoid `CompletableFuture.runAsync`

bion (changes requested):

server/app/controllers/CiviFormController.java
reemaka: Done

server/app/controllers/CiviFormController.java
reemaka: Done

bion (approved):

Issue #4872, opened 2023-05-16
Add v0 stub for Program docs


avaleske:
Add some basic v0 docs so AR can see the question names to expect, and the possible values of their answers.

Spec:
- Docs are available at a path like `/api/v1/docs/program-slug` or `/programs/program-slug/docs` (Probably the former)
- Access to docs doesn't require auth
- Docs should show the shape of the application portion of the json response (the entire response shape may be helpful here, but at least the application shape)
- Docs should show the available options for selection questions
- Docs should show the expected format for all answers

Not in scope for this issue:
- Showing the spec for draft programs (that's #4945)
- Using OpenAPI, unless it's very easy (that's  #4946)

Implementation notes:
- We can probably just loop through the program definition and build a json blob, similar to how the current API code does it, and then render that blob in a regular view along with other information about the program json shape.
- https://github.com/civiform/civiform/blob/main/server/app/controllers/api/ProgramApplicationsApiController.java
- https://github.com/civiform/civiform/blob/main/server/app/services/export/JsonExporter.java

PR #5111, opened 2023-06-16
Add `supportedQuestion` to `QuestionType`

michaelzetune:
### Description

Adds a new `supportedQuestion` field to `QuestionType` to make it more clear which question types map to which `Question`s.

Previously, the `Question` was responsible for saying which `QuestionType`s it can represent:

https://github.com/civiform/civiform/blob/9517d9346b401c1e5a7914e47385aa9db0fa0262/server/app/services/applicant/question/SingleSelectQuestion.java#L31-L34

However, this is error prone:
* There could be more than one `Question` that covers a given `QuestionType`, which is illegal 
   * (multiple `QuestionType`s for one `Question` is ok as above)
* We could be missing a `Question` for a given `QuestionType`.

Instead, we now make the `Question` class a part of the `QuestionType`:

https://github.com/civiform/civiform/blob/cf2880023f18c4077797b5b8f0f9f67ab1654bf7/server/app/services/question/types/QuestionType.java#L6-L24

This is also much cleaner as we can remove the `validQuestionTypes()` method.

In addition, remove the `isMultiOptionType` field as it is unnecessary given we only have one.

## Release notes

None.

#### General

   - No need; the new guarantees that are brought as part of this PR are done at compile-time.
   - Not necessary.

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

Relates to #4872 as it will clean up question types in a way that will help with autogenerated docs.


server/app/services/applicant/question/Question.java
```diff
@@ -29,13 +28,15 @@ public abstract class Question {
...
-              "Question is not a question of the following types: [%s]: %s (type: %s)",
-              Joiner.on(", ").join(validQuestionTypes().stream().toArray()),
-              applicantQuestion.getQuestionDefinition().getQuestionPathSegment(),
-              applicantQuestion.getQuestionDefinition().getQuestionType()));
+              "%s is not equal to question type %s's supported %s",
+              getClass(),
+              supportedQuestionType,
+              supportedQuestionType.getSupportedQuestion().toString()));
     }
   }
```
michaelzetune, line 41: The new error message looks like:

```
java.lang.RuntimeException: Class public final class services.applicant.question.TextQuestion is not equal to question type ADDRESS's supported class class services.applicant.question.AddressQuestion
```

avaleske: suggestion: this might be clearer if there format string was
```
The question type \"%s\" is not equal to question type supported by \"%s\", which is %s
```

avaleske (approved):
This is great thank you!

server/app/services/applicant/question/Question.java
michaelzetune: Made a slight modification to your suggestion and committed.

PR #5123, opened 2023-06-20
Update fmt prettier check to include Markdown

nb1701:
This updates the prettier check in bin/fmt to include Markdown files in addition to TS and JSON. Additionally, it optimizes the call to prettier to only operate on changed and untracked files, speeding it up a bit.

#### General

### DIFF
--- a/formatter/fmt
+++ b/formatter/fmt
@@ -39,11 +39,11 @@ function get_modified_java_files() {
   ) | grep -E ".+\.java$"
 }
 
-function get_modified_ts_and_json_files() {
+function get_modified_ts_json_and_md_files() {
   (
     get_tracked_modified_files
     get_untracked_files
-  ) | grep -E ".+\.(ts|json)$"
+  ) | grep -E ".+\.(ts|json|md)$"
 }
 
 function get_modified_python_files() {
@@ -61,8 +61,9 @@ else
   echo 'Done formatting java'
 fi
 
-if [[ -z $(get_modified_ts_and_json_files) ]]; then
-  echo 'No modified TypeScript or JSON files found'
+modified_ts_json_md=$(get_modified_ts_json_and_md_files)
+if [[ -z $modified_ts_json_md ]]; then
+  echo 'No modified TypeScript, JSON or Markdown files found'
 else
   echo 'Start formatting with prettier'
   # prettier is installed as node module in `formatter` directory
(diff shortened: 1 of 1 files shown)


formatter/fmt
```diff
@@ -61,8 +61,9 @@ else
   echo 'Done formatting java'
 fi
 
-if [[ -z $(get_modified_ts_and_json_files) ]]; then
-  echo 'No modified TypeScript or JSON files found'
+modified_ts_json_md=$(get_modified_ts_json_and_md_files)
+if [[ -z $modified_ts_json_md ]]; then
```
gwendolyngoetz, line 65:
```suggestion
if [[ -z "${modified_ts_json_md}" ]]; then
```

formatter/fmt
```diff
@@ -61,8 +61,9 @@ else
   echo 'Done formatting java'
 fi
 
-if [[ -z $(get_modified_ts_and_json_files) ]]; then
-  echo 'No modified TypeScript or JSON files found'
+modified_ts_json_md=$(get_modified_ts_json_and_md_files)
```
gwendolyngoetz, line 64:
```suggestion
modified_ts_json_md="$(get_modified_ts_json_and_md_files)"
```

formatter/fmt
```diff
@@ -61,8 +61,9 @@ else
...
+if [[ -z $modified_ts_json_md ]]; then
+  echo 'No modified TypeScript, JSON or Markdown files found'
```
gwendolyngoetz, line 66:
```suggestion
  echo 'No modified TypeScript, JSON, or Markdown files found'
```

formatter/fmt
```diff
@@ -72,7 +73,7 @@ else
     --config ../.prettierrc.js \
     --ignore-path ../.prettierignore \
     --ignore-unknown \
-    ../
+    $(printf "../%s " ${modified_ts_json_md[@]})
```
gwendolyngoetz, line 76:
```suggestion
    "$(printf "../%s " "${modified_ts_json_md[@]}")"
```

gwendolyngoetz: @nb1701 Do we anticipate any of these files having spaces in the file name?

gwendolyngoetz (changes requested):
@nb1701 Some updates per the Bash style guide.

Here's the explanations:
- https://google.github.io/styleguide/shellguide.html#variable-expansion
- https://google.github.io/styleguide/shellguide.html#quoting

gwendolyngoetz (approved):

Issue #3286, opened 2022-08-26
Prefix error messages with "Error:"


jeanniefu:
From our accessibility advisors - Accessibility criteria: "Color is not used as the sole means of indicating the error message".

Currently, we display error messages in red without another indication that is is an error. This does not work well for users who are color-blind. We should include the text "Error" before the message.

melanie-exygy:
@sijiayam I'm working on this as a first issue.  Do you think this applies to toast error messages, error pages or both?  Right now toast errors look like this:  
![image](https://github.com/civiform/civiform/assets/134556723/c3a7fad8-3ff3-4302-87b5-936387a7c4f4)
and error pages look like this:  
![image](https://github.com/civiform/civiform/assets/134556723/3eef9f55-8036-4a06-bc42-069ada14d891)

PR #5127, opened 2023-06-21
Adding "Error: " as prefix on error toast messages, for accessibility

melanie-exygy:
### Description

Adding "Error: " as prefix on error toast messages, for accessibility.  This way, identifying an error isn't solely based on the color of the toast.

#### General

#### User visible changes

### Issue(s) this completes

Fixes #<[3286](https://github.com/civiform/civiform/issues/3286)>;


dkatzz (approved):
You can remove the "new features" section to the description since it doesn't apply in this case (and other checkboxes that don't apply) (see [here](https://docs.civiform.us/contributor-guide/developer-guide/technical-contribution-guide#guidelines-for-pull-requests)).

PR #5132, opened 2023-06-23
Add eligibility status to questions on PDF export

### LINKED ISSUE #4455: Show eligibility status on the question level of admin application PDF export
This is a follow-up to adding this information to the [applications view](https://github.com/civiform/civiform/issues/4446). We should also show question level eligibility on the PDF export.

dkatzz:
### Description

Add eligibility status to questions on PDF export when eligibility conditions are applied to the program and question.

#### General

### Issue(s) this completes

Fixes #4455

### DIFF
--- a/server/app/services/export/PdfExporter.java
+++ b/server/app/services/export/PdfExporter.java
@@ -25,6 +25,9 @@ import models.Application;
 import services.applicant.AnswerData;
 import services.applicant.ApplicantService;
 import services.applicant.ReadOnlyApplicantProgramService;
+import services.program.EligibilityDefinition;
+import services.program.ProgramBlockDefinitionNotFoundException;
+import services.program.ProgramDefinition;
 
 /** PdfExporter is meant to generate PDF files. */
 public final class PdfExporter {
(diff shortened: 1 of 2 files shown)


server/app/services/export/PdfExporter.java
```diff
@@ -126,9 +130,38 @@ private byte[] buildPDF(
         Paragraph time =
             new Paragraph("Answered on : " + date, FontFactory.getFont(FontFactory.HELVETICA, 10));
         time.setAlignment(Paragraph.ALIGN_RIGHT);
+        Paragraph eligibility = new Paragraph();
+        if (isEligibilityEnabledInProgram) {
+          try {
+            Optional<EligibilityDefinition> eligibilityDef =
+                programDefinition.getBlockDefinition(answerData.blockId()).eligibilityDefinition();
+            if (eligibilityDef.isPresent()
+                && eligibilityDef
```
bion, line 139: **nitpick (blocking):** with `map` and then `orElse(false)` I don't think you need `eligibilityDef.isPresent() &&`?

bion (approved):

server/app/services/export/PdfExporter.java
dkatzz: Yes good catch - thanks!

PR #5135, opened 2023-06-26
Add an is_admin column to the applications table.

reemaka:
### Description

Add an is_admin column to the applications table which indicates whether or not the application belongs to an admin. As part of the program preview feature (#240), draft applications will be created for admins who are previewing a program as an applicant. Although admin applications will never be submitted, the new is_admin column will make it easy to filter out draft admin applications for future use cases such as tracking abandoned applications.

#### General

### Issue(s) this completes

Part of #240


server/test/auth/CiviFormProfileMergerTest.java
```diff
@@ -62,9 +62,10 @@ public void setup() {
 
     account.setApplicants(Collections.singletonList(applicant));
 
+    when(applicant.getAccount()).thenReturn(account);
```
rockycodes, line 65: question (non-blocking): Just curious why you moved this?

rockycodes (approved):
lgtm!

server/test/auth/CiviFormProfileMergerTest.java
reemaka: The Applicant object needs to have an Account before the Application is created on line 67, otherwise there will be a null exception when getAccount() is called in the Application constructor. This is only an issue for tests since in reality the Applicant will always have an associated Account.

rockycodes: ah gotcha. thanks for explaining!

PR #5137, opened 2023-06-26
Disallow wildcard imports with ErrorProne

michaelzetune:
### Description

Disallows wildcard imports with ErrorProne in our `build.sbt` using `-Xep:WildcardImport:ERROR`


bion (approved):
Thanks!

PR #5151, opened 2023-06-28
Adding prefix to error toast messages, localized when applicant-facing, for accessibility

melanie-exygy:
### Description

- Added "Error: " as a prefix on toast messages, as an accessibility enhancement.  For toast messages which were already being localized, the prefix will also be localized.  For other admin-facing toast messages, the prefix is added in English.
-  Turned ToastMessage.java into more of a factory class for consistency in how we create toasts and so that new error toast messages will include the prefix.
- Note that by moving this prefix logic server-side, any client-side-only toast messages will not include the prefix.  So far, I have only seen one instance of a client-side-only toast and that was during admin predicates editing.  I will create a tech debt item to address that edge-case.  

## Release notes

Added "Error: " as a prefix on toast messages, as an accessibility enhancement.

#### General

#### User visible changes

- [x] Followed steps to [internationalize new strings](https://docs.civiform.us/contributor-guide/developer-guide/internationalization-i18n#internationalization-for-application-strings)

1. As a CF admin, set a visibility condition on a program block
2. Publish the program
3. Log out
4. Open the application for the program
5. Fill out the question with an answer that doesn't display the next block. Save and go to the review page.
6. Clone the tab
7. In the new tab, change the answer to the question to an answer that would display the next block and save.
8. Go back to the original tab and hit submit
9. An error toast should show saying the application has changed.  It should begin with "Error: " and this prefix should be localized.

### Issue(s) this completes

Fixes [#3286](https://github.com/civiform/civiform/issues/3286);


server/app/views/components/ToastMessage.java
melanie-exygy, line 1: These are the core changes in this PR.  Start here.

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -250,17 +247,19 @@ private CompletionStage<Result> submitInternal(
                 if (cause instanceof ApplicationSubmissionException) {
                   Call reviewPage =
                       routes.ApplicantProgramReviewController.review(applicantId, programId);
+                  Messages msgProvider = messagesApi.preferred(request);
                   String errorMsg =
-                      messagesApi
-                          .preferred(request)
-                          .at(MessageKey.BANNER_ERROR_SAVING_APPLICATION.getKeyName());
+                      msgProvider.at(
```
melanie-exygy, line 252: This is other key change of this PR and is an example of how localized error messages will be handled.

server/app/views/components/ToastMessage.java
```diff
@@ -44,7 +44,8 @@ public static ToastMessage alert(String message) {
     return new ToastMessage(message, ToastType.ALERT);
   }
 
-  public static ToastMessage error(String message) {
+  public static ToastMessage error(String message, boolean isLocalized) {
```
michaelzetune, line 47: We shouldn't have an `isLocalized` param here; if a language isn't localized yet, we will default to English, so "Error: " will be prepended until we get that TOAST_ERROR_MSG_OUTLINE string translated later, anyways. We don't want to remember to change this later on.

server/app/views/components/ToastMessage.java
```diff
@@ -44,7 +44,8 @@ public static ToastMessage alert(String message) {
...
+    if (!isLocalized) message = "Error: " + message;
```
michaelzetune, line 48: I prefer not to force clients to use TOAST_ERROR_MSG_OUTLINE themselves, if every error toast will use it. How about this implementation which does it on behalf of the clients, so files like `ApplicantProgramReviewController` don't have to?

```java
  public static ToastMessage error(String message, Messages messages) {
    String errMessageWithPrefix = messages.at(
      MessageKey.TOAST_ERROR_MSG_OUTLINE.getKeyName(), message);
    return new ToastMessage(errMessageWithPrefix, ToastType.ERROR);
  }
```

michaelzetune (changes requested):

michaelzetune (changes requested):

server/app/views/components/ToastMessage.java
melanie-exygy: The reason for the `isLocalized` param isn't for languages that aren't localized yet, it's for all the error toasts on the admin side which aren't being localized at all (aren't using the Play MessageProvider).  This is the case for all but two of the error toasts.

michaelzetune: Got it - sorry I didn't pay closer attention. In that case, can we make two methods, `ToastMessage.errorNonLocalized()` and `ToastMessage.error()` ? I think that will be easier to understand than a raw boolean passed in as the second param.

My second comment still stands for `ToastMessage.error()`, especially as we may get more toast errors on the applicant side in the future.

server/app/views/components/ToastMessage.java
melanie-exygy: This seems like a very good idea, although I'll have to modify it to accommodate all the admin error toasts which aren't using `play.i18n.Messages.`  I guess I could check if `messages` is null rather than having a third `isLocalized` param.

michaelzetune: For the admin error toasts, `ToastMessage.errorNonLocalized()` would not have a `Messages` param. So it would be a minimal change, for those, right?

server/app/views/components/ToastMessage.java
melanie-exygy: I agree that having two methods will be clearer.  Thanks for the good idea!  I'll go ahead and add that.

server/app/views/components/ToastMessage.java
michaelzetune: Applicant error toasts call `ToastMessage.error(String message, Messages messages)`

Admin error toasts call `ToastMessage.errorNonLocalized(String message)`

So no need for a isLocalized param. Make sense? Am I missing anything?

melanie-exygy: Yep, makes a lot of sense.  I think that will work well.

browser-test/image_snapshots/admin_predicates_test/predicate-error.png
michaelzetune, line 1: Understood why this screenshot changed. Just curious, why aren't there any screenshot diffs with "Error: " being introduced? Do we just not have any tests covering that?

michaelzetune (approved):

browser-test/image_snapshots/admin_predicates_test/predicate-error.png
melanie-exygy: That's a good callout.  The reason is that I already added "Error: " to two tests in my first PR to address this issue.  See:  [#5127](https://github.com/civiform/civiform/pull/5127/files).

PR #5163, opened 2023-06-29
Remove setting default ValidationPredicates

michaelzetune:
### Description

Currently, we require a `ValidationPredicates` to construct a `QuestionDefinitionConfig`:

https://github.com/civiform/civiform/blob/3717ab8fec91c414c018d467f039937705c279e4/server/app/services/question/types/QuestionDefinitionConfig.java#L10-L33

However, the vast majority of the time, this is set to the default one, like:

https://github.com/civiform/civiform/blob/3717ab8fec91c414c018d467f039937705c279e4/server/test/forms/NameQuestionFormTest.java#L39-L48

Instead, we can simply set it to the default in the `QuestionDefinitionConfig` abstract constructor, using a new `getDefaultValidationPredicates()` abstract method. This allows us to delete all the code like `.setValidationPredicates(NameQuestionDefinition.NameValidationPredicates.create())`, which there is **a lot** of.

## Release notes

None.

#### General

   - Not applicable
   - Not applicable
   - Not applicable

#### User visible changes

None.

#### New Features

None.

None.

### Issue(s) this completes

None.

### DIFF
--- a/server/app/services/question/types/QuestionDefinition.java
+++ b/server/app/services/question/types/QuestionDefinition.java
@@ -174,7 +174,7 @@ public abstract class QuestionDefinition {
 
   /** Get the validation predicates. */
   public final ValidationPredicates getValidationPredicates() {
-    return config.validationPredicates().orElse(getDefaultValidationPredicates());
+    return config.validationPredicates().orElseGet(this::getDefaultValidationPredicates);
   }
 
   /** Serialize validation predicates as a string. This is used for persisting in database. */
--- a/server/app/services/question/types/QuestionDefinition.java
+++ b/server/app/services/question/types/QuestionDefinition.java
@@ -24,6 +24,10 @@ public abstract class QuestionDefinition {
   private final QuestionDefinitionConfig config;
 
   protected QuestionDefinition(QuestionDefinitionConfig config) {
+    if (config.validationPredicates().isEmpty()) {
+      config = config.toBuilder().setValidationPredicates(getDefaultValidationPredicates()).build();
+    }
+
     this.config = config;
   }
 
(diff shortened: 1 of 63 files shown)


server/app/services/question/types/QuestionDefinition.java
```diff
@@ -24,6 +24,10 @@ public abstract class QuestionDefinition {
   private final QuestionDefinitionConfig config;
 
   protected QuestionDefinition(QuestionDefinitionConfig config) {
+    if (config.validationPredicates().isEmpty()) {
+      config = config.toBuilder().setValidationPredicates(getDefaultValidationPredicates()).build();
+    }
+
```
michaelzetune, line 30: Technically, we could remove the above and leave `validationPredicates()` as empty in `QuestionDefinitionConfig`, since `getValidationPredicates()` on line 176 is handling that case anyways. Any thoughts welcome.

server/app/services/question/types/QuestionDefinitionConfig.java
```diff
@@ -53,20 +53,12 @@ public interface RequiredQuestionText {
   }
 
   public interface RequiredQuestionHelpText {
-    RequiredValidationPredicates setQuestionHelpText(LocalizedStrings questionHelpText);
-  }
-
-  public interface RequiredValidationPredicates {
-    Builder setValidationPredicates(QuestionDefinition.ValidationPredicates validationPredicates);
+    Builder setQuestionHelpText(LocalizedStrings questionHelpText);
   }
```
michaelzetune, line 57: Removing `RequiredValidationPredicates` makes `validationPredicates()` not required to be set at compile-time.

server/app/services/question/types/QuestionDefinition.java
```diff
@@ -170,17 +174,20 @@ public ImmutableSet<Locale> getSupportedLocales() {
 
   /** Get the validation predicates. */
   public final ValidationPredicates getValidationPredicates() {
-    return config.validationPredicates();
+    return config.validationPredicates().orElse(getDefaultValidationPredicates());
```
swatkat1, line 177: I think we should be using orElseGet() here instead of orElse(). The orElse() is evaluated, even when having a non-empty Optional.

michaelzetune: Changed.

server/app/services/question/types/QuestionDefinition.java
swatkat1: I think it's safe to remove.

server/app/services/question/types/QuestionDefinitionConfig.java
swatkat1: QQ- then why was it implemented before? Can you please explain

server/app/services/question/types/QuestionDefinition.java
michaelzetune: Removed.

PR #5170, opened 2023-06-30
Clear the redirect target if the user navigates to the index page.

### LINKED ISSUE #5167: Admin cannot access programs without going through preview
**Describe the bug**
When I login as a program admin via dev tools I cannot see the full list of programs until I complete the program I landed on.

**To Reproduce**

1. Load the homepage
2. Click the dev tools and login as a CiviForm admin
3. See program page with no way to get to the list of programs

Note: This bug seems inconsistent so you might need to load up a specific program page first (eg https://staging-aws.civiform.dev/programs/test), then go to the home page, then login as an admin via dev tools.

**Expected behavior**
When logging in as an admin I expect to see the page with the list of programs.

**Screenshots**
Video: https://github.com/civiform/civiform/assets/24319951/e1cb5e08-98e1-4afe-919a-31c4fc896b29

- OS: iOS
- Browser: seeing this in both Chrome and Firefox

reemaka:
### Description

This is a patch for #5167. The REDIRECT_TO_SESSION_KEY is set when the navigating to a program by the slug and doesn't get cleared properly. This is only a temporary fix that clears it if the user goes to the index page. We should follow up with a solution that fixes the underlying problem. Filed #5171 to track that.

#### General

### Issue(s) this completes

Fixes #5167

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramsController.java
+++ b/server/app/controllers/applicant/ApplicantProgramsController.java
@@ -1,6 +1,7 @@
 package controllers.applicant;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static controllers.CallbackController.REDIRECT_TO_SESSION_KEY;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import auth.CiviFormProfile;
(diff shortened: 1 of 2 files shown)


michaelzetune (approved):
Thanks for the fix. I'll assign myself to #5171 since I think I may have caused it [here](https://github.com/civiform/civiform/pull/4877/files#diff-511191c0bef138ec469baae2a63f913fd4769fdc3999a2f89131c1eb9a616988).

PR #5172, opened 2023-06-30
Fix Program Slug Redirect

### LINKED ISSUE #5171: programBySlug redirect does not get cleared properly
**Describe the bug**
If you navigate directly to a program by the slug with `/programs/<slug>`, RedirectController.programBySlug() adds itself as a redirect destination in REDIRECT_TO_SESSION_KEY. The key only gets cleared if the user follows the expected path. If they load a different program by ID, the key is not cleared and the user can be unexpectedly redirected to the initial program. This happens even if the initial program doesn't exist and results in a ProgramNotFound error.

There are also no tests for the redirect, so we should add those.

This is related to #5167 and is slightly mitigated by #5170, but we should revisit the way REDIRECT_TO_SESSION_KEY works to fix the underlying issue.

**To Reproduce**

1. Go to https://staging-aws.civiform.dev/programs/nonexistent-program
2. See program not found exception
3. Go to https://staging-aws.civiform.dev/programs/simple-program (has to be a program that does exist)
4. Close the log in dialog.
5. Click `End Session`
6. See program not found exception for nonexistent-program.

**Expected behavior**
1. Ending session should always redirect to program index page instead of to a specific program.
2. After navigating to any other screen in the app, the user should never be redirected to programBySlug.

michaelzetune:
### Description

`RedirectController.programBySlug` inserts a `REDIRECT_TO_SESSION_KEY` if the user doesn't have a profile yet, so that they are redirected to the `/programs/slug-link` URL again after getting a guest profile.

Previously we did not clear this out after sending the user to the full program link, which led to #5171. This fixes that by `removingFromSession` the `REDIRECT_TO_SESSION_KEY` before redirecting them to the full program link.

## Release notes

Fixes a bug with program slug links.

#### General

#### User visible changes

None.

#### New Features

None.

### Issue(s) this completes

Fixes #5171

### DIFF
--- a/browser-test/src/program_deep_link.test.ts
+++ b/browser-test/src/program_deep_link.test.ts
@@ -135,4 +135,21 @@ describe('navigating to a deep link', () => {
 
     await logout(page)
   })
+
+  it('Going to a deep link does not retain redirect in session', async () => {
+    await resetContext(ctx)
+    const {page, browserContext} = ctx
+    await browserContext.clearCookies()
+
+    // Go to a deep link
+    await gotoEndpoint(page, '/programs/test-deep-link')
+    await page.click('text="Continue to application"')
+
+    // Logging out should not take us back to /programs/test-deep-link, but rather
+    // to the program index page.
+    await logout(page)
+    expect(await page.innerText('h1')).toContain(
+      'Save time when applying for benefits',
+    )
+  })
 })
--- a/server/app/controllers/applicant/RedirectController.java
+++ b/server/app/controllers/applicant/RedirectController.java
@@ -75,8 +75,12 @@ public final class RedirectController extends CiviFormController {
     Optional<CiviFormProfile> profile = profileUtils.currentUserProfile(request);
 
     if (profile.isEmpty()) {
-      Result result = redirect(routes.CallbackController.callback(GuestClient.CLIENT_NAME).url());
-      result = result.withSession(ImmutableMap.of(REDIRECT_TO_SESSION_KEY, request.uri()));
+      // If there isn't currently a session, create a guest session using the CallbackController,
+      // and add a session key that asks it to redirect back here when the profile will be present.
+      Result result =
+          redirect(routes.CallbackController.callback(GuestClient.CLIENT_NAME).url())
+              .withSession(ImmutableMap.of(REDIRECT_TO_SESSION_KEY, request.uri()));
+
       return CompletableFuture.completedFuture(result);
     }
 
@@ -107,32 +111,35 @@ public final class RedirectController extends CiviFormController {
                         // for this program, redirect to program version associated
                         // with that application if so.
                         if (programForExistingApplication.isPresent()) {
+                          long programId = programForExistingApplication.get().id();
                           return CompletableFuture.completedFuture(
-                              redirect(
-                                      controllers.applicant.routes.ApplicantProgramReviewController
-                                          .review(
-                                              applicantId,
-                                              programForExistingApplication.get().id()))
-                                  .flashing("redirected-from-program-slug", programSlug));
+                              goToActiveProgram(programId, applicantId, programSlug, request));
+                        } else {
+                          return programService
+                              .getActiveProgramDefinitionAsync(programSlug)
+                              .thenApply(
+                                  activeProgramDefinition ->
+                                      goToActiveProgram(
+                                          activeProgramDefinition.id(),
+                                          applicantId,
+                                          programSlug,
+                                          request));
                         }
-
-                        return redirectToActiveProgram(applicantId, programSlug);
                       },
                       httpContext.current());
             },
             httpContext.current());
   }
 
-  private CompletionStage<Result> redirectToActiveProgram(long applicantId, String programSlug) {
-    return programService
-        .getActiveProgramDefinitionAsync(programSlug)
-        .thenApplyAsync(
-            (activeProgramDefinition) ->
-                redirect(
-                        controllers.applicant.routes.ApplicantProgramReviewController.review(
-                            applicantId, activeProgramDefinition.id()))
-                    .flashing("redirected-from-program-slug", programSlug),
-            httpContext.current());
+  private Result goToActiveProgram(
+      long programId, long applicantId, String programSlug, Http.Request request) {
+    return redirect(
+            controllers.applicant.routes.ApplicantProgramReviewController.review(
+                applicantId, programId))
+        .flashing("redirected-from-program-slug", programSlug)
+        // If we had a redirectTo session key that redirected us here, remove it so that it doesn't
+        // get used again.
+        .removingFromSession(request, REDIRECT_TO_SESSION_KEY);
   }
 
   private CompletionStage<Optional<ProgramDefinition>> getProgramVersionForApplicant(
--- a/server/app/controllers/applicant/RedirectController.java
+++ b/server/app/controllers/applicant/RedirectController.java
@@ -125,14 +125,9 @@ public final class RedirectController extends CiviFormController {
                                           programSlug,
                                           request))
                               .exceptionally(
-                                  ex -> {
-                                    Throwable cause = ex.getCause().getCause();
-                                    if (cause instanceof ProgramNotFoundException) {
-                                      return notFound()
-                                          .removingFromSession(request, REDIRECT_TO_SESSION_KEY);
-                                    }
-                                    throw new RuntimeException(ex);
-                                  });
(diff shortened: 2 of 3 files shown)


browser-test/src/program_deep_link.test.ts
```diff
@@ -135,4 +135,21 @@ describe('navigating to a deep link', () => {
...
+    await gotoEndpoint(page, '/programs/test-deep-link')
+    await page.click('text="Continue to application"')
+
+    // Logging out should not take us back to /programs/test-deep-link, but rather
+    // to the program index page.
+    await logout(page)
+    expect(await page.innerText('h1')).toContain(
+      'Save time when applying for benefits',
+    )
+  })
```
michaelzetune, line 154: Confirmed the above fails before the changes to `RedirectController`, and passes after.

server/app/controllers/applicant/RedirectController.java
```diff
@@ -75,8 +75,12 @@ public CompletionStage<Result> programBySlug(Http.Request request, String progra
...
 
     if (profile.isEmpty()) {
-      Result result = redirect(routes.CallbackController.callback(GuestClient.CLIENT_NAME).url());
-      result = result.withSession(ImmutableMap.of(REDIRECT_TO_SESSION_KEY, request.uri()));
+      // If there isn't currently a session, create a guest session using the CallbackController,
+      // and add a session key that asks it to redirect back here when the profile will be present.
+      Result result =
+          redirect(routes.CallbackController.callback(GuestClient.CLIENT_NAME).url())
+              .withSession(ImmutableMap.of(REDIRECT_TO_SESSION_KEY, request.uri()));
+
```
michaelzetune, line 83: Above is a no-op refactor for clarity.

server/app/controllers/applicant/RedirectController.java
```diff
@@ -107,32 +111,35 @@ public CompletionStage<Result> programBySlug(Http.Request request, String progra
...
-    return programService
-        .getActiveProgramDefinitionAsync(programSlug)
-        .thenApplyAsync(
-            (activeProgramDefinition) ->
-                redirect(
-                        controllers.applicant.routes.ApplicantProgramReviewController.review(
-                            applicantId, activeProgramDefinition.id()))
-                    .flashing("redirected-from-program-slug", programSlug),
-            httpContext.current());
+  private Result goToActiveProgram(
```
reemaka, line 134: This method no longer takes you to the active program and should be renamed

server/app/controllers/applicant/RedirectController.java
```diff
@@ -107,32 +111,35 @@ public CompletionStage<Result> programBySlug(Http.Request request, String progra
...
+      long programId, long applicantId, String programSlug, Http.Request request) {
+    return redirect(
+            controllers.applicant.routes.ApplicantProgramReviewController.review(
+                applicantId, programId))
+        .flashing("redirected-from-program-slug", programSlug)
```
reemaka, line 139: Does this show a toast to applicants? Do we want that?

server/app/controllers/applicant/RedirectController.java
```diff
@@ -107,32 +111,35 @@ public CompletionStage<Result> programBySlug(Http.Request request, String progra
...
-                              redirect(
-                                      controllers.applicant.routes.ApplicantProgramReviewController
-                                          .review(
-                                              applicantId,
-                                              programForExistingApplication.get().id()))
-                                  .flashing("redirected-from-program-slug", programSlug));
+                              goToActiveProgram(programId, applicantId, programSlug, request));
+                        } else {
+                          return programService
+                              .getActiveProgramDefinitionAsync(programSlug)
```
reemaka, line 119: Can you add handling for the case where there's no active program found? We should still clear the session in that case

reemaka (commented):
Does this fix work for admin roles and for the steps to reproduce listed in #5171?

server/app/controllers/applicant/RedirectController.java
michaelzetune: Renamed to `redirectToReviewPage`

server/app/controllers/applicant/RedirectController.java
michaelzetune: Nope, it's used to show the login modal:

https://github.com/civiform/civiform/blob/c3a918af28eded0b86c794c40ea5b73ae84bd38e/server/app/controllers/applicant/ApplicantProgramReviewController.java#L126-L148

server/app/controllers/applicant/RedirectController.java
michaelzetune: We clear the session in both cases, calling `redirectToReviewPage` (previously `goToActiveProgram`), which has the `removingFromSession` line.

server/app/controllers/applicant/RedirectController.java
michaelzetune: Previously we had `.flashing("redirected-from-program-slug", programSlug)` in both cases, so we've just consolidated it in to one method.

server/app/controllers/applicant/RedirectController.java
reemaka: getActiveProgramDefinitionAsync() can throw a ProgramNotFoundException. I'm referring to handling that exception.

michaelzetune:
> Does this fix work for admin roles and for the steps to reproduce listed in #5171?

It does work for your steps to reproduce. What do you mean by for admin roles? Do admins also use deep links somehow?

reemaka:
> It does work for your steps to reproduce. 

Cool, thanks for verifying that!

> What do you mean by for admin roles? Do admins also use deep links somehow?

For example, if you go to a deep link and then log in as an admin, like Rocky did in https://github.com/civiform/civiform/issues/5167

server/app/controllers/applicant/RedirectController.java
michaelzetune: Added. It doesn't seem possible to modify the `Request` directly, only to call `removingFromSession` on a new `Result`. So we have to return a `notFound()` (a Result), and the pass the exception's message. But since it's buried two layers deep, we get this:

image

We can call `ex.getCause().getCause()` to get the underlying message, but that couples the code with the cause  being two layers deep, so the simplest thing to do might be to leave the exception as-is. Open to your thoughts if I missed something.

michaelzetune:
> > It does work for your steps to reproduce.
> 
> Cool, thanks for verifying that!
> 
> > What do you mean by for admin roles? Do admins also use deep links somehow?
> 
> For example, if you go to a deep link and then log in as an admin, like Rocky did in #5167

I can't seem to replicate the issue on main...could you pull this branch and check?

server/app/controllers/applicant/RedirectController.java
reemaka: I think that's fine, just want to make sure they're not stuck in a loop.

server/app/controllers/applicant/RedirectController.java
```diff
@@ -107,32 +111,39 @@ public CompletionStage<Result> programBySlug(Http.Request request, String progra
...
-            httpContext.current());
+  private Result redirectToReviewPage(
+      long programId, long applicantId, String programSlug, Http.Request request) {
+    return redirect(
+            controllers.applicant.routes.ApplicantProgramReviewController.review(
+                applicantId, programId))
+        .flashing("redirected-from-program-slug", programSlug)
+        // If we had a redirectTo session key that redirected us here, remove it so that it doesn't
+        // get used again.
+        .removingFromSession(request, REDIRECT_TO_SESSION_KEY);
```
reemaka, line 146: **optional:** I think you could revert my patch from earlier now that this is in place.

server/app/controllers/applicant/RedirectController.java
reemaka, line 1: Are there any tests you could add to RedirectControllerTest?

reemaka (approved):
checked out your branch and verified it works if you login as admin, thanks!

server/app/controllers/applicant/RedirectController.java
michaelzetune: I think we're better off keeping it since it's better practice to never have a `redirectTo` when landing on the index page.

Ideally we would avoid redirects in `LogoutController`, but the version of that class that we're using is Play's, and it doesn't seem worth it to try to mess with it.

server/app/controllers/applicant/RedirectController.java
michaelzetune: Added some tests for existing and non-existing programs.

Issue #5146, opened 2023-06-27
When visibility or eligibility condition contains quotes, unhandled exception occurs


nb1701:
**Describe the bug**
When a visibility condition string contains quotes, an unhandled exception occurs when a user goes to apply for a program. This seems like something an admin may mistakenly do, and in fact, it is something that is very easily triggered by editing existing visibility conditions.  When you go to edit an existing condition, if it contains a string to match, it is displayed with "" around it even when that wasn't what was input originally.  When a user edits the visibility condition, not changing the string, and saves it, that string now has "" "" around it.  We keep appending "" every time.

Note that this also occurs if you set up an eligibility condition in the same manner.

**To Reproduce**

1. Create a new program
2. Add two screens. The first one containing the name question, the second one containing the DOB question.
3. On the second screen, create a visibility condition, such as "Screen is hidden if first name is equal to abc"
4. Save condition.
5. Edit the visibility condition.
6. Note how it now shows the name as "abc" with the quotes.  Change it to screen is shown, or something, without changing the string.
7. If you edit the condition again, you'll note now the string is ""abc"", but you don't need to make any more changes.
8. Publish the program.
9. Apply to the program as an applicant.
10. Unhandled exception occurs

**Expected behavior**
1. Do not fill in the existing visibility/eligibility condition strings with "" around it.
2. Validate that visibility/eligibility condition strings do not contain quotes before accepting the input.

**Screenshots**
![image](https://github.com/civiform/civiform/assets/122842472/37514f52-636f-4c2d-b092-c3f174217d3b)

PR #5174, opened 2023-07-03
Fix use of quotes in predicates

nb1701:
When saving a string used in a predicate, we must wrap double quotes around the string in order for it to be handled correctly. When editing an existing predicate, it would render the string with the double quotes intact. This meant that when it saved the predicate, it would wrap it again in double quotes. This would cause an unhandled exception when a user went to apply for a program.

With these changes, we now render the predicate string without double quotes. Then upon saving, we strip the value of all double quotes, then wrap it with double quotes. This should ensure the value is always in the format of "value" and won't result in JSON pathing errors.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/5146

### DIFF
--- a/server/app/services/program/predicate/PredicateValue.java
+++ b/server/app/services/program/predicate/PredicateValue.java
@@ -135,7 +135,26 @@ public abstract class PredicateValue {
         .orElse("<obsolete>");
   }
 
+  /**
+   * Remove any double quotes currently in the string, as this would interfere with how we store the
+   * predicate, then surround the string with a pair of double quotes.
+   *
+   * @param s the string to surround in double quotes
+   * @return the same string with double quotes removed, then surrounded by double quotes
+   */
   private static String surroundWithQuotes(String s) {
-    return "\"" + s + "\"";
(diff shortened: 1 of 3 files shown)


server/app/services/program/predicate/PredicateValue.java
```diff
@@ -135,7 +136,28 @@ private static String parseMultiOptionIdToText(
...
+   * Get the stringified value of the predicate without surrounding quotes. Used for displaying the
+   * predicate string in the UI. Only removes surrounding quotes from plain string predicate types.
+   * Lists of strings will still contain double quotes around each item in the list.
+   *
+   * @return A plain string with surrounding double quotes removed, or the result of value() if it
+   *     is not a plain string type.
+   */
+  public String valueWithoutSurroundingQuotes() {
+    return type() == OperatorRightHandType.STRING
+        ? Joiner.on("").join(value().split("\""))
```
bion, line 160: **question:** here as well?

bion (approved):

PR #5182, opened 2023-07-05
Add criteria for picking a new version number

nb1701:
This adds a list of items to look for when decided if a new release should increment the minor or patch version.

### DIFF
--- a/.github/ISSUE_TEMPLATE/oncall-rotation.md
+++ b/.github/ISSUE_TEMPLATE/oncall-rotation.md
@@ -21,6 +21,15 @@ Resources:
 
 ## Release
 
+- [ ] Choose a new version number in the format `vX.Y.Z`. If any of the following apply to the release, bump the minor version (Y). Otherwise, bump the patch version (Z). Do not bump the major version (X) without discussing with #engineering, as this will need more extensive communication and planning.
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/oncall-rotation.md
```diff
@@ -21,6 +21,15 @@ Resources:
 
 ## Release
 
+- [ ] Choose a new version number in the format `vX.Y.Z`. If any of the following apply to the release, bump the minor version (Y). Otherwise, bump the patch version (Z). Do not bump the major version (X) without discussing with #engineering, as this will need more extensive communication and planning.
+
+  - A brand new feature (feature flagged or not)
```
bion, line 26: **nitpick:** `(feature flagged or not)` maybe clarify that the minor version should be bumped when the feature is generally available as opposed to released behind a flag?

bion (approved):

.github/ISSUE_TEMPLATE/oncall-rotation.md
bion: > I suppose it is (in theory anyway) completely isolated and not user-visible when disabled when it's behind the feature flag

This is just as often not the reality though 😆. Maybe we should bump it both when the feature is introduced behind the flag and when the flag is removed? Minor versions are cheap.

Whatever we do just want it to be explicit here.

Issue #5139, opened 2023-06-27
Python version mismatch error when running deploy system


bion:
A civic enttity deployer encountered the following messages when deploying:

```
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of requests to determine which version is compatible with other requirements. This could take a while.
ERROR: Package 'env-var-docs' requires a different Python: 3.9.16 not in '>=3.10'
WARNING: You are using pip version 21.2.3; however, version 23.1.2 is available.
You should consider upgrading via the '/civiform/civiform-deploy/checkout/.venv/bin/python3 -m pip install --upgrade pip' command.
```

Followed by `ERROR: Package 'env-var-docs' requires a different Python: 3.9.16 not in '>=3.10'` and the deployment command failing.

This is likely due to the version of Python available on the deployer's machine. If that's the case, a couple options:

A. generate whatever output is needed from env-var-docs and store it somewhere so that the deploy code doesn't need to run it and therefore doesn't need Python >= 3.10
B. dockerize the deployment commands so the deployer only needs docker installed and doesn't need to worry about other dependencies (this work is partially completed)
C. Require deployers to install Python >=3.10

bion:
@nb1701 IIRC the parser uses some features from >=3.10 and it'd be fairly annoying to downgrade

PR #5202, opened 2023-07-07
Use older Union syntax to relax Python version requirement

nb1701:
The only reason we needed Python 3.10+ was the usage of the | syntax for type unions. This replaces it with the older Union[] style, so that we can relax the version requirement. Some users are on OSes that don't include the latest versions of Python in their system repos.

#### General

- [ ] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release | database >

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/5139


bion:
@nb1701 IIRC we need >=3.10 to run [mypy](https://mypy-lang.org/)

Issue #4908, opened 2023-05-19
Client emails should not be required for TI Flow


ldinhcos:
**Is your feature request related to a problem? Please describe.**
Navigators mentioned clients do not typically use email and that this requirement hinders the experience. Navigators mentioned they monitor statuses through their own email and do not always need to include clients’ addresses.

**Describe the solution you'd like**
Change the requirements of emails and make it optional. 

**Additional context**

**Done when**
What conditions need to be met before this issue can be closed?

elisekalstad:
![Image](https://github.com/civiform/civiform/assets/86082659/2637773f-233c-42bb-951a-d69f69c741f6)

elisekalstad:
Needs technical review of level of effort for removing email as the identifier in database. @swatkat1 -- let's chat about taking the initial review. And partnering with @nb1701 as needed.

elisekalstad:
@nb1701 --I don't want to change it directly in the file, but here is my suggestion. @ldinhcos --What do you think?

  "Add an email address for your client to receive status updates about their application"
                    + " automatically. Without an email, you or your community-based organization will be"
                    + " responsible for communicating updates to your client.")

Rationale: TIs don't actually call themselves TIs, rather they are community-based orgs. Include more direct language.

ldinhcos:
Thanks Nick & Elise.

I saw the original copy and I like Elise's updates in it as it calls out the acronym of CBOs clearly instead of the verbiage of TI's as it may confuse people and also appreciate that it appoints them as responsible for sharing status updates.
________________________________

PR #5215, opened 2023-07-11
Allow Trusted Intermediary clients not to have an email address

nb1701:
### Description

This makes the email address field optional when adding a new client in the TI dashboard. Because we already support guest accounts without an email address, it works to simply remove the validation enforcing a value for the email address.

This also shows the DOB as required, since it always has been, and needed the red * by it.

Additionally, this changes the input type of the email field to an email type. This adds some validation to ensure that if a value is entered, it looks like an actual email address.

Finally, this modifies the TI unit test slightly to clean up state between tests. Previously, you'd be unable to run this test multiple times as the accounts created by the tests would persist and you'd get failures due to trying to add more accounts with duplicate email addresses.

Also contains a commit to update .gitignore with auto-generated files in the .settings directory.

#### General

Add new clients as a TI, both with and without email addresses and ensure all the flows work correctly.

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/4908

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ logs/
 !/browser-test/sbt_cache/.keep
 sbt_cache/*
 !sbt_cache/.keep
+server/.settings/org.eclipse.*
 target/
 test-support/sbt_cache/
 


swatkat1:
This is really cool!!

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -67,14 +66,6 @@ public Form<AddApplicantToTrustedIntermediaryGroupForm> addNewClient(
     return form;
   }
 
-  private Form<AddApplicantToTrustedIntermediaryGroupForm> validateEmailAddress(
-      Form<AddApplicantToTrustedIntermediaryGroupForm> form) {
```
swatkat1, line 71: I don't think we should get rid of it. If the email address is going to be optional, we still want to make sure it is unique.

swatkat1:
Can you please add a Browser test to create a Client without email and check if it's getting created successfully?

browser-test/src/support/ti_dashboard.ts
```diff
@@ -39,7 +39,7 @@ export class TIDashboard {
 
   async expectDashboardContainClient(client: ClientInformation) {
     const row = this.page.locator(
-      `.cf-admin-question-table-row:has-text("${client.emailAddress}")`,
```
swatkat1, line 42: Instead of deleting this line, we can instead pass a flag to check if the email address needs to be checked or not as it tests the function in other clients who have provided email address.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -57,6 +57,42 @@ describe('Trusted intermediaries', () => {
...
+  })
+
+  it('expect client cannot be added with invalid email address', async () => {
+    const {page, tiDashboard} = ctx
+    await loginAsTrustedIntermediary(page)
+
+    await tiDashboard.gotoTIDashboardPage(page)
+    await waitForPageJsLoad(page)
+    const client: ClientInformation = {
+      emailAddress: 'bademail',
```
swatkat1, line 85: Here also, can you leave it blank (two clients with blank email)

swatkat1 (approved):

Issue #4975, opened 2023-05-26
Add answers to the enumerator question itself to the API


avaleske:
The API response includes the answers for repeated questions, but it doesn't include the enumerator question itself.
https://github.com/civiform/civiform/blob/bed9f40095830e158825b8dad73d846c97cba270/server/app/services/export/JsonExporter.java#L122-L128

For example, the responses to this question are not included in the API:
<img width="589" alt="image" src="https://github.com/civiform/civiform/assets/1295824/48e86b1a-fc3b-4d44-94ef-fe77d489b60f">

But the answers to the repeated questions asked about each household member on the following screens _are_ included.

More in the original API TDD:
https://docs.google.com/document/d/1yVMv-FA_UKrftsstSMlL5RW_4VyIn4lwqc0I57zAL6c/edit#bookmark=id.da3r0qcbyhwm

If an enumerator question has no answers, so there are no repeated entities, it should be included as an empty array, like 
```js
"application": {
  "applicant_household_members": []
}
```

Edit by @michaelzetune: to add this we'll have to update `QuestionJsonPresenter` for the API. Once done, we'll also eventually want to update `QuestionJsonSampler`, which powers the API docs - see #5235.

PR #5235, opened 2023-07-13
Create `QuestionJsonSampler`

michaelzetune:
### Description

This PR creates `QuestionJsonSampler`, which is the precursor to API generated docs.

The goal of the class is simple: for any given `Question`, seed some sample data then use the corresponding `QuestionJsonPresenter` to show JSON for it. In a future PR, API generated docs will use this class to present the API Response Preview.

Note that the following don't return sample JSON:
* `STATIC` questions: these don't have responses from a user.
* `ENUMERATOR` questions: these are not supported by the API, so they shouldn't be sampled. See #4975.
* Enumerat**ed** questions: (any question where the `enumeratorId` is set): this turns out to be *very* complicated, so I will punt it to a future PR.

## Release notes

None.

#### General

#### User visible changes

None.

#### New Features

None yet.

None.

### Issue(s) this completes

Relates to #4872, #5238.

### DIFF
--- a/server/app/services/export/QuestionJsonSampler.java
+++ b/server/app/services/export/QuestionJsonSampler.java
@@ -38,6 +38,11 @@ import services.question.types.QuestionType;
 public interface QuestionJsonSampler<Q extends Question> {
 
   default CfJsonDocumentContext getSampleJson(QuestionDefinition questionDefinition) {
+    if (questionDefinition.getEnumeratorId().isPresent()) {
+      // TODO(#5238): support enumerated questions.
+      return new CfJsonDocumentContext();
+    }
+
     ProgramQuestionDefinition programQuestionDefinition =
         ProgramQuestionDefinition.create(questionDefinition, Optional.empty());
     ApplicantData applicantData = new ApplicantData();
--- a/server/app/services/export/QuestionJsonSampler.java
+++ b/server/app/services/export/QuestionJsonSampler.java
@@ -1,7 +1,7 @@
 package services.export;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static services.export.JsonExporter.exportToJsonApplication;
+import static services.export.JsonExporter.exportEntriesToJsonApplication;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -56,7 +56,7 @@ public interface QuestionJsonSampler<Q extends Question> {
     ImmutableMap<Path, ?> entries = getJsonPresenter().getJsonEntries(question);
     CfJsonDocumentContext jsonApplication = new CfJsonDocumentContext();
 
-    entries.entrySet().forEach(entry -> exportToJsonApplication(jsonApplication, entry));
+    exportEntriesToJsonApplication(jsonApplication, entries);
 
     return jsonApplication;
   }
@@ -207,7 +207,7 @@ public interface QuestionJsonSampler<Q extends Question> {
       ImmutableMap<Path, ?> entries = getJsonPresenter().getJsonEntries(question);
       CfJsonDocumentContext jsonApplication = new CfJsonDocumentContext();
 
-      entries.entrySet().forEach(entry -> exportToJsonApplication(jsonApplication, entry));
+      exportEntriesToJsonApplication(jsonApplication, entries);
 
       return jsonApplication;
     }
--- a/server/app/services/export/QuestionJsonSampler.java
+++ b/server/app/services/export/QuestionJsonSampler.java
@@ -1,0 +1,521 @@
+package services.export;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static services.export.JsonExporter.exportToJsonApplication;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.Optional;
+import javax.inject.Inject;
+import services.CfJsonDocumentContext;
+import services.Path;
+import services.applicant.ApplicantData;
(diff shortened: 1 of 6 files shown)


server/test/services/export/QuestionJsonSamplerTest.java
```diff
@@ -0,0 +1,297 @@
...
+public class QuestionJsonSamplerTest extends ResetPostgres {
+
+  private QuestionJsonSampler.Factory questionJsonSamplerFactory;
+
+  @Before
+  public void setUp() {
+    questionJsonSamplerFactory = instanceOf(QuestionJsonSampler.Factory.class);
+  }
+
+  @Test
```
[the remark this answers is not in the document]
avaleske: Yeah I can see why you wanted Java 20

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,525 @@
...
+    ProgramQuestionDefinition programQuestionDefinition =
+        ProgramQuestionDefinition.create(questionDefinition, Optional.empty());
+    ApplicantData applicantData = new ApplicantData();
+    ApplicantQuestion applicantQuestion =
+        new ApplicantQuestion(programQuestionDefinition, applicantData, Optional.empty());
+
+    Q question = getQuestion(applicantQuestion);
+    addSampleData(applicantData, question);
+
+    @SuppressWarnings("unchecked")
```
avaleske, line 55: Can we add a comment explaining why we're suppressing the warning?

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,525 @@
...
+      ProgramQuestionDefinition programQuestionDefinition =
+          ProgramQuestionDefinition.create(questionDefinition, Optional.empty());
+      ApplicantData applicantData = new ApplicantData();
+      ApplicantQuestion applicantQuestion =
+          new ApplicantQuestion(programQuestionDefinition, applicantData, Optional.empty());
+
+      Q question = getQuestion(applicantQuestion);
+      addSampleData(applicantData, question);
+
+      @SuppressWarnings("unchecked")
```
avaleske, line 206: ditto here

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+    @Override
+    public EmailQuestion getQuestion(ApplicantQuestion applicantQuestion) {
+      return applicantQuestion.createEmailQuestion();
+    }
+
+    @Override
+    public void addSampleData(ApplicantData applicantData, EmailQuestion question) {
+      applicantData.putString(question.getEmailPath(), "homer.simpson@springfield.gov");
```
avaleske, line 280: lol

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,525 @@
...
+  class EmptyJsonSampler implements QuestionJsonSampler<Question> {
+
+    @Override
+    public CfJsonDocumentContext getSampleJson(QuestionDefinition questionDefinition) {
+      return new CfJsonDocumentContext();
+    }
+
+    @Override
+    public Question getQuestion(ApplicantQuestion applicantQuestion) {
+      return applicantQuestion.createCurrencyQuestion();
```
avaleske, line 298: Should this be the currency question?

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+    @Override
+    public void addSampleData(ApplicantData applicantData, Question question) {
+      // no-op
+    }
+
+    @Override
+    public QuestionJsonPresenter getJsonPresenter() {
+      return null;
```
avaleske, line 308: Should this return the EmptyJsonPresenter?

avaleske (approved):
I like that we're using the same code to generate the sample response as we do for the actual response!

server/app/services/export/QuestionJsonSampler.java
michaelzetune: Done.

server/app/services/export/QuestionJsonSampler.java
michaelzetune: Not quite, we don't want to couple "empty" with static content question, since other question types (currently, enumerator too) should be able to use it.

But returning currency question was a mistake, updated it to return null. Note this is not a problem, since the `getSampleJson` method uses `getQuestion` to pass to it to `addSampleData`, whose implementation is a no-op (below) for `EmptyJsonSampler`.

server/app/services/export/QuestionJsonSampler.java
michaelzetune: There isn't an `EmptyJsonPresenter`, and no need for one. Since we override the default `getSampleJson` in `EmptyJsonSampler`, it always returns an empty `CfJsonDocumentContext` and we don't need to worry about it.

server/app/services/export/QuestionJsonSampler.java
```diff
@@ -0,0 +1,519 @@
...
+      applicantData.putString(question.getWellKnownIdPath(), "23214");
+      applicantData.putString(question.getServiceAreaPath(), "springfield_county");
+    }
+
+    @Override
+    public QuestionJsonPresenter getJsonPresenter() {
+      return addressJsonPresenter;
+    }
+  }
+
```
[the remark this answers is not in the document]
michaelzetune: This ended up not being needed; removed.

PR #5261, opened 2023-07-19
Removing logger.error for phone question formatting errors

### LINKED ISSUE #5258: PhoneQuestion Logging Erroneous Errors
### Errors being generated on the review page
The application review page is causing errors to be logged by the Phone Question type.

This is the [error being logged](https://github.com/civiform/civiform/blob/8b273dc0f45e8be2365402cadd1d5db4abf967a8/server/app/services/applicant/question/PhoneQuestion.java#L126C1-L130).

You don't have to fill out the application, just load the initial starting point.

###  Don't log application data
Additionally we should not include getPhoneNumberValue in the logging message in order to keep applicant data from ending up in the logs.

### Log Level
The `getAnswerString` method looks to rely on the try/catch to determin if the value is valid. If it's not valid it throws, we eat the exception, and use a placeholder string. I don't think this probably needs to be logging at all then. If we keep the logging it should be lowered to warning at the highest and not error. Trace or info may be more appropriate. Since it's being handled having it as error make reviewing the logs more noisy for unhandled errors.

## To Reproduce

- As Admin
    - Create a new question of phone type
    - Create a new program
    - Add phone type question to program
   - Publish
   - Logout
- As Applicant
    - Apply to program as quest
    - Errors logged here on the program review page

**Expected behavior**
- Don't log applicant a data
- Lower the log level
- Don't generate errors on the program review page.

swatkat1:
### Description

Updating logger.error to logger.warn for phone formatting errors

#### General

### Issue(s) this completes

Fixes #5258

### DIFF
--- a/server/app/services/applicant/question/PhoneQuestion.java
+++ b/server/app/services/applicant/question/PhoneQuestion.java
@@ -111,7 +111,10 @@ public final class PhoneQuestion extends Question {
   public Path getCountryCodePath() {
     return applicantQuestion.getContextualizedPath().join(Scalar.COUNTRY_CODE);
   }
-
+  /**
+   * Returns formatted phone number or defaults to "-" if formatting fails. Format example:+1
+   * 206-648-2489
+   */
   @Override
   public String getAnswerString() {
     try {
--- a/server/app/services/applicant/question/PhoneQuestion.java
+++ b/server/app/services/applicant/question/PhoneQuestion.java
@@ -112,8 +112,8 @@ public final class PhoneQuestion extends Question {
     return applicantQuestion.getContextualizedPath().join(Scalar.COUNTRY_CODE);
   }
   /**
-   * Returns formatted phone number or defaults to "-" if formatting fails. Format example:+1
-   * 206-648-2489
+   * Returns formatted phone number or defaults to "-" if formatting fails. An example of formatted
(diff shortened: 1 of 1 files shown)


server/app/services/applicant/question/PhoneQuestion.java
```diff
@@ -123,12 +120,7 @@ public String getAnswerString() {
               getPhoneNumberValue().orElse(""), getCountryCodeValue().orElse(""));
       return PHONE_NUMBER_UTIL.format(phoneNumber, PhoneNumberUtil.PhoneNumberFormat.INTERNATIONAL);
     } catch (NumberParseException e) {
-      logger.error(
-          "Unable to retrieve or parse phone number "
-              + getPhoneNumberValue().orElse("")
-              + "for country_code "
-              + getCountryCodeValue().orElse(""));
+      return "-";
```
gwendolyngoetz, line 123: Blocking: Can you add a javadoc comment that says `Returns formatted phone number or defaults to "-" if formatting fails. Format example: __ADD_EXAMPLE_HERE__` and replace that placeholder with an actual example of the formatted number.

swatkat1: done

gwendolyngoetz (approved):

PR #5270, opened 2023-07-20
Simplify `QuestionDefinitionBuilder`

michaelzetune:
### Description

`QuestionDefinitionBuilder` exists as technical debt, now that `QuestionDefinition` is constructed with a `QuestionDefinitionConfig.Builder`.

It's very hard to remove, since it has so many usages throughout CiviForm. This PR works towards making `QuestionDefinitionBuilder` use `QuestionDefinitionConfig` internally, so that there is at least less risk of divergence in behavior.

Changes:
* Instead of storing all of the `QuestionDefinition` vars in `QuestionDefinitionBuilder`, we now just store a `QuestionDefinitionConfig.Builder`.
* Stop constructing the `QuestionDefinitionConfig` for each question type; reuse the original one in the switch statement.
* In `QuestionDefinition`, use `config` as little as possible (i.e. use `this.getName()` instead of `config.getName()`.

Note: `questionHelpText` was a tricky thing to fix here. In `QuestionDefinitionConfig.Builder`, we had it as a required field, both at compile time and AutoValue's runtime checks. However, it turns out that the many usages of `QuestionDefinitionBuilder` allow `questionHelpText` to be unset, defaulting to `LocalizedStrings.empty()`. In order to accomodate this, we modify the `QuestionDefinitionConfig` AutoValue to allow `questionHelpText` to be unset. Using some clever internal methods, we can avoid the complexity of an `Optional<LocalizedStrings>`, which we don't want since the `LocalizedStrings` itself can be empty anyways.

## Release notes

None.

#### General

   - Refactor, not necessary

### Issue(s) this completes

Progress towards #5271.

### DIFF
--- a/server/app/services/question/types/QuestionDefinition.java
+++ b/server/app/services/question/types/QuestionDefinition.java
@@ -170,7 +170,7 @@ public abstract class QuestionDefinition {
       return config.questionText().locales();
     } else {
       return ImmutableSet.copyOf(
-          Sets.intersection(config.questionText().locales(), config.questionHelpText().locales()));
+          Sets.intersection(config.questionText().locales(), getQuestionHelpText().locales()));
     }
   }
 
@@ -285,7 +285,7 @@ public abstract class QuestionDefinition {
 
   @Override
   public int hashCode() {
-    return Objects.hash(config.id());
+    return Objects.hash(getId());
   }
 
   /** Two QuestionDefinitions are considered equal if all of their properties are the same. */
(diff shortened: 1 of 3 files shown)


server/app/services/question/types/QuestionDefinition.java
```diff
@@ -170,7 +170,7 @@ public ImmutableSet<Locale> getSupportedLocales() {
       return config.questionText().locales();
     } else {
       return ImmutableSet.copyOf(
-          Sets.intersection(config.questionText().locales(), config.questionHelpText().locales()));
```
michaelzetune, line 173: Here and elsewhere in this file, using the internal getters avoids divergence if we change the way we get help text or other fields.

server/app/services/question/types/QuestionDefinitionBuilder.java
```diff
@@ -19,41 +19,42 @@
 import services.question.types.TextQuestionDefinition.TextValidationPredicates;
 
 /**
- * Provides helper functions to build a QuestionDefinition.
+ * DEPRECATED. Provides helper functions to build a {@link QuestionDefinition}.
  *
- * <p>TODO(#4872): Remove this class in favor of QuestionDefinitionConfig.Builder.
+ * <p>TODO(#5271): Remove this class in favor of {@link QuestionDefinitionConfig.Builder}.
  */
 public final class QuestionDefinitionBuilder {
```
michaelzetune, line 27: As evident below, we stop storing the fields that are in `QuestionDefinitionConfig`, instead storing the `QuestionDefinitionConfig` itself to reduce duplication.

server/app/services/question/types/QuestionDefinitionConfig.java
```diff
@@ -49,16 +53,12 @@ public interface RequiredDescription {
   }
```
michaelzetune, line 54: Previously, `RequiredQuestionText` was a clever mechanism to ensure that we set the question help text at compile time. Since `QuestionDefinitionBuilder` has many clients that don't set this, we should remove it as it doesn't match the business logic.

yotommy (approved):
Nice simplifications!

PR #5285, opened 2023-07-24
Bug Fix - Adding tombstoned questions to a program creates invalid state

### LINKED ISSUE #5281: Adding tombstoned questions to a program creates invalid state
1. create two programs, A and B, and one question. Add the question to program A and publish all drafts.
remove the question from program A
1. open screen editing page for program B, open the add question sidebar, see the question
1. in a separate tab, navigate to the questions index and archive the question
1. in the tab from step 3, add the question to program B
1. publish all drafts

Result: [RuntimeException: java.util.concurrent.CompletionException: java.lang.RuntimeException: Question not found for Program 1396]

We should instead throw an exception when attempting to add a question already marked for archival to a program.

Related to: https://github.com/civiform/civiform/issues/5278

swatkat1:
### Description

With two tabs, it is possible to add a tomb-stoned question to an existing program resulting in "Question not found" error. This PR adds another layer of check at ProgramBlockValidation.java to ensure tomb-stoned questions aren't added accidently. 

## Release notes
Bug Fix  for "Adding tombstoned questions to a program creates invalid state"

#### General

### Issue(s) this completes

Fixes #5281


server/app/services/ProgramBlockValidation.java
```diff
@@ -48,26 +57,28 @@ public enum AddQuestionResult {
    * creating to ensure that no one can maliciously mess up data by sending specially crafted
    * request (or in case we mess up client-side checks).
    */
-  public static AddQuestionResult canAddQuestion(
+  public AddQuestionResult canAddQuestion(
       ProgramDefinition program, BlockDefinition block, QuestionDefinition question) {
+    if (versionRepository
+        .getDraftVersion()
+        .getTombstonedQuestionNames()
+        .contains(question.getName())) return AddQuestionResult.QUESTION_TOMBSTONED;
```
bion, line 65: **nitpick:** https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used
```suggestion
        .contains(question.getName())) {
        return AddQuestionResult.QUESTION_TOMBSTONED;
      }
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -48,26 +57,28 @@ public enum AddQuestionResult {
    * creating to ensure that no one can maliciously mess up data by sending specially crafted
    * request (or in case we mess up client-side checks).
    */
-  public static AddQuestionResult canAddQuestion(
+  public AddQuestionResult canAddQuestion(
       ProgramDefinition program, BlockDefinition block, QuestionDefinition question) {
+    if (versionRepository
+        .getDraftVersion()
```
bion, line 63: **issue (blocking):** this is going to query the database for the current draft version each time the method is called, which we do once per question in `ProgramQuestionBank`.

The least disruptive way I can think of to address this is to inject a `ProgramBlockValidationFactory` in the places that currently consume `ProgramBlockValidation`. The factory then creates an instance of `ProgramBlockValidation` which can be created and reused locally to prevent N+1 database queries i.e. the instance can be retrieved before the call to `return params.questions().stream()` and reused inside the call to `filter()`

bion (changes requested):

server/app/services/ProgramBlockValidation.java
swatkat1: I think I understand what your saying - instead of creating an instance of VersionRepository for each call, creating an instance of ProgramBlockValidation and reusing the same object will be prevent excess DB calls. Am I getting it rt?

swatkat1: Also, should the Unit test be updated?

server/app/services/ProgramBlockValidationFactory.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @Inject
+  public ProgramBlockValidationFactory(ProgramBlockValidation programBlockValidation) {
+    this.programBlockValidation = checkNotNull(programBlockValidation);
+  }
+
+  public AddQuestionResult canAddQuestion(
+      ProgramDefinition program, BlockDefinition block, QuestionDefinition question) {
+    return programBlockValidation.canAddQuestion(program, block, question);
+  }
```
bion, line 23: **suggestion (blocking):** This is still the same behavior as before, just with another class in between. The goal is to avoid calling database query methods in a loop and instead call them only once. In this case, the database query method we want to avoid calling repeatedly is inside `versionRepository.getDraftVersion()`.

Factory classes should have a single method that creates a new instance of the class they're a factory for. In this case, `ProgramBlockValidationFactory` should have a `create` method that returns a new instance of `ProgramBlockValidation`:
```suggestion
  @Inject
  public ProgramBlockValidationFactory(VersionRepository versionRepository) {
    this.versionRepository = checkNotNull(versionRepository);
  }

  public ProgramBlockValidation create() {
    Version currentDraft = versionRepository.getDraftVersion();
    return new ProgramBlockValidation(currentDraft);
  }
```

Then, in `ProgramQuestionBank.java`:
```java
  private Stream<QuestionDefinition> filterQuestions() {
    ProgramBlockValidation programBlockValidator = programBlockValidationFactory.create();
    return params.questions().stream()
        .filter(
            q ->
                programBlockValidator.canAddQuestion(params.program(), params.blockDefinition(), q)
                    == ProgramBlockValidation.AddQuestionResult.ELIGIBLE);
                programBlockValidator.canAddQuestion(
                        params.program(), params.blockDefinition(), q)
                    == AddQuestionResult.ELIGIBLE);
  }
  ```

**Review by bion — CHANGES_REQUESTED** [review]

File: server/app/services/ProgramBlockValidationFactory.java (hunk 3 above)
Comment (swatkat1) [reply server/app/services/ProgramBlockValidationFactory.java:23]: This make sense. Thank you for clarifying

File: server/app/services/ProgramBlockValidation.java [hunk 4]
```diff
@@ -10,8 +11,11 @@
 /** Helper class for performing validation related to creating or modifying program blocks. */
 public final class ProgramBlockValidation {
 
-  private ProgramBlockValidation() {}
+  private final Version version;
 
+  public ProgramBlockValidation(Version version) {
+    this.version = version;
```
bion, line 17: **nitpick:** 
```suggestion
    this.version = checkNotNull(version);
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -48,26 +53,31 @@ public enum AddQuestionResult {
    * creating to ensure that no one can maliciously mess up data by sending specially crafted
    * request (or in case we mess up client-side checks).
    */
-  public static AddQuestionResult canAddQuestion(
+  public AddQuestionResult canAddQuestion(
       ProgramDefinition program, BlockDefinition block, QuestionDefinition question) {
+    System.out.println("question" + question.getName());
+    for (String s : version.getTombstonedQuestionNames()) {
+      System.out.println("SSSS" + s);
+    }
```
bion, line 61: **nitpick:** remove print statements

server/app/services/ProgramBlockValidationFactory.java
```diff
@@ -0,0 +1,22 @@
+package services;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import javax.inject.Inject;
+import models.Version;
+import repository.VersionRepository;
+
+public final class ProgramBlockValidationFactory {
```
bion, line 9: **suggestion (blocking):** add javadoc for this public class

server/app/services/ProgramBlockValidation.java
```diff
@@ -48,26 +53,31 @@ public enum AddQuestionResult {
    * creating to ensure that no one can maliciously mess up data by sending specially crafted
    * request (or in case we mess up client-side checks).
    */
-  public static AddQuestionResult canAddQuestion(
+  public AddQuestionResult canAddQuestion(
```
bion, line 56: **suggestion (blocking):** update the javadoc for this method

server/test/services/ProgramBlockValidationFactoryTest.java
```diff
@@ -0,0 +1,52 @@
...
+            .withBlock()
+            .withRequiredQuestionDefinition(questionForTombstone.getQuestionDefinition())
+            .buildDefinition();
+    assertThat(
+            programBlockValidation.canAddQuestion(
+                program,
+                program.getLastBlockDefinition(),
+                questionForTombstone.getQuestionDefinition()))
+        .isEqualTo(services.ProgramBlockValidation.AddQuestionResult.QUESTION_TOMBSTONED);
+  }
```
bion, line 51: **suggestion (blocking):** move this test into server/test/services/ProgramBlockValidationTest.java since it is focused on asserting the behavior of the `ProgramBlockValidation` class.

I don't think the factory class needs a unit test

bion (changes requested):

server/test/services/ProgramBlockValidationFactoryTest.java
swatkat1: oh ok..

server/app/services/ProgramBlockValidation.java
```diff
@@ -51,14 +53,11 @@ public enum AddQuestionResult {
    * <p>This should be checked both during rendering to make sure admins don't see ineligible
    * questions when creating blocks. Also it should be checked during the actual server-side block
    * creating to ensure that no one can maliciously mess up data by sending specially crafted
-   * request (or in case we mess up client-side checks).
+   * request (or in case we mess up client-side checks). It also ensures that a question tombstoned
+   * in another browser tab, doesn't get accidently added to a program.
```
bion, line 57: **suggestion:** 
```suggestion
   * request (or in case we mess up client-side checks). It also ensures that a question is not tombstoned
   * (marked for deletion) in the current draft.
```

server/app/services/ProgramBlockValidationFactory.java
```diff
@@ -6,6 +6,12 @@
 import models.Version;
 import repository.VersionRepository;
 
+/**
+ * Helper class for creating a ProgramBlockValidation Object.The reason this factory class is needed
+ * because - for every question in the question picker, we will need to check if the question is
+ * tombstoned, resulting in n+1 DB queries. To avoid it, we create a factory object injecting a
+ * VersionRepository and calling getDraftVersion() only once during the object creation.
```
bion, line 13: **nitpick:** 
```suggestion
 * Factory for creating `ProgramBlockValidation` instances. 
 * <p>Needed because - for every question in the question picker, we will need to check if the question is
 * tombstoned, resulting in n+1 DB queries. To avoid this, we create a factory object injecting a
 * VersionRepository and calling getDraftVersion() only once per request.
```

server/app/services/ProgramBlockValidation.java
```diff
@@ -46,28 +53,30 @@ public enum AddQuestionResult {
...
    */
-  public static AddQuestionResult canAddQuestion(
+  public AddQuestionResult canAddQuestion(
       ProgramDefinition program, BlockDefinition block, QuestionDefinition question) {
+    if (version.getTombstonedQuestionNames().contains(question.getName())) {
+      return AddQuestionResult.QUESTION_TOMBSTONED;
+    }
```
bion, line 63: **suggestion (non-blocking):** in this or a follow up PR, it'd also be good to ensure that the question IS in either the current draft version or current active version.

bion (changes requested):

server/app/services/ProgramBlockValidation.java
swatkat1: I am new to this side of the code. for your comment on "in this or a follow up PR, it'd also be good to ensure that the question IS in either the current draft version or current active version." --> Is this information available in the version object or should I be looking else where? I see this class "ActiveAndDraftQuestions"has a list of all active and draft questions

bion: It isn't available on `Version`, so you'll likely need to inject something else and pass it in through the factory, fine to do in a separate PR or I can do it if you'd rather

bion (approved):

Issue #5282, opened 2023-07-24
Ensure updating a question never creates duplicate drafts


bion:
Follow up on https://github.com/civiform/civiform/issues/5209

PR #5286, opened 2023-07-24
Use canonical identifiers for question and program translations

bion:
### Description

UI for editing translations use specific question/program (QP) IDs, so that when the form is rendered and submitted it is with regard to a specific version of a QP. This could potentially cause data inconsistency issues if a user has browser pages open to stale versions of the question since the code to update a translation calls [QuestionService#update](https://github.com/civiform/civiform/blob/use-canonical-references-for-programs-and-questions/server/app/services/question/QuestionService.java) which calls [createOrUpdateDraft](https://github.com/civiform/civiform/blob/use-canonical-references-for-programs-and-questions/server/app/services/question/QuestionService.java#L147) -- since the draft has a new ID, multiple drafts could be created this way.

This PR changes all IDs in URLs in the questions and programs translation UI to use the QP name, which is immutable across multiple versions, to ensure that the regardless of stale browser state, the server loads the current draft for the question or program, or throws a `BadRequestException` if no draft is available.

A possible follow up to this PR would be to create a draft is one does not exist yet for a better admin experience, but it should a rare occurrence  (only possible with stale browser state) given that the button to access the translation UI is only available after creating a new draft of the QP.

#### General

### Issue(s) this completes

Contributes to https://github.com/civiform/civiform/issues/5282

### DIFF
--- a/server/app/controllers/admin/AdminQuestionTranslationsController.java
+++ b/server/app/controllers/admin/AdminQuestionTranslationsController.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import auth.Authorizers;
 import auth.ProfileUtils;
+import controllers.BadRequestException;
 import controllers.CiviFormController;
 import forms.translation.EnumeratorQuestionTranslationForm;
 import forms.translation.MultiOptionQuestionTranslationForm;
@@ -23,8 +24,8 @@ import services.CiviFormError;
 import services.ErrorAnd;
 import services.TranslationLocales;
 import services.question.QuestionService;
+import services.question.ReadOnlyQuestionService;
 import services.question.exceptions.InvalidUpdateException;
-import services.question.exceptions.QuestionNotFoundException;
 import services.question.exceptions.UnsupportedQuestionTypeException;
 import services.question.types.QuestionDefinition;
 import services.question.types.QuestionType;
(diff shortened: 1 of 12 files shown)


server/app/controllers/admin/AdminQuestionTranslationsController.java
```diff
@@ -120,49 +103,49 @@ public CompletionStage<Result> edit(Http.Request request, long questionId, Strin
    *     same {@link QuestionTranslationView} with error messages
    */
   @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
-  public CompletionStage<Result> update(Http.Request request, long questionId, String locale) {
+  public Result update(Http.Request request, String questionName, String locale) {
```
bion, line 106: 💁‍♂️ made this synchronous to make the code easier to read since it's admin-facing

server/app/views/admin/TranslationFormView.java
```diff
@@ -53,7 +53,7 @@ protected final DivTag renderLanguageLinks(long entityId, Locale currentlySelect
    * Given the ID of the entity to translate and a locale for translation, returns a link
    * destination URL for the edit form to translate the entity in the given locale.
    */
-  protected abstract String languageLinkDestination(long entityId, Locale locale);
+  protected abstract String languageLinkDestination(String entityId, Locale locale);
```
gwendolyngoetz, line 56: (blocking) Can you update entityId to be entityName along with the javadoc

server/app/controllers/admin/AdminQuestionTranslationsController.java
```diff
@@ -87,28 +82,16 @@ public Result redirectToFirstLocale(Http.Request request, long questionId) {
...
-        .getReadOnlyQuestionService()
-        .thenApplyAsync(
-            readOnlyQuestionService -> {
-              try {
-                QuestionDefinition definition =
-                    readOnlyQuestionService.getQuestionDefinition(questionId);
-                return ok(translationView.render(request, localeToEdit, definition));
-              } catch (QuestionNotFoundException e) {
-                return notFound(e.getMessage());
-              }
```
gwendolyngoetz, line 109: (praise) Nice simplification here! And thank you for removing the notFound returning a  simple string!!!

gwendolyngoetz (changes requested):

gwendolyngoetz:
Are there any browser tests that need updating?

server/app/controllers/admin/AdminQuestionTranslationsController.java
bion: 😄

server/app/views/admin/TranslationFormView.java
bion: Done, good call

bion:
> Are there any browser tests that need updating?

I don't think so -- there's browser test coverage of editing translations, so this is covered.

gwendolyngoetz (approved):
LGTM

Issue #5262, opened 2023-07-19
API Generated Docs


michaelzetune:
API Generated Docs intend to clearly document what API consumers can expect the response for a given program to look like for the `/applications` endpoint.

[Design document](https://docs.google.com/document/d/1mv6YTk0-Pt9ygLsZHEHu2RXqkzO3sIyAJ7VCl6Vz-v0/edit?resourcekey=0-GnlC2ufUkH7QtM8i1O_p1A#heading=h.xf9qz34jakum)

Sub-issues:
* #4872
* #4945
* #4946

#5253 is the initial PR. It adds API docs gated under the flag `api_generated_docs_enabled`.

TODO after that PR:
- [x] Consider implementing Sijia's frontend suggestions ([mock](https://github.com/civiform/civiform/assets/30369272/f875d0f0-334d-429f-ab11-6834dd3535dd) of some of these)
   - Can wrap columns if need more than 3 columns
   - Can remove "Question" from "Question text" etc
   - For lower underscore names, we can keep the font consistent, don't need to use monospace
   - Bigger size and bold for name of question
   - Add accordion for the help text at the top of the API docs page
   - Implemented in #5284
- [x] Add other fields outside of `application` to the API Response Preview
   - Implemented in #5294 and #5297
- [x] Support draft programs (#4945)
   - Implemented in #5310
- [ ] Add copy near selectable options (maybe a tooltip) that makes it clear that the names of options may change/have changed, without changing users' selections. That is, the text of their selection may not match what they saw when they applied, only the ID of the selection is consistent. (#5350)
- [ ] Add "Published time" and "Removed time" field to each question to make it clear when that question was published or removed from each program (so API consumers understand that it won't be on applications earlier/later than those time). ([Design doc link](https://docs.google.com/document/d/1mv6YTk0-Pt9ygLsZHEHu2RXqkzO3sIyAJ7VCl6Vz-v0/edit?resourcekey=0-GnlC2ufUkH7QtM8i1O_p1A#bookmark=id.5eu7swcmtuym)) (#5351)
- [ ] Support enumerated questions (already in API, need to add to API docs) (#5238)
- [ ] Support enumerator questions (not in API nor API docs) (#4975)
- [ ] Support opening the API docs link in a new browser (i.e. incognito mode). Currently, opening the CiviForm home page in incognito, unauthenticated, followed by opening the API docs page (without logging in as an admin) **works**, but opening the API docs page in incognito mode directly does not work. (#5353)
[...]

PR #5287, opened 2023-07-24
Use `QuestionAnswerer` in `QuestionJsonSampler`

michaelzetune:
### Description

Greatly simplify `QuestionJsonSampler` by using `QuestionAnswerer` to answer questions.

Details:
- Move `QuestionAnswerer` from `support` to `services.question` directory (originally it was only for tests)
- Call `QuestionAnswerer` in the `addSampleData` method of each `QuestionJsonSampler`.

## Release notes

None.

#### General

   - Existing `api_docs.test.ts` and `QuestionJsonSamplerTest` ensure the change is consistent with previous behavior.

#### User visible changes

None.

#### New Features

None.

None.

### Issue(s) this completes

Relates to #5262

### DIFF
--- a/server/app/controllers/api/ApiDocsController.java
+++ b/server/app/controllers/api/ApiDocsController.java
@@ -48,10 +48,10 @@ public final class ApiDocsController {
 
     ImmutableSet<String> allProgramSlugs = programService.getAllProgramSlugs();
 
+    ProgramDefinition programDefinition;
     try {
-      ProgramDefinition programDefinition =
+      programDefinition =
           programService.getActiveProgramDefinitionAsync(programSlug).toCompletableFuture().join();
-      return ok(docsView.render(request, programDefinition, allProgramSlugs));
     } catch (Exception e) {
       return notFound(
           String.format(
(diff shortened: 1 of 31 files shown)


server/app/controllers/api/ApiDocsController.java
```diff
@@ -48,16 +48,18 @@ public Result docsForSlug(Http.Request request, String programSlug) {
...
     } catch (Exception e) {
       return notFound(
           String.format(
               "No active programs found for %s. Please create and publish a program with this slug"
                   + " to continue.",
               programSlug));
     }
+
+    return ok(docsView.render(request, programDefinition, allProgramSlugs));
   }
```
michaelzetune, line 64: Earlier I ran into an exception in the page rendering that said "No active programs found..." so I moved the `return ok(...)` outside the try so that the exception will be more clear if it were to happen again.

avaleske (approved):
LGTM but check with @bion to make sure there's not some reason other than inertia that we weren't already using the QuestionAnswerer in prod code.

bion:
> LGTM but check with @bion to make sure there's not some reason other than inertia that we weren't already using the QuestionAnswerer in prod code.

Inertia mostly. One concern is that it bypasses the normal system for writing an applicant's answers so we lose validations and metadata updates. That's not a concern in this case since we're not writing real applicant data.

bion (approved):

PR #5292, opened 2023-07-25
Detect applications that are duplicates (modulo timestamps).

yotommy:
### Description

Detect applications that are duplicates (modulo timestamps).

## Release notes

This PR adds code to detect that a draft application being submitted contains identical answers to the currently-active application for this applicant and program. The scope of this PR is to log the condition; future PRs may affect the UX.

Though it would be cleanest to compare answers via their Java representations, these are not available at this point in the flow; only the JSON representation is available. Thus we compare the JSON representations of the applicant data, after clearing some timestamp metadata.

See [Design: Detecting Unintentional Duplicate Applications](https://docs.google.com/document/d/1nUGatPTxcL0bCuGY3e2hTajPAYufQJpXXnbpgXi4Xs4/edit?usp=sharing).

#### General

As a logged-in user, edit an existing application. Do not change any answers before re-submitting.

As a guest user, complete an application. Then use the browser back button and re-submit.

In either case, check the server logs for `warn`-level messages that contain the string `"duplicate = true"`.

### DIFF
--- a/server/app/services/applicant/ApplicantData.java
+++ b/server/app/services/applicant/ApplicantData.java
@@ -173,6 +173,7 @@ public class ApplicantData extends CfJsonDocumentContext {
     // The `updated_at` timestamp for an answer should not be considered when
     // comparing answers.
     try {
+      // The ".." in the path scans the entire document.
       applicantData.getDocumentContext().set("$..updated_at", 0);
     } catch (PathNotFoundException unused) {
       // Metadata may be missing in unit tests. No harm, no foul.
--- a/server/app/services/applicant/ApplicantData.java
+++ b/server/app/services/applicant/ApplicantData.java
@@ -156,7 +156,7 @@ public class ApplicantData extends CfJsonDocumentContext {
 
   /**
    * Returns `true` if the answers in `other` match the answers in the current object. Ignores
-   * certain metadata (such as timestamps) when comparing answers.
+   * `updated_at` timestamps when comparing answers.
    */
   public boolean isDuplicateOf(ApplicantData other) {
     // Copy data and clear fields not required for comparison.
--- a/server/app/services/applicant/ApplicantData.java
+++ b/server/app/services/applicant/ApplicantData.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableMap;
+import com.jayway.jsonpath.PathNotFoundException;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


server/app/services/applicant/ApplicantData.java
```diff
@@ -152,4 +153,29 @@ public Optional<LocalDate> getDateOfBirth() {
...
+
+    return thisApplicantData.asJsonString().equals(otherApplicantData.asJsonString());
+  }
+
+  private static void clearFieldsNotRequiredForComparison(ApplicantData applicantData) {
+    // The `updated_at` timestamp for an answer should not be considered when
+    // comparing answers.
+    try {
+      applicantData.getDocumentContext().set("$..updated_at", 0);
+    } catch (PathNotFoundException unused) {
```
yotommy, line 177: There's an argument for being more defensive here and catching everything. Advice welcome.

server/app/services/applicant/ApplicantData.java
```diff
@@ -152,4 +153,29 @@ public Optional<LocalDate> getDateOfBirth() {
...
+    clearFieldsNotRequiredForComparison(otherApplicantData);
+
+    return thisApplicantData.asJsonString().equals(otherApplicantData.asJsonString());
+  }
+
+  private static void clearFieldsNotRequiredForComparison(ApplicantData applicantData) {
+    // The `updated_at` timestamp for an answer should not be considered when
+    // comparing answers.
+    try {
+      applicantData.getDocumentContext().set("$..updated_at", 0);
```
bion, line 176: **suggestion:** please include a comment explaining that the `..` scans the whole document

server/app/services/applicant/ApplicantData.java
bion: IMO this is fine

server/app/services/applicant/ApplicantData.java
```diff
@@ -152,4 +153,29 @@ public Optional<LocalDate> getDateOfBirth() {
   public void setDateOfBirth(String dateOfBirth) {
     putDate(WellKnownPaths.APPLICANT_DOB, dateOfBirth);
   }
+
+  /**
+   * Returns `true` if the answers in `other` match the answers in the current object. Ignores
+   * certain metadata (such as timestamps) when comparing answers.
```
bion, line 159: **suggestion (blocking):** recommend being a little more explicit:
```suggestion
   * Returns `true` if the answers in `other` match the answers in the current object. Ignores
   * `updated_at` timestamps when comparing answers.
```

bion (approved):

server/app/services/applicant/ApplicantData.java
yotommy: Done.

PR #5294, opened 2023-07-25
Add all fields to API Response Preview

michaelzetune:
### Description

This PR adds non-`"application"` data to the json object in the API docs response preview. See the screenshots for what those are.

To do this, we expose the `buildJsonApplication` method in `JsonExporter` that is already being used for real API responses, to `ProgramJsonSampler`. It can then use this in order to create the full sample JSON, just like we do so for the real API.

#### General

   - Updated tests

#### User visible changes

### Issue(s) this completes

Relates to #5262


server/app/services/export/JsonExporter.java
```diff
@@ -79,59 +81,73 @@ private CfJsonDocumentContext buildJsonApplication(
       Application application, ProgramDefinition programDefinition) {
     ReadOnlyApplicantProgramService roApplicantProgramService =
         applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
+
```
michaelzetune, line 84: Everything in this file is functionally the same as before.

Now, instead of exposing `public static void exportEntriesToJsonApplication(CfJsonDocumentContext jsonApplication, ImmutableMap<Path, Optional<?>> entries)` for `ProgramJsonSampler` to use, we make that private and expose `CfJsonDocumentContext buildJsonApplication(JsonExportData jsonExportData)`. 

This is much more straightforward: give me a `JsonExportData` and I'll give you a `CfJsonDocumentContext` using that data. `ProgramJsonSampler` can easily build a `JsonExportData` and rely on this method to produce the sample JSON data.

server/test/services/export/ProgramJsonSamplerTest.java
```diff
@@ -23,28 +31,59 @@ public class ProgramJsonSamplerTest extends ResetPostgres {
 
   private ProgramJsonSampler programJsonSampler;
 
-  /**
-   * We need to mock the AutoValue in order to mock on streamQuestionDefinitions(), and this is
-   * substantially simpler than seeding the complex ProgramDefinition AutoValue with data.
-   */
-  @SuppressWarnings("DoNotMockAutoValue")
-  @Mock
-  private ProgramDefinition mockProgramDefinition;
```
michaelzetune, line 32: Since we need a couple more pieces of data from the test `ProgramDefinition`, let's stop mocking it and actually create one, for better test fidelity.

swatkat1: Cool!

swatkat1 (approved):

PR #5297, opened 2023-07-25
Use `getResponseJson` in `ApiDocsView` so that API Docs have full API response

michaelzetune:
### Description

Before this PR, the API docs page only shows part of the API response, namely the part under `"payload"`. Technically, this is enough, since the only other part of the the response is the `nextPageToken`. However, in the name of completeness, we want to show the full response so that API consumers understand what they are getting back from CiviForm.

This PR uses *the same code as the API itself* to generate this preview, so there will not be divergence between them. As an added bonus, the screenshot tests will operate as an implicit "golden test" in case the API were to change in a way we didn't notice.

<details>
<summary>Before</summary>

image
</details>

<details>
<summary>After</summary>

image

</details>

## Release notes

Improves the API Response Preview on API docs.

#### General

#### User visible changes

### Issue(s) this completes

Relates to #5262.


browser-test/image_snapshots/api_docs_test/api-docs-page-accordion-open.png
michaelzetune, line 1: Before, we would only show everything inside `payload` in the gray box. Now, we show the entire response.

server/app/controllers/api/CiviFormApiController.java
```diff
@@ -53,31 +49,4 @@ protected void assertHasProgramReadPermission(Http.Request request, String progr
       throw new UnauthorizedApiRequestException(apiKey, programSlug);
     }
   }
-
```
michaelzetune, line 56: Moved to `JsonExporter` (which makes a lot more sense anyways)

server/app/services/CfJsonDocumentContext.java
```diff
@@ -600,18 +597,6 @@ public String asJsonString() {
     return jsonData.jsonString();
   }
 
-  public String asPrettyJsonString() {
```
michaelzetune, line 603: Moved to `JsonPrettifier`

server/app/services/export/JsonExporter.java
```diff
@@ -147,6 +155,33 @@ CfJsonDocumentContext buildJsonApplication(JsonExportData jsonExportData) {
     return jsonApplication;
   }
 
+  public String getResponseJson(
```
michaelzetune, line 158: Unmodified from `CiviFormApiController`

server/app/services/export/JsonPrettifier.java
```diff
@@ -0,0 +1,20 @@
+package services.export;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+public class JsonPrettifier {
+  public static String asPrettyJsonString(String original) {
```
michaelzetune, line 8: Moved and adapted from`CfJsonDocumentContext`

server/app/services/export/JsonPrettifier.java
```diff
@@ -0,0 +1,20 @@
+package services.export;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
```
dkatzz, line 6: **suggestion:** Can you add a javadoc

dkatzz (approved):
Nice!

server/app/services/export/JsonPrettifier.java
michaelzetune: Done.

PR #5298, opened 2023-07-25
Allow env-var-docs parser package to run on older versions of python

rockycodes:
### Description

These changes undo syntax used in python version 3.10 and allow us to support older versions of python back to version 3.8. This unblocks turning on server variable generation for AR who is on Python 3.9.16

#### General

- [x] Performed manual testing (tested deployment with these changes)

Test this branch by running `bin/env-var-docs-run-tests`.

Can also test that server variable auto-generation works by building this branch into a Docker image and deploying it with server variable auto-generation enabled. There are many steps to this, so talk to Rocky if you want to test in this way.


bion (approved):

rasmi:
Hey Rocky, do you think 3.7+ is possible? The AWS Cloud Shell is running 3.7.16 and it's convenient to be able to run upgrades from there.

rockycodes:
Good question. @nb1701 actually did most of this work and I seem to remember it breaking on versions older than 3.8 but tagging him in for thoughts.

PR #5310, opened 2023-07-26
Show draft programs in API Docs

### LINKED ISSUE #4945: Show draft program spec in docs
Expand the program-specific API docs to also show the the expected response for the draft version of a program.

michaelzetune:
### Description

This PR adds on an additional dropdown in the API docs for the version of a program you want to view: Active or Draft. If that version is not available, we show a simple error message.

## Release notes

Adds version selection to API docs.

#### General

#### User visible changes

### Issue(s) this completes

Relates to #5262, fixes #4945.


browser-test/src/api_docs.test.ts
```diff
@@ -9,20 +9,64 @@ import {BASE_URL} from './support/config'
...
+    await page.goto(BASE_URL)
+    await loginAsAdmin(page)
+
+    await adminPrograms.publishAllDrafts()
+    await page.click('text=API docs')
+
+    await validateScreenshot(page, 'comprehensive-program-active-version')
+
+    await page.selectOption('#select-slug', {value: 'minimal-sample-program'})
+    await validateScreenshot(page, 'minimal-program-active-version')
```
bion, line 31: **suggestion (blocking):** in cases where the test manipulates the UI and then does a screenshot assertion, please insert an assertion on the updated text content of the page before doing the screenshot validation. This avoids a race condition where if the server is too slow to update the screenshot will be of the pre-updated UI.

server/app/views/api/ApiDocsView.java
```diff
@@ -83,31 +84,50 @@ public Content render(
...
       Http.Request request) {
 
     SelectTag slugsDropdown =
         select()
             .withId("select-slug")
-            .withClasses("border", "border-gray-300", "rounded-lg", "p-2", "ml-2")
-            .attr("onchange", "window.location.href = '/api/docs/v1/' + this.value");
-
+            .withClasses("border", "border-gray-300", "rounded-lg", "p-2", "mx-2")
+            .attr("onchange", "window.location.href = '/api/docs/v1/' + this.value + '/active'");
```
bion, line 104: **suggestion (blocking):** please move this code into a typescript file and follow the pattern used for registering DOM event handlers used on other pages

bion (changes requested):
Please add unit tests for the code changes in `VersionRepository` and `ProgramServiceImpl`

server/app/views/api/ApiDocsView.java
michaelzetune: Done.

server/app/views/api/ApiDocsView.java
```diff
@@ -83,31 +84,50 @@ public Content render(
...
+                    .withValue("active")
+                    .withCondSelected(request.uri().contains("/active")))
+            .with(
+                option("Draft")
+                    .withValue("draft")
+                    .withCondSelected(request.uri().contains("/draft")))
+            .attr(
+                "onchange",
+                "window.location.href = '/api/docs/v1/' +"
+                    + " document.getElementById('select-slug').value + '/' + this.value");
```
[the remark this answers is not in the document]
michaelzetune: Done.

server/app/assets/javascripts/api_docs.ts
```diff
@@ -0,0 +1,34 @@
+import {addEventListenerToElements} from './util'
+
+class ApiDocs {
+  // Redirects the user to the active version of the program slug they just selected upon
+  // clicking the select program slug.
+  static attachSlugEventListener() {
+    addEventListenerToElements('#select-slug', 'change', (event: Event) => {
+      console.log('test log')
```
bion, line 8: **suggestion (blocking):** remove console log

server/app/repository/ProgramRepository.java
```diff
@@ -173,6 +173,32 @@ public CompletableFuture<Program> getActiveProgramFromSlug(String slug) {
...
   }
 
+  /** Get the current draft program with the provided slug. */
+  public CompletableFuture<Program> getDraftProgramFromSlug(String slug) {
+    return supplyAsync(
+        () -> {
+          for (Program program : database.find(Program.class).where().isNull("slug").findList()) {
+            program.getSlug();
+            program.save();
+          }
```
bion, line 183: **issue (blocking):** This is unrelated to the name and javadoc of the method. In general reads should not alter database state unless they're explicitly named `findOrCreate`. If it's necessary for the feature to work please refactor for clarity, otherwise remove.

server/app/controllers/api/ApiDocsController.java
```diff
@@ -34,32 +36,49 @@ public ApiDocsController(
...
 
-    ProgramDefinition programDefinition;
     try {
-      programDefinition =
-          programService.getActiveProgramDefinitionAsync(programSlug).toCompletableFuture().join();
+      ProgramDefinition programDefinition =
+          getProgramDefFn.apply(programSlug).toCompletableFuture().join();
+      return Optional.of(programDefinition);
+
     } catch (Exception e) {
```
bion, line 80: **suggestion (blocking):** I know you didn't add this line, but we should **never** catch `Exception` like this, please update to a more specific exception type.

server/app/repository/ProgramRepository.java
```diff
@@ -173,6 +173,32 @@ public CompletableFuture<Program> getActiveProgramFromSlug(String slug) {
...
+
+          Optional<Version> version = versionRepository.get().getDraftVersion();
+
+          if (version.isEmpty()) {
+            throw new RuntimeException(new ProgramNotFoundException(slug));
```
bion, line 188: **suggestion (blocking):** 

1. would prefer this throws a checked exception so the caller decides how to handle it rather than the repository deciding for the caller that this will cause a server error.

2. use a different exception type -- `ProgramNotFoundException` doesn't match the exceptional case here that there is no draft version.

server/app/repository/ProgramRepository.java
```diff
@@ -173,6 +173,32 @@ public CompletableFuture<Program> getActiveProgramFromSlug(String slug) {
...
+          }
+
+          ImmutableList<Program> draftPrograms = version.get().getPrograms();
+          List<Program> programsMatchingSlug =
+              database.find(Program.class).where().eq("slug", slug).findList();
+          return draftPrograms.stream()
+              .filter(programsMatchingSlug::contains)
+              .findFirst()
+              .orElseThrow(() -> new RuntimeException(new ProgramNotFoundException(slug)));
```
bion, line 197: **suggestion (blocking):** throw the checked exception

server/app/repository/ProgramRepository.java
```diff
@@ -173,6 +173,32 @@ public CompletableFuture<Program> getActiveProgramFromSlug(String slug) {
...
+
+          Optional<Version> version = versionRepository.get().getDraftVersion();
+
+          if (version.isEmpty()) {
+            throw new RuntimeException(new ProgramNotFoundException(slug));
+          }
+
+          ImmutableList<Program> draftPrograms = version.get().getPrograms();
+          List<Program> programsMatchingSlug =
+              database.find(Program.class).where().eq("slug", slug).findList();
```
bion, line 193: **suggestion (blocking):** I don't think this second query is necessary? Looks like you could loop through `draftPrograms` and find the one with the matching `slug`

bion (changes requested):

server/app/repository/ProgramRepository.java
michaelzetune: For (1), this is tricky because Java does not allow checked exceptions in CompletableFutures, even if we declare `throws ProgramNotFoundException` in the method signature (since the method itself isn't throwing it). That's probably why the existing `getActiveProgramFromSlug` above is wrapping it in a `RuntimeException` as well.

For this I could make this method synchronous (avoiding the supplyAsync block) since there isn't any async code calling it right now, or we could leave the unchecked exception as-is. What do you think?

For (2), I could still make the underlying `ProgramNotFoundException` a different one, perhaps a new `ProgramDraftNotAvailableException`?

server/app/repository/ProgramRepository.java
michaelzetune: Same question as above.

server/app/controllers/api/ApiDocsController.java
michaelzetune: Updated to RuntimeException. I'll leave this comment unresolved until we figure out the best approach to throwing in the other method, since this might change.

michaelzetune: https://github.com/civiform/civiform/pull/5315 😄

server/app/repository/ProgramRepository.java
bion: 🙈 that is not going to make it easier for me to sleep this weekend. Thanks for pointing that out.

server/app/controllers/api/ApiDocsController.java
bion: Thanks!

server/app/repository/ProgramRepository.java
bion: Looking at Program and ProgramDefinition, I think this was added originally because the `slug` column wasn't populated for all programs at one point, so this code was included as a hacky way of doing a data migration. All new programs have the column populated though and it's been a while, so I think it's safe to remove this code.

server/app/repository/ProgramRepository.java
bion: > For this I could make this method synchronous (avoiding the supplyAsync block) since there isn't any async code calling it right now, or we could leave the unchecked exception as-is. What do you think?

I'd prefer this -- better to avoid the complexity of async when we don't need it.

bion: For (2): the important thing is to communicate to callers what could go wrong and what actually went wrong when it goes wrong. The problem wasn't that they asked for a program that doesn't exist, it's that they should not call this method when no draft version exists, so the exception and javadoc should reflect that.

server/app/repository/ProgramRepository.java
michaelzetune: I'll do this in a separate PR so that it's easy to roll back if necessary.

server/app/repository/ProgramRepository.java
michaelzetune: Made `getDraftProgramFromSlug` and `getDraftProgramDefinition` synchronous. Created and used new `ProgramDraftNotFoundException`.

server/app/repository/ProgramRepository.java
michaelzetune: `ProgramDraftNotFoundException` is now thrown

server/app/repository/ProgramRepository.java
michaelzetune: https://github.com/civiform/civiform/pull/5321

bion (approved):

PR #5315, opened 2023-07-27
Prohibit catching `Exception` or `throws`-ing `RuntimeException`

michaelzetune:
### `CatchingUnchecked`

This PR enables ErrorProne's `CatchingUnchecked` bug pattern, which prevents us from having a `catch (Exception e)` style block. We do so by enabling `-Xep:CatchingUnchecked:ERROR` in `build.sbt`.

https://errorprone.info/bugpattern/CatchingUnchecked

> This catch block catches `Exception`, but can only catch unchecked exceptions. Consider catching RuntimeException (or something more specific) instead so it is more apparent that no checked exceptions are being handled.

### `ThrowsUncheckedException`

While we're here, I also enabled `ThrowsUncheckedException`, which is less pressing but also enfoces better practice. It prevents us from declaring `throws RuntimeException` (and subclasses of it), which may misleadingly imply that callers will need to handle these exceptions at compile-time.

https://errorprone.info/bugpattern/ThrowsUncheckedException

> Use the Javadoc @throws tag to document each exception that a method can throw, but do not use the throws keyword on unchecked exceptions.

### General

### Issue(s) this completes

None.

### DIFF
--- a/server/app/auth/oidc/OidcClientProvider.java
+++ b/server/app/auth/oidc/OidcClientProvider.java
@@ -222,7 +222,7 @@ public abstract class OidcClientProvider implements Provider<OidcClient> {
 
     try {
       client.init();
-    } catch (Exception e) {
+    } catch (RuntimeException e) {
       logger.error("Error while initilizing OIDC provider", e);
       throw e;
     }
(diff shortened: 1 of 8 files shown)


bion (approved):
Thanks for improving this!

PR #5317, opened 2023-07-27
Fix payload display on API Docs

michaelzetune:
### Description

Previously, API docs showed the `"payload"` keyword as mapping to an object, instead of a list. This is incorrect, as the API shows a list of applications. Turns out this was a discrepency between how `ProgramJsonSampler` and `JsonExporter` create the JSON. 

The simplest solution is to have `ProgramJsonSampler` emulate the logic for multiple applications the way `JsonExporter` does. This is lazy though; it's better to have expose the higher-level functions in `JsonExporter` for `ProgramJsonSampler` to use directly, which is what we do here.

We also do a bit of refactoring of `JsonExporter` to help modularize the single/multi application logic. I've added comments in the diff to explain this.

## Release notes

None.

#### General

#### User visible changes

### Issue(s) this completes

Relates to #5262.


server/app/controllers/api/ProgramApplicationsApiController.java
```diff
@@ -115,11 +115,10 @@ public CompletionStage<Result> list(
                 throw new RuntimeException(e);
               }
 
-              String applicationsJson =
-                  jsonExporter.export(programDefinition, paginationResult).getLeft();
+              String applicationsJson = jsonExporter.export(programDefinition, paginationResult);
 
               String responseJson =
-                  jsonExporter.getResponseJson(
```
michaelzetune, line 122: Same method, just renamed.

server/app/services/export/JsonExporter.java
```diff
@@ -71,21 +70,28 @@ public Pair<String, PaginationResult<Application>> export(
     return export(programDefinition, paginationResult);
   }
 
-  public Pair<String, PaginationResult<Application>> export(
+  public String export(
```
avaleske, line 73: nonblocking: While you're in here, can we disambiguate between these two `export()` methods? Either with a javadoc and/or the name?

iiuc one is for the admin export view, and the other is for the API and provides pagination information.

avaleske (approved):

Issue #4944, opened 2023-05-24
Error when advancing to next screen in application


rasmi:
From @sophiatareen:
**Describe the bug**
Error when advancing to the next screen after filling out information as a TI in a demo/staging environment.

**To Reproduce**
1. Log in as a TI
2. Go to existing client with previously entered data (e.g. Soph Ta)
3. Start Rental Assistance Program, screen 1
4. Enter requested data (e.g. DOB, Address); Save and next

**Expected behavior**
CiviForm advances to the next screen. Instead, an error is shown (see below):

![unnamed](https://github.com/civiform/civiform/assets/2267370/ee12eb4e-746a-4bb7-89a9-80088664bb1f)

Stack trace from logs:
```
[[31merror[0m] p.h.DefaultHttpErrorHandler - 
```

```
! @81k9do3g8 - Internal server error, for (GET) [/applicants/2554/programs/27/review] ->
```

```
play.api.UnexpectedException: Unexpected exception[NoSuchElementException: No value present]
	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:358)
	at play.api.http.HttpErrorHandlerExceptions.throwableToUsefulException(HttpErrorHandler.scala)
	at play.http.DefaultHttpErrorHandler.throwableToUsefulException(DefaultHttpErrorHandler.java:246)
	at play.http.DefaultHttpErrorHandler.onServerError(DefaultHttpErrorHandler.java:182)
	at controllers.ErrorHandler.onServerError(ErrorHandler.java:102)
	at play.core.j.JavaHttpErrorHandlerAdapter.onServerError(JavaHttpErrorHandlerAdapter.scala:32)
	at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:441)
	at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:433)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:490)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[...]

PR #5320, opened 2023-07-28
Redirect user to home on APRC failures

yotommy:
### Description

In cases where the `ApplicantProgramReviewController` cannot fetch the user profile, or the user is unauthorized, redirect the user to the home page.

## Release notes

In cases where the user profile cannot be fetched, or the user is unauthorized, redirect the user to the home page. This case could occur if a tab is left open overnight and the browser session expires.

If a user profile is not available, then it doesn't make sense to create a guest browser session and try the action again, since the new account will not be authorized to see the application for the existing account. Sending the user to the homepage seems reasonable for this case.

Similarly, if the user fails authorization for any reason, we currently just return 401 and nothing shows in the browser. Redirecting to the home page also makes sense here.

This addresses two out of three stack traces reported in #4944.

#### General

### DIFF
--- a/server/test/controllers/HasBooleanHeaderArgumentMatcher.java
+++ b/server/test/controllers/HasBooleanHeaderArgumentMatcher.java
@@ -1,0 +1,27 @@
+package controllers;
+
+import org.mockito.ArgumentMatcher;
+import play.mvc.Http;
+
+/**
+ * ArgumentMatcher that matches an Http.Request containing a specified header
+ * whose value is "true".
+ */
+public class HasBooleanHeaderArgumentMatcher implements ArgumentMatcher<Http.Request> {
+  private final String headerName;
+
+  HasBooleanHeaderArgumentMatcher(String headerName) {
+    this.headerName = headerName;
+  }
+
+  @Override
+  public boolean matches(Http.Request argument) {
+    if (argument == null || !argument.hasHeader(headerName)) return false;
+    return Boolean.parseBoolean(argument.header(headerName).get());
+  }
+
+  @Override
+  public String toString() {
+    return String.format("[HTTP Request has header %s containing \"true\"]", headerName);
+  }
+}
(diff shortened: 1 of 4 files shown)


server/test/controllers/HasBooleanHeaderArgumentMatcher.java
```diff
@@ -0,0 +1,26 @@
...
+public class HasBooleanHeaderArgumentMatcher implements ArgumentMatcher<Http.Request> {
+  private final String headerName;
+
+  HasBooleanHeaderArgumentMatcher(String headerName) {
+    this.headerName = headerName;
+  }
+
+  @Override
+  public boolean matches(Http.Request argument) {
+    if (argument == null || !argument.hasHeader(headerName)) return false;
```
bion, line 18: **nitpick:** https://google.github.io/styleguide/javaguide.html#s4.1-braces
```suggestion
    if (argument == null || !argument.hasHeader(headerName)) {
      return false;
    }
```

server/test/controllers/HasBooleanHeaderArgumentMatcher.java
```diff
@@ -0,0 +1,26 @@
+package controllers;
+
+import org.mockito.ArgumentMatcher;
+import play.mvc.Http;
+
+/**
+ * ArgumentMatcher that matches an Http.Request containing a specified header whose value is "true".
+ */
+public class HasBooleanHeaderArgumentMatcher implements ArgumentMatcher<Http.Request> {
```
bion, line 9: **nitpick (blocking):** 
```suggestion
public final class HasBooleanHeaderArgumentMatcher implements ArgumentMatcher<Http.Request> {
```

bion (approved):
two small items, otherwise looks good

yotommy:
The updated [coverage report](https://app.codecov.io/github/civiform/civiform/commit/3fc566f35c13c38136dbfc6b25b69feb3a85606e/blob/server/app/controllers/applicant/ApplicantProgramReviewController.java) shows the new code paths are covered by tests.

PR #5331, opened 2023-08-01
Handle getting state abbreviation for possible secondary field

### LINKED ISSUE #5291: Address Standarization does not correct for State
**Describe the bug**

Addresses that are entered with the wrong state are not corrected during the address standardization window. 

1. Enter a correct Address in WA. Enter any other state than WA. 
2. When the standardization page asks you to select a correct address, you can select an address with the original incorrect state. 

**Expected behavior**

All Seattle addresses should at least correct to WA for the state. 

**Screenshots**

**Additional context**

gwendolyngoetz:
### Description

Get the state abbreviation from the attributes json node. On public ESRI this is under the RegionAbbr field, but some custom implementations may have it under the Region field.

## Release notes

#### General

### Issue(s) this completes

Fixes  #5291

### DIFF
--- a/server/app/services/geo/esri/EsriClient.java
+++ b/server/app/services/geo/esri/EsriClient.java
@@ -16,7 +16,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.function.BiFunction;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.libs.Json;
@@ -119,7 +119,6 @@ public abstract class EsriClient {
               ImmutableList.Builder<AddressSuggestion> suggestionBuilder = ImmutableList.builder();
               for (JsonNode candidateJson : json.get("candidates")) {
                 JsonNode location = candidateJson.get("location");
-                JsonNode attributes = candidateJson.get("attributes");
                 AddressLocation addressLocation =
                     AddressLocation.builder()
                         .setLongitude(location.get("x").asDouble())
@@ -127,19 +126,8 @@ public abstract class EsriClient {
                         .setWellKnownId(wkid)
                         .build();
 
-                Address candidateAddress =
-                    Address.builder()
-                        .setStreet(attributes.get("Address").asText())
-                        .setLine2(
-                            attributes.get("SubAddr") == null || attributes.get("SubAddr").isEmpty()
-                                ? address.getLine2()
-                                : attributes.get("SubAddr").asText())
-                        .setCity(attributes.get("City").asText())
-                        .setState(getStateAbbreviationFromAttributes(attributes, address))
-                        .setZip(attributes.get("Postal").asText())
-                        .build();
-                // Suggestion must be a fully formed address.
-                // Sometimes the esri service returns only a partially formed address.
+                Address candidateAddress = mapAddressAttributesJson(candidateJson, address);
+
                 if (candidateAddress.getStreet().isEmpty()
                     || candidateAddress.getCity().isEmpty()
                     || candidateAddress.getState().isEmpty()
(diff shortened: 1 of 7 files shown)


server/app/services/geo/esri/EsriClient.java
```diff
@@ -179,6 +177,35 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
...
   }
 
+  /**
+   * Get the state abbreviation from the attributes json node. on Public ESRI this is under the
+   * RegionAbbr field, but some custom implementations may have it under the Region field.
+   */
+  private String getStateAbbreviationFromAttributes(JsonNode attributes, Address address) {
+    BiFunction<JsonNode, String, String> getValue =
+        (jsonNode, keyName) ->
+            (jsonNode.get(keyName) == null || jsonNode.get(keyName).isNull())
```
dkatzz, line 187: **question:** Why do we need both of these null checks? Should it be isEmpty like above?

server/app/services/geo/esri/EsriClient.java
```diff
@@ -179,6 +177,35 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
...
+                ? ""
+                : jsonNode.get(keyName).asText();
+
+    // Start with getting the value from the RegionAbbr, this is the
+    // default used by ESRI for state abbreviations
+    String result = getValue.apply(attributes, "RegionAbbr");
+
+    // Assume if it's not two characters it is either empty or the
+    // full state name and pull from the Region value which
```
dkatzz, line 196: **suggestion:** Remove which (or finish sentence)

dkatzz (approved):

dkatzz:
Nit: you can remove some of the sections in the PR description that aren't being used.

server/app/services/geo/esri/EsriClient.java
gwendolyngoetz: One checks if the object is null, the other checks for a null node type. I'm going to write some more tests, I think using isEmpty is actually a bug.

server/app/services/geo/esri/EsriClient.java
gwendolyngoetz: done

server/app/services/geo/esri/EsriClient.java
gwendolyngoetz: @dkatzz, Ok fixed a whole different bug than I planned. Turns out `isEmpty` on scalars always returns true. It's really looking at children in arrays/objects. The `isNull` method is the correct thing to use. I've pull things apart a little more to fix it and switched to a more targeted series of tests to check the mapping.

server/app/services/geo/esri/EsriClient.java
```diff
@@ -179,6 +165,75 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
             });
   }
 
+  /** Maps the json values within the node to an address */
```
dkatzz, line 168: **nitpick:** Capitalize to JSON (here and below)

server/app/services/geo/esri/EsriClient.java
```diff
@@ -179,6 +165,75 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
...
+        // returned by the ESRI instance and we don't want to lose the value.
+        .setLine2(getNodeTextValueOrDefault(attributes, "SubAddr", address::getLine2))
+        .setCity(getNodeTextValueOrEmpty(attributes, "City"))
+        .setState(getStateAbbreviationFromAttributes(attributes, address))
+        .setZip(getNodeTextValueOrEmpty(attributes, "Postal"))
+        .build();
+  }
+
+  /**
+   * Get the state abbreviation from the attributes json node. on Public ESRI this is under the
```
dkatzz, line 185: **nitpick:** Capitalize to On

server/test/services/geo/esri/EsriClientTest.java
```diff
@@ -177,4 +182,83 @@ public void getAddressSuggestionsWithError() throws Exception {
     assertThat(group.getWellKnownId()).isEqualTo(0);
     assertThat(group.getOriginalAddress()).isEqualTo(address);
   }
+
```
dkatzz, line 185: **suggestion:** Could you add a test where both Region and RegionAbbr are both > 2 letters?

server/test/services/geo/esri/EsriClientTest.java
```diff
@@ -177,4 +182,83 @@ public void getAddressSuggestionsWithError() throws Exception {
...
+            + "  \"Region\": \"WA\","
+            + "  \"RegionAbbr\": \"Washington\","
+            + "  \"Postal\": \"11111-expected\""
+            + "}}";
+
+    runMapAddressAttributesJson(
+        json, "street-expected", "line2-user", "city-expected", "WA", "11111-expected");
+  }
+
+  private void runMapAddressAttributesJson(
```
dkatzz, line 237: **suggestion:** I'm not sure what the best name is, but thoughts on calling this something like runAndVerifyMapAddressAttributesJson or something? I think with the name its unclear this function is doing the asserts as well.

server/app/services/geo/esri/EsriClient.java
dkatzz: Ah makes sense - thanks for the update! The way you pulled things apart LGTM - just some minor comments.

PR #5332, opened 2023-08-01
Allow api docs screen to handle missing questions

gwendolyngoetz:
### Description

Render the empty question sample option if `NULL_QUESTION` type

## Release notes

#### General


avaleske (commented):
The json API doesn't currently export the null question type. Do we need to add this there as well?

gwendolyngoetz:
Abandoning this because it's not a strictly under an admin route and thus should stay broken until the program is fixed.

PR #5346, opened 2023-08-03
Adding sort on getAllProgramSlugs

gwendolyngoetz:
### Description

Sort the slugs in getAllProgramSlugs

#### General

### DIFF
--- a/server/app/services/program/ProgramServiceImpl.java
+++ b/server/app/services/program/ProgramServiceImpl.java
@@ -99,6 +99,7 @@ public final class ProgramServiceImpl implements ProgramService {
   public ImmutableSet<String> getAllProgramSlugs() {
     return getAllProgramNames().stream()
         .map(MainModule.SLUGIFIER::slugify)
+        .sorted()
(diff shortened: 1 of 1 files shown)


avaleske (approved):
Should the other `getAllProgramNames()` and related methods also be sorted?

gwendolyngoetz:
I believe everything show to the user should have a sort on it. Poor UX to have to scour through a list. In this case the only other place using the `getAllProgramSlugs` method is in the api key generator and I checked and it looks better now too.

avaleske:
> I believe everything shown to the user should have a sort on it. Poor UX to have to scour through a list. In this case the only other place using the `getAllProgramSlugs` method is in the api key generator and I checked and it looks better now too.

Ah sorry to clarify: I agree. I was asking if, while you're there, you thought `getAllProgramNames()` should also sort its output.

gwendolyngoetz:
I think it should, but that feels like it has the potential for extra work beyond just adding the sort.

PR #5349, opened 2023-08-03
Update env-var-docs markdown generation for containerization

nb1701:
Now that the script to generate the markdown runs inside a container, we need to change how this action works. This now clones into the civiform-release directory inside of the civiform-main directory, since the civiform-main directory is volume mounted into the container. It also changes the ENV_VAR_DOCS_PATH to point to the relative location rather than the path on the action runner.

This additionally removes the step of setting up Python. Since it is containerized, the action runner doesn't need to have it installed.

#### General


rockycodes (approved):
lgtm!

PR #5368, opened 2023-08-04
Question Translations Save button will load the same view on success

### LINKED ISSUE #5377: Save translations without being kicked out of translation screen
When adding translations to programs and questions, after hitting save for a language the translation screen closes, requiring program admin to navigate back to the translation screen

Admin should be able to click save after updating a language and be able to then select the next language to continue adding translations rather that getting sent back to the question or program screen.

swatkat1:
### Description

Saving the translation is rerouting the admins to the question edit page. This PR changes the flow and save it but doesn't reroute. This PR also add a "Back" button for easy maneuverability. 

## Release notes

Question Translations Save button will load the same view on success.

#### General

- [x] Extended the documentation, if necessary

### Issue(s) this completes

Fixes #5377


server/app/views/admin/TranslationFormView.java
```diff
@@ -84,6 +85,12 @@ protected final FormTag renderTranslationForm(
       Locale locale,
       String formAction,
       ImmutableList<DomContent> formFieldContent) {
+    String redirectUrl = "";
+    if (formAction.contains("program")) {
+      redirectUrl = controllers.admin.routes.AdminProgramController.index().url();
+    } else {
+      redirectUrl = controllers.admin.routes.AdminQuestionController.index().url();
+    }
```
gwendolyngoetz, line 93:
```suggestion
    String redirectUrl = 
      formAction.contains("program")) 
        ? controllers.admin.routes.AdminProgramController.index().url() 
        : controllers.admin.routes.AdminQuestionController.index().url();
```

server/app/views/admin/TranslationFormView.java
```diff
@@ -84,6 +85,12 @@ protected final FormTag renderTranslationForm(
       Locale locale,
       String formAction,
       ImmutableList<DomContent> formFieldContent) {
+    String redirectUrl = "";
+    if (formAction.contains("program")) {
```
gwendolyngoetz, line 89: Is there any other way to determine the between program and question that doesn't rely on the formAction?

server/app/views/admin/TranslationFormView.java
```diff
@@ -96,7 +103,13 @@ protected final FormTag renderTranslationForm(
...
                             locale.getDisplayLanguage(LocalizedStrings.DEFAULT_LOCALE)))
                     .withId("update-localizations-button")
-                    .withClasses(ButtonStyles.SOLID_BLUE));
+                    .withClasses(ButtonStyles.SOLID_BLUE),
+                div(a().withHref(redirectUrl)
+                        .with(
+                            button("Back")
+                                .withId("back-to-edit")
+                                .withClasses(ButtonStyles.OUTLINED_TRANSPARENT)))
+                    .withClasses("mt-4"));
```
gwendolyngoetz, line 112: (blocking) Place the buttons on the same line with the back button first. The html should be something like

```html
<div class="flex flex-row">
   <button>Back</button>
   <button>Save</button>
</div>
```

server/app/views/admin/TranslationFormView.java
```diff
@@ -96,7 +103,13 @@ protected final FormTag renderTranslationForm(
                             "Save %s updates",
                             locale.getDisplayLanguage(LocalizedStrings.DEFAULT_LOCALE)))
                     .withId("update-localizations-button")
-                    .withClasses(ButtonStyles.SOLID_BLUE));
+                    .withClasses(ButtonStyles.SOLID_BLUE),
+                div(a().withHref(redirectUrl)
+                        .with(
+                            button("Back")
+                                .withId("back-to-edit")
```
gwendolyngoetz, line 110: Since the redirectUrl is going to the list screen the Id name is confusing. It would make more sense to call it `back-to-list-button`.

gwendolyngoetz (changes requested):

server/app/views/admin/TranslationFormView.java
swatkat1: done!

server/app/views/admin/TranslationFormView.java
swatkat1: Done!

server/app/views/admin/TranslationFormView.java
swatkat1: changed it to accept a boolean instead (/*isProgramEdit=*/)

server/app/views/admin/TranslationFormView.java
```diff
@@ -83,20 +84,33 @@ protected final FormTag renderTranslationForm(
...
-                submitButton(
-                        String.format(
-                            "Save %s updates",
-                            locale.getDisplayLanguage(LocalizedStrings.DEFAULT_LOCALE)))
-                    .withId("update-localizations-button")
-                    .withClasses(ButtonStyles.SOLID_BLUE));
+                div()
+                    .withClasses("flex", "flex-row", "gap-x-2")
+                    .with(
+                        a().withHref(redirectUrl)
```
gwendolyngoetz, line 103: Putting a button inside an a tag isn't valid html5 and could cause some weirdness with them both trying to do actions. Simple fix though just move the button styles onto the a tag and remove the button tag.

gwendolyngoetz (changes requested):

server/app/views/admin/TranslationFormView.java
swatkat1: done!

PR #5375, opened 2023-08-07
Add application revision state to JSON export

### LINKED ISSUE #5029: Add application lifecycle stage to the API
This will need a small design doc to ensure we're on the same page about how to expose enums via the API

yotommy:
### Description

Add application revision state to JSON export.

## Release notes

Add application revision state to JSON export. This is a string derived from the lifecycle stage of an application that is meant to be meaningful to the API consumer:

| LifeCycleStage | Revision State |
|--------|--------|
| `ACTIVE` | `CURRENT` |
| `OBSOLETE` | `OBSOLETE` |

Applications with a `LifeCycleStage` value of `DRAFT` or `DELETED` are not currently exported, so revision state is not supported for those values.

#### General

### Issue(s) this completes

Fixes #5029


server/app/models/Application.java
```diff
@@ -52,6 +53,17 @@ public class Application extends BaseModel {
...
+      ImmutableMap.of(
+          LifecycleStage.DRAFT,
+          "NOT_SUBMITTED",
+          LifecycleStage.ACTIVE,
+          "CURRENT",
+          LifecycleStage.OBSOLETE,
+          "OBSOLETE",
+          LifecycleStage.DELETED,
+          "DELETED");
+
```
gwendolyngoetz, line 66: This should get encapsulated into it's own method on the `LifecycleStage` enum. Like a `getDescription` method or something of that ilk. 

```
LifecycleStage stage  = LifecycleStage.DRAFT;
String description = stage.getDescription();
```

server/app/services/export/ProgramJsonSampler.java
```diff
@@ -56,6 +56,7 @@ public CfJsonDocumentContext getSampleJson(ProgramDefinition programDefinition)
             // for those tests, otherwise to the actual program ID.
             .setLanguageTag(Locale.US.toLanguageTag())
             .setSubmitterType(SubmitterType.APPLICANT)
+            .setSubmissionStatus("CURRENT")
```
gwendolyngoetz, line 59: This could then become `LifecycleStage.ACTIVE.getDescription()` based on my comment on moving the enum. Then we don't have to remember to update it in the future.

gwendolyngoetz (changes requested):

server/app/models/Application.java
yotommy: You know, I had it this way before and then moved it to Application, because I thought the submission status seemed related to the application. But looking at your suggestion now, I agree. Done!

gwendolyngoetz:
If you can add a test for `Unknown` that would be nice, but otherwise LGTM.

gwendolyngoetz (approved):

yotommy:
I don't really see a way of testing `UNKNOWN` at present. It's there in case someone adds a new `LifeCycleStage` value and forgets to update the mapping. This could be addressed through Java switch expressions, since they are exhaustive, but we can't use this yet since we are on JDK 11.

avaleske:
> I don't really see a way of testing `UNKNOWN` at present. It's there in case someone adds a new `LifeCycleStage` value and forgets to update the mapping. This could be addressed through Java switch expressions, since they are exhaustive, but we can't use this yet since we are on JDK 11.

I have thoughts here but I'm still mid-review pls hold 📞

server/app/models/LifecycleStage.java
```diff
@@ -17,6 +18,13 @@ public enum LifecycleStage {
 
   private final String stage;
 
+  private static ImmutableMap<LifecycleStage, String> descriptions =
```
avaleske, line 21: issue: These values aren't really tied to the LifecycleStage model, and are only relevant when we're exporting applications, so I don't know that this is the right place for them.

issue: By declaring these as strings, we end up matching strings in a few different places, such as tests or when building the `ApplicationJsonExportData`.

suggestion: Can we declare this as an enum in services/export/enums, similar to the `SubmitterType` enum? And then call `.toString()` on it when we actually want to export it?

server/app/models/LifecycleStage.java
```diff
@@ -17,6 +18,13 @@ public enum LifecycleStage {
...
+      ImmutableMap.of(
+          LifecycleStage.DRAFT, "NOT_SUBMITTED",
+          LifecycleStage.ACTIVE, "CURRENT",
+          LifecycleStage.OBSOLETE, "OBSOLETE",
```
avaleske, line 25: suggestion: If we move this to services/export/enums, and then do the mapping from LifecycleStage to this in the json export code, then this will only ever be used in the context of the export. That means we can just throw `NotImplemented` if the LifecycleStage is anything other than Active or Obsolete, since the export code should never even see Draft or Deleted applications.

server/test/services/export/JsonExporterTest.java
```diff
@@ -174,6 +174,9 @@ private void testApplicationTopLevelAnswers(
         "$[" + resultIndex + "].applicant_id", application.getApplicant().id);
     resultAsserter.assertValueAtPath("$[" + resultIndex + "].application_id", application.id);
     resultAsserter.assertValueAtPath("$[" + resultIndex + "].language", "en-US");
+    resultAsserter.assertValueAtPath(
```
avaleske, line 177: issue: It's not clear from the TODO at the top of the file, but this test is being replaced soon.

suggestion: Instead, can we add a test in the style of `export_whenSubmitterIsTi_tiTopLevelFieldsAreSet()` for each of the `submission_status` types?

server/app/services/export/JsonExporter.java
```diff
@@ -181,6 +182,8 @@ private CfJsonDocumentContext buildSingleApplicationJson(
                 jsonApplication.putString(
                     submitTimePath, dateConverter.renderDateTimeDataOnly(submitTime)),
             () -> jsonApplication.putNull(submitTimePath));
+    jsonApplication.putString(
+        Path.create("submission_status"), applicationJsonExportData.submissionStatus());
```
avaleske, line 186: nitpick: I'm open to being convinced I'm wrong, but I don't _love_ `submission_status` for this field name. It sounds to me like it's the application's status in the submission process, and would have values like "in progress", "accepted", "in review" etc, which is not what this is.

Maybe something along the lines of `revision` or `revision_type`? (Although "type" here is kind of odd too... hmm...)

See also the Google AIPs, which use "revision": https://google.aip.dev/162

This also leaves room for a future "revision_set_id` or `application_set_id` field that ties revisions of the same application together

avaleske (changes requested):
Thank you for working on this!

I owe the team a "principles for API features" to explain some of my reasoning here and I haven't written that yet, so sorry about that.

server/app/services/export/JsonExporter.java
avaleske: `revision_state` could work here actually, instead of `revision_type`

(Inspired by https://google.aip.dev/216)

yotommy: Done

server/test/services/export/JsonExporterTest.java
yotommy: Done

server/app/models/LifecycleStage.java
yotommy: I like the enum idea, and I belatedly remember you mentioning this at the beginning. Done.

server/test/services/export/AbstractExporterTest.java
```diff
@@ -703,5 +703,21 @@ public Application submit() {
 
       return application;
     }
+
+    public Application obsolete() {
```
avaleske, line 707: nit (not-blocking): Can we name this with a verb or preposition to match the other methods in this "builder"

avaleske (approved):
This looks great, thank you!

avaleske:
Also +@bion to double check new API field name, since once this is submitted we're stuck with the name we've chosen.

server/test/services/export/AbstractExporterTest.java
avaleske: Also, might be helpful to have setting it obsolete be distinct from submitting it, if possible? I'm not sure how the lifecycle stage currently gets updated in prod code.

yotommy: This was my intention, invoking the arguably cringy [obsolete-as-a-verb](https://www.merriam-webster.com/dictionary/obsolete) usage.

Issue #3287, opened 2022-08-26
Move focus to error message after applicant tries to submit invalid data


jeanniefu:
From our accessibility advisors - when an applicant tries to submit invalid data and we display an error message, we should consider moving focus onto the error message. We have both question level and field level errors. Their suggestions are:
* Move focus onto the question level error message using focus(), but ensure that we add tabindex="-1" to not cause it to go on tab focus order.
* (OR) move focus onto the field that has the error. This should then read out to the screen reader users that an error occurred here.

jeanniefu:
#3563 is required to meet WCAG 2.1 standards. This issue is a nice to have feature. 

Proposal for this issue is: if there is only one input on the page with an error, move focus to that element so the error is read out. (If there are multiple errors, then use the current behavior implemented in #3563).

See #3563 for more discussion.

jeanniefu:
We also received feedback during AR a11y testing that this would be helpful.

PR #5383, opened 2023-08-09
Move focus to input field when a single question on an application block has errors

melanie-exygy:
### Description

Per our accessibility advisors' recommendation, moving focus to the first field with an error when a single question on an application block has errors.  This directs the screen reader directly to the field that needs to be updated.  The screen reader then reads the question and the error message.  When there are multiple questions with errors, focus is not redirected.  Rather, we have kept the existing behavior in which the screen reader gives a vocal alert of there being multiple errors on the page.

## Release notes

Per our accessibility advisors' recommendation, moving focus to the first field with an error when a single question on an application block has errors.  This directs the screen reader directly to the field that needs to be updated.  The screen reader then reads the question and the error message.  When there are multiple questions with errors, focus is not redirected.  Rather, we have kept the existing behavior in which the screen reader gives a vocal alert of there being multiple errors on the page.

#### General

#### User visible changes

1. Ensure a program has a required question.
2. Turn on a screen reader.
3. Begin filling out the application and hit "Next" without filling out the required question.
4. Notice that the screen reader goes directly to the missing field.

### Issue(s) this completes

Fixes #<[3287](https://github.com/civiform/civiform/issues/3287)>;


server/app/views/questiontypes/AddressQuestionRenderer.java
```diff
@@ -115,6 +115,14 @@ protected DivTag renderInputTags(
       cityField.forceAriaInvalid();
       stateField.forceAriaInvalid();
       zipField.forceAriaInvalid();
+      if (params
+          .errorDisplayMode()
+          .equals(ApplicantQuestionRendererParams.ErrorDisplayMode.DISPLAY_SINGLE_ERROR)) {
+        streetAddressField.focusOnError();
```
rockycodes, line 121: question (blocking): what happens when you add "focus" to all of the fields? i tried it out and it focused on the first field for me (streetAddressField) even when I had filled that one out. I'm curious if we need to add `focusOnError` to all of the fields, or maybe just the first one. (apologies if we talked about this yesterday and I am not remembering!)

server/test/views/components/SelectWithLabelTest.java
```diff
@@ -48,6 +48,13 @@ public void createSelect_rendersSelectedOption() {
     assertThat(selectWithLabel.getSelectTag().render()).doesNotContain("<optgroup>");
   }
 
+  @Test
```
rockycodes, line 51: suggestion (non-blocking): I don't know that you need this test since `SelectWithLabel` is using the `focusOnError` method on `FieldWithLabel` (if I am understanding correctly). I think it's enough to have the `FieldWithLabel` test.

server/test/views/questiontypes/RadioButtonQuestionRendererTest.java
```diff
@@ -100,4 +103,19 @@ public void render_withAriaLabels() {
     assertThat(result.render().matches(".*fieldset aria-describedby=\"[A-Za-z]{8}-description\".*"))
         .isTrue();
   }
+
+  @Test
+  public void renderWithErrors_andSingleErrorMode_hasAutofocus() {
```
rockycodes, line 108: question (non-blocking): Why did you leave out a test for the absence of "autofocus" in this test suite? I don't know that you need to include it, just curious about the reasoning.

rockycodes: I'm also curious about your overall testing strategy. Why did you add tests for Radio and Checkbox but not others types like Text? (Apologies again if we talked about this and I don't remember 🙈 )

rockycodes (changes requested):
Overall this looks great! Very clear. I have a couple of clarifying questions before I approve.

server/app/views/components/FieldWithLabel.java
```diff
@@ -571,6 +576,7 @@ FieldErrorsInfo applyAttrsGenFieldErrorsInfo(T fieldTag) {
     fieldTag.condAttr(!ariaIds.isEmpty(), "aria-describedby", StringUtils.join(ariaIds, " "));
     fieldTag.condAttr(
         shouldForceAriaInvalid || fieldErrorsInfo.hasFieldErrors, "aria-invalid", "true");
+    fieldTag.condAttr(focusOnError, "autofocus", "");
```
rockycodes, line 579:
```suggestion
    fieldTag.condAttr(focusOnError, Attr.AUTOFOCUS, "");
```

server/app/views/questiontypes/AddressQuestionRenderer.java
rockycodes: Ah gotcha, that makes sense! I'm thinking about what would be the most clear for another dev coming across this. The multiple `focusOnError` calls would make me think that functionality is already in place. But it would also be confusing to only have the focus on the first one. I defer to your judgement, but what do you think about adding a comment to explain?

rockycodes: Sounds good!

PR #5394, opened 2023-08-09
Updating add user as program admin error message

gwendolyngoetz:
### Description

On the add user as a program admin page it displays an error message if the user doesn't not yet have an admin account. Updating the error message to provide more info the the CiviForm Admin on why they can't add them and what to do.

From
`user@example.org does not have an admin account and cannot be added as a Program Admin.`

To
`Cannot add user@example.org as a Program Admin because they do not have an admin account. Have the user log in as admin on the home page, then they can be added as a Program Admin.`

#### General


melanie-exygy (approved):
Looks good!

Issue #3571, opened 2022-09-29
(WCAG 2.1) Set focus to specific input field after user clicks Q from application summary page.


jeanniefu:
Required to meet WCAG 2.1 guidelines. Identified during an accessibility audit against WCAG 2.1.

**WCAG SC**: [2.4.3. Focus Order](https://www.w3.org/TR/WCAG21/#focus-order)

**Observed results**: Focus is not set explicitly to the correct form field when "Start here" was activated for that particular application question (in this case - What is your favorite color?)

**How to fix**: After the user clicks on a specific link from the preview page, we should set focus to that particular input field in the application block. (e.g. "What is your favorite color?" input field).

![image](https://user-images.githubusercontent.com/9094240/192914038-f0b81250-827e-4612-a765-03fb4d9af522.png)

rockycodes:
Completed by: https://github.com/civiform/civiform/pull/5409

PR #5409, opened 2023-08-11
Focus on question input when applicant clicks on question from review screen

rockycodes:
### Description

Per accessibility recommendations, when an applicant clicks on a question from the application review screen, we want the focus to be set explicitly on the input for the question they clicked on. This PR adds that functionality.

#### General

- [x] Performed manual testing (only on Chrome because Firefox interacts poorly with tab order)

1. Create and publish a program with multiple questions on a single screen
2. As an applicant, click "apply"
3. Turn on a screen reader
4. From the program summary screen, click on any question
5. The cursor should land you on the input for the question you clicked on and the screen reader should start reading from that question
6. You should also see two params in the url `questionName` and `questionType`

https://github.com/civiform/civiform/assets/24319951/5d471f5e-2ac2-4353-8a19-fa56f5fb56bd

### Issue(s) this completes

https://github.com/civiform/civiform/issues/3571


server/conf/routes
```diff
@@ -149,7 +149,7 @@ GET     /applicants/:applicantId/programs/:programId
 GET     /applicants/:applicantId/programs/:programId/edit                                      controllers.applicant.ApplicantProgramsController.edit(request: Request, applicantId: Long, programId: Long)
 GET     /applicants/:applicantId/programs/:programId/review                                   controllers.applicant.ApplicantProgramReviewController.review(request: Request, applicantId: Long, programId: Long)
 POST    /applicants/:applicantId/programs/:programId/submit                                    controllers.applicant.ApplicantProgramReviewController.submit(request: Request, applicantId: Long, programId: Long)
-GET     /applicants/:applicantId/programs/:programId/blocks/:blockId/edit                      controllers.applicant.ApplicantProgramBlocksController.edit(request: Request, applicantId: Long, programId: Long, blockId: String)
+GET     /applicants/:applicantId/programs/:programId/blocks/:blockId/edit                      controllers.applicant.ApplicantProgramBlocksController.edit(request: Request, applicantId: Long, programId: Long, blockId: String, questionName: java.util.Optional[String], questionType: java.util.Optional[String])
```
rockycodes, line 152: This was the only way I could think of to have the question edit screen know which question the applicant clicked on. There also is already a pattern of checking for optional params to manage conditional logic in the application. But editing the url comes with annoying side affects like needed to pass in `Optional.empty()` everywhere `.edit` is invoked. I'm open to other ideas for how to do this if folks have them!

server/test/views/questiontypes/DateQuestionRendererTest.java
```diff
@@ -0,0 +1,110 @@
...
+import play.i18n.Lang;
+import play.i18n.Messages;
+import services.LocalizedStrings;
+import services.applicant.ApplicantData;
+import services.applicant.question.ApplicantQuestion;
+import services.question.types.DateQuestionDefinition;
+import services.question.types.QuestionDefinitionConfig;
+import views.questiontypes.ApplicantQuestionRendererParams.ErrorDisplayMode;
+
+public class DateQuestionRendererTest {
```
rockycodes, line 22: I added test files for question types that didn't have them already.

melanie-exygy: Thanks for adding these tests!  Yes, there's some redundancy, but I agree that each question type should have it's own test in case of any differences.

server/app/views/questiontypes/RadioButtonQuestionRenderer.java
```diff
@@ -80,6 +84,7 @@ private DivTag renderRadioOption(
             .withName(selectionPath)
             .withValue(String.valueOf(option.id()))
             .withCondChecked(checked)
+            .condAttr(applicantSelectedQuestion(questionName, questionType), "autofocus", "")
```
melanie-exygy, line 87: Suggestion (non-blocking):  Returning the suggestion you made to me of using `Attr.AUTOFOCUS`.

rockycodes: good catch! i updated it in all the places.

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -219,19 +220,23 @@ private DivTag renderQuestionSummary(AnswerData data, Messages messages, long ap
...
+                      applicantId, data.programId(), data.blockId(), questionName, questionType)
                   .url())
           .setText(messages.at(MessageKey.LINK_EDIT.getKeyName()))
           .setIcon(Icons.EDIT, LinkElement.IconPosition.START);
     } else {
       editElement
           .setHref(
               routes.ApplicantProgramBlocksController.edit(
-                      applicantId, data.programId(), data.blockId())
+                      applicantId, data.programId(), data.blockId(), questionName, questionType)
```
gwendolyngoetz, line 239: Are the links being urlencoded? Question name could have special characters in it.

server/conf/routes
gwendolyngoetz: I wish we could do this without query parameters, but alas..

Why do we need the questionType? The question name should be unique.

gwendolyngoetz (changes requested):

gwendolyngoetz:
I think the focused element needs to have some love. It's very difficult to see which one has the blue border color. If we're going to have it be more dynamic it should be much easier to tell where my cursor is.

image

server/app/views/applicant/ApplicantProgramSummaryView.java
rockycodes: Good question, I'm not sure. I'll look into it.

server/conf/routes
rockycodes: I don't know why I thought questions of different types could have the same name. If that's not true then we don't need it. I'll look into it, thanks for pointing that out!

rockycodes:
@gwendolyngoetz agreed about the focus element needing some more love. Is it alright with you if I make an issue for that and address it in another pr?

server/app/views/questiontypes/DropdownQuestionRenderer.java
```diff
@@ -38,22 +38,24 @@ protected DivTag renderInputTag(
...
-                    .sorted(Comparator.comparing(LocalizedQuestionOption::order))
-                    .map(
-                        option ->
-                            SelectWithLabel.OptionValue.builder()
-                                .setLabel(option.optionText())
-                                .setValue(String.valueOf(option.id()))
-                                .build())
-                    .collect(ImmutableList.toImmutableList()))
-            .setAriaDescribedByIds(ariaDescribedByIds);
+        (SelectWithLabel)
```
melanie-exygy, line 41: (non-blocking) Rather than casting, I'd suggest overriding the method in `SelectWithLabel`, like this other method, to make full use of inheritance:    
```
@Override
  public SelectWithLabel setAriaDescribedByIds(ImmutableList<String> ariaDescribedByIds) {
    super.setAriaDescribedByIds(ariaDescribedByIds);
    return this;
  }
```
  
  You won't have to do this, though, if you remove the `maybeFocusOnInput` method, like I suggest in my other comment.

server/conf/routes
melanie-exygy: I had the same concerns, but also couldn't think of a better way to do it.

server/app/views/components/FieldWithLabel.java
```diff
@@ -370,6 +373,24 @@ public void focusOnError() {
     this.focusOnError = true;
   }
 
+  /**
+   * Determines whether or not a user arrived on the edit page by clicking on a specific question.
+   * If they clicked on a specific question, we set the autofocus to the input for that question.
+   */
+  public FieldWithLabel maybeFocusOnInput(
```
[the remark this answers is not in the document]
rockycodes: Good idea! I'll try that out.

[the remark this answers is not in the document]
rockycodes: done

server/app/views/questiontypes/DropdownQuestionRenderer.java
rockycodes: done

server/app/views/applicant/ApplicantProgramSummaryView.java
rockycodes: Looks like they are url encoded by the Java Play reverse routing. I tried out the admin id `special!` and the url generated was `http://localhost:9000/applicants/264/programs/41/blocks/1/edit?questionName=special%21`. Autofocus worked when clicking through to edit that question.

melanie-exygy (approved):

gwendolyngoetz (approved):

PR #5433, opened 2023-08-15
Allow searching by applicant email in Program Admin application search

### LINKED ISSUE #5324: Filtering admin applications view by applicant email filters by TI email instead
**Describe the bug**
The `ProgramRepository.getApplicationsForAllProgramVersions()` method searches by the email of the TI that submitted the application, not the applicant's email.

The field `submitter_email` in https://github.com/civiform/civiform/blob/2c0526aea1c5ce3c79cc2ff4762b3b0fc1590b32/server/app/repository/ProgramRepository.java#L254 is the email of the TI.

**To Reproduce**

1. Submit an application with a known email
2. In /admin/programs/{program id}/applications, search for the application by email address
3. See that there are no results.

**Additional context**
The tests didn't catch this cause they're manually setting the `submitter_email` on the application to the applicant's email, which is incorrect:
https://github.com/civiform/civiform/blob/2c0526aea1c5ce3c79cc2ff4762b3b0fc1590b32/server/test/repository/ProgramRepositoryTest.java#L358

Changing this map:
https://github.com/civiform/civiform/blob/2c0526aea1c5ce3c79cc2ff4762b3b0fc1590b32/server/test/repository/ProgramRepositoryTest.java#L345
to 
```java
.map(a -> a.getApplicant().getAccount().getEmailAddress())
```
works for most of the tests, except for the test cases on line 299 and 304 (test cases 6 and 9, zero-indexed) from
https://github.com/civiform/civiform/blob/2c0526aea1c5ce3c79cc2ff4762b3b0fc1590b32/server/test/repository/ProgramRepositoryTest.java#L292-L314
because they're searching by email address and looking for a match in the `submitter_email` field, which is no longer set.

### LINKED ISSUE #5413: Search `email_address` on Accounts table when Program Admin search applications
When Program Admin search applications by email, the current logic searches for an entered email in the “submitter_email” column on the Applications table. This field is set to the TI email when a TI submits an application. In addition to searching this field, we want to also search for “email_address” on the Accounts table (e.g. by using [this method](https://github.com/civiform/civiform/blob/db05b35664453a4e6c5927d85f2a9529ad43ed16/server/app/repository/UserRepository.java#L61)). “email_address” on the Accounts table is what is set when a user creates an account.

rockycodes:
### Description

Currently, searching for applications as a Program Admin only returns applications where the email being searched matches the email of the TI who submitted the application on behalf of an applicant. This PR adds the logic to also search the applicant's email on the accounts table which will be filled in if an applicant has created an account and logged in, or if the TI included an email for the applicant when adding them. This PR also adds the ability to search TI and applicant email by a partial string, similar to how name search functionality works now.

#### General

- [x] unit and/or browser tests which fail without the change (if possible)

Applying as a TI
1. Create a program and publish it
2. As a TI, add a client with email address and complete an application for that client for the program
3. As a Program Admin, navigate to review the applications for that program
4. Search by the email address of the client added, you should see the application returned
5. Search by the full or partial email address of the TI, you should see the application returned
6. Search by a bogus email address, you should see no errors and the application should not be returned

Applying as an Applicant
1. Create a program and publish it
2. Log in to/create an account for CiviForm and submit an application for the program
3. As a Program Admin, navigate to review the applications for that program
4. Search by the full or partial email address you used to log in, you should see the application returned
6. Search by a bogus email address, you should see no errors and the application should not be returned

### Issue(s) this completes

Fixes #5324; Fixes #5413

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -258,11 +258,13 @@ public final class ProgramRepository {
       if (search.matches("^\\d+$")) {
         query = query.eq("id", Integer.parseInt(search));
       } else {
+        Optional<Long> applicantId = getApplicantIdByEmail(search);
         String firstNamePath = getApplicationObjectPath(WellKnownPaths.APPLICANT_FIRST_NAME);
         String lastNamePath = getApplicationObjectPath(WellKnownPaths.APPLICANT_LAST_NAME);
         query =
             query
                 .or()
+                .eq("applicant_id", applicantId.isPresent() ? applicantId.get() : null)
                 .ieq("submitter_email", search)
                 .raw(firstNamePath + " || ' ' || " + lastNamePath + " ILIKE ?", "%" + search + "%")
                 .raw(lastNamePath + " || ' ' || " + firstNamePath + " ILIKE ?", "%" + search + "%")
@@ -338,5 +340,15 @@ public final class ProgramRepository {
     result.append("')");
 
     return result.toString();
+  }
+
+  private Optional<Long> getApplicantIdByEmail(String email) {
+    List<Account> accounts =
+        database.find(Account.class).where().ieq("email_address", email).findList();
+    if (accounts.isEmpty()) {
+      return Optional.empty();
+    } else {
+      return Optional.of(accounts.get(0).ownedApplicantIds().get(0));
+    }
   }
 }
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -14,7 +14,6 @@ import io.ebean.Query;
 import io.ebean.SqlRow;
 import io.ebean.Transaction;
 import io.ebean.TxScope;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -238,6 +237,7 @@ public final class ProgramRepository {
         database
             .find(Application.class)
             .fetch("program")
+            .fetch("applicant")
             .orderBy("id desc")
             .where()
             .in("program_id", allProgramVersionsQuery(programId))
@@ -259,13 +259,12 @@ public final class ProgramRepository {
       if (search.matches("^\\d+$")) {
         query = query.eq("id", Integer.parseInt(search));
       } else {
-        List<Long> applicantIds = getApplicantIdsByEmail(search);
         String firstNamePath = getApplicationObjectPath(WellKnownPaths.APPLICANT_FIRST_NAME);
         String lastNamePath = getApplicationObjectPath(WellKnownPaths.APPLICANT_LAST_NAME);
         query =
             query
                 .or()
-                .in("applicant_id", applicantIds)
+                .raw("applicant.account.emailAddress ILIKE ?", "%" + search + "%")
                 .raw("submitter_email ILIKE ?", "%" + search + "%")
                 .raw(firstNamePath + " || ' ' || " + lastNamePath + " ILIKE ?", "%" + search + "%")
                 .raw(lastNamePath + " || ' ' || " + firstNamePath + " ILIKE ?", "%" + search + "%")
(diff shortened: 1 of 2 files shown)


server/app/repository/ProgramRepository.java
```diff
@@ -258,11 +258,13 @@ public PaginationResult<Application> getApplicationsForAllProgramVersions(
       if (search.matches("^\\d+$")) {
         query = query.eq("id", Integer.parseInt(search));
       } else {
+        Optional<Long> applicantId = getApplicantIdByEmail(search);
```
avaleske, line 261: This is going to require the email address to be an exact match. Is that ok? If not, we could join the Applicant table onto the current query and search for partial matches with the rest of the where clauses below.

As written, this also results in an extra query. That's probably fine, but figured I'd call it out.

avaleske (approved):
LGTM assuming the thing I raised in the comment isn't an issue.

server/app/repository/ProgramRepository.java
rockycodes: Doesn't `.ieq("submitter_email", search)` require it to be a perfect match? I think a partial match is better, but I'm not sure how to implement that.

I also think the fewer queries the better, but I'm not sure how to do that since email is stored on the `accounts` table. What do you think?

avaleske: we could join the Applicant table onto this query, right?

rockycodes: Good call, I'll try that.

rockycodes: Argh, I think they won't let me join them because ebean has a rule about [not allowing more than one ](https://ebean.io/docs/query/fetch#rules)`oneToMany` joins. Account is `oneToMany` with Applicant (which [according to this comment](https://github.com/civiform/civiform/blob/4f196c7e83d5fc7fc1c14145161056852128423a/server/app/models/Applicant.java#L28) it actually shouldn't be) and Applicant is `oneToMany` with Applications. Does my reading of this match your understanding?

re: Searching for a partial email match, I tested it out and the current logic for searching my TI email requires it to be a perfect match which seems intentional based on [this test case](https://github.com/civiform/civiform/blob/793e05d7e9e9f0533dd7e68d9925619c822544a4/server/test/repository/ProgramRepositoryTest.java#L291). That said, I think it's a better user experience for email search to mirror name search, so I updated this to search partial strings for applicant emails. I left the current logic that TI emails must be an exact match. What do you think? Is that confusing?

rockycodes: Ok cool, I'll update the TI email search to match too. I agree it makes sense for all search to work the same way.

Doing some further reading on the joins, I think the error I was getting ("invalid fetch path") is actually unrelated to the `toMany` joins. But my searching for it led me to this [StackOverflow](https://stackoverflow.com/questions/51882391/ebean-query-operator-fetch-isnt-working-when-i-use-more-than-one-relationship) where they talk about `fetch` doing multiple queries under the hood when you're joining more than one `toMany` relationship.

rockycodes: I updated search for both TI and applicant emails to be findable by a partial match.

I paired with @nb1701 a bit on how to join the accounts table, and I'm pretty sure we would have to write some raw sql that makes this much less readable. Since this is on the admin side, I'm opting to keep the extra db hit in favor of keeping the code simpler.

rockycodes: Jk I figured out how to do a join with fetch! If at first you don't succeed, write 40 million comments on your own PR about it 😄

avaleske: Amazing! So what was the end result? Is `fetch()` doing multiple queries underneath? It's still easier to read then doing the query separately ourselves.

bion: @avaleske see https://ebean.io/docs/query/aggregation#fetch

Issue #5299, opened 2023-07-25
Turn off Azure tests and Azure staging deploy


rockycodes:
There have been several improvements to the deployment system this year that are AWS specific. Azure staging deployments have been failing for at least 6 months and Azure tests add additional overhead when submitting a PR. We do not currently have any CEs planning to deploy with Azure, so per [a conversation in our eng weekly meeting](https://github.com/civiform/cloud-deploy-infra/pull/220) on 7/24/23, we have decided to turn off Azure tests (except critical ones that would be the most likely to break -- eg. file upload) and turn off the automatic deployment for now.

Work involved:
[Azure tests](https://github.com/civiform/civiform/blob/62903d109cbe537bf4ff3c4149ad94c60ab7d91f/.github/workflows/tests.yaml#L122-L179) - https://github.com/civiform/civiform/pull/5454
[Azure deployment](https://github.com/civiform/civiform-staging-deploy/blob/main/.github/workflows/azure_deploy.yaml) - disabled manually in github
[Update docs](https://docs.civiform.us/it-manual/sre-playbook/initial-deployment) - https://github.com/civiform/docs/pull/372 and updated dev docs with notes eg. https://github.com/civiform/civiform/wiki/Deploying-with-Dev-Azure

PR #5454, opened 2023-08-18
Rockyfine/pare down azure ci tests

rockycodes:
### Description

As a part of the work to [deprecate active Azure support](https://github.com/civiform/civiform/issues/5299), this PR adds a script to only run the Azure tests we think are most likely to break (in this case, file upload) and adds that script to `.github/workflows/tests.yaml` instead of running the full suite of browser tests.

Note: after PR is approved I'll remove the require `run_browser_tests_azure` tests from [here](https://github.com/civiform/civiform/settings/branch_protection_rules/18907456) to allow tests to complete.

#### General

- [x] Performed manual testing (ran script locally)

### DIFF
--- a/bin/run-browser-tests-azure
+++ b/bin/run-browser-tests-azure
@@ -4,7 +4,6 @@
 
 import subprocess
 
-# Select test files we specifically want to test in Azurite
 def get_test_files():
     test_files = ['src/file.test.ts'] # file upload tests
     test_files.sort()
--- a/bin/run-browser-tests-azure
+++ b/bin/run-browser-tests-azure
@@ -4,9 +4,9 @@
 
 import subprocess
 
-# Gets a list of all the test files in the browser-test directory.
+# Select test files we specifically want to test in Azurite
 def get_test_files():
-    test_files = ['src/file.test.ts']
+    test_files = ['src/file.test.ts'] # file upload tests
     test_files.sort()
     return test_files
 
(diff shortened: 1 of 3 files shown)


bin/run-browser-tests-azure
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env python3
+
+# DOC: Run a subset of browser tests that we want to specifically test in Azure in CI mode.
+
+import subprocess
+
+
+def get_test_files():
+    test_files = ['src/file.test.ts']  # file upload tests
+    test_files.sort()
```
gwendolyngoetz, line 10: Do you need this sort?

gwendolyngoetz (commented):
Can you remove the [matrix from the Azure part of the tests_skip.yaml](https://github.com/civiform/civiform/blob/rockyfine/pare-down-azure-ci-tests/.github/workflows/tests_skip.yaml#L31-L37) too so they stay aligned.

bin/run-browser-tests-azure
rockycodes: Definitely not atm since it's just the one file lol. I copied it from https://github.com/civiform/civiform/blob/9ddd1479403e5f11944709fe0bec4ad23e36083a/bin/run-browser-tests-ci-batch#L35 thinking we would want a sort if we had more than one file, but since these likely won't be batched I don't think we need it. What do you think?

gwendolyngoetz: Let's remove it. If we get to the point of having so many tests to run we need batch again probably means it's time to just start running everything.

gwendolyngoetz (changes requested):

bin/run-browser-tests-azure
rockycodes: done

gwendolyngoetz (approved):

PR #5463, opened 2023-08-22
Adding accessiblity to accordion per accessibility audit

### LINKED ISSUE #3560: (WCAG 2.1) Accordion text styling for static text questions is not accessible

**WCAG SC**:  [2.1.1. Keyboard](https://www.w3.org/TR/WCAG21/#keyboard)

**Observed results**: For static text questions, we support some light formatting including an accordion style format See "What are eligibility requirements?" in screenshot). This needs some additional fixes to be accessible:
- It is not focusable or cannot be activated by keyboard only.
- Does not convey role (e.g. button) and state (expanded/collapsed)

**How to fix**:
- Ensure all interactive elements receive keyboard focus or can be activated with the keyboard. Use `<button>` element or if using ARIA, see https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-faq.html for example
- Ensure custom controls provide proper textual name, role, and state information. See https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-faq.html for example

image

melanie-exygy:
### Description

Following WCAG [2.1.1. Keyboard](https://www.w3.org/TR/WCAG21/#keyboard), adding the ability to operate the accordion widget through the keyboard only.  Also, made the accordion focusable.  The accordion now conveys role (e.g. button) and state (expanded/collapsed).  The accordion is used in three places:  static text questions (in the application flow), static text question preview (in admin question creation) and in the API docs help section.

## Release notes

Following WCAG [2.1.1. Keyboard](https://www.w3.org/TR/WCAG21/#keyboard), adding the ability to operate the accordion widget through the keyboard only.  Also, made the accordion focusable.  The accordion now conveys role (e.g. button) and state (expanded/collapsed).  The accordion is used in three places:  static text questions (in the application flow), static text question preview (in admin question creation) and in the API docs help section.

#### General

#### User visible changes

1. From the homepage, click "Apply" on the "Comprehensive Sample Program".
2. On the summary page, click "Continue"
3. Turn a screen reader on.
4. Tab through the first questions page till you get to the "What are the Eligibility Requirements?" accordion.  Once there, note the focus box around the accordion.
5. Also, notice that the screen reader announces that you are on a collapsed button and reads the button text.
6. Click "Enter" or the space-bar to expand or collapse the accordion.

image

### Issue(s) this completes

Fixes #3560


server/app/assets/stylesheets/styles.css
```diff
@@ -29,14 +29,10 @@
     @apply bg-white text-gray-400 border border-gray-400;
   }
 
-  .cf-accordion-visible > .cf-accordion-header > .cf-accordion-button {
+  button[aria-expanded='true'] > .cf-accordion-button {
     @apply rotate-180;
```
gwendolyngoetz, line 33: Does @apply actually do anything? I recall a blog post a while ago that said it was removed. Confirmed by [caniuse](https://caniuse.com/?search=%40apply).

browser-test/image_snapshots/static_text_test/static-text-accordion-open.png
gwendolyngoetz, line 1: Can you add an equal amount of padding to the left of the "Accordion Title" so it doesn't hit up directly against the left side?

melanie-exygy: It's hitting against the side of the focus box, not the container, so I don't think padding would fix that.  I can look into it, though, if you think that it's important for the focus box.

gwendolyngoetz: If it can be done in like 20 minutes and still keep the heading and body text align it would be great. Otherwise don't bother.

server/app/assets/stylesheets/styles.css
melanie-exygy: I didn't know this, but, apparently, this is a [Tailwind annotation](https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply).

browser-test/image_snapshots/static_text_test/static-text-accordion-open.png
melanie-exygy: Sure, will give it a shot.

browser-test/src/api_docs.test.ts
```diff
@@ -102,6 +108,7 @@ describe('Viewing API docs', () => {
     expect(await page.textContent('html')).toContain(
       'Program and version not found',
     )
+    await page.waitForTimeout(300) // ms
```
rockycodes, line 111: question (non-blocking): did you add the additional `waitForTimeout` calls because of the issue with inconsistent screenshots commented on below?

browser-test/src/api_docs.test.ts
```diff
@@ -122,16 +129,6 @@ describe('Viewing API docs', () => {
     // causing inconsistent screenshots.
     await page.waitForTimeout(300) // ms
 
-    // Wait for the specific text to become visible in the accordion.
```
rockycodes, line 125: question (non-blocking): why did you remove this?

browser-test/src/static_text.test.ts
```diff
@@ -41,6 +57,24 @@ describe('Static text question for applicant flow', () => {
...
+  it('displays the accordion', async () => {
+    const {page, applicantQuestions} = ctx
+    await applicantQuestions.applyProgram(programName)
+
+    expect(await page.innerHTML('.cf-applicant-question-text')).toContain(
+      'cf-accordion',
+    )
+  })
+
+  it('expands accordion when accordion header clicked', async () => {
```
rockycodes, line 69: suggestion (non-blocking): combine this test with `validate screenshot when accordion open`

server/app/assets/javascripts/accordion.ts
```diff
@@ -1,37 +1,74 @@
...
+  }
+
+  onBlur() {
+    this.buttonNode.classList.remove('focus')
+  }
 }
 
 export function init() {
-  new AccordionController()
+  const disclosureButtons: NodeListOf<Element> = document.querySelectorAll(
```
rockycodes, line 67: question(non-blocking): Just curious why you restructured it this way instead of keeping the exported `init` function only responsible for creating the instance. There seems to be a pattern in the other files of keeping the exported `init` function very minimal, but I don't have a strong preference about this.

server/test/views/components/AccordionTest.java
```diff
@@ -0,0 +1,43 @@
+package views.components;
```
rockycodes, line 1: praise: thanks for adding a test!

rockycodes (approved):
lgtm! just a couple of non-blocking questions + suggestions. feel free to ping me again if you make changes and want me to look at them.

browser-test/src/api_docs.test.ts
melanie-exygy: Yes, specifically because of the open/close caret icon taking a few millis to rotate.

browser-test/src/api_docs.test.ts
melanie-exygy: With the new way of implementing the accordion, there is no longer a `cf-accordion-visible` CSS class and the text in the content div is always present.  What changes is the `display` styling (going from `none` to `block`).  I couldn't figure out how to evaluate whether that styling was present, plus we're already testing that the content is displayed in the screenshot validation.  I can give it another try if you feel that the screenshot validation isn't enough.

browser-test/src/static_text.test.ts
melanie-exygy: Good idea.  Doing that now.

server/app/assets/javascripts/accordion.ts
melanie-exygy: The idea was to only initialize this class when accordion functionality is needed, which can slightly improve page load performance.

browser-test/image_snapshots/static_text_test/static-text-accordion-open.png
melanie-exygy: Turns out that it was an easy fix.  Added the padding.

browser-test/src/api_docs.test.ts
rockycodes: Gotcha. Thanks for explaining! I think the screenshot is enough since it will change if the text changes.

server/app/assets/javascripts/accordion.ts
rockycodes: Nice! Good idea.

browser-test/src/static_text.test.ts
```diff
@@ -41,6 +58,15 @@ describe('Static text question for applicant flow', () => {
     await applicantQuestions.seeStaticQuestion(staticText)
   })
 
+  it('displays the accordion', async () => {
```
rockycodes, line 61: suggestion (non-blocking): Sorry I missed this before, but you could also combine this test with `validate screenshot when accordion closed` which might be nice. One test for accordion open, one test for accordion closed.

melanie-exygy: Sure, will change that.

PR #5491, opened 2023-08-30
enable applicants to download and print completed applications

### LINKED ISSUE #4506: Enable applicants to download and print completed applications
A group of Trusted Intermediaries mentioned that it's helpful to print the completed program application for a resident/client to take with them. Residents often want the physical form to take with them. They mentioned it can be helpful if the residents seeks services from multiple CBOs, they can take their application to the next TI.

An option on the application confirmation page that allows the applicant (TI or resident) to download then print the completed application.

**Additional context**
It would be helpful if it included all completed questions as well as a printout of any uploaded documents. There may be data privacy concerns with the email option, would want to check with Privacy on this. 

**Done when**
Resident or TI can completed application. 

image

zarifzm:
### Description

Download button added to application confirmation page, allowing user to get a PDF of the application contents.

### Release notes

APPLICATION_EXPORTABLE(default = false) as a new feature flag that enables this feature.
- Download button visibility(true = visible)
- Download function in UpsellController(true = enabled)

#### General

#### User visible changes

- [x] Followed steps to [internationalize new strings](https://docs.civiform.us/contributor-guide/developer-guide/internationalization-i18n#internationalization-for-application-strings)

Complete an application and click on the 'Download' button to generate a PDF of the application contents.

### Issue(s) this completes

Fixes #4506


server/app/services/applications/PdfExporterService.java
```diff
@@ -0,0 +1,51 @@
...
+    checkNotNull(configuration);
+    checkNotNull(deploymentType);
+  }
+
+  public PdfExporter.InMemoryPdf generatePdf(Long applicationId, ProgramDefinition program) {
+    Optional<Application> applicationMaybe =
+        applicationService.getApplication(applicationId, program);
+    if (applicationMaybe.isEmpty()) {
+      throw new NoSuchElementException(
+          String.format("Application %d does not exist.", applicationId));
```
bion, line 40: **suggestion (blocking):** if an application isn't found it's because the applicationId is invalid, which is a responsibility of the request meaning the error should result in a 400. Throwing this uncaught exception will produce a 500, however. Instead throw a typed exception, catch it in the controller, and throw a `BadRequestException` which will result in a 400.

server/app/services/applications/PdfExporterService.java
```diff
@@ -0,0 +1,51 @@
...
+      DeploymentType deploymentType,
+      PdfExporter pdfExporter) {
+    this.applicationService = checkNotNull(applicationService);
+    this.pdfExporter = checkNotNull(pdfExporter);
+
+    checkNotNull(configuration);
+    checkNotNull(deploymentType);
+  }
+
+  public PdfExporter.InMemoryPdf generatePdf(Long applicationId, ProgramDefinition program) {
```
bion, line 35: **suggestion (blocking):** pass the application object in instead of the ID, this class should not be responsible for retrieving data from the DB

bion (changes requested):

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.typesafe.config.Config;
+import java.util.Optional;
+import javax.inject.Inject;
+import models.Application;
+import repository.ApplicationRepository;
+import services.DeploymentType;
+import services.program.ProgramDefinition;
+
+// NON_ABSTRACT_CLASS_ALLOWS_SUBCLASSING ApplicationService
```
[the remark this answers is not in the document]
zarifzm: https://docs.google.com/document/d/1nc8yrtWRN5hARsxKdJPwHlaCJrjYIp-4sn77Iiy_vjw/edit?disco=AAAA3lhJMGs

The getApplication(..) is moved from ProgramAdminApplicationService as the functionality will be used in multiple user contexts. The ProgramAdminApplicationService extends ApplicationService; therefore, ApplicationService cannot be a final class.

browser-test/src/support/applicant_questions.ts
```diff
@@ -493,6 +502,14 @@ export class ApplicantQuestions {
     await this.clickSubmit()
   }
 
+  async downloadFromConfirmationPage() {
+    // Assert that we're on the review page.
+    await this.expectConfirmationPage()
+
+    // Click on download button.
+    await this.clickDownload()
```
bion, line 510: **suggestion (blocking):** this clicks the download button but makes no assertion that the download succeeds. Please add that assertion using the playwright Download API: https://playwright.dev/docs/downloads

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +177,42 @@ public CompletionStage<Result> considerRegister(
...
+        return CompletableFuture.completedFuture(unauthorized("Invalid credentials"));
+      }
+    } else {
+      Account currentAccount = profile.getAccount().join();
+      if (currentAccount.getManagedByGroup().isEmpty()) {
+        return CompletableFuture.completedFuture(unauthorized("Invalid credentials"));
+      }
+    }
+    Optional<Application> applicationMaybe =
+        applicationService.getApplication(applicationId, program);
```
bion, line 204: **suggestion (blocking):** there are several places in this code that are making still making synchronous database calls. Please rewrite using async calls to access the DB.

https://github.com/civiform/civiform/pull/5491#discussion_r1316393352

server/app/services/applications/ApplicationService.java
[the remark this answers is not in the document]
bion: In that case I'd recommend composition over inheritance and have `ProgramAdminApplicationService` inject `ApplicationService` as a dependency.

bion (changes requested):

server/app/controllers/applicant/UpsellController.java
```diff
@@ -200,13 +201,14 @@ public CompletionStage<Result> download(
         return CompletableFuture.completedFuture(unauthorized("Invalid credentials"));
       }
     }
-    Optional<Application> applicationMaybe =
-        applicationService.getApplication(applicationId, program);
-    if (applicationMaybe.isEmpty()) {
+    CompletableFuture<Optional<Application>> applicationMaybe =
+        applicationService.getApplication(applicationId, program.join()).toCompletableFuture();
```
bion, line 205: **issue (blocking):** adding `join()` here is making this previously async code blocking

bion (changes requested):
@zarifzm let's find a time to chat about async vs blocking IO. There are several places in this PR that are doing blocking IO in places that shouldn't to be in compliance with our development standards.

server/app/controllers/applicant/UpsellController.java
swatkat1: moved the join to the service class. We do need to check the application if it has the right program name.

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,54 @@
...
+  public Optional<Application> getApplication(
+      Optional<Application> application, ProgramDefinition program) {
+    if (application.isEmpty()
+        || !application
+            .get()
+            .getProgram()
+            .getProgramDefinition()
+            .adminName()
+            .equals(program.adminName())) return Optional.empty();
+    else return application;
```
bion, line 36: **suggestion (blocking):** Please review the Java style guide: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used
```suggestion
    if (application.isEmpty()
        || !application
            .get()
            .getProgram()
            .getProgramDefinition()
            .adminName()
            .equals(program.adminName())) {
      return Optional.empty();
    } else {
      return application;
    }
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -813,7 +813,7 @@ public ImmutableList<Application> getApplications(TimeFilter submitTimeFilter) {
    */
   public CompletionStage<ApplicationPrograms> relevantProgramsForApplicant(
       long applicantId, CiviFormProfile requesterProfile) {
-    // Note: The Program model associated with the application is eagerly loaded.
+    // Note: The Program model associated with the application is eagerly loaded.F
```
bion, line 816: **nitpick:** fix typo

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,54 @@
...
+  private final HttpExecutionContext httpExecutionContext;
+
+  @Inject
+  ApplicationService(
+      ApplicationRepository applicationRepository, HttpExecutionContext httpExecutionContext) {
+    this.applicationRepository = checkNotNull(applicationRepository);
+    this.httpExecutionContext = checkNotNull(httpExecutionContext);
+  }
+
+  public Optional<Application> getApplication(
```
bion, line 27: **suggestion (blocking):** This method is named `getApplication` but its implementation is limited to checking if the application is associated with the program passed in. This is validation logic that should instead be performed wherever this check is necessary.

If you would like a convenience method for retrieving the program name of an application I recommend adding it to `Application` e.g.

```java
  public String getProgramName() {
    return getProgram().getProgramDefinition().adminName();
  }
```

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,54 @@
...
+      long applicationId, CompletableFuture<ProgramDefinition> program) {
+    CompletableFuture<Optional<Application>> maybeApplication =
+        applicationRepository.getApplication(applicationId).toCompletableFuture();
+    return CompletableFuture.allOf(maybeApplication, program)
+        .thenApplyAsync(
+            v -> {
+              return getApplication(maybeApplication.join(), program.join());
+            },
+            httpExecutionContext.current());
+  }
```
bion, line 53: **suggestion (blocking):** Also here, recommend moving the validation logic out of this service

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -239,4 +221,9 @@ public Optional<String> getNote(Application application) {
         .findFirst()
         .map(app -> app.getDetails().noteEvent().get().note());
   }
+
+  public Optional<Application> getApplication(long applicationId, ProgramDefinition program) {
```
bion, line 225: **suggestion (blocking):** missing javadoc for this public method

bion (changes requested):

server/app/services/applications/ApplicationService.java
zarifzm: Renamed method to `validateApplication`.

Both the applicant scenario and the admin client scenario have a check for if the application has the right program name, so I believe it should stay in the service.

server/app/services/applications/ApplicationService.java
zarifzm: Both the applicant scenario and the admin client scenario have a check for if the application has the right program name, so I believe it should stay in the service.

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +177,56 @@ public CompletionStage<Result> considerRegister(
...
+            .toCompletableFuture();
+
+    CompletableFuture<Optional<Application>> applicationMaybe =
+        applicationService.getApplicationAsync(applicationId, program).toCompletableFuture();
+
+    return CompletableFuture.allOf(applicationMaybe, program, account)
+        .thenApplyAsync(
+            check -> {
+              PdfExporter.InMemoryPdf pdf =
+                  pdfExporterService.generatePdf(applicationMaybe.join().get());
```
bion, line 210: **suggestion (blocking):** There is a security vulnerability here where someone signed in as an applicant could make a request to this endpoint with an `applicationId` for an application that isn't theirs. Add a check that ensures the account requesting the application has permission to read it.

bion (changes requested):
Unit tests are failing

bion:
@zarifzm looks like this is missing a feature flag for the release

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +177,64 @@ public CompletionStage<Result> considerRegister(
...
+        profileMaybe.orElseThrow(
+            () -> new NoSuchElementException("User authorized as applicant but no profile found"));
+
+    CompletableFuture<ApplicantPersonalInfo> applicantPersonalInfo =
+        applicantService.getPersonalInfo(applicantId).toCompletableFuture();
+
+    CompletableFuture<Account> account =
+        applicantPersonalInfo
+            .thenComposeAsync(
+                v -> checkApplicantAuthorization(request, applicantId), httpContext.current())
```
bion, line 199: **suggestion (blocking):** there is still a security vulnerability here because it's using the user-provided `applicantId`. Instead use the applicant ID from the `account` resolve via their `profile` i.e. their session cookie.

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,54 @@
...
+  private final HttpExecutionContext httpExecutionContext;
+
+  @Inject
+  ApplicationService(
+      ApplicationRepository applicationRepository, HttpExecutionContext httpExecutionContext) {
+    this.applicationRepository = checkNotNull(applicationRepository);
+    this.httpExecutionContext = checkNotNull(httpExecutionContext);
+  }
+
+  /** Validates the given application is associated with the given program. */
```
bion, line 27: **question (blocking):** why are we validating that the application is associated with a given program?

bion (changes requested):

server/app/controllers/applicant/UpsellController.java
swatkat1: We cannot use this approach as it won't work in TI path. When a TI is applying on behalf of a client, the Profile is still TI, but the applicant Id belongs to the Client.

bion: Gotcha, in the case the `applicantId` provided in the URL needs to be validated as either the current user's applicant ID or as associated with an account that is managed by the current user's TI org.

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +177,64 @@ public CompletionStage<Result> considerRegister(
...
+    CiviFormProfile profile =
+        profileMaybe.orElseThrow(
+            () -> new NoSuchElementException("User authorized as applicant but no profile found"));
+
+    CompletableFuture<ApplicantPersonalInfo> applicantPersonalInfo =
+        applicantService.getPersonalInfo(applicantId).toCompletableFuture();
+
+    CompletableFuture<Account> account =
+        applicantPersonalInfo
+            .thenComposeAsync(
```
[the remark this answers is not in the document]
zarifzm: Good catch. Removed

server/app/controllers/applicant/UpsellController.java
swatkat1: I believe the checkApplicantAuthorization(applicantId) checks the same. @zarifzm  - Please confirm

server/app/services/applications/ApplicationService.java
zarifzm: https://github.com/civiform/civiform/pull/3180

server/app/controllers/applicant/UpsellController.java
swatkat1: I think the checkApplicantAuthorization does it. @zarifzm - please confirm

zarifzm: It does. https://github.com/civiform/civiform/blob/main/server/app/auth/CiviFormProfile.java#L203

server/app/services/applications/ApplicationService.java
zarifzm: "Preventing program admins from accessing applications that aren't part of the provided program" (https://github.com/civiform/civiform/pull/3180)

I separated out the validation logic for the applicant/TI client scenario and the program admin scenario

server/conf/helper/feature-flags.conf
```diff
@@ -30,3 +30,7 @@ intake_form_enabled = ${?INTAKE_FORM_ENABLED}
 # api generated docs tab
 api_generated_docs_enabled = false
 api_generated_docs_enabled = ${?API_GENERATED_DOCS_ENABLED}
+
+# completed application exportable
+application_exportable = true
```
swatkat1, line 35: @zarifzm  -  the feature flags are turned off by default.  We should turn it on env. basis. Example - first turn it on in staging env, test it out, then turn it on in prod.

browser-test/src/support/applicant_questions.ts
```diff
@@ -323,6 +323,18 @@ export class ApplicantQuestions {
...
   }
 
+  async clickDownload() {
+    const [downloadEvent] = await Promise.all([
+      this.page.waitForEvent('download'),
+      this.page.click('text="Download"'),
+    ])
+    const path = await downloadEvent.path()
+    if (path === null) {
+      throw new Error('download failed')
```
swatkat1, line 333: Can we add another case to check the length of file >0

browser-test/image_snapshots/application_navigation_test/program-submission-guest.png
bion, line 1: **issue (blocking):** The download icon isn't vertically aligned with the text. Also it's a little odd to have the icon for only one of the buttons.

browser-test/src/support/applicant_questions.ts
```diff
@@ -323,6 +323,18 @@ export class ApplicantQuestions {
     await waitForPageJsLoad(this.page)
   }
 
+  async clickDownload() {
+    const [downloadEvent] = await Promise.all([
+      this.page.waitForEvent('download'),
+      this.page.click('text="Download"'),
+    ])
+    const path = await downloadEvent.path()
+    if (path === null || readFileSync(path, 'utf8').length == 0) {
```
bion, line 332: **suggestion (blocking):** https://google.github.io/styleguide/jsguide.html#features-equality-checks
```suggestion
    if (path === null || readFileSync(path, 'utf8').length === 0) {
```

server/app/models/Application.java
```diff
@@ -93,6 +93,10 @@ public Program getProgram() {
     return this.program;
   }
 
+  public String getProgramName() {
```
bion, line 96: **suggestion (blocking):** add javadoc for this new public method. All public methods should have javadoc, though some are still missing in the codebase.

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +181,66 @@ public CompletionStage<Result> considerRegister(
...
+              if (!profile.isTrustedIntermediary()) {
+                if (!applicationMaybe
+                    .join()
+                    .get()
+                    .getApplicant()
+                    .getAccount()
+                    .equals(account.join())) {
+                  return unauthorized();
+                }
+              }
```
bion, line 222: **question (blocking):** You've already checked authorization on L203, no?

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,49 @@
...
+  ApplicationService(
+      ApplicationRepository applicationRepository, HttpExecutionContext httpExecutionContext) {
+    this.applicationRepository = checkNotNull(applicationRepository);
+    this.httpExecutionContext = checkNotNull(httpExecutionContext);
+  }
+
+  /** Validates that the given application is not empty or without a program name. */
+  public Optional<Application> validateApplication(Optional<Application> application) {
+    if (application.isEmpty() || application.get().getProgramName().isEmpty()) {
+      return Optional.empty();
```
bion, line 29: **suggestion (blocking):** AFAIK the system never creates applications that are not associated with a program, I don't think this method is necessary?

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -239,4 +217,24 @@ public Optional<String> getNote(Application application) {
...
+  /** Validates that the given application is part of the given program. */
+  private Optional<Application> validateProgram(
+      Optional<Application> application, ProgramDefinition program) {
+    if (application.isEmpty()
+        || application.get().getProgramName().isEmpty()
+        || !application.get().getProgramName().equals(program.adminName())) {
+      return Optional.empty();
+    }
+    return application;
+  }
```
bion, line 239: **issue (blocking):** Returning an empty optional is not an obvious way of indicating that a given piece of data or a given object is invalid. Better alternatives would be to either throw an exception or return an error object such as https://github.com/civiform/civiform/blob/main/server/app/services/ErrorAnd.java

bion (changes requested):

server/app/models/Application.java
```diff
@@ -93,6 +93,11 @@ public Program getProgram() {
     return this.program;
   }
 
+  /** Returns the program's admin name. */
```
bion, line 96: **suggestion:** 
```suggestion
  /** Returns the admin name of the program this application is associated with. */
```

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,40 @@
...
+  ApplicationService(
+      ApplicationRepository applicationRepository, HttpExecutionContext httpExecutionContext) {
+    this.applicationRepository = checkNotNull(applicationRepository);
+    this.httpExecutionContext = checkNotNull(httpExecutionContext);
+  }
+
+  /**
+   * Retrieves the application with the given ID and validates that it is associated with the given
+   * program.
+   */
```
bion, line 29: **issue (blocking):** this javadoc is no longer accurate

server/app/services/applications/ApplicationService.java
```diff
@@ -0,0 +1,40 @@
...
+  public CompletionStage<Optional<Application>> getApplicationAsync(long applicationId) {
+    CompletableFuture<Optional<Application>> maybeApplication =
+        applicationRepository.getApplication(applicationId).toCompletableFuture();
+    return CompletableFuture.allOf(maybeApplication)
+        .thenApplyAsync(
+            v -> {
+              return maybeApplication.join();
+            },
+            httpExecutionContext.current());
```
bion, line 38: **suggestion (blocking):** the body of this function is overcomplicated for what it's doing. Can it it be replaced with `return applicationRepository.getApplication(applicationId);`?

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +181,56 @@ public CompletionStage<Result> considerRegister(
...
+    CiviFormProfile profile =
+        profileMaybe.orElseThrow(
+            () -> new NoSuchElementException("User authorized as applicant but no profile found"));
+
+    CompletableFuture<Account> account =
+        program
+            .thenComposeAsync(
+                v -> checkApplicantAuthorization(request, applicantId), httpContext.current())
+            .thenComposeAsync(v -> profile.getAccount(), httpContext.current())
+            .toCompletableFuture();
```
bion, line 205: **suggestion (blocking):** 

1. the `account` is not actually needed, it's only consumed by the `allOf` below but then its value is never read after that.
2. there's no reason to wait on the `program` completion stage resolving before making the call to `checkApplicantAuthorization`

bion (changes requested):
@zarifzm looks like several of the browser tests are failing. Please check in with @sijiayam to ensure the UI changes in this work match their expectations.

server/app/controllers/applicant/UpsellController.java
```diff
@@ -165,4 +180,45 @@ public CompletionStage<Result> considerRegister(
               throw new RuntimeException(ex);
             });
   }
+
+  /** Download a PDF file of the application to the program. */
+  @Secure
+  public CompletionStage<Result> download(
```
bion, line 186: **suggestion (blocking):** Please add new test cases in https://github.com/civiform/civiform/blob/main/server/test/controllers/applicant/UpsellControllerTest.java for this new method including a test case with an authenticated applicant or TI without permission to view the application they're attempting to download.

bion (changes requested):
getting close! missing unit testing for the new controller action

bion (changes requested):

swatkat1 (approved):
I think you got the overall changes done. Well done Zarif!

bion (approved):
Good work!

server/app/views/applicant/ApplicantUpsellCreateAccountView.java
```diff
@@ -54,24 +60,35 @@ public Content render(
...
                 /* bypassMessage= */ MessageKey.BUTTON_CONTINUE_WITHOUT_AN_ACCOUNT)
             .build();
-
+    ATag downloadButton = new ATag();
+    if (settingsManifest.getApplicationExportable(request)) {
+      downloadButton =
+          new ATag()
+              .withHref(redirectUrl)
+              .with(
+                  button("Download PDF")
```
bion, line 77: **suggestion (blocking):** this hard-coded UI string (and any others added in this PR) need to be added to the i18n system

bion (changes requested):

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -239,4 +218,30 @@ public Optional<String> getNote(Application application) {
         .findFirst()
         .map(app -> app.getDetails().noteEvent().get().note());
   }
+
+  /**
+   * Retrieves the application with the given ID and validates that it is associated with the given
+   * program.
+   */
+  public Optional<Application> getApplication(long applicationId, ProgramDefinition program) {
```
bion, line 226: **question (blocking):** where is this new method used?

bion (changes requested):

server/app/services/applications/ProgramAdminApplicationService.java
bion: Ah I see, it's used by existing code.

bion (approved):

PR #5493, opened 2023-08-30
Remove database call to get question data on index page if not needed

dkatzz:
### Description

We look up question data on the program admin page in order to determine eligibility status, but we don't need to do this if the user hasn't started any applications.

This reduces the number of question calls from 1x per program to 1x total.

There are further improvements we can make, but this is the first step.

This change also removes a call to get applicant info, which doesn't seem to be used.

#### General

### Issue(s) this completes

#5514

### DIFF
--- a/server/app/services/applicant/ApplicantService.java
+++ b/server/app/services/applicant/ApplicantService.java
@@ -819,32 +819,30 @@ public final class ApplicantService {
             .getApplicationsForApplicant(
                 applicantId, ImmutableSet.of(LifecycleStage.DRAFT, LifecycleStage.ACTIVE))
             .toCompletableFuture();
-    CompletableFuture<ImmutableList<ProgramDefinition>> activeProgramDefinitionsFuture =
-        userRepository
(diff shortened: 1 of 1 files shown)


server/app/services/applicant/ApplicantService.java
```diff
@@ -819,32 +819,30 @@ public CompletionStage<ApplicationPrograms> relevantProgramsForApplicant(
             .getApplicationsForApplicant(
                 applicantId, ImmutableSet.of(LifecycleStage.DRAFT, LifecycleStage.ACTIVE))
             .toCompletableFuture();
-    CompletableFuture<ImmutableList<ProgramDefinition>> activeProgramDefinitionsFuture =
-        userRepository
```
dkatzz, line 823: **thought:** Doesn't seem like we are doing anything with the applicant info here, so should be able to remove this? Unless I'm missing something?

dkatzz: Great, thank you!

PR #5552, opened 2023-09-07
Tooltip on the "Administrative Identifier" field in the question creation form

### LINKED ISSUE #5329: Explain question "admin name" is used for API export, on question creation view
It would be helpful to CiviForm admins to know that the "admin name" for a question is what it will be called in the API export.

This is done when:
- There's help text on the question creation page that explains what the admin name is used for
- The question creation page shows a preview of how the admin name will be formatted in the api response

melanie-exygy:
### Description

Added a tooltip on the "Administrative Identifier" field in the question creation form.  This tooltip clarifies how the identifier, aka question name, will appear in the API response.

#### General

#### User visible changes

1. As a CF admin, go to the Questions tab.
2. Click on "Create new Question".
3. Hover over the "!" icon next to the "Administrative Identifier" field label.
4. Note the following tooltip message.
image

### Issue(s) this completes

Fixes #5329


server/app/views/admin/questions/QuestionEditView.java
```diff
@@ -521,6 +522,12 @@ private DivTag administrativeNameField(String adminName, boolean editExistingQue
         .setId("question-name-input")
         .setFieldName(QUESTION_NAME_FIELD)
         .setLabelText("Administrative identifier. This value can't be changed later")
+        .setToolTipText(
```
avaleske, line 525: Do we also want to note that hyphens will be stripped?

server/app/views/admin/questions/QuestionEditView.java
```diff
@@ -521,6 +522,12 @@ private DivTag administrativeNameField(String adminName, boolean editExistingQue
...
+            "This is the name with which the question will be identified in the API export.  "
```
avaleske, line 526: suggestion (optional): "This will be used to identify questions in the API and CSV export. It will be formatted..."

reason: it's slightly more casual language

avaleske (approved):
This is great, thanks!

I see your point about the live preview being more work. I was thinking it could go directly under the input in a single line like "Visible in the API as `question_name`" but since that'd require frontend js it can be a separate PR if we decide to do it.

If it's not going to be a fast-follow, though, can we make a new issue for it?

server/app/views/admin/questions/QuestionEditView.java
melanie-exygy: Good idea.  I like that better.  Will change the wording.

server/app/views/admin/questions/QuestionEditView.java
melanie-exygy: Yes, good catch.  I'll add that.

melanie-exygy:
> This is great, thanks!
> 
> I see your point about the live preview being more work. I was thinking it could go directly under the input in a single line like "Visible in the API as `question_name`" but since that'd require frontend js it can be a separate PR if we decide to do it.
> 
> If it's not going to be a fast-follow, though, can we make a new issue for it?

I've created the following issue for the live preview:  #5556.  Feel free to edit the description or add tags if needed.

PR #5596, opened 2023-09-12
Show preview of formatted question name in the admin create question form

### LINKED ISSUE #5556: Live preview of question name after formatting
Admin users may not know how the "Administrative identifier" on a new question will appear in API and CSV exports.  We have a tooltip which gives an example of formatting, but a live preview of the formatted question name could help make it even clearer.

On new question creation, the live preview would appear directly under the input in a single line which says: "Visible in the API as <question_name>".  The formatted <question_name> would be updated based on user input.

**Additional context**
This issue is a follow-up to #5329 and this PR:  [5552](https://github.com/civiform/civiform/pull/5552).

**Done when**
A live preview of the formatted question name appears under the "Administrative identifier" input box in the new question form.

melanie-exygy:
### Description

When a CF admin creates a new question, they must enter an "Administrative Identifier" (aka question name) for the question.  The admin now sees a live preview of how the identifier will appear in the API.  The preview is below the input field and shows the formatted version of the identifier.

image

## Release notes

When a CF admin creates a new question, they must enter an "Administrative Identifier" (aka question name) for the question.  The admin now sees a live preview of how the identifier will appear in the API.  The preview is below the input field and shows the formatted version of the identifier.

#### General

#### User visible changes

1.  As an admin, go to the "Questions" tab and click on the "Create new question" button.
2. Select any type of question.
3. Start typing a question name in the "Administrative Identifier" input field.
4. Note the preview below.

### Issue(s) this completes

Fixes #5556


gwendolyngoetz (commented):
Is there something specific driving this? I really don't like that the name formatting logic is getting duplicated into the client.

I'm also concerned that someone may type in the name, copy down the api formatted value, and not notice when they get told that a particular name is in use. This shows as a toast message when saving. And it looks like saving would wipe out the preview.

The api version of the name is actually never shown anywhere. I want to say there's a bug on it, but can't recall for certain. The result from `questionForm.getQuestionName()` isn't the value used for the api. That's actually going to be on  `QuestionDefinition`'s `getQuestionNameKey` method.

It doesn't show the api name on the question list or detail screen. Take a look at `http://localhost:9000/api/docs/v1` to see an example of api name being used.

My suggestion would be to cut the live preview since it may not be accurate upon saving. Keep the `Visible in the api` label, I like that. And have it display the api name. I would also put the api label into the Question list page.

melanie-exygy:
> Is there something specific driving this? I really don't like that the name formatting logic is getting duplicated into the client.

This was requested by Arkansas, I believe. The preview was requested as part of this issue:  #5329.  I created #5556 so that I could address the client-side changes in a separate issue.  @avaleske is it accurate that Arkansas requested this?
I also don't like that the formatting logic is duplicated in the client.  I can't think of a way to avoid this if we do decide to preview given that it's dynamic. 
> 
> I'm also concerned that someone may type in the name, copy down the api formatted value, and not notice when they get told that a particular name is in use. This shows as a toast message when saving. And it looks like saving would wipe out the preview.

If the name is in use, wouldn't they have to go back to the input field and type in a different name?  If so, they would see that the preview is changing as they type, no?
> 
> The api version of the name is actually never shown anywhere. I want to say there's a bug on it, but can't recall for certain. The result from `questionForm.getQuestionName()` isn't the value used for the api. That's actually going to be on `QuestionDefinition`'s `getQuestionNameKey` method.

I think I copied the formatting from `getQuestionNameKey`.  Was there something I missed?
> 
> It doesn't show the api name on the question list or detail screen. Take a look at `http://localhost:9000/api/docs/v1` to see an example of api name being used.

Yes, I noticed that.  I double-checked that what the preview shows is what is being shown in the example api response in the api docs.  Are you saying that the wording needs to be clearer?  Maybe `Visible in the api response as: ` ?
> 
> My suggestion would be to cut the live preview since it may not be accurate upon saving. Keep the `Visible in the api` label, I like that. And have it display the api name. I would also put the api label into the Question list page.

Good idea to add it to the Question List page.  As for not showing the preview, maybe we should discuss with Austin since he's more familiar with what Arkansas is asking for?

avaleske:
My intention for this had been to help admins deal with that fact that the api name doesn't format as one might expect. That is:
- underscores and dashes are cleared instead of being preserved in the api name
- spaces become underscores

I thought this could result in someone naming a question "applicant-primary-car-purchase-date", expecting it to look that way in the API, and instead getting a the difficult to work with API name of `applicantprimarycarpurchasedate`.

For Arkansas specifically:
- They didn't request it directly, but they had expressed frustration with the formatting of the api question names and shared that they had reverse engineered how to get them to format the way they wanted in the API.
- Since the name is not changeable after saving the question, I wanted to show up-front what it might look like.

That said, if there's a better way to help admin pick good names, I'm all for it. Would it be helpful to have Product/UX take a pass at this problem?

Re other points that were raised:

>> I'm also concerned that someone may type in the name, copy down the api formatted value, and not notice when they get told that a particular name is in use. This shows as a toast message when saving. And it looks like saving would wipe out the preview.

> If the name is in use, wouldn't they have to go back to the input field and type in a different name? If so, they would see that the preview is changing as they type, no?

@gwendolyngoetz, would triggering the preview on pageload, not just on keystrokes, fix this? I figured admins would notice the form failed to submit, and see the preview update when they pick a new question name.

>>  I really don't like that the name formatting logic is getting duplicated into the client.

I don't love that either, but in this case, it probably fine since this regex can't change without breaking a lot of things. We should leave a comment in both places noting the other place it's used, though.

gwendolyngoetz:
Alright, all good points. I'd be okay with approving this if we do the follow:
- (blocking) Add a comment at both places noting the duplicate logic as Austin suggests
- (optional) Re-populate the API name if it turns out that the name already exists
- (blocking) Add the API name to the question list and detail. It's usually not when I'm making it question that I need to know the API name, but when I go back to find it later. This can be a separate PR if that's easier.

How does that sound?

melanie-exygy:
> Alright, all good points. I'd be okay with approving this if we do the follow:
> 
> * (blocking) Add a comment at both places noting the duplicate logic as Austin suggests
> * (optional) Re-populate the API name if it turns out that the name already exists
> * (blocking) Add the API name to the question list and detail. It's usually not when I'm making a question that I need to know the API name, but when I go back to find it later. This can be a separate PR if that's easier.
> 
> How does that sound?

That sounds like a good plan to me.  I've asked @isabelcshaw and @sijiayam to look at this and give their guidance.  I'll wait to see what they say before changing anything.  

Just to clarify, when you suggest re-populating the API name if it already exists, do you mean clearing the input field (which would also clear the preview)?  If so, I wonder if users might forget exactly what they had tried before and end up trying it again.  By leaving their previous attempt in the input field, they can slightly modify it, making it unique.  As they modify, the preview will be updated below.  It updates on keystrokes.

gwendolyngoetz:
> Just to clarify, when you suggest re-populating the API name if it already exists, do you mean clearing the input field (which would also clear the preview)?

No, not the name textbox that should stay as it. I mean re-populate what was added previous in the preview label.

melanie-exygy:
> > Just to clarify, when you suggest re-populating the API name if it already exists, do you mean clearing the input field (which would also clear the preview)?
> 
> No, not the name textbox that should stay as it. I mean re-populate what was added previous in the preview label.

Got it.  Yes, I agree.  The preview should re-populate.

isabelcshaw:
Chatted with @melanie-exygy (thank you!) and this approach sounds good to me. 

The one ask is to pull out adding the API name to the question bank list view. Definitely agree that it should be on the question detail / preview page (as it appears when you create the question). Just a little hesitant to add more info to those question components unless it is really, consistently high value. Open to pushback here! 

For the moment, we'll make a new issue to split that out and pull in Sijia.

gwendolyngoetz:
It will definitely make it more cumbersome if you are looking for the api name for multiple questions, but I'm fine deferring to it on the list screen for now.

server/app/assets/javascripts/main.ts
```diff
@@ -303,6 +304,34 @@ export function init() {
...
+  const formatQuestionName = (unformatted: string) => {
+    const formatted = unformatted
+      .toLowerCase()
+      .replace(/[^a-zA-Z ]/g, '')
+      .replace(/\s/g, '_')
+    return formatted
+  }
+
+  // Give a live preview of how the question name will be formatted in exports
+  const questionNameInput = document.getElementById('question-name-input')
```
gwendolyngoetz, line 317:
```suggestion
  const questionNameInput: HTMLInputElement | null = document.getElementById('question-name-input')
```

Are you able to add the type annotation here like you have for `formattedOutput`? That should negate the need for casting `questionNameInput as HTMLInputElement` further down.

server/app/assets/javascripts/util.ts
```diff
@@ -47,3 +47,17 @@ export function assertNotNull<T>(
   }
   return value
 }
+
+// Throttles a function's execution.  Helpful for functions triggered by keystrokes.
+export function debounce<A = unknown>(
+  func: (...params: A[]) => void,
+  timeout = 200,
```
gwendolyngoetz, line 54: Does 200ms feel laggy if you're typing fast?

melanie-exygy: You're right, it does feel laggy.  I'll lower it.  Will try 100ms.

server/app/assets/javascripts/main.ts
melanie-exygy: Good catch.  I'll update that.

melanie-exygy: Tried setting the type on `questionNameInput` when it's declared as you suggested, but I got this error:  
image
So, I think casting is my only option.  Unless you can think of another way around this?

gwendolyngoetz: Oh bummer. No worries. It's fine as is then.

server/app/assets/javascripts/util.ts
melanie-exygy: Lowering it to 50 ms

gwendolyngoetz (approved):

browser-test/src/admin_program_creation.test.ts
```diff
@@ -34,6 +34,7 @@ describe('program creation', () => {
         '### This is an example of some static text with formatting',
     })
 
+    await page.waitForTimeout(100) // ms
```
gwendolyngoetz, line 37: What's the timeout for?

melanie-exygy: The test was being flaky.  The snapshot sometimes had time to load the API name preview and sometimes not.  This gives it time to ensure that our snapshot always has the API name preview.

bion: IIRC the selector timeout is >=1,000ms

melanie-exygy: The selector will find the element, but the inner text will be blank.

bion: You could add an HTML class or empty element when the contents loads for the tests to look for

melanie-exygy: Added the HTML class but the timing of the snapshot was still off.  I decided to remove the debounce because it felt laggy anyway.  Without the debounce, the test works fine.

avaleske: Makes sense, it's not a very expensive function so running it on every keystroke is likely fine.

melanie-exygy: @bion Are you ok with this solution and me closing the conversation?

gwendolyngoetz: It's not crossing the wire so removing the debounce is probably for the best anyway.

Issue #888, opened 2021-04-26
Enable ability for bold/underline/etc for question descriptions


neel-a-mehta:
In #886 we discuss the importance of linkifying URLs in question descriptions to make it easier for program admins to link to external sources, such as maps of daycare centers or tables of income criteria.

Instead of doing a hacky RegEx-y solution, instead we can let Civiform admins write questions in Markdown or using a rich-text HTML editor ([or both](https://simplemde.com/)). This may increase complexity on the Civiform admin side, but it would enable a lot more creative control and let them convey more information.

neel-a-mehta:
#979 is an example of something we'd like to do with the markdown editor too -- adding images.

glorialiou:
Note: Rochester has mentioned this has something they would like.

msprenke:
Needs PRD

elisekalstad:
Would like to add hyperlinks to custom messages (statuses AND success message)

reemaka:
Just confirming, this is specifically for question text and help text, is that correct? i.e. the fields here:

![image](https://github.com/civiform/civiform/assets/7339672/0ed80be7-cf8c-4733-ab73-486b6524d5ba)
(screenshot is for date picker question, but same fields apply to all question types)

msprenke:
Yes Question & help and we'd like to add custom confirmation message text

reemaka:
The custom confirmation message is for the program, correct?

elisekalstad:
> The custom confirmation message is for the program, correct?

Yes! The custom confirmation message is specific to each program. We'd like to be able to add formatting to that message. Right now it's hard to parse out a URL or give emphasis to specific parts of the message.

reemaka:
Makes sense, thanks for clarifying!

reemaka:
I think a tricky part here is how to integrate this with translations. We'll need to think about how to handle that.

msprenke:
For translations the text formatting should be the same as for the main question.  Allow the user of bold/underline

PR #5495, opened 2023-08-31
Bug Fix - static text formatting

swatkat1:
### Description
Bugfix - Static text does not have the same formatting as texts on other question types.

## Release notes

Bugfix - Static text does not have the same formatting as texts on other question types.

#### General

Read the full guidelines for PRs [here](https://docs.civiform.us/contributor-guide/developer-guide/technical-contribution-guide#guidelines-for-pull-requests)

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)

### Issue(s) this completes

Fixes #5296

PR #5644, opened 2023-09-19
Implement markdown parsing for program descriptions, confirmation screen messages, and question text

### LINKED ISSUE #5434: Allow for additional customization on the confirmation screen at application submission
Right now, when a CiviForm admin makes a program, they can add a custom message that will be shown on the confirmation page after an application has been submitted. The goal is to use this message to explain next steps of the application process and/or highlight other programs to apply for. 

Given the importance of this handoff and the feedback we have received that it is not as clear as expected, we'd like to expand the functionality of this space to allow for linking out to other pages outside of CiviForm and the text in the submission view to be easier to read. 

This issue is complete with the admin can:  

1. Add links to text that allow for navigation to sites outside of CiviForm 
2. Add custom formatting of the text in the form of bold, underlined, and text size

rockycodes:
### Description

This PR implements the ability to use Markdown for formatting in admin text fields. The text fields affected are:
- Program Description (in multiple places for admin and applicant)
- Regular Question Text (on the summary page and as they're going through the application)
- Static Question Text (for applicants)
- Static Question Preview (for admin)
- Confirmation Screen (for applicants)

The main logic to review is in `TextFormatter.java` and `preview.ts`. The rest of the PR is updating/adding calls to the formatter and fixing tests. I've added comments in the PR to explain what pages are affected by calls to `formatText`.

To implement Markdown parsing, I used the already imported [commonmark-java](https://github.com/commonmark/commonmark-java) library for the Settings Panel (plus the `AutoLinkExtension` to preserve the ability to autodetect links). I imported Google's [OWASP Java HTML Sanitizer](https://github.com/OWASP/java-html-sanitizer) to sanitize the output before rendering.

For parsing markdown in `preview.ts`, I imported [markdown-it](https://www.npmjs.com/package/markdown-it). Both `commonmark-java` and `markdown-it` use [CommonMark](https://commonmark.org/) as the basis for their Markdown parsing so results should be the same.

_Functionality changes:_
With these changes, using ### to indicate an accordion is deprecated, but all other functionality should be the same. One change to call out is that urls are detected slightly differently with these changes -- urls starting with `www` will not be automatically detected (they must start with `http` or `https`). Also, emails will be autodetected as links, which is different from the current functionality. Unfortunately, the AutoLinkExtension library does not allow control over what is autodetected as a link. All existing deployments approved of these changes.

_Note:_ 
Both issues related to this change specifically requested the ability to underline text, which is generally discouraged in Markdown because underlined text can easily be confused with a link. I'll follow up with product if bold/italics are good enough and open a separate issue for adding underline if it's still needed.

## Release notes

This change allows CiviForm Admin to use Markdown to apply formatting to questions, program descriptions, and custom confirmation screen messages. Formatting options include italics, bolding, changing font size, adding a text link, and creating lists.

#### General

Documentation will be updated separately (tracked by [this issue](https://github.com/civiform/civiform/issues/5903)).

1. Test the backend functionality by editing the program description, custom confirmation message, and/or question text. Changes should show up for admin and applicants.
2. Test the frontend functionality by creating/editing a static question type. As you edit the text using Markdown you should see the appropriate formatting render in the window next to it.

### Issue(s) this completes

Fixes #5434; #888


server/app/assets/javascripts/preview.ts
```diff
@@ -463,11 +463,23 @@ class PreviewController {
...
-      listItem.textContent = item
+      listItem.appendChild(PreviewController.parseMarkdown(item))
       listTag.appendChild(listItem)
     })
     return listTag
   }
+
+  private static parseMarkdown(currentLine: string): Element {
+    const md = new MarkdownIt()
+    const parser = new DOMParser()
```
gwendolyngoetz, line 474: Do you know what the recommended practice for instantiating these two object are? I see `parseMarkdown` used in a tight loop and I just want to be cognizant of memory and garbage collection. Does the static typescript method get transpiled to not make it an issue?

server/app/assets/javascripts/preview.ts
```diff
@@ -77,6 +78,7 @@ class PreviewController {
   private static accordionContent = '>'
   private static accordionHeader = '### '
   private static bulletedItem = '* '
+  private static headerIndicator = '#'
```
gwendolyngoetz, line 81: What's this do?

server/app/assets/javascripts/preview.ts
```diff
@@ -463,11 +463,23 @@ class PreviewController {
...
       listTag.appendChild(listItem)
     })
     return listTag
   }
+
+  private static parseMarkdown(currentLine: string): Element {
+    const md = new MarkdownIt()
+    const parser = new DOMParser()
+    let html
+    if (currentLine[0] == this.headerIndicator) {
```
bion, line 476: **nitpick:** use `===` https://google.github.io/styleguide/jsguide.html#features-equality-checks

server/app/views/components/TextFormatter.java
```diff
@@ -50,6 +51,8 @@ public enum UrlOpenAction {
   private static final String ACCORDION_CONTENT = ">";
   private static final String ACCORDION_HEADER = "### ";
   private static final String BULLETED_ITEM = "* ";
+  private static final char HEADER = '#';
+  private static final CiviFormMarkdown civiFormMarkdown = new CiviFormMarkdown();
```
bion, line 55: **nitpick:** constant
```suggestion
  private static final CiviFormMarkdown CIVI_FORM_MARKDOWN = new CiviFormMarkdown();
```

server/app/assets/javascripts/preview.ts
gwendolyngoetz, line 476: (blocking) Please add a guard clause to make sure `currentLine[0]` can't meltdown the markdown if currentLine is null or an empty string.

bion (changes requested):

server/app/views/CiviFormMarkdown.java
```diff
@@ -17,6 +19,37 @@ public String render(String markdown) {
     return RENDERER.render(markdownRootNode);
   }
 
+  /**
+   * Removes p tags that are added by default by the CommonMark library. Logic was pulled from this
+   * commit which is the recommended way to solve this problem:
+   * https://github.com/commonmark/commonmark-java/commit/8b0fd7c73afaf1756edb6412c15a75cc423a6ba9
+   */
```
gwendolyngoetz, line 26: Please include details on why the `p` tags are getting removed

server/app/views/components/TextFormatter.java
```diff
@@ -184,7 +187,15 @@ private static DivTag buildAccordion(String title, String accordionContent) {
 
   private static UlTag buildList(List<String> items) {
     UlTag listTag = ul().withClasses("list-disc", "mx-8");
-    items.forEach(item -> listTag.with(li().withText(item)));
+    items.forEach(item -> listTag.with(li(applyMarkdown(item))));
     return listTag;
   }
+
+  private static DomContent applyMarkdown(String content) {
+    if (content.charAt(0) == TextFormatter.HEADER) {
```
gwendolyngoetz, line 195: (blocking) Add a guard clause here also to make sure `content` isn't null or empty and that `charAt(0)` can't blow up.

server/app/views/components/TextFormatter.java
```diff
@@ -50,6 +51,8 @@ public enum UrlOpenAction {
   private static final String ACCORDION_CONTENT = ">";
   private static final String ACCORDION_HEADER = "### ";
   private static final String BULLETED_ITEM = "* ";
+  private static final char HEADER = '#';
```
gwendolyngoetz, line 54: Same question as I left in the typescript. What is this indicating?

server/test/views/components/TextFormatterTest.java
```diff
@@ -170,14 +170,14 @@ public void listRendersCorrectly() {
   }
 
   @Test
-  public void rendersRecursively() {
+  public void rendersRecursivelyWithMarkdown() {
     String text =
         "Cheesecake Recipe\n"
```
gwendolyngoetz, line 175: Can you include a test case for markdown formatted urls. ```[link text](the_hyperlink)```. This has been a huge request from folks in Seattle.

gwendolyngoetz (changes requested):

server/app/views/components/TextFormatter.java
```diff
@@ -50,6 +51,8 @@ public enum UrlOpenAction {
   private static final String ACCORDION_CONTENT = ">";
   private static final String ACCORDION_HEADER = "### ";
```
melanie-exygy, line 52: Suggestion (non-blocking): In the static questions, I think it will be confusing that we're using these markdown symbols to define the accordion rather than to create a blockquote ('>') or an H3('###').  I suggest using non-markdown characters to delineate the accordion.  This could be a fast-follow. 
IIRC, there were no static questions in the Seattle db when @gwendolyngoetz last checked.  If that's the case for other civic entities, then we wouldn't have to worry about backward compatibility and could just switch out the symbols.

melanie-exygy (commented):
issue (blocking): Links seem to be handled differently in each of the markdown libraries.  If I use the usual markdown format for a link, it looks like this after parsing with markdown-it: 

image
but it looks like this after parsing with commonmark-java:

image

server/app/views/components/TextFormatter.java
gwendolyngoetz: When I check last there were no uses of accordion in Seattle. The only one I can think of specifically is on the api docs page.

My suggestion for new characters:
`ACCORDION_CONTENT="🪗";`
`ACCORDION_HEADER = "🎩";`

gwendolyngoetz:
> issue (blocking): Links seem to be handled differently in each of the markdown libraries. If I use the usual markdown format for a link, it looks like this after parsing with markdown-it:

The formatting issue with links on the Java side is probably because `TextFormatter.createLinksAndEscapeText` is called.

melanie-exygy (commented):
suggestion (non-blocking):  It might be good to add a test or two in `static_text.test.ts` with some basic markdown.

melanie-exygy (commented):
Suggestion (non-blocking):  This would need to be vetted by Product/UX, but now that we're adding markdown to the program descriptions, I would suggest parsing the markdown for the description on each program card in the program list view.  Otherwise, the admins will see the markdown like this, which I think looks messy.  This would be a separate issue, though.
image

melanie-exygy (commented):
issue (blocking): For the program description, bullet points aren't rendered.  ("email questions", "phone questions" and "etc." should be bullet points,)

image

server/test/views/components/TextFormatterTest.java
gwendolyngoetz: Also please add a test case that has `<script>alert('bad-time');</script>` in the markdown and verifies that there is no script tag in the generated html.

server/app/assets/javascripts/preview.ts
rockycodes: Great question, and I don't know. I feel like instantiating an object in a tight loop is probably not the recommended practice though so I'll do some digging. That's for catching/calling this out! The examples in the docs instantiate it right below the import statement which felt weird to me but might be the way to do it.

server/app/assets/javascripts/preview.ts
rockycodes: `renderInline` and the corresponding custom Java method I wrote `renderWithoutWrappingPTags` trim the wrapping tags off the block, regardless of what they are. This meant that header tags were getting trimmed and headers weren't showing up as expected. So I added this var + the `if` check in `parseMarkdown` to get around that. If it detects a header, it calls the regular `render` method that doesn't trim the wrapping tags. It's kinda funky and I'm open to other ideas of how to handle this if you have them.

server/app/views/components/TextFormatter.java
rockycodes: Lol those characters!! I agree this will be confusing, and would actually prevent users from using H3s. I would rather do it as a fast follow so that we can quickly revert in case it messes up existing values. I'll make an issue for it.

server/app/views/components/TextFormatter.java
rockycodes: I explained above, but lmk if you have more questions about this one in particular.

server/test/views/components/TextFormatterTest.java
rockycodes: The url one is a tricky one. The current logic in `createLinksAndEscapeText` detects urls automatically, and I implemented the markdown parsing so it only applies to strings surrounding the urls, not the urls themselves. Can you tell me more about the need for markdown formatted urls though? Have folks in Seattle wanted urls to be formatted differently that they are now with the automatic detection?

server/app/assets/javascripts/preview.ts
gwendolyngoetz: Makes sense. It did just occur to me that we probably should not allow for an H1 to be added here since there should only ever be a single H1 on a page. 

We even log it if there are [more than one found](https://github.com/civiform/civiform/blob/6339bd864ed472cbabfc86b7b0563e4a4075c754/server/app/views/applicant/ApplicantLayout.java#L113), which I haven't been good at reporting.

server/test/views/components/TextFormatterTest.java
gwendolyngoetz: Yeah they have. Being able to have custom link text, something friendlier than a long "scary" url, gets brought up often. I always point at #888 so duplicate issues don't get filed.

server/app/assets/javascripts/preview.ts
rockycodes: Oh good call! I could add some logic that automatically turns it into `##` or just trims off the H1. I need to check how big/small that allows us to make the text in that case, since that's a requirement. My other idea was to write some custom logic to apply the tailwind text classes but would be great to not have to write more custom logic. What do you think?

server/test/views/components/TextFormatterTest.java
rockycodes: That's good to know and complicates things a little more. Since it all runs through `formatText` we would need a way to distinguish if folks want the "scary" urls, or the markdown ones, which would involve a UI change. Unless we want to apply a rule based on where the url is rendering (program description vs. confirmation screen vs. static question) or we want to do away with scary urls all together which I could talk with Sijia about. What do you think?

server/app/assets/javascripts/preview.ts
rockycodes: done

rockycodes: done

server/app/views/CiviFormMarkdown.java
rockycodes: done

server/app/views/components/TextFormatter.java
rockycodes: done

rockycodes:
> issue (blocking): Links seem to be handled differently in each of the markdown libraries. If I use the usual markdown format for a link, it looks like this after parsing with markdown-it:
> 
> image but it looks like this after parsing with commonmark-java:
> 
> image

@melanie-exygy This is because of `TextFormatter.createLinksAndEscapeText` as Gwen said, which automatically detects and formats urls. So my changes aren't intended to parse Markdown for urls. There's no corresponding `createLinksAndEscapeText` function in `preview.ts` so there's no way to preview links correctly (this is what I meant in the first bullet point under "Notes" in my PR description). That said, Gwen and I are chatting on this PR about giving the ability to use Markdown for urls so that might change.

melanie-exygy:
> > issue (blocking): Links seem to be handled differently in each of the markdown libraries. If I use the usual markdown format for a link, it looks like this after parsing with markdown-it:
> > image but it looks like this after parsing with commonmark-java:
> > image
> 
> @melanie-exygy This is because of `TextFormatter.createLinksAndEscapeText` as Gwen said, which automatically detects and formats urls. So my changes aren't intended to parse Markdown for urls. There's no corresponding `createLinksAndEscapeText` function in `preview.ts` so there's no way to preview links correctly (this is what I meant in the first bullet point under "Notes" in my PR description). That said, Gwen and I are chatting on this PR about giving the ability to use Markdown for urls so that might change.

That makes sense as to why the markdown for links isn't being parsed on the Java side.  I suggest that we consider removing `TextFormatter.createLinksAndEscapeText` if possible now that we have the ability to parse Markdown links.  That would be dependent on verifying that `commonmark-java` handles links the way that we want.  Again, this could be a separate issue.  Note that `markdown-it` does create a navigable link, but it doesn't underline the link text or give it a blue color, so that's a bug we'd need to deal with.  Obviously, all of this is also dependent on your conversation with Gwen.

rockycodes:
> issue (blocking): For the program description, bullet points aren't rendered. ("email questions", "phone questions" and "etc." should be bullet points,)
> 
> image

@melanie-exygy I looked into this and realized that `createLinksAndEscapeText` was being called from the program index view and not `formatText` so bullet points were actually never working in that view! With the markdown implementation, it would be weird that some formatting would work, but other formatting wouldn't so I updated this to call `formatText`. `formatText` calls `createLinksAndEscapeText` with the same values that were previously being passed in here.

rockycodes:
> > > issue (blocking): Links seem to be handled differently in each of the markdown libraries. If I use the usual markdown format for a link, it looks like this after parsing with markdown-it:
> > > image but it looks like this after parsing with commonmark-java:
> > > image
> > 
> > 
> > @melanie-exygy This is because of `TextFormatter.createLinksAndEscapeText` as Gwen said, which automatically detects and formats urls. So my changes aren't intended to parse Markdown for urls. There's no corresponding `createLinksAndEscapeText` function in `preview.ts` so there's no way to preview links correctly (this is what I meant in the first bullet point under "Notes" in my PR description). That said, Gwen and I are chatting on this PR about giving the ability to use Markdown for urls so that might change.
> 
> That makes sense as to why the markdown for links isn't being parsed on the Java side. I suggest that we consider removing `TextFormatter.createLinksAndEscapeText` if possible now that we have the ability to parse Markdown links. That would be dependent on verifying that `commonmark-java` handles links the way that we want. Again, this could be a separate issue. Note that `markdown-it` does create a navigable link, but it doesn't underline the link text or give it a blue color, so that's a bug we'd need to deal with. Obviously, all of this is also dependent on your conversation with Gwen.

@melanie-exygy I chatted this over with Gwen and the only blocker is that currently we auto-detect links so we would want to keep that functionality. I dug into it a little and `commonmark-java` has an extension we can use for auto-detecting links. I didn't find similar functionality for `markdown-it` but since links currently aren't auto-formatted in the static text preview, maybe that's ok. If we decide we really need it I could look into another library. I'm going to try this out today!

server/test/views/components/TextFormatterTest.java
rockycodes: Update that Gwen and I talked this over and decided that the optimal solution is to do away with the custom auto-detecting functionality and just use Markdown URL syntax. This would give admins the ability to format text and URLs which is something they've been asking for! I want to make sure we maintain the ability to auto-detect URLs so I'll look into that before deprecating this custom logic.

server/app/assets/javascripts/preview.ts
```diff
@@ -261,9 +233,7 @@ class PreviewController {
       questionType &&
       questionType.textContent === PreviewController.STATIC_QUESTION_TEXT
     if (useAdvancedFormatting) {
-      const contentElement = PreviewController.formatText(text, true)
-      contentElement.classList.add('text-sm')
-      contentElement.classList.add('font-normal')
```
rockycodes, line 266: Removing these classes to be in line with what will actually render. They were removed from the backend in this PR: https://github.com/civiform/civiform/pull/5495/files#diff-02a0a04dcd13a5a42384a58108ed56ad7ded9a847d5f0084e2285f7134dd92ca

server/app/views/admin/programs/ProgramBaseView.java
```diff
@@ -30,7 +31,13 @@ protected final DivTag renderProgramInfo(ProgramDefinition programDefinition) {
             .withId("program-title")
             .withClasses("text-3xl", "pb-3");
     DivTag description =
-        div(programDefinition.localizedDescription().getDefault()).withClasses("text-sm");
+        div()
+            .with(
+                TextFormatter.formatText(
```
rockycodes, line 36: This is a new implementation of the formatter to properly format the program description when an admin is editing the screens in a program. Untitled

server/app/views/applicant/ApplicantProgramInfoView.java
```diff
@@ -74,7 +74,9 @@ private DivTag topContent(String programTitle, String programInfo, Messages mess
                 BaseStyles.TEXT_SEATTLE_BLUE, "text-2xl", "font-semibold", "text-gray-700", "mt-4");
 
     // "Markdown" the program description.
-    ImmutableList<DomContent> items = TextFormatter.formatText(programInfo, false);
+    ImmutableList<DomContent> items =
+        TextFormatter.formatText(
```
rockycodes, line 78: This is a program details page that I think might be deprecated. It's accessible by going to `/applicants/:applicantId/programs/:programId`, but I don't think you can get here by clicking anywhere on the home page. I'm updating it for now, but I think we might be repurposing this page in the future. See more discussion in [this comment thread](https://docs.google.com/document/d/1NvtSTFWYG3Q5sVBZHNsjWLbd8X9BC5baVYA37fqzMS0/edit?disco=AAAA7CfyKII)

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -160,11 +160,13 @@ public Content render(Params params) {
...
   private DivTag renderQuestionSummary(AnswerData data, Messages messages, long applicantId) {
-    DivTag questionPrompt = div(data.questionText()).withClasses("font-semibold");
-    if (!data.applicantQuestion().isOptional()) {
-      questionPrompt.with(ViewUtils.requiredQuestionIndicator());
-    }
-    DivTag questionContent = div(questionPrompt).withClasses("pr-2");
+    DivTag questionContent =
+        div(div()
+                .with(
+                    TextFormatter.formatText(
```
rockycodes, line 166: This will format the question text on the applicant summary/review screen where it lists all the questions.

server/app/views/components/ProgramCardFactory.java
```diff
@@ -72,7 +72,12 @@ public DivTag renderCard(Request request, ProgramCardData cardData) {
                                 "font-bold",
                                 "text-xl"))
                     .with(
-                        p(programDescriptionText)
+                        div()
+                            .with(
+                                TextFormatter.formatText(
```
rockycodes, line 77: This formats the program description for admin on the page that lists all their programs 
image

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -423,10 +423,10 @@ private LiTag programCard(
                 .withClasses(ReferenceClasses.APPLICATION_CARD_TITLE, "text-lg", "font-semibold")
                 .withText(program.localizedName().getOrDefault(preferredLocale));
     ImmutableList<DomContent> descriptionContent =
-        TextFormatter.createLinksAndEscapeText(
+        TextFormatter.formatText(
```
rockycodes, line 426: This formats the program description on the home page for an applicant. 
image

server/app/views/components/TextFormatter.java
```diff
@@ -1,190 +1,121 @@
...
-    return listTag;
+  private static String addRequiredIndicatorInsidePTag(String markdownText) {
+    int indexOfPTag = markdownText.lastIndexOf("</p>");
+    String stringWithRequiredIndicator = ViewUtils.requiredQuestionIndicator().toString() + "</p>";
+    return markdownText.substring(0, indexOfPTag) + stringWithRequiredIndicator;
+  }
+
+  private static String sanitizeHtml(String markdownText) {
+    PolicyFactory customPolicy =
+        new HtmlPolicyBuilder()
```
rockycodes, line 74: This will strip any elements and attributes not on the "allowed list" before returning the html. This disallows script tags for example.

server/app/views/components/TextFormatter.java
```diff
@@ -1,190 +1,121 @@
...
+    return new HtmlChangeListener<Object>() {
+      @Override
+      public void discardedTag(Object ctx, String elementName) {
+        logger.warn(String.format("HTML element: \"%s\" was caught and discarded.", elementName));
+      }
+
+      @Override
+      public void discardedAttributes(Object ctx, String tagName, String... attributeNames) {
+        for (String attribute : attributeNames) {
+          logger.warn(String.format("HTML attribute: \"%s\" was caught and discarded.", attribute));
```
rockycodes, line 116: Log if we catch elements or attributes that we don't allow

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -102,9 +102,9 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
     }
 
     ImmutableList<DomContent> questionTextDoms =
-        TextFormatter.createLinksAndEscapeText(
+        TextFormatter.formatText(
```
rockycodes, line 105: Formats text on an individual question for the applicant

server/app/views/questiontypes/StaticContentQuestionRenderer.java
```diff
@@ -26,7 +26,11 @@ public DivTag render(ApplicantQuestionRendererParams params) {
     DivTag questionTextDiv =
         div()
             .withClasses(ReferenceClasses.APPLICANT_QUESTION_TEXT, "mb-2", "font-bold", "text-xl")
-            .with(TextFormatter.formatText(question.getQuestionText(), false));
+            .with(
+                TextFormatter.formatText(
```
rockycodes, line 30: Formats text for a static type question for the applicant

server/app/views/applicant/ApplicantUpsellCreateAccountView.java
```diff
@@ -105,7 +105,9 @@ public Content render(
                 div()
                     .with(
                         TextFormatter.formatText(
-                            customConfirmationMessage.getOrDefault(locale), true))
+                            customConfirmationMessage.getOrDefault(locale),
```
rockycodes, line 108: Formats custom text used on the confirmation screen for the applicant after they submit an application

server/app/views/questiontypes/ApplicantQuestionRendererImpl.java
```diff
@@ -70,9 +70,9 @@ public final DivTag render(ApplicantQuestionRendererParams params) {
                         ReferenceClasses.APPLICANT_QUESTION_HELP_TEXT,
                         ApplicantStyles.QUESTION_HELP_TEXT)
                     .with(
-                        TextFormatter.createLinksAndEscapeText(
+                        TextFormatter.formatText(
```
rockycodes, line 73: Formats text on question help text.

melanie-exygy:
Seems like ordered lists, block quotes, h3s and horizontal rule aren't available.  Is this expected?  Also, unordered lists are indented and cause following items to be indented too.  Here's the markdown I wrote, followed by the parsed program description:

image

image

melanie-exygy:
If the markdown is too long, it overflows the card on the Program Index page:

image

server/app/views/components/TextFormatter.java
```diff
@@ -1,190 +1,121 @@
...
+  private static String sanitizeHtml(String markdownText) {
+    PolicyFactory customPolicy =
+        new HtmlPolicyBuilder()
+            .allowElements(
+                "p", "div", "h2", "h3", "h4", "h5", "h6", "a", "ul", "li", "span", "svg", "br",
+                "em", "strong", "code", "path", "pre")
+            // Per accessibility best practices, we want to disallow adding h1 headers to
+            // ensure the page does not have more than one h1 header
+            // https://www.a11yproject.com/posts/how-to-accessible-heading-structure/
+            // This logic changes h1 headers to h2 headers which are still larger than the default
```
melanie-exygy, line 81: If we're changing h1s to h2s, should we also change h2s to h3s so that there is a relative difference in size?  Otherwise, a user may not understand why "#" and "##" produce the same size headers.

server/app/views/components/TextFormatter.java
```diff
@@ -1,190 +1,121 @@
...
-      // Add the URL.
-      var urlTag =
-          a().withText(url.getOriginalUrl())
-              .withHref(url.getFullUrl())
-              .withClasses(
-                  "text-blue-900", "font-bold", "opacity-75", "underline", "hover:opacity-100");
-
-      if (urlOpenAction == UrlOpenAction.NewTab) {
-        urlTag
-            .withTarget("_blank")
```
gwendolyngoetz, line 120: (nonblocking) Maybe I'm just missing it. How does the new parser know when to add `target="_blank"`?

server/app/views/components/TextFormatter.java
gwendolyngoetz: (nonblocking) In what circumstances would this logging trigger? If it's rare cool, but if it will spam the logs with warnings Info might be better. The only time I can think of ever wanting to look would being logged is if maybe it's removing a `script` tag I happen to notice, but unless someone on the team has an issue, I probably wouldn't ever look at these.

gwendolyngoetz (approved):

melanie-exygy:
issue (blocking):  On regular question text, when the question is required, some markdown elements prevent all or part of the markdown from showing up .  For example, this markdown on the address question doesn't show up on the summary page:

image

image

However, if I reverse the unordered list and the bold, markdown does show, but only the first two lines:
image

image

suggestion:  Prevent the required question asterisk from being parsed with the markdown.  When the question is optional, all of the markdown is shown without any problem.

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -160,11 +160,13 @@ public Content render(Params params) {
 
   /** Renders {@code data} including the question and any existing answer to it. */
   private DivTag renderQuestionSummary(AnswerData data, Messages messages, long applicantId) {
-    DivTag questionPrompt = div(data.questionText()).withClasses("font-semibold");
-    if (!data.applicantQuestion().isOptional()) {
```
melanie-exygy, line 164: suggestion (blocking):  Adding the required question indicator after formatting the text (similar to this) might solve the bug with the markdown on required questions that I described in the other comment.

rockycodes: smart! i'll try that.

server/app/views/components/TextFormatter.java
rockycodes: This would trigger if someone tries to pass in markdown that isn't on our "allow" list. I don't see that happening a whole lot unintentionally, so "warn" level felt right. That said, I'm making an assumption and who knows what will happen when actual humans use it! Would you prefer to default to info level and up it to warn later if that feels right or vice versa?

server/app/views/components/TextFormatter.java
```diff
@@ -1,190 +1,121 @@
...
-      } else if (line.length() > 0) {
-        ImmutableList<DomContent> lineContent =
-            TextFormatter.createLinksAndEscapeText(
-                line, UrlOpenAction.NewTab, /*addRequiredIndicator= */ false);
-        builder.add(div().with(lineContent));
-      } else if (preserveEmptyLines) {
-        builder.add(div().withClasses("h-6"));
+      if (line.isBlank()) {
+        // Add an empty space character so the blank line is preserved
+        lines[i] = "&nbsp;\n";
```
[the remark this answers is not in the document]
rockycodes: Can you say more about this?

server/app/views/components/TextFormatter.java
rockycodes: Currently, it's [always adding it ](https://github.com/civiform/civiform/pull/5644/files#diff-603ea7dbf066a2af47b7d00b6c0af09e817f85787870903c714dcbca60dc2dbbR44). I made that call because we aren't using `SameTab` anywhere right now, and I think design wise there's a preference for opening in a new tab, but I could be wrong about that. Do you think I should add the ability to specify same vs new tab back in or hold off until the need comes up?

server/app/views/components/TextFormatter.java
rockycodes: I'm torn about this. My instinct is to not change more than we need to, but I agree it could be confusing. I'll think on it some more!

server/app/views/components/TextFormatter.java
gwendolyngoetz: Having external links open in a new tab is perfect. I was just wondering what the mechanism is that sets it. The link didn't take me directly to a spot, not sure where it was pointing.

server/app/views/components/TextFormatter.java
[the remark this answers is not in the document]
gwendolyngoetz: Ha. Github actually rendered the < br /> in my comment.

I was just asking how line breaks done by an admin in a textarea get converted to < br /> tags. Now sure if it was done by the markdown library or custom.

server/app/views/components/TextFormatter.java
rockycodes: Oh weird! I was trying to point you to line 34 (44 now) in `CiviformMarkdown.java` where we have `attributes.put("target", "_blank");`

server/app/views/components/TextFormatter.java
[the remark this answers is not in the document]
rockycodes: Ha! Silly github. It's done by the call to `.softbreak("<br/>")` on the HtmlRenderer builder in `CiviFormMarkdown.java`. https://www.javadoc.io/doc/com.atlassian.commonmark/commonmark/0.9.0/org/commonmark/renderer/html/HtmlRenderer.Builder.html#softbreak-java.lang.String-

server/app/views/components/TextFormatter.java
gwendolyngoetz: Oh I must have missed it. Thanks!

melanie-exygy:
> Seems like ordered lists, block quotes, h3s, h4s, h5s and horizontal rule aren't available. Is this expected?
> 
> issue (blocking): Also, unordered lists are indented and cause following items to be indented too. Here's the markdown I wrote, followed by the parsed program description:
> 
> image
> image

@rockycodes pointed-out that the indentation after a list is normal Markdown behavior.  Indentation can be reset by adding a new line.  I verified that this indentation reset works as expected.  So, consider the above "issue (blocking)" as unblocked!

server/app/views/components/TextFormatter.java
rockycodes: I'm leaning towards leaving this as is. In our communication we can specify that h1 headers aren't allowed and direct people to use h2 headers which should clear up confusion. Otherwise we would have to change all of the smaller headers as well. What do you think?

server/app/views/components/TextFormatter.java
gwendolyngoetz: In that case let's keep it at warning for now. If it becomes too spammy we can lower it later. Thanks!

server/app/views/applicant/ApplicantProgramSummaryView.java
rockycodes: I realized the issue was that I was only attaching the required indicator to p tags! I've got a fix I'll push up.

server/app/views/components/TextFormatter.java
melanie-exygy: That makes sense to me, especially since the only header we support right now is h2.

melanie-exygy:
> issue (blocking): If the markdown is too long, it overflows the card on the Program Index page:
> 
> image

Confirmed that this problem is no longer reproducible.

melanie-exygy (approved):

PR #5657, opened 2023-09-22
"Return" should trigger an action in Applications Search form

### LINKED ISSUE #5464: "Return" should trigger an action
As a Program Admin, I want to be able to use my "Return/Enter" key to trigger an action. When I am searching for an application by name, email or ID, I want to be able to hit "Return" and see the search result. I currently need to click on the "Filter" button with my mouse each time.

A Program Admin can hit "Return" and have their input trigger an action in CiviForm, such as Filter.

**Additional context**
image

(Would expect the blue-highlighted button to correspond the hitting "Return".)

**Done when**
I can hit "Return" on my keyboard and have my search results display.

swatkat1:
### Description

"Return" should trigger an action in Applications Search form

## Release notes

"Return" should trigger an action in Applications Search form

#### General

### Issue(s) this completes

Fixes #5464

### DIFF
--- a/server/app/assets/javascripts/main.ts
+++ b/server/app/assets/javascripts/main.ts
@@ -262,8 +262,15 @@ function attachStopPropogationListenerOnFormButtons() {
 /**
  * Disables default browser behavior where pressing Enter on any input in a form
  * triggers form submission. See https://github.com/civiform/civiform/issues/3872
+ * Except when the ignore element is present, the default form action would work
+ * See https://github.com/civiform/civiform/issues/5464
  */
 function disableEnterToSubmitBehaviorOnForms() {
+  const ignoreElement = document.getElementById('cf-search-applications')
+  if (ignoreElement) {
+    return
+  }
+
   addEventListenerToElements('form', 'keydown', (e: KeyboardEvent) => {
     const target = (e.target as HTMLElement).tagName.toLowerCase()
     // If event originated from a button, link, or textarea, it should proceed with the default action.
--- a/server/app/assets/javascripts/main.ts
+++ b/server/app/assets/javascripts/main.ts
@@ -262,19 +262,20 @@ function attachStopPropogationListenerOnFormButtons() {
 /**
  * Disables default browser behavior where pressing Enter on any input in a form
  * triggers form submission. See https://github.com/civiform/civiform/issues/3872
- * Except when the ignore element is present, the default form action would work
+ * Except when the overrideDisableSubmitOnEnter attribute is present for the form, the default form action would work
  * See https://github.com/civiform/civiform/issues/5464
  */
 function disableEnterToSubmitBehaviorOnForms() {
-  const ignoreElement = document.getElementById('cf-search-applications')
-  if (ignoreElement) {
-    return
-  }
-
   addEventListenerToElements('form', 'keydown', (e: KeyboardEvent) => {
     const target = (e.target as HTMLElement).tagName.toLowerCase()
-    // If event originated from a button, link, or textarea, it should proceed with the default action.
+    const overrideDisableSubmitOnEnter = document
+      .getElementById((e.target as HTMLElement).id)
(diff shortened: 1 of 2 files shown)


server/app/assets/javascripts/main.ts
```diff
@@ -262,8 +262,15 @@ function attachStopPropogationListenerOnFormButtons() {
 /**
  * Disables default browser behavior where pressing Enter on any input in a form
  * triggers form submission. See https://github.com/civiform/civiform/issues/3872
+ * Except when the ignore element is present, the default form action would work
+ * See https://github.com/civiform/civiform/issues/5464
```
gwendolyngoetz, line 266: After you seem my notes below please update this comment to include instructions on how to bypass the custom behavior that blocks hitting enter.

server/app/assets/javascripts/main.ts
```diff
@@ -262,8 +262,15 @@ function attachStopPropogationListenerOnFormButtons() {
...
  */
 function disableEnterToSubmitBehaviorOnForms() {
+  const ignoreElement = document.getElementById('cf-search-applications')
+  if (ignoreElement) {
+    return
+  }
```
gwendolyngoetz, line 272: You're on the right track. Since you're looking for a specific ID this won't work well on forms with multiple input fields. 

Rather than tag each input we can just tag a specific form. I added an example below that adds a custom `data-` attribute to the form element. The `data-` attribute should not conflict with any existing or future libraries.

This way you just have to add that attribute to a form and the whole form will now work as browsers were intended to function by default.

server/app/assets/javascripts/main.ts
```diff
@@ -262,8 +262,15 @@ function attachStopPropogationListenerOnFormButtons() {
...
+
   addEventListenerToElements('form', 'keydown', (e: KeyboardEvent) => {
     const target = (e.target as HTMLElement).tagName.toLowerCase()
     // If event originated from a button, link, or textarea, it should proceed with the default action.
```
gwendolyngoetz, line 276: Add this after `const target`
```typescript
const overrideDisableSubmitOnEnter = document.getElementById((e.target as HTMLElement).id)
      ?.closest('form')
      ?.hasAttribute('data-override-disable-submit-on-enter');
```

Add this to the `if`
```typescript
overrideDisableSubmitOnEnter === false &&
```

server/app/views/admin/programs/ProgramApplicationListView.java
```diff
@@ -211,6 +211,7 @@ private FormTag renderSearchForm(
                                 .withClasses("flex"))),
             FieldWithLabel.input()
                 .setFieldName(SEARCH_PARAM)
+                .setId("cf-search-applications")
```
gwendolyngoetz, line 214: Remove `setId` on the input field

Add this to the form
```.attr("data-override-disable-submit-on-enter")```

**Review by gwendolyngoetz — CHANGES_REQUESTED** [review]

File: server/app/assets/javascripts/main.ts (hunk 3 above)
Comment (swatkat1) [reply server/app/assets/javascripts/main.ts:276]: thanks @gwendolyngoetz

File: server/app/assets/javascripts/main.ts [hunk 5]
```diff
@@ -262,12 +262,20 @@ function attachStopPropogationListenerOnFormButtons() {
 /**
  * Disables default browser behavior where pressing Enter on any input in a form
  * triggers form submission. See https://github.com/civiform/civiform/issues/3872
+ * Except when the overrideDisableSubmitOnEnter attribute is present for the form, the default form action would work
```
gwendolyngoetz, line 265:
```suggestion
 * Except when the data-override-disable-submit-on-enter attribute is present for the form, the default form action would work
```

gwendolyngoetz (changes requested):
Just update this string and I'll approve

gwendolyngoetz (approved):

Issue #5724, opened 2023-10-04
Refactor code to call function to get data from function in versionRepository instead of Version class


dkatzz:
**Describe the solution you'd like**
Related to [Caching frequent database queries](https://docs.google.com/document/d/1COr52i3ouS-hGcVL4SqoDLYDdrCwir3rGKbkjE3Tfgw/edit#heading=h.bknox3ib8wpb), we want to refactor our code to create functions in versionRepository to get question and program data instead of getting that data from the Version class.

This refactor won't change any behavior but will make it easier for us to introduce the caching changes later.

PR #5729, opened 2023-10-04
Continue refactoring code to make getQuestionsForVersion an instance method

dkatzz:
### Description

https://github.com/civiform/civiform/pull/5711 refactored places that call `version.getQuestions()` to use a new function `getQuestionsForVersion()`. We'll want to do this in the version class as well and make the method an instance method, which is handled in this PR. By changing it into an instance method, we had to refactor and move some other functions as well. 

See more information in [doc](https://docs.google.com/document/d/1COr52i3ouS-hGcVL4SqoDLYDdrCwir3rGKbkjE3Tfgw/edit#heading=h.bknox3ib8wpb)

#### General

### Issue(s) this completes

Related to https://github.com/civiform/civiform/issues/5724

### DIFF
--- a/server/app/models/Version.java
+++ b/server/app/models/Version.java
@@ -13,6 +13,7 @@ import javax.persistence.JoinTable;
 import javax.persistence.ManyToMany;
 import javax.persistence.Table;
 import play.data.validation.Constraints;
+import repository.VersionRepository;
 import services.program.ProgramDefinition;
 import services.question.exceptions.QuestionNotFoundException;
 import services.question.types.QuestionDefinition;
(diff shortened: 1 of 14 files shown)


server/app/models/Version.java
```diff
@@ -122,7 +123,7 @@ public Optional<Program> getProgramByName(String name) {
    * can exist in a version.
    */
   public Optional<Question> getQuestionByName(String name) {
-    return getQuestions().stream()
+    return VersionRepository.getQuestionsForVersion(this).stream()
```
bion, line 126: **issue (non-blocking):** I missed it in the last PR, but FYI this method will need to become an instance method rather than static in order to consume the cache.

In general it's best to avoid public static methods since they require changes to all consumers when they end up needing state from the class they're associated with.

bion (approved):

server/app/models/Version.java
dkatzz: Ah you're right, sorry for missing that too! I think that means I'll have to do a refactor that involves changing these static methods to instance methods as well, since those call the getQuestionsForVersion function

dkatzz: Ended up making that change as part of this PR, so it was a larger refactor.

bion (approved):

PR #5743, opened 2023-10-06
Rename UserRepository to AccountRepository

bion:
The system does not have a resource called "user", the closest thing we have is "account" along with "applicant", with the former being most closely associated with the identity of a specific user.


yotommy (approved):
Thanks for this clarifying change!

gwendolyngoetz (approved):

PR #5749, opened 2023-10-06
Update feature_flag.md

nb1701:


dkatzz (approved):
Agree this doesn't belong here, but if we're adding a config flag for terraform we would still want to make changes like this (https://github.com/civiform/cloud-deploy-infra/commit/9d17356ff1fa1f3a16c97608cc00cbd4c7c11ffe). It probably belongs in its own section (or file) like "Adding a terraform config flag"

dkatzz:
I think it makes sense on that page, but maybe a section called "Adding a flag in terraform"

I think the steps there would just be adding it in these 3 places (https://github.com/civiform/cloud-deploy-infra/commit/9d17356ff1fa1f3a16c97608cc00cbd4c7c11ffe) (but in whichever directory it belongs) because these all need to be set in the deployment file as opposed to SettingsManifest, since these need to be created as part of the deployment process.

dkatzz:
> I'll merge this and make a note for myself to fix that up

Sounds good, thank you!!

PR #5752, opened 2023-10-06
Add screen to prevent duplicate application submissions

### LINKED ISSUE #4869: Warn applicants before resubmitting application if answers haven't changed
Right now, resubmitting an application without changing any answers is possible. AR is concerned that this leads to people resubmitting applications and creating unnecessary duplicates.

Present a Modal that states that "none of your answers have changed" and confirms the user intends resubmit the application.

Highlight the answers that changed on the review screen.

**Additional context**
AR requests this feature.,

**Done when**
Above.

rockycodes:
### Description

This PR prevents submission of duplicate applications and adds a screen to be shown to applicants if they attempt to re-submit an application without making any changes to it.

image

## Release notes

With this change, applicants will no longer be able to re-submit applications without editing them. If an applicant attempts to submit a duplicate application, they will be shown a screen letting them know there are no changes to save and asking if they would like to continue editing or exit the application.

#### General

#### User visible changes

- [x] Followed steps to [internationalize new strings](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#internationalization-for-application-strings) (waiting on translations to merge)

1. Seed program and question data
2. Log in as an admin and publish the program(s), log out
3. As an applicant (logged in or out), complete an application and submit
4. Click "Apply to another program"
5. Click "Edit" on the program you've already applied for
6. Click "Submit" without making any changes (also try opening up a question and saving again without making changes, behavior should be the same)
7. See screen telling you no changes were made

### Issue(s) this completes

Fixes #4869


server/app/services/applicant/exception/DuplicateApplicationException.java
```diff
@@ -0,0 +1,11 @@
+package services.applicant.exception;
+
+/**
+ * Thrown if an application is submitted that is a duplicate of a previously submitted one. Triggers
+ * a screen to be shown to the applicant asking them to continue editing or exit the application.
+ */
+public class DuplicateApplicationException extends RuntimeException {
```
yotommy, line 7: I'm not very familiar with the best practices regarding Exception handling, so this may be off base.

Would it be safer to make this a checked exception? Having an Exception inadvertently thrown by a controller will result an error in the browser, which doesn't make anyone happy.

browser-test/src/application_navigation.test.ts
```diff
@@ -350,6 +350,60 @@ describe('Applicant navigation flow', () => {
...
+      await applicantQuestions.clickNext()
+      await applicantQuestions.submitFromReviewPage()
+
+      // Submit the application again without editing it
+      await applicantQuestions.returnToProgramsFromSubmissionPage()
+      await applicantQuestions.clickApplyProgramButton(programName)
+      await applicantQuestions.submitFromReviewPage()
+
+      // See the duplicate submissions page
+      await validateScreenshot(page, 'duplicate-submission-page')
```
avaleske, line 386: Can we verify that we're on the page by looking for some identifying text, like a header, as well?

The screenshot won't fail the test until it compares them at the end.

browser-test/src/application_navigation.test.ts
```diff
@@ -350,6 +350,60 @@ describe('Applicant navigation flow', () => {
...
+      await applicantQuestions.clickEdit()
+
+      // Edit the application but insert the same values as before and submit.
+      await applicantQuestions.answerDateQuestion('2021-11-01')
+      await applicantQuestions.answerEmailQuestion('test1@gmail.com')
+      await applicantQuestions.clickNext()
+      await applicantQuestions.submitFromReviewPage()
+
+      // See the duplicate submissions page
+      await validateScreenshot(page, 'duplicate-submission-page')
```
avaleske, line 401: ditto here

server/app/services/applicant/exception/DuplicateApplicationException.java
avaleske: Thinking about _Effective Java_ items 69 and 70, I'm not sure we should be using exceptions here at all.

- 69 says that "exceptions should be exceptional" and checking if this is a duplicate feels more like a "normal control flow" sort of thing. I could be convinced that this is no less exceptional then "program not found", which we do use an exception for, but at the moment a 404 seems more exceptional then submitting a duplicate program
- 70 says that we should used checked exceptions for recoverable conditions and runtime exceptions for programming errors

Links (I couldn't find a great, linkable version of _Effective Java_ so this will have to do):
- https://github.com/AlphaWang/alpha-effective-java-3e/blob/master/10_exceptions/item_69_use_exceptions_only_for_exceptional_conditions.md#item-69-use-exceptions-only-for-exceptional-conditions
- https://github.com/AlphaWang/alpha-effective-java-3e/blob/master/10_exceptions/item_70_use_checked_exceptions_for_recoverable_conditions_and_runtime_exceptions_for_programming_errors.md

server/app/repository/ApplicationRepository.java
```diff
@@ -128,14 +129,16 @@ private Application submitApplicationInternal(
           continue;
         }
         boolean isDuplicate = applicant.getApplicantData().isDuplicateOf(app.getApplicantData());
+        if (isDuplicate) {
+          throw new DuplicateApplicationException();
```
avaleske, line 133: I added a note to the exception class with more detail, but I'm not sure we should be using an exception here.

That said, I'm not immediately sure how to refactor this to avoid this strategy but happy to talk further or be convinced otherwise.

server/app/views/components/ButtonStyles.java
```diff
@@ -91,4 +91,12 @@ public final class ButtonStyles {
   // Just like CLEAR_WITH_ICON, but in dropdowns we want to remove the rounded corners.
   public static final String CLEAR_WITH_ICON_FOR_DROPDOWN =
       StyleUtils.removeStyles(CLEAR_WITH_ICON, "rounded-full");
+
+  public static final String LINK_STYLE =
+      StyleUtils.joinStyles(
+          StyleUtils.removeStyles(
```
avaleske, line 97: suggestion: I find this a little hard to reason about. Could we instead duplicate the contents of `BUTTON_BASE_OUTLINE_SEMIBOLD` here?

Also, `LINK_STYLE` is a little generic, could we name with something like `LINK_STYLED_BUTTON`

avaleske (changes requested):
Overall lgtm but two questions:
- I had questions about whether exceptions were the right control flow to use here.
- iirc we thought one reason this was happening was because people didn't realize they had already submitted the application, right? If that's the case, do we want the copy to say something like "already submitted"?

rockycodes:
@avaleske re:
> 
>     * iirc we thought one reason this was happening was because people didn't realize they had already submitted the application, right? If that's the case, do we want the copy to say something like "already submitted"?

I'll start a thread in Slack with you and Sijia to discuss!

server/app/services/applicant/exception/DuplicateApplicationException.java
gwendolyngoetz: I never much cared for the `ProgramNotFoundException` or `QuestionNotFoundException`. Something like a Null Object or other pattern could have probably taken the place for those and provided more flexibility. I'd recommend we don't continue to propagate this pattern, but if we have infrastructure components already build to use that pattern I'd go with it in this case and have a larger refactor issue made.

Also checked exception are the devil and no one will convince me otherwise.

browser-test/src/application_navigation.test.ts
```diff
@@ -350,6 +350,60 @@ describe('Applicant navigation flow', () => {
...
     })
+
+    it('shows "no changes" page when a duplicate application is submitted', async () => {
+      const {page, applicantQuestions} = ctx
+      await applicantQuestions.applyProgram(programName)
+
+      // Fill out application and submit.
+      await applicantQuestions.answerDateQuestion('2021-11-01')
+      await applicantQuestions.answerEmailQuestion('test1@gmail.com')
+      await applicantQuestions.clickNext()
```
melanie-exygy, line 361: question (non-blocking):  Why do we need to .clickNext() twice?

server/app/services/applicant/exception/DuplicateApplicationException.java
melanie-exygy: I agree that this is an unusual use of exceptions, but, in this case, I don't see how it will cause issues other than being annoying.  I wanted to suggest that we refactor by having the `ApplicationRepository.submitApplicationInternal()` return a null object, but then we'd have no way of knowing why the method returned null, assuming there can be reasons other than a duplicate application.

Let's definitely make it a checked exception, though.

server/app/views/applicant/PreventDuplicateSubmissionView.java
```diff
@@ -0,0 +1,91 @@
...
+            .with(
+                p(messages.at(
+                        MessageKey.TITLE_NO_CHANGES_TO_SAVE.getKeyName(),
+                        roApplicantProgramService.getProgramTitle()))
+                    .withClasses("text-black", "font-bold", "text-xl", "pb-8"))
+            .with(p(messages.at(MessageKey.CONTENT_NO_CHANGES.getKeyName())).withClasses("pb-8"))
+            .with(
+                div()
+                    .withClasses(
+                        "flex", "flex-col", "gap-4", StyleUtils.responsiveSmall("flex-row"))
```
melanie-exygy, line 56: nit (non-blocking): I think "flex-col" and "flex-row" cancel each other out here, no?  Since "flex-row" is the default, you could remove both of those classes.

server/app/views/components/ButtonStyles.java
melanie-exygy: suggestion (blocking):  Can we use one of the link styles that we already have so that our styles are consistent?  There's `LinkElement.java` and `ApplicantStyles.LINK`.  Seems like we could use one of the transparent buttons like we do in our other modals where we have "Continue to application" or "Continue without an account" (ButtonStyles.OUTLINED_TRANSPARENT).  @sijiayam could we use an existing style and mimic another modal?

server/app/services/applicant/exception/DuplicateApplicationException.java
avaleske: I think the refactor would look like creating an `ApplicationSubmissionResult` object that contains a field for the submitted application and a field for any errors, and returning that up the tree instead of the `Application` that we currently do. Similarly, methods like `validateApplicationForSubmission()` would need to return something like a `ApplicationValidationResult`.

Given that it would probably be more confusing to have some errors bubble up with the current exception strategy and others with this wrapper object, I'm ok with leaving it as you have it in the PR now.

server/app/repository/ApplicationRepository.java
```diff
@@ -128,14 +129,16 @@ private Application submitApplicationInternal(
           continue;
```
avaleske, line 129: Do we need to loop through old applications if we only care about comparing with the most recent one? Or is it possible that there could be multiple `ACTIVE` applications?

server/test/controllers/applicant/ApplicantProgramReviewControllerTest.java
```diff
@@ -240,6 +240,40 @@ public void submit_incomplete_showsError() {
         .contains("There's been an update to the application");
   }
 
+  @Test
```
melanie-exygy, line 243: praise:  nice job on writing a thorough test that covers both cases

server/app/services/applicant/exception/DuplicateApplicationException.java
melanie-exygy: @avaleske I like that idea for the refactor.  My vote is that we create a separate tech debt issue to refactor all of these un-exceptional exceptions.  :-)

rockycodes: I like that idea for a refactor a lot @avaleske! I'll follow @melanie-exygy suggestion and create an issue to do that before closing out this work.

server/app/repository/ApplicationRepository.java
rockycodes: @melanie-exygy can you say more about why you think a log would be helpful? I removed the log since this PR prevents saving dupes to the db. I thought it would be extra noise to have a log saying an applicant attempted to submit a duplicate.

server/app/views/applicant/PreventDuplicateSubmissionView.java
rockycodes: I'm pretty sure we need both of these since we want the buttons to be in a column when the screen is small and a row otherwise (I copied this from the [Ineligible Screen](https://github.com/civiform/civiform/blob/3fb9bad36d53d0e0216f8e4255779bde5cd1684e/server/app/views/applicant/IneligibleBlockView.java#L98)). The "flex-col" says the buttons should always be in a column, but then the `responsiveSmall("flex-row")` overrides that to say that any screen size bigger than "small" will have them in a row. Let me know if you understand differently!

server/app/services/applicant/exception/DuplicateApplicationException.java
rockycodes: I added this issue: https://github.com/civiform/civiform/issues/5775

Feel free to comment/add to it.

server/app/repository/ApplicationRepository.java
gwendolyngoetz: Since the exception is used for control flow and we are handling it with a screen for the user I'm not seeing much benefit of logging it. At most maybe just info level.

If I'm looking through my prod logs would there be anything I need to do or report about a warning? If not it creates extra noise and distraction from other potential issues.

server/app/repository/ApplicationRepository.java
rockycodes: That's a good question. Based on the code here, it looks like we mark all the old apps as `OBSOLETE` before saving the new one as `ACTIVE` so there should only ever be one `ACTIVE` version. This loop ignores `OBSOLETE` so those would never match and the `DRAFT` application is the one we're looking at, so that would be skipped as well. The only other lifecycle stage is `DELETED` which is never used. So, long story short, I think we're only ever comparing to the single `ACTIVE` application in the db in this loop. cc @yotommy who originally added the dupes check here. What do you think?

browser-test/src/application_navigation.test.ts
rockycodes: Done

server/app/repository/ApplicationRepository.java
rockycodes: I looked at the existing logs and and an info log might be helpful, otherwise there's no indication that the applicant did not successfully submit:
image

There's no action to take from this, so I think info level is good. I'll add that.

melanie-exygy: I was thinking that we might want to know if users keep trying to create duplicates so that we can address the UX root cause of their confusion.  It's probably unnecessary, though, as you and Gwen said, since we're preventing duplicates from happening.  Feel free to leave it out.

rockycodes: Done! Added an info log. I had to update the `logback.xml` file to allow info level logs from `ApplicationRepository`. Lmk if you see any problems with doing that.

server/app/views/applicant/PreventDuplicateSubmissionView.java
melanie-exygy: You're right.  Thanks for explaining!

melanie-exygy (approved):

PR #5753, opened 2023-10-09
Add version cache implementation gated behind feature flag

dkatzz:
### Description

This change adds caching for for data associated with versions, including questions and programs (gated behind a feature flag). We aren't caching versions themselves, since we don't seem to be requesting any version data except previous version data (from admin flow) and active data. The questionList and programList calls are the most expensive, so going to begin by caching those.

See [doc](https://docs.google.com/document/d/1COr52i3ouS-hGcVL4SqoDLYDdrCwir3rGKbkjE3Tfgw/edit) for more information.

#### General

#### New Features

### Issue(s) this completes

### DIFF
--- a/server/app/repository/VersionRepository.java
+++ b/server/app/repository/VersionRepository.java
@@ -62,8 +62,8 @@ public final class VersionRepository {
   private final DatabaseExecutionContext databaseExecutionContext;
 
   private final SettingsManifest settingsManifest;
-  private SyncCacheApi questionsByVersionCache;
-  private SyncCacheApi programsByVersionCache;
+  private final SyncCacheApi questionsByVersionCache;
+  private final SyncCacheApi programsByVersionCache;
 
   @Inject
   public VersionRepository(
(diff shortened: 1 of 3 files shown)


dkatzz:
@bion and @gwendolyngoetz - added you both to this, but there is also an open thread about doing it the other way instead. Figured we could do it this way for now (since the work was mostly done already), but happy to do a follow up with the other method.

server/app/repository/VersionRepository.java
```diff
@@ -58,12 +61,23 @@ public final class VersionRepository {
   private final ProgramRepository programRepository;
   private final DatabaseExecutionContext databaseExecutionContext;
 
+  private final SettingsManifest settingsManifest;
+  private SyncCacheApi questionsByVersionCache;
+  private SyncCacheApi programsByVersionCache;
```
bion, line 66: **nitpick (blocking):** 
```suggestion
  private final SyncCacheApi questionsByVersionCache;
  private final SyncCacheApi programsByVersionCache;
```

bion (approved):
I think this is a fine way to do the feature flag

gwendolyngoetz (approved):

PR #5760, opened 2023-10-10
Make "Choose File" button keyboard accessible.

### LINKED ISSUE #5523: (WCAG 2.1) "Choose File" Control is not keyboard accessible.
OBSERVED RESULTS:

- The "Choose File" control for file upload is not keyboard accessible and lacks operability.
- The control does not have a clear role of a button or link, which can confuse users relying on keyboard navigation.
- Screenshot

EXPECTED RESULTS:

- The "Choose File" control should be made keyboard accessible and operable, allowing users to interact with it using keyboard input.
- Add a valid role attribute to the "Choose File" control to indicate its purpose. Consider using <button> <a> or role="button" and tabindex=”0” to convey that it's a button for selecting files.

STEPS TO REPRODUCE:

- Navigate to [URL](https://demo.civiform.dev/)
- Navigate to and activate the ‘Apply’ for the Comprehensive sample program.
- Navigate to and activate “What is your address?”.
- Fillup the form fields and continue till step 8.
- On the step 8 Upload anything from your computer, navigate to the "Choose File" control using TAB and OR SHIFT tab key.
- Observe that the "Choose File" Control is not keyboard accessible.

ENVIRONMENT DETAILS:

- Device(s) (brand, OS version, test key or release key): Windows 10 Enterprise (64-bit),MacOS Ventura Version 13.4
- Screen reader name and version: JAWS 2022.2211.7, NVDA 2022.3, VoiceOver
- Browser name and version: Chrome Version 113.0.5672.93

caitlinshk:
### Description

Make "Choose File" button keyboard accessible.

#### General

#### User visible changes

- [x] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method -- **Looks like test already exists in browser-test/src/file.test.ts**

As an applicant, open a file upload question and verify keyboard tabbing includes the "Choose file" button.

### Issue(s) this completes

Fixes #5523

### DIFF
--- a/server/app/views/questiontypes/FileUploadQuestionRenderer.java
+++ b/server/app/views/questiontypes/FileUploadQuestionRenderer.java
@@ -73,6 +73,7 @@ public class FileUploadQuestionRenderer extends ApplicantSingleQuestionRenderer 
             label()
                 .withFor(fileInputId)
                 .withText(messages.at(MessageKey.BUTTON_CHOOSE_FILE.getKeyName()))
+                .attr("role", "button")
                 .attr("tabindex", 0)
                 .withClasses(ButtonStyles.OUTLINED_TRANSPARENT, "w-44", "mt-2", "cursor-pointer"));
   }
--- a/server/app/views/questiontypes/FileUploadQuestionRenderer.java
+++ b/server/app/views/questiontypes/FileUploadQuestionRenderer.java
@@ -2,6 +2,7 @@ package views.questiontypes;
 
 import static j2html.TagCreator.div;
 import static j2html.TagCreator.label;
+import static j2html.TagCreator.span;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


server/app/views/questiontypes/FileUploadQuestionRenderer.java
```diff
@@ -73,6 +73,7 @@ protected DivTag renderInputTag(
             label()
```
rockycodes, line 73: question (potentially blocking): What do you think about making this a button? I noticed when I was tab navigating I can navigate to "Choose File" now, but I can't interact with it (I expected that hitting "enter" would allow me to upload a file). I think changing this to a button might resolve that, but I'm not sure.

rockycodes (changes requested):
I have a question/thought about being able to interact with the "Choose File" button.

server/app/views/questiontypes/FileUploadQuestionRenderer.java
rockycodes: Actually, I take it back, you can interact with it with "control option space" which is how you interact with buttons! So I think the functionality is fine, but I'm still curious what you think of making it a button. Right now the screen reader tells me I'm on a "group" which isn't super clear.

rockycodes: Ah I gotcha. I'll spend a little time futzing with it, but seems like it might not be possible.

caitlinshk: Awesome thanks! Eager to hear if you find anything :)

rockycodes: I think this should work!
`.attr("role", "button")`

When I tried it out locally with my screen reader it recognized it as a button and prompted me to interact with it. Here is an [SO post](https://stackoverflow.com/questions/55197141/how-to-make-screenreader-not-recognise-an-element-as-an-element-group) that helped me understand the role of "role" if that's helpful!

caitlinshk: Ah, just the "role" alone didn't work for me locally, so I got rid of that. But having both "role" and "tabindex" works and I agree that this label should have the button role!

caitlinshk: Updated to put the role on a `div` element, since https://www.w3.org/TR/html-aria/  indicates `label`s aren't allowed to have roles. (Thanks for finding that doc Rocky!)

caitlinshk:
Let me know what you think about this change with the span!

rockycodes (approved):
Wahoo it works! I'm able to interact with the "button" with my keyboard so I think we can call this good to go.

Issue #5758, opened 2023-10-10
Allow x-axis scrolling in eligibility conditions screen


paripatel07:
**Is your feature request related to a problem? Please describe.**
There is a 4 condition limitation on eligibility condition screen. This makes it confusing for users to have to go between multiple screens for complex conditions

**Describe the solution you'd like**
This bug is to allow x-axis scrolling so multiple screens are not needed

**Additional context**
[Add any other context or screenshots about the feature request here.
](https://docs.google.com/document/d/1VLlxrAih-5PEBJEq2o_EwS5brllRCUZBGUGtRkgljqI/edit?resourcekey=0-Ogri95TZL0u7sTJRj7___Q&disco=AAAA7kqNPu4)

**Done when**
Users can add more than 4 conditions on one page

PR #5766, opened 2023-10-10
Admin can use more than 4 questions in a predicate

bion:
### Description

Enables x-axis scrolling in predicate configurator to allow more than four questions to be used in a single predicate. The UX here is not ideal, but this change makes the feature more capable for the time being. We anticipate a holistic redesign of the predicate configuration UX to be prioritized within the next quarter or two.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/5758

### DIFF
--- a/browser-test/src/admin_predicates.test.ts
+++ b/browser-test/src/admin_predicates.test.ts
@@ -471,18 +471,6 @@ describe('create and edit predicates', () => {
         'Screen 1',
       )
 
-      // Select all questions
-      for (const question of questions) {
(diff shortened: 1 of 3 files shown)


avaleske (approved):
I think I'm getting notified for every PR sent to civiform-developers but since I'm here:
Do we want to do anything slightly clever with this carousel to ensure it's clear that it's scrollable and some cards might be off-screen? Like by adjusting the padding so part of a card is always peaking in from offscreen?

Or somehow forcing the scroll bar to be visible?

PR #5695, opened 2023-09-29
Create filter to store a session id.

yotommy:
### Description

Add a filter that stores a session id in the browser session.

For OIDC logout, we want to store ID tokens in the database for an applicant. If they are logged in from more than one browser, then we need to index the ID tokens so that the proper one can be retrieved at logout time. We will use this new session key for the key.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [ ] Extended the README / documentation, if necessary

### Issue(s) this completes

Relates to #4359.

Issue #4359, opened 2023-03-10
Standardize Logout Process


michaelzetune:
## Current State

Pac4j handles the logout process for us with [`OidcLogoutActionBuilder`](https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/logout/OidcLogoutActionBuilder.java). Among other things, this class creates a Logout action that includes an ID token as a query param (`id_token_hint`). Importantly, this is part of the framework code that is supposed to handle a complex piece of authentication *for us*.

We override this class with [`CiviformOidcLogoutActionBuilder`](https://github.com/civiform/civiform/blob/main/server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java), which overrides `getLogoutAction`. You can see the diff between these [here](https://www.diffchecker.com/fTivK5qk/). Notably, the logic is *mostly* duplicated. The differences are:

1. We use a [`CustomOidcLogoutRequest`](https://github.com/civiform/civiform/blob/main/server/app/auth/oidc/CustomOidcLogoutRequest.java) instead of the standard `LogoutRequest` that pac4j uses. I've simplified this class in earlier PRs, but the reason for its existence at this point is (1) to allow us to insert a custom post-logout redirect URL and (2) to fix a URI fragment issue for Seattle.
2. We allow a `postLogoutRedirectParam` to be used from the application config (retrieved in the `CiviformOidcLogoutActionBuilder` constructor), which later gets passed into the `CustomOidcLogoutRequest`.
3. The `OidcLogoutActionBuilder::getLogoutAction` takes as a third parameter a `UserProfile`. In the standard implementation, this is always an `instanceof` `OidcProfile` - and this fact is really an implementation detail of pac4j. However, we expect the in our overidden implementation for the `UserProfile` to be an `instanceof` `CiviFormProfileData`. As a consequence, we cannot replicate the logic to get the ID token from the `OidcProfile` [the way that pac4j does](https://github.com/pac4j/pac4j/blob/c3ef714701086053e7be95cf93f51f995d9df643/pac4j-oidc/src/main/java/org/pac4j/oidc/logout/OidcLogoutActionBuilder.java#L53) in the base implementation.

## Next Steps

[...]

PR #5772, opened 2023-10-11
Inject Providers into ErrorHandler

yotommy:
### Description

Replace some injected instances with providers in `ErrorHandler`.

While developing [Create filter to store a session id](https://github.com/civiform/civiform/pull/5695), the `test_prod` check was failing when trying to start the server.

The cause was that PR added a `Filter`, for which the constructor took a `Materializer` as a parameter. This caused the `Materializer` to be instantiated earlier than it would have otherwise, which causes it to create an `ErrorHandler`. This class has some injected dependencies that themselves have transient dependencies on the database, which is not yet initialized.

The fix is to inject `Provider`s for those dependencies so that they can be dereferenced after server startup.

#### General

### Issue(s) this completes

Related to #4359

### DIFF
--- a/server/app/controllers/ErrorHandler.java
+++ b/server/app/controllers/ErrorHandler.java
@@ -45,8 +45,8 @@ import views.errors.NotFound;
 @Singleton
 public class ErrorHandler extends DefaultHttpErrorHandler {
 
-  private final InternalServerError internalServerErrorPage;
-  private final NotFound notFoundPage;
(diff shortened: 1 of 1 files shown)


yotommy:
* main
  * **PR #5772** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5772" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/5772?utm_source=stack-comment).

PR #5776, opened 2023-10-12
Turn on version caching for browser tests

dkatzz:
### Description

Enable version caching for browser tests. This involves clearing when the database is emptied during the browser test run.  We use the AsyncCacheApi to clear the cache, since it has a removeAll function. We also sometimes call to getVersion or Question data not using the cache during the publishing flow, to ensure we aren't caching the version right as it becomes active.

#### General

### Issue(s) this completes

#5845


server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -93,11 +110,34 @@ public Result clear() {
...
+      clearCache();
+    }
     resetTables();
     return redirect(routes.DevDatabaseSeedController.index().url())
         .flashing("success", "The database has been cleared");
   }
 
+  /** Remove all content from the cache. */
+  public void clearCache() {
+    if (!isDevOrStaging) return;
```
bion, line 123: **nitpick:** 
```suggestion
    if (!isDevOrStaging) {
      return;
    }
```

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -93,11 +110,34 @@ public Result clear() {
...
+   * clear.
+   */
+  private void clearVersionCache(SyncCacheApi cache, Version activeVersion) {
+    for (int num = 1; num <= activeVersion.id; num++) {
+      if (cache.get(String.valueOf(num)).isPresent()) {
+        cache.remove(String.valueOf(num));
+      }
+    }
+  }
+
```
bion, line 140: **suggestion (blocking):** please move this logic to an existing or new class focused on encapsulating caching behavior

bion (changes requested):

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
dkatzz: I realized I can use the AsyncCacheApi to do this and I don't have to change other references, so updated to do that!

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -93,11 +105,23 @@ public Result clear() {
...
+    if (settingsManifest.getVersionCacheEnabled()) {
+      clearCache();
+    }
     resetTables();
     return redirect(routes.DevDatabaseSeedController.index().url())
         .flashing("success", "The database has been cleared");
   }
 
+  /** Remove all content from the cache. */
+  public void clearCache() {
```
bion, line 117: **nitpick (blocking):** 
```suggestion
  private void clearCache() {
```

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -93,11 +105,23 @@ public Result clear() {
...
+    if (!isDevOrStaging) {
+      return;
```
bion, line 119: **nitpick:** this is already guarded in the action before this method gets called

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -93,11 +105,23 @@ public Result clear() {
...
+    }
+    programsByVersionCache.removeAll();
+    questionsByVersionCache.removeAll();
```
bion, line 122: **suggestion (blocking):** call `join()` on these to ensure they complete before the request finishes, otherwise errors won't get surfaced to the caller

bion (changes requested):

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
dkatzz: Yeah, that's true - removed!

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
dkatzz: Good call - updated!

bion (approved):

PR #5804, opened 2023-10-16
Update question search to only look at relevant question text.

### LINKED ISSUE #4157: Modify question search to only search pertinent fields
**Describe the bug**
Currently, the question search feature checks all text in the cf-question-bank-element class, which includes things like the "Add" button. Ideally it would only search pertinent fields like the question text, admin ID, etc.

**To Reproduce**
Sign in as admin, go to the question bank in program edit page. Search "add", for example, and note that all questions still show up since they all have an add button.

caitlinshk:
### Description

Update question search to only look at relevant question text.

#### General

#### User visible changes

(1) As an admin, open the "Add questions to program" page. Search for "add" and "admin" and verify that not all questions show up (e.g. for "Add", the "What is your address?" question should still appear, but not all questions should appear). Search for other keywords and verify the matching questions show up.

(2) As an admin, open the question bank page. Verify the old filtering behavior -- filter matches any text within the question element -- still works. For example, searching "draft" should show all draft questions. (Applying this "relevant filter text" idea to the question bank seemed more complicated because that relevant text should likely also include things like draft vs. active, # programs included, etc.)

### Issue(s) this completes

Fixes #4157

### DIFF
--- a/browser-test/src/admin_program_creation.test.ts
+++ b/browser-test/src/admin_program_creation.test.ts
@@ -393,7 +393,7 @@ describe('program creation', () => {
       ])
   })
 
-  it('filter questions in question bank based on text', async () => {
+  fit('filter questions in question bank based on text', async () => {
     const {page, adminQuestions, adminPrograms} = ctx
 
     await loginAsAdmin(page)
@@ -410,6 +410,8 @@ describe('program creation', () => {
     await adminPrograms.addProgram(programName)
     await adminPrograms.editProgramBlock(programName)
     await adminPrograms.openQuestionBank()
+
+    await page.pause();
     await page.locator('#question-bank-filter').fill('fi')
     expect(await adminPrograms.questionBankNames()).toEqual(['first question'])
     await page.locator('#question-bank-filter').fill('se')
--- a/browser-test/src/admin_program_creation.test.ts
+++ b/browser-test/src/admin_program_creation.test.ts
@@ -369,7 +369,7 @@ describe('program creation', () => {
     ])
   })
 
-  it('all questions shown on question bank open', async () => {
+  it('all questions shown on question bank before filtering', async () => {
     const {page, adminQuestions, adminPrograms} = ctx
 
     await loginAsAdmin(page)
--- a/browser-test/src/admin_program_creation.test.ts
+++ b/browser-test/src/admin_program_creation.test.ts
@@ -393,7 +393,7 @@ describe('program creation', () => {
       ])
   })
 
-  fit('filter questions in question bank based on text', async () => {
+  it('filter questions in question bank based on text', async () => {
     const {page, adminQuestions, adminPrograms} = ctx
 
     await loginAsAdmin(page)
@@ -411,11 +411,12 @@ describe('program creation', () => {
     await adminPrograms.editProgramBlock(programName)
     await adminPrograms.openQuestionBank()
 
-    await page.pause();
     await page.locator('#question-bank-filter').fill('fi')
     expect(await adminPrograms.questionBankNames()).toEqual(['first question'])
+
     await page.locator('#question-bank-filter').fill('se')
     expect(await adminPrograms.questionBankNames()).toEqual(['second question'])
+
     await page.locator('#question-bank-filter').fill('')
     expect(await adminPrograms.questionBankNames()).toEqual([
       'second question',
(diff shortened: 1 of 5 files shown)


caitlinshk:
The failing test revealed that there's another place where we filter questions that I didn't update correctly. Working on a fix now.

browser-test/src/admin_program_creation.test.ts
```diff
@@ -386,21 +386,204 @@ describe('program creation', () => {
...
+    await adminPrograms.openQuestionBank()
+
+    await page.locator('#question-bank-filter').fill('qf-desc')
+    expect(await adminPrograms.questionBankNames()).toEqual(['first question'])
+
+    await page.locator('#question-bank-filter').fill('qs-desc')
+    expect(await adminPrograms.questionBankNames()).toEqual(['second question'])
+  })
+
+  /**
```
dkatzz, line 506: **praise:** Thanks for adding these extra tests!!

dkatzz (commented):
Will take another look once the other test gets fixed, but LGTM overall! Great job!

browser-test/src/admin_program_creation.test.ts
```diff
@@ -369,7 +369,7 @@ describe('program creation', () => {
     ])
   })
 
-  it('filter questions in question bank', async () => {
+  it('all questions shown on question bank open', async () => {
```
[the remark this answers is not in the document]
caitlinshk: I like it, done.

caitlinshk:
I updated `questionBank.ts` to fall back to using the `questionElement.innerText` if the data attribute wasn't there. The question bank displayed on the admin "All questions" page has some additional information that I think is still worth including as filter criteria, so I didn't want to update that page to include the data attribute.

browser-test/src/admin_program_creation.test.ts
```diff
@@ -386,21 +386,204 @@ describe('program creation', () => {
...
+    const programName = 'Test program'
+    await adminPrograms.addProgram(programName)
+    await adminPrograms.editProgramBlock(programName)
+    await adminPrograms.openQuestionBank()
+
+    await page.locator('#question-bank-filter').fill('admin note')
+
+    expect(await adminPrograms.questionBankNames()).toEqual([])
+  })
+
```
bion, line 586: **praise:** thank you for the thorough testing!

server/app/assets/javascripts/questionBank.ts
```diff
@@ -139,11 +140,25 @@ class QuestionBankController {
     )
     questions.forEach((question) => {
       const questionElement = question as HTMLElement
-      const questionContents = questionElement.innerText
+
+      let questionFilterText
```
dkatzz, line 144: **thought:** Not sure if this would make it cleaner, but you can also do something like:

```
const questionFilterText = questionElement.hasAttribute(
          QuestionBankController.RELEVANT_FILTER_TEXT_DATA_ATTR,
        ) ? questionElement.getAttribute(
          QuestionBankController.RELEVANT_FILTER_TEXT_DATA_ATTR,
        ) : questionElement.innerText
```

server/app/assets/javascripts/questionBank.ts
```diff
@@ -139,11 +140,25 @@ class QuestionBankController {
...
+        )
+      } else {
+        questionFilterText = questionElement.innerText
+      }
+
       questionElement.classList.toggle(
         'hidden',
         filterString.length > 0 &&
-          !questionContents.toUpperCase().includes(filterString),
+          (questionFilterText == null ||
```
dkatzz, line 160: **question:** In what case might this be null?

caitlinshk: `getAttribute` is documented as possibly returning `null`. Even though I check it first using `hasAttribute`, the compiler doesn't know that and still requires a null check.

We could also rewrite this as:
```
let questionFilterText = questionElement.getAttribute(
          QuestionBankController.RELEVANT_FILTER_TEXT_DATA_ATTR,
        )
if (questionFilterText == null) {
  questionFilterText = questionElement.innerText
}
```

To (I think) remove the null check. Thoughts?

(Typescript doesn't have a `?:` operator like [Kotlin](https://kotlinlang.org/docs/null-safety.html#elvis-operator) right?)

dkatzz: Oh got it, makes sense! There is actually something called a nullish coalescing operator that looks at the left side and if its null it will return what is on the right side. So in this case it would be:

```
const questionFilterText = questionElement.getAttribute( QuestionBankController.RELEVANT_FILTER_TEXT_DATA_ATTR, ) ?? questionElement.innerText
```

That seems like probably the best option to me but let me know what you think! 

Here are some examples of it being used: https://github.com/search?q=org%3Aciviform%20%22%3F%3F%22&type=code

caitlinshk: `??` is exactly what I was hoping for, thanks! Done.

server/app/assets/javascripts/questionBank.ts
caitlinshk: Resolved using the `??` operator.

dkatzz (approved):
LGTM - Great job!

Issue #5054, opened 2023-06-12
"Logged in as <name>" on auth header shows applicant's info when TI impersonating applicant


michaelzetune:
**Describe the bug**
When a TI is applying on behalf of an applicant, we show `(applying as: <applicant name>)` on the header (expected), but we also show `Logged in as <applicant name>` on the top right. 

**To Reproduce**

1. Go to TI portal
2. Go to an applicant's dashboard

**Expected behavior**

The top right should show the TI's auth info instead.

**Screenshots**

<img width="1980" alt="image" src="https://github.com/civiform/civiform/assets/30369272/95280600-66b4-4373-80dd-d2188c1e6e4a">

**Desktop (please complete the following information):**

- OS: [e.g. iOS]: macOS
- Browser [e.g. chrome, safari]: Chrome
- Version [e.g. 22]: 114.0.5735.90

**Additional context**

None.

PR #5853, opened 2023-10-25
Fix TI seeing client's name in top nav login state

bion:
### Description

This fixes a bug wherein the Trusted Intermediary incorrectly sees their client's (applicant's) name in the upper right corner for "Logged in as _____". There is already a banner at the top stating "You are applying for [first], [last]".

This bugfix displays the TI's email address instead. Why not their name? We have no guarantee that the TI's name is available, most of the time it won't be. Their email address should always be available though.

**Technical note: consistent email on CiviFormProfile**

A tricky aspect of fixing this was making the TI's email address available in `ApplicantLayout`. The available state in the layout class includes the TI's profile and the applicant's `applicant` record. Adding new params to the already param-heavy render method would have involved updating many call sites. One option could have been to load the TI's `account` record using their profile, but this would have necessitated making a blocking DB call in the middle of the view code.

The alternative taken here involves updating CiviForm's profile creation logic to ensure all `CiviFormProfileData` (the data serialized into the pac4j session) have the email address for their account stored under the `email` attribute. This was previously inconsistent since different identity providers store the email under different attributes, e.g. IDCS stores it under `user_emailid`.

By ensuring all authenticated profiles have an email under `email` we can use the `getEmail()` method on the parent class, `CommonProfile`.

#### General

- [x] Test with IDCS on Seattle staging
- [ ] Test with auth0 on aws-staging

1. Login as a Trusted Intermediary using an identity provider (NOT using the dev tools link)
1. Navigate to a client's "Applicant Dashboard"
1. Confirm that the Trusted Intermediary's email address appears in the upper right corner instead of the client's (applicant's) name

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/5054


server/app/auth/CiviFormProfile.java
```diff
@@ -152,6 +151,8 @@ public CompletableFuture<Void> setAuthorityId(String authorityId) {
    * @return the future of the database operation
    */
   public CompletableFuture<Void> setEmailAddress(String emailAddress) {
+    profileData.setEmail(emailAddress);
```
swatkat1, line 154: We have a number of cases where the below error ""Profile already contains an email address:" is thrown in SEA. What will happen to the profileData if an existing email address is added? Will the object be garbage collected?

bion: > What will happen to the profileData if an existing email address is added?

The profile's lifespan is the same as the session's, so if their session is destroyed the data stored in the profile will be gone with it.

server/app/auth/ProfileFactory.java
```diff
@@ -99,6 +100,7 @@ private CiviFormProfileData create(Role[] roleList) {
     return p;
   }
 
+  @VisibleForTesting
```
yotommy, line 103: **suggestion (blocking):** Prefer `@RestrictedApi` since `@VisibleForTesting` is only informational.

server/app/auth/oidc/CiviformOidcProfileCreator.java
```diff
@@ -114,16 +114,17 @@ protected CiviFormProfileData mergeCiviFormProfile(
         .forEach(role -> civiformProfile.getProfileData().addRole(role));
     adaptForRole(civiformProfile, roles);
 
+    String emailAddress =
+        getEmail(oidcProfile)
+            .orElseThrow(
```
yotommy, line 119: **question (blocking):** Is throwing an exception in this case new behavior?

server/app/auth/oidc/CiviformOidcProfileCreator.java
```diff
@@ -114,16 +114,17 @@ protected CiviFormProfileData mergeCiviFormProfile(
...
+                () -> new InvalidOidcProfileException("Unable to get email from profile."));
+
     // If the civiformProfile is a trusted intermediary, bypass remaining merging because
     // we don't want to actually merge the guest profile into theirs.
     if (isTrustedIntermediary(civiformProfile)) {
-      return civiformProfile.getProfileData();
+      return civiformProfile.getProfileData().setEmail(emailAddress);
```
yotommy, line 125: **suggestion (non-blocking):** Add a comment that explains why you are setting the email address here.

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -386,17 +386,34 @@ private DivTag authDisplaySection(
...
-      String logoutLink = org.pac4j.play.routes.LogoutController.logout().url();
-      return outsideDiv.with(
-          div(loggedInAsMessage).withClasses("text-sm"),
-          a(messages.at(MessageKey.BUTTON_LOGOUT.getKeyName()))
-              .withId("logout-button")
-              .withHref(logoutLink)
-              .withClasses(ApplicantStyles.LINK));
     }
+
+    String accountIdentifier =
```
yotommy, line 391: **suggestion (non-blocking):** Use a comment to explain why we are displaying email address for TIs rather than name.

server/app/auth/CiviFormProfileData.java
```diff
@@ -26,6 +27,26 @@ public CiviFormProfileData(Long accountId) {
     this.setId(accountId.toString());
   }
 
+  /**
+   * Sets the "canonical" email field in the profile data. Some identity providers use non-standard
+   * attribute names for email. We use the attribute name provided by pac4j here to ensure all
+   * profiles store the email in the same place to make it is accessible via {@code
+   * CommonProfile.getEmail()}.
+   */
+  public CiviFormProfileData setEmail(String email) {
```
yotommy, line 36: **praise:** This is a clever approach: using a well-known location in the `pac4j` profile.

yotommy (commented):
This is a clever fix for the bug.

Did you consider storing the TI email address in a session cookie? I'm not sure if this would be easier to implement, but it might be easier to understand. It would also incur a slight penalty in cookie size compared to the case where the `UserProfile` already contained the email address in the canonical field.

server/app/auth/CiviFormProfile.java
swatkat1: Line 161 raises ProfileMergeConflictException. Should profileData.setEmail() happen after it?

server/app/auth/oidc/CiviformOidcProfileCreator.java
bion: No

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -386,17 +386,34 @@ private DivTag authDisplaySection(
...
+            .withClasses(ApplicantStyles.LINK));
+  }
+
+  private String tiEmailForDisplay(CiviFormProfile profile) {
+    // CommonProfile.getEmail() can return null, so we guard that with a generic
+    // display string.
+    String email =
+        Optional.ofNullable(profile.getProfileData().getEmail()).orElse("Trusted Intermediary");
+
+    // To ensure a consistent string with browser snapshots, we override the
```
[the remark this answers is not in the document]
bion: I'm unfortunately not aware of a tool that will enforce "if this changes, also change that", comments are the best we can do. I think browser snapshot tests will also fail if one of these gets changed without the other though.

server/app/auth/oidc/CiviformOidcProfileCreator.java
bion: Done

server/app/views/applicant/ApplicantLayout.java
bion: Done

server/app/auth/CiviFormProfileData.java
bion: thanks!

server/app/views/applicant/ApplicantLayout.java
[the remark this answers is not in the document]
yotommy: Maybe in the future we could create a helper class that creates a fake TI and validate the email address pattern.

yotommy (approved):
Nice!

(There is still one browser test to fix)

server/app/auth/CiviFormProfile.java
bion: I don't *think* that will change much since if there's an exception the whole session is probably broken, but there's no harming in setting it later on the idea we only update state anywhere after performing the check 👍🏻.

swatkat1 (approved):

Issue #5845, opened 2023-10-24
Turn on caching capabilities for browser tests


dkatzz:
**Describe the solution you'd like**
Enable caching for browser tests, but ensure that the cache is cleared when the database is seeded in between test runs.

PR #5863, opened 2023-10-26
Enable program cache in browser tests

dkatzz:
### Description

Enable program cache in browser tests. When looking up the program, we check if there is a draft version and in that case we don't use the cache, so an incorrect version doesn't get loaded. In the future, we can further optimize this by either: 

1. Only looking up data from the cache if we're in the applicant flow and calling a separate function from the admin flow that doesn't use the cache if there is a draft version
2. Checking if the user is an admin when attempting to lookup a program and if so then check if there is a draft version and only use the cache if so, and if not always using the cache

#### General

### Issue(s) this completes

Related to #5845

### DIFF
--- a/server/app/repository/VersionRepository.java
+++ b/server/app/repository/VersionRepository.java
@@ -145,7 +145,7 @@ public final class VersionRepository {
           .forEach(
               program -> {
                 draft.addProgram(program);
-                removeCacheForProgram(String.valueOf(program.id));
+                removeCacheForProgram(program.id);
               });
 
       // Associate any active questions that aren't present in the draft with the draft.
@@ -178,7 +178,7 @@ public final class VersionRepository {
               programToDelete -> {
                 draft.removeTombstoneForProgram(programToDelete);
                 draft.removeProgram(programToDelete);
-                removeCacheForProgram(String.valueOf(programToDelete.id));
+                removeCacheForProgram(programToDelete.id);
               });
 
       // Move forward the ACTIVE version.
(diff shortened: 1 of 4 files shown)


server/app/repository/VersionRepository.java
```diff
@@ -145,10 +145,7 @@ private Version publishNewSynchronizedVersion(PublishMode publishMode) {
           .forEach(
               program -> {
                 draft.addProgram(program);
-                if (settingsManifest.getProgramCacheEnabled()) {
-                  versionsByProgramCache.remove(String.valueOf(program.id));
-                  programCache.remove(String.valueOf(program.id));
-                }
+                removeCacheForProgram(String.valueOf(program.id));
```
bion, line 148: **suggestion (non-blocking):** instead of calling `String.valueOf` in several places, overload `removeCacheForProgram` so it accepts taking a long.

server/app/repository/VersionRepository.java
```diff
@@ -507,6 +510,13 @@ private void removeCacheForVersion(String versionKey) {
     }
   }
 
+  private void removeCacheForProgram(String programKey) {
```
bion, line 513: **suggestion (non-blocking):** `uncacheProgram`? idk

bion (approved):

server/app/repository/VersionRepository.java
dkatzz: Updated!

server/app/repository/VersionRepository.java
dkatzz: We also have removeCacheForVersion, so I feel like if we change this method name we should change that one too to uncacheVersion. I'll do both of these in a separate PR.

Issue #5532, opened 2023-09-05
(WCAG 2.1) Change in Focus outline is barely visible on input fields


dkatzz:
OBSERVED RESULTS:

- The change from black to blue focus outline is hard to see, especially on comboboxes where there's no cursor like there is in input text boxes.
- Applicable for all form fields.
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/1ffe9cd9-ee1d-402e-8465-6be31d640a85)
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/cc9c1111-9ac9-4b21-b4cf-f1a0bc276507)

EXPECTED RESULTS:

- The focus indicator outline should meet a minimum contrast ratio of 3:1 to the background.
- Keyboard focus indicator should be visually distinct. In this case, implementing the keyboard focus indicator larger than the default black border can improve visibility.

STEPS TO REPRODUCE:

- Navigate to [URL](https://demo.civiform.dev/)
- Navigate to and activate the ‘Apply’ for the Comprehensive sample program.
- Navigate to and activate “What is your address?”.
- Navigate to each form field using keyboard TAB or SHIFT TAB key.
- Observe the focus outline for input fields and comboboxes.

ADDITIONAL INFORMATION:

- [ W3org Reference - 2.4.11: Focus Appearance (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html)

ENVIRONMENT DETAILS:

- Device(s) (brand, OS version, test key or release key): Windows 10 Enterprise (64-bit),MacOS Ventura Version 13.4
- Screen reader name and version: JAWS 2022.2211.7, NVDA 2022.3, VoiceOver
- Browser name and version: Chrome Version 113.0.5672.93

melanie-exygy:
Once #5662 is completed, our focus color will be #2491ff, which is the USWDS default.  This will give us a contrast ratio of 3.19 in respect to the white background, which is acceptable.

![image](https://github.com/civiform/civiform/assets/134556723/eabed08a-f7e1-4623-9e8f-a2ce5a63fd42)

I propose closing this issue once #5662 is complete and adding a note to #5533 to ensure that the focus color is not changed to one that doesn't contrast enough with the white background.

Issue #5533, opened 2023-09-05
(WCAG 2.1) Focus outline color does not meet minimum contrast ratio of 3:1 


dkatzz:
OBSERVED RESULTS:

- The change in focus indicator / focus outline color for the following buttons do not meet minimum contrast ratio of 3:1 (Foreground: ##000000 Background: #113F9F, Contrast Ratio: 2.23).
- The black outline is barely visible to the user.
- Apply buttons, Continue button, Previous button, Save and next button, Add household member
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/ac1b4823-3aa7-4
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/93edac91-286d-40f3-861c-bbc0dfdca4c0)
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/5e5aac3a-72dc-4437-9574-eea265ce4533)
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/5f5c2ff4-8d55-499b-a875-217b85af9d81)
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/3ca35c34-79dd-4dbc-9ddf-fec0cb31a801)
- ![Screenshot](https://github.com/civiform/civiform/assets/86739416/14740e86-51e5-42d2-b0cf-4a80100b2230)

EXPECTED RESULTS:

- The focus indicator outline should meet a minimum contrast ratio of 3:1 to the background. Keyboard focus indicator should be visually distinct..

STEPS TO REPRODUCE:

- Navigate to [URL](https://demo.civiform.dev/)
- Navigate to the ‘Apply’ links using keyboard TAB or SHIFT TAB key.
- Observe the focus outline color do not meet minimum contrast ratio of 3:1

ADDITIONAL INFORMATION:

- [ W3org Reference - 2.4.11: Focus Appearance (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html)

ENVIRONMENT DETAILS:

- Device(s) (brand, OS version, test key or release key): Windows 10 Enterprise (64-bit),MacOS Ventura Version 13.4
- Screen reader name and version: JAWS 2022.2211.7, NVDA 2022.3, VoiceOver
- Browser name and version: Chrome Version 113.0.5672.93OBSERVED RESULTS:OBSERVED RESULTS:

melanie-exygy:
[...]

PR #5870, opened 2023-10-26
Import USWDS component library

### LINKED ISSUE #5662: Import USWDS component library
Allow CiviForm to make use of U.S Web Design System (USWDS) components by importing them rather than recreating them from scratch.  By doing this, we will save significant development hours.  It will also allow us to consistently follow UX best practices and recommendations, notably, in the areas of accessibility and responsiveness.  This is in contrast to each component being created separately, with the chance that a developer may forget to include certain accessibility features. 

**Additional context**
Follow the approved design here:
[TDD](https://docs.google.com/document/d/1EzBlasVoe9C3D0_AquYOsGy5FYubIHCpPeZFSm5ZGYk/edit)

1. Import USWDS node module.
2. Modify Webpack config to convert Sass files into CSS and bundle the USWDS files.
3. Add USWDS stylesheet and script to our base HTML.

**Done when**
The USWDS library is available in CiviForm

melanie-exygy:
### Description

This PR pulls in the [U.S Web Design System (USWDS)](https://designsystem.digital.gov/about/) component library.  Once this library is available to us, we will be able to easily use any USWDS out-of-the-box components or styles.  With this PR, the only style change is the color and thickness of the focus box. 

## Release notes

This PR pulls in the [U.S Web Design System (USWDS)](https://designsystem.digital.gov/about/) component library.  Once this library is available to us, we will be able to easily use any USWDS out-of-the-box components or styles. 

#### General

#### User visible changes

1.  Open the developer tools in your browser.
2. Go to the Network tab.
3. Load the CiviForm app.
4. Note the presence of `uswds.min.css`, `uswds.bundle.js` and `uswds-init.min.js`
5. Also, use the tab key and notice the new focus styles:
image

### Issue(s) this completes

Fixes #5662; #5886


server/app/views/components/Accordion.java
```diff
@@ -57,7 +57,7 @@ public DivTag getContainer() {
             .withClasses(
                 ReferenceClasses.ACCORDION_BUTTON,
                 "transition-all",
-                "duration-300",
+                "duration-100",
```
melanie-exygy, line 60: I decreased the spinning caret transition time on the accordion to make tests less flaky and because I saw no visual improvement with a slower caret spin.

bion:
@melanie-exygy does this PR introduce visual changes apart from the selected state?

Also does this PR fix either of these by chance?

- https://github.com/civiform/civiform/issues/5532
- https://github.com/civiform/civiform/issues/5533

server/app/assets/stylesheets/styles.css
```diff
@@ -14,6 +14,10 @@
     @apply text-xl;
   }
 
+  body {
+    color: rgb(0 0 0);
+  }
+
```
bion, line 20: suggestion (blocking): please add a comment clarifying the need for this rule

server/app/assets/stylesheets/uswds/_uswds-theme.scss
```diff
@@ -0,0 +1,19 @@
+/*
+----------------------------------------
+USWDS with settings overrides
+----------------------------------------
+Uncomment the following lines and add a list of changed settings
+in the form $setting: value,
```
bion, line 6: suggestion (blocking): nothing is commented below, so this comment seems inappropriate

server/app/views/HtmlBundle.java
```diff
@@ -191,6 +192,9 @@ private FooterTag renderFooter() {
     }
     footerTag.with(viewUtils.makeLocalJsTag(jsBundle.getJsPath()));
 
+    // Add USWDS scripts
+    footerTag.with(viewUtils.makeLocalJsTag(USWDS_FILEPATH));
```
bion, line 196: question (non-blocking): what's the difference between `dist/uswds.bundle` and `javascripts/uswds/uswds-init.min` and why are they added the the DOM in different places?

server/package-lock.json
bion, line 1: quite a relief to discover this file is the reason for the PR being >1k LOC 😂

server/postcss.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
bion, line 1: suggestion (non-blocking): consider adding a comment here explaining what this is doing and why we need it

server/project/TypescriptBuilder.scala
```diff
@@ -54,18 +56,18 @@ object TypescriptBuilder extends AutoPlugin {
...
         cacheDir,
         streamsVal.log
       )
-      // transform compiles JS files to PathMapping and add "javascripts" as part of the file name.
-      // That way JS files fille be added in /assets/javascripts/*.js in final output.
+      // transform compiles JS files to PathMapping and add "dist" as part of the file name.
+      // That way JS files fille be added in /assets/dist/*.js in final output.
       val jsFiles: Seq[PathMapping] =
-        compiledJsFiles.map({ f => (f, "javascripts/" + f.getName) })
+        compiledJsFiles.map({ f => (f, "dist/" + f.getName) })
```
bion, line 70: question (blocking): why rename the output from `javascripts` to `dist`?

bion (changes requested):
Thanks for getting this figured out! Excited to use it!

melanie-exygy:
> @melanie-exygy does this PR introduce visual changes apart from the selected state?

No, all of the visual changes are because of the new focus styling.  I was able to avoid any other visual differences by adjusting the uswds settings in `server/app/assets/stylesheets/uswds/_uswds-theme.scss` and by adding the default tailwind body color to our `styles.css` so that it has precedence over the uswds color themes.  I'll add comments in those files explaining this.
> 
> Also does this PR fix either of these by chance?
> 
> * [(WCAG 2.1) Change in Focus outline is barely visible on input fields #5532](https://github.com/civiform/civiform/issues/5532)
> * [(WCAG 2.1) Focus outline color does not meet minimum contrast ratio of 3:1  #5533](https://github.com/civiform/civiform/issues/5533)

Please see the comments that I added to both of those issues.  The new USWDS focus color does increase the contrast ratio, but not enough to meet the 3:1 standard.  I will close #5532 once this PR has been merged, but I suggest keeping #5533 so that we can change our focus color to be more contrasting.  It will be a very easy change, but we'll need to get input from UX.

server/app/assets/stylesheets/styles.css
melanie-exygy: Great callout.  I realize now that I need to add more comments in general.  Will add one here.

server/app/assets/stylesheets/uswds/_uswds-theme.scss
melanie-exygy: Thanks, will fix these comments.

server/app/views/HtmlBundle.java
melanie-exygy: `uswds-init.min` is a small JavaScript file (less than 1 KB minified) that helps the browser know if the USWDS JavaScript library is loading properly.  `uswds.bundle` is the main uswds JavaScript file.  They are in two different places in the DOM because I followed the recommendations here:  [uswds docs](https://designsystem.digital.gov/documentation/developers/#using-uswds-css-and-javascript-in-your-project-2).

Note that I didn't opt to bundle `uswds-init.min` with Webpack because it's already bundled and is very small.  I figured that it was unlikely to change, so copied the minified file directly into our project.  Does that seem reasonable to you?

server/postcss.config.js
melanie-exygy: Good idea!  Will add a comment here too.

server/project/TypescriptBuilder.scala
melanie-exygy: I renamed it because the output is no longer only JS.  The compiled CSS file is output there too.  `dist` seemed like an appropriate name given that it's the default Webpack output file name.  It also seemed like a clearer name for bundling output because it differentiates it from source code.

I attempted to find a way to split the output into two different directories, but wasn't able to find a way to configure that in Webpack.  It also seemed unnecessary.

server/webpack.config.js
```diff
@@ -1,13 +1,31 @@
...
+
 module.exports = {
   mode: 'production',
   devtool: 'source-map',
   stats: 'errors-only',
   module: {
     rules: [
+      {
+        test: /\.scss$/,
+        use: [
```
melanie-exygy, line 11: Loaders run from the bottom up.  Here is what these loaders are doing:
- sass-loader:  Converts Sass into CSS.
- postcss-loader:  Adds vendor prefixes (like `-webkit-`, `-moz-` or `-ms-`) so that styles are compatible with all browsers, even older ones.  USWDS strongly recommends autoprefixing.  See the highlighted note [here](https://designsystem.digital.gov/documentation/getting-started/developers/phase-two-compile/#loading-uswds-in-your-sass-entry-point-2).
- css-loader:  Collects all the CSS which resulted from the Sass compilation above and puts it into a single string.
- MiniCssExtractPlugin:  Puts that CSS string into a file.  See the plugin filename setting on line 48.

Note that all of these loaders are 3rd-party packages which don't come with Webpack.  That's why I added them to our dependencies in` package.json`.

server/project/TypescriptBuilder.scala
```diff
@@ -19,9 +19,11 @@ import scala.sys.process.{Process, ProcessLogger}
   * been updated since 2018 and uses Typescript 2 while current Typescript
   * version is 4.
   *
-  * Our custom plugin invokes `tsc` compiler directly. `tsconfig.json` is setup
-  * to compile code from `app/assets/javascripts` directory and outputs result
-  * to `target/web/typescript/javascripts` folder.
+  * Our custom plugin invokes `tsc` compiler directly through the Webpack
```
melanie-exygy, line 22: I considered renaming this file, the task name and some of the variables since it now builds CSS as well as converting Typescript to JS.  I opted against it since compiling TS is still the main objective and the CSS is "extra".  I was also hesitant to rename because this PR is already adding a big change and I didn't want to further complicate it.  

Please let me know if you think this is acceptable or if the names will be confusing.

server/package.json
```diff
@@ -16,12 +16,17 @@
     "@types/jest": "29.5.6",
     "@typescript-eslint/eslint-plugin": "6.7.5",
     "@typescript-eslint/parser": "6.7.5",
-    "autoprefixer": "10.4.16",
+    "autoprefixer": "^10.4.16",
```
melanie-exygy, line 19: All of these dev dependencies are for the Webpack bundling.  Please see my comment on `webpack.config.js` for more details on each one.

server/app/assets/stylesheets/uswds/_uswds-theme.scss
```diff
@@ -0,0 +1,20 @@
...
+USWDS with settings overrides
+----------------------------------------
+Customize USWDS styles here or prevent USWDS utilities (custom HTML classes
+scoped to a single CSS property) from being applied by setting 'output' to 'false'.
+See https://designsystem.digital.gov/documentation/settings/ and
+https://designsystem.digital.gov/utilities/.
+*/
+
+@use 'uswds-core' with (
+  $theme-font-path: '../../../../node_modules/@uswds/uswds/dist/fonts',
```
melanie-exygy, line 12: When the Webpack `sass-loader` compiles Sass into CSS, it doesn't update the font and image urls to be relative to the Sass entry point.  For us, that's `app/assets/stylesheets/uswds/styles.scss` [sass-loader documentation](https://designsystem.digital.gov/documentation/getting-started/developers/phase-two-compile/#loading-uswds-in-your-sass-entry-point-2).  The fonts and images are still in the uswds directory under node_modules.  So, I specify the correct paths here in the uswds settings.

server/project/TypescriptBuilder.scala
rockycodes: I think leaving keeping this class name is fine since, as you said, the CSS is "extra" and you mention it in the comment.

server/webpack.config.js
rockycodes: Great explanation! Thank you!

rockycodes (commented):
Just commenting for now until we can chat about it so I understand better. But overall I think this looks great!

server/app/assets/stylesheets/styles.css
bion: Thanks!

server/app/assets/stylesheets/uswds/_uswds-theme.scss
bion: Please move this github comment into a code comment here.

server/postcss.config.js
bion: Thanks!

server/project/TypescriptBuilder.scala
bion: How is the CSS "extra"? Will CiviForm not break if it fails to build CSS?

server/project/TypescriptBuilder.scala
bion: Thanks, that makes sense to me! Sounds like an apt rename. I think I disagree with the decision to leave this as `TypescriptBuilder` given it's now responsible for stylesheets as well.

server/webpack.config.js
bion: Please move this github comment into code comments

bion (changes requested):

server/project/TypescriptBuilder.scala
melanie-exygy: That's a good point, Bion.  What I meant is that Webpack is primarily a JavaScript bundler and most of the logic in the file is still around TypeScript/JS.  But, I'll rename the file so as not to cause any confusion.

server/app/views/JsBundle.java
```diff
@@ -27,6 +27,6 @@ public enum JsBundle {
   }
 
   public String getJsPath() {
-    return jsPath;
+    return "dist/" + jsPath;
```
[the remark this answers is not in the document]
rockycodes: We discussed and we need to add the `dist` here so that the admin and applicant bundle paths are correct. They previously were prepended by `javascripts` in `ViewUtils.java`

server/app/views/ViewUtils.java
```diff
@@ -75,10 +73,8 @@ public ScriptTag makeWebJarsTag(String assetsRoute) {
   /**
    * Generates an HTML link tag for loading the CSS file found at public/stylesheets/[filename].css.
```
rockycodes, line 74: Should the path in this comment be `public/dist/[filename].css`? I think you said "no" when we discussed offline, but I didn't understand why.

server/app/assets/javascripts/uswds/uswds-init.min.js
```diff
@@ -0,0 +1,19 @@
...
+    window.addEventListener(
+      'load',
+      function e() {
+        window.uswdsPresent &&
+          (clearTimeout(n), t(), window.removeEventListener('load', e, !0))
+      },
+      !0,
+    )
+})()
+//# sourceMappingURL=uswds-init.min.js.map
```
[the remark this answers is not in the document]
rockycodes: We discussed offline and this code is directly copied from USWDS, so I think leaving this line in is fine. We think it's meant to tell us where the map gets written to but we're not sure. 

Melanie opted to copy this code here instead of including it in the bundle since it's so small and likely won't change much, which I think is fine. Also the USWDS styles seem to load fine without it. She will add a comment with a link to where the source code for this file lives for context.

server/app/assets/stylesheets/uswds/styles.scss
```diff
@@ -0,0 +1,3 @@
+@forward 'uswds-theme';
+@forward 'uswds';
```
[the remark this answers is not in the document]
melanie-exygy: Good question!  Let me explain.  `@forward` is a Sass annotation that behaves like an import statement would in JS.  [Sass docs](https://sass-lang.com/documentation/at-rules/forward/).  So, when Webpack starts converting the Sass and bundling our CSS, it starts with the `app/assets/stylesheets/uswds/styles.scss` as the entry point (See the `webpack-config.js` entrypoint list) includes all the styles from the referenced files (`uswds-theme.scss`, `uswds-theme-custom-styles.scss` and even `node_modules/@uswds/uswds/dist/scss/stylesheets/uswds.scss`).  It is able to find that last file in the node modules because we told it to search this path in the sass-loader configuration in `webpack-config.js`:
image

server/app/assets/javascripts/uswds/uswds-init.min.js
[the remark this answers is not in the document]
melanie-exygy: I was curious about that commented-out line as well.  First things first, what is that map file that Webpack outputs alongside every bundled file?  The map file is the "decoder ring" for the minified JS file.  It allows the browser dev tools to show you the un-minified JS for debugging.  When Webpack creates the map file, it inserts this comment at the end of the related minified JS file so that the browser knows where to find the source map.

server/app/views/ViewUtils.java
melanie-exygy: You're right, I do need to change the path in that comment.  It should say "found at public/main/[path].css".  Here are the two paths it will receive as an argument:
- `stylesheets/tailwind`
- `dist/uswds.min`

`tailwind.css` ends up under` public/main/stylesheets` because when SBT builds it copies everything from these assets directories directly into `public/main`.
image

`uswds.min.css` end up under `public/main/dist` because it was bundled with Webpack and that is the output folder we specified.

bion (approved):

server/app/views/HtmlBundle.java
bion: Yes, thank you for the explanation

server/app/assets/stylesheets/uswds/styles.scss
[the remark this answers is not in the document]
rockycodes: Gotcha! Thank you for explaining :)

server/app/assets/javascripts/uswds/uswds-init.min.js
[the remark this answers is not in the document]
rockycodes: Very cool! Thanks for explaining and adding the comments in!

rockycodes (approved):

PR #5875, opened 2023-10-26
Always build prod with linux/amd64

nb1701:
As it turns out, we can build prod on M-series macs as long as we pass --platform=linux/amd64 in both stages. You can technically run it on macs too, though it won't run super well. It doesn't really work with the dev image, as compileIncremental gets all screwy trying to do its thing under emulation, so we should really only do this for prod images.


bion (approved):
WOMM

PR #5881, opened 2023-10-27
Adding some extra error trapping to aid in problem areas

gwendolyngoetz:
### Description
- Provides info on what scope the GITHUB_TOKEN needs.
- Adds extra error handling to give more feed back on artifact download failures

### DIFF
--- a/bin/sync-failed-browser-test-images
+++ b/bin/sync-failed-browser-test-images
@@ -138,7 +138,7 @@ if (($? != 0)); then
   readonly FOUND_FILE="$(find "${ARTIFACT_FILE}" -type f -size -30720c)"
 
   if (($? != 0)); then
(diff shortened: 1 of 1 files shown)


bin/sync-failed-browser-test-images
```diff
@@ -3,9 +3,12 @@
 # DOC: Sync failed github action browser-test snapshots with local system.
 # DOC: With no arguments this defaults to the current active branch or
 # DOC: pass in the first argument to specify a branch name
-
 source bin/lib.sh
 
+# bin/lib.sh sets this to -e, but we want to trap errors to provided
+# users with better feedback
+set +e
```
bion, line 10:
```suggestion
# bin/lib.sh sets this to -e, but we want to trap errors to provide
# users with better feedback
set +e
```

bion (approved):

PR #5882, opened 2023-10-27
Move a11y focus to modal on modal open.

### LINKED ISSUE #4196: Modal accessibility
**Describe the bug**
When a modal opens on the page, the focus does not shift to the modal and the user has to tab through the elements behind the modal before being able to interact with the modal. This could make for a confusing experience for folks using screen readers.

Also, the modal fails the accessibility test [here](https://github.com/civiform/civiform/blob/7f29b40800422ec8b854c7d2ac9ab783f9afa420/browser-test/src/support/index.ts#L476) with the message that not all content is contained within landmarks (https://dequeuniversity.com/rules/axe/4.6/region?application=axeAPI)

**To Reproduce**

1. Go to the block edit view
2. Use the tab button to navigate the page
3. Hit "Enter" when you have the "Delete screen" button selected
4. Use the tab button to continue navigating the page
5. You will have to tab through all elements behind the modal before getting to the modal

**Expected behavior**
When a user opens the modal and continues navigating the page via manual tabbing the elements on the modal should be the next ones selected.

**Screenshots**
You can see that "Edit visibility condition" is selected when I hit "tab" after opening the modal
image

- OS: iOS
- Browser Firefox
- Version 109.0.1

**Additional context**

caitlinshk:
### Description

This change (1) Moves a11y focus to the modal when the modal is opened; and (2) Sets the modal title as the label for the modal so that the title is read out by the screen reader when the modal is opened.

There's still a few improvements that could be made, like preventing the user from accessing the content behind the modal while the modal is open. I tried a few different ways of doing that with no luck, and this is at least an improvement over the current implementation.

#### General

#### User visible changes

Turn on VoiceOver, then open a modal. Verify focus moves to the modal and the modal title is read out by the screen reader.

### Issue(s) this completes

Fixes #4196


caitlinshk:
Rocky, assigning to you since you created https://github.com/civiform/civiform/issues/4196. Specifically, I'm curious if you think this addresses modal accessibility enough to mark #4196 as fixed or if you think there's more to do. Thanks!

caitlinshk:
I'm a bit worried that only some of the modal screenshots show the a11y focus outline. Specifically, I'm worried this might introduce some flakiness where the screenshot is sometimes taken before focus is acquired, and sometimes taken after. I ran the tests twice and it was the same 4 screenshots that were changed both times, so hopefully it's not flaky?

PR #5883, opened 2023-10-27
Enable question cache in browser tests

dkatzz:
### Description

Enable question cache in browser tests.

#### General

### Issue(s) this completes

Related to #5845


dkatzz:
We are no longer planning to cache question data (at least for now) because when we look up a question it is on the admin side and not the applicant side.

PR #5884, opened 2023-10-27
Allow cache to be cleared in dev and staging instances of the app

### LINKED ISSUE #5844: Ensure cache can be cleared in dev instance
**Description**
We should create an endpoint and/or button for users to clear the cache in a dev or staging instance of CiviForm.

Related to https://github.com/civiform/civiform/issues/5843

dkatzz:
### Description

Add an endpoint and button to clear the cache in dev and staging instances of the app. We would be reloading the cache by navigating to the index page, so we instead navigate to a blank page.

#### General

### Issue(s) this completes

Fixes #5844


gwendolyngoetz (approved):

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -111,20 +111,38 @@ public Result clear() {
...
   }
 
   /** Remove all content from the cache. */
-  private void clearCache() {
-    programsByVersionCache.removeAll().toCompletableFuture().join();
-    questionsByVersionCache.removeAll().toCompletableFuture().join();
-    programCache.removeAll().toCompletableFuture().join();
-    versionsByProgramCache.removeAll().toCompletableFuture().join();
+  public Result clearCache() {
+    if (!isDevOrStaging) {
```
yotommy, line 122: What do you think about allowing this in prod? I could imagine that it _might_ help in some weird circumstances, and hitting it accidentally wouldn't do _that_ much damage.

dkatzz: Thanks for the comment! Agreed that it wouldn't do much harm. The only reason I didn't is because in many cases prod environments will have more than one task running and if there is, this wouldn't clear the cache in all tasks. I'm happy to enable it though and add that as part of the message. "The cache has been cleared. NOTE: if you are clearing the cache in a deployed instance and you have more than one task running, you should re-deploy the application to ensure the cache from every task is cleared." Let me know what you think.

yotommy: Excellent point. In that case, disallowing here make sense to me.

gwendolyngoetz: If you put it in prod it should go somewhere other than under /dev

dkatzz: Yes, good point! Not going to do this for now, but if we decide to do it at some point, I'll make that change.

Issue #5860, opened 2023-10-25
Feature flag for Universal Questions


nb1701:
See [Releasing a Feature Behind a Feature Flag](https://github.com/civiform/civiform/wiki/Feature-Flags#releasing-a-feature-behind-a-feature-flag) for more details.

### Flag name

`UNIVERSAL_QUESTIONS`

### Feature description

This enables setting the UNIVERSAL QuestionTag on a question and all the bits of the UI around universal questions.  

### Checklist

- [x] Feature flag created, feature guarded with flag in code
- [x] Flag [enabled in dev](https://github.com/civiform/civiform/blob/main/server/conf/application.dev.conf)
- [x] Flag [disabled for browser tests](https://github.com/civiform/civiform/blob/main/server/conf/application.dev-browser-tests.conf)
- [x] Feature written and guarded with flag, including unit and browser tests that manipulate the state of the flag as needed. Enable the flag for browser tests when it's ready to be tested.
- [x] Flag enabled in staging and Seattle staging
- [x] Seattle given option to test manually
- [x] Flag default changed to true, with PR release notes mentioning the change and what feature it enables. <Edit this issue and note the date it was enabled here>
- [x] Verified feature is enabled in production build. <Edit this issue and note the date it was enabled here>
- [x] Flag enabled in production for at least a month
- [x] Flag removed and feature code unguarded

Issue #5810, opened 2023-10-17
Add Universal Question slider and persist the state


nb1701:
Per the [mock](https://www.figma.com/file/KqfHZYPS3Ffd93aJsfkNYZ/Universal-Questions?type=design&node-id=293-1092&mode=design&t=mGC8OAmup0UNw6Gw-0), add a "Universal question" section to "Question settings" in the question creation/edit view that allows the user to mark the question as a universal question.  This should be persisted in the database as a QuestionTag, per the [TDD](https://docs.google.com/document/d/1-2stuFeIKWkiDCHZlnytsG8c0q0G_Dl3dnplSY4Wy5E/edit#heading=h.bknox3ib8wpb).

PR #5887, opened 2023-10-27
Add Universal question tag and toggle

nb1701:
### Description

- Our current toggle reloads the page (or, rather, part of the page via htmx) every time it is toggled because we need to record the state of the toggle in the program block edit view, which doesn't have a "submit" kind of button. Here, we want the toggle to behave more like a checkbox. This adds a toggle that is defined much like our current toggle, but sets the state of a hidden checkbox underneath and is toggled via typescript.  We should be able to rework our current toggle to use this new ViewUtils.makeToggleButton later on.  This is currently styled exactly the same as our current toggle, which doesn't quite match the mocks.  There may be some tweaks pending final approval from @sijiayam . Note that this PR originally contained a brand new toggle component written mostly in CSS.  The commit remains in the PR history, but reverted, in case we decide we want to go that direction instead.  However, making the toggle a button is better from an accessibility standpoint.  
- Adds a toggle to the question create/edit view that sets whether the question is a universal question.
- Adds a `isUniversal` field to `QuestionForm`, which binds to the state of the toggle used in the view.
- Adds a new `QuestionTag` called `UNIVERSAL`.  When present for a question, that question will be treated as a universal question, with changes in the UI that will be added in later PRs.
- Adds a `universal` property and `setUniversal` function to QuestionDefinitionConfig, a `setUniversal` function to `QuestionDefinitionBuilder`, and a `isUniversal` function to `QuestionDefinition`.  The `universal` property will be default `false`, set via the config builder.
- When a `Question` is created by passing in a `QuestionDefinition` to the constructor, the `UNIVERSAL` QuestionTag will get added if `QuestionDefinition.isUniversal()` is true.
- Adds a UNIVERSAL_QUESTIONS feature flag.
- Optimizes the code in some of the unit tests

#### General

#### User visible changes

Note: I did not leave the validateAccessibility function in the test, since it fails for other things on the page, but verified there are no warnings for the added components.

#### New Features

- [x] Add new FeatureFlag env vars to `server/conf/helper/feature-flags.conf`

1. Turn on the feature flag in the settings UI
2. Create a new question with universal set/unset, and ensure it remains set/unset when editing the question.
3. Edit the question, toggling universal, and ensure the state persists.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/5860 https://github.com/civiform/civiform/issues/5810


server/app/views/ViewUtils.java
```diff
@@ -232,4 +234,67 @@ public PTag renderEditOnText(String prefix, Optional<Instant> time) {
   public static SpanTag requiredQuestionIndicator() {
     return span(rawHtml("&nbsp;*")).withClasses("text-red-600", "font-semibold");
   }
+
+  public static ButtonTag makeToggleButton(
```
bion, line 238: suggestion (blocking): add javadoc for this new public method

server/app/views/ViewUtils.java
```diff
@@ -232,4 +234,67 @@ public PTag renderEditOnText(String prefix, Optional<Instant> time) {
...
+                                enabled ? "right-1" : "left-1",
+                                "top-1",
+                                "w-6",
+                                "h-6",
+                                "rounded-full",
+                                "cf-toggle-nub",
+                                toggleInstanceNubClass)));
+    if (text.isPresent()) {
+      button.withText("Set as a universal question");
+    }
```
bion, line 297: issue (blocking): this text should be added somewhere else since this method is generic

server/conf/env-var-docs.json
```diff
@@ -717,6 +717,11 @@
         "mode": "HIDDEN",
         "description": "Enables logic to populate more fields in OIDC logout requests.",
         "type": "bool"
+      },
+      "UNIVERSAL_QUESTIONS": {
+        "mode": "ADMIN_WRITEABLE",
+        "description": "Enables setting the universal question state on questions.",
```
bion, line 723: suggestion (non-blocking): consider providing more info here for admins who might not be familiar with universal questions

bion (changes requested):

server/app/assets/javascripts/toggle.ts
```diff
@@ -1,4 +1,5 @@
 /** This class controls the toggling of toggle components. */
+import {addEventListenerToElements} from './util'
```
bion, line 2: nit: space makes things a little easier to read
```suggestion
import {addEventListenerToElements} from './util'

```

server/app/views/ViewUtils.java
```diff
@@ -232,4 +234,83 @@ public PTag renderEditOnText(String prefix, Optional<Instant> time) {
...
+                                enabled ? "right-1" : "left-1",
+                                "top-1",
+                                "w-6",
+                                "h-6",
+                                "rounded-full",
+                                "cf-toggle-nub")
+                            .withCondId(idPresent, nubId)));
+    if (text.isPresent()) {
+      button.withText(text.get());
+    }
```
bion, line 313: suggestion (non-blocking): 
```suggestion
    text.ifPresent(button::withText);
```

bion (approved):

PR #5889, opened 2023-10-30
Adding optional namespace to image build scripts

gwendolyngoetz:
### Description

Adds an optional namespace override for the docker image build scripts. Also adds more details to the docs comment.

This will simplify the instructions listed in https://github.com/civiform/civiform/wiki/Deploy-System-Developer-Guide#building-your-docker-image-locally

Currently does not support pushing to alternate registries due to the `docker login` in `lib/docker.sh` logging into default registry (docker.io).

Since it defaults to the original value existing usage in GitHub actions do not need to be modified.

#### Examples

`bin/build-dev` defaults to current value of "civiform" and would be `civiform/civiform-dev:latest`
`NAMESPACE=alt-namespace bin/build-dev` will use "alt-namespace" and would be `alt_namespace/civiform-dev:latest`

#### General

### DIFF
--- a/bin/build-browser-tests
+++ b/bin/build-browser-tests
@@ -1,18 +1,23 @@
 #! /usr/bin/env bash
 
-# DOC: Build a new docker image for running the browser tests and optionally push to Docker Hub if PUSH_IMAGE is set to anything.
+# DOC: Build a new docker image for running the browser tests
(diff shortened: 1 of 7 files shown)


yotommy (approved):
Thank you!

Issue #5811, opened 2023-10-17
Add Universal question badge to Questions tab


nb1701:
Per the [mocks](https://www.figma.com/file/KqfHZYPS3Ffd93aJsfkNYZ/Universal-Questions?type=design&node-id=1-27&mode=design&t=mGC8OAmup0UNw6Gw-0), add a badge to universal questions in the Questions tab.  The badge should show the state of either the active version of the question, or the draft version of the question, if it exists.  Additionally, change the sorting so that universal questions appear before other questions, as noted in the [TDD](https://docs.google.com/document/d/1-2stuFeIKWkiDCHZlnytsG8c0q0G_Dl3dnplSY4Wy5E/edit#heading=h.bknox3ib8wpb).

PR #5932, opened 2023-11-06
Add universal badge to questions and sort with universal on top

nb1701:
### Description

This adds a `ViewUtils` function to create a universal question badge.  This is used in this PR to add the badge to the Questions tab when a question is marked as a universal question, and can be reused in future PRs for other views.  It adds a new icon from the CiviForm icon library in Figma, and the badge includes in the name of the question type in it.  This also modifies the sorting so that universal questions appear before non-universal ones.  Within both groups, questions are sorted by last modified as they are today.

Additionally, this modifies the `label` field of several `QuestionType`s.  Some included the word `Field`, some did not, some included other words that weren't necessary.  This change makes the labels more concise and less awkward to use in the UI, such as in the universal badge.

#### General

#### User visible changes

Note: I did not leave the `validateAccessibility` step in the test, as several other elements are failing since this is admin-facing, but I verified the badge does not trigger any issues.

#### New Features

Test with universal set and unset for a variety of questions.  Ensure sorting is correct (universal by last modified, then the rest by last modified).

### Issue(s) this completes

https://github.com/civiform/civiform/issues/5811

### DIFF
--- a/server/app/views/ViewUtils.java
+++ b/server/app/views/ViewUtils.java
@@ -27,7 +27,6 @@ import java.time.Instant;
 import java.util.Optional;
 import javax.inject.Inject;
 import services.DateConverter;
-import services.question.types.QuestionType;
 import views.components.Icons;
 import views.components.LinkElement;
 import views.style.BaseStyles;
@@ -162,7 +161,7 @@ public final class ViewUtils {
    * @return Tag representing the badge. No classes should be added to the returned tag as it will
    *     overwrite existing classes due to how Jhtml works.
    */
-  public static PTag makeBadge(ProgramDisplayType status, String... extraClasses) {
+  public static PTag makeLifecycleBadge(ProgramDisplayType status, String... extraClasses) {
     String badgeText = "", badgeBGColor = "", badgeFillColor = "";
     switch (status) {
       case ACTIVE:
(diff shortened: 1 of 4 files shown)


server/app/views/ViewUtils.java
```diff
@@ -301,4 +302,21 @@ public static ButtonTag makeToggleButton(
     text.ifPresent(button::withText);
     return button;
   }
+
+  public static DivTag makeUniversalBadge(QuestionType questionType) {
```
melanie-exygy, line 306: suggestion(non-blocking):  This could easily be a more generic `makeBadge(String content)` so that we would have a more content-agnostic and reusable UI component.  I do see that there's already a `makeBadge` function.  That could be renamed to `makeProgramStatusBadge`.

server/tailwind.config.js
```diff
@@ -55,6 +55,9 @@ module.exports = {
           light: '#ffefc9',
           DEFAULT: '#6b4b00',
         },
+        'civiform-teal': {
```
melanie-exygy, line 58: praise:  Great idea to add this to what's currently serving as our design tokens.

server/app/services/question/types/QuestionType.java
```diff
@@ -21,21 +21,21 @@
 
 /** Defines types of questions supported. */
 public enum QuestionType {
-  ADDRESS("Address Field", AddressQuestion.class),
+  ADDRESS("Address", AddressQuestion.class),
```
melanie-exygy, line 24: praise: Great idea to make these labels more concise and to-the-point.  I think it makes them more user-friendly.

melanie-exygy (commented):
Looks good to me!  I just had one small, non-blocking suggestion.

melanie-exygy (approved):

PR #5935, opened 2023-11-06
Add additional data to database query metrics for prometheus 

### LINKED ISSUE #5958: Add query location information to Prometheus
**Description**
Having  more detailed information on our query metrics will be helpful to further evaluate our caching implementation.

We should add the filename and function that is executing a database query. For queries that are through java persistence annotations, we can have the class name.

dkatzz:
### Description

Add a profile location and class name to our database query metrics, so we can better see which queries are still being executed to analyze the benefit of the cache and how we can further optimize it.

This adds this data specifically to the VersionRepository and ProgramRepository class, but we will extend this to other repository files in the future. 

#### General

- [ ] Extended the README / documentation, if necessary (will do as a follow up once this is in production)

### Issue(s) this completes

Fixes #5958

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -46,6 +46,7 @@ import services.settings.SettingsManifest;
  */
 public final class ProgramRepository {
   private static final Logger logger = LoggerFactory.getLogger(ProgramRepository.class);
+  private final QueryUtils queryUtils = new QueryUtils("ProgramRepository");
 
   private final Database database;
   private final DatabaseExecutionContext executionContext;
@@ -81,7 +82,13 @@ public final class ProgramRepository {
   }
 
   private Optional<Program> lookupProgramSync(long id) {
-    return database.find(Program.class).where().eq("id", id).findOneOrEmpty();
+    return database
+        .find(Program.class)
+        .setLabel("Program.findById")
+        .setProfileLocation(queryUtils.createProfileLocation("lookupProgramSync"))
+        .where()
+        .eq("id", id)
+        .findOneOrEmpty();
   }
 
   public Program insertProgramSync(Program program) {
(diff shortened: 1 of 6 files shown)


server/app/repository/ProgramRepository.java
```diff
@@ -46,6 +46,7 @@
  */
 public final class ProgramRepository {
   private static final Logger logger = LoggerFactory.getLogger(ProgramRepository.class);
+  private final QueryUtils queryUtils = new QueryUtils("ProgramRepository");
```
avaleske, line 49: nit (non-blocking, optional): `QueryUtils` feels like it'd be a static class. Is there a name for this that might not imply that? Something like `QueryProfileLocationBuilder`?

Or maybe `QueryUtils` is static and here we can call a `getProfileLocationBuilder()` method on it?

avaleske (approved):
- Is there documentation we should update with this new pattern?
- Are there any static analysis tools we can use to make sure the labels are unique?

server/app/repository/ProgramRepository.java
dkatzz: It can be a static variable here (updated to add), but also renamed to QueryProfileLocationBuilder, since that better represents what its doing. Let me know what you think!

avaleske:
Can you add an issue to the PR description or remove that line?

dkatzz:
> * Is there documentation we should update with this new pattern?

Yeah, I'll plan to update the grafana dashboard default setup to include graphs with these data points, but I'll do this as a follow up once this is in production. I'll also look to see where this can be included in the dev docs.

> * Are there any static analysis tools we can use to make sure the labels are unique?

In this case, since we're creating the prometheus data point, it will only have the labels we create for it, so this should be fine as is. https://prometheus.io/docs/practices/naming/#labels

dkatzz:
> Can you add an issue to the PR description or remove that line?

Updated thank you!

avaleske (approved):

PR #5946, opened 2023-11-08
Remove applicant id from URLs for index action

yotommy:
### Description

Currently, routes for applicant actions include the applicant id in the URL. We want to remove these for applicant actions to provide cleaner URLs for sharing and bookmarking. To accomplish this, we will store the applicant id in the user profile. A profile attribute is set in `ProfileFactory.createNewApplicant()` which will achieve the goal for every newly-created profile. If a user with an existing (legacy) profile authenticates successfully, then we also set the applicant id in the `CiviFormProfileMerger`.

This PR only affects the `ApplicantProgramsController.index()` action; subsequent PRs will address the other applicant actions.

This PR renames the existing `index()` method to `indexWithApplicantId()`. A new `index()` method extracts the applicant ID from the profile and then calls `indexWithApplicantId()`. The corresponding `routes` are configured appropriately.

Note that we want to retain the existing routes when _TIs_ are taking actions of an applicant's behalf.  Also, some applicants will have existing profiles that do not yet contain the applicant id, so we need to use the existing routes for them as well. Since routes for applicant actions are computed in a number of views as well as controllers (for redirects), the new `ApplicantRoutes` class will centralize the conditional logic for computing the appropriate route for a given situation. There are also some changes in plumbing to pass the user profile to views when this computation is required. Note that another reason to retain the existing routes, particularly for the home page, is that existing users may have the current URLs bookmarked.

In addition to the unit tests, this change was tested on a dev instance with an existing profile in the browser cookie. See instructions below.

### Release notes

Remove applicant ID from home page URL

#### General

- In a dev instance, check out `main`
- Navigate to `http://localhost:9000`
- This will redirect to the home page with the applicant ID in the URL. It will also create a user profile that does not include the applicant id
- Check out this branch
- Again, navigate to `http://localhost:9000`
- Again, this should redirect to the home page with the applicant ID in the URL
- Click "End session" in upper right. This will create a new user profile that _does_ include the applicant id
- Verify that the browser is now on `http://localhost:9000/programs` (no applicant id)

### Issue(s) this completes

Relates to #5905

### DIFF
--- a/server/app/auth/CiviFormProfile.java
+++ b/server/app/auth/CiviFormProfile.java
@@ -1,11 +1,11 @@
 package auth;
 
+import static java.util.Comparator.comparing;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
-import java.util.Comparator;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -48,7 +48,8 @@ public class CiviFormProfile {
 
   /** Get the latest {@link ApplicantModel} associated with the profile. */
   public CompletableFuture<ApplicantModel> getApplicant() {
-    if (profileData.containsAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME)) {
+    if (settingsManifest.getNewApplicantUrlSchemaEnabled()
+        && profileData.containsAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME)) {
       long applicantId =
           profileData.getAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME, Long.class);
       return accountRepository
@@ -64,11 +65,12 @@ public class CiviFormProfile {
     // which requires an extra db fetch.
     return this.getAccount()
         .thenApplyAsync(
-            (a) ->
-                a.getApplicants().stream()
-                    .min(Comparator.comparing(ApplicantModel::getWhenCreated))
-                    .orElseThrow(),
-            httpContext.current());
+            (account) -> getApplicantForAccount(account).orElseThrow(), httpContext.current());
+  }
+
+  private Optional<ApplicantModel> getApplicantForAccount(AccountModel account) {
+    // Accounts (should) correspond to a single applicant.
+    return account.getApplicants().stream().min(comparing(ApplicantModel::getWhenCreated));
   }
 
   /** Look up the {@link AccountModel} associated with the profile from database. */
@@ -280,5 +282,36 @@ public class CiviFormProfile {
                   String.format(
                       "Account %s is not authorized to access program %s.", getId(), programName));
             });
+  }
+
+  /**
+   * Stores applicant id in user profile.
+   *
+   * <p>This allows us to know the applicant id instead of having to specify it in the URL path, or
+   * looking up the account each time and finding the corresponding applicant id.
+   */
+  void storeApplicantIdInProfile(Long applicantId) {
+    if (!settingsManifest.getNewApplicantUrlSchemaEnabled()) {
+      return;
+    }
+
+    if (!profileData.containsAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME)) {
+      profileData.addAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME, applicantId);
+    }
+  }
+
+  /**
+   * Stores applicant id corresponding to this account in user profile.
+   *
+   * <p>This allows us to know the applicant id instead of having to specify it in the URL path, or
+   * looking up the account each time and finding the corresponding applicant id.
+   */
+  void storeApplicantIdInProfile(AccountModel account) {
+    if (!settingsManifest.getNewApplicantUrlSchemaEnabled()) {
+      return;
+    }
+
+    Long applicantId = getApplicantForAccount(account).orElseThrow().id;
+    storeApplicantIdInProfile(applicantId);
   }
 }
--- a/server/app/auth/CiviFormProfile.java
+++ b/server/app/auth/CiviFormProfile.java
@@ -4,6 +4,7 @@ import static java.util.Comparator.comparing;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
+import auth.controllers.MissingOptionalException;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import java.util.Optional;
@@ -56,7 +57,8 @@ public class CiviFormProfile {
           .lookupApplicant(applicantId)
           .thenApply(
               optionalApplicant -> {
-                return optionalApplicant.orElseThrow();
+                return optionalApplicant.orElseThrow(
+                    () -> new MissingOptionalException(ApplicantModel.class));
               })
           .toCompletableFuture();
     }
@@ -65,7 +67,10 @@ public class CiviFormProfile {
     // which requires an extra db fetch.
     return this.getAccount()
         .thenApplyAsync(
-            (account) -> getApplicantForAccount(account).orElseThrow(), httpContext.current());
+            (account) ->
+                getApplicantForAccount(account)
+                    .orElseThrow(() -> new MissingOptionalException(ApplicantModel.class)),
+            httpContext.current());
   }
 
   private Optional<ApplicantModel> getApplicantForAccount(AccountModel account) {
@@ -311,7 +316,10 @@ public class CiviFormProfile {
       return;
     }
 
-    Long applicantId = getApplicantForAccount(account).orElseThrow().id;
+    Long applicantId =
+        getApplicantForAccount(account)
+            .orElseThrow(() -> new MissingOptionalException(ApplicantModel.class))
+            .id;
     storeApplicantIdInProfile(applicantId);
   }
 }
--- a/server/app/auth/CiviFormProfile.java
+++ b/server/app/auth/CiviFormProfile.java
@@ -15,6 +15,7 @@ import models.AccountModel;
 import models.ApplicantModel;
 import play.libs.concurrent.HttpExecutionContext;
 import play.mvc.Http.Request;
+import repository.AccountRepository;
 import repository.DatabaseExecutionContext;
 import services.settings.SettingsManifest;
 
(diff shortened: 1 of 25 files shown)


yotommy:
* main
  * **PR #5946** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5946" target="_blank">Graphite</a>  👈
    * **PR #5905** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5905" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/5946?utm_source=stack-comment).

server/app/auth/CiviFormProfile.java
```diff
@@ -44,6 +45,27 @@ public CiviFormProfile(
...
+            Applicant applicant = new Applicant();
+            applicant.id = applicantId;
+            try {
+              applicant.refresh();
+            } catch (EntityNotFoundException e) {
+              // Unlike AccountNotFoundException, ApplicantNotFoundException is not a
+              // RuntimeException.
+              throw new RuntimeException(new ApplicantNotFoundException(applicantId));
+            }
+            return applicant;
```
bion, line 62: suggestion (blocking): use AccountRepository.lookupApplicant instead

In general we're trying to move toward encapsulating all query logic into repository classes where we'll have an easier time optimizing it and controlling how data is loaded

server/app/auth/CiviFormProfileMerger.java
```diff
@@ -81,10 +81,22 @@ private Optional<CiviFormProfile> mergeApplicantAndGuestProfile(
         // Merge the two applicants and prefer the newer one.
         guestProfile = Optional.of(mergeProfiles(applicantInDatabase.get(), guestProfile.get()));
       }
+      // Ideally, the applicant id would already be populated in `guestProfile`. However, there
+      // could be profiles in user sessions that were created before we started populating this
+      // info.
+      storeApplicantIdInProfile(guestProfile.orElseThrow(), applicantInDatabase.orElseThrow());
```
bion, line 87: question (blocking): when might this throw?

server/app/auth/ProfileFactory.java
```diff
@@ -38,18 +39,31 @@ public ProfileFactory(
...
       Provider<ApiKeyService> apiKeyService,
-      Provider<AccountRepository> userRepositoryProvider,
+      Provider<AccountRepository> accountRepositoryProvider,
       SettingsManifest settingsManifest) {
     this.dbContext = Preconditions.checkNotNull(dbContext);
     this.httpContext = Preconditions.checkNotNull(httpContext);
     this.versionRepositoryProvider = Preconditions.checkNotNull(versionRepositoryProvider);
     this.apiKeyService = Preconditions.checkNotNull(apiKeyService);
-    this.userRepositoryProvider = Preconditions.checkNotNull(userRepositoryProvider);
+    this.accountRepositoryProvider = Preconditions.checkNotNull(accountRepositoryProvider);
```
bion, line 48: praise: thanks for renaming

server/app/repository/AccountRepository.java
```diff
@@ -423,4 +425,16 @@ public void updateSerializedIdTokens(
...
+   * Stores applicant id in user profile.
+   *
+   * <p>This allows us to know the applicant id instead of having to specify it in the URL path,
+   * or looking up the account each time and finding the corresponding applicant id.
+   */
+  public void populateApplicantIdInProfile(AccountModel account, CiviFormProfileData profileData) {
+    // Accounts correspond to a single applicant.
+    Long applicantId = account.ownedApplicantIds().stream().findAny().orElseThrow();
+    profileData.addAttribute(APPLICANT_ID_ATTRIBUTE_NAME, applicantId);
+  }
```
bion, line 456: issue (blocking): repository classes are for database interaction, maybe `CiviFormProfileData` is a good place for this method?

server/app/controllers/applicant/ApplicantRoutes.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Returns the URL corresponding to the applicant index action.
+   *
+   * @param profile - Profile corresponding to the logged-in user (applicant or TI).
+   * @param applicantId - ID of applicant for whom the action should be performed.
+   * @return Route for the index action.
+   */
+  public static Call index(CiviFormProfile profile, long applicantId) {
```
bion, line 33: suggestion (if-minor, non-blocking): Static methods are resistant to extension/refactoring since if there becomes a need for them to depend on something non-static (e.g. the settings manifest) all call sites will need to be updated. Consider making this a public instance method and injecting an instance of `applicantRoutes` where ever it's needed.

server/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -215,4 +225,16 @@ public CompletionStage<Result> edit(Request request, long applicantId, long prog
...
+    CiviFormProfileData profileData =
+        profileUtils.currentUserProfile(request).orElseThrow().getProfileData();
+
+    if (profileData == null) {
+      return Optional.empty();
+    }
+
+    return Optional.ofNullable(
+        profileData.getAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME, Long.class));
+  }
```
bion, line 239: suggestion (blocking): this will be reused in future PRs that handle the other applicant routes, consider moving somewhere shared such as the parent controller class or `ProfileUtils`

server/app/controllers/applicant/ApplicantRoutes.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public final class ApplicantRoutes {
+
+  // There are two cases where we want to use the URL that contains the applicant id:
+  // - TIs performing actions on behalf of applicants.
+  // - The applicant has a profile that does not (yet) include the applicant id.
+  //   This case will eventually go away once existing profiles have expired and been replaced.
+  private static boolean includeApplicantIdInRoute(CiviFormProfile profile) {
+    boolean isTi = profile.isTrustedIntermediary();
+    // TODO(yotommy): add a counter to monitor this case
```
bion, line 19: suggestion (non-blocking): do this now, now that the counter PR is ready? otherwise file an issue for it `TODO([issue #])`

bion (changes requested):

server/app/auth/CiviFormProfileMerger.java
yotommy: These two accesses are guaranteed not to throw by the surrounding logic.

server/app/controllers/applicant/ApplicantRoutes.java
yotommy: I had thought about making the instances, but thought it would be a messy refactor. Guice to the rescue! (I had not considered this.) Done, and this simplifies the counter tests.

server/app/auth/CiviFormProfile.java
yotommy: Sigh, yes, I should remember this by now, instead of just borrowing a pattern from nearby. Done

server/app/repository/AccountRepository.java
yotommy: Done

server/app/controllers/applicant/ApplicantProgramsController.java
yotommy: Done

server/app/controllers/applicant/ApplicantRoutes.java
yotommy: Done, and simpler now that we are using instance methods rather than static methods.

server/app/auth/CiviFormProfileMerger.java
bion: Thanks!

server/app/controllers/applicant/ApplicantRoutes.java
bion: Ah sorry to hassle you on this, but that isn't an improvement for maintainability since with the factory method we can't inject the c'tor. Guice will use the default constructor without you needing to create one, so you can do:

```java
  private final ApplicantRoutes applicantRoutes;

  @Inject
  public ApplicantInformationController(
     ...,
     ApplicantRoutes applicantRoutes
       ) {
    super(profileUtils, versionRepository);
    ...
    this.applicantRoutes = checkNotNull(applicantRoutes); 
  }

  ...
                redirectLink = applicantRoutes.index(profile, applicantId).url();
```

Doing this now just makes it easier to add state in the future. If you'd rather not inject it in all the places that's fine, YAGNI is valid.

bion (approved):

server/app/controllers/applicant/ApplicantRoutes.java
bion: thanks!

bion (approved):

server/app/auth/CiviFormProfileData.java
```diff
@@ -18,6 +19,8 @@
  */
 public class CiviFormProfileData extends CommonProfile {
 
+  private static final long serialVersionUID = 3142603030317816700L;
```
bion, line 22: suggestion (blocking): please add some context for why this value in particular and what it is doing

bion (approved):
overall looks really good, thank you!

server/app/auth/CiviFormProfileData.java
yotommy: Done in parent PR #6063

server/app/auth/CiviFormProfile.java
```diff
@@ -30,27 +31,46 @@ public class CiviFormProfile {
...
+    if (settingsManifest.getNewApplicantUrlSchemaEnabled()
+        && profileData.containsAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME)) {
+      long applicantId =
+          profileData.getAttribute(ProfileFactory.APPLICANT_ID_ATTRIBUTE_NAME, Long.class);
+      return accountRepository
+          .lookupApplicant(applicantId)
+          .thenApply(
+              optionalApplicant -> {
+                return optionalApplicant.orElseThrow();
+              })
```
bion, line 60: **nitpick:** 
```suggestion
          .thenApply(Optional::orElseThrow)
```

server/app/controllers/applicant/ApplicantInformationController.java
```diff
@@ -48,13 +50,15 @@ public ApplicantInformationController(
...
-      VersionRepository versionRepository) {
+      VersionRepository versionRepository,
+      ApplicantRoutes applicantRoutes) {
     super(profileUtils, versionRepository);
     this.httpExecutionContext = httpExecutionContext;
     this.messagesApi = messagesApi;
     this.repository = repository;
     this.formFactory = formFactory;
     this.layout = layout;
+    this.applicantRoutes = applicantRoutes;
```
bion, line 61: **nitpick:** these should all have `checkNotNull`

server/app/controllers/applicant/ApplicantInformationController.java
```diff
@@ -90,7 +94,8 @@ public CompletionStage<Result> setLangFromBrowser(Http.Request request, long app
                             /* page= */ Optional.of(1))
                         .url();
               } else {
-                redirectLink = routes.ApplicantProgramsController.index(applicantId).url();
+                CiviFormProfile profile = profileUtils.currentUserProfile(request).orElseThrow();
```
bion, line 97: **suggestion (blocking):** This PR uses `orElseThrow` quite a bit. It'd be helpful for each of those places to include an error message that will assist in error reporting and debugging when a civic entity reports a incident. Even something as simple as `orElseThrow(() -> new RuntimeException("No profile found!")`

bion (approved):

server/app/auth/CiviFormProfile.java
yotommy: Thanks for this suggestion. But based on your other suggestion (throwing meaningful exceptions in `.orElseThrow()`), I am retaining this syntax.

server/app/controllers/applicant/ApplicantInformationController.java
yotommy: Done

server/app/controllers/applicant/ApplicantInformationController.java
yotommy: Added a `MissingOptionalException` and used it throughout.

PR #5960, opened 2023-11-10
Refactor Program to ProgramModel

### LINKED ISSUE #5961: Refactor our classes within the "Model" directory to include Model in the class name
**Description**
Update our Model classes to have "Model" in the class name, so it is clearer when we are accessing the database.

dkatzz:
### Description

Refactor the "Program" class to be "ProgramModel", so it is clear when we may be accessing the database.

We'll do this to the other Models as well as a follow up.

I didn't rename the associated variables, but can if we think that is the right approach.

#### General

### Issue(s) this completes

Fixes #5961


avaleske (approved):
Wow we really just pass the models around everywhere don't we

dkatzz:
> Wow we really just pass the models around everywhere don't we

We really do!

Issue #5676, opened 2023-09-26
[Epic] Add images to program cards


melanie-exygy:
### Epic Overview

Residents, CBOs, and City Staff have all mentioned that having images in the program card would help to know who the program serves. It's quicker to digest images than text.  We would like admins to be able to upload an image that displays on the program card.  See #3783.  

### Goal

Feedback from users as well as findings from behavioral economics research have identified that our current program cards are difficult for residents and TIs to quickly parse for important information. Users are having to click into program details to understand the program's value proposition or area focus (nutrition, transportation, etc). In order to deliver on our discoverability (and accessibility) goals, we need to remove this cognitive burden. While we will investigate multiple areas for enhancement, we'll begin with the ability to add images. 

### Key Deliverables

- There is an upload box on the program edit page. The upload supports JPEG, PNG, GIF images at a minimum. #5782 
- We have the ability to store the program images.
- The images appear on program cards on the program index  #page.  #5781 

### Documentation

PRD creation issue:  #5686 
UX mocks
TDD creation issue:   #5680

caitlinshk:
This feature is finished and default-enabled in production. https://github.com/civiform/civiform/issues/5919 will track removing the flag eventually, but I think we can mark this as complete!

PR #5964, opened 2023-11-10
Add "Edit program image" button to program details page.

caitlinshk:
### Description

This change adds a frontend skeleton for #5676. Specifically:
1) Adds an "Edit program image" button on the program edit homepage.
2) When that button is clicked, opens a new page under the route `admin/programs/{id}/image` to add/modify/remove the program image. This page currently just has a title and nothing else.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)

#### User visible changes

- [x] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [x] Manually tested at 200% size
- [x] Manually evaluated tab order

#### New Features

- [x] Conditioned new functionality on a [FeatureFlag](https://github.com/civiform/civiform/wiki/Feature-Flags)
- [x] Wrote browser tests with the feature flag off and on, etc.

### Instructions for manual testing

- Disable program_card_images flag then go to a program edit page -> verify "Edit program image" button doesn't appear
- Enable program_card_images flag then go to a program edit page -> verify "Edit program image" button appears.
- Click on "Edit program image" button -> verify a new page appears (with no content, yet :) )

The same steps should also work for setting up an eligibility/visibility condition and editing an existing eligibility/visibility condition.

### Issue(s)
Fixes #5996

Epic: #5676

PR #5963, opened 2023-11-10
Put program header button rendering all in one place.

caitlinshk:
### Description

Updates `ProgramBaseView` to do all the header button rendering. Specific admin program views now just pass the buttons that they want to the render method.

For #5676 we're going to add a third "Edit program image" button, and this refactor helps add that button more easily.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).
- [x] Created unit and/or browser tests which fail without the change (if possible) -- **Existing browser tests should cover**
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)

### Instructions for manual testing

As an admin:
1) View a program that doesn't have a draft -> verify button says "Edit program" not "Edit program details". Verify clicking on "Edit program" reloads the page to then show the program as editable.
2) View a draft program -> verify "Edit program details" and "Preview as applicant" buttons appear.
2) Add or edit a visibility condition -> verify "Edit program details" button appears.
3) Add or edit an eligibility condition -> verify "Edit program details" button appears.

### Issue(s) this completes
Pre-refactor for #5676

PR #5976, opened 2023-11-15
Add program image alt text to admin translations page.

### LINKED ISSUE #5998: [Program Card Images] [Admin] Add alt text to translations page
Sub-issue of main epic issue #5676

caitlinshk:
### Description

Updates the admin translations page to include image description translations if an image description is set.

If there's no image description set, then image description isn't included in the translations page at all.

#### General

#### User visible changes

1. Edit a program > Click "Edit program image" > provide an image description.
2. Edit a program's translations > Verify there's a new "Program image description" field shown for each language on the translations page.
3. Verify you can save translated image descriptions for each language.

### Issue(s)
Fixes #5998 

Epic: #5676


caitlinshk:
* main
  * **PR #5963** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5963?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5964** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5964?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5967** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5976** <a href="https://app.graphite.dev/github/pr/civiform/civiform/5976?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/5976?utm_source=stack-comment).

server/app/services/program/ProgramService.java
```diff
@@ -659,14 +659,22 @@ public ErrorAnd<ProgramDefinition, CiviFormError> updateLocalization(
             .setLocalizedConfirmationMessage(
                 programDefinition
                     .localizedConfirmationMessage()
-                    .updateTranslation(locale, localizationUpdate.localizedConfirmationMessage()))
-            .setStatusDefinitions(
-                programDefinition.statusDefinitions().setStatuses(toUpdateStatusesBuilder.build()))
-            .build()
-            .toProgram();
+                    .updateTranslation(locale, localizationUpdate.localizedConfirmationMessage()));
+    if (programDefinition.localizedSummaryImageDescription().isPresent()
```
rockycodes, line 663: question/suggestion (blocking): I find this part a little hard to read. Is it possible to handle translations the same way that name, description, and confirmation message are? If not, I recommend factoring this logic out into a separate function.

server/app/views/admin/programs/ProgramTranslationView.java
```diff
@@ -97,29 +97,8 @@ private ImmutableList<DomContent> formFields(
...
-                            program.localizedDescription()),
-                        fieldWithDefaultLocaleTextHint(
-                            FieldWithLabel.input()
-                                .setFieldName(
-                                    ProgramTranslationForm.CUSTOM_CONFIRMATION_MESSAGE_FORM_NAME)
-                                .setLabelText("Custom Confirmation Screen Message")
-                                .setValue(updateData.localizedConfirmationMessage())
-                                .getInputTag(),
-                            program.localizedConfirmationMessage()))));
+                    getApplicantVisibleProgramDetailFields(program, updateData)));
```
rockycodes, line 100: praise: I really like that you factored this out!

server/app/views/admin/programs/ProgramTranslationView.java
```diff
@@ -176,4 +155,46 @@ private ImmutableList<DomContent> formFields(
...
+                    program.localizedDescription()),
+                fieldWithDefaultLocaleTextHint(
+                    FieldWithLabel.input()
+                        .setFieldName(ProgramTranslationForm.CUSTOM_CONFIRMATION_MESSAGE_FORM_NAME)
+                        .setLabelText("Custom Confirmation Screen Message")
+                        .setValue(updateData.localizedConfirmationMessage())
+                        .getInputTag(),
+                    program.localizedConfirmationMessage()));
+
+    // Only add the summary image description input to the page if a summary image description is
```
rockycodes, line 186: praise: I really like this design choice! It would be nice to have this conditional rendering with the confirmation message too.

browser-test/src/admin_program_translation.test.ts
```diff
@@ -148,6 +152,123 @@ describe('Admin can manage translations', () => {
...
+    await adminPrograms.gotoDraftProgramManageTranslationsPage(programName)
+    await adminTranslations.selectLanguage('Spanish')
+    await adminTranslations.expectProgramImageDescriptionTranslation(
+      'Spanish image description',
+    )
+
+    await disableFeatureFlag(page, 'program_card_images')
+  })
+
+  it('deleting summary image description deletes all translations', async () => {
```
rockycodes, line 237: question (non-blocking): Is this testing that the translations are deleted because the translation field will only show if they are present?

rockycodes (changes requested):
Just one requested change in `ProgramService`. Mostly I'm curious why you chose a different pattern for managing translations than the other translated fields use.

server/app/views/admin/programs/ProgramTranslationView.java
caitlinshk: Thank you! Let me know if you want me to file a GH issue about it.

server/app/services/program/ProgramService.java
caitlinshk: Great question. Like you'd mentioned in the other PR, the name and description are required program fields but the confirmation message is not. For confirmation message, we use `LocalizedStrings.of(DEFAULT_LOCALE, "")` to represent that we don't have any custom confirmation message. Since the message will always be an instance of `LocalizedStrings`, it's then easy to always call `localizedConfirmationMessage().updateTranslation(...)` here. For the program image description, I'm instead using `Optional.empty()` to represent that we don't have any description, which means that we have to check the optional everywhere and that's why this part is harder to read.

I think the question is:
1) Do we prefer the `LocalizedStrings.of(DEFAULT_LOCALE, "")` or `Optional.empty()` strategy for representing these non-required program fields?
- If we prefer `LocalizedStrings.of(DEFAULT_LOCALE, "")`, then I can update this PR and the previous PR to use that instead.
- If we prefer `Optional.empty()`, I can definitely pull this out into a helper method. And, that points to another question: should we go back and change the confirmation message implementation to also be `Optional.empty()`?

browser-test/src/admin_program_translation.test.ts
caitlinshk: Yes exactly! I added a bit more to the comment near the end of the method -- let me know if it makes sense or if there's a better way to explain.

server/app/services/program/ProgramService.java
rockycodes: Thank you so much for explaining! If an admin sets an image description and then goes back and clears it out, is it reset to `Optional.empty` or is it an empty string in that case? I think it's the former but want to make sure. Tbh I'm relatively new to using `Optional.empty` and I'm curious what you see as the tradeoffs. I like that it makes it super clear when we want to render the translations field, but it does seem to add some extra code.

server/app/views/admin/programs/ProgramTranslationView.java
rockycodes: That would be great! Thanks!

browser-test/src/admin_program_translation.test.ts
rockycodes: Thanks perfect. Thank you!

server/app/services/program/ProgramService.java
caitlinshk: Yep it's reset to `Optional.empty()` (lines 903-905 of https://github.com/civiform/civiform/pull/5967/files#diff-33f37dc160bda78d0b594ff49d80c4ffacca5eefe12b4f1f37ef3c1f9105f546).

There are a bunch of different camps for when/how to use Optional, ranging from "never" to "only for return types but never for parameters or fields" to "frequently". https://nipafx.dev/inside-java-newscast-19/ and https://nipafx.dev/java-optional-opinions talk about the different camps and tradeoffs, and [this Reddit thread](https://www.reddit.com/r/java/comments/13rv1t2/why_exactly_does_brian_goetz_not_recommend_the/) has a lot of opinions :D Most of the discussion is around `null` vs `Optional`, whereas we're talking more about `Optional` vs "object whose internal fields are empty".

My 2 cents (and this really is just my opinion): I mostly agree with what's said in https://nipafx.dev/java-optional-opinions/#expressing-intent -- I think `Optional` is better than `null` and better than "object with empty fields" at expressing "this value is not set, and that's perfectly valid". For the image description in particular:
- I think the problem with null is that you need to hope that each caller remembers to check whether it's null, whereas `Optional` forces callers to check.
- I think using the object with empty fields is possibly misleading, because on first glance it looks like the field is always present, but certain configurations actually mean it's not present.

That being said, I do think it's easy to get carried away with using optionals too frequently, which makes the code too verbose and too difficult to understand (like you said), so we should be careful about where we use them. The image description felt like a good place to use them because the it is literally an optional field, but if we like using `null` or an empty `LocalizedStrings` better, than I'm happy to adjust!

caitlinshk: I added this as a specific implementation detail in https://github.com/civiform/civiform/issues/6007 if that looks good to you?

rockycodes (approved):

caitlinshk:
### Merge activity

* **Nov 21, 5:11 PM**: @caitlinshk started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/5976).

PR #5977, opened 2023-11-15
Replace our accordion with the USWDS accordion component

melanie-exygy:
### Description

This PR replaces our custom accordion element with one from the U.S. Web Design System library.  The reason for this is so that we can adopt the USWDS super banner, which relies on the accordion.  Currently, our accordion is only used in one place, the API docs page, so this change only has a small impact.  I've also re-used most of our custom accordion styles to reduce any visual changes.  

## Release notes

This PR replaces our custom accordion element with one from the U.S. Web Design System library.  The reason for this is so that we can adopt the USWDS super banner, which relies on the accordion. 

1. Log in as a CF Admin.
2. Click on the "API Docs" tab.
3. Note the "How does this work?" accordion button.
4. Click or tab to the accordion and press the space bar to open.
image


browser-test/src/api_docs.test.ts
```diff
@@ -124,10 +118,8 @@ describe('Viewing API docs', () => {
     // Select minimal sample program so the screenshot will be smaller.
     await page.selectOption('#select-slug', {value: 'minimal-sample-program'})
 
+    // Opening the accordion
```
rockycodes, line 121: question (non-blocking): why do we no longer need the `waitForTimeout` calls?

rockycodes: Ohh maybe because with the USWDS classes it's all done with css now?

server/app/views/api/ApiDocsView.java
```diff
@@ -278,9 +278,7 @@ private DivTag getNotes(Http.Request request) {
                 + " Static content questions are not shown on the API Response Preview because"
                 + " they do not include answers to questions."));
 
-    Accordion accordion = new Accordion().setTitle("How does this work?");
-    accordion.addContent(notesTag);
-
-    return accordion.getContainer();
+    return AccordionFactory.buildAccordion(
```
rockycodes, line 281: question (non-blocking): why rename this to AccordionFactory?

server/app/views/components/AccordionFactory.java
```diff
@@ -0,0 +1,56 @@
...
+import j2html.tags.ContainerTag;
+import j2html.tags.specialized.DivTag;
+import java.util.Optional;
+import views.style.ReferenceClasses;
+
+/**
+ * Utility class for building an accordion. This is a modified version of the USWDS accordion
+ * component here: https://designsystem.digital.gov/components/accordion/.
+ */
+public final class AccordionFactory {
```
rockycodes, line 15: suggestion (non-blocking): I have a slight preference for editing the existing class rather than creating a new one since there's some overlap, but I'm curious to know your thinking.

rockycodes (changes requested):
Overall this looks great! I have a few questions I want to chat about before approving.

The only potential blocker is that the header text formatting has changed ("How does this work?"). Is that intentional? It looks pretty different from how it was previously and different than other designs we usually go with so I want to double check.

browser-test/src/api_docs.test.ts
melanie-exygy: The `waitForTimeout` calls were because of the transition time it took for the "^" icon to spin when we open/closed the accordion.  With USWDS, they just swap "+" for "-" or vice versa, without transition time.

server/app/views/api/ApiDocsView.java
melanie-exygy: I found it misleading to have a utility class called Accordion that, when instantiated via the constructor, didn't provide you with an accordion component.  It could have been easy for someone not to realize that they had to call `getContainer`.  I also didn't like the name "container" because, to me, what was meant by "container" was the actual accordion element. 

I feel that it's clearer now.  If you need an accordion, you call AccordionFactory.buildAccordion().  Yes, it isn't following the factory design pattern, but it's a simple factory class.

server/app/views/components/AccordionFactory.java
melanie-exygy: Essentially, I did edit the existing class, but, because I changed the name of the class (see my reasoning above), the filename had to be changed and version control sees this as a completely new file.

melanie-exygy:
> Overall this looks great! I have a few questions I want to chat about before approving.
> 
> The only potential blocker is that the header text formatting has changed ("How does this work?"). Is that intentional? It looks pretty different from how it was previously and different than other designs we usually go with so I want to double check.

I agree that the text looks pretty different.  I left it that way because I wasn't sure how intentional our original text styling was, given that we likely didn't have much UX input at the time.  I figured that USWDS had put more thought into their text styles.  Hm...  I can try to make it more similar to how it was, just so I don't disrupt anything.  Normally, I'd defer to Sijia, but they're at the CF/Exygy meetings today.

server/app/views/api/ApiDocsView.java
rockycodes: That makes sense to me! Thanks for explaining.

rockycodes (approved):
Good call about the former formatting maybe not being intentional. I don't think it's worth you fiddling with the classes to get them to match, but I would prefer to hold off on merging until Sijia has approved.

sijiayam:
@rockycodes I reviewed with Melanie and have approved the differences in visual design. I am comfortable/like the text styles that are coming from the USWDS library.

Issue #5972, opened 2023-11-14
Create Universal questions section and fix sorting in questions list view


nb1701:
After some discussion with @sijiayam , we're going to apply the same sort of design to the question list view as we have in mocks for the question bank, where there is a separate universal questions section at the top and then an "all other questions" section below.

Additionally, the first pass at implementation of the universal badge and sorting in the questions list view did not take into account the sorting options in the dropdown box, which is implemented client-side in the questionBank.ts file.  This needs to be fixed so that universal questions always stay at the top in the universal questions section and are also filtered appropriately when a user is searching for questions.

PR #6017, opened 2023-11-21
Questions list view universal section and sorting

nb1701:
### Description

This makes five changes.

1. There was a bug where when you had a published (lifecycle stage ACTIVE) universal question, if you edited it and unset universal, it would not remove universal from the question until you unset it again.  This was due to how we were creating a new draft of a question, where we copy all of the QuestionTags from the old question directly.  We still have to do this since we don't maintain the state of the export tags in the QuestionDefinition (yet), but now we'll explicitly remove the universal tag if the new QuestionDefinition#isUniversal is false.
2. In the questions list view, universal questions are now in their own separate section.  If no questions are marked universal, it will look like it does today.  If one or more are, you will see them in their own section with a USWDS banner and a "all other questions" section for the rest.  This was in the mocks for the question bank but not the questions list view.  After discussing with @sijiayam , we decided the questions list view should have it as well.
3. Since we now have multiple sections we need to sort independently, this modifies the typescript code to sort all of the sections.
4. Moves the browser test to `civiform_admin_question_list` which is more where it belongs.
5. Adds a line to build.sbt to allow printing to stdout from inside tests. Normally, nothing would get printed when tests are run in forking mode (which is the default).  With this, we can `System.out.println` to our heart's desire.

## Release notes

#### General

#### User visible changes

It looks...not so great on mobile and 200% with elements overlapping, but that's unrelated to this PR.

#### New Features

1. Ensure that removing universal from a published universal question removes it successfully the first time in the new draft.
2. Ensure the questions list view displays universal and other questions appropriately.
3. Ensure the questions list view sorts the question sections appropriately.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/5972


server/app/views/style/BaseStyles.java
```diff
@@ -30,6 +30,23 @@ public final class BaseStyles {
   public static final String BG_CIVIFORM_PURPLE_LIGHT = "bg-civiform-purple-light";
   public static final String TEXT_CIVIFORM_YELLOW = "text-civiform-yellow";
   public static final String BG_CIVIFORM_YELLOW_LIGHT = "bg-civiform-yellow-light";
+  public static final String BG_CIVIFORM_TEAL = "bg-civiform-teal";
+
+  /////////////////////////////////////////////////////////////////////////////////////////////////
+  // USWDS component classes
```
melanie-exygy, line 36: question (non-blocking):  I'm pondering whether there's an advantage to adding the USWDS styles here.  Seems like the class names are already clear.

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -187,22 +188,45 @@ private Pair<DivTag, ImmutableList<Modal>> renderAllQuestionRows(
...
+    ImmutableList<DomContent> universalQuestionContent = universalQuestionRows.build();
+    ImmutableList<DomContent> nonArchivedQuestionContent = nonArchivedQuestionRows.build();
+    ImmutableList<DomContent> archivedQuestionContent = archivedQuestionRows.build();
+    DivTag questionContent = div();
+    if (!universalQuestionContent.isEmpty()) {
+      questionContent.with(
+          div()
+              .withId("questions-list-universal")
+              .withClasses(ReferenceClasses.SORTABLE_QUESTIONS_CONTAINER)
+              .with(h2("Universal questions").withClasses("mt-8", "font-semibold"))
```
melanie-exygy, line 209: suggestion (non-blocking):  Since you reuse the "mt-8" and "font-semibold" classes on each of these h2s, it could be nice to add a style to our `BaseStyles` such as `SEMIBOLD_HEADER`.

server/app/views/style/BaseStyles.java
melanie-exygy: I don't really understand that comment because we use plenty of styles that aren't in `BaseStyles.java` and, as far as I can tell, they aren't stripped out in production.  Regardless, these USWDS styles won't come from the `tailwind.css` stylesheet, they'll be coming from the separate `uswds.min.css` file.

melanie-exygy (approved):

Issue #5812, opened 2023-10-17
Modify question bank for universal questions


nb1701:
Per the [mocks](https://www.figma.com/file/KqfHZYPS3Ffd93aJsfkNYZ/Universal-Questions?type=design&node-id=1-27&mode=design&t=mGC8OAmup0UNw6Gw-0), add a badge to universal questions in the question bank, and the separate "universal questions" with the informational text and the "all other questions" section.  The badge should show the state of either the active version of the question, or the draft version of the question, if it exists, as noted in the [TDD](https://docs.google.com/document/d/1-2stuFeIKWkiDCHZlnytsG8c0q0G_Dl3dnplSY4Wy5E/edit#heading=h.bknox3ib8wpb).

Also, ensure the sorting chosen in the dropdown box works correctly for each sorting option.

PR #6056, opened 2023-11-27
Add universal section to question bank

nb1701:
### Description

Similar to https://github.com/civiform/civiform/pull/6017, this adds a separate universal questions section to the top of the questions bank when the feature flag is set and universal questions are present.  Otherwise, the bank looks the same as before.

This also changes the look of the rows in the question bank based on feedback from @sijiayam.

This removes some extraneous `disableFeatureFlag` calls in the browser tests.  Turns out we don't need to do this to clean up the test since the settings are reset between each test.
https://github.com/civiform/civiform/blob/main/server/app/controllers/dev/seeding/DevDatabaseSeedController.java#L151-L156
https://github.com/civiform/civiform/blob/main/server/app/controllers/dev/FeatureFlagOverrideController.java#L41-L61

This removes some extraneous styles from BaseStyles, since we don't still need to define things in there to prevent the tailwind extractor from removing them.

#### General

#### User visible changes

#### New Features

1. Add a couple regular questions and a couple universal questions.
2. With the feature flag on, ensure the universal questions are in their own section, and ensure sorting works correctly for both sections.
3. With the feature flag off, ensure no separate universal section, and all questions are sorted correctly.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/5812

### DIFF
--- a/server/app/views/components/ProgramQuestionBank.java
+++ b/server/app/views/components/ProgramQuestionBank.java
@@ -33,6 +33,7 @@ import services.program.BlockDefinition;
 import services.program.ProgramDefinition;
 import services.question.types.QuestionDefinition;
 import views.ViewUtils;
+import views.style.AdminStyles;
 import views.style.ReferenceClasses;
 
 /** Contains methods for rendering question bank for an admin to add questions to a program. */
@@ -167,7 +168,7 @@ public final class ProgramQuestionBank {
           div()
               .withId("question-bank-universal")
               .withClasses(ReferenceClasses.SORTABLE_QUESTIONS_CONTAINER)
-              .with(h2("Universal questions").withClasses("mt-8", "font-semibold"))
+              .with(h2("Universal questions").withClasses(AdminStyles.SEMIBOLD_HEADER))
               .with(
                   ViewUtils.makeAlertInfoSlim(
                       "We recommend using all universal questions in your program for personal and"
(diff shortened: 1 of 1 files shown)


server/app/views/components/ProgramQuestionBank.java
```diff
@@ -145,25 +148,49 @@ private FormTag questionBankPanel() {
...
         filterQuestions()
             .sorted(
                 Comparator.<QuestionDefinition, Instant>comparing(
                         qdef -> qdef.getLastModifiedTime().orElse(Instant.EPOCH))
                     .reversed()
                     .thenComparing(qdef -> qdef.getName().toLowerCase(Locale.ROOT)))
             .collect(ImmutableList.toImmutableList());
-
+    ImmutableList<QuestionDefinition> universalQuestions =
+        allQuestions.stream().filter(q -> q.isUniversal()).collect(ImmutableList.toImmutableList());
```
melanie-exygy, line 161: suggestion (non-blocking):  It would be nice for the formatting on line 161 to match what's on lines 163-165.  Is it the formatter that's doing that?

server/app/views/components/ProgramQuestionBank.java
```diff
@@ -145,25 +148,49 @@ private FormTag questionBankPanel() {
...
+    ImmutableList<QuestionDefinition> nonUniversalQuestions =
+        allQuestions.stream()
+            .filter(q -> !q.isUniversal())
+            .collect(ImmutableList.toImmutableList());
+    if (!universalQuestions.isEmpty() && showUniversal) {
+      contentDiv.with(
```
melanie-exygy, line 167: suggestion (non-blocking):  The creation of the question bank subsection div could be separated into a helper function to clean things up and for a little bit less duplication.

melanie-exygy (commented):
Looks good overall.  Just have a few non-blocking suggestions.

melanie-exygy (approved):

PR #6093, opened 2023-12-01
Improvements to the focus outline

### LINKED ISSUE #5947: USWDS Adding Outline Around All Focusable Elements
We recently added the USWDS component library. It is making all hyperlinks and form controls that have focus display a large outline. 

It's not so bad for form controls, but individual cases need some love. See below for a couple examples, but everything should be checked individually.

I believe the purpose of this is to aid in navigation via keyboard. That's actually great cause I use the keyboard for navigation at times. 

But in most cases we should probably be using the css pseudo class `:focus-visible` and not `:focus`. Using :focus-visible` will allow for keyboard navigation to behave as it currently does with the outline showing, but mouse users will see it as it was before without the outline.

### Hyperlinks

For links it's pretty annoying to have the border just show up when you press the mouse button down and it gains focus. 

image

### Buttons

When using a mouse buttons now have a hover color, and then the outline when clicked. You end up with three/four states that appear at different points.

#### Button state examples
image
image
image
image

### Radio Buttons and Checkboxes

Radio buttons and Checkboxes also don't look good. We should have the outline appearing around the label instead of the input itself.

#### Outline showing around radio button only (current)
image

#### Outline showing around the label (ideal)
image

melanie-exygy:
### Description

Added a white offset between interactive elements and the focus outline.  This is especially helpful for color contrast around blue buttons since our focus outline is also blue.

Also, on checkboxes and radio buttons, moved the focus outline to the label rather than the button itself.

- Notice the offset between interactive elements and the focus outline.  For example:

image

- Check radio button and checkbox questions in an application.  Make sure the focus outline is on the label rather than the checkbox or button.

image

- Also check the radio buttons when editing a program:

image

### Issue(s) this completes

Fixes #5947


bion (approved):
looks great!

PR #6100, opened 2023-12-01
Make bin/help more helpful

gwendolyngoetz:
### Description

The previous `bin/help` was getting crufty. 
- The formatting wasn't aligning if a file had multiple `#DOC` tokens
- Grepping wouldn't always show the filename
- Only one filename would be printed if there were multiple files with no `#DOC` help text

This replacement to `bin/help` now defaults to printing like this:
image

*Colors and fonts are subject to the system they are run on and may not match these screenshots*

Additionally the optional flags will allow:
- `-n` to not print colors
- `-s` to print the file name and help text on the same line
- `-h` prints info on these flags

Using `-n` prints in boring black and white. 

Using `-s` will allow for grepping (searching) the help text and have the file name always included. Files missing help text are not display when using this so excess noise isn't included in matches from grep.

image

All filenames missing help text will now be printed when not in `-s` mode
image

### DIFF
--- a/bin/help
+++ b/bin/help
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 # DOC: Print descriptions of each bin script.
-# DOC: Use --help for optional usage flags
+# DOC: Use -h for help to show optional usage flags
 
 source bin/lib.sh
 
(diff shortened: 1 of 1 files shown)


bin/help
```diff
@@ -1,36 +1,102 @@
...
+  ${0} [options] 
 
-for f in bin/* bin/**/*; do
-  # Files containing library code for scripts have .sh or .py file extension,
-  # don't include those or directories.
-  if [[ -d "${f}" || "${f}" =~ .+\.sh || "${f}" =~ .+\.py ]]; then
-    continue
-  fi
+OPTIONS
+  -h    shows help
```
yotommy, line 13: *Optional, non-blocking*: "shows this help message"

yotommy (approved):
Nice, thanks!

bin/help
gwendolyngoetz: done

Issue #6186, opened 2023-12-12
Feature flag: suggest programs on application confirmation page


bion:
### Flag name

`SUGGEST_PROGRAMS_ON_APPLICATION_CONFIRMATION_PAGE`

### Feature description

Add programs to the confirmation screen that an applicant sees after finishing an application, so they are easily able to find another relevant program to apply to.

### Checklist

- [ ] Feature flag created, feature guarded with flag in code
- [ ] Flag [enabled in dev](https://github.com/civiform/civiform/blob/main/server/conf/application.dev.conf)
- [ ] Flag [disabled for browser tests](https://github.com/civiform/civiform/blob/main/server/conf/application.dev-browser-tests.conf)
- [ ] Feature written and guarded with flag, including unit and browser tests that manipulate the state of the flag as needed. Enable the flag for browser tests when it's ready to be tested.
- [ ] Flag enabled in staging and Seattle staging
- [ ] Seattle given option to test manually
- [ ] Flag default changed to true, with PR release notes mentioning the change and what feature it enables. <Edit this issue and note the date it was enabled here>
- [ ] Verified feature is enabled in production build. <Edit this issue and note the date it was enabled here>
- [ ] Flag enabled in production for at least a month
- [ ] Flag removed and feature code unguarded

rockycodes:
Closing because we want this to stay as a feature flag indefinitely.

PR #6187, opened 2023-12-12
Add program upsell feature flag

bion:
Adds the feature flag for https://github.com/civiform/civiform/issues/6186


server/conf/helper/feature-flags.conf
```diff
@@ -60,3 +60,7 @@ universal_questions = ${?UNIVERSAL_QUESTIONS}
 # Program card images
 program_card_images = false
 program_card_images = ${?PROGRAM_CARD_IMAGES}
+
+# Upsell programs
+suggest_programs_on_application_confirmation_page = false
```
azhangggggg, line 65: Not a comment, more like a question - kinda curious about why we set the bool value twice. does this mean it's default false unless `SUGGEST_PROGRAMS_ON_APPLICATION_CONFIRMATION_PAGE` is not `false`?

azhangggggg (approved):

server/conf/helper/feature-flags.conf
bion: It's defaulting it to `false` and then overriding with a value from the environment if the environment variable is set. More info here: https://www.playframework.com/documentation/2.9.x/ConfigFile#Substitutions

Issue #6158, opened 2023-12-08
Fix flaky api docs tests on probers


melanie-exygy:
**Describe the bug**
Several of the tests in `api_docs.test.ts` intermittently fail on probers.  They pass locally and in GH.

**Details**
- See this deploy job for the failure [logs](https://github.com/civiform/civiform-staging-deploy/actions/runs/7122543951).
- Started failing intermittently 2 weeks ago, Nov.22nd.
        - We made a lot of changes that day.
	- Lots of dependency updates, including Jest
	- All the refactoring of model classes to include “Model” in the class name
- On Nov. 16th,  I removed the timeouts before screenshots when replacing the accordion component with the USWDS accordion component since the reason the timeouts had been added in the first place was because of the caret transition time being inconsistent.  With USWDS, we don’t have that caret transition anymore.

- Nature of the failures: 
    - It isn’t the snapshots that are failing.
    - When the minimal-sample-program is selected from the dropdown, the comprehensive program details are still shown below.  
    - It’s a delay in loading the page.  Each dropdown selection reloads the entire page, but the test doesn't seem to wait for that reload to finish.

**Possible solutions**

- We could add a timeout before the expectation, as a band-aid.  Can’t await page content because the promise seems to resolve with the old data before the page has a chance to reload.
- It would be much better to find out what is causing this delay on the probers.
     - Could there be a delay in fetching program details?

melanie-exygy:
It's interesting that the "Views active API docs without logging in" test always passes even though it is doing much the same thing as the other tests.

gwendolyngoetz:
When I finish updating this test I'm going to remove the environment lock out. If it still is an issue we can add it back.

PR #6189, opened 2023-12-12
Skipping unreliable api docs tests on probers for now

melanie-exygy:
### Description

Skipping unreliable api docs tests on probers for now so as not to block deployment, but intend to fix the problem with this issue: 
#6158


rockycodes (approved):
lgtm!

PR #6203, opened 2023-12-15
formatted long string in screen name and description

### LINKED ISSUE #5001: Long program screen names or screen descriptions cause formatting issues
**Describe the bug**
Long screen names cause text to be cut off on the right side of the page and the "Delete screen" button to be out of view. 

**To Reproduce**

1. As a CiviForm admin, go to the Program Dashboard.
2. Select a draft program (or create one) and click on "Edit".  
3. Click the "Add Screen" button on the left.
4. Click "Edit screen name and description".
5. In the "Screen name" text box, type a very long name and/or a very long description. 
6. Click the "Save" button.
7. Notice that the screen name is cut off and you must scroll to the right to see the "Delete screen" button.  Also, notice how the text and "Delete" button are outside of the white container element.

**Expected behavior**
I expect all the text and buttons to stay within the view, without having to scroll to the right.  See below for more technical suggestions.

**Screenshots**
image

- OS: iOS
- Browser Chrome
- Version 113.0.5672.126

Did not verify on mobile.

**Additional context**
As a technical note, I would suggest adding input field length limitations.  This will also prevent other bugs.  I would also suggest that long descriptions wrap rather than continuing off the page.

azhangggggg:
### Description

Formatted screen name and description so that the `Delete screen` button stays in screen when the name and description of the screen is too long.

## Release notes

#### General

#### User visible changes

### Issue(s) this completes

Fixes #5001


melanie-exygy:
Did a quick manual test from your branch and it seems the problem is still there:  
image

Might be good to add a browser test with a snapshot.

Also, did you see the comments on the issue?  What would you think of limiting the number of characters on the name and description fields?

melanie-exygy:
I think that we need to limit the max widths for both the block list on the left and for the block info display.  Then, we also need to make sure the delete button is always within the container.  Honestly, this page needs a major style clean-up, so everything is made a little bit more challenging.  I think fixed max-widths would do the trick for now.

browser-test/src/admin_program_viewing.test.ts
```diff
@@ -144,4 +144,69 @@ describe('admin program view page', () => {
...
+    await adminPrograms.expectQuestionCardWithLabel(
+      'date-q',
+      'required question',
+    )
+    await adminPrograms.expectQuestionCardWithLabel(
+      'email-q',
+      'required question',
+    )
+
+    await validateScreenshot(page, 'view-program-block-2-with-long-description')
```
melanie-exygy, line 205: I'm not seeing this new snapshot pdf in the PR.

azhangggggg: Hi I saw multiple of your comments at once, reply is below.

Might be good to add a browser test with a snapshot. 
--> Browser test added, along with the snapshot.

Also, did you see the comments on the issue? What would you think of limiting the number of characters on the name and description fields? 
--> For key issue (making sure delete button is within the container) I think my fix is adequate. Yes I did. Open to adding a limit to the number of characters if you have a number in mind. Though it sounds like a product decision.

I think that we need to limit the max widths for both the block list on the left and for the block info display. Then, we also need to make sure the delete button is always within the container. Honestly, this page needs a major style clean-up, so everything is made a little bit more challenging. I think fixed max-widths would do the trick for now.
 --> I limited the max-width for the block list on the left. It doesn't seem like limiting it for the block info display is necessary given the delete button is always within the container now. 

I'm not seeing this new snapshot pdf in the PR.
--> It's there now.

browser-test/image_snapshots/admin_program_viewing_test/program-view-universal-questions.png
melanie-exygy, line 1: question:  What are the visual changes on the three snapshots that don't have the long screen name?  I'm not able to tell.  I wish Github would show the diff.

melanie-exygy: Tried to get the diff locally and wasn't able to.  Since we can't tell the difference visually and the tests pass, I think it's safe to call this good.

melanie-exygy (approved):

browser-test/src/admin_program_viewing.test.ts
melanie-exygy: Agree that the character limit should be a product decision.  Let's table that for now.

PR #6244, opened 2023-12-20
Reduce unnecessary calls to create and publish a program

dkatzz:
### Description

When we load program associations (which we do whenever we load the program pages), if that program is active, we are currently doing a publishing dry run of an active and draft version each time that loads twice (which makes many db calls we don't need). We should only need to publish a dry run of a program in the admin flow. With this PR, we are now checking if the program is a draft in a way that doesn't make as many calls and if it is active, we don't load the extra data and do the publishing dry run. We also add some comments to make it more clear when this is happening and remove some print statements from a related test (and make some updates to publishing). We should take a holistic approach on how to separate applicant and admin calls, but this should be a big step in reducing latency for the applicant.

#### General

- [ ] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).

### Issue(s) this completes

### DIFF
--- a/server/app/services/program/ProgramService.java
+++ b/server/app/services/program/ProgramService.java
@@ -9,7 +9,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Streams;
 import com.google.inject.Inject;
 import controllers.BadRequestException;
 import forms.BlockForm;
@@ -226,9 +225,9 @@ public final class ProgramService {
   private CompletionStage<ProgramDefinition> syncProgramAssociations(ProgramModel program) {
     VersionModel activeVersion = versionRepository.getActiveVersion();
     VersionModel maxVersionForProgram =
-      programRepository.getVersionsForProgram(program).stream()
-        .max(Comparator.comparingLong(p -> p.id))
-        .orElseThrow();
+        programRepository.getVersionsForProgram(program).stream()
+            .max(Comparator.comparingLong(p -> p.id))
+            .orElseThrow();
     // If the max version is greater than the active version, it is a draft
     if (maxVersionForProgram.id > activeVersion.id) {
       // This method makes multiple calls to get questions for the active and
@@ -1652,9 +1651,9 @@ public final class ProgramService {
    */
   private CompletionStage<ProgramDefinition> syncProgramDefinitionQuestions(
       ProgramDefinition programDefinition) {
-    // Note: This method is also used for non question updates.  We should
-    // have a focused method for that because getReadOnlyQuestionService() makes
-    // multiple calls to get question data for the active and draft versions.
+    // Note: This method is also used for non question updates.
+    // TODO(#6249) We should have a focused method for that because getReadOnlyQuestionService()
+    // makes multiple calls to get question data for the active and draft versions.
     return questionService
         .getReadOnlyQuestionService()
         .thenApplyAsync(
--- a/server/app/services/program/ProgramService.java
+++ b/server/app/services/program/ProgramService.java
@@ -13,6 +13,7 @@ import com.google.common.collect.Streams;
 import com.google.inject.Inject;
 import controllers.BadRequestException;
 import forms.BlockForm;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dkatzz:
* `main`
  * **PR #6244** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6244?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #6245** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6244?utm_source=stack-comment).

server/app/services/program/ProgramService.java
```diff
@@ -223,27 +223,26 @@ public ImmutableList<ProgramDefinition> getAllProgramDefinitionVersions(long pro
   }
 
   private CompletionStage<ProgramDefinition> syncProgramAssociations(ProgramModel program) {
-    if (isActiveOrDraftProgram(program)) {
+    VersionModel activeVersion = versionRepository.getActiveVersion();
```
yotommy, line 226: **suggestion (non-blocking):** Consider putting this in a helper method. If the helper method is called something like `isDraftVersion()`, then no comment is required here.

yotommy, line 226: **suggestion (non-blocking):** Consider putting this in a helper method. If the helper method is called something like `isDraftVersion()`, then no comment is required here.

server/app/services/program/ProgramService.java
```diff
@@ -223,27 +223,26 @@ public ImmutableList<ProgramDefinition> getAllProgramDefinitionVersions(long pro
...
+    VersionModel maxVersionForProgram =
+        programRepository.getVersionsForProgram(program).stream()
+            .max(Comparator.comparingLong(p -> p.id))
```
yotommy, line 229: **suggestion (non-blocking):** I *think* this could be replaced with `Program::id`. You can decide if that's better or not.

server/app/services/program/ProgramService.java
yotommy: Oh, now I see that you use `maxVersionForProgram` below, so it's good the way it is.

server/app/services/program/ProgramService.java
```diff
@@ -1652,8 +1651,9 @@ private static ErrorAnd<BlockDefinition, CiviFormError> createEmptyBlockDefiniti
    */
   private CompletionStage<ProgramDefinition> syncProgramDefinitionQuestions(
       ProgramDefinition programDefinition) {
-    // Note: This method is also used for non question updates.  It'd likely be
-    // good to have a focused method for that.
+    // Note: This method is also used for non question updates.  We should
```
yotommy, line 1654: **nitpick (blocking):** Add TODO?

server/app/services/program/ProgramService.java
```diff
@@ -223,27 +223,26 @@ public ImmutableList<ProgramDefinition> getAllProgramDefinitionVersions(long pro
...
 
   private CompletionStage<ProgramDefinition> syncProgramAssociations(ProgramModel program) {
-    if (isActiveOrDraftProgram(program)) {
+    VersionModel activeVersion = versionRepository.getActiveVersion();
+    VersionModel maxVersionForProgram =
+        programRepository.getVersionsForProgram(program).stream()
+            .max(Comparator.comparingLong(p -> p.id))
+            .orElseThrow();
+    // If the max version is greater than the active version, it is a draft
+    if (maxVersionForProgram.id > activeVersion.id) {
```
yotommy, line 232: **praise:** This is a great example of a "why" comment. I love finding these when I read unfamiliar code.

yotommy (approved):

server/app/services/program/ProgramService.java
dkatzz: Yep exactly :) was thinking about doing that but then it would involve one more db call to do it in the function and to get the maxVersion for below.

server/app/services/program/ProgramService.java
dkatzz: Tried but I think because id is a property not a method in this case it didn't work.

server/app/services/program/ProgramService.java
dkatzz: Updated! Thanks

PR #6245, opened 2023-12-20
Update cache clearing behavior

dkatzz:
### Description

Remove cases where we're clearing the cache when we don't need to. We thought this was needed, but really we were making unnecessary calls to create and drafts when we didn't need to. Since we aren't doing that anymore after https://github.com/civiform/civiform/pull/6244, we can remove this here. Additionally, we should never clear the cache outside of the dev instance because if there is more than one task, this wouldn't work as intended (entity may exist in one cache even if it was cleared in another). Added a comment to make that clear.

#### General

### Issue(s) this completes

### DIFF
--- a/server/app/controllers/dev/seeding/DevDatabaseSeedController.java
+++ b/server/app/controllers/dev/seeding/DevDatabaseSeedController.java
@@ -134,7 +134,16 @@ public class DevDatabaseSeedController extends Controller {
     return ok("The cache has been cleared");
   }
 
+  /**
+   * Clear cache if it is enabled in settings.
+   *
+   * <p>Note: this is not safe to do in most deployed instances, because there may be multiple
+   * tasks, but we assume all dev instances only have one task.
+   */
   private void clearCacheIfEnabled() {
+    if (!isDevOrStaging) {
+      return;
+    }
     if (settingsManifest.getVersionCacheEnabled()) {
       programsByVersionCache.removeAll().toCompletableFuture().join();
       questionsByVersionCache.removeAll().toCompletableFuture().join();
--- a/server/app/repository/VersionRepository.java
+++ b/server/app/repository/VersionRepository.java
@@ -18,7 +18,6 @@ import io.ebean.Transaction;
 import io.ebean.TxScope;
 import io.ebean.annotation.TxIsolation;
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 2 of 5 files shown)


dkatzz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/civiform/civiform/6245?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #6244** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6244?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #6245** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6245?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6245?utm_source=stack-comment).

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
```diff
@@ -134,6 +134,12 @@ public Result clearCache() {
     return ok("The cache has been cleared");
   }
 
+  /**
+   * Clear cache if it is enabled in settings.
+   *
+   * <p>Note: this is not safe to do in most deployed instances, because there may be multiple
```
yotommy, line 140: **question:** Is it worth protecting this operation with a condition like [this](https://github.com/civiform/civiform/blob/e85e7fc39062cd294e9481ceca59771edbd37c53/server/app/controllers/dev/seeding/DevDatabaseSeedController.java#L61)?

yotommy (approved):

server/app/controllers/dev/seeding/DevDatabaseSeedController.java
dkatzz: Good call out - both callers of this function call that method, so I left it out here, but do you think its worth adding here as well? This is a private function, so it would just be new callers within this file that would need to add a call to the isDevOrStaging method. I can lean either way :)

yotommy: My suggestion is to put it here as well. A bit of extra defensive coding to protect against future refactors where the existing checks may get lost.

dkatzz: Yeah that makes sense to me! Updated

PR #6267, opened 2023-12-22
Create a cache for the full program definition

dkatzz:
### Description

The full program definition is an expensive database lookup operation, since it involves looking up question data.

We should set this program definition in the cache when the program is active or obsolete.

We can create a follow up PR to refactor programModel.getProgramDefinition() calls to use this cached program definition if it is present and also ensure syncProgramDefinitionQuestions and its callers use the cache if its available (if not a draft program).

This PR enables the question_cache_enabled feature flag in tests, and updates some tests to create draft programs instead of active ones when edits are still being made to the program (because otherwise it will be cached)

## Release notes

Get and set full program definition data from the cache (with the question data included), if the question cache feature flag is enabled.

#### General

#### New Features

- [x] Wrote browser tests with the feature flag off and on, etc. (all browser tests have this feature flag enabled)

### Issue(s) this completes

#6466

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -147,8 +147,13 @@ public final class ProgramRepository {
    * <p>Draft program definition data must not be set in the cache.
    */
   public void setFullProgramDefinitionCache(long programId, ProgramDefinition programDefinition) {
-    if (settingsManifest.getQuestionCacheEnabled()) {
-      programDefCache.set(String.valueOf(programId), programDefinition);
+    if (settingsManifest.getQuestionCacheEnabled()
+        // We only set the cache if it hasn't yet been set for the ID.
+        && getFullProgramDefinitionFromCache(programId).isEmpty()) {
+      // We should never set the cache for draft programs.
+      if (!versionRepository.get().isDraftProgram(programId)) {
+        programDefCache.set(String.valueOf(programId), programDefinition);
+      }
     }
   }
 
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -130,15 +130,15 @@ public final class ProgramRepository {
    * Gets the program definition that contains the related question data from the cache (if
    * enabled).
    */
-  public Optional<ProgramDefinition> getProgramDefinitionFromCache(ProgramModel program) {
-    if (settingsManifest.getProgramCacheEnabled()) {
+  public Optional<ProgramDefinition> getFullProgramDefinitionFromCache(ProgramModel program) {
+    if (settingsManifest.getQuestionCacheEnabled()) {
       return programDefCache.get(String.valueOf(program.id));
     }
     return Optional.empty();
   }
 
-  public Optional<ProgramDefinition> getProgramDefinitionFromCache(long programId) {
-    if (settingsManifest.getProgramCacheEnabled()) {
+  public Optional<ProgramDefinition> getFullProgramDefinitionFromCache(long programId) {
+    if (settingsManifest.getQuestionCacheEnabled()) {
       return programDefCache.get(String.valueOf(programId));
     }
     return Optional.empty();
@@ -149,8 +149,8 @@ public final class ProgramRepository {
    *
    * <p>Draft program definition data must not be set in the cache.
    */
-  public void setProgramDefinitionCache(long programId, ProgramDefinition programDefinition) {
-    if (settingsManifest.getProgramCacheEnabled()) {
+  public void setFullProgramDefinitionCache(long programId, ProgramDefinition programDefinition) {
+    if (settingsManifest.getQuestionCacheEnabled()) {
       programDefCache.set(String.valueOf(programId), programDefinition);
     }
   }
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -131,10 +131,7 @@ public final class ProgramRepository {
    * enabled).
    */
   public Optional<ProgramDefinition> getFullProgramDefinitionFromCache(ProgramModel program) {
-    if (settingsManifest.getQuestionCacheEnabled()) {
-      return programDefCache.get(String.valueOf(program.id));
-    }
-    return Optional.empty();
+    return getFullProgramDefinitionFromCache(program.id);
   }
 
   public Optional<ProgramDefinition> getFullProgramDefinitionFromCache(long programId) {
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -54,6 +54,7 @@ public final class ProgramRepository {
   private final Provider<VersionRepository> versionRepository;
   private final SettingsManifest settingsManifest;
   private final SyncCacheApi programCache;
+  private final SyncCacheApi programDefCache;
   private final SyncCacheApi versionsByProgramCache;
 
   @Inject
@@ -62,12 +63,14 @@ public final class ProgramRepository {
       Provider<VersionRepository> versionRepository,
       SettingsManifest settingsManifest,
       @NamedCache("program") SyncCacheApi programCache,
+      @NamedCache("program-definition") SyncCacheApi programDefCache,
       @NamedCache("program-versions") SyncCacheApi versionsByProgramCache) {
     this.database = DB.getDefault();
     this.executionContext = checkNotNull(executionContext);
     this.versionRepository = checkNotNull(versionRepository);
     this.settingsManifest = checkNotNull(settingsManifest);
     this.programCache = checkNotNull(programCache);
+    this.programDefCache = checkNotNull(programDefCache);
     this.versionsByProgramCache = checkNotNull(versionsByProgramCache);
   }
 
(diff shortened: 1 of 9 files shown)


dkatzz:
* `main`
  * **PR #6267** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6267?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6267?utm_source=stack-comment).

server/app/repository/ProgramRepository.java
```diff
@@ -123,6 +126,35 @@ public ImmutableSet<String> getAllProgramNames() {
     return names.build();
   }
 
+  /**
+   * Gets the program definition that contains the related question data from the cache (if
+   * enabled).
+   */
+  public Optional<ProgramDefinition> getFullProgramDefinitionFromCache(ProgramModel program) {
+    if (settingsManifest.getQuestionCacheEnabled()) {
+      return programDefCache.get(String.valueOf(program.id));
```
caitlinshk, line 135: Question (non-blocking): I'm assuming the cache only supports lookups with strings, not with longs, which is why you need the `String.valueOf`?

server/app/repository/ProgramRepository.java
```diff
@@ -123,6 +126,35 @@ public ImmutableSet<String> getAllProgramNames() {
...
+    if (settingsManifest.getQuestionCacheEnabled()) {
+      return programDefCache.get(String.valueOf(programId));
+    }
+    return Optional.empty();
+  }
+
+  /**
+   * Sets the program definition that contains the related question data in the cache (if enabled).
+   *
+   * <p>Draft program definition data must not be set in the cache.
```
caitlinshk, line 150: question (blocking): Is there a way to assert that the `programDefinition` isn't a draft before saving it in the cache?

server/app/services/program/ProgramService.java
```diff
@@ -1424,9 +1444,13 @@ public CompletionStage<ImmutableList<ProgramDefinition>> syncQuestionsToProgramD
                   if (!programDef.hasEligibilityEnabled()) {
                     return programDef;
                   }
+                  Long programId = programDef.id();
```
caitlinshk, line 1447: Nitpick (non-blocking): Could this be a `long` instead of a `Long`?

server/conf/application.conf
```diff
@@ -303,7 +303,7 @@ play.ws {
 #
 play.cache {
   # Specific caches can be injected using the @NamedCache annotation.
-  bindCaches = ["api-keys", "monthly-reporting-data", "version-programs", "version-questions", "program", "program-versions"]
+  bindCaches = ["api-keys", "monthly-reporting-data", "version-programs", "version-questions", "program", "program-versions","program-definition"]
```
caitlinshk, line 306: Nitpick: Add a space before `"program-definition"`?

server/test/services/program/ProgramServiceTest.java
```diff
@@ -1063,6 +1070,31 @@ public void getProgramDefinitionAsync_constructsQuestionDefinitions() throws Exc
     assertThat(foundQuestion).isInstanceOf(NameQuestionDefinition.class);
   }
 
+  @Test
+  public void getProgramDefinitionAsync_setsCacheWithQuestionDefinitions() throws Exception {
+    QuestionDefinition question = nameQuestion;
+    ProgramDefinition program =
+        ProgramBuilder.newActiveProgram()
```
caitlinshk, line 1077: Question (non-blocking): Why is this test able to use `newActiveProgram` but other tests needed to be changed to use `newDraftProgram`?

server/app/repository/ProgramRepository.java
```diff
@@ -123,6 +126,35 @@ public ImmutableSet<String> getAllProgramNames() {
...
+   */
+  public void setFullProgramDefinitionCache(long programId, ProgramDefinition programDefinition) {
```
caitlinshk, line 152: question (non-blocking): Are there any times we should remove a program definition from the cache?

server/test/repository/ProgramRepositoryTest.java
```diff
@@ -58,13 +59,15 @@ public void setup() {
...
     versionsByProgramCache = instanceOf(SyncCacheApi.class);
     repo =
         new ProgramRepository(
             instanceOf(DatabaseExecutionContext.class),
             Providers.of(versionRepo),
             mockSettingsManifest,
             programCache,
+            programDefCache,
             versionsByProgramCache);
   }
```
caitlinshk, line 72: question (blocking): Is it worth adding a short test here that verifies the new set & get methods work? (And maybe that they don't do anything if the flag is off?)

server/app/services/program/ProgramService.java
```diff
@@ -1403,8 +1421,10 @@ public CompletionStage<ImmutableList<ProgramDefinition>> syncQuestionsToProgramD
     for (ProgramDefinition programDef : programDefinitions) {
       ProgramModel p = programDef.toProgram();
       p.refresh();
-      // We only need to get the question data if the program has eligibility conditions.
-      if (programDef.hasEligibilityEnabled()) {
+      // We only need to get the question data if the program has eligibility conditions and the
```
caitlinshk, line 1424: question (non-blocking): I have a question mostly for my own understanding. I know that if an applicant starts applying to a program, then an admin makes changes to the program and publishes a new version, the applicant should still see the old version. Each time an admin makes a new version of a program, it gets a new program ID right? So the cache should still work correctly for applicants filling out old versions of programs because the ID will still be tied to the old version of the program definition in the cache?

caitlinshk (changes requested):

server/app/repository/ProgramRepository.java
dkatzz: Yeah, exactly! https://www.playframework.com/documentation/2.6.x/api/java/play/cache/SyncCacheApi.html

server/app/repository/ProgramRepository.java
dkatzz: Yeah this is a good question. The reason I didn't do it is because we only call this function if it isn't a draft and finding out if a program is a draft requires an extra lookup to get the active version and then get the versions for this program ID, so I thought it may not be worth the extra call. But if we think this may be used in situations it shouldn't it may be worth doing this. This is what the extra code would look like. Let me know what you think!

```
    VersionModel activeVersion = versionRepository.getActiveVersion();
    VersionModel maxVersionForProgram =
        programRepository.getVersionsForProgram(program).stream()
            .max(Comparator.comparingLong(p -> p.id))
            .orElseThrow();
    // If the max version is less than or equal to the active version, it is not a draft
    if (maxVersionForProgram.id <= activeVersion.id) {
```

server/app/services/program/ProgramService.java
dkatzz: oops yes thank you!!

server/conf/application.conf
dkatzz: Updated thanks!!

server/test/services/program/ProgramServiceTest.java
dkatzz: Good question! Basically, we should set it as a draft if we're still planning to make changes to the program (which the other tests do) and we don't want to cache it. In this case, we don't make changes after creating the program and we want it to be cached, so we can set it as active. I added a test case as well to ensure if its a draft program, it doesn't get cached.

server/app/repository/ProgramRepository.java
dkatzz: We only need to remove it from the cache in the DevDatabaseSeedController so that for the browser test runs the cache is cleared when the database is reset. In application code, we don't have to clear the cache because once a program is set to active or obsolete, the program data for that ID will never change.

server/test/repository/ProgramRepositoryTest.java
dkatzz: Yep good call - added!

server/app/services/program/ProgramService.java
dkatzz: Yeah that is all correct! Once a program is published, there will be no changes to that program (when looked up by ID) and when that program is edited, it will keep the name and get a new ID. The only time a program can change is if its in draft mode, which is why we only cache programs that are active or obsolete.

server/app/repository/ProgramRepository.java
caitlinshk: I think it's probably still worth it, just because it's easy for someone in the future to call this method without remembering to read the documentation.

This method will only be called when an admin does a publishing action, right? So the extra call shouldn't happen that frequently and that seems worth it for the extra safety guarantee. If this method could also be called on the applicant side then I agree with you it wouldn't be worth it.

server/app/repository/ProgramRepository.java
caitlinshk: Got it, thanks!!

server/app/services/program/ProgramService.java
caitlinshk: Excellent, thanks for the clarification!

server/test/repository/ProgramRepositoryTest.java
```diff
@@ -100,6 +111,42 @@ public void lookupProgram_usesCacheWhenEnabled() {
     assertThat(programCache.get(String.valueOf(two.id))).hasValue(found);
   }
 
+  @Test
+  public void getFullProgramDefinitionFromCache_getsFromCacheWhenPresent() {
+    Mockito.when(mockSettingsManifest.getQuestionCacheEnabled()).thenReturn(true);
+    ProgramModel program = resourceCreator.insertActiveProgram("testInCache");
+    repo.setFullProgramDefinitionCache(program.id, program.getProgramDefinition());
+    Optional<ProgramDefinition> programDefFromCache =
```
caitlinshk, line 119: Nitpick (non-blocking): It's nice to leave an empty line between arrange / act / assert stages, so might be just a tad more readable if you have an empty line above the `Optional<ProgramDefinition> programDefFromCache = repo.getFullProgramDefinitionFromCache(program);` line.

server/test/services/program/ProgramServiceTest.java
caitlinshk: Ah got it! Does the `.withBlock().withRequiredQuestionDefinition(question)` not count as making changes after creating the program?

caitlinshk (approved):

server/test/repository/ProgramRepositoryTest.java
dkatzz: Good call - updated!

server/test/services/program/ProgramServiceTest.java
dkatzz: It'll all get built as one I think and the program def won't be cached until we make the call to get the program definition.

server/app/repository/ProgramRepository.java
caitlinshk: Ah got it, thanks for the clarification that it could happen on the applicant side.

PR #6287, opened 2024-01-02
Fix single program publish bug

### LINKED ISSUE #6194: Unable to Publish Programs With Certain Characters in Name
----
This only occurs when the `UNIVERSAL_QUESTIONS` flag is enabled.
----

Seattle's database has been around quite a while. As such some older data does not necessarily conform to newer validation requirements as the UI now requires.

I've found the older program "Gold Card (HSD)" which I put into draft mode on my local machine and the test server generate javascript errors and prevent `Publish all drafts` and `Publish` working for this program. Other programs that don't have this text can be published individually.

I tried swapping the modal id from using `program.adminName` to `program.slug` but that caused the publish button to disappear from the screen.

# Javascript Error

image

# Database Date

image

rockycodes:
### Description

When the Universal Questions feature flag is turned on, single program publish doesn't work with program admin names that include invalid characters. These characters are no longer allowed to be saved as `adminName` but this breaks with legacy data which impacts at least Seattle.

This fix changes the logic to reference `program.slug` instead of `program.adminName` when creating the reference name for the modal. `program.slug` is a [created from the admin name](https://github.com/civiform/civiform/blob/5e665380c10d7db32b1e8f1abdc264ada2455f68/server/app/services/program/ProgramDefinition.java#L495) with special characters stripped out.

#### General

1. Login as an admin and create a program
2. From the admin program index page, attempt to publish that one program
3. Program should publish successfully

Also try this with legacy data that includes invalid characters in `adminName` (Eg. "Gold Card (HSD)")

### Issue(s) this completes

Fixes #6194 ;

### DIFF
--- a/browser-test/src/admin_program_list.test.ts
+++ b/browser-test/src/admin_program_list.test.ts
@@ -173,12 +173,12 @@ describe('Program list page.', () => {
     await adminPrograms.addQuestionFromQuestionBank(nameQuestion)
 
     await adminPrograms.gotoAdminProgramsPage()
-    await page.click('#program-one-publish-modal-button')
-    expect(await page.innerText('#program-one-publish-modal')).toContain(
+    await page.click('#publish-modal-program-one-button')
+    expect(await page.innerText('#publish-modal-program-one')).toContain(
       'Are you sure you want to publish program one and all of its draft questions?',
     )
     // Warning should not show because there are no universal questions
-    expect(await page.innerText('#program-one-publish-modal')).not.toContain(
+    expect(await page.innerText('#publish-modal-program-one')).not.toContain(
       'Warning: This program does not use all recommended universal questions.',
     )
     await validateScreenshot(page, 'publish-single-program-modal-no-warning')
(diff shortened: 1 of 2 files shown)


browser-test/src/admin_program_list.test.ts
```diff
@@ -173,12 +173,12 @@ describe('Program list page.', () => {
     await adminPrograms.addQuestionFromQuestionBank(nameQuestion)
 
     await adminPrograms.gotoAdminProgramsPage()
-    await page.click('#program-one-publish-modal-button')
-    expect(await page.innerText('#program-one-publish-modal')).toContain(
+    await page.click('#publish-modal-program-one-button')
```
gwendolyngoetz, line 176: Non-blocking: Consider putting all the various `#publish-modal-xxx` and `#publish-modal-xxx-button` strings into a variable.

server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -263,7 +263,7 @@ private ImmutableList<Modal> buildPublishSingleProgramModals(
                           "flex", "flex-col", StyleUtils.responsiveMedium("flex-row"), "py-4");
 
               return Modal.builder()
-                  .setModalId(program.adminName() + "-publish-modal")
+                  .setModalId("publish-modal-" + program.slug())
```
gwendolyngoetz, line 266: Blocking: Recommend creating a method you can pass a program to that will build the modal ID so it doesn't get out of sync between the spots you need it.

gwendolyngoetz (changes requested):

server/app/views/admin/programs/ProgramIndexView.java
rockycodes: good idea. done.

browser-test/src/admin_program_list.test.ts
rockycodes: done

gwendolyngoetz (approved):

PR #6311, opened 2024-01-08
Update release date in oncall issue template

gwendolyngoetz:
Update release date in oncall issue template from Wednesday to Tuesday


avaleske (approved):
Ah I only searched for "wednesday"!

PR #6315, opened 2024-01-08
Annazhang include hidden questions in application pdf

azhangggggg:
### Description

## Release notes

#### General

- [ ] Added the correct label: < feature | enhancement | bug | dependencies | infrastructure | ignore-for-release | database >
- [ ] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/services/export/PdfExporter.java
```diff
@@ -184,6 +192,26 @@ private byte[] buildPDF(
           document.add(eligibility);
         }
       }
+      if (answersOnlyHidden.size() > 0) {
```
gwendolyngoetz, line 195:
```suggestion
      if (!answersOnlyHidden.isEmpty()) {
```

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -363,7 +418,8 @@ private ImmutableList<Block> getBlocks(
               blockDefinition,
               applicantData,
               maybeRepeatedEntity);
-      if (includeBlockIfTrue.test(block)) {
+      if ((includeBlockIfTrue.test(block) && !includeHidden)
+          || (!includeBlockIfTrue.test(block) && includeHidden)) {
```
gwendolyngoetz, line 422: (blocking) This is a getting a little hard to follow. Can you move the check into its own method and include comments on what it's doing.

Something like

```Java
/**
 *
 */
private Boolean includeBlockInList(Predicate<Block> includeBlockIfTrue, Boolean includeHidden) {
   return (includeBlockIfTrue.test(block) && !includeHidden) || (!includeBlockIfTrue.test(block) && includeHidden);
}
```

or maybe more verbose for future readers

```Java
/**
 *
 */
private Boolean includeBlockInList(Predicate<Block> includeBlockIfTrue, Boolean includeHidden) {
  // What case does this cover?
  if (includeBlockIfTrue.test(block) && !includeHidden) {
    return true;
  }

  // What case does this cover?
  if (!includeBlockIfTrue.test(block) && includeHidden) {
    return true;
  }

  // What case does this cover?
   return false;
}
```

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -275,6 +325,9 @@ public ImmutableList<AnswerData> getSummaryDataOnlyActive() {
         String questionText = applicantQuestion.getQuestionText();
         String questionTextForScreenReader = applicantQuestion.getQuestionTextForScreenReader();
         String answerText = applicantQuestion.getQuestion().getAnswerString();
+        if (hiddenBlocks) {
+          answerText = "N/A";
```
gwendolyngoetz, line 329: (question) Does this need to be localized?

gwendolyngoetz (changes requested):

server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -117,8 +129,11 @@ public interface ReadOnlyApplicantProgramService {
   /** Returns if the block has an eligibility predicate. */
   boolean blockHasEligibilityPredicate(String blockId);
 
-  /** Returns if the block eligibility criteria are met. */
-  boolean isBlockEligible(String blockId);
+  /** Returns if the hidden block eligibility criteria are met. */
+  boolean isHiddenBlockEligible(String blockId);
```
avaleske, line 133: *question*: Is this used anywhere outside of a test? If not, do we need it?

server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -65,6 +71,9 @@ public interface ReadOnlyApplicantProgramService {
    */
   ImmutableList<ApplicantQuestion> getIneligibleQuestions();
 
+  /** Get the hidden block with the given block ID if there is one. It is emptry if there isn't. */
```
avaleske, line 74: *nit* (typo): empty

server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -35,6 +35,12 @@ public interface ReadOnlyApplicantProgramService {
    */
   ImmutableList<Block> getAllActiveBlocks();
 
+  /**
+   * Get the {@link Block}s for this program and applicant. This includes blocks that are hidden
```
avaleske, line 39: *suggestion*: This should say it's "only" hidden blocks, not that it "includes" hidden blocks, right?

server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -92,6 +101,9 @@ public interface ReadOnlyApplicantProgramService {
    */
   Optional<Block> getFirstIncompleteBlockExcludingStatic();
 
+  /** Returns summary data for each question in this application. */
```
avaleske, line 104: *suggestion*: This is only summary data for the hidden blocks, right, not for "each question"? This comment should specify that it only returns summary data for the hidden blocks.

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -152,10 +164,18 @@ public boolean isBlockEligible(String blockId) {
...
+      allActiveBlockList = getBlocks(this::showBlock, /* includeHidden= */ false);
     }
-    return allBlockList;
+    return allActiveBlockList;
+  }
+
+  @Override
+  public ImmutableList<Block> getAllHiddenBlocks() {
+    if (allHiddenBlockList == null) {
+      allHiddenBlockList = getBlocks(this::showBlock, /* includeHidden= */ true);
```
avaleske, line 176: *question*: Does `includeHidden` mean that it returns active _and_ hidden blocks, or only hidden blocks?

If it's active and hidden, then calls above this might get active blocks unexpectedly since `getAllHiddenBlocks()` implies you're only getting the hidden ones.

*suggestion*:
If it's only hidden blocks, then we should rename the parameter to `onlyHidden` so it's more clear.

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
avaleske: +1.

Another way to simplify this might be to remove the check for `includeHidden` and flip the logic in the includeBlockIfTrue predicate. so then line 168 might look like
`allActiveBlockList = getBlocks((block) -> showBlock(block));`

but line 178 might look like
`allHiddenBlockList = getBlocks(block -> !showBlock(block));`

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
avaleske: See also the note below about combining `includeHidden` with the `includeIfTrue` predicate.

server/app/services/export/PdfExporter.java
```diff
@@ -184,6 +192,26 @@ private byte[] buildPDF(
...
+        document.add(Chunk.NEWLINE);
+        Paragraph hiddenText =
+            new Paragraph(
+                "Hidden Questions : ", FontFactory.getFont(FontFactory.HELVETICA_BOLD, 12));
```
avaleske, line 199: _suggestion_: If there are other section headings in the doc, can we match their font size? It's currently the same font size as a question, which makes it hard to skim.

server/test/services/export/PdfExporterTest.java
```diff
@@ -150,6 +153,30 @@ public void validatePDFExport_OptionalFileUploadWithoutFile()
     }
   }
 
+  @Test
+  public void validatePDFExport_visibility() throws IOException, DocumentException {
```
avaleske, line 157: *suggestion*: it should be clear from the test name what failed, if the test fails. so something like `validatePdfExport_hiddenQuestionAreIncluded` might be clearer

*also nit, optional*: Three letter acronyms should only have the first letter capitalized, but then that wouldn't match the other method names in this file. Up to you which way you want to go.

avaleske (changes requested):
Thanks for these changes!

I have a few suggestions but overall it looks good.

Also, are there tests for the controllers? As it is, we wouldn't catch a bug that accidentally started including the hidden questions in the user's PDF export

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
azhangggggg: Changes made. Thank you both for the suggestions!

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
azhangggggg: Not this doesn't. changes made.

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
```diff
@@ -165,15 +161,16 @@ private boolean isBlockEligible(Block block) {
...
+      allActiveBlockList = getBlocks((block) -> showBlock(block));
     }
     return allActiveBlockList;
   }
 
   @Override
   public ImmutableList<Block> getAllHiddenBlocks() {
     if (allHiddenBlockList == null) {
-      allHiddenBlockList = getBlocks(this::showBlock, /* includeHidden= */ true);
+      //
```
avaleske, line 172: nit (blocking): can we remove this commented line?

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
avaleske: I think @gwendolyngoetz was asking if the "N/A" text needs to be translated into the user's language, not if it should not be a local variable.

The PDF probably should be translated, but other text in it is not so we shouldn't block on it. For example, the labels on each field: https://github.com/civiform/civiform/blob/cce0552baec4746f118423a64f6c37c09c7f2c5d/server/app/services/export/PdfExporter.java#L113C16-L113C31

avaleske (approved):
Thanks for these changes!

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
azhangggggg: yea for sure. deleted.

server/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java
azhangggggg: Thanks for providing more context!

PR #6408, opened 2024-01-19
Fixed TI client search bug so that cleared search parameters return the full client list

melanie-exygy:
### Description

When searching for clients on the TI dashboard, if the user cleared their search parameters, an empty list was returned.  This meant that if a TI searched for a client and then wanted to see the full list afterwards, they were not able to do that.  This PR fixes that bug so that when a client clears their search parameters, they are able to see the full list of clients.

1. Log in as a trusted intermediary and go to the TI dashboard.
2. Search for a client either by name or date of birth.
3. Delete the content from the search fields and click "Search".
4. Note that the full client list is shown.

### DIFF
--- a/server/app/services/ti/TrustedIntermediaryService.java
+++ b/server/app/services/ti/TrustedIntermediaryService.java
@@ -256,7 +256,12 @@ public final class TrustedIntermediaryService {
   public TrustedIntermediarySearchResult getManagedAccounts(
       SearchParameters searchParameters, TrustedIntermediaryGroupModel tiGroup) {
     ImmutableList<AccountModel> allAccounts = tiGroup.getManagedAccounts();
-    if (searchParameters.nameQuery().isEmpty() && searchParameters.dateQuery().isEmpty()) {
+    if ((searchParameters.nameQuery().isEmpty() && searchParameters.dateQuery().isEmpty()) ||
+      (searchParameters.nameQuery().isPresent() &&
+        searchParameters.nameQuery().get().isEmpty() &&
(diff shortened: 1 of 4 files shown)


server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -256,7 +256,13 @@ public Form<EditTiClientInfoForm> updateClientInfo(
   public TrustedIntermediarySearchResult getManagedAccounts(
       SearchParameters searchParameters, TrustedIntermediaryGroupModel tiGroup) {
     ImmutableList<AccountModel> allAccounts = tiGroup.getManagedAccounts();
-    if (searchParameters.nameQuery().isEmpty() && searchParameters.dateQuery().isEmpty()) {
+    /* Filtering to transform empty strings into empty optionals.
+    Empty parameters should return all accounts. */
+    Optional<String> filteredNameQuery =
+        searchParameters.nameQuery().filter(s -> !s.isEmpty() && !s.isBlank());
```
swatkat1, line 262: are we handling null cases?

melanie-exygy: Since the `nameQuery` and `dateQuery` fields are of type `Optional`, they're initialized as empty `Optional`s rather than `null`.  Also, in practice, we set them to empty strings as a default here:  https://github.com/civiform/civiform/blob/40e4fa72e40a2ef0fb7f7c1117753bb9d23d6086/server/app/views/applicant/TrustedIntermediaryDashboardView.java#L118.

swatkat1 (approved):

PR #6425, opened 2024-01-22
[Program Card Images] Add CTA from program image upload page to translations page

### LINKED ISSUE #6001: [Program Card Images] [Admin] Add CTA from program image page to program translations page
Sub-issue of main epic issue #5676

caitlinshk:
### Description

For program images, admins need to provide an alt-text description of the image and that alt-text needs to be translated to all the languages supported by the CE. This PR adds a button from the program image page to the translations page to encourage admins to add translations.

#### General

#### User visible changes

1. Enable `PROGRAM_CARD_IMAGES` flag.
2. As an admin, navigate to the program images upload page.
3. Verify there's a "Manage translations" button that's disabled if there's no description set but enabled if there is a description.
4. Click the "Manage translations" button and verify it takes you to the translations page. Verify that the translations page has a entry for the program description to be translated into each locale.

### Issue(s) this completes

Fixes #6001 

Part of epic #5676


caitlinshk:
* `main`
  * **PR #6425** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6425?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6425?utm_source=stack-comment).

rockycodes (approved):
This looks great! From testing this flow, I realized it's possible to add alt text even if I haven't uploaded an image which is a little weird. Should that be the case or should we disable the alt text section unless an image is uploaded?

caitlinshk:
> This looks great! From testing this flow, I realized it's possible to add alt text even if I haven't uploaded an image which is a little weird. Should that be the case or should we disable the alt text section unless an image is uploaded?

So, the plan is actually do to the *opposite*: Because we want to require admins to provide alt-text, we're planning on disabling the image upload until alt-text is provided. It does feel a little backwards -- I agree with your intuition that admins want to first upload an image then write the text to describe it. But, I couldn't figure out a good way to enforce the alt-text requirement otherwise.

One option is to change the program card view to not show the image if there's no alt-text, but I think it's a bit hard to convey that to the admin -- they'd upload an image and save it, but the image still wouldn't show on the card preview and I think that's more confusing.

If you have ideas for requiring alt-text and also keeping a good user flow I'm very open to them!

Note: https://github.com/civiform/civiform/pull/6447 is the PR that makes alt-text required if you wanted to try that out and see what you think. I want to figure out one more UI change, which is why I haven't officially sent it for review yet.

rockycodes:
> So, the plan is actually do to the opposite: Because we want to require admins to provide alt-text, we're planning on disabling the image upload until alt-text is provided. It does feel a little backwards -- I agree with your intuition that admins want to first upload an image then write the text to describe it. But, I couldn't figure out a good way to enforce the alt-text requirement otherwise.

Ah I see what you mean, that's tricky. I tried out your other PR and it makes sense to me to require the alt text first. One idea I have is to flip the order of the elements so alt text comes first vertically on the page. I think that would help make it clear that needs to be filled in first.

Another couple of ideas I have if we don't want people to be able to save alt text without an image (which maybe doesn't actually matter?):
1) Disable alt text until an image is uploaded, but throw an error if they try to save the image without adding alt text (this is also kind of convoluted)
2) Have one save button for both image and alt text and throw an error if either is blank (I feel like that might require some significant refactor at this stage, so maybe not worth it)

caitlinshk:
I like flipping the element order so the alt-text is first and the file upload is second. Ron and Shreya wanted the image upload to be first because that's the most important part of the page, but I think the user flow makes more sense to have alt-text first and image upload second. I'm going to make the file upload UI larger, so if it's second on the page but larger I think that'll be fine.

For being able to save alt text without an image: I'm not sure how to prevent that if we're requiring users to save alt text before they upload an image. We unfortunately can't have one save button for both because the file upload needs to send the file to AWS first, and once AWS responds that the file was successfully uploaded the AWS response won't have anything else that was filled out in the form. (That's why the file upload question needs to be on its own screen with no other question.)

WDYT about switching the element order in the other PR and being okay with the admin saving alt-text but no image?

rockycodes:
Sounds good to me!

PR #6225, opened 2023-12-19
Add main page text revisions to messages file for translation

dkatzz:
### Description

Add text to the messages file that we plan to use, based on suggestions from Leslie https://github.com/civiform/civiform/issues/5893. 

See full before and after text [here](https://docs.google.com/spreadsheets/d/1KkGB8AsvxuSEJipVXnZIuXqshMUSh6KjBldr-Zv_gkI/edit#gid=0)

These will be swapped in for the "before" texts once we get the translations back (likely in early - mid January)

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).

PR #6429, opened 2024-01-22
Revise messages on main page

### LINKED ISSUE #5893: Main Page Text Revisions
- Change From: Save time when applying for benefits
 - To: Save time finding and applying for programs and services. 
 
 - Change From: Log in to your account, or create an account, and apply for many benefits without having to enter the same information again. You can also edit applications at any time and check your application statuses.
 - To: You can continue as a guest or create an account. An account allows you to save your information, edit applications and check application statuses. 
 
 - Change From: Find services that can help you
 - To: Get Started

In general, Arkansas would like to stay away from the term "benefits" whenever possible.

azhangggggg:
### Description

Made changes on some messages on main page, see the details [here](https://docs.google.com/spreadsheets/d/1KkGB8AsvxuSEJipVXnZIuXqshMUSh6KjBldr-Zv_gkI/edit#gid=0)

We had originally added these translations to the messages file in order to get them translated before making this change, so we're deleting the legacy translations and renaming messages that have the suffix of `new`.

#### General

#### User visible changes

### Issue(s) this completes

Fixes #5893


server/conf/i18n/messages
azhangggggg, line 1: note: I kept the variable name `findServicesSection` as is since it makes more sense than `getStartedSection` imo. But can make the change if the reviewer thinks that it's better.

server/conf/i18n/messages
```diff
@@ -88,16 +88,16 @@ content.findPrograms=Find programs
 content.description=CiviForm helps you find benefits you may be eligible for in {0}. Get started by choosing an application below.
 # Content that explains the benefit of using CiviForm. We'll fill in the name of the city or state.
 content.findProgramsDescription=CiviForm helps you find programs you may be eligible for in {0}. Get started by choosing an application below.
-content.saveTime=Save time when applying for benefits
+content.saveTime=Save time finding and applying for programs and services
 # Content on the home page to describe the benefit of CiviForm
 content.saveTimeServices=Save time applying for programs and services
```
dkatzz, line 93: **suggestion (blocking):** There are already translations for all of these below the original message, so instead of changing these files, we should the reference in the message key files. https://github.com/civiform/civiform/blob/ceb257a20b349ce637e46b14083b7c6b7f5314ef/server/app/services/MessageKey.java#L67. Updating that should keep the images the same, but will ensure that we have the translations that were done in https://github.com/civiform/civiform/issues/5948. By changing this file, it only changes the english version of the strings.

dkatzz: Here is the PR before the translation request https://github.com/civiform/civiform/pull/6225/files. We should determine if it makes sense to use the new key name in `MessageKey.java` instead or replace the key (like if it says `new`). If we replace the key, we should change it in the other message files as well.

rockycodes (approved):

server/conf/i18n/messages
dkatzz: Updated to change the keys or use the updated keys.

PR #6433, opened 2024-01-23
Combining the Phone validation method from TIService class and PhoneQuestion class

### LINKED ISSUE #6382: Combine the phone number validations in PhoneQuestion.java and TrustedIntermediaryService.java into a Util class.
**Describe the bug**
Combine the phone number validations in PhoneQuestion.java and TrustedIntermediaryService.java into a Util class.

swatkat1:
### Description

Combining the Phone validation method from TIService class and PhoneQuestion class.

### Release notes

Combining the Phone validation method from TIService class and PhoneQuestion class.

#### General

### Issue(s) this completes

Fixes #6382


browser-test/src/phone.test.ts
```diff
@@ -137,14 +137,28 @@ describe('phone question for applicant flow', () => {
...
         'United States',
         '123###1212',
       )
 
+      await applicantQuestions.clickNext()
+      const countryCodeId = '.cf-phone-number'
+      expect(await page.innerText(countryCodeId)).toContain(
+        'The phone number must contain only ten digits',
+      )
+    })
```
caitlinshk, line 153: Nitpick (non-blocking): Can we keep any empty line between each test case? (It looks like this file currently doesn't have empty lines between each case, so if you'd rather follow the format of the rest of the file and *not* have an empty line between each test case, fine by me!)

browser-test/src/phone.test.ts
```diff
@@ -137,14 +137,28 @@ describe('phone question for applicant flow', () => {
         'This phone number is invalid',
       )
     })
-    it('invalid characters in phone numbers', async () => {
+    it('invalid length of phone numbers', async () => {
```
caitlinshk, line 140: Question (blocking): My initial thought was that since this test is only for invalid length not invalid characters, maybe we should remove the invalid characters so that this test case is just focused on length. But then I noticed that line 168 also has a test case for invalid length, so now I'm not quite sure what this is testing.

I'm wondering: Is this test different in some way from the test case on line 168? Is this testing a phone number has the right length *before* removing invalid characters but the wrong length *after* removing the invalid characters? If so, can we name this test case something clearer like 'invalid length of phone number when only valid characters included'?

(Feel free to word-smith that test name, I think we just need to make it clear how this test is different from the test on line 168.)

server/app/services/PhoneValidationUtils.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.i18n.phonenumbers.PhoneNumberUtil;
+import com.google.i18n.phonenumbers.Phonenumber;
+import java.util.Optional;
+
+/** Utility Class for all phone validations. */
+public final class PhoneValidationUtils {
+
+  private static final PhoneNumberUtil PHONE_NUMBER_UTIL = PhoneNumberUtil.getInstance();
+
+  public static Optional<MessageKey> validatePhoneNumber(
```
caitlinshk, line 13: Suggestion (blocking): Can you add some documentation specifying what this method returns and when? E.g. `@returns a MessageKey representing the validation error that occurred if the phone number is invalid, or an Optional.empty() if the phone number is valid`.

server/app/services/applicant/question/PhoneQuestion.java
```diff
@@ -20,9 +21,8 @@
 public final class PhoneQuestion extends Question {
 
   private Optional<String> phoneNumberValue;
-  private Optional<String> countryCodeValue;
-
   private static final PhoneNumberUtil PHONE_NUMBER_UTIL = PhoneNumberUtil.getInstance();
+  private Optional<String> countryCodeValue;
```
caitlinshk, line 25: Question (non-blocking): Does this field need to be moved?

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -143,21 +142,25 @@ private Form<EditTiClientInfoForm> validatePhoneNumber(Form<EditTiClientInfoForm
...
+          return form.withError(FORM_FIELD_NAME_PHONE, "A phone number must contain only digits");
+        case PHONE_VALIDATION_INVALID_LENGTH:
+          return form.withError(
+              FORM_FIELD_NAME_PHONE, "A phone number must contain only 10 digits");
+        case PHONE_VALIDATION_INVALID_PHONE_NUMBER:
+          return form.withError(FORM_FIELD_NAME_PHONE, "This phone number is not valid");
+        case PHONE_VALIDATION_NUMBER_NOT_IN_COUNTRY:
+          return form.withError(FORM_FIELD_NAME_PHONE, "This phone number is not valid for US");
+        default:
+          return form.withError(FORM_FIELD_NAME_PHONE, "An unknown error occurred " + error.get());
```
caitlinshk, line 162: Suggestion (blocking): I'm hoping we don't have to copy-paste the validation messages here. Could I suggest a slightly different format for this, where we just use the `MessageKey` provided?

```
Messages messages = messagesApi.preferred(request);
if (error.isPresent()) {
  return form.withError(FORM_FIELD_NAME_PHONE, messages.at(error.get().getKeyName()));
}
return form;
```

This code assumes 1) that we can inject `MessagesApi` into the constructor of this class; and 2) That we have a valid `Request` object that we can use for `MessagesApi.preferred(request)`. I think we can do 1 and 2, but let me know if that's not the case!

server/conf/i18n/messages
```diff
@@ -241,6 +241,7 @@ validation.phoneCountryCodeRequired=Country must be selected
 validation.phoneNumberMustContainNumbersOnly=Phone number cannot contain non-number characters
 validation.invalidPhoneNumberProvided=This phone number is invalid
 validation.phoneMustBeLocalToCountry=The phone you have provided does not belong to the country
+validation.phoneNumberMustContainOnlyTenDigits=The phone number must contain only ten digits
```
caitlinshk, line 244: Question (blocking): Because we're adding a new string, do we need to flag-guard this so we don't show untranslated strings to users?

server/test/services/applicant/question/PhoneQuestionTest.java
```diff
@@ -92,7 +92,7 @@ public void withInvalidApplicantData_missingRequiredFields() {
   }
 
   @Test
-  @Parameters({"5552123333", "123asc1234"})
+  @Parameters({"5552123333", "1231234567"})
```
caitlinshk, line 95: Question (blocking): Is looks like this *is* a valid phone number? Is this not a valid phone number for some reason?

caitlinshk (changes requested):

server/app/services/applicant/question/PhoneQuestion.java
swatkat1: no cannot be. It is the Class parameter. The PhoneUtil is needed to for the getAnswerString method

server/conf/i18n/messages
swatkat1: what is flag-guard?

server/test/services/applicant/question/PhoneQuestionTest.java
swatkat1: both are invalid. "The short answer is that most 555 numbers are not working numbers, so real people won't be harassed if diehard fans try to call them. "

server/app/services/ti/TrustedIntermediaryService.java
swatkat1: I have done one better. Passed the messages object to the service instead of request object. This reduces the scope of the overall change.

server/app/services/applicant/question/PhoneQuestion.java
caitlinshk: Oh sorry, I meant: It looks like this was moved from line 23 to line 25, can it be moved back to line 23?

server/test/services/applicant/question/PhoneQuestionTest.java
caitlinshk: Got it thanks!

server/app/services/ti/TrustedIntermediaryService.java
caitlinshk: Ooh yes even better than I imagined, great!

server/conf/i18n/messages
caitlinshk: I think Danielle answered this in slack but: I meant that there's some delay between adding the new string here and actually getting translations for it, so I'm concerned that during that in-between period, users may start seeing this error message not translated to their language.

I think the best approach from https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#internationalization-for-application-strings is to write a separate PR that adds the message, then submit this PR once all the message translations come through?

caitlinshk (approved):

server/conf/i18n/messages
caitlinshk: Chatted offline, decided to re-use an existing error message so we don't need to wait for translations.

PR #6443, opened 2024-01-23
Update styling on application bottom bar so it flows better

### LINKED ISSUE #6437: Button text not consistent on narrow screen
**Describe the bug**
In the applicant view, the review and previous buttons have different text alignment than skip and save and next. This only appears on a narrow screen, when 'Save and next' is on multiple lines

**To Reproduce**
On applicant view, make the screen narrow and view any question

**Expected behavior**
The text in all of the bottom buttons should be centered, or the buttons should flow onto multiple lines

**Screenshots**
7UjtUZ5EvAM8fTJ

greg-hyde:
### Description

Updates the alignment on the bottom bar so it flows better in narrow screen sizes. The two changes are:
1) replace the empty divs with justify-content: end as a method for right aligning the buttons (there was extra padding on mobile because of the empty divs, hence the updated browser screenshots)
2) add flex-wrap so that if the screen is narrow enough that all the content doesn't fit on one line but not so narrow that we make each button take up the entire width, we flow the buttons instead of making the buttons oddly tall with strange text alignment (see bug for example)

Desktop 
8YNBUKCncN9nVBr

Mildly narrow
54mPnuu3HHXF8G7

Narrow
7KVSAW3KSw7oycr

#### General

#### User visible changes

### Issue(s) this completes

Fixes #6437


caitlinshk (changes requested):
question (blocking): Is there an easy way to write a screenshot test for the example you have in the PR description for the "mildly narrow" case?

greg-hyde:
> question (blocking): Is there an easy way to write a screenshot test for the example you have in the PR description for the "mildly narrow" case?

So yes it wouldn't be too difficult to do screenshot tests at that size. I'd want to do it for all the applicant view though (take a large, medium, and small screenshot for each test). The hesitation that I have is that the "mildly narrow" kind of depends on what changes are being made. In this case, the mildly narrow case exists because 1) the UI switches to flowing vertically under size M (tailwind's definition of medium), and 2) the text of these 4 buttons happens to be wider than the dividing line between S and M screen sizes. Together those two conditions create the odd middle ground case. In a more general sense, for other UI elements, the dividing lines creating odd middle grounds could happen at other screen widths (for instance, some style kicks in at size L or S or something).

So moral of the story, I think it's totally reasonable, and just comes down to what size we think would catch most things (this particular case happens between 768 and 854 px wide, in english at least)

caitlinshk (approved):

PR #6447, opened 2024-01-23
[Program Card Images] Polish the admin UI

caitlinshk:
### Description

A whole host of UI polish to the program image upload UI based on feedback from Ron and Shreya. A list of changes:
- Move the alt-text form to be above the image upload form
- Minor update to the alt-text form label
- Disable the image upload form until alt-text has been provided (and add text explaining that to the user)
- Minor update to the file max size hint text
- Add additional hint text about image being cropped to 16x9 (based on feedback from Seattle)
- Make the file upload box take the whole column width (that's the `_uswds-theme-custom-styles.scss` change)
- Center the program card preview within its column

#### General

#### User visible changes

As an admin, navigate to the program image upload page and check out the cool new digs :)

### Issue(s) this completes

Part of epic #5676


caitlinshk:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/civiform/civiform/6447?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #6425** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6425?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #6447** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6447?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6447?utm_source=stack-comment).

server/app/views/admin/programs/ProgramImageView.java
```diff
@@ -199,7 +185,7 @@ private DivTag createImageDescriptionForm(
                     makeCsrfTokenInputTag(request),
                     FieldWithLabel.input()
                         .setFieldName(ProgramImageDescriptionForm.SUMMARY_IMAGE_DESCRIPTION)
-                        .setLabelText("Enter image description (Alt Text)")
+                        .setLabelText("Enter image description (Alt Text) (Required) *")
```
rockycodes, line 188: suggestion (non-blocking): Use required indicator https://github.com/civiform/civiform/blob/cce0552baec4746f118423a64f6c37c09c7f2c5d/server/app/views/ViewUtils.java#L235

rockycodes (changes requested):
Looks good! Just one question about `max-width: none` on the `uswds` class directly.

I also wanted to raise that with the current behavior it's possible to save the alt text without adding an image which results in something like this
image

I think there's enough feedback that folks probably won't do this, but I wanted to raise it in case you weren't aware and think it might be an issue.

caitlinshk:
> I also wanted to raise that with the current behavior it's possible to save the alt text without adding an image which results in something like this

Hmm, I don't think that should be happening? We only add an image to the view if the actual file key is present: https://github.com/civiform/civiform/blob/cce0552baec4746f118423a64f6c37c09c7f2c5d/server/app/views/ProgramImageUtils.java#L45-L47.

What could've happened that I've experienced before is that if I add an image to a program then kill & restart `bin/run-dev` later, the image will be gone from cloud storage because we kill and restart LocalStack which is what stores the images. So then if you look at the program with the new `bin/run-dev`, the image can't be found so the browser falls back to showing alt text. Can you try creating a new program, only filling in alt text, and see if it still repros?

rockycodes:
> Can you try creating a new program, only filling in alt text, and see if it still repros?

I just tried that and am still seeing the alt text show up in the preview and when publishing the program.

image

Is that not happening for you?

caitlinshk:
Weird, that behavior isn't happening to me... let me look further.

caitlinshk:
Paired offline, and we realized that you can click "Save image" even if you haven't uploaded an image and that will cause the program card to show just the alt text like the screenshot Rocky showed. https://github.com/civiform/civiform/pull/6465 will disable the button unless there's actually an image there, and we may also want to throw an error somewhere before saving to the database if the file key is empty.

server/app/assets/stylesheets/uswds/_uswds-theme-custom-styles.scss
```diff
@@ -24,3 +24,8 @@ See the radio-option below as an example.
 .cf-radio-option:focus-within {
   @include u-border('blue-40v');
 }
+
+/* Make the file input element for program image upload fill the area it's given. */
+.usa-file-input {
```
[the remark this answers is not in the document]
caitlinshk: Good question. We can't set it directly in `FileUploadViewStrategy` because of how the USWDS component works. In `FileUploadViewStrategy`, the only divs we create are one classed as `usa-form-group` and another classed as `usa-file-input`. But, USWDS actually creates a lot more items than just those two (see below). I tried setting an ID on the `<input>` element created in `FileUploadViewStrategy`, but that ID turns out to apply to the element with the `usa-file-input__input` class, but we actually need the `max-width` override on the element with the `usa-file-input` class.

I updated the style to be `#program-image-upload-file-input .usa-file-input` so that it only modifies file inputs within the program image upload div and it looks like that works! And shouldn't affect other file uploads.

```
<div class="usa-file-input usa-file-input--disabled">
  <div class="usa-file-input__target">
    <div class="usa-file-input__box"></div>
    <div class="usa-file-input__instructions" aria-hidden="true">
      <span class="usa-file-input__drag-text">Drag file here or</span> 
      <span class="usa-file-input__choose">choose from folder</span>
    </div>
    <input id="program-image-upload-file-input" type="file" name="file" class="usa-file-input__input" aria-describedby="file-input-hint-0 file-input-hint-1 " accept="image/jpeg,image/png" disabled="" aria-label="Drag file here or choose from folder">
  </div>
</div>
```

server/app/views/admin/programs/ProgramImageView.java
caitlinshk: Ooh nice, done.

rockycodes (approved):
lgtm!

PR #6496, opened 2024-01-29
Updating the format of the client search on the TI dashboard

### LINKED ISSUE #6214: Update search components to USWDS components
### Describe the solution:
Update the input fields for search and filters on the TI dashboard top increase usability.

### Figma mock: 
https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=816%3A494&mode=design&t=A8ppwAkGj6ALGDd0-1

### Screenshots:

Image

### Done when:
- The search inputs are laid out horizontally on desktop
- The search fields have a h4 search title
- The datepicker uses the [USWDS datepicker field](https://designsystem.digital.gov/components/date-picker/)
- The search fild used the [USWDS input field](https://designsystem.digital.gov/components/text-input/)

melanie-exygy:
### Description

Updated the client search on the TI dashboard to be aligned horizontally, have an h4 "Search" title, use USWDS input fields and use the USWDS Memorable Date component rather than the native date picker.

Many of the changes in this PR are related to the query parameters being changed from just name and dob to: name, day, month, and year.  The reason for these new search parameters is that the Memorable Date component has three fields rather than just one.

## Release notes

Updated the client search on the TI dashboard according to the new UX design and to use the USWDS Memorable Date component rather than the native date picker.

1. Go to the TI dashboard
2. Add a couple of clients if you don't have any yet
3. Search for clients either by name or by date of birth (if by dob, complete all 3 fields)
4. Observe that the corresponding clients appear in the table below.

image

### Issue(s) this completes

Fixes #6214


browser-test/src/trusted_intermediary.test.ts
```diff
@@ -392,7 +392,7 @@ describe('Trusted intermediaries', () => {
     }
     await tiDashboard.createClient(client3)
 
-    await tiDashboard.searchByDateOfBirth(client3.dobDate)
+    await tiDashboard.searchByDateOfBirth('10', '12', '2021')
```
swatkat1, line 395: Can a test case like date = 1956 - blank - blank be added too?

melanie-exygy: Good idea, yes, I'll add one.  Thanks!

melanie-exygy: Added.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -276,8 +271,12 @@ private ImmutableList<AccountModel> searchAccounts(
         .filter(
             account ->
                 ((account.newestApplicant().get().getApplicantData().getDateOfBirth().isPresent()
-                        && searchParameters.dateQuery().isPresent()
-                        && !searchParameters.dateQuery().get().isEmpty()
+                        && searchParameters.dayQuery().isPresent()
```
swatkat1, line 274: Non-blocking  - Would it be better to add this check in the SearchParam class iteself? eg. if day param is empty --> make is Optional.Empty else Optional.of(day)?

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -251,13 +251,8 @@ public Form<EditTiClientInfoForm> updateClientInfo(
   public TrustedIntermediarySearchResult getManagedAccounts(
```
swatkat1, line 251: Please update the Java docs of the method to include new changes like- "" .... If both
   are empty or incomplete dates are passed - an unfiltered list of accounts is returned..."

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -296,4 +299,27 @@ private ImmutableList<AccountModel> searchAccounts(
                                 searchParameters.nameQuery().get().toLowerCase(Locale.ROOT)))))
         .collect(ImmutableList.toImmutableList());
   }
+
+  private boolean checkIfEmptySearchParams(SearchParameters searchParameters) {
```
swatkat1, line 303: I think this handles all the cases for returning a full list including partial dates. If yes, why should we check again during Account filtering?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -100,34 +105,36 @@ public Content render(
 
   private FormTag renderSearchForm(Http.Request request, SearchParameters searchParameters) {
     return form()
-        .withClass("w-1/4")
+        .withId("ti-search-form")
```
swatkat1, line 108: Is there way to add a flash message or let the user know- "we can only search on complete dates" I think the TI will think, they can search by year for easier use.

melanie-exygy: That's a good point.  Let me ask Sijia what they would recommend.

server/app/services/ti/TrustedIntermediaryService.java
melanie-exygy: `checkIfEmptySearchParams` checks if any single one of the date fields is empty/null whereas in `searchAccounts` it checks all the date fields to make sure none of them are empty/null.  I could remove the empty string checks (not the null, though) from `searchAccounts` but, in that case, if any date field had an empty string, the user would see the error message "Please enter date in MM/dd/yyyy format", similar to them making any other formatting mistake.  

Come to think of it, that might be a good way to address the missing date field error message that you mention below.  I'd need to change the error message wording, but I think I should do that anyway.  I also really don't like that the current error message shows up on a blank page like this:
image

I'll ask Sijia about that too.

swatkat1: I agree!

server/app/services/ti/TrustedIntermediaryService.java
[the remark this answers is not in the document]
melanie-exygy: Good idea.  I added a couple of helper functions to make it cleaner.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
melanie-exygy: Upon Sijia's recommendation, I added an error message when the dob is incomplete and the name field is empty.  The missing fields also have a red outline so that the user knows which ones are missing:
image

server/app/services/ti/TrustedIntermediaryService.java
melanie-exygy: Done

server/app/services/ti/TrustedIntermediaryService.java
melanie-exygy: I consolidated the empty/blank/null checks into this method: `TrustedIntermediaryService.findMissingSearchParams` so that I am only doing those checks once.  It returns a list of missing parameters which I can then use to know specifically which parameters are missing (Name, Day, Month, or Year).

server/app/services/ti/TrustedIntermediaryService.java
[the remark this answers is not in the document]
melanie-exygy: The helper function is `TrustedIntermediaryService.validateAndConvertSearchParamDOB`

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -238,11 +240,13 @@ public Form<EditTiClientInfoForm> updateClientInfo(
...
-   *     between the Account holder's name and the nameQuery is performed. If dateQuery is present -
-   *     a match between the Account holder's Date of Birth and the dateQuery is performed - the
-   *     matched results are collected and sent as an Immutable List.
+   * @param searchParameters - This object contains a nameQuery, a dayQuery, a monthQuery and a
+   *     yearQuery String. If all are empty, an unfiltered list of accounts is returned. If
+   *     nameQuery is present, a match between the Account holder's name and the nameQuery is
+   *     performed. If dayQuery, monthQuery and yearQuery are present, a match between the Account
+   *     holder's Date of Birth and the date queries is performed. The matched results are collected
+   *     and sent as an Immutable List. If name query is empty and any of the date queries are
+   *     empty, an empty list is returned.
```
swatkat1, line 249: Is an empty list returned or an unfiltered list?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: beautiful!!!

server/app/services/ti/TrustedIntermediaryService.java
melanie-exygy: An empty list of clients is returned if the name query is empty and only some of the date queries are empty.  See this test:  https://github.com/civiform/civiform/blob/a8b31f19e072de158bc2df1c5d4f9131def20b76/server/test/services/ti/TrustedIntermediaryServiceTest.java#L375

I'll change the wording of the Javadoc so that it's more clear.

melanie-exygy: Javadoc updated

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -251,49 +255,103 @@ public Form<EditTiClientInfoForm> updateClientInfo(
...
+                    || (!missingParams.contains(SearchParameters.ParamTypes.NAME)
                         && account
                             .getApplicantName()
                             .toLowerCase(Locale.ROOT)
                             .contains(
                                 searchParameters.nameQuery().get().toLowerCase(Locale.ROOT)))))
         .collect(ImmutableList.toImmutableList());
   }
+
+  private static List<SearchParameters.ParamTypes> findMissingSearchParams(
```
swatkat1, line 298: Please revert static.

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -251,49 +255,103 @@ public Form<EditTiClientInfoForm> updateClientInfo(
...
+      missing.add(SearchParameters.ParamTypes.MONTH);
+    }
+    if (filteredYearQuery.isEmpty()) {
+      missing.add(SearchParameters.ParamTypes.YEAR);
+    }
+
+    return missing;
+  }
+
+  public static boolean validateSearch(SearchParameters searchParameters) {
```
swatkat1, line 329: Please revert static. Also any reason why is it public?

melanie-exygy: This is `public static` so that it can be called as a utility function from the `TrustedIntermediaryDashboardView.java`.

server/app/services/ti/TrustedIntermediaryService.java
melanie-exygy: This is static because this function is called from `validateSearch`, which is static for the reason described below.

swatkat1 (approved):
This will be huge to the TIs!

PR #6532, opened 2024-01-31
Localize TI dashboard

### LINKED ISSUE #5535: Localize the TI dashboard
It would be nice to make the TI dashboard more accessible to non-English speakers.

We could offer translations for the text on the TI dashboard.

**Done when**
- All text on the TI Dashboard is translated into the languages that we support.

azhangggggg:
### Description

Added messages to localized TI dashboard via new [design](https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=1-19&mode=design&t=XLOFF9Oa0csHcEZO-0).

#### General

#### i18n
- [x] Add the default string to [messages](https://github.com/civiform/civiform/blob/main/server/conf/i18n/messages) and include a comment with extra context for the translators above it (like [this](https://github.com/civiform/civiform/blob/ee16d08c04ab0e8b0c5bbaf9432718d5be4b164c/server/conf/i18n/messages#L34)).
- [x] Do not touch the other messages.xx files, such as messages.tl. Transifex will update these files automatically when translations are available. For now, all foreign languages will fall back to English strings.
- [x] Add the key from the messages file to [MessageKey.java](https://github.com/civiform/civiform/blob/main/server/app/services/MessageKey.java)

### Issue(s) this completes

Fixes #5535


server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+
+title.acctSettings=Account settings
+title.allClients=All Clients
+title.clientList=Client list
+title.createClient=Create a new client
+title.displayingAllClients=Displaying all clients
+title.orgMembers=Organization members
+
+toast.asterisk=Required fields are marked with an asterisk
+toast.dobExample=For example: January 18,1986
```
melanie-exygy, line 141: suggestion (blocking):  Since this is part of the label, I suggest calling this variable `label.dobExample`, rather than `toast`.  Also, note that the date is January 28 rather than January 18.

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+
```
melanie-exygy, line 142: suggestion (blocking):  With the tweaks of the design this morning, we'll need to add this message too:  `label.nameExample=For example:  Gu or Darren or Darren Gu`.

We're also missing, as far as I can see, "Month", "Day", "Year", "Select", "Clear search", an "Edit" link, "Name", "Email", "Account Status", "Phone number", "New client successfully created.", "Your new client {0} has been added to your client list. You can now begin submitting applications for this client.", "Start an application", "Edit client info", "This account cannot be deleted since it has open applications."

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+label.nameOrEmail=Name(s) or email
+label.notes=Notes
+label.contactInfo=Contact Information
+
+link.aboutCiviform=About CiviForm
+link.accessibilityStatement=Accessibility statement
+link.backToClientList=Back to client list
+link.privacyPolicy=Privacy Policy
+
+title.acctSettings=Account settings
```
melanie-exygy, line 133: suggestion (non-blocking):  These aren't title elements, so I'd suggest marking them as `header` or `link` in the case of "Client list" and "Account settings".

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+title.allClients=All Clients
+title.clientList=Client list
+title.createClient=Create a new client
+title.displayingAllClients=Displaying all clients
```
melanie-exygy, line 137: suggestion (blocking):  This one needs a parameter since it can say "Displaying 2 clients" when the list is filtered.  We'll also need a singular version for when there's one client.  That one can be "Displaying 1 client".

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+title.orgMembers=Organization members
+
+toast.asterisk=Required fields are marked with an asterisk
```
melanie-exygy, line 140: suggestion (blocking):  We don't need this one because it is already in the messages file since it's used in other places.

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+button.keepAcct=Keep account
+button.nextPage=Next
+button.save=Save
+button.search=Search
+button.viewApplications=View applications
+
+content.numberOfAppSubmitted={0} applications submitted
+content.warning=There are no applications submitted for this client, so no applications will be lost.
+
+dialog.deleteConfirmation=Are you sure you want to delete this client's account?
```
melanie-exygy, line 121: suggestion (blocking):  For some reason (maybe an escape character?), you need to put two apostrophes here.  "...to delete this client''s account?".  See other examples in the messages file.

server/conf/i18n/messages
```diff
@@ -97,6 +97,49 @@ title.submittedPrograms=Submitted
...
+
+label.dob=Date of birth
+label.nameOrEmail=Name(s) or email
+label.notes=Notes
+label.contactInfo=Contact Information
+
+link.aboutCiviform=About CiviForm
```
melanie-exygy, line 128: suggestion (non-blocking):  I don't think the page footer is part of the TI dashboard work.  It should be a separate effort to create the footer and internationalize it.  So, you probably don't need to include "About CiviForm", "Accessibility statement" and "Privacy Policy".

server/conf/i18n/messages
azhangggggg: done.

server/conf/i18n/messages
azhangggggg: done.

server/conf/i18n/messages
azhangggggg: done.

server/conf/i18n/messages
azhangggggg: done.

server/conf/i18n/messages
```diff
@@ -97,49 +97,61 @@ title.submittedPrograms=Submitted
...
+header.clientList=Client list
+
+label.acctStatus=Account Status
 label.contactInfo=Contact Information
+label.day=Day
+label.dobExample=For example: January 28,1986
+label.email=Email
+label.month=Month
+label.name=Name
+label.nameExample=ex. Darren Gu, Gu or darren.gu@gmail.com
```
melanie-exygy, line 140: suggestion (blocking):  This one was updated today too.  It should be "For example: Gu or Darren or Darren Gu".

server/conf/i18n/messages
```diff
@@ -97,49 +97,61 @@ title.submittedPrograms=Submitted
...
 button.viewApplications=View applications
 
+content.clientCreated=You new client {0} has been added to your client list. You can now begin submitting applications for this client.
 content.numberOfAppSubmitted={0} applications submitted
 content.warning=There are no applications submitted for this client, so no applications will be lost.
 
-dialog.deleteConfirmation=Are you sure you want to delete this client's account?
+dialog.deleteConfirmation=Are you sure you want to delete this client''s account?
 
-label.dob=Date of birth
```
melanie-exygy: suggestion (blocking):  We still need "Date of birth" for the label in the Add Client form.

azhangggggg: done.

server/conf/i18n/messages
azhangggggg: ah i see, it was literally `ex. Darren Gu, Gu or darren.gu@gmail.com` just one hour ago. updated.

melanie-exygy (approved):

server/conf/i18n/messages
```diff
@@ -97,6 +97,62 @@ title.submittedPrograms=Submitted
...
+button.deleteAcct=Delete account
+button.keepAcct=Keep account
+button.nextPage=Next
+button.save=Save
+button.search=Search
+button.select=Select
+button.startApp=Start an application
+button.viewApplications=View applications
+
+content.clientCreated=You new client {0} has been added to your client list. You can now begin submitting applications for this client.
```
dkatzz, line 124: **question:** I don't have access to the figma files, but should this be "Your new client"?

melanie-exygy: Oh, good catch!  Yes, it should be.  I'll push the change, although looks like I may have just missed the cutoff for the translations batch.  We probably won't need this right away, though.

PR #6546, opened 2024-02-01
Fix for Application count mismatch between CVS exports and Reporting

### LINKED ISSUE #5824: Reporting submissions don't match with CSV export
**Describe the bug**
As a Program Admin, I use the Reporting Dashboard to see how many applications are submitted each month. The total number of submissions doesn't match with the CSV export of the same data.

For example:
HSD's Gold Card for the month of 9/2023

- 76 submissions per the reporting table
- 162 submissions per the CSV export

**To Reproduce**

1. Go to the Reporting Dashboard
2. Click on a program
3. See the number of submissions by month
4. Compare that number to the CSV export 
5. See that numbers aren't aligned

**Expected behavior**
I would expect the numbers to be the same

**Screenshots**
Below and CSV attached

image
[Gold Card (HSD)_Example.csv](https://github.com/civiform/civiform/files/13046696/Gold.Card.HSD._Example.csv)

swatkat1:
### Description

Fix for Application count mismatch between CVS exports and Reporting.
Fix 1- Internally we store all submit time as UTC. On trying to the retrieve the records for dd-mm-yyyy, we should also covert it to UTC time zone so that we will get the same results.
Fix 2 - When converting java.sql.Timestamp into MM/yyyy format for reporting, there is no need to add time zone information as the time stamp is always set to the beginning of the month eg. 01-01-2024 00:00:000. So SimpleDateTimeFormatter is enough to extract the MM/yyyy format.  

## Release notes

Fix for Application count mismatch between CVS exports and Reporting.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #5824

### DIFF
--- a/server/app/modules/DurableJobModule.java
+++ b/server/app/modules/DurableJobModule.java
@@ -87,7 +87,7 @@ public final class DurableJobModule extends AbstractModule {
         DurableJobName.REPORTING_DASHBOARD_MONTHLY_REFRESH,
         persistedDurableJob ->
             new ReportingDashboardMonthlyRefreshJob(reportingRepository, persistedDurableJob),
-        new RecurringJobExecutionTimeResolvers.Sunday2Am());
+        new RecurringJobExecutionTimeResolvers.FirstOfMonth2Am());
 
     durableJobRegistry.register(
         DurableJobName.UNUSED_ACCOUNT_CLEANUP,
--- a/server/app/modules/DurableJobModule.java
+++ b/server/app/modules/DurableJobModule.java
@@ -87,7 +87,7 @@ public final class DurableJobModule extends AbstractModule {
         DurableJobName.REPORTING_DASHBOARD_MONTHLY_REFRESH,
         persistedDurableJob ->
             new ReportingDashboardMonthlyRefreshJob(reportingRepository, persistedDurableJob),
-        new RecurringJobExecutionTimeResolvers.FirstOfMonth2Am());
+        new RecurringJobExecutionTimeResolvers.Sunday2Am());
 
     durableJobRegistry.register(
         DurableJobName.UNUSED_ACCOUNT_CLEANUP,
--- a/server/app/services/DateConverter.java
+++ b/server/app/services/DateConverter.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.inject.Inject;
 import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.LocalDate;
@@ -11,8 +12,6 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.util.Calendar;
-import java.util.TimeZone;
 
 /** Utility class for converting dates between different formats. */
 public final class DateConverter {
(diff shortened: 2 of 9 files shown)


server/app/modules/DurableJobModule.java
```diff
@@ -87,7 +87,7 @@ public DurableJobRegistry provideDurableJobRegistry(
         DurableJobName.REPORTING_DASHBOARD_MONTHLY_REFRESH,
         persistedDurableJob ->
             new ReportingDashboardMonthlyRefreshJob(reportingRepository, persistedDurableJob),
-        new RecurringJobExecutionTimeResolvers.FirstOfMonth2Am());
+        new RecurringJobExecutionTimeResolvers.Sunday2Am());
```
gwendolyngoetz, line 90: (blocking) If we're changing the refresh rate the job name needs to be renamed otherwise it's confusing when you see it happening weekly. Change it from `REPORTING_DASHBOARD_MONTHLY_REFRESH` to `REPORTING_DASHBOARD_REFRESH` then it won't matter what we make the refresh rate. Change the enum value and the class name please.

server/app/services/DateConverter.java
```diff
@@ -59,12 +59,12 @@ public LocalDate getCurrentDateForZoneId() {
...
   /**
    * Parses a string containing a ISO-8601 date (i.e. "YYYY-MM-DD") and converts it to an {@link
-   * Instant} at the beginning of the day in local time zone.
+   * Instant} at the beginning of the day in UTC time zone.
    *
    * @throws DateTimeParseException if dateString is not well-formed.
    */
   public Instant parseIso8601DateToStartOfDateInstant(String dateString) {
-    return parseIso8601DateToLocalDate(dateString).atStartOfDay(zoneId).toInstant();
+    return parseIso8601DateToLocalDate(dateString).atStartOfDay(ZoneId.of("UTC")).toInstant();
```
gwendolyngoetz, line 67: (blocking) I'm concerned about the ramifications this may have with other things that are calling it already expecting a local time result. Can you dig into that and document what's using it now and if this change will negatively affect it?

gwendolyngoetz (changes requested):

server/app/services/DateConverter.java
swatkat1: Hi @gwendolyngoetz, I see this getting used in three places -1. ApiKeyService.java--> this method is used to set the expiration of the API Key. 2. ProgramApplicationsApiController.java - to determine, fromDate and toDate to find the applications. 3. AdminApplicationController.java --> Where the applications filters are converted to instant to filter the result. Given this small scope, I don't not see much issue in changing this. Let me know your thoughts. Link- https://github.com/search?q=repo%3Aciviform%2Fciviform%20parseIso8601DateToStartOfDateInstant&type=code

avaleske: That's a good callout.

Currently the behavior is that a day starts on midnight local time, and I don't think we should change that. If someone has a job that expects to get "yesterday's applications" that they're running at 1am PST, changing this to UTC will cause that job to fetch a different set of applications (from 4pm two days before until 4pm yesterday)

Similarly, having the admin filter be UTC could be confusing because an application filled out at 9pm PST wouldn't show up in the filter for yesterday, it'd show up in the filter for today.

avaleske: Instead, could we update the reporting to use local time?

gwendolyngoetz: That what I was thinking. Thanks Austin.

@swatkat1 I'm going to suggest that you rename the existing method `parseIso8601DateToStartOfDateInstant` to `parseIso8601DateToStartOfLocalDateInstant` and keep the code as it was originally. Then create a new called `parseIso8601DateToStartOfUtcDateInstant` that does what your UTC change do for your use case.

It's a tiny amount of duplication, but more explicit than having the caller have to pass in the zoneId every time.

gwendolyngoetz: I'm just here so github will let me review changes.

gwendolyngoetz (changes requested):

server/app/services/DateConverter.java
swatkat1: Thank you @gwendolyngoetz and @avaleske . Updated the PR.

server/test/services/DateConverterTest.java
```diff
@@ -18,7 +19,7 @@ public class DateConverterTest {
   @Test
   public void workingWith_Iso8601_dates() {
```
gwendolyngoetz, line 20:
```suggestion
  public void parseIso8601DateToLocalDateInstant_isSuccessful() {
```

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

Issue #6398, opened 2024-01-17
Add JS for dynamically showing/hiding question edit UI components


nb1701:
#### Describe the solution presented in this issue.
Per the TDD:

- [ ] When a question has a PAI toggle set, but then the user unsets the universal toggle, programatically set the PAI toggle off, hide the toggle, then show the alert about universal not being set.
- [ ] When a user toggles universal on, show the PAI toggle and hide the alert
- [ ] When creating a new question, when an enumerator is chosen for the question, hide the PAI section entirely
- [ ] When creating a new question, when deselecting an enumerator for the question, show the PAI section again

[Spike commit](https://github.com/civiform/civiform/pull/6202/commits/df935dba4f802b614b23fde4276652acd6d44380)

Eventually, we want to rewrite this behavior in HTMX, but will start with this bit of JS code to get v1 of this feature out the door.

#### Additional Context
[TDD](https://docs.google.com/document/d/13xTLRl82gDD0VajM68gVdOGUV6alXi_mtluf1uUT1Lo/edit)

#### Done when:
What criteria need to be met for the issue to be finished? 
- [ ] When flipping the universal toggle on and off, the PAI toggle/alert is shown/hidden appropriately
- [ ] When selecting/deselecting an enumerator on question creation, the PAI section is shown/hidden appropriately
- [ ] When editing a universal question that has a PAI tag set, when unsetting universal and saving the question, the PAI tags are removed as well
- [ ] Tests added for these changes

Issue #6397, opened 2024-01-17
Modify question create/edit UI to add toggles for PAI tags


nb1701:
#### Describe the solution presented in this issue.
Per the TDD and mocks, add the toggles to the question create/edit UI to be able to set each of the defined tags on their appropriate question types.

- [x] Modify `ViewUtils#makeToggleButton` and `ViewUtils#makeAlertInfoSlim` to allow these components to be hidden ([Spike commit](https://github.com/civiform/civiform/pull/6202/commits/410d2549886d298fc4428d29fe86e25543309552)
- [ ] Modify `QuestionEditView` to build the Primary Applicant Info section, setting the state of the components being hidden as appropriate (don't worry about dynamically showing and hiding things based on the state of the universal toggle for now) ([Spike commit](https://github.com/civiform/civiform/pull/6202/commits/ae36e27291b3f523ae123235312606b870903ecb)
  - If a different question already has the tag set, show the alert noting this and hide the toggle
  - If universal is not set, show the alert noting this and hide the toggle
  - Use the field name in the `PrimaryApplicantInfoTag` so that `QuestionForm` can bind to it automatically
  - If the question is being edited and has an enumerator set, do not show the whole PAI section

#### Additional Context
[TDD](https://docs.google.com/document/d/13xTLRl82gDD0VajM68gVdOGUV6alXi_mtluf1uUT1Lo/edit)

#### Done when:
What criteria need to be met for the issue to be finished? 
- [ ] Can set the name PAI tag on a universal name type question, and this being set/unset persists
- [ ] Can set the DOB PAI tag on a universal date type question, and this being set/unset persists
- [ ] Can set the email PAI tag on a universal email type question, and this being set/unset persists
- [ ] Can set the phone number PAI tag on a universal phone type question, and this being set/unset persists
- [ ] Toggle is hidden when editing an existing question if it is not marked universal (don't worry about changing state on toggle before submit) and alert is shown instead
- [ ] Toggle is hidden when creating or editing a question if a different question already has the tag set
- [ ] PAI section (toggle + text) is not shown when editing a question with an enumerator

PR #6549, opened 2024-02-02
Add Primary Applicant Info toggles to the question edit view

nb1701:
### Description

This adds the Primary Applicant Information toggles to the question edit view.  If the universal toggle is not set, it shows an alert saying the PAI tag can't be set because it is not a universal question.  If the universal toggle is set, but the tag is already set on another question, it shows an alert saying so.  Upon question creation, when an enumerator is set, the PAI section is hidden.  All of this is done dynamically via a bit of JS code.

This also changes the name of the feature flag, since it was created before we finalized the name of this feature.  I promise it won't change again :)

#### General

#### User visible changes

#### New Features

1. Enable the PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED feature flag.
2. Ensure the alerts and toggles are shown and hidden according to the description.
3. Ensure the state of the toggles persist after saving and publishing.
4. Optionally, check the questions table in the database to verify the appropriate QuestionTag was added to the list for the question.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/6397
https://github.com/civiform/civiform/issues/6398


server/app/views/admin/questions/QuestionEditView.java
```diff
@@ -457,10 +468,94 @@ private DomContent buildUniversalQuestion(QuestionForm questionForm) {
...
+              DivTag tagSubsection = div().withClass("cf-primary-applicant-info-subsection");
+
+              if (differentQuestionHasTag) {
+                // Only include the alert. We'll swap the text based on the state of the
+                // universal toggle in admin_question_edit.ts.
+                String otherQuestionName =
+                    currentQuestionForTag.map(QuestionDefinition::getName).orElseThrow();
+                String alreadySetAlertText =
+                    String.format(
+                        "You cannot edit this setting since this property is already set on a"
```
avaleske, line 502: *very nit (phrasing)*: Here and on line 507 we use "since", but on line 508 we use "because". 

I think it'd read slightly cleaner if we used "because" everywhere, (or at least used the same conjunction for both).

Also "this property" refers to the PAI toggle, right? imo naming it is helpful, so "You cannot make this question a Primary Applicant Info question because the question named "%s" is already set as a Primary Applicant Info question." would be more clear.

server/app/assets/javascripts/admin_question_edit.ts
```diff
@@ -8,6 +8,78 @@ class AdminQuestionEdit {
...
+      'input',
+      (event: Event) => {
+        const target = event.target as HTMLInputElement
+        const setHidden = target.value !== ''
+        primaryApplicantInfoSection.toggleAttribute('hidden', setHidden)
+      },
+    )
+  }
+
+  addUniversalToggleHandler(primaryApplicantInfoSection: HTMLElement) {
```
avaleske, line 33: *Question (blocking)*: Do screenreaders do the right thing when we unhide elements or do we need to wrap all this in an `aria-live`?

server/app/views/admin/questions/QuestionEditView.java
```diff
@@ -457,10 +468,94 @@ private DomContent buildUniversalQuestion(QuestionForm questionForm) {
...
+                    tagSubsection
+                        .with(
+                            ViewUtils.makeAlertInfoSlim(
+                                initialAlertText,
+                                /* hidden= */ false,
+                                /* classes...= */ "cf-primary-applicant-info-alert",
+                                "usa-alert-primary-applicant-info"))
+                        .withData("already-set-alert", alreadySetAlertText)
+                        .withData(
+                            "non-universal-already-set-alert", nonUniversalAlreadySetAlertText);
```
avaleske, line 525: *suggestion (blocking but I could be convinced otherwise)*: I think this view code, and the JS that manipulates the DOM it renders, would be simpler if we put everything we might need on the page, but hidden.

Instead of stashing the error messages we might need in a `data` attribute and pulling them out in the JS, I think we should include a few versions of the alert in the DOM and unhide the one that applies.

That simplifies this class because then we can structure it like
```java
DivTag alertNotUniversalSection = ...
DivTag alertNotUniversalAndAlreadySetSection = ...
DivTag alertAlreadySetSection = ...

tagSubsection
  .condWith(alertNotUniversal, alertNotUniversalSection)
  .condWith(alertNotUniversalAndAlreadySet, alertNotUniversalAndAlreadySetSection)
  .condWith(alertAlreadySet, alertAlreadySetSection)
  .condWith(includeToggle, toggleButton);
```

And if we're ok having two alert boxes, we don't need the middle permutation.

server/app/views/admin/questions/QuestionEditView.java
```diff
@@ -457,10 +468,94 @@ private DomContent buildUniversalQuestion(QuestionForm questionForm) {
...
+                                        /* text= */ Optional.of(
+                                            primaryApplicantInfoTag.getDisplayName()))))
+                        .with(
+                            ViewUtils.makeAlertInfoSlim(
+                                "You cannot edit this setting since the question is not a universal"
+                                    + " question.",
+                                /* hidden= */ questionForm.isUniversal(),
+                                /* classes...= */ "cf-primary-applicant-info-alert",
+                                // This removes the top 1rem margin built in to the USWDS component
+                                "usa-alert-primary-applicant-info"));
```
avaleske, line 552: *thought, non-blocking*: what naming convention do we want to introduce for classes that modify USWDS styles? Do we want to go in a tailwind-ish direction and name it something like `clear-usa-alert-top-margin`?

browser-test/src/support/admin_questions.ts
```diff
@@ -79,7 +94,7 @@ export class AdminQuestions {
   }
 
   async clickSubmitButtonAndNavigate(buttonText: string) {
-    await this.page.click(`button:has-text("${buttonText}")`)
+    await this.page.click(`button:text-is("${buttonText}")`)
```
avaleske, line 97: *praise*: I didn't know about `text-is`, it's def better

browser-test/src/question_lifecycle.test.ts
```diff
@@ -524,4 +528,218 @@ describe('normal question lifecycle', () => {
       'my_test_question',
     )
   })
+
```
avaleske, line 531: *suggestion (blocking)*: imo we should move these into a new primary_applicant_info_test file. a) because we'll have a bunch of other PAI related tests with Rocky's work and b) this file is getting too big and c) even though some of these tests are about ensuring that creating a draft, publishing, creating another draft, etc does the right thing, that's not the core goal of these tests.

avaleske (changes requested):
Large thoughts
- I left some comments about how to structure the JS and View code in a way that might be simpler
- Are we already doing the server-side validation of the updated question?

Small thoughts:
- I had a few naming suggestions
- I had a few error message phrasing suggestions.

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -211,7 +211,8 @@ private Pair<DivTag, ImmutableList<Modal>> renderAllQuestionRows(
               .with(
                   ViewUtils.makeAlertInfoSlim(
                       "We recommend using all universal questions in your program for personal and"
-                          + " contact information questions."))
+                          + " contact information questions.",
```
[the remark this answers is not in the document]
sijiayam: Looks good to me

server/app/services/question/PrimaryApplicantInfoTag.java
```diff
@@ -16,31 +16,30 @@ public enum PrimaryApplicantInfoTag {
...
+      "If this is turned on, the name collected from this question will be used to identify an"
+          + " applicant and search for their application."),
   APPLICANT_PHONE(
       QuestionTag.PRIMARY_APPLICANT_PHONE,
       QuestionType.PHONE,
       "primaryApplicantPhone",
-      "Applicant Phone Number",
-      "Setting this property will make the application searchable by phone number.");
+      "Set answer as applicant phone number",
+      "If this is turned on, admins can search for an application by this collected phone number.");
```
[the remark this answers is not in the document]
sijiayam: @nb1701 We can move the word collected

server/app/views/ViewUtils.java
```diff
@@ -363,24 +371,31 @@ public static DivTag makeUniversalBadge(
...
+                    h4().withClass("usa-alert__heading").withText(title.orElse("")))
                 .with(p().withClass("usa-alert__text").withText(text)));
   }
 
-  public static DivTag makeAlertInfoSlim(String text) {
-    return makeAlert(text, Optional.empty(), BaseStyles.ALERT_INFO, BaseStyles.ALERT_SLIM);
+  public static DivTag makeAlertInfoSlim(String text, boolean hidden, String... classes) {
+    return makeAlert(
+        text,
+        hidden,
```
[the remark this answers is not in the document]
avaleske: Yeah I think as it is is fine!

server/app/views/admin/questions/QuestionEditView.java
sijiayam: These suggestions look great @avaleske . Lets move ahead with them. @nb1701

server/app/views/ViewUtils.java
```diff
@@ -363,24 +371,31 @@ public static DivTag makeUniversalBadge(
...
     return div()
+        .withCondHidden(hidden)
         .withClasses("usa-alert", String.join(" ", classes))
         .with(
             div()
                 .withClasses("usa-alert__body")
                 .condWith(
-                    maybeTitle.isPresent(),
-                    h4().withClass("usa-alert__heading").withText(maybeTitle.orElse("")))
+                    title.isPresent(),
```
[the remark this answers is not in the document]
avaleske: lol I had added it too 😆

server/app/views/admin/questions/QuestionEditView.java
avaleske: I found that easier to reason about! I get the page-weight concern, but I think it'll mostly be gzipped out, and I don't have concerns about hiding stuff in the dom until we need it. (that's what we do with modals, after all)

avaleske: Is the "may be null" comment because the element might not be on the page if we're in the "it's already set on another question" condition?

avaleske (approved):
Thanks for these changes!

I left a few non-blocking grammar comments but addressing them might be out of scope for this since they're about general style across the app

PR #6558, opened 2024-02-05
Fixing Tailwind and Updating to Latest Version

gwendolyngoetz:
#### Fixing Tailwind Generation
Adding --watch=always and setting tailwind to first version with that flag. The `--watch=always` added at [this version of tailwind](https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.2.5).

This update bring us to the latest release of Tailwind, v3.4.1.

#### New Order in Generated CSS File
Tailwind is now generating the css in a bit of a different order. Nothing we can do about it; they do not guarantee order. Typically this isn't a problem because everything has the same specificity unless a modifier is added. However we haven't been very careful in how we apply classes. 

It looks to be common to have multiple classes on an element modifying the same css property. The main problem child at this time is with the `background-color`. With things being reordered it changes what is actually applied since the spot lower in the css file supersedes settings defined higher up.

##### The New Order, Originally was Opposite
```css
.bg-civiform-white {
  --tw-bg-opacity: 1;
  background-color: rgb(248 249 250 / var(--tw-bg-opacity));
}

.bg-transparent {
  background-color: transparent;
}
```

#### 
The additional issue is that the `.bg-transparent` class is in a different order in the generated css file. The button has these two classes on it. Before they were in this order, but now they've swapped location.

#### TI Dashboard
Another issue I see is that `BaseHtmLView#renderHeader` has `mb-4` on it.  The TI dashboard adds `mb-0` to reduce the gap between the `h1` and `h2`. This no longer works and the gap stays. Because `mb-4` is added via j2html `withClasses`, but you can't remove a class that was already added. Because the header is used everywhere I'm not looking to change it. This only affects the TI Dashboard and it's not really the end of the world, it just looks a little sad because of the excess whitespace.

#### Admin Question Edit Readonly Inputs
Setting readonly admin questions to use the `read-only` tailwind modifier to get readonly input colors working again.

#### Admin Question Edit Validation Border
The new generated css must have the validation error red border in a different spot as on the admin question edit screen validation error input border is now showing all by itself.

#### General

### Issue(s) this completes

Related https://github.com/civiform/civiform/pull/6353


server/app/views/components/FieldWithLabel.java
```diff
@@ -583,8 +583,8 @@ private <T extends Tag<T> & IName<T> & IDisabled<T>> void generalApplyAttrsClass
         .withClasses(
             getFieldClasses(fieldTag, hasFieldErrors),
             // TODO(#5623): Use unified styles for disabled inputs
-            this.readOnly ? "text-gray-500" : "",
-            this.readOnly ? "bg-gray-100" : "")
+            this.readOnly ? "read-only:text-gray-500" : "",
+            this.readOnly ? "read-only:bg-gray-100" : "")
```
gwendolyngoetz, line 587: Why: Conflict between `bg-white` and `bg-gray-100`. The `read-only` increases specificity when `input readonly="readonly"`. Including `text-gray-500` for to prevent the same issue down the road.

server/app/views/style/BaseStyles.java
```diff
@@ -185,11 +185,7 @@ public final class BaseStyles {
 
   public static final String ADMIN_LOGIN =
       StyleUtils.joinStyles(
-          "bg-transparent",
-          "text-black",
-          "underline",
-          "font-bold",
-          StyleUtils.hover("bg-gray-200", "opacity-90"));
+          "text-black", "underline", "font-bold", StyleUtils.hover("bg-gray-200", "opacity-90"));
```
gwendolyngoetz, line 188: Why: The new ordering in the css file causes `bg-transparent` to load later now upsetting the visuals.

server/app/views/components/ButtonStyles.java
```diff
@@ -64,12 +63,12 @@ public final class ButtonStyles {
       StyleUtils.joinStyles(BUTTON_BASE_SOLID_BLUE_SEMIBOLD, "text-xl");
 
   public static final String OUTLINED_TRANSPARENT =
-      StyleUtils.joinStyles(BUTTON_BASE_OUTLINE_SEMIBOLD, "text-base");
+      StyleUtils.joinStyles(BUTTON_BASE_OUTLINE_SEMIBOLD, "text-base", "bg-transparent");
```
gwendolyngoetz, line 66: Why: The new ordering in the css file causes `bg-transparent` to load later now upsetting the visuals.

server/app/views/admin/programs/ProgramBlocksView.java
```diff
@@ -931,13 +931,7 @@ private DivTag renderReadOnlyLabel(String label) {
...
         div()
             .withClasses(
-                "flex",
-                "gap-2",
-                "items-center",
-                "text-gray-400",
-                "font-medium",
-                "bg-transparent",
-                "rounded-full")
+                "flex", "gap-2", "items-center", "text-gray-400", "font-medium", "rounded-full")
```
gwendolyngoetz, line 934: Why: The new ordering in the css file causes `bg-transparent` to load later now upsetting the visuals. This will be the same reason as the other `bg-transparent` removals.

server/app/views/components/ButtonStyles.java
gwendolyngoetz, line 66: Why: The new ordering in the css file causes `bg-transparent` to load later now upsetting the visuals. Unlike the other `bg-transparent` changes removing the class, this one is adding it for the same general reason.

server/app/views/components/ButtonStyles.java
```diff
@@ -97,7 +96,9 @@ public final class ButtonStyles {
   public static final String LINK_STYLE =
       StyleUtils.joinStyles(
           StyleUtils.removeStyles(ApplicantStyles.LINK, "text-sm"),
-          "bg-transparent",
           "border-none",
           "hover:bg-opacity-0");
+
+  public static final String LINK_STYLE_PUBLISH_PROGRAM =
```
gwendolyngoetz, line 102: Why: Adding a new style because we have some places in the application that have issues with `bg-transparent` being there and others that have issues if it isn't there.

server/app/views/components/ButtonStyles.java
```diff
@@ -97,7 +96,9 @@ public final class ButtonStyles {
   public static final String LINK_STYLE =
       StyleUtils.joinStyles(
           StyleUtils.removeStyles(ApplicantStyles.LINK, "text-sm"),
-          "bg-transparent",
           "border-none",
           "hover:bg-opacity-0");
+
+  public static final String LINK_STYLE_WITH_TRANSPARENCY =
```
gwendolyngoetz, line 102: Why: Same issues with `bg-transparency`. Unfortunately there some places need it, some don't. Creating this new variable to be the one with transparency and updating the couple spots that need it. The existing one without `bg-transparency` stays on all other existing buttons.

avaleske (approved):
Amazing thank you!

PR #6590, opened 2024-02-07
Reverting changes in CSV export filter to use LocalDateTime

### LINKED ISSUE #5824: Reporting submissions don't match with CSV export

swatkat1:
### Description

Reverting changes in CSV export filter to use LocalDateTime. Commit Id - f6367df
In this PR I am reverting Fix-1.

Original Description  - 
Fix for Application count mismatch between CVS exports and Reporting.
Fix 1- Internally we store all submit time as UTC. On trying to the retrieve the records for dd-mm-yyyy, we should also covert it to UTC time zone so that we will get the same results.
Fix 2 - When converting java.sql.Timestamp into MM/yyyy format for reporting, there is no need to add time zone information as the time stamp is always set to the beginning of the month eg. 01-01-2024 00:00:000. So SimpleDateTimeFormatter is enough to extract the MM/yyyy format.

## Release notes

Reverting the Fix-1 from original PR - Commit Id - f6367df

#### General

### Issue(s) this completes

Fixes #5824

### DIFF
--- a/server/app/controllers/admin/AdminApplicationController.java
+++ b/server/app/controllers/admin/AdminApplicationController.java
@@ -226,7 +226,7 @@ public final class AdminApplicationController extends CiviFormController {
         .map(
             s -> {
               try {
-                return dateConverter.parseIso8601DateToStartOfUTCDateInstant(s);
+                return dateConverter.parseIso8601DateToStartOfLocalDateInstant(s);
               } catch (DateTimeParseException e) {
                 throw new BadRequestException("Malformed query param");
               }
--- a/server/app/services/DateConverter.java
+++ b/server/app/services/DateConverter.java
@@ -66,16 +66,6 @@ public final class DateConverter {
     return parseIso8601DateToLocalDate(dateString).atStartOfDay(zoneId).toInstant();
   }
 
-  /**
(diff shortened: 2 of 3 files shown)


gwendolyngoetz (approved):
I agree with the move to keep the method name changed with local in it. Removes ambiguity if you don't notice it in javadoc.

shreyachatterjee00:
@elisekalstad @swatkat1 we are good with this from the Exygy side, go ahead!

Issue #6399, opened 2024-01-17
Modify ApplicantModel to get/set data in new columns


nb1701:
#### Describe the solution presented in this issue.
Modify the `ApplicantModel` to add new fields for all of the new columns, as well as getter/setters for each of these fields.

Additionally, add the `ApplicantModel` to the `ApplicantData` object.  This was debated in the TDD, and is not ideal, but we think it is probably the least tangled way to give access to this data in all the places that need it.  Also modify the `getApplicantName` and `setUserName` functions to use the data from the `ApplicantModel` fields, rather than using `WellKnownPaths`.  Then remove `WellKnownPaths`.

Spike commits [1](https://github.com/civiform/civiform/pull/6202/commits/643c8fd7d88b8121b51a2691c07f60f2fa04613e) [2](https://github.com/civiform/civiform/pull/6202/commits/bf0f44b606082f83cb85de06b875f470916d4e4d) [3](https://github.com/civiform/civiform/pull/6202/commits/1b65bd98c8fe8393e7d19bd589609acfd41b6681)

#### Additional Context
[TDD](https://docs.google.com/document/d/13xTLRl82gDD0VajM68gVdOGUV6alXi_mtluf1uUT1Lo/edit)

#### Done when:
What criteria need to be met for the issue to be finished? 
- [ ] Code written along with tests
- [ ] `WellKnownPaths` removed

PR #6599, opened 2024-02-07
Modify ApplicantModel and ApplicantData for Primary Applicant Info

nb1701:
### Description

This modifies ApplicantModel to include the new columns for Primary Applicant Info. Note that empty (`""`) or blank (whitespace only) strings will get coerced into a null value so that the getters can return the correct Optional value.  This also modifies existing ApplicantData functions to set values in both the PAI columns as well as the Well Known Paths.  The get functions first try to get the data from the PAI columns in ApplicantInfo, and fall back to the WKP data if it is not found.  This is done so that existing functionality is not lost before this feature is complete, and to minimize the amount of changes needed in the many parts of the code currently operating off of ApplicantData.  Once the feature is complete, we will remove use of the Well Known Paths and then rely only on the PAI column data and modify all the call sites then.

The ApplicantData object now also includes the ApplicantModel object. This was debated in the TDD if this was a good idea, but we landed on doing this to simplify code that currently only has access to the ApplicantData object, so we don't have to pass down the ApplicantModel object separately.

#### General

### Issue(s) this completes

https://github.com/civiform/civiform/issues/6399

### DIFF
--- a/browser-test/src/primary_applicant_info.test.ts
+++ b/browser-test/src/primary_applicant_info.test.ts
@@ -231,7 +231,7 @@ test.describe('primary applicant info questions', () => {
     )
   })
 
-  test.only('logging in does not overwrite name with OIDC-provided name', async () => {
+  test('logging in does not overwrite name with OIDC-provided name', async () => {
     const {page, adminQuestions, adminPrograms, applicantQuestions} = ctx
 
     await loginAsAdmin(page)
@@ -255,19 +255,28 @@ test.describe('primary applicant info questions', () => {
     await applicantQuestions.answerNameQuestion('Geordi', 'LaForge')
     await applicantQuestions.clickNext()
     await applicantQuestions.submitFromReviewPage()
-    
+
     await logout(page)
     await loginAsProgramAdmin(page)
 
     await adminPrograms.viewApplications('test')
-    expect(await adminPrograms.selectApplicationCardForApplicant('LaForge')).not.toBeNull()
+    await expect(
+      page.locator(adminPrograms.selectApplicationCardForApplicant('LaForge')),
+    ).toBeVisible()
 
     await logout(page)
-    await loginAsTestUser(page, 'a:has-text("Log in")', false, 'LaForge, Geordi')
+    await loginAsTestUser(
+      page,
+      'a:has-text("Log in")',
+      false,
+      'LaForge, Geordi',
+    )
     await logout(page)
     await loginAsProgramAdmin(page)
 
     await adminPrograms.viewApplications('test')
-    expect(await adminPrograms.selectApplicationCardForApplicant('LaForge')).not.toBeNull()
+    await expect(
+      page.locator(adminPrograms.selectApplicationCardForApplicant('LaForge')),
+    ).toBeVisible()
   })
 })
--- a/browser-test/src/support/index.ts
+++ b/browser-test/src/support/index.ts
@@ -363,9 +363,7 @@ export const loginAsTestUser = async (
   if (displayName === '') {
     displayName = testUserDisplayName()
   }
-  await page.waitForSelector(
-    `:has-text("Logged in as ${displayName}")`,
-  )
+  await page.waitForSelector(`:has-text("Logged in as ${displayName}")`)
 }
 
 async function loginAsTestUserSeattleStaging(page: Page, loginButton: string) {


server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -279,15 +279,28 @@ private ImmutableList<AccountModel> searchAccounts(
...
+                            || account
+                                .newestApplicant()
+                                .get()
+                                .getApplicantData()
+                                .getDateOfBirthAtWellKnownPath()
+                                .get()
+                                .equals(maybeDOB.get())))
                     || (!missingParams.contains(SearchParameters.ParamTypes.NAME)
                         && account
                             .getApplicantName()
```
rockycodes, line 306: suggestion (maybe blocking): I think `account.getApplicantName` calls out to `applicantData.getApplicant` name so you might need to handle both cases in `account.getApplicantName` as well.

server/app/views/applicant/EditTiClientView.java
```diff
@@ -95,20 +100,43 @@ private DivTag renderEditClientForm(
...
+    ApplicantData applicantData = applicant.getApplicantData();
     FormTag formTag =
         form()
             .withId("edit-ti")
             .withMethod("POST")
             .withAction(routes.TrustedIntermediaryController.updateClientInfo(account.id).url());
+    // TODO (#5503): Remove using WKP functions when removing feature flag
+    Optional<String> firstName =
+        settingsManifest.getPrimaryApplicantInfoQuestionsEnabled(request)
+            ? applicant.getFirstName()
```
rockycodes, line 113: question (maybe blocking): do we need to check that this is present and use `.get` before setting it as a value below?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -306,8 +306,7 @@ private DivTag renderCardContactInfo(AccountModel account) {
     if (newestApplicant.isEmpty()) {
       return div();
     }
-    ApplicantData applicantData = newestApplicant.get().getApplicantData();
-    Optional<String> maybePhoneNumber = applicantData.getPhoneNumber();
+    Optional<String> maybePhoneNumber = newestApplicant.get().getPhoneNumber();
```
rockycodes, line 309: question (blocking): Do you need to handle the WKP phone number here as well?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -385,13 +384,15 @@ private DivTag renderClientDateOfBirth(AccountModel account) {
     if (newestApplicant.isEmpty()) {
       return div();
     }
+    // TODO (#5503): Remove checking at WKP when removing feature flag
+    Optional<LocalDate> currentDobAtWKP =
+        newestApplicant.get().getApplicantData().getDateOfBirthAtWellKnownPath();
+    Optional<LocalDate> currentDobAtPIA = newestApplicant.get().getDateOfBirth();
```
rockycodes, line 390: suggestion (non-blocking): Use `newestApplicant.get().getDateOfBirth()` directly instead of naming it `currentDobAtPIA`. I think it's clear enough what's going on and then you don't have to refactor it later.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -75,8 +75,7 @@ public void testUpdateClientInfo_ThrowsApplicantNotFoundException()
     account.setManagedByGroup(tiGroup2);
     account.save();
 
-    assertThat(testApplicant.get().getApplicantData().getDateOfBirth().get().toString())
-        .isEqualTo("2022-07-18");
+    assertThat(testApplicant.get().getDateOfBirth().get().toString()).isEqualTo("2022-07-18");
```
rockycodes, line 78: suggestion (non-blocking): keep both tests in here for now and remove the WKP test cases when you remove those code paths.

server/test/repository/AccountRepositoryTest.java
```diff
@@ -75,24 +73,33 @@ public void updateClientNameTest_AllNonEmpty() {
...
+        account,
+        applicantUpdateTest,
+        "Dow",
+        "James",
+        "John",
+        "1234567890",
+        "note",
+        "email@address.com",
+        "2020-10-10");
+    assertThat(applicantUpdateTest.getFirstName().get()).isEqualTo("Dow");
```
rockycodes, line 85: suggestion (non-blocking): same suggestion as above. test that these values are set in both places for now.

server/test/repository/ApplicationRepositoryTest.java
```diff
@@ -118,8 +118,10 @@ public void createOrUpdateDraftApplication_updatesExistingDraft() {
     // Since this is a draft application, it shouldn't have a submit time set.
     assertThat(appDraft2.getSubmitTime()).isNull();
     // Applicant data is not saved in the application until it is submitted.
-    assertThat(appDraft1.getApplicantData().getApplicantName()).isEmpty();
-    assertThat(appDraft2.getApplicantData().getApplicantName()).isEmpty();
+    // Though we are not saving the applicant name to a well known path anymore,
```
rockycodes, line 121: suggestion (non-blocking): delete the extra comment. It's helpful in the context of this PR but will be come outdated as soon as we remove the WKP code paths.

server/test/services/ti/TrustedIntermediaryServiceTest.java
```diff
@@ -271,7 +271,7 @@ public void addClient_WithEmptyEmailAddress() {
             .filter(acct -> acct.getApplicantName().equals("Email, No"))
             .findFirst()
             .get();
-    assertThat(account.getApplicants().get(0).getApplicantData().getDateOfBirth().get().toString())
+    assertThat(account.getApplicants().get(0).getDateOfBirth().get().toString())
```
rockycodes, line 274: suggestion (non-blocking): same comment as above, keep tests for both code paths.

rockycodes (changes requested):
Overall this looks great! Well done making a potentially hairy situation very easy to follow.

server/app/services/ti/TrustedIntermediaryService.java
rockycodes: TI client names show up now!

server/app/services/applicant/ApplicantData.java
```diff
@@ -224,4 +272,8 @@ private static void clearFieldsNotRequiredForComparison(ApplicantData applicantD
       // Metadata may be missing in unit tests. No harm, no foul.
     }
   }
+
+  public Optional<LocalDate> getDeprecatedDateOfBirth() {
```
rockycodes, line 276: nitpick (non-blocking): This can be private now, it's only used in this file.

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -24,40 +27,72 @@ public void hasPreferredLocale_onlyReturnsTrueIfPreferredLocaleIsSet() {
...
   public void getApplicantName_exists() {
-    ApplicantData data = new ApplicantData();
+    ApplicantData data = createNewApplicantData();
     data.setUserName("First Last");
     assertThat(data.getApplicantName()).isEqualTo(Optional.of("Last, First"));
   }
 
+  // TODO (#5503): Remove this when we remove this check from ApplicantData#getApplicantName
+  @Test
+  public void getApplicantName_fallsBackToWKPWhenNameIsEmail() {
```
rockycodes, line 64:
```suggestion
  public void getApplicantName_fallsBackToWellKnownPathWhenNameIsEmail() {
```

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -24,40 +27,72 @@ public void hasPreferredLocale_onlyReturnsTrueIfPreferredLocaleIsSet() {
...
+    applicant.setAccount(account);
+
+    applicant.save();
+    return new ApplicantData(applicant);
+  }
+
   @Test
   public void getApplicantName_exists() {
-    ApplicantData data = new ApplicantData();
+    ApplicantData data = createNewApplicantData();
```
rockycodes, line 57: suggestion (non-blocking): Do you want to add a test that values fall back to WKP when PAI values aren't available? Maybe that's overkill since it would basically just be testing that `.or` call.

server/app/services/applicant/ApplicantData.java
```diff
@@ -135,15 +199,19 @@ public void setUserName(String displayName) {
   }
 
   public void setUserName(
-      String firstName, @Nullable String middleName, @Nullable String lastName) {
+      String firstName, Optional<String> middleName, Optional<String> lastName) {
```
rockycodes, line 202: comment (non-blocking): It's definitely not for you to address in this PR, but it's confusing to me to have a `setUserName` function and an `updateUserName` function. Maybe we can address this when we remove the WKP logic.

rockycodes (changes requested):
Just one blocking comment about an if statement, but overall I think this looks great!

server/test/services/applicant/ApplicantDataTest.java
rockycodes: Oh wow! Glad you tried it.

server/test/services/applicant/ApplicantDataTest.java
```diff
@@ -134,12 +182,52 @@ public void isDuplicateWithMetadata_returnsTrue() {
...
+            "{\"applicant\":{\"name\":{\"first_name\":\"First\",\"last_name\":\"Last\",\"program_updated_in\":2,\"updated_at\":1690293297676}}}",
+            new ApplicantModel());
 
     assertThat(data1.isDuplicateOf(data2)).isTrue();
     assertThat(data2.isDuplicateOf(data1)).isTrue();
   }
+
+  // TODO (#5503): Remove this when we remove the feature flag
+  @Test
+  public void fallsBackToWellKnownPathWhenPrimaryApplicantInfoDoesNotExist() {
```
rockycodes, line 198: praise: great test addition! extra credit for the names 🖖

rockycodes (changes requested):
Apologies to go back to you on this again! I thought of a potential problem with reworking `setUserName` that we might want to account for.

server/app/services/applicant/ApplicantData.java
```diff
@@ -66,84 +76,185 @@ public void setPreferredLocale(Locale locale) {
...
   public void setUserName(
-      String firstName, @Nullable String middleName, @Nullable String lastName) {
-    if (!hasPath(WellKnownPaths.APPLICANT_FIRST_NAME)) {
-      putString(WellKnownPaths.APPLICANT_FIRST_NAME, firstName);
+      String firstName, Optional<String> middleName, Optional<String> lastName, boolean overwrite) {
+    Path firstPath = WellKnownPaths.APPLICANT_FIRST_NAME;
+    Path middlePath = WellKnownPaths.APPLICANT_MIDDLE_NAME;
+    Path lastPath = WellKnownPaths.APPLICANT_LAST_NAME;
+    if (!overwrite
+        && (applicant.getFirstName().isPresent()
```
rockycodes, line 235: suggestion (non-blocking): this logic is hard to parse. maybe save it in a var called `haveFirstName` first and use that?

server/app/services/applicant/ApplicantData.java
```diff
@@ -66,84 +76,185 @@ public void setPreferredLocale(Locale locale) {
...
         // fallthrough
       default:
         // Too many names - put them all in first name.
         firstName = displayName;
     }
-    setUserName(firstName, middleName, lastName);
+    setUserName(firstName, middleName, lastName, overwrite);
   }
 
+  public void setUserName(String displayName) {
```
rockycodes, line 210: comment (non-blocking): I realized this method is only used in tests/test setup. We could refactor those calls to use the one that looks for first middle last so we could get rid of one of the `setUserName` functions. I don't think you need to do that in this PR, but I might look at that when I'm doing system actions if that makes sense to you.

rockycodes (approved):
Wahoo finish line! I don't love having 4 `setUserName` functions, but I don't really see a way around it for now. I left a couple of non-blocking comments that you should feel free to resolve whenever!

PR #6600, opened 2024-02-08
Adding tests to cover using address as a visibility condition

### LINKED ISSUE #6505: Address Service Area not working as Visibility Condition
**Describe the bug**
When crafting a visibility condition using an address question the service area is used as criteria for show/hide.  The condition never shows the question as it should.  It is always hidden.

This is needed when an application may have a different question for a Seattle resident vs someone who is outside the city but within the service area for a Seattle utility or department.

**To Reproduce**

1. Create a new program question screen after an address question
2. Add a question to the screen
3. Click Create visibility condition 
4. Select the address question 
5. The service area criteria will be displayed (currently only Seattle is available service area), select SHOW
6. Save Criteria
7. Publish program

**Screenshots**
Image

gwendolyngoetz:
### Description

Adding these two tests to cover using Address eligibility as a Visibility Condition that is NOT gated. This works and I don't want to find it gets broken at some point. I'm not certain, but I don't think this is working because it was designed this way. Getting it working is quite convoluted. I suspect it's a happy accident of overlapping features. 

I'll write up some documentation in a separate PR. I'm attaching videos of the output for reference when I write docs.

#### General

### Issue(s) this completes
Fixes #6505

[ admin setup of the program.webm](https://github.com/civiform/civiform/assets/195162/0730ddea-2d05-4b91-806e-0e599e724ac9)
[when address is eligible show hidden screen.webm](https://github.com/civiform/civiform/assets/195162/5fdf19b9-8ae6-4735-8e77-520e8b09c353)
[when address is not eligible do not show hidden screen.webm](https://github.com/civiform/civiform/assets/195162/e5f7a41a-ef92-407b-aeb0-cef80e3d1d84)

### DIFF
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -1697,7 +1697,7 @@ describe('Applicant navigation flow', () => {
       await applicantQuestions.answerTextQuestion('answer 2')
       await applicantQuestions.clickNext()
 
-      await applicantQuestions.expectAddressHasBeenCorrected(
+      await applicantQuestions.expectQuestionAnsweredOnReviewPage(
         questionAddress,
         'Address In Area',
       )
@@ -1726,7 +1726,7 @@ describe('Applicant navigation flow', () => {
       await applicantQuestions.answerTextQuestion('answer 2')
       await applicantQuestions.clickNext()
 
-      await applicantQuestions.expectAddressHasBeenCorrected(
+      await applicantQuestions.expectQuestionAnsweredOnReviewPage(
         questionAddress,
         'Nonlegit Address',
       )
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -1589,6 +1589,10 @@ describe('Applicant navigation flow', () => {
       const {page, adminQuestions, adminPrograms, adminPredicates} = ctx
       await loginAsAdmin(page)
       await enableFeatureFlag(page, 'esri_address_correction_enabled')
+      await enableFeatureFlag(
+        page,
+        'esri_address_service_area_validation_enabled',
+      )
 
       // Create Questions
       await adminQuestions.addAddressQuestion({
@@ -1673,7 +1677,6 @@ describe('Applicant navigation flow', () => {
 
     it('when address is eligible show hidden screen', async () => {
       const {page, applicantQuestions} = ctx
-      await enableFeatureFlag(page, 'esri_address_correction_enabled')
       await applicantQuestions.applyProgram(programName)
 
       // Fill out application and submit.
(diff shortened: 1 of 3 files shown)


browser-test/src/application_navigation.test.ts
```diff
@@ -1438,5 +1438,161 @@ describe('Applicant navigation flow', () => {
...
+    const questionText1 = 'text-test-q-one'
+    const questionText2 = 'text-test-q-two'
+    const screen1 = 'Screen 1'
+    const screen2 = 'Screen 2'
+    const screen3 = 'Screen 3'
+
+    beforeAll(async () => {
+      const {page, adminQuestions, adminPrograms, adminPredicates} = ctx
+      await loginAsAdmin(page)
+      await enableFeatureFlag(page, 'esri_address_correction_enabled')
```
rockycodes, line 1453: suggestion (maybe blocking): enable `ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ENABLED` as well. I think this is working without that because they're both set to true in `application.dev.conf`

browser-test/src/application_navigation.test.ts
```diff
@@ -1438,5 +1438,161 @@ describe('Applicant navigation flow', () => {
...
+
+      await adminQuestions.addTextQuestion({
+        questionName: questionText2,
+        questionText: questionText2,
+      })
+
+      // Create Program
+      await adminPrograms.addProgram(programName)
+
+      // Attach questions to program
```
rockycodes, line 1474: praise: helpful comments!

browser-test/src/application_navigation.test.ts
```diff
@@ -1438,5 +1438,161 @@ describe('Applicant navigation flow', () => {
...
+      // Publish Program
+      await adminPrograms.gotoAdminProgramsPage()
+      await adminPrograms.publishProgram(programName)
+
+      await logout(page)
+    })
+
+    it('when address is eligible show hidden screen', async () => {
+      const {page, applicantQuestions} = ctx
+      await enableFeatureFlag(page, 'esri_address_correction_enabled')
```
rockycodes, line 1538: question (non-blocking): do you need to enable the feature flag again since you enable it in the `beforeAll`?

browser-test/src/application_navigation.test.ts
```diff
@@ -1438,5 +1438,161 @@ describe('Applicant navigation flow', () => {
...
+        '',
+        'Seattle',
+        'WA',
+        '98109',
+        0,
+      )
+      await applicantQuestions.clickNext()
+      await applicantQuestions.expectVerifyAddressPage(true)
+      await applicantQuestions.clickNext()
+      await applicantQuestions.answerTextQuestion('answer 1')
```
rockycodes, line 1553: suggestion (non-blocking): Add an expect here that tests that the screen showing is indeed the hidden screen. Alternatively, add a comment saying something like "this screen will be hidden if the address is not eligible".

rockycodes (changes requested):
This looks great! Props on figuring out it works like this. Just one blocking comment about enable the other feature flag and a couple of suggestions.

browser-test/src/application_navigation.test.ts
gwendolyngoetz: Good catch

browser-test/src/application_navigation.test.ts
gwendolyngoetz: Given it's running in the same textcontext probably not.

rockycodes (approved):

PR #6620, opened 2024-02-10
Don't run address visiblity test on staging

gwendolyngoetz:
These tests fail if not using the mock web services since data is different from live endpoints. Don't run them on staging environments.

All this does is wrap the tests in an `if (isLocalDevEnvironment())`. The diff looks busy, but the test itself remains unchanged.

#### General

### DIFF
--- a/browser-test/src/application_navigation.test.ts
+++ b/browser-test/src/application_navigation.test.ts
@@ -1576,164 +1576,170 @@ describe('Applicant navigation flow', () => {
     })
   })
 
-  describe('using address as visibility condition', () => {
-    const programName = 'Test program for address as visibility condition'
-    const questionAddress = 'address-test-q'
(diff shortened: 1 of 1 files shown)


dkatzz (commented):
These work on the normal staging probers, right? Just not on the Seattle ones since its setup differently?

gwendolyngoetz:
They will only run on on local builds or in the the GitHub actions and not against any staging site. Each staging site has a different configuration so we can't count on the tests passing anywhere but against the mock-web-services.

dkatzz (approved):

dkatzz:
> They will only run on on local builds or in the the GitHub actions and not against any staging site. Each staging site has a different configuration so we can't count on the tests passing anywhere but against the mock-web-services.

Makes sense! Thank you!

PR #6547, opened 2024-02-01
Have Previous button save answers or show an error modal.

caitlinshk:
### Description

This builds on https://github.com/civiform/civiform/pull/6544 to also have the "Previous" button in the application flow to save the applicant's answers so that applicants don't lose any work. This PR does that by doing the following:

1) Updates the Previous button to submit the form using the `ApplicantProgramBlocksController#update` action (which is what the "Save & next" button currently uses and what the "Review" button uses in https://github.com/civiform/civiform/pull/6544). The Previous button provides a new parameter `ApplicantRequestedAction.PREVIOUS_BLOCK` to indicate the user wants to go to the previous block after the update.
3) If `#update` says that there were validation errors (like missing required questions) and the user clicked the Previous button, re-render the block with a new `errorDisplayMode=DISPLAY_ERRORS_WITH_MODAL_PREVIOUS`.
4) When `ApplicantProgramBlockEditView` re-renders and receives that new display mode, it will render the validation errors and also render and a modal asking the user whether they'd like to go back and edit the answers or proceed to the previous page without saving.
    - Clicking "Go back and edit" will just close the modal, allowing the applicant to see the validation errors and fix them.
    - Clicking "Go to previous page without saving" will take the applicant to the previous page and not save any answers.
    
Items that will be in future PRs:
- Updating the Previous button on address correction
- Updating the Previous button on file upload

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers).
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)

#### User visible changes

- [x] Followed steps to [internationalize new strings](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#internationalization-for-application-strings) -- **This is protected behind a feature flag, so the strings are used but not translated yet. I'll only enable the feature flag once the strings have been translated.**
- [x] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [x] Manually tested at 200% size
- [x] Manually evaluated tab order

#### New Features

[...]

Issue #6450, opened 2024-01-23
Clicking Review Without Saving Deletes all Data


lftaylor2023:
Using Arkansas' "start here" form as an example, when logged in as a guest if you complete all 11 questions but hit "Review" rather than "Save and Next," all data is deleted and you have to start over. This may be the intended functionality but I think we need to anticipate that applicants are going to do this from time to time. I haven't tried this as a logged in user yet because I'm running out of throw away email accounts. I'm assuming that the functionality will be the same as a logged in user who is using Civiform for the first time and hasn't save any data in any forms. 

Possible solutions, in no order of precedence: 

- Force the form data to always be cached in the browser when clicking "review" before "save and next."
- Add text below the row of buttons that warns users that if they choose "review" without first clicking "save and next," that their data will be lost and they will have to start over. 
- Don't show the "review" button until the form data has been saved. 

Additionally, it doesn't seem necessary to have "review, "previous" and "save and next" buttons on a form that consists of just one screen. There's no "previous" and no "next," in reality. So perhaps in scenarios when a form is only one screen, only present the user with the option to "Save and Review" their form entries.  

Leslie Taylor, State of Arkansas, ARData

caitlinshk:
It looks like we've had this reported multiple times in the past: https://github.com/civiform/civiform/issues/3601, https://github.com/civiform/civiform/issues/2470, and https://github.com/civiform/civiform/issues/1670. I agree this is confusing and we should fix this. The simplest fix is probably a confirmation modal saying that data won't be saved? Based on [Bion's comment here](https://github.com/civiform/civiform/issues/2470#issuecomment-1138033181), also saving the data when clicking "Review" begs the question of what to do if the data they entered fails validation.

lftaylor2023:
Can we cache everything that passes validation, flag everything that doesn't, and show those fields to the applicant as needing to be revised (red text that alludes to what the problem is)?

caitlinshk:
So have the same experience as if you click "Save and Next" but have validation errors? Probably! We'll talk with our designers and product managers to confirm what behavior should happen but that seems reasonable to me.

shreyachatterjee00:
Aligned that having the "Review" button having the same behavior as "Save and next" is a good fix for now. We'll also do some more thinking holistically from a UX perspective about how these buttons and navigation work overall.

caitlinshk:
[...]

PR #6651, opened 2024-02-15
Save uploaded file when clicking Review or Previous [behind SAVE_ON_ALL_ACTIONS flag]

caitlinshk:
### Description

To address https://github.com/civiform/civiform/issues/6450, we'd like to save any user data whenever *any* button is clicked in an application. Previous PRs have made the "Review" and "Previous" buttons save data for the typical block:
 - Review button: https://github.com/civiform/civiform/pull/6544
 - Previous button: https://github.com/civiform/civiform/pull/6547

However, those PRs didn't update the "Review" and "Previous" buttons for the file upload block because the file upload block has some custom logic. This PR updates the file upload block to save data when "Review" and "Previous" buttons are clicked.

#### How saving on all actions works

Previously, a user's answers were only saved when a user clicked "Save&next". So, the "Save&next" button invoked an endpoint called `ApplicantProgramBlocksController#update`, which would (1) Save the user's data to the database; and (2) Redirect the user to the next block in the program.

To have *all* buttons save the user's data, we've updated all buttons to use that `ApplicantProgramBlocksController#update` endpoint. We've added a parameter to the endpoint called `applicantRequestedAction` which indicates *where* the redirect should go ("Save&next" passes in `NEXT_BLOCK`, "Previous" passes in `PREVIOUS_BLOCK`, "Review" passes in `REVIEW_PAGE`.)

 #### How file upload works

When a user uploads a file then clicks "Save & next", the HTML form is submitted to AWS (or other cloud storage provider), *not* CiviForm. AWS stores the file in a bucket, and if the file was successfully stored then AWS invokes the URL specified by the `<input name="success_action_redirect">` in the HTML form. We currently set the `success_action_redirect` URL to point to `ApplicantProgramBlocksController#updateFile`. That endpoint will (1) Save the file key in the database so that we know how to fetch the file back from AWS later; and (2) Redirect the user to the next block in the program.

Now that all the buttons should save the file, all the buttons will need to follow a similar pattern of submitting the HTML form to AWS First. They'll also need the `#updateFile` endpoint to handle redirecting to different places.

#### File upload question requirements

Because the file upload question first sends the HTML form to AWS, the file upload question can't be on a block with any other questions because AWS will throw away all the other question inputs. File upload questions are always on a block by themselves.

### Expected behaviors

See the manual testing directions for expected behaviors for each of the buttons.

### Implementation

Given those background details above, this PR makes a few changes:
1) Updates all buttons on the file upload question page to be `submitButton`s so that they send the HTML form to AWS first, then trigger `ApplicantProgramBlocksController#updateFile`.
1) Updates `ApplicantProgramBlocksController#updateFile` to also take in an `applicantRequestedAction` parameter so that it can redirect to different places based on which button was clicked.
3) Updates the TypeScript on the file upload question page to change what the `<input name="success_action_redirect"> is based on which button was clicked.
4) Updates the TypeScript on the file upload question page to allow the "Review" and "Previous" buttons to do the correct redirect even if the user didn't upload a file.

#### General

#### User visible changes

#### New Features

Notably: The Previous and Review buttons work correctly whether or not the user uploaded a file. The Save&next button will only work if the user *has* uploaded a file.

Basic buttons:
 - Save&next:
     -  *without* an uploaded file -> an error is shown and the user stays on the block. **Note:** This is true for both required file upload questions and optional file upload questions. Optional file upload questions include a separate Skip button to let the user proceed to the next block without uploading a file.
    - *with* an uploaded file -> file saved, user taken to next block
 - Previous:
      -  *without* an uploaded file -> user taken to previous block
      - *with* an uploaded file -> filed saved, user taken to previous block
 - Review:
      - *without* an uploaded file -> user taken to review page
     - *with* an uploaded file -> filed saved, user taken to review page
 
Bonus buttons:
 - Continue (only shown if you have previously uploaded a file then come back to the question):
      -  *without* an uploaded file -> user taken to next block
      - *with* an uploaded file -> file **not** saved, user taken to next block. **Note:** I think this behavior is worth changing, but that's out-of-scope for this PR.
 - Skip (only shown on optional file upload questions that don't already have a previously uploaded file):
      -  *without* an uploaded file -> user taken to next block
      - *with* an uploaded file -> file **not** saved, user taken to next block
 - Delete (only shown on optional file upload questions and if you have previously uploaded a file then come back to the question):
     - deletes previously uploaded file, user taken to next block

### Issue(s) this completes

Part of #6450


server/test/views/questiontypes/ApplicantQuestionRendererFactoryTest.java
```diff
@@ -43,7 +46,10 @@ public void rendererExistsForAllTypes(QuestionType type) throws UnsupportedQuest
 
     ApplicantQuestionRendererFactory factory =
         new ApplicantQuestionRendererFactory(
-            new ApplicantFileUploadRenderer(new AwsFileUploadViewStrategy(), applicantRoutes));
+            new ApplicantFileUploadRenderer(
+                new AwsFileUploadViewStrategy(),
+                applicantRoutes,
+                new SettingsManifest(ConfigFactory.parseMap(ImmutableMap.of()))));
```
dkatzz, line 52: **suggestion:** can you do `instanceOf(SettingsManifest.class)`? Here and below?

dkatzz (commented):
Thanks for the thorough testing and explanations!! Some comments but nothing major.

dkatzz:
The description is super helpful! One thing that would be helpful to call out is that file upload questions will be the only ones on a particular block. It took me a minute to remember that :)

server/app/assets/javascripts/file_upload.ts
```diff
@@ -40,6 +51,61 @@ export function init() {
...
+  )
+
+  if (fileInput.value != '') {
+    modifySuccessActionRedirect(buttonTarget, blockForm)
+  } else {
+    const redirectWithoutFile = buttonTarget.dataset.redirectWithoutFile
+    if (redirectWithoutFile) {
+      // If there's no file uploaded but the button provides a redirect
+      // that can be used even when there's no file, invoke that redirect.
+      // See ApplicantFileUploadRenderer.java.
```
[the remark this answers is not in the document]
caitlinshk: Done.

server/app/assets/javascripts/file_upload.ts
```diff
@@ -40,6 +51,61 @@ export function init() {
...
+      window.location.href = redirectWithoutFile
+      // This will prevent form submission, which is important because we
+      // don't want to send an empty file to cloud storage providers or
+      // store an empty file key in our database.
+      e.preventDefault()
+    }
+  }
+  return true
```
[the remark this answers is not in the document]
caitlinshk: Don't think so! Removed.

server/app/assets/javascripts/file_upload.ts
```diff
@@ -40,6 +51,61 @@ export function init() {
...
+}
+
+/**
+ * Modifies the "success_action_redirect"-named <input> to have the correct redirect
+ * location based on the button that was clicked.next
+ *
+ * Context: When a user submits a file upload, the <form> data is first sent to the
+ * cloud storage provider to store the file. Once the file is successfully uploaded,
```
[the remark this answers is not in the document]
caitlinshk: Good idea, done.

server/app/views/applicant/ApplicantFileUploadRenderer.java
```diff
@@ -47,14 +53,38 @@ public final class ApplicantFileUploadRenderer extends ApplicationBaseView {
   private static final String FILEUPLOAD_SKIP_BUTTON_ID = "fileupload-skip-button";
   private static final String FILEUPLOAD_CONTINUE_BUTTON_ID = "fileupload-continue-button";
 
+  /**
+   * A data key that points to a redirect URL that should be used if the user has uploaded a file.
+   * Should be set on each action button if the SAVE_ON_ALL_ACTIONS flag is enabled.
+   *
+   * <p>Should be kept in sync with file_upload.ts.
```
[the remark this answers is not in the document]
caitlinshk: Done.

server/test/views/questiontypes/ApplicantQuestionRendererFactoryTest.java
caitlinshk: `instanceOf` is only defined in `ResetPostgres` and this test class currently doesn't extend `ResetPostgres`. I can make it extend `ResetPostgres` if that's preferred, but I figured if it didn't have to then the test might be faster.

dkatzz: Ah got it! There was recently this [PR](https://github.com/civiform/civiform/pull/6694) related to adding ResetPostgres to more tests - I wonder if @gwendolyngoetz has thoughts here.

dkatzz (approved):

server/test/views/questiontypes/ApplicantQuestionRendererFactoryTest.java
gwendolyngoetz: They were expecting a database schema to exist. My database at the time was completely empty. Running the test would then fail since nothing elsewhere had already created the schema. In `bin/sbt-test` run `testOnly package.name.to.YourTest` to run the specific test. You might not see any problems if your unit test database already exists though since the postgres volume is persistent.

caitlinshk: Ah got it, thanks for the clarification! I did a `docker system prune && docker volume prune` then ran the test and it still passed so I don't think it needs `ResetPostgres` but happy to add it if we want that to be the test standard.

gwendolyngoetz: If you still have created containers `docker ps -a` pruning wouldn't have removed existing volumes.

caitlinshk: Oh I deleted the containers first, I should've mentioned that above.

Issue #6464, opened 2024-01-25
Standardize number of digits on date/time fields


elisekalstad:
As a CiviForm Admin I need to export the demographic data to update metrics such as time to complete applications. It is difficult to analyze the date/time field for Create Time and Submit Time because the times have a different number of characters. Excel and Tableau are both struggling (or maybe it's me!) to read them as "times". 

#### Describe the solution you'd like.
Would it be possible to consistently use 6 digits to display the time? 

Current: 6:53:78
Proposed: **0**6:53:78

I'd like to be able to export the date/time and have each cell have the same number of values in it.

#### What priority is this?

Low: I can struggled without it, but it's a bit cumbersome.

elisekalstad:
This is also an issue as a Program Admin downloading the CSV for applications.

swatkat1:
@elisekalstad  - If it's a CSV issue, can we get a sign off from other entities too?

avaleske:
Confirmed with Matthew at Arkansas that since they're not fully live yet we can switch to an ISO date format without needing to put it behind a feature flag first. Once we know what release it's in we can let them know.

I _think_ the best format to use is the `ISO_OFFSET_DATE_TIME` from https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

swatkat1:
@nb1701 and @avaleske --> Elise is OOO till mid march. I am adding @msprenke for visibility.

PR #6667, opened 2024-02-15
Use two-digit hour in exports

nb1701:
This string not being a predictable length for every entry caused problems for Excel and Tableau to automatically read the data as timestamps. This changes it to always print a two-digit hour, prepending with a 0 as needed.

https://github.com/civiform/civiform/issues/6464


avaleske (approved):
Yet another thing I wish I had noticed when I started API stuff last summer. My "current status" audit was... not thorough.

It's _probably_ fine, but before merging we should:
- a) Test what happens when parsing two digit hour datetime with a single digit hour format, like `LocalDate.parse(text, "yyyy/MM/dd 'at' h:mm a z")`. If it works fine then there's less concern here.
- b) Let the civic entities know this change is coming and get their signoff, especially Arkansas. Matthew at Arkansas will be the one with the most knowledge of their API consuming code. (And if they're ok with it, changing it to an ISO date would be especially nice.)

Alternatively, we could add new time fields to the CSV and JSON API called `create_time_iso` and use an ISO format like we should've from the start, and eventually deprecate the current fields.

fwiw, the more formal process for changing an in-use API feature that I'm proposing is:
https://docs.google.com/document/d/1aKQ55fRdCxvYd9XMd2p9PJxQuSuEIPJoE9-O-pSjWsQ/edit#bookmark=id.qt2b8u6lfpgv

isabelcshaw:
@nb1701 @avaleske Matthew confirmed that he's good with using an ISO format

avaleske:
> @nb1701 @avaleske Matthew confirmed that he's good with using an ISO format

just said the same on the issue! https://github.com/civiform/civiform/issues/6464

avaleske (approved):
Thank you!

avaleske:
Oh please also update the api docs:
- https://github.com/civiform/docs/blob/main/docs/it-manual/api/program-applications.md#create_time
- https://github.com/civiform/docs/blob/main/docs/it-manual/api/program-applications.md#submit_time

PR #6668, opened 2024-02-16
Use sentence case

nb1701:
We mostly use sentence case for titles, headers, and button text, but sometimes we use title case. This standardizes things to use sentence case. Note that this does not fix strings that are localized.


server/app/views/BaseHtmlLayout.java
```diff
@@ -150,7 +150,8 @@ public Content render(HtmlBundle bundle) {
     } else {
       bundle.setTitle(String.format("%s — %s", currentTitle, getTitleSuffix()));
     }
-    bundle.addMetadata(meta().withName("thanks").withContent("Thank you Bion ❤️"));
+    // Best practice: add ❤️ every time you touch this file :)
```
rockycodes, line 153: ❤️

rockycodes (approved):

PR #6670, opened 2024-02-16
Find country code based on phone number

gwendolyngoetz:
### Description

Adding ability to find country code based on phone number.

We will be removing the country dropdown from the phone question control in the future and this will test each of the supported countries and selects the one that is valid or returns the a validation error message.

Adding unit tests.

#### General

### DIFF
--- a/server/app/services/PhoneValidationResult.java
+++ b/server/app/services/PhoneValidationResult.java
@@ -1,0 +1,73 @@
+package services;
+
+import com.google.auto.value.AutoValue;
+import java.util.Optional;
+
+/**
+ * {@link PhoneValidationResult} contains the phone number, country code if match found, and,
+ * validation message in the event validation fails
+ */
+@AutoValue
+public abstract class PhoneValidationResult {
+  public static Builder builder() {
+    return new AutoValue_PhoneValidationResult.Builder();
+  }
+
+  public abstract Optional<String> getPhoneNumber();
+
+  public abstract Optional<String> getCountryCode();
+
+  public abstract Optional<MessageKey> getMessageKey();
+
+  /**
+   * Convenience method to create a {@link PhoneValidationResult} object when a validation checks
+   * pass and a country code is found.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @param countryCode {@link Optional<String>}} of the Country Code in <a
+   *     href="https://www.javadoc.io/doc/com.googlecode.libphonenumber/libphonenumber/8.4.1/com/google/i18n/phonenumbers/PhoneNumberUtil.html">CLDR
+   *     format</a>
+   * @return a {@link PhoneValidationResult} object when a validation checks pass
+   */
+  public static PhoneValidationResult create(
+      Optional<String> phoneNumber, Optional<String> countryCode) {
+    return PhoneValidationResult.builder()
+        .setPhoneNumber(phoneNumber)
+        .setCountryCode(countryCode)
+        .build();
+  }
+
+  /**
+   * Convenience method to create a {@link PhoneValidationResult} object when a validation check
+   * fails.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @param messageKey {@link MessageKey} pertaining to the failed validation check
+   * @return a {@link PhoneValidationResult} object when a validation check fails
+   */
+  public static PhoneValidationResult createError(
+      Optional<String> phoneNumber, MessageKey messageKey) {
+    return PhoneValidationResult.builder()
+        .setPhoneNumber(phoneNumber)
+        .setMessageKey(Optional.of(messageKey))
+        .build();
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


server/app/services/PhoneValidationUtils.java
```diff
@@ -7,45 +7,110 @@
...
+  }
+
+  /**
+   * Gets an Optional string of Phone number and CountryCode values and runs validations on them.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @return an {@link Optional<String>} representing the country code if the phone number is valid,
+   *     or an Optional.empty() if the phone number is invalid
+   */
+  public static Optional<String> determineCountryCode(Optional<String> phoneNumber) {
```
avaleske, line 57: **suggestion (non-blocking)**: not-blocking cause it might come up as you do the rest of this refactor, but I'd prefer we return a more clear type, rather than assigning meaning to the empty optional.

Also we're assigning the opposite meaning to the empty optional than we do in `validatePhoneNumber()`, which is likely going to be confusing to reason about at the callsite.

Options for a return type could include adding a `phoneNumber` field to `PhoneValidationResult` and passing that out of the class, or using `ErrorAnd` to wrap the error and the result:
https://github.com/civiform/civiform/blob/b10353fdb25c81f2c65e29935b8c96bcefa02839/server/app/services/ErrorAnd.java#L13-L13
We do both of these things elsewhere so I think either is fine.

server/app/services/PhoneValidationUtils.java
```diff
@@ -7,45 +7,110 @@
...
-      return Optional.of(MessageKey.PHONE_VALIDATION_NUMBER_REQUIRED);
+    var result = validate(phoneNumber, countryCode);
+    return result.messageKey;
+  }
+
+  /**
+   * Gets an Optional string of Phone number and CountryCode values and runs validations on them.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @return an {@link Optional<String>} representing the country code if the phone number is valid,
```
avaleske, line 54: **suggestion**: As mentioned elsewhere, can we add a note about the expected format of the country code in this javadoc?

server/app/services/PhoneValidationUtils.java
```diff
@@ -7,45 +7,110 @@
...
+   * @param countryCode {@link Optional<String>}} of the Country Code
+   * @return an {@link Optional<MessageKey>} representing the validation error that occurred if the
    *     phone number is invalid, or an Optional.empty() if the phone number is valid
    */
   public static Optional<MessageKey> validatePhoneNumber(
       Optional<String> phoneNumber, Optional<String> countryCode) {
-    if (phoneNumber.isEmpty()) {
-      return Optional.of(MessageKey.PHONE_VALIDATION_NUMBER_REQUIRED);
+    var result = validate(phoneNumber, countryCode);
+    return result.messageKey;
```
avaleske, line 47: As mentioned on `determineCountryCode()`, it would be nice to return a type that's more clear about what happened. At the very least, we probably shouldn't have an empty optional here mean the opposite of what it does on `determineCountryCode()`.

avaleske (approved):
Approving since this is part of a larger refactor but I shared some thoughts about how we're passing the validation results back to the callsites.

server/app/services/PhoneValidationUtils.java
```diff
@@ -7,45 +7,110 @@
...
+
+    return Optional.empty();
+  }
+
+  /**
+   * Gets an {@link PhoneValidationResult} object with details on the validation result and valid
+   * country code value.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @param countryCode {@link Optional<String>}} of the Country Code
```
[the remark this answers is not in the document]
gwendolyngoetz: I'm not fond of point at a specific version, but after looking at the cldr unicode site this is going to be much easier for people to find the info.

server/app/services/PhoneValidationUtils.java
gwendolyngoetz: I'm expecting to remove the existing `validatePhoneNumber` that only returns an `Optional<MessageKey>` leaving just one that will return the `PhoneValidationResult`.

server/app/services/PhoneValidationResult.java
```diff
@@ -0,0 +1,73 @@
+package services;
+
+import com.google.auto.value.AutoValue;
+import java.util.Optional;
+
+/**
+ * {@link PhoneValidationResult} contains the phone number, country code if match found, and,
```
avaleske, line 7: **nit** extra comma

avaleske (approved):
Thanks for the changes!

server/app/services/PhoneValidationUtils.java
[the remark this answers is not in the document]
avaleske: Yeah I tried to find the info on the CLDR site too...

avaleske (approved):

PR #6676, opened 2024-02-17
Removing Country Code DropDown from Phone Question

### LINKED ISSUE #6525: Default country code for phone numbers

Most of our applicants (if not all of them!) will have a U.S. phone number. Is it possible to set the "Country" on the phone number question type to default to "United States" automatically? Alternatively, could we just hide the "Country" selector? It's an extra question that doesn't impact the input of the phone number; that is, no matter what the person selects, they are still prompted to enter a 10-digit number. 

As an applicant, I do not want to answer the "County" question when asked for my phone number.

#### What priority is this?

P2-we'd like to switch out the phone number question that is currently in use in all program forms (currently just a text question type) with the actual phone number question type. But it would be helpful if that change doesn't add an additional question for residents to complete. So we'd like to remove the "Country" question before switching the questions in all program forms. 

We would also like to align the format of the Seattle Login phone number question with the CiviForm phone number question--both should be 10-digit phone numbers. The Seattle Login does not ask for country code, so this information is extraneous.

#### Associated civic entity

Seattle

#### User Group

Feedback is coming from the CiviForm Admins. Residents have not yet tested this feature.

#### Additional Context
Current State and Options
image

Seattle Login Phone Formatting:
image

#### Done when:

Applicants no longer need to select a Country.

gwendolyngoetz:
### Description

Removing Country Code DropDown from Phone Question. Additionally this fixes the input formatting when there are multiple phone number questions on the same screen.

We now call the new method which will determine the County Code and it's validity based on the supported Country Code options (US and CA at this time).

#### General

- [x] Updated units where needed

### Issue(s) this completes

Fixes #6525


server/app/assets/javascripts/phone.ts
```diff
@@ -19,11 +19,13 @@ function formatPhone(element: HTMLElement) {
 }
 
 export function init() {
-  const element = document.getElementById('cf-phone-number')
-  const countryElement = document.getElementById('cf-phone-country-code')
-  if (element && countryElement) {
-    formatPhone(element)
-  }
+  document.querySelectorAll('[id^=cf-phone-number]').forEach((el) => {
+    // Our questions automatically include an element matching id appended with "-errors"; filter them out.
```
rockycodes, line 23: praise: helpful comment! And good to know.

rockycodes: suggestion (non-blocking): add a sentence to this comment explaining that the phone elements will have a unique id with a number prepended on the end which is why you have to select them this way and then filter out the `error` ones.

server/app/views/questiontypes/PhoneQuestionRenderer.java
```diff
@@ -78,26 +56,16 @@ protected DivTag renderInputTags(
                     phoneQuestion.getPhoneNumberPath(), ImmutableSet.of()))
             .setScreenReaderText(applicantQuestion.getQuestionTextForScreenReader())
             .addReferenceClass(ReferenceClasses.PHONE_NUMBER)
```
rockycodes, line 58: question (non-blocking): do we still need this reference class?

rockycodes (approved):
looks great! just a couple of non-blocking comments.

server/app/views/questiontypes/PhoneQuestionRenderer.java
gwendolyngoetz: At least for now. A lot of tests started to fall apart when I removed/changed it.

PR #6677, opened 2024-02-17
DOM text reinterpreted as HTML {Patch}

Shivam7-1:
patch for this issue: https://issuetracker.google.com/issues/324680545
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML. Always be cautious when dealing with user input or dynamic content to prevent security risks.


gwendolyngoetz:
Thank you @ Shivam7-1

PR #6679, opened 2024-02-20
Update pagination component on the TI dashboard to the USWDS component

### LINKED ISSUE #6211: Update pagination component to USWDS component
### Describe the solution:
Update the UI of the pagination of the table list so that the controls are clear and easy to use.

### Figma mock: 
https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=816%3A494&mode=design&t=A8ppwAkGj6ALGDd0-1

### Screenshots:

Image

### Done when:
- Users are able to paginate to a refreshed client list if there are more than 10 clients displayed
- Pagination matches the [USWDS pagination component](https://designsystem.digital.gov/components/pagination/)

melanie-exygy:
### Description

Replacing our custom pagination with the [USWDS pagination component](https://designsystem.digital.gov/components/pagination/), which is vetted for usability and accessibility.  This PR only replaces the pagination on the TI dashboard.  Pagination is also used in the Program Admin applications view, but UX needs to confirm changes for that page.  Once confirmed, we can easily switch over to the USWDS component for that view as well.

The rules for the desired behavior of the pagination component are:

- The component features a maximum of seven slots.
- Each slot can contain a navigation item or an overflow indicator.
- The first slot is always the first page of the set.
- If there are fewer than seven pages in the set, show only that number of slots.
- The component should always show the first page and current page.
- Show the next page, previous page, and last page if those pages exist.
- Display the same number of slots for each page in the set.

For examples of these desired behaviors, see [pagination behaviors](https://designsystem.digital.gov/components/pagination/#using-the-pagination-component-2).

Manual testing is impractical for any more than 2 pages because you would need to manually create over 20 new clients.  I relied on browser tests for multiple page scenarios.  However, if you'd like to test 2 pages manually, here are the steps:

1. Log in as a TI and go to the TI dashboard.
2. Create new clients until you have 11 of them.
3. Go to the bottom of the client card list to see the pagination.
4. The current page should be highlighted in black.
5. When on page 1, you should not see a "Previous" button.
6. When on page 2, you should not see a "Next" button.
7. There should be 10 client cards on the first page and 1 on the 2nd page.

### Issue(s) this completes

Fixes #6211


browser-test/src/support/ti_dashboard.ts
```diff
@@ -176,6 +189,18 @@ export class TIDashboard {
     )
   }
 
+  async expectPageNumberButtonOrNot(pageNum: string, isPresent: boolean) {
```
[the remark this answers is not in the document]
rockycodes: Alternatively you could call the expect methods directly in the test, or create two helper methods.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -753,4 +753,233 @@ describe('Trusted intermediaries', () => {
...
+      )
+    })
+
+    it('shows 2 pages when there are 12 clients', async () => {
+      const {page, tiDashboard} = ctx
+      await loginAsTrustedIntermediary(page)
+      await tiDashboard.gotoTIDashboardPage(page)
+      await waitForPageJsLoad(page)
+
+      await tiDashboard.createMultipleClients('myname', 11)
```
rockycodes, line 799: question (non-blocking): should the be 12? the test name says "when there are 12 clients".

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -753,4 +753,233 @@ describe('Trusted intermediaries', () => {
...
+
+      // Page 1 should still only show 10 clients
+      const cardCount = await page.locator('.usa-card__container').count()
+      expect(cardCount).toBe(10)
+
+      expect(await page.innerHTML('.usa-pagination__list')).not.toContain(
```
rockycodes, line 805: suggestion (non-blocking): add a comment like you have above eg `// No "Previous" button, but there should be a "Next" button`

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -753,4 +753,233 @@ describe('Trusted intermediaries', () => {
...
+      await tiDashboard.expectPageNumberButtonOrNot('5', true)
+      await tiDashboard.expectPageNumberButtonOrNot('6', true)
+      await tiDashboard.expectPageNumberButtonOrNot('7', true)
+      await tiDashboard.expectPageNumberButtonOrNot('8', false)
+
+      // Going to page 7
+      await page.click('[aria-label=Page7]')
+      expect(await page.innerHTML('.usa-current')).toContain('7')
+
+      expect(await page.innerHTML('.usa-pagination__list')).not.toContain(
```
rockycodes, line 864: suggestion (non-blocking): add a comment about what `usa-pagination_overflow` is. I think elipses?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -753,4 +753,233 @@ describe('Trusted intermediaries', () => {
...
+        'usa-pagination__overflow',
+      )
+
+      await validateScreenshot(
+        page.locator('.usa-pagination'),
+        'ti-pagination-two-ellipses',
+      )
+    })
+
+    it('shows one ellipses on the left when more than 7 pages and current page is one of the last 4 pages', async () => {
```
rockycodes, line 951: praise: super thorough testing!

rockycodes (approved):
I have a bunch of suggestions to help with clarity, but nothing blocking. This is a great improvement!

server/app/views/BaseHtmlView.java
```diff
@@ -146,6 +230,52 @@ protected DivTag renderPaginationDiv(
...
+  }
+
+  private LiTag renderNextPageButton(int page, Function<Integer, Call> linkForPage) {
+    return li().withClass("usa-pagination__item usa-pagination__arrow")
+        .with(
+            a().withClass("usa-pagination__link usa-pagination__next-page")
+                .attr("aria-label", "Next page")
+                .withHref(linkForPage.apply(page + 1).url())
+                .with(
+                    span("Next").withClass("usa-pagination__link-text"),
```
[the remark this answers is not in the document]
melanie-exygy: Unfortunately, the reason they have to be in a different order is that, going from left to right, the icon renders first for the "Previous" section and the icon renders last for the "Next" section.  I think the easiest way is to show how it will need to appear visually:  
image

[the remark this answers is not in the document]
rockycodes: Oh duh! I'm silly 😄

server/app/views/BaseHtmlView.java
```diff
@@ -124,7 +134,81 @@ private static String getCsrfToken(Http.Request request) {
...
+                .condWith(
+                    pageCount > 7 && page > 4 && page < pageCount - 3,
+                    renderPaginationEllipses(),
+                    renderPaginationPageButton(page - 1, false, linkForPage),
+                    renderPaginationPageButton(page, true, linkForPage),
+                    renderPaginationPageButton(page + 1, false, linkForPage),
+                    renderPaginationEllipses(),
+                    renderPaginationPageButton(pageCount, false, linkForPage))
+
+                /* If the page count is > 7 and the current page is <= 4,
```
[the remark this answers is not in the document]
melanie-exygy: Great ideas.  I made all of these changes.

browser-test/src/support/ti_dashboard.ts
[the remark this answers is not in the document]
melanie-exygy: Understood.  I split the helper method into two.

browser-test/src/trusted_intermediary.test.ts
melanie-exygy: Good catch!  I changed the test description to say there were 11 clients.

browser-test/src/trusted_intermediary.test.ts
melanie-exygy: Done.

browser-test/src/trusted_intermediary.test.ts
melanie-exygy: Added a comment about ellipses whenever `usa-pagination_overflow` is used.

PR #6701, opened 2024-02-21
Gated on feature flag, render blank page

greg-hyde:
### Description

When the north star feature flag is enabled, render a new blank(ish) page in Thymeleaf for the block edit view. This PR mostly shows the structure of a controller/java view/html view for a page

#### General

#### New Features


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -371,20 +377,26 @@ public CompletionStage<Result> previousWithApplicantId(
...
-                            programId,
-                            blockId,
-                            inReview,
-                            roApplicantProgramService,
-                            block.get(),
-                            personalInfo,
-                            ApplicantQuestionRendererParams.ErrorDisplayMode.HIDE_ERRORS,
-                            applicantRoutes,
-                            profile)));
+                ApplicationBaseView.Params applicationParams =
```
greg-hyde, line 380: This is something I assume might change down the road (or in this PR 🤷 ). We need basically the same info for the two views. It's probably worth moving ApplicationBaseView.Params into a utility class somewhere, since ApplicationBaseView is a superclass of ApplicantProgramBlockEditView and I haven't made the NorthStar view extend those classes since we don't need the utility rendering functions, so it's a little odd to reference from the north star view. Open to opinions.

server/app/views/applicant/ApplicantBaseFragment.html
```diff
@@ -0,0 +1,24 @@
+<head th:fragment="pageHeader">
```
greg-hyde, line 1: This file will probably split into more than one as time goes on. For now it's just a place for fragments that are shared across multiple pages, but the footer at least will grow into its own component at some point, and there will be more fragments like this

server/app/controllers/applicant/ApplicantProgramBlocksController.java
avaleske: At first glance I like this pattern of using a data object to pass the necessary info the the view, it keeps the method signatures a little cleaner.

otoh, it does make it a little harder to tell what's actually being used, and I don't think IntelliJ is smart enough to notice that a field on the data object is unused.

if we keep these data objects, I'd agree that moving them into a class on their own makes sense.

server/app/views/applicant/ApplicantBaseFragment.html
```diff
@@ -0,0 +1,24 @@
...
+<div th:fragment="requiredFieldsExplanation">
+  <!-- This uses utext because the parameter passed into the message is html,
+    it is a placeholder span with the red asterisk. We need it to be html and not
+    raw text to recreate the red style -->
+  <p th:utext="#{content.requiredFieldsNote(~{applicant/ApplicantBaseFragment :: requiredAsterisk})}"></p>
+</div>
+
+<th:block th:fragment="requiredAsterisk">
+    <span class="text-red-600 font-semibold"> *</span>
+</th:block>
```
avaleske, line 24: nit re newlines

server/app/views/applicant/ApplicantProgramBlockEditView.html
```diff
@@ -0,0 +1,12 @@
+<html>
```
avaleske, line 1: If we're going to call the java classes `View`s I think we should call these something else, like `ApplicantProgramBlockEditTemplate.html` to help disambiguate.

server/app/views/applicant/NorthStarApplicantBaseView.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(Request request) {
+    ThymeleafModule.PlayThymeleafContext context = playThymeleafContextFactory.create(request);
+    context.setVariable("tailwindStylesheet", assetsFinder.path("stylesheets/tailwind.css"));
+    context.setVariable("uswdsStylesheet", assetsFinder.path("dist/uswds.min.css"));
+    context.setVariable("adminJsBundle", assetsFinder.path("dist/admin.bundle.js"));
+    context.setVariable("uswdsJsInit", assetsFinder.path("javascripts/uswds/uswds-init.min.js"));
+    context.setVariable("uswdsJsBundle", assetsFinder.path("dist/uswds.bundle.js"));
+    context.setVariable("ApiDocsController", controllers.api.routes.ApiDocsController);
```
avaleske, line 35: Is this being used in the template? Should we consider developing some linting to catch unused context variables?

server/app/views/applicant/ApplicantBaseFragment.html
```diff
@@ -0,0 +1,24 @@
...
+
+<div th:fragment="requiredFieldsExplanation">
+  <!-- This uses utext because the parameter passed into the message is html,
+    it is a placeholder span with the red asterisk. We need it to be html and not
+    raw text to recreate the red style -->
+  <p th:utext="#{content.requiredFieldsNote(~{applicant/ApplicantBaseFragment :: requiredAsterisk})}"></p>
+</div>
+
+<th:block th:fragment="requiredAsterisk">
+    <span class="text-red-600 font-semibold"> *</span>
```
avaleske, line 23: It seems like this would be easier as a shared css class we can just use on a `<span class="required-asterisk"> *</span>` in each question, instead of a template fragment we pull into every question

server/app/views/applicant/NorthStarApplicantBaseView.java
greg-hyde: nope it wasn't used. I haven't looked into any linting solutions, but that's a good point

server/app/views/applicant/ApplicantBaseFragment.html
greg-hyde: Yes that would certainly be better. I actually didn't know we used any css classes directly, server/app/assets/stylesheets/styles.css seems like the only place, is that right?

avaleske: I _think_ that's the only place, outside of the USWDS classes. tailwind also has a way to wrap a few tailwind classes in an outer class.

looks like there might actually be a uswds class that applies? `usa-hint usa-hint--required` based on the form example:
https://designsystem.digital.gov/templates/form-templates/address-form/

greg-hyde: So we can do that, but we still need the template to use as a value passed into the requiredFieldsExplanation. Might be worth just using the span in other usages though

server/app/controllers/applicant/ApplicantProgramBlocksController.java
greg-hyde: This PR seems big enough already to me, so I'll plan to factor it out in a follow up, since it's referenced in a bunch of places

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
```diff
@@ -0,0 +1,12 @@
...
+    <body>
+        <div>
+            <form id="cf-block-form" th:action="${formAction}" method="POST">
+                <input hidden th:value="${csrfToken}" name="csrfToken">
+                <div th:replace="~{applicant/ApplicantBaseFragment :: requiredFieldsExplanation}"></div>
+            </form>
+        </div>
+        <footer th:replace="~{applicant/ApplicantBaseFragment :: pageFooter}"></footer>
+    </body>
+</html>
```
avaleske, line 12: **suggestion**: Can we add a newline at the end of the file?

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
```diff
@@ -0,0 +1,12 @@
+<html>
```
avaleske, line 1: **question**: This file doesn't start with `<!DOCTYPE html>`, but is it still required at the start of the file, in practice? Technically W3C says it is, but do browsers still even have a quirks mode anymore we'd need to steer them clear of?

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
```diff
@@ -0,0 +1,12 @@
...
+    <head th:replace="~{applicant/ApplicantBaseFragment :: pageHeader}"></head>
+    <body>
+        <div>
+            <form id="cf-block-form" th:action="${formAction}" method="POST">
+                <input hidden th:value="${csrfToken}" name="csrfToken">
+                <div th:replace="~{applicant/ApplicantBaseFragment :: requiredFieldsExplanation}"></div>
+            </form>
+        </div>
+        <footer th:replace="~{applicant/ApplicantBaseFragment :: pageFooter}"></footer>
```
avaleske, line 10: **question (thinking out loud, I don't actually have strong opinions here)**: With the pattern in this PR, there's a template for each page that then pulls in the header and footer fragments into itself from the `ApplicantBaseFragment`.

Why is that better then a pattern like this one?
https://www.thymeleaf.org/doc/tutorials/3.1/usingthymeleaf.html#layout-inheritance
in which there's an outer `layout` fragment that contains the header and footer, and the body content is inserted into it.

(It might be that it actually doesn't matter and we just picked the one that was simplest at first, I'm just curious if there were additional reasons.)

avaleske (approved):
Had some things I was curious about but LGTM!

Thanks for doing this!

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
greg-hyde: done

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
greg-hyde: done

server/app/views/applicant/ApplicantProgramBlockEditTemplate.html
greg-hyde: For me, when I was making this PR at least, I think it comes down to readability, but I also don't a strongly held opinion. When I look at that example in the docs you link to, it doesn't feel like it would be immediately obvious to someone new to Thymeleaf what DOM structure that would generate. The obvious advantage of that pattern is that it's easier to reuse the more pages you have (although slightly less simple when there is variance among pages, like different page titles.)

I don't think this needs to be a permanent decision, it would be pretty easy to do that as a separate refactor after the fact as we add more pages.

avaleske: sgtm!

PR #6708, opened 2024-02-22
Removing Jest from Browser Tests

gwendolyngoetz:
### Description

Removing Jest from Browser Tests and using Playwright Test directly.

This looks large but the majority of changes here are simply adding an import and prepending `test.` to functions such as `describe`, `beforeAll`, `beforeEach`, `afterEach`, and `afterAll`.

#### Verified in my Changes
- [x] All tests still pass.
- [x] Creates trace files and saves them as artifacts.
- [x] Still creates and diffs screenshots and saves them as artifacts.
- [x] Still validates, logs accessibility issues, fails the test, and saves them as artifacts.
- [x] Still creates videos and saves them as artifacts. Currently the files names are UUIDs because I can't get the running test name due to how we create the test context outside of tests. I have a plan to resolve this, but it requires some refactoring that is best done in a separate PR.

> [!NOTE]
> My hope is for videos to become less needed. The trace files can be downloaded from the artifact list, unzipped, and the run `npx playwright show-report path-to-extracted-zip-file-directory/html-output` which will start the server to view the trace. I plan on making a script similar to the `bin/sync-failed-browser-test-images` script in the future to make it less of a manual process.

> [!NOTE]
> I'll provide further instruction for the engineers and schedule a demo in the near future

#### General


browser-test/src/support/admin_questions.ts
```diff
@@ -789,10 +789,11 @@ export class AdminQuestions {
...
     }
 
-    assert(options)
-    for (let index = 0; index < options.length; index++) {
+    expect(options).toBeTruthy()
+
+    for (let index = 0; index < options!.length; index++) {
       await this.page.click('#add-new-option')
-      await this.fillMultiOptionAnswer(index, options[index])
+      await this.fillMultiOptionAnswer(index, options![index])
```
rockycodes, line 796: question (non-blocking): what does adding `!` here do?

browser-test/src/support/index.ts
```diff
@@ -90,20 +91,28 @@ function makeBrowserContext(
     // will only be used when debugging failures.
     const dirs = ['tmp/videos']
     if ('expect' in global && expect.getState() != null) {
-      const testPath = expect.getState().testPath
+      // const testPath = expect.getState().testPath
```
rockycodes, line 94: nitpick: delete commented out line

browser-test/src/support/index.ts
```diff
@@ -90,20 +91,28 @@ function makeBrowserContext(
...
       }
+
       const testFile = testPath.substring(testPath.lastIndexOf('/') + 1)
       dirs.push(testFile)
+
       // Some test initialize context in beforeAll at which point test name is
       // not set.
-      const testName = expect.getState().currentTestName
+
+      // const testName = expect.getState().currentTestName
```
rockycodes, line 107: nitpick: delete commented out line

rockycodes (approved):
lgtm! just a couple of non-blocking questions and nitpicks.

browser-test/src/support/admin_questions.ts
gwendolyngoetz: This tells the Typescript compiler to be chill with its not null check, otherwise it complains. The previous `expect` on line 792 will fail if it's null. I changed it because it was causing failures when running.

browser-test/playwright.config.ts
```diff
@@ -0,0 +1,21 @@
+import {defineConfig} from '@playwright/test'
+
+export default defineConfig({
+  // globalTimeout: 180000,
```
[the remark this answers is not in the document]
gwendolyngoetz: The globalTimeout is the length the entire run of all tests combined can be. We don't need it right now. I had it set too low and tests were failing.

Issue #6113, opened 2023-12-04
Store session id in user profile


yotommy:
**Is your feature request related to a problem? Please describe.**
To support OIDC logout, we introduced a session id to use as a key for a map containing id tokens. This is populated by a [Play filter](https://www.playframework.com/documentation/2.8.x/Filters) that sees all incoming requests. Since Play sessions are immutable, a redirect is required to add the id to the session ([reference](https://www.playframework.com/documentation/2.9.x/JavaSessionFlash#Working-with-Cookies)).

However, some routes are accessed by clients that do not support redirection, so we maintain a set of [excluded route prefixes](https://github.com/civiform/civiform/blob/5dde21b31807360d2ea1a2c22667647bc34e70a2/server/app/filters/SessionIdFilter.java#L20-L21). Twice, this has resulted in bugs because additional exclusions where required (`/playIndex`, `/metrics`). The current approach is fragile.

**Describe the solution you'd like**
The session id is required for authenticated users. Instead of storing the session id directly in the session, store it in the pac4j profile. This will eliminate the need for a redirect and will store the data in a relevant scope (rather than unscoped in the session).

yotommy:
With #6736, we get the desired behavior. However I will keep this issue open to remove the migration-only code after sufficient time has passed.

gwendolyngoetz:
Do we consider enough time to have passed?

yotommy:
Indeed, it is time to remove migration-only code.

yotommy:
This should no longer be necessary: https://github.com/civiform/civiform/blob/e7d16006f0050f5e7f8120388d08017611876969/server/app/auth/oidc/CiviformOidcLogoutActionBuilder.java#L114-L115

If we want to be super-safe, we could log a warning here and make sure that this code path isn't executing before we remove it.

PR #6736, opened 2024-02-26
Set session ID in OIDC profile

yotommy:
### Description

Store the session ID in as an OIDC profile attribute.

We store a session ID to help us retrieve the OIDC ID token at logout time. Before this PR, the session ID was added directly to the Play user session by the `SessionIdFilter`. This caused some bugs because the filter was being invoked by some requesters that didn't honor redirects. We have since encapsulated our [`NonUserRoutePrefixes`](https://github.com/civiform/civiform/blob/8314f7442c9e93ed60bb3260f2a37ed1ab71417b/server/app/filters/SessionIdFilter.java#L27), but this is still somewhat brittle, as someone can introduce a new such prefix and forget to update this class.

Since the session ID is only used for OIDC logout, we can store it as an OIDC profile attribute. That guarantees that it will only be created when needed. It also prevents the need for a [redirect](https://github.com/civiform/civiform/blob/8314f7442c9e93ed60bb3260f2a37ed1ab71417b/server/app/filters/SessionIdFilter.java#L43-L48) from the filter to just to store the ID in the session.

This PR removes the `SessionIdFilter`, since it is no longer needed.

This PR also adds some code to handle the case where user sessions already contain the session ID. In that case, we use the value from the session as the session ID in the profile. In the normal case, where the session ID is not present in the user session, we create a random session ID when the profile is created. Similarly, when we are building a logout request, we first check the profile for the session ID. If it is not present in the profile (because the profile was created before this PR was deployed), we use the value from the user session.

Key classes:
- Session ID in set in profile in the `CiviformOidcProfileCreator`. This is tested in a corresponding unit test.
- Session ID is retrieved from the profile by `CiviformOidcLogoutActionBuilder`. This is tested in a corresponding unit test.

#### General

### Issue(s) this completes

Relates to #6113


server/app/auth/oidc/CiviformOidcProfileCreator.java
```diff
@@ -162,17 +165,50 @@ protected CiviFormProfileData mergeCiviFormProfile(
...
               })
           .join();
     }
 
     return civiformProfile.getProfileData();
   }
 
-  private Optional<String> getSessionId(WebContext context) {
+  private void setSessionIdInProfile(CiviFormProfile profile, WebContext context) {
     PlayWebContext playWebContext = (PlayWebContext) context;
```
gwendolyngoetz, line 177: (question) Is it possible to recover if this somehow wasn't a PlayWebContext? If not probably just leave this, if we can maybe do some type checking and prevent the hard cast from failing.

yotommy: It is not possible to recover.

We already make this cast in a few locations: https://github.com/search?q=repo%3Aciviform%2Fciviform+%22%28PlayWebContext%29%22&type=code

gwendolyngoetz (approved):

dkatzz (approved):

PR #6762, opened 2024-02-28
Update address correction view to use 'Go back and edit' and 'Confirm Address' buttons

caitlinshk:
### Description

This PR updates the address correction view buttons if the `SAVE_ON_ALL_ACTIONS` flag is on. Instead of having "Review", "Previous", and "Save&next" like other block pages, it now has two custom buttons: "Go back and edit" and "Confirm address".
 - "Go back and edit" will take the user back to the block with the address question so they can change their address.
 - "Confirm address" will save the address the user chose and then take the user to whatever page they'd previously requested when they were on the block with the address question:
        - If they'd chosen "Review", then they'd go to the review page.
         - If they'd chosen "Previous", then they'd go to the block *before* the block with the address question.
         - If they'd chosen "Save & Next", then they'd go to the block *after* the block with the address question.

This helps simplify navigation when the `SAVE_ON_ALL_ACTIONS` flag is on.

To do this, this PR updates `AddressCorrectionBlockView` to take in an `ApplicantRequestedAction`, which specifies what button the user clicked after answering the address question (review, previous, or next). When the user clicks "Confirm address", that `ApplicantRequestedAction` is passed to the `#confirmAddress` route. That route eventually calls `renderErrorOrRedirectToRequestedPage`, which already knows how to direct the user based on the `ApplicantRequestedAction`.

Note: A lot of the diff in this PR is from moving the address correction tests out of `application_navigation.test.ts` and into their own `application_address_correction.test.ts` file. That new address correction test file is still over 1000 lines of code, so I think it's helpful to split it out into its own file.

## Release notes

**This change is protected behind the SAVE_ON_ALL_ACTIONS flag, which is not yet enabled.**

#### General

#### User visible changes

#### New Features

#### Testing when `SAVE_ON_ALL_ACTIONS` flag is off

1. Answer an address question with correction enabled with "Legit Address" as the street.
2a. Click Review -> verify you're taken to the review page and your address isn't saved.
2b. Click Previous -> verify you're taken to the previous block and your address isn't saved.
2c. Click Save&next -> verify you're taken to the address correction page. Verify the buttons on the bottom are "Review", "Previous", and "Save&next".

3. On the address correction page, choose one of the suggestions.
4a. Click Review -> verify you're taken to the review page and the suggested address you chose isn't saved.
4b. Click Previous -> verify you're taken to the previous block and the suggested address you chose isn't saved.
4c. Click Save&next -> verify you're taken to the next block. Verify the suggested address you chose *was* saved.

#### Testing when `SAVE_ON_ALL_ACTIONS` flag is on

1. Answer an address question with correction enabled with "Legit Address" as the street.
2a. Click Review -> verify you're taken to the address correction page. Verify the buttons on the bottom are "Go back and edit" and "Confirm address".
2b. Click Previous -> verify same as 2a
2c. Click Save&next -> verify same as 2b

3. On the address correction page, choose one of the suggestions.
4a. Click Confirm address -> verify you're taken to the correct page based on the button you'd chosen in step 2. Verify the suggested address you chose *was* saved.
4b. Click Go back and edit -> verify you're taken to the block with the address question. Verify your original address (not the suggested address you chose) is listed in the address question.

### Issue(s) this completes

Part of #6450

### DIFF
--- a/browser-test/src/support/applicant_questions.ts
+++ b/browser-test/src/support/applicant_questions.ts
@@ -344,7 +344,7 @@ export class ApplicantQuestions {
   }
 
   async clickConfirmAddress() {
-    await this.page.getByRole('button', { name: 'Confirm address' }).click()
+    await this.page.getByRole('button', {name: 'Confirm address'}).click()
     await waitForPageJsLoad(this.page)
   }
 
@@ -354,7 +354,7 @@ export class ApplicantQuestions {
   }
 
   async clickGoBackAndEdit() {
-    await this.page.getByRole('button', { name: 'Go back and edit' }).click()
+    await this.page.getByRole('button', {name: 'Go back and edit'}).click()
     await waitForPageJsLoad(this.page)
   }
 


caitlinshk:
* **#6762** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6762?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6762?utm_source=stack-comment).

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -833,14 +824,6 @@ private CompletionStage<Result> renderErrorOrRedirectToRequestedPage(
...
     }
     if (applicantRequestedAction == ApplicantRequestedAction.PREVIOUS_BLOCK) {
-      if (onAddressCorrectionPage) {
-        // When an applicant is on the address correction view and clicks "Previous", we want them
-        // to go back to the block with the address question. Address correction isn't a defined
-        // block in the program definition, so `blockId` represents the block with the address
-        // question and we just need to go back to that block.
-        return getBlockPage(profile, applicantId, programId, blockId, inReview, flashingMap);
-      }
-
```
caitlinshk, line 843: Wanted to give some detail on why this was removed.

Earlier, we had thought that the address correction page would still have the three classic buttons of "Review", "Previous", and "Save and next". If you click "Previous" while on the address correction page, we wanted to take the user back to the block **with** the address question, not the block **before** the address question. This if block and the `onAddressCorrectionPage` boolean was being used to handle that.

Now that we've removed the "Previous" button from the address correction page, we don't need this custom logic to handle that case. (And note that if the `SAVE_ON_ALL_ACTIONS` flag is off, we still handle this correctly in https://github.com/civiform/civiform/blob/e87bdabb0a49eabdc042edfc314b04f894adfa86/server/app/views/applicant/AddressCorrectionBlockView.java#L265-L267.)

browser-test/src/support/applicant_questions.ts
```diff
@@ -343,11 +343,21 @@ export class ApplicantQuestions {
     await waitForPageJsLoad(this.page)
   }
 
+  async clickConfirmAddress() {
+    await this.page.click('text="Confirm address"')
```
gwendolyngoetz, line 347:
```suggestion
    await this.page.getByRole('button', { name: 'Confirm address' }).click()
```

(blocking) The `page.click` method is discouraged/deprecated. Playwright considers [using locators to be the best practice](https://playwright.dev/docs/best-practices#use-locators). Prefer using one of the `page.getXXXX` methods and only fallback to `page.locator` as a last resort.

browser-test/src/support/applicant_questions.ts
```diff
@@ -343,11 +343,21 @@ export class ApplicantQuestions {
...
+    await waitForPageJsLoad(this.page)
+  }
+
   async clickEdit() {
     await this.page.click('text="Edit"')
     await waitForPageJsLoad(this.page)
   }
 
+  async clickGoBackAndEdit() {
+    await this.page.click('text="Go back and edit"')
```
gwendolyngoetz, line 357:
```suggestion
    await this.page.getByRole('button', { name: 'Go back and edit' }).click()
```

(blocking) See notes above

browser-test/src/application_address_correction.test.ts
```diff
@@ -0,0 +1,1272 @@
+import {test} from '@playwright/test'
```
gwendolyngoetz, line 1: (praise) Good idea splitting this off

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

PR #6766, opened 2024-02-28
Linting: Prefer web-first assertions

gwendolyngoetz:
### Description

Fixing eslint issues to align with playwright best practices. Rule: prefer-web-first-assertions.

Fun fact: The original code uses `page.isHidden(selector)` which returns the first match even if there are multiple. Probably not what most people would expect.

Both `page.isHidden` and `page.isVisible` are [deprecated](https://playwright.dev/docs/api/class-page#deprecated).

## Release notes

#### General

### DIFF
--- a/browser-test/src/address.test.ts
+++ b/browser-test/src/address.test.ts
@@ -54,13 +54,13 @@ test.describe('address applicant flow', () => {
         '54321',
       )
       let error = page.locator('.cf-address-street-1-error')
-      expect(await error.isHidden()).toEqual(true)
+      await expect(error).toBeHidden()
       error = page.locator('.cf-address-city-error')
-      expect(await error.isHidden()).toEqual(true)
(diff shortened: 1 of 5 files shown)


melanie-exygy (commented):
This is really helpful.  It's going to be so nice to standardize our assertions and only do things one way.  Thanks for doing this somewhat tedious work.

Could we also apply the linter rule to `.isChecked()` in places like [this](https://github.com/civiform/civiform/blob/c7404c452abeecd5b832b8f6fff59224f4a63961/browser-test/src/support/admin_settings.ts#L30)?

gwendolyngoetz:
@melanie-exygy Eventually yes, all the deprecated things should get caught. I'm just making a few changes at a time until everything is clean.

melanie-exygy (approved):

PR #6768, opened 2024-02-28
Use USWDS alert for "Please select a file" error on file upload question

caitlinshk:
### Description

If a user is on a file upload question and clicks "Save & next" without uploading a file, we display an error saying "Please select a file". This PR updates that error to use the USWDS error alert UI component.

#### General

- [x] Created unit and/or browser tests which fail without the change (if possible) -- **Existing tests should already cover this**

#### User visible changes

1. As an applicant, open a file upload question.
2. Don't upload a file, then click "Save&next".
3. Verify the USWDS error alert is shown. (I also verified that the alert is announced by my VoiceOver screen reader.)
4. Upload a file.
5. Verify the USWDS error is hidden as soon as your file is uploaded.


caitlinshk:
* **#6768** <a href="https://app.graphite.dev/github/pr/civiform/civiform/6768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/civiform/civiform/6768?utm_source=stack-comment).

server/app/assets/javascripts/file_upload.ts
```diff
@@ -139,12 +141,10 @@ function validateFileUploadQuestion(blockForm: Element): boolean {
  * but the user wants to continue to the next page.
  */
 function showFileSelectionError(
-  errorDiv: Element,
+  errorDiv: HTMLElement,
   fileInput: HTMLInputElement,
 ) {
-  errorDiv.classList.remove('hidden')
-  // Add ariaLive label so error is announced to screen reader.
-  errorDiv.ariaLive = 'polite'
```
caitlinshk, line 147: I removed this because `ViewUtils#makeAlert` adds this for me. 🎉

server/app/views/applicant/ApplicantFileUploadRenderer.java
```diff
@@ -122,10 +123,13 @@ public DivTag signedFileUploadFields(
...
     result.with(createFileInputFormElement(fileInputId, ariaDescribedByIds, hasErrors));
     result.with(
-        div(fileUploadQuestion.fileRequiredMessage().getMessage(params.messages()))
-            .withId(fileInputId + "-required-error")
-            .withClasses(
-                ReferenceClasses.FILEUPLOAD_ERROR, BaseStyles.FORM_ERROR_TEXT_BASE, "hidden"));
+        ViewUtils.makeAlertSlim(
+            fileUploadQuestion.fileRequiredMessage().getMessage(params.messages()),
+            // file_upload.ts will un-hide this error if needed.
+            /* hidden= */ true,
```
melanie-exygy, line 129: question (non-blocking):  Can you think of a way that we could include/not include this in the DOM rather than hiding it?  This might be a good candidate for HTMX.  I ask because of this guidance in the USWDS Alert docs:  

image

If you don't want to tackle this as part of this issue (which I would completely understand), would it be possible to create a Github issue?

melanie-exygy (approved):
Looks good to me.  I just had one non-blocking question.

server/app/views/applicant/ApplicantFileUploadRenderer.java
caitlinshk: Oh I didn't realize there was that guidance, good question! Would HTMX let us update the TypeScript to do "if this form has no file uploaded, go fetch the error DOM from this endpoint and add it"? If so, then HTMX seems like it would let us follow this guidance.

I would rather tackle that in a separate PR instead of this PR because we're still exploring HTMX for our application and don't have standard use patterns yet, so I think it'd be better to do once we have standard guidance. Created https://github.com/civiform/civiform/issues/6804 to track!

melanie-exygy: That makes sense.  Thanks for creating a new issue.  

As a partial answer to your question, my understanding is that HTMX would fetch the error alert from the HTML rather than the TypeScript.  I think we could do something similar to inline field validation, which is definitely possible  [with HTMX](https://htmx.org/examples/inline-validation/).

caitlinshk: Ah got it! That's a great link and seems like it'd be what we want. Thanks!

PR #6840, opened 2024-03-07
Using the USWDS pagination component on the Program Admin applications list view

### LINKED ISSUE #6690: Use USWDS pagination component for Program Admin application list
Replace the pagination component in the Program Admin application list view with the USWDS component.  Also, consider changing the current page indicator to a blue background rather than black.

#### Additional Context
In #6679, we added a utility function that creates a USWDS pagination component: [code](https://github.com/civiform/civiform/blob/ae755a1c3f6c2f5f94ee2584965d7a12b3645dc8/server/app/views/BaseHtmlView.java#L153).  We can easily use that method in ProgramApplicationListView.java, but the resulting styles don't seem to fit as well.  

#### Done when:
- [ ] The Program Admin application list uses `BaseHtmlView.renderPagination`
- [ ] Pagination is under the application list rather than above the search.
- [ ] Pagination doesn't show when there is only a single page.
- [ ] UX has approved the style changes
- [ ] The old utility function has been removed from the code:  [BaseHtmlView.renderPaginationDivOldWay](https://github.com/civiform/civiform/blob/ae755a1c3f6c2f5f94ee2584965d7a12b3645dc8/server/app/views/BaseHtmlView.java#L225)

melanie-exygy:
### Description

Making use of the USWDS pagination component for the Program Admin applications list.  This also included a couple of cosmetic changes requested by UX: 
- pagination is now under the applications list rather than above the search
- If there is only a single page of applications, the pagination component does not show

image

Manual testing is a little bit tedious if you have less than 10 applications saved for a program.  To see pagination, you'll need to create over 10 applications for a program, or rely on the browser test.

1. Log in as a Program Admin and go to the Programs page
2. Click the "Applications" link on a program.
3. If you have less than 10 applications on that program, notice that pagination isn't visible.
4. Create over 10 applications to see pagination at the bottom of the list.

### Issue(s) this completes

Fixes #6690


server/app/views/BaseHtmlView.java
```diff
@@ -220,34 +218,6 @@ protected NavTag renderPagination(int page, int pageCount, Function<Integer, Cal
...
-        div(paginationText).withClasses("leading-3", "float-left", "inline-block", "p-2", "m-4"));
-    if (pageCount > page) {
-      div.with(
-          new LinkElement().setText("→").setHref(linkForPage.apply(page + 1).url()).asButton());
-    } else {
-      div.with(new LinkElement().setText("∅").asButtonNoHref());
-    }
-    return div.with(br());
-  }
-
```
gwendolyngoetz, line 250: (praise) Deleted!

browser-test/src/support/admin_programs.ts
```diff
@@ -1188,4 +1188,8 @@ export class AdminPrograms {
     await this.page.getByText('Back').click()
     await this.page.waitForLoadState()
   }
+
+  async isPaginationVisible(): Promise<boolean> {
+    return this.page.locator('.usa-pagination').isVisible()
```
gwendolyngoetz, line 1193: (question) Are we likely to ever need two pagers on the screen?

browser-test/src/admin_application_statuses.test.ts
```diff
@@ -58,6 +58,34 @@ test.describe('view program statuses', () => {
...
+
+        // Navigate to the submitted application as the program admin.
+        await loginAsProgramAdmin(page)
+      }
+
+      // Navigate to the applications list
+      await adminPrograms.viewApplications(programWithoutStatusesName)
+
+      expect(await ctx.adminPrograms.isPaginationVisible()).toBe(true)
+      expect(page.locator('.usa-pagination__button:has-text("2")'))
```
gwendolyngoetz, line 87: (non-blocking) Consider taking a screenshot here so we can have a visual diff with the pager.

browser-test/image_snapshots/admin_application_statuses_test/application-view-with-default.png
gwendolyngoetz, line 1: (non-blocking) If it's quick and easy to do consider reducing the vertical whitespace between `test-program-with-a-default-status` and the `Applications submitted` on the left application list side.

browser-test/src/admin_application_statuses.test.ts
gwendolyngoetz: Oh if you need this test to have more than ten you may have an off by one on your loop only generating 10. Not sure if there is another application created elsewhere.

melanie-exygy: Good idea to add a screenshot since all the other tests only have one page of applications.  

For the other concern, there's a single application that's being created in the `beforeAll` [here](https://github.com/civiform/civiform/blob/898fd655c2fc493f1505222535348a4a71030ab9/browser-test/src/admin_application_statuses.test.ts#L32).  I'll add a comment about that.

browser-test/src/support/admin_programs.ts
melanie-exygy: Probably not likely, but I'll add a more specific locator just in case.

gwendolyngoetz: I'd recommend using `page.getByRole`

```typescript
await page.getByRole('navigation', { name: 'whatever-the-name' })`
```

https://playwright.dev/docs/api/class-page#page-get-by-role

gwendolyngoetz: Assuming the pager has an aria role/label on it. I don't recall what the new one has on it.

browser-test/image_snapshots/admin_application_statuses_test/application-view-with-default.png
melanie-exygy: Sure.  I've decreased the space there.  I still left a bit of space because I like it, but, hopefully, it's a compromise.  :-)

browser-test/src/admin_application_statuses.test.ts
melanie-exygy: Screenshot added and a comment about the beforeAll.

browser-test/src/support/admin_programs.ts
melanie-exygy: Done

gwendolyngoetz (approved):

PR #6864, opened 2024-03-11
Pass in sbt version to dockerfiles

gwendolyngoetz:
### Description

Our dev and prod dockerfiles thought they were setting the sbt version from the environment variable defined in `bin/lib.sh`. Turns out they were not. This adds in the dockerfile arg needed as well as passing in the sbt version in the bin/build-nnnn scripts.

I am removing the fallback value. If it can't get the value from the SBT_VERSION environment variable (which comes from `server/project/build.properties`) I want it to fail to build. Not quietly use an old version.

In order for this to work in the GitHub actions I'm updating the test.yaml to call the build scripts instead of using direct docker build commands.

#### General

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,10 +17,12 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+
       - name: Build test container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -t civiform-dev --cache-from docker.io/civiform/civiform-dev:latest ./
+        run: bin/build-dev
+
       - name: Build dev-oidc
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -195,10 +201,15 @@ jobs:
...
+
       - name: Build prod container
         env:
           DOCKER_BUILDKIT: 1
-        run: docker build -f prod.Dockerfile -t civiform:prod --cache-from docker.io/civiform/civiform:latest ./
+        run: bin/build-prod
+
+      - name: Set local prod container tag
+        run: docker tag civiform:latest civiform:prod
+
```
gwendolyngoetz, line 212: Why: `prod-simulator-compose.yml` is looking for that tag. While switching it to the local built version `civiform:latest` I'm not prepared at this moment to make sure there are no issues with it using `latest`.

Issue #6869, opened 2024-03-11
Allow jmeter tests to be run through github


dkatzz:
#### Description

Integrate with github actions to be able to run jmeter tests in the `load-test` directory. This will allow contributors to run tests without installing any tooling.

PR #6870, opened 2024-03-11
Add a github action to run jmeter test

dkatzz:
### Description

We added some jmeter tests to github, but these can currently only be run by installing jmeter and running locally. We should allow these to be run through a github action.

#### General

### Issue(s) this completes

#6869

### DIFF
--- a/.github/workflows/run_jmeter_test.yaml
+++ b/.github/workflows/run_jmeter_test.yaml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ github.event.inputs.branch_name }}
-      - name: JMeter Test
+      - name: JMeter Test - ${{ github.event.inputs.test_to_run }}
         uses: QAInsights/PerfAction@v5.6.2
         with:
           test-plan-path: load_test/${{ github.event.inputs.test_to_run }}.jmx
--- a/.github/workflows/run_jmeter_test.yaml
+++ b/.github/workflows/run_jmeter_test.yaml
@@ -28,7 +28,7 @@ jobs:
           test-plan-path: load_test/${{ github.event.inputs.test_to_run }}.jmx
           args: ""
       - name: Upload Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: jmeter-results
           path: result.jtl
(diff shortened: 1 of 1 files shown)


.github/workflows/run_jmeter_test.yaml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch_name }}
+      - name: JMeter Test
+        uses: QAInsights/PerfAction@v5.6.2
+        with:
+          test-plan-path: load_test/${{ github.event.inputs.test_to_run }}.jmx
+          args: ""
+      - name: Upload Results
+        uses: actions/upload-artifact@v3
```
gwendolyngoetz, line 31: (blocking) v4 is the current version

dkatzz: Good catch, thanks! Updated.

.github/workflows/run_jmeter_test.yaml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch_name }}
+      - name: JMeter Test
+        uses: QAInsights/PerfAction@v5.6.2
+        with:
+          test-plan-path: load_test/${{ github.event.inputs.test_to_run }}.jmx
+          args: ""
```
gwendolyngoetz, line 29: question (non-blocking): Do we need this if we aren't setting it?

gwendolyngoetz (approved):

gwendolyngoetz:
FYI, if you eventually decide to add a cron schedule the values from `github.event.inputs` will be null.

.github/workflows/run_jmeter_test.yaml
dkatzz: I think we do based on this example https://github.com/marketplace/actions/perfaction-for-jmeter#example-1-with-no-arguments?

dkatzz:
> FYI, if you eventually decide to add a cron schedule the values from `github.event.inputs` will be null.

Good point! I added a default value

gwendolyngoetz:
Defaults in the workflow dispatch section don't carry over to other entries in the `on`

gwendolyngoetz:
Yeah, nothing to update right now. Just was an FYI for the future.

PR #6875, opened 2024-03-12
Fix for TI dashboard client cards needing to only show submitted applications

### LINKED ISSUE #6834: TI Dashboard: Incorrect # of Applications Submitted
**Describe the bug**
As a TI, I can see a summary of the submitted applications per client on the TI dashboard. The summary looks to include "In Progress" applications as a part of "Submitted Applications"

**To Reproduce**

1. Login as a TI
2. Complete an application on behalf of a client
3. Start an application on behalf of a client, but don't submit it
4. See that the TI Dashboard lists 2 completed applications

**Expected behavior**
I would expect it to show the number of completed applications and the number of in progress applications separately

**Screenshots**
**1 application submitted, 1 is in progress**
image

**Shows as 2 submitted applications**
image

**Additional context**

melanie-exygy:
### Description

On the TI dashboard client cards, "# submitted applications" was showing all started applications rather than just the submitted ones.  This fixes that.

image

1. Log in as a TI.
2. Create a client if you don't already have one.
3. On the client card, click "View applications".
4. Submit an application.
5. Go back to the TI dashboard.
6. Note that the client card now says "1 application submitted" with the program name below.
7. Click "View applications" again.
8. Start an application and click "Save and Next" but don't click "Submit".
9. Go back to the TI dashboard.
10. Note that the client card still says "1 application submitted".

### Issue(s) this completes

Fixes #6834


server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -367,11 +368,16 @@ private DivTag renderCardApplications(AccountModel account) {
...
 
-    ImmutableList<ApplicationModel> newestApplicantApplications =
-        newestApplicant.get().getApplications();
-    int applicationCount = newestApplicantApplications.size();
+    ImmutableList<ApplicationModel> submittedApplications =
+        newestApplicant.get().getApplications().stream()
+            .filter(
+                application ->
+                    application.getSubmitTime() != null
+                        && application.getSubmitTime().compareTo(Instant.now()) <= 0)
```
dkatzz, line 376: **suggestion:** I think it would be helpful to add a comment here explaining this code. If an application isn't submitted, is the submit time another instant?

Not sure if this is better, but another option could be using the lifecycle stage, which may be a little clearer. Seems like any submitted application would have a lifecycle stage as active https://github.com/civiform/civiform/blob/63d2163c3b2019804751bda444d1d63676ceffd2/server/app/repository/ApplicationRepository.java#L74-L163

browser-test/image_snapshots/trusted_intermediary_test/back-link-leads-to-ti-dashboard.png
dkatzz, line 1: **question:** Do you see any diff in these? I'm trying to figure out what it is, but having trouble :)

dkatzz (commented):
Trying to figure out the diffs :) and one other idea around approach but otherwise LGTM! Thanks!

browser-test/image_snapshots/trusted_intermediary_test/back-link-leads-to-ti-dashboard.png
melanie-exygy: I don't see any diffs.  When I first created the PR, the spacing after the h1 was smaller, which I think was because I hadn't restarted the test environment locally and it was still comparing my screenshots to pre-Tailwind update screenshots.  But then, I checked in the updated snapshots, so I really don't see why Github is still acting as if there are diffs.  I honestly think it's a Github problem.  If you want, I guess I could put my changes on a new branch and create a new PR.  Pretty sure it wouldn't show image diffs in that case.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
melanie-exygy: Great idea to use the `LifecycleStage.ACTIVE`.  I've switched to using that.

PR #6905, opened 2024-03-14
Allow guest emails to be used for application updates

### LINKED ISSUE #4094: Guest account options for status updates
**As a Program Admin**, I am unable to send any automated communications (such as status updates) to applicants who come in as a _guest_. The work-around is to send emails or call the applicants to follow-up on their status. For some programs with small numbers of applications, this is do-able. For other programs, such as Fresh Bucks, they do not have the staffing capacity to do individual follow-ups to every applicant who came in as a _guest_.

**As an applicant**, I want to know the status of my application but the only way I can receive automated status updates currently, is if I create an account or log in. If I come into CiviForm as a _guest_, I will not receive automated messages about the status of my application.

A way to send automate communications (such as status updates) to applicants who do not wish to create an account or log in.

**Additional context**
@gwendolyngoetz Is looking into the statistics on the number of applicants in 2022 who submitted an application as a _guest_ and those who logged in.

Every City program asks applicants to provide an email address at some point in the application. But while we capture these email addresses, we cannot connect automated messaging to them. 

**Done when**
There is a simple way to keep _guests_ informed about their application status.

rockycodes:
### Description

This change allows guest emails to be used when sending emails in CiviForm (eg. when an application is submitted or when the application status changes). It uses the value saved in the db when an applicant answers the email question that is marked as the "Primary Applicant Info" email question, whether or not that applicant is logged in.

When an applicant is logged in and answers the primary email question with a different email address, an email is sent to both addresses.

When an applicant is logged in and answers the primary email question with the same email address they used to log in, only one email is sent to that address.

This is hidden behind the PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED feature flag.

The main logic changes are contained in `ApplicantPersonalInfo`, `ApplicantService` and `ProgramAdminApplicationView`. Most of the other files are just changed to pass down `request` for use when checking the feature flag or updating calls to `ApplicantPersonalInfo.ofGuestUser` which now requires a representation to be passed in.

#### General

#### New Features

1. Login as a CiviForm Admin and turn the `PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED` flag on
2. Create a program
3. Create a couple of statuses for the program
4. Create an email type question and toggle on "Set as universal question" and "Set answer as applicant email address"
5. Add the email question to the program and publish
6. As an applicant, apply to the program as a guest and logged in
7. As a logged in applicant, try answering the question with the same email address and a different email address to test the different cases
8. Login as a Program Admin and go to "Applications"
9. Click on one of the applications and update the status of that application
10. You should see a popup that includes the correct email address(es)

### Issue(s) this completes

Fixes #4094


browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+            emailBody,
+            programWithStatusesName,
+          )
+        }
+      })
+      test('only one email is displayed and sent for a logged in user that has answered the PAI email question with the same email they used to login', async () => {
+        const {page, applicantQuestions, adminPrograms} = ctx
+
+        await loginAsTestUser(page)
+        await applicantQuestions.applyProgram(programWithStatusesName)
```
rockycodes, line 495: This test is consistently flaky when I run in locally and times out trying to find `continue-application-button`. Any idea what's up with that?
Untitled

gwendolyngoetz: When this passes, how long did the retry take? It could be there's something that sometimes takes longer to run. If you think it's slow you can modify this test to be `test.slow('only one email....blah blah`. Tests marked as slow get three times the amount of timeout they usually get. That's where I'd start first.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+      test('both email addresses are displayed and two emails are sent for a logged in user that has answered the PAI email question with a different email', async () => {
+        const {page, applicantQuestions, adminPrograms} = ctx
+        const otherTestUserEmail = 'other@example.com'
+
+        await loginAsTestUser(page)
+        await applicantQuestions.applyProgram(programWithStatusesName)
+        await applicantQuestions.answerEmailQuestion(otherTestUserEmail)
+        await applicantQuestions.clickNext()
+        await applicantQuestions.submitFromReviewPage()
+        await logout(page)
```
gwendolyngoetz, line 451:
```suggestion
        await test.step(`Apply to program ${programWithStatusesName} as test user`, async () => {
            await loginAsTestUser(page)
            await applicantQuestions.applyProgram(programWithStatusesName)
            await applicantQuestions.answerEmailQuestion(otherTestUserEmail)
            await applicantQuestions.clickNext()
            await applicantQuestions.submitFromReviewPage()
            await logout(page)
        })
```

I recommend you use [test.step](https://playwright.dev/docs/release-notes#-steps-api-with-teststep) to group the logical parts of your test.

Further down other steps for things like 
- view application as program admin
- do the thing with the modal
- verify emails sent

Notice the 'do the thing with the modal'. I wrote that because it's not super clear exactly what is going on just glancing at it. A good label will help us know what that steps does.

gwendolyngoetz: Oh and steps can return values if that is ever needed

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+        await loginAsProgramAdmin(page)
+        await adminPrograms.viewApplications(programWithStatusesName)
+        await adminPrograms.viewApplicationForApplicant(testUserDisplayName())
+
+        const acccountEmailsBefore = supportsEmailInspection()
+          ? await extractEmailsForRecipient(page, testUserDisplayName())
+          : []
+        const applicantEmailsBefore = supportsEmailInspection()
+          ? await extractEmailsForRecipient(page, otherTestUserEmail)
+          : []
```
gwendolyngoetz, line 462: This looks like it could go inside the `if (supportsEmailInspection)` block starting on line 476.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+        await adminPrograms.viewApplicationForApplicant('Guest')
+
+        const emailsBefore = supportsEmailInspection()
+          ? await extractEmailsForRecipient(page, guestEmail)
+          : []
+
+        const modal =
+          await adminPrograms.setStatusOptionAndAwaitModal(emailStatusName)
+
+        const notifyCheckbox = await modal.$('input[type=checkbox]')
```
gwendolyngoetz, line 420: `$` is deprecated. You should switch to a [locator](https://playwright.dev/docs/locators), preferable using one of the `page.getByXxx` methods and only fall back to `page.locator` as a last resort.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+          ['Email'],
+        )
+        await adminPrograms.publishAllDrafts()
+        await logout(page)
+      })
+      test.beforeEach(async () => {
+        const {page} = ctx
+        // Logout before each test so we can start from the homescreen as an applicant
+        await logout(page)
+      })
```
gwendolyngoetz, line 391: FYI. In the future when I get fixtures going, we don't to do this thing. Each test will be isolated.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+        })
+        await adminPrograms.editProgram(programWithStatusesName)
+        await adminPrograms.editProgramBlock(
+          programWithStatusesName,
+          'block description',
+          ['Email'],
+        )
+        await adminPrograms.publishAllDrafts()
+        await logout(page)
+      })
```
gwendolyngoetz, line 386: Using `beforeAll` and `afterAll` creates a shared state between the tests which can lead to tests relying on other tests.

browser-test/src/admin/admin_application_statuses.test.ts
gwendolyngoetz: Fun fact: one reason locators are better is because if they find multiple match they error. Older ways will go with the first one found :|

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: Great callout, I'll see if that's what's happening.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: This actually has to be on it's own to capture the number of emails before we do another send.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: That makes sense and explains why I was seeing some bleed over. Would you recommend I move this logic to `beforeEach`?

browser-test/src/admin/admin_application_statuses.test.ts
gwendolyngoetz: Putting that in a `test.step('Gather emails before doing the thing you are going')` would be a boon.

gwendolyngoetz: How you can return the const variables from a step https://timdeschryver.dev/blog/keep-your-playwright-tests-structured-with-steps#steps

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: It's failing because I'm logging is as the same test user in both tests, and since the data isn't cleared between tests, I need to click "Edit" instead of "Continue" to edit the previously submitted value. It's surprising to me that it's passing on retry and in CI. I wonder if the data shared between tests is handled differently in those environments? I'm not sure how to get around this.

browser-test/src/admin/admin_application_statuses.test.ts
gwendolyngoetz: Yeah, I would let the `beforeEach` handle it. One thing to keep in mind is that way up at the top your ctx is created with clearDb=false so you'll probably have to call the method to clear the database yourself.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: I can update the test to click the correct buttons, but then there's an issue with needing to create a new status since one was already set for this user which means I need to update a bunch of snapshots.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+      test.beforeEach(async () => {
+        const {page} = ctx
+        // Logout before each test so we can start from the homescreen as an applicant
+        await logout(page)
+      })
+      test.afterAll(async () => {
+        const {page} = ctx
+        await disableFeatureFlag(
```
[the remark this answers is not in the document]
gwendolyngoetz: ctx at the top is created with clearDb=false so this probably does need to be kept

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: There were too many errors from test context carrying over so I ended up factoring this out into it's own `describe` that clears the db between tests.

browser-test/src/admin/admin_application_statuses.test.ts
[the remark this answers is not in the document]
rockycodes: This is done. I factored the tests out into their own section that clears the db every time.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -359,6 +361,171 @@ test.describe('view program statuses', () => {
...
+        const emailsBefore = supportsEmailInspection()
+          ? await extractEmailsForRecipient(page, guestEmail)
+          : []
+
+        const modal =
+          await adminPrograms.setStatusOptionAndAwaitModal(emailStatusName)
+
+        const notifyCheckbox = await modal.$('input[type=checkbox]')
+        if (!notifyCheckbox) {
+          throw new Error('Expected a checkbox input')
```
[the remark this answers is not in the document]
rockycodes: Done. I tried factoring the logic out into a helper method but it kept thinking `notifyCheckbox` was null and not allowing me to call methods on it (even when nested in the if check). So I just refactored it in all the places in this file.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: Done.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: I couldn't get the locator working so I left this with `$` for now.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: Done.

server/app/services/applicant/ApplicantService.java
```diff
@@ -587,24 +594,33 @@ CompletionStage<ApplicationModel> submitApplication(
                       .orElse(CompletableFuture.completedFuture(null));
 
               ApplicantPersonalInfo applicantPersonalInfo = applicantLabelFuture.join();
-              Optional<String> applicantEmail =
+              Optional<ImmutableSet<String>> applicantEmails =
                   applicantPersonalInfo.getType() == ApplicantType.LOGGED_IN
                       ? applicantPersonalInfo.loggedIn().email()
-                      : Optional.empty();
+                      : applicantPersonalInfo.getType() == ApplicantType.TI_PARTIALLY_CREATED
```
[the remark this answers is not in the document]
rockycodes: That's a good question. I copied this over the logic you had in your working branch, but now I'm realizing maybe there's a reason we don't want emails going to folks that TIs added? I'll check with product on this.

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -225,11 +225,11 @@ test.describe('view program statuses', () => {
           const modal =
             await adminPrograms.setStatusOptionAndAwaitModal(emailStatusName)
           const notifyCheckbox = await modal.$('input[type=checkbox]')
-          if (!notifyCheckbox) {
-            throw new Error('Expected a checkbox input')
+          expect(notifyCheckbox).not.toBeNull()
+          if (notifyCheckbox) {
+            await notifyCheckbox.uncheck()
+            expect(await notifyCheckbox.isChecked()).toBe(false)
           }
```
gwendolyngoetz, line 232:
```suggestion
          expect(notifyCheckbox).not.toBeNull()
          await notifyCheckbox.uncheck()
          expect(await notifyCheckbox.isChecked()).toBe(false)
```

Since this is now asserting that `notifyCheckbox` is not null you don't need the `if` block.

gwendolyngoetz (changes requested):

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
rockycodes: Update: We're pretty sure this is a bug. Previously, [emails would be sent](https://github.com/civiform/civiform/blame/bfb86b42f8a945535e76dafea41232f7b5f49ac2/server/app/services/applicant/ApplicantService.java#L499-L511) if there was an email present (which would be true for TI Clients if the TI adds an email). Moreover, when updating the application status for a client added by a TI, their email will show up in the modal, but attempting to send the email results in an error because that case is not accounted for in `updateStatus`.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -437,14 +455,35 @@ private DomContent renderStatusUpdateConfirmationModalEmailSection(
...
+            span(applicantNameWithApplicationId)
+                .withClasses("font-semibold", ReferenceClasses.BT_APPLICATION_ID),
             span(" of this change at "),
-            span(maybeApplicantEmail.orElse("")).withClass("font-semibold"));
+            span(emailString).withClass("font-semibold"));
+  }
+
+  private String generateEmailString(
+      Optional<String> optionalAccountEmail, Optional<String> optionalApplicantEmail) {
+    if (optionalAccountEmail.isEmpty() && optionalApplicantEmail.isPresent()) {
```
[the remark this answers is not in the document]
rockycodes: I really struggled with making this logic readable! What you wrote is much better, but we still need to account for the situation where they are the same. But maybe I could make a set and then join them or something.

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
rockycodes: Done.

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -129,17 +130,20 @@ public void setStatus(
...
       // Notify the applicant.
       ApplicantPersonalInfo personalInfo =
-          applicantService.getPersonalInfo(applicant.id).toCompletableFuture().join();
-      Optional<String> applicantEmail =
+          applicantService.getPersonalInfo(applicant.id, request).toCompletableFuture().join();
+      Optional<ImmutableSet<String>> applicantEmail =
           personalInfo.getType() == ApplicantType.LOGGED_IN
               ? personalInfo.loggedIn().email()
-              : Optional.empty();
+              : personalInfo.getType() == ApplicantType.TI_PARTIALLY_CREATED
```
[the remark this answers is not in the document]
rockycodes: Done.

server/app/views/admin/programs/ProgramApplicationView.java
```diff
@@ -411,11 +419,21 @@ private Modal renderStatusUpdateConfirmationModal(
...
-      StatusDefinitions.Status status) {
+      StatusDefinitions.Status status,
+      Request request) {
     InputTag sendEmailInput =
         input().withType("checkbox").withName(SEND_EMAIL).withClasses(BaseStyles.CHECKBOX);
-    Optional<String> maybeApplicantEmail =
+
+    Optional<String> optionalAccountEmail =
         Optional.ofNullable(application.getApplicant().getAccount().getEmailAddress());
+    Optional<String> optionalApplicantEmail = Optional.empty();
```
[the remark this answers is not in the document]
rockycodes: Done.

server/app/views/admin/programs/ProgramApplicationView.java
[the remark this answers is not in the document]
rockycodes: Done.

server/app/views/applicant/IneligibleBlockView.java
```diff
@@ -141,12 +141,14 @@ public Content render(
...
         roApplicantProgramService.getApplicantData().getApplicantName();
+    Representation.Builder representationBuilder = Representation.builder();
+
     return layout.renderWithNav(
         request,
         applicantName.isPresent()
             ? ApplicantPersonalInfo.ofLoggedInUser(
-                Representation.builder().setName(applicantName).build())
-            : ApplicantPersonalInfo.ofGuestUser(),
+                representationBuilder.setName(applicantName).build())
```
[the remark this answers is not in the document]
rockycodes: Great catch! This is not a good way to check if a user is logged in, if a guest answered the pre-seeded name question, they would also have a name present here. I fixed this.

server/app/views/applicant/PreventDuplicateSubmissionView.java
```diff
@@ -85,13 +85,14 @@ public Content render(
...
         roApplicantProgramService.getApplicantData().getApplicantName();
+    Representation.Builder representationBuilder = Representation.builder();
 
     return layout.renderWithNav(
         request,
         applicantName.isPresent()
             ? ApplicantPersonalInfo.ofLoggedInUser(
-                Representation.builder().setName(applicantName).build())
-            : ApplicantPersonalInfo.ofGuestUser(),
+                representationBuilder.setName(applicantName).build())
```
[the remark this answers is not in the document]
rockycodes: Done.

server/app/views/admin/programs/ProgramApplicationView.java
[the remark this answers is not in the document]
rockycodes: Ohh I see that now. I somehow missed the `set` part of yours. I'll try that out.

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -1616,25 +1647,77 @@ public void getPersonalInfo_applicantWithManyNames() {
...
         .isEqualTo(
             ApplicantPersonalInfo.ofLoggedInUser(
                 Representation.builder()
-                    .setEmail("test@example.com")
+                    .setEmail(ImmutableSet.of("test@example.com"))
                     .setName("First Second Third Fourth")
                     .build()));
   }
 
+  @Test
```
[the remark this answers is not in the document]
rockycodes: Done.

browser-test/src/admin/admin_application_statuses.test.ts
rockycodes: you would think that, but when i try to remove it i get an "is possible null" error 
image

gwendolyngoetz: Use `notifyCheckbox!.`

gwendolyngoetz: Or keep the `if` block for now

rockycodes: Ohh good idea! TIL about typescript's `!`. Updated.

gwendolyngoetz (approved):
You've done a fantastic jobs with the browser tests. I have not looked at anything but the tests so approving but with that caveat.

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -129,16 +128,24 @@ public void setStatus(
...
           applicantService.getPersonalInfo(applicant.id, request).toCompletableFuture().join();
-      Optional<ImmutableSet<String>> applicantEmail =
-          personalInfo.getType() == ApplicantType.LOGGED_IN
-              ? personalInfo.loggedIn().email()
-              : personalInfo.getType() == ApplicantType.TI_PARTIALLY_CREATED
-                  ? personalInfo.tiPartiallyCreated().email()
-                  : personalInfo.guest().email();
-      if (applicantEmail.isPresent()) {
-        applicantEmail
+      Optional<ImmutableSet<String>> applicantEmails;
```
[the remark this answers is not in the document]
rockycodes: Good idea. Done!

elisekalstad:
Woah!!! This is amazing!!! Nice work, @rockycodes !

PR #6929, opened 2024-03-15
DOM text reinterpreted as HTML Added encodeURIComponent  {Patch}

Shivam7-1:
By using `encodeURIComponent()`, it will prevent the risk of unintended interpretation of special characters in the URL construction. This method automatically encodes special characters, ensuring that they are treated as part of the URL string rather than being interpreted as HTML or JavaScript code. This helps prevent potential security vulnerabilities, such as injection attacks or unexpected behavior, by encoding the input as part of the URL path.


Shivam7-1:
Hi @avaleske Could You Please Review Above PR
Thanks

Issue #5510, opened 2023-09-01
Show "(optional)" indicator for optional fields 


melanie-exygy:
**Describe the problem**

Per [UX standards](https://designsystem.digital.gov/templates/form-templates/ ), optional fields should be labeled with the text "(optional)".  We have already started this on the TI page, but now need to apply this consistently throughout the application.

**We will know this is complete when:**

We have added "(optional)" to the labels in the following places:

- All form fields which are optional.  This includes fields in each block of the application process.
- Next to each optional question label in the application form.  (@sijiayam please confirm this.  In the example below, should we also add "(optional)" to the right of "what is your phone number?". 

![Pasted Graphic](https://github.com/civiform/civiform/assets/134556723/954c0c9c-61d8-4451-b008-caa90d532e3a)
- Form fields and, possibly, question labels (depending on the decision for the above item) in the CF Admin program preview.
- In the questions list on the program summary page (@sijiayam please confirm).  This refers to the page below:

![image](https://github.com/civiform/civiform/assets/134556723/226b1b37-d504-443b-9c78-eb889263d22b)
- On the search option fields of the TI page (to be confirmed):

![Pasted Graphic 1](https://github.com/civiform/civiform/assets/134556723/40654b44-5327-4e24-83ab-14136d65f527)
- In the CF Admin question edit form
- In the CF "Create new Trusted Intermediary form":

![image](https://github.com/civiform/civiform/assets/134556723/f9ba3a8c-0862-45af-b7b1-c5ab4d9e8b80)

**Note:**
Check that "optional" doesn't appear in the Program Admin application search box:
![image](https://github.com/civiform/civiform/assets/134556723/b7c6bbd3-8edc-4b82-bfef-1fd50c0b8a52)

sijiayam:
@melanie-exygy I've responded to your questions below. The scope of the  adding the (optional) indicator should touch any applicant facing form fields, admin form fields. The indicator should not affect any search or filters. 

- [ ] Next to each optional question label in the application form (yes)
- [ ] In the CF "Create new Trusted Intermediary form": (yes)
- [ ] In the CF Admin question edit form (yes)
- [ ] Questions list (no)
- [ ] (optional)` does not need to be added to any search fields or filters

ssandbekkhaug:
Do we have mocks for how the label/indicator should look in North Star?

PR #6937, opened 2024-03-15
Add new TI Client - in a new page  😸 

### LINKED ISSUE #5985: Update the add client workflow for TIs
This issue covers all the work required to shift over the "add client" workflow from its current state to the experience prototyped in [the mocks. ](https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=1%3A19&mode=design&t=ZISoVvmXCxVgUVF7-1).  Note that, as of 12/20/23, our plan for the mocks has changed.  Rather than a "New client" modal, we will have a new client page with a link to return back to the TI dashboard.  

This includes: 
1. Removing the current component 
2. Adding the "add new client" button 
3. Adding the add new client form in its own page with a return link (@Sijia mocks are TBD)
4. Adding inline field validation error messages rather than showing them in a toast notification. (all validations, including error state for the client already exists.  (This may already be completed as part of #4956 .)
5. Adding this success alert when successfully creating a client:

Image

swatkat1:
### Description

Update the add client workflow for TIs

## Release notes

Update the add client workflow for TIs

#### General

#### User visible changes

### Issue(s) this completes

Fixes #5985


browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
caitlinshk, line 1: question: Should this screenshot also be showing an error that the email is invalid? It looks like the old screenshot didn't show an error either, but I'm wondering if it maybe should?

browser-test/image_snapshots/trusted_intermediary_test/edit-client-information-with-all-fields.png
caitlinshk, line 1: question (non-blocking): Do you know why the "Email address" label is black, but the other labels are dark gray? Should they all be the same?

browser-test/image_snapshots/trusted_intermediary_test/org-members-table-many.png
caitlinshk, line 1: question (not blocking): The two table screenshots look basically the same, do you know what's changed? No problem to keep the new screenshots, just curious.

browser-test/image_snapshots/trusted_intermediary_test/ti-dashboard-chinese.png
caitlinshk, line 1: question (not blocking): Do we need a screenshot for the add new client page in Chinese?

browser-test/src/support/ti_dashboard.ts
```diff
@@ -25,24 +25,32 @@ export class TIDashboard {
...
-    await this.page.fill('label:has-text("Date of birth")', client.dobDate)
-    await this.page.click('text="Add"')
+    await this.page.getByRole('button', {name: 'Add new client'}).click()
+    await waitForPageJsLoad(this.page)
+    await this.page.fill('#email-input', client.emailAddress)
+    await this.page.fill('#first-name-input', client.firstName)
+    await this.page.fill('#middle-name-input', client.middleName)
+    await this.page.fill('#last-name-input', client.lastName)
+    await this.page.fill('#date-of-birth-input', client.dobDate)
+    if (client.notes != undefined && client.phoneNumber != undefined) {
```
caitlinshk, line 35: suggestion: Should we check the notes and phone number independently? e.g.

```
if (client.notes != undefined) {
  await this.page.fill('#ti-note-input', client.notes)
}
if (client.phoneNumber != undefined) {
  await this.page.fill('#phone-number-input', client.phoneNumber)
}
```

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -124,7 +121,41 @@ public Result dashboard(
...
+      return unauthorized();
+    }
+    Optional<TrustedIntermediaryGroupModel> trustedIntermediaryGroup =
+        accountRepository.getTrustedIntermediaryGroup(civiformProfile.get());
+    if (trustedIntermediaryGroup.isEmpty()) {
+      return notFound();
+    }
+    if (!trustedIntermediaryGroup.get().id.equals(id)) {
+      return unauthorized();
+    }
```
caitlinshk, line 136: Question (non-blocking): Do we do similar authorization checks in other methods in this controller? Is there an easy way to pull them out into a helper function that can be reused?

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -160,31 +197,49 @@ public Result addApplicant(Long id, Http.Request request) {
...
+                /* nameQuery= */ Optional.empty(),
+                /* dayQuery= */ Optional.empty(),
+                /* monthQuery= */ Optional.empty(),
+                /* yearQuery= */ Optional.empty(),
+                /* page= */ Optional.of(1)))
+        .flashing(
+            "success",
+            String.format(
+                "Successfully added new client: %s %s",
+                form.value().get().getFirstName(), form.value().get().getLastName()));
```
caitlinshk, line 238: question: It looks like Melanie's https://github.com/civiform/civiform/pull/6969 PR changes this success behavior, do you expect that Melanie's changes will go in first and then you'll update the behavior here?

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -195,26 +250,31 @@ public Result updateClientInfo(Long id, Http.Request request) throws ApplicantNo
...
-              Optional.of(form)));
+              /* request= */ request,
+              /* messages= */ messagesApi.preferred(request),
+              /* accountIdToEdit= */ Optional.of(id),
+              /* applicantIdOfTi= */ civiformProfile
+                  .get()
+                  .getApplicant()
+                  .toCompletableFuture()
+                  .join()
+                  .id,
```
caitlinshk, line 276: suggestion (non-blocking): It looks like you fetch the `applicantIdOfTi` the same way in a lot of these methods. Is it worthwhile to have a helper method that does it that can be reused?

server/app/controllers/ti/TrustedIntermediaryController.java
```diff
@@ -195,26 +250,31 @@ public Result updateClientInfo(Long id, Http.Request request) throws ApplicantNo
...
     form =
         tiService.updateClientInfo(
             form, trustedIntermediaryGroup.get(), id, messagesApi.preferred(request));
     if (form.hasErrors()) {
       return ok(
           editTiClientView.render(
-              trustedIntermediaryGroup.get(),
-              ApplicantPersonalInfo.ofLoggedInUser(
+              /* tiGroup= */ trustedIntermediaryGroup.get(),
+              /* personalInfo= */ ApplicantPersonalInfo.ofLoggedInUser(
```
caitlinshk, line 261: praise: thank you for adding the `/* param= */` everywhere, it's super helpful!

server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -74,60 +75,34 @@ public Form<AddApplicantToTrustedIntermediaryGroupForm> addNewClient(
...
-    }
-    return form;
-  }
-
-  private Optional<String> validateDateOfBirth(String dob) {
+  private Form<TiClientInfoForm> validateDateOfBirth(Form<TiClientInfoForm> form) {
+    String dob = form.value().get().getDob();
     if (Strings.isNullOrEmpty(dob)) {
-      return Optional.of("Date of Birth required");
+      return form.withError(FORM_FIELD_NAME_DOB, "Date of Birth required");
```
caitlinshk, line 81: praise: Great improvement to be able to use the `form.withError` everywhere!

server/app/views/applicant/EditTiClientView.java
```diff
@@ -54,26 +55,34 @@ public Content render(
       ApplicantPersonalInfo personalInfo,
       Http.Request request,
       Messages messages,
-      Long accountId,
-      Optional<Form<EditTiClientInfoForm>> editTiClientInfoForm) {
-
+      Optional<Long> accountIdToEdit,
+      Long applicantIdOfTi,
+      Optional<Form<TiClientInfoForm>> tiClientInfoForm) {
+    Optional<AccountModel> mayBeAccountModel = Optional.empty();
```
caitlinshk, line 61: Nitpick: We just recently decided that optionals should be named `optionalAccountModel` instead of `maybeAccountModel` - https://github.com/civiform/civiform/wiki/Development-standards#naming-of-optional-variables-and-functions-that-return-optional.

server/app/views/applicant/EditTiClientView.java
```diff
@@ -90,101 +99,115 @@ private ATag renderBackLink() {
...
-      Optional<Form<EditTiClientInfoForm>> form,
+      Optional<Form<TiClientInfoForm>> form,
       Messages messages) {
-    ApplicantData applicantData = account.newestApplicant().get().getApplicantData();
-    FormTag formTag =
-        form()
-            .withId("edit-ti")
-            .withMethod("POST")
-            .withAction(routes.TrustedIntermediaryController.updateClientInfo(account.id).url());
+    Optional<ApplicantData> mayBeApplicantData = Optional.empty();
```
caitlinshk, line 108: Similar nitpick here about naming.

server/app/views/applicant/EditTiClientView.java
```diff
@@ -90,101 +99,115 @@ private ATag renderBackLink() {
...
             FieldWithLabel.input()
-                .setId("edit-middle-name-input")
+                .setId("middle-name-input")
                 .setFieldName("middleName")
-                .setLabelText("Middle name")
-                .setValue(applicantData.getApplicantMiddleName()),
+                .setLabelText(
+                    messages.at(MessageKey.NAME_LABEL_MIDDLE.getKeyName())
+                        + " "
+                        + messages.at(MessageKey.CONTENT_OPTIONAL.getKeyName()))
```
caitlinshk, line 147: question (non-blocking): It's generally discouraged to concatenate strings together because different languages may have words in different orders. But, it also doesn't seem worthwhile to have to create a new string for "Middle name (optional)". So, this is probably fine for now?

It looks like we also want to add this (optional) indicator for applicants filling out optional questions: https://github.com/civiform/civiform/issues/5510. I'm not sure how we're going to handle internationalization there either. Are we doing to define a message that's `"{0} (optional)"` so that we can plug in any question text into the `{0}` placeholder?

@melanie-exygy do you have any thoughts here?

server/app/views/applicant/EditTiClientView.java
```diff
@@ -214,16 +237,58 @@ private DivTag renderEditClientForm(
                 .withClasses("border", "border-gray-300", "shadow-md", "w-1/2", "mt-6"));
   }
 
+  private String getDefaultDob(Optional<ApplicantData> mayBeApplicantData) {
```
caitlinshk, line 240: praise: I think these helper methods are really useful to keep the form creation easy to read!

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -130,6 +126,12 @@ public Content render(
     return layout.renderWithNav(request, personalInfo, messages, bundle, currentTisApplicantId);
   }
 
+  private ATag renderAddNewClientButton(Messages messages, Long tiGroupId) {
+    return new ATag()
+        .withHref(routes.TrustedIntermediaryController.showAddClientForm(tiGroupId).url())
+        .with(button(messages.at(MessageKey.BUTTON_ADD_NEW_CLIENT.getKeyName())));
+  }
```
caitlinshk, line 133: Question (non-blocking): Asking because I genuinely just don't know the answer: For buttons that redirect to other pages, do we prefer to put a `<button>` inside an `<a>`, or do we prefer to use `BaseHtmlView.asRedirectElement(button(), redirectUrl)`?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -220,7 +222,7 @@ private DivTag renderTIMembersTable(TrustedIntermediaryGroupModel tiGroup, Messa
                         this::renderTIRow))));
   }
 
-  private DivTag renderAddNewForm(
+  /* private DivTag renderAddNewForm(
```
caitlinshk, line 225: nitpick: Remove this commented-out code

caitlinshk (changes requested):
Overall this looks great!! Mostly non-blocking questions and praise :)

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-dob.png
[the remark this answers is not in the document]
melanie-exygy: Oh, I see that the issue doesn't have the correct link for the mocks.  That is so unfortunate.  Here are the mocks:  [TI dashboard mocks](https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=1%3A19&mode=design&t=ZISoVvmXCxVgUVF7-1).  I'll update the link in the issue, even though it's quite late for that.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -54,9 +54,20 @@ test.describe('Trusted intermediaries', () => {
...
+    await waitForPageJsLoad(page)
+
+    await page.fill('#email-input', client.emailAddress)
+    await page.fill('#first-name-input', client.firstName)
+    await page.fill('#middle-name-input', client.middleName)
+    await page.fill('#last-name-input', client.lastName)
+    await page.fill('#date-of-birth-input', client.dobDate)
+
+    await page.getByRole('button', {name: 'Save'}).click()
     await validateScreenshot(page, 'dashboard-add-client-invalid-dob')
```
melanie-exygy, line 67: nit (non-blocking):  Could we remove the word "dashboard" from the screenshot name now that the screenshot is of a separate "add client" page?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -215,7 +219,7 @@ test.describe('Trusted intermediaries', () => {
     await page.waitForSelector('h2:has-text("Edit Client")')
     await page.click('text=Back to client list')
     await waitForPageJsLoad(page)
-    await page.waitForSelector('h2:has-text("Add Client")')
+    await page.waitForSelector('h4:has-text("Search")')
     await validateScreenshot(page, 'back-link-leads-to-ti-dashboard')
```
melanie-exygy, line 223: suggestion (non-blocking):  The screenshot seems unnecessary here since we aren't testing any of the styles on the dashboard page.  The selector on line 222 should be sufficient to know that the back link is working.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -573,7 +585,7 @@ test.describe('Trusted intermediaries', () => {
     await tiDashboard.clickOnViewApplications()
     await page.click('#ti-clients-link')
 
-    expect(await page.innerText('#add-client')).toContain('Add Client')
+    expect(await page.innerText('#ti-search-form')).toContain('Search by name')
```
melanie-exygy, line 588: suggestion (non-blocking):  rather than verifying the text, you could verify that the innerHtml has the id `name-search`.  This is so that we don't have to update this test if the text ever changes.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: I changed it!. Please let me know. Even I am not sure!

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: It a race condition 🐎

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-clients-no-email.png
melanie-exygy, line 1: suggestion (blocking):  Part of the issue was to replace the success toast.  I just noticed that the issue still said to use a success modal but the UX mocks have changed and now we should show a success banner.  I updated the issue with a screenshot of the banner from the mocks.  Personally, I'd be fine with the success alert being a fast-follow, especially since the issue was unfortunately never updated with the recent UX plans.

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: Most likely she will go first and I will merge my changes.

server/app/controllers/ti/TrustedIntermediaryController.java
swatkat1: Moved to a helper. :)

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
melanie-exygy, line 1: question (blocking):  Should there be a validation error message here?

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-dob.png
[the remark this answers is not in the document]
swatkat1: Hi @melanie-exygy, I thought this should be simple but I think we should do it as a fast followup. The USWDS needs a few things before moving over - 1. Ability to handle a list of field errors (DOB can have errors like - Date of Birth should be less than 150 years ago , date in should be in the past etc.) 2. They should have "required" field markers.  I started tackling this but realized , it needs more support before switching over.

browser-test/src/support/ti_dashboard.ts
swatkat1: Sure!

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
swatkat1: Hi Melanie, per my analysis- https://github.com/civiform/civiform/pull/6937#discussion_r1532776458 This error is not getting captured in the Screenshots too :(

swatkat1: Also this is not getting captured in screemshots.

server/app/views/applicant/EditTiClientView.java
melanie-exygy: @caitlinshk I think concatenation is fine here because "(optional)" is a separate phrase or indicator, so even in another language, the placement should be the same.  For the applicant optional questions, using `{0} (optional)` would be fine, although I don't see the advantage of that over concatenating, honestly.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
melanie-exygy: This is going to be a USWDS button (using their button CSS classes), so we'll need it to be a button element.

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
melanie-exygy: Understood.  Since the error is showing for users, that's what counts most.  Thanks for looking into it!  I think we can go without the screenshot capturing it.

browser-test/image_snapshots/trusted_intermediary_test/edit-client-information-with-all-fields.png
swatkat1: the email input is html email type input while other are just plain text input. May be the difference is there.. I didn't adjust the form nor the field settings.

browser-test/image_snapshots/trusted_intermediary_test/ti-dashboard-chinese.png
swatkat1: Done!

browser-test/image_snapshots/trusted_intermediary_test/org-members-table-many.png
swatkat1: Not too sure.

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-clients-no-email.png
swatkat1: Should i get rid of this toast message fully? I have added Success Alerts on Add Client page

server/app/views/applicant/TrustedIntermediaryDashboardView.java
swatkat1: @melanie-exygy  - Do you have an example in the codebase?

swatkat1: Changed it to BaseHtmlView.redirectButton(id,text,url)--> which returns a ButtonTag. Please let me know if it works.

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-clients-no-email.png
swatkat1: Spoke offline- I have added the success alert but some work is needed and will be addressed as a follow-up item.

browser-test/image_snapshots/trusted_intermediary_test/add-clients-no-email.png
melanie-exygy, line 1: suggestion (blocking):  So sorry, I'm just noticing now that the "Add new client" button is overlapping the horizontal line.  It should be in line with the "All Clients" header, like this:
image

browser-test/image_snapshots/trusted_intermediary_test/add-client-invalid-dob.png
melanie-exygy, line 1: suggestion (blocking):  Phone number and Notes should also show "(optional)".

server/app/views/applicant/TrustedIntermediaryDashboardView.java
melanie-exygy: I thought that it needed to be a button to apply the USWDS button classes, but it doesn't.  Here's an example from the client cards:  [link styling example](https://github.com/civiform/civiform/blob/4902bcd1793739a9add7d59fc358aabb2c65976c/server/app/views/applicant/TrustedIntermediaryDashboardView.java#L445)

melanie-exygy: Since in this case, the mocks are showing a blue button, you'd just need `.withClasses("usa-button")`.  I'd suggest using an ATag rather than a button because otherwise you'll have to fight against our default button styles to be able to apply the USWDS one.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -59,9 +59,20 @@ test.describe('Trusted intermediaries', () => {
...
+
+    await page.getByRole('button', {name: 'Save'}).click()
+    await validateScreenshot(page, 'add-client-invalid-dob')
+    await tiDashboard.gotoTIDashboardPage(page)
+    await waitForPageJsLoad(page)
     await tiDashboard.expectDashboardNotContainClient(client)
-    await validateScreenshot(page, 'dashboard-add-client-invalid-dob')
   })
 
   test('expect Dashboard Contain New Client', async () => {
```
melanie-exygy, line 78: suggestion (blocking):  I don't see the add client success alert in any of the snapshots.  Could you add it in this test or one of the others where we add a new client?

server/app/views/applicant/EditTiClientView.java
```diff
@@ -57,30 +57,39 @@ public Content render(
...
-      Long accountId,
-      Optional<Form<EditTiClientInfoForm>> editTiClientInfoForm,
-      boolean isSuccessfulSave) {
-
+      Optional<Long> accountIdToEdit,
+      Long applicantIdOfTi,
+      Optional<Form<TiClientInfoForm>> tiClientInfoForm) {
+    Optional<AccountModel> optionalAccountModel = Optional.empty();
+    String title = "Add new Client";
+    String pageHeader = "Add client";
```
melanie-exygy, line 65: suggestion (blocking):  Could you please use the localized message string here and also for the "Edit client" page header text on line 72?

server/app/views/applicant/EditTiClientView.java
```diff
@@ -57,30 +57,39 @@ public Content render(
...
             .addMainContent(
                 renderHeader(tiGroup.getName(), "py-12", "mb-0", "bg-gray-50"),
                 hr(),
-                renderSubHeader("Edit client").withId("edit-client").withClass("my-4"),
+                renderSubHeader(pageHeader).withId(pageId).withClass("my-4"),
                 renderBackLink(),
-                renderSuccessAlert(
-                    isSuccessfulSave,
-                    messages.at(MessageKey.BANNER_CLIENT_INFO_UPDATED.getKeyName())),
+                renderSuccessAlert(isSuccessfulSave, successToast),
```
melanie-exygy, line 86: suggestion (blocking):  The success alert styles for adding a new client are different than for editing.  Please use `ViewUtils.makeAlert()` rather than `ViewUtils.makeAlertSlim()` for the new client alert.  There's also more text, which you'll find in our `messages.conf` file.  
image

browser-test/image_snapshots/trusted_intermediary_test/add-clients-no-email.png
swatkat1: updated

server/app/views/applicant/EditTiClientView.java
swatkat1: sure!

swatkat1: There is no title "Edit client information". There is no page header for either. I see this messages.at(MessageKey.TITLE_CREATE_CLIENT.getKeyName()); so updated this one.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -38,13 +38,38 @@ test.describe('Trusted intermediaries', () => {
...
-    await tiDashboard.expectSuccessAlert()
+    await tiDashboard.expectSuccessAlertOnUpdate()
     await validateScreenshot(page.locator('main'), 'edit-client-success-alert')
 
     await page.click('#ti-dashboard-link')
     await waitForPageJsLoad(page)
     await tiDashboard.expectDashboardContainClient(updatedClient)
   })
+  test('verify success toast screenshot on adding new client', async () => {
+    const {page, tiDashboard} = ctx
```
caitlinshk, line 49: nitpick: Gwen recently showed us how to include the `{page, tiDashboard}` in the test line:

`test('verify success toast screenshot on adding new client', async ({page, tiDashboard}) => {`

@gwendolyngoetz  do we want to start using that convention for all new tests?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -305,7 +325,7 @@ test.describe('Trusted intermediaries', () => {
       lastName: 'last',
       dobDate: '2023-07-11',
     }
-    await page.getByRole('button', {name: 'Add new client'}).click()
+    await page.locator('#add-new-client').click()
```
caitlinshk, line 328: Nitpick: Is it worth defining a helper method for clicking the "Add new client" button since we do it in multiple tests?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -330,9 +385,13 @@ test.describe('Trusted intermediaries', () => {
   test('dashboard contains required indicator note and optional marker', async () => {
```
caitlinshk, line 385: Nitpick: Should this test name be changed to not have "dashboard"?

server/app/views/applicant/EditTiClientView.java
caitlinshk: Sounds good!

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
caitlinshk: Understood! Maybe we don't need the screenshot then if the error isn't being shown? Maybe we just want an `expect` statement that expects that we're still on the add client page?

caitlinshk:
Left a few more nitpicks and resolved old comments. Will let Melanie do the approval since Melanie is super familiar with all the TI work!

browser-test/src/trusted_intermediary.test.ts
gwendolyngoetz: I committed support for using these custom fixtures earlier this week, but won't have time until tonight or this weekend to update the wiki and provide more detailed instructions on how to migrate.

It's fine to use them going forward, but it's best to convert an entire file all at once. Otherwise you run the risk of using the page from ctx (old way) with the page as a fixture (new way) which could result it weird test results.

Converting the entire file should be done as a separate PR on test files that are very large too.

browser-test/src/trusted_intermediary.test.ts
gwendolyngoetz: @caitlinshk summoned me on this PR so everything I say is her fault.

Don't change these locators. Playwright best practices are to use `page.getByRole` first, other `page.getByXxx` methods second, and only fallback to element id as a last resort.

The reason being that Playwright looks up elements based on w3c accessibility name which is akin to how assistive technologies see the elements and it makes for less fragile tests.

gwendolyngoetz (changes requested):

caitlinshk:
> Don't change these locators. Playwright best practices are to use page.getByRole first, other page.getByXxx methods second, and only fallback to element id as a last resort.

I think Swathi had to change this because the page is translated, so the text "Add new client" won't appear on this page when it's in a different language.

gwendolyngoetz:
> I think Swathi had to change this because the page is translated, so the text "Add new client" won't appear on this page when it's in a different language.

If it's to get a screenshot in a different language navigate to the page you want, change the language from the dropdown, and then do the screenshot

browser-test/src/trusted_intermediary.test.ts
swatkat1: I tried getByRole() but this is not a button but an ATag. I am not sure how to get it. I checked https://playwright.dev/docs/locators#locate-by-role  doesn't seem to have it. @gwendolyngoetz and @caitlinshk  - Do you know?

browser-test/src/trusted_intermediary.test.ts
swatkat1: Ahh!! Dashboard is coming back to kick me!!!

browser-test/src/trusted_intermediary.test.ts
gwendolyngoetz: If it's an `<a>` you can use the role 'link'

swatkat1: Thanks @gwendolyngoetz

melanie-exygy (approved):
Thanks for sticking with this big chunk of work, @swatkat1 !  It's looking really good and is such a cleaner user experience.

caitlinshk (approved):

caitlinshk (approved):

gwendolyngoetz (approved):

browser-test/image_snapshots/trusted_intermediary_test/dashboard-add-client-invalid-email.png
swatkat1: Ahh make sense. Will add it.

swatkat1: Edit- I had updated the test to check the screenshot and go back to the TIDashboard to make sure the Client is not in the list. Please let me know if this coverage is enough.

caitlinshk: That sounds good!

Issue #6510, opened 2024-01-30
Detect when Docker is running out of disk space


nb1701:
When running a local dev instance (run-dev or browser tests), if docker runs out of disk space, the database container will fail to start properly.  The failure signature is 
```
PersistenceException: SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 30001ms.
```
But this isn't very clear that the problem is that Docker does not have any more disk space in which to allocate the volume needed for the postgres container.

We should have the script either detect this error condition and print a message saying what the problem probably is.  Or, it should detect that Docker is about to run out of available disk space, and offer to do `docker image prune` for you.

dkatzz:
We should ensure this works for codespaces as well.

PR #6951, opened 2024-03-18
Add docker disk space check for MacOS

nb1701:
This adds a docker disk space check for MacOS. Because it is not straightforward to check this on Docker Desktop for Mac, it checks the settings.json file to find the max disk space allocation, and compares against the output of `docker system df`. This isn't entirely accurate, because there are some overlapping disk space usage (e.g. images that use common layers), but it's close enough. This will be more straightforward for Linux if we decide to implement it there.

Addresses https://github.com/civiform/civiform/issues/6510

### DIFF
--- a/bin/lib/docker.sh
+++ b/bin/lib/docker.sh
@@ -265,7 +265,7 @@ function docker::check_disk_space() {
   # close enough for detecting if we're close to running out of space.
   settings_json="${HOME}/Library/Group Containers/group.com.docker/settings.json"
   if [[ -f "${settings_json}" && "$(uname -s)" == "Darwin" ]]; then
-    max_size=$(jq '.diskSizeMiB' "$settings_json" | awk '{print $1 * 1024 * 1024}')
+    max_size=$(jq '.diskSizeMiB * 1024 * 1024' "$settings_json")
     used=0
     while IFS= read -r size; do
       # Extract size and unit from the size string


gwendolyngoetz (changes requested):

bin/lib/docker.sh
```diff
@@ -256,3 +256,48 @@ function docker::dev_and_test_server_sbt_command() {
...
+
+function docker::check_disk_space() {
+  # TODO: Figure out the right way to do this on Linux.
+  # Should be easier because it uses the actual file system.
+  # Note that this is not really accurate for MacOS because it's
+  # summing up a list of things that might have some overlap, but it's
+  # close enough for detecting if we're close to running out of space.
+  settings_json="${HOME}/Library/Group Containers/group.com.docker/settings.json"
+  if [[ -f "${settings_json}" && "$(uname -s)" == "Darwin" ]]; then
+    max_size=$(jq '.diskSizeMiB' "$settings_json" | awk '{print $1 * 1024 * 1024}')
```
gwendolyngoetz, line 268:
```suggestion
    max_size=$(jq ".diskSizeMiB * 1024 * 1024" "$settings_json")
```

gwendolyngoetz (approved):
I still think trying to do this if docker compose failed to become healthy, but that doesn't appear to be as easy as checking the return code :/

bin/lib/docker.sh
gwendolyngoetz: @nb1701  this is completely optional

PR #6984, opened 2024-03-20
Updating header tests to use custom civiform fixtures

gwendolyngoetz:
### Description

Extends Playwright test with custom fixtures and updates the header tests to use them.

#### General

### DIFF
--- a/browser-test/playwright.config.ts
+++ b/browser-test/playwright.config.ts
@@ -1,4 +1,5 @@
 import {defineConfig} from '@playwright/test'
+import { BASE_URL } from './src/support/config'
 
 // For details see: https://playwright.dev/docs/api/class-testconfig
 
@@ -27,6 +28,8 @@ export default defineConfig({
   use: {
     trace: 'on-first-retry',
     video: process.env.RECORD_VIDEO ? 'on-first-retry' : 'off',
+    // Fall back support config file until it is removed
+    baseURL: process.env.BASE_URL || BASE_URL // 'http://civiform:9000'
   },
   reporter: [
     ['list', {printSteps: true}],
--- a/browser-test/playwright.config.ts
+++ b/browser-test/playwright.config.ts
@@ -1,5 +1,5 @@
 import {defineConfig} from '@playwright/test'
-import { BASE_URL } from './src/support/config'
+import {BASE_URL} from './src/support/config'
 
 // For details see: https://playwright.dev/docs/api/class-testconfig
 
@@ -29,7 +29,7 @@ export default defineConfig({
     trace: 'on-first-retry',
     video: process.env.RECORD_VIDEO ? 'on-first-retry' : 'off',
     // Fall back support config file until it is removed
-    baseURL: process.env.BASE_URL || BASE_URL // 'http://civiform:9000'
+    baseURL: process.env.BASE_URL || BASE_URL, // 'http://civiform:9000'
   },
   reporter: [
     ['list', {printSteps: true}],
--- a/browser-test/src/header.test.ts
+++ b/browser-test/src/header.test.ts
@@ -1,56 +1,59 @@
-import {test, expect} from '@playwright/test'
+import {test, expect} from './support/civiform_fixtures'
 import {
-  createTestContext,
   loginAsTestUser,
   validateScreenshot,
   validateAccessibility,
+  logout,
 } from './support'
 
-test.describe('Header', () => {
-  const ctx = createTestContext(/* clearDb= */ false)
+test.describe('Header', {tag: ['@uses-fixtures']}, () => {
+  /**
+   * @todo (#4360) add a "Not logged in, guest mode disabled" test once we can get to the programs page without logging in, for an entity without guest mode.
+   */
+  test('Check screenshots and validate accessibility', async ({page}) => {
+    await test.step('Take a screenshot as a guest', async () => {
+      await logout(page)
+      await validateScreenshot(
+        page.getByRole('navigation'),
+        'not-logged-in-guest-mode-enabled',
+      )
+    })
 
-  test('Not logged in, guest mode enabled', async () => {
-    const {page} = ctx
-    await validateScreenshot(
-      page.getByRole('navigation'),
-      'not-logged-in-guest-mode-enabled',
+    await test.step('Take a screenshot as the test user', async () => {
+      await loginAsTestUser(page)
+      await validateScreenshot(page.getByRole('navigation'), 'logged-in')
+    })
+
+    await test.step('Passes accessibility test', async () => {
+      await validateAccessibility(page)
+    })
+  })
+
+  test('Government banner', async ({page}) => {
+    const usaBannerLocator = page.getByRole('banner').locator('.usa-banner')
+    const usaBannerContentLocator = usaBannerLocator.locator(
+      '.usa-banner__content',
     )
-  })
+    const usaBannerButtonLocator = usaBannerLocator.getByRole('button', {
+      name: "Here's how you know",
+    })
 
-  // TODO(#4360): add a "Not logged in, guest mode disabled" test once we
-  // can get to the programs page without logging in, for an entity without
-  // guest mode.
+    await test.step('Page loads with the banner visible and collapsed', async () => {
+      await expect(usaBannerLocator).toContainText(
+        'This is an official government website.',
+      )
+      await expect(usaBannerContentLocator).toBeHidden()
+    })
 
-  test('Logged in', async () => {
-    const {page} = ctx
-    await loginAsTestUser(page)
-    await validateScreenshot(page.getByRole('navigation'), 'logged-in')
-  })
+    await test.step('Clicking the button expands the banner', async () => {
+      await usaBannerButtonLocator.click()
+      await expect(usaBannerContentLocator).toBeVisible()
+      await validateScreenshot(page.getByRole('navigation'), 'banner-expanded')
+    })
 
-  test('Passes accessibility test', async () => {
-    const {page} = ctx
-    await validateAccessibility(page)
-  })
-
-  test('Displays the government banner', async () => {
-    const {page} = ctx
-    expect(await page.textContent('section')).toContain(
-      'This is an official government website.',
-    )
-  })
-
-  test('Government banner expands when clicked and closes when clicked again', async () => {
-    const {page} = ctx
-    // The banner is initially closed
-    await expect(page.locator('.usa-banner__content')).toBeHidden()
-    // Click to expand the banner
-    await page.click('.usa-banner__button')
-
-    await expect(page.locator('.usa-banner__content')).toBeVisible()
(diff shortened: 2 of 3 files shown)


browser-test/playwright.config.ts
```diff
@@ -27,6 +28,8 @@ export default defineConfig({
   use: {
     trace: 'on-first-retry',
     video: process.env.RECORD_VIDEO ? 'on-first-retry' : 'off',
+    // Fall back support config file until it is removed
+    baseURL: process.env.BASE_URL || BASE_URL // 'http://civiform:9000'
```
gwendolyngoetz, line 32: Long term, settings will get migrated over here. The baseURL in particular should be set here as Playwright will automatically use that as the default.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
-import {test, expect} from '@playwright/test'
+import {test, expect} from './support/civiform_fixtures'
```
gwendolyngoetz, line 1: This file extends the built-in Playwright test allowing the use of custom fixtures.

Instead of 

```typescript
test('Check screenshots and validate accessibility', async () => {
  const page = {ctx}
  // do stuff
}
```

We just tell the test what we want. Basically DI.

```typescript
test('Check screenshots and validate accessibility', async ({page}) => {
  // do stuff
}
```

Eventually everything will use this and not the default playwright one.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
-
-  // TODO(#4360): add a "Not logged in, guest mode disabled" test once we
-  // can get to the programs page without logging in, for an entity without
-  // guest mode.
-
-  test('Logged in', async () => {
-    const {page} = ctx
-    await loginAsTestUser(page)
-    await validateScreenshot(page.getByRole('navigation'), 'logged-in')
+test.describe('Header', {tag: ['@uses-fixtures']}, () => {
```
gwendolyngoetz, line 9: tag `@uses-fixtures` will be used to mark which files have been migrated over and will go away once everything is done.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
+  /**
+   * @todo (#4360) add a "Not logged in, guest mode disabled" test once we can get to the programs page without logging in, for an entity without guest mode.
+   */
```
gwendolyngoetz, line 12: JSDoc. We should use it where it makes sense https://jsdoc.app/

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
+    })
+
+    await test.step('Take a screenshot as the test user', async () => {
+      await loginAsTestUser(page)
+      await validateScreenshot(page.getByRole('navigation'), 'logged-in')
+    })
+
+    await test.step('Passes accessibility test', async () => {
+      await validateAccessibility(page)
+    })
```
gwendolyngoetz, line 29: Do more asserts in a single test. It's recommended by Playwright and it shaved 3.5s off this file 

image

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
-
-    await expect(page.locator('.usa-banner__content')).toBeVisible()
-    await validateScreenshot(page.getByRole('navigation'), 'banner-expanded')
-    // Click again to close the banner
-    await page.click('.usa-banner__button')
-
-    await expect(page.locator('.usa-banner__content')).toBeHidden()
+    const usaBannerButtonLocator = usaBannerLocator.getByRole('button', {
+      name: "Here's how you know",
+    })
```
gwendolyngoetz, line 39: Saving a reference to the parent-est locator and doing searches on that means we don't have to traverse the full page DOM every time. Makes it easier to read and can be better for performance.

gwendolyngoetz: Using `page.getByRole` or other `page.getByXxx` methods is the preferred way to look up things, `getByRole` especially. This is because it's less fragile to look for things the way accessibility tools would.

browser-test/src/support/civiform_fixtures.ts
```diff
@@ -0,0 +1,91 @@
...
+    await request.post('/dev/seed/clear')
+    await page.goto('/programs')
+    await waitForPageJsLoad(page)
+    await page.locator('#warning-message-dismiss').click()
+
+    // Run the Test
+    await use(page)
+
+    // AfterEach
+    // - none -
```
gwendolyngoetz, line 87: The `use` call is the handoff to a single test itself. Comments are here to indicate the flow, with aftereach left for illustrative purposes.

browser-test/src/support/civiform_fixtures.ts
```diff
@@ -0,0 +1,91 @@
...
+  },
+})
+
+export {expect} from '@playwright/test'
```
gwendolyngoetz, line 91: Re-export this so we only need to import test and expect on a single line. Plus it will allow for extending expect down the road without having to go back and update every file.

browser-test/src/support/civiform_fixtures.ts
```diff
@@ -0,0 +1,91 @@
...
+  adminQuestions: AdminQuestions
+  adminProgramStatuses: AdminProgramStatuses
+  applicantQuestions: ApplicantQuestions
+  adminPredicates: AdminPredicates
+  adminTranslations: AdminTranslations
+  adminProgramImage: AdminProgramImage
+  applicantFileQuestion: ApplicantFileQuestion
+  tiDashboard: TIDashboard
+  adminTiGroups: AdminTIGroups
+}
```
gwendolyngoetz, line 29: Each class used on this is considered a fixture class

browser-test/src/support/civiform_fixtures.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const test = base.extend<CiviformFixtures>({
+  adminApiKeys: async ({page}, use) => {
+    await use(new AdminApiKeys(page))
+  },
```
gwendolyngoetz, line 34: Set up an individual fixture

browser-test/src/support/civiform_fixtures.ts
```diff
@@ -0,0 +1,91 @@
...
+  adminTiGroups: async ({page}, use) => {
+    await use(new AdminTIGroups(page))
+  },
+
+  page: async ({page, request}, use) => {
+    // BeforeEach
+    await request.post('/dev/seed/clear')
+    await page.goto('/programs')
+    await waitForPageJsLoad(page)
+    await page.locator('#warning-message-dismiss').click()
```
gwendolyngoetz, line 81: Always do these before running each test.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
-    const {page} = ctx
-    await validateAccessibility(page)
-  })
-
-  test('Displays the government banner', async () => {
-    const {page} = ctx
-    expect(await page.textContent('section')).toContain(
-      'This is an official government website.',
+  test('Government banner', async ({page}) => {
+    const usaBannerLocator = page.getByRole('banner').locator('.usa-banner')
```
gwendolyngoetz, line 33: Technically there *should* only be a single banner role on a page. This will fail if there is more than one banner role. Some work needs to be done in the rendering side of things to make it more clear the banner (that is the first <header>)  and it's child nav have more meaningful aria settings.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
...
+  /**
+   * @todo (#4360) add a "Not logged in, guest mode disabled" test once we can get to the programs page without logging in, for an entity without guest mode.
+   */
+  test('Check screenshots and validate accessibility', async ({page}) => {
+    await test.step('Take a screenshot as a guest', async () => {
+      await logout(page)
+      await validateScreenshot(
+        page.getByRole('navigation'),
+        'not-logged-in-guest-mode-enabled',
+      )
```
gwendolyngoetz, line 19: I hate the `bin/fmt` typescript wrapping settings.

browser-test/src/header.test.ts
```diff
@@ -1,56 +1,59 @@
-import {test, expect} from '@playwright/test'
+import {test, expect} from './support/civiform_fixtures'
 import {
-  createTestContext,
   loginAsTestUser,
   validateScreenshot,
   validateAccessibility,
+  logout,
```
gwendolyngoetz, line 6: long term these individual methods will get grouped into logical fixtures of their own

browser-test/playwright.config.ts
```diff
@@ -27,6 +28,8 @@ export default defineConfig({
   use: {
     trace: 'on-first-retry',
     video: process.env.RECORD_VIDEO ? 'on-first-retry' : 'off',
+    // Fall back support config file until it is removed
+    baseURL: process.env.BASE_URL || BASE_URL, // 'http://civiform:9000'
```
gwendolyngoetz, line 32: Long term, settings will get migrated over here. The baseURL in particular should be set here as Playwright will automatically use that as the default.

browser-test/src/header.test.ts
rockycodes: Nice!

rockycodes (approved):
These comments are soooo helpful. This looks good to me. I recommend sharing this PR out with the whole team for context/example of how to use some of the new options for browser tests.

browser-test/src/header.test.ts
melanie-exygy: I like how much cleaner this is.

browser-test/src/header.test.ts
melanie-exygy: Smart!

browser-test/src/header.test.ts
melanie-exygy: I didn't know very much about the banner role till just now.  If I'm reading these docs correctly, [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role), wouldn't the parent `header` element that contains both the "This is a gov website" `section` and the `div` that represents our CF header be the one with the banner role?

browser-test/src/header.test.ts
melanie-exygy: That's really good to know.  Can you help me understand why "name" is used?  I don't see a `name` attribute on this element.

melanie-exygy (approved):

browser-test/src/header.test.ts
gwendolyngoetz: @melanie-exygy that's my understanding, but it assumes no one manually sets the role of banner on something else.

browser-test/src/header.test.ts
gwendolyngoetz: @melanie-exygy Name is the accessible name defined by w3c specs. Playwright has a good visual example [here](https://playwright.dev/docs/locators#locate-by-role) with more details.

PR #7009, opened 2024-03-22
Add "Clear search" button and display number of search items on TI client search

### LINKED ISSUE #6509: Add button to clear search parameters on the TI dashboard client search
We would like to add a button which allows users to clear search parameters on the TI dashboard client search.

#### Link to Figma mocks: 
[mocks](https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=843-5233&mode=design&t=hVhnoT2n3WeX8UFe-0)

image

#### Additional Context
This is related to issue #6214. 

#### Done when:
- [ ] Clicking the "Clear search" link clears the search parameters for the Name and DOB fields.

melanie-exygy:
### Description

Adding a quick way to clear the search fields on the TI dashboard by clicking on "Clear search".  Also, displaying the number of search items found.  If no search filter is applied, the message will say "Displaying all clients".  If there are validation errors in the search input fields, we don't show the "Displaying # clients" message.

I also threw in a couple of corrections to some i18n message strings so that they match the UX mocks.

image

1. Log in as a TI.
2. Add a few clients if you don't already have any.
4. Notice "Displaying all clients" under the search.
5. Search for a client by name and/or date.
6. Notice that it now says "Displaying 1 client".
7. Edit 2 clients so they have the same birthday.
8. Search by that birthday.
9. Note that it now says "Displaying 2 clients".
10. Click "Clear search" to empty the search fields.
11. Enter a month, but no other search terms and click "Search".
12. Note that the "Displaying ... clients" message doesn't show.

### Issue(s) this completes

Fixes #6509


browser-test/src/trusted_intermediary.test.ts
```diff
@@ -408,7 +408,7 @@ test.describe('Trusted intermediaries', () => {
       firstName: 'first2',
       middleName: 'middle',
       lastName: 'last2',
-      dobDate: '2021-11-07',
+      dobDate: '2021-12-07',
```
azhangggggg, line 411: question (non-blocking): curious about the purpose of changing this TI's dob?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -493,39 +511,13 @@ test.describe('Trusted intermediaries', () => {
...
     await waitForPageJsLoad(page)
 
+    await expect(
+      page.getByRole('heading', {name: 'Displaying 1 client'}),
+    ).toBeVisible()
     await tiDashboard.expectDashboardContainClient(client1)
     await tiDashboard.expectDashboardNotContainClient(client2)
   })
 
-  test('empty search parameters returns all clients', async () => {
```
azhangggggg, line 500: Thanks for adding this button to clear search. Question: What is the expected behavior when search with empty search parameters?

melanie-exygy: The expected behavior is to display all clients in an unfiltered list.

melanie-exygy: I removed this test because I added that test scenario in the 'search For Client In TI Dashboard' test (see line 442).  I consolidated the two tests for better test performance, as Gwen suggested.

azhangggggg: I see. makes sense. thanks.

azhangggggg (approved):

PR #7027, opened 2024-03-23
DOM text reinterpreted as HTML {Patch}

Shivam7-1:
Patch issue link:  https://issuetracker.google.com/issues/330971625

By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML. Always be cautious when dealing with user input or dynamic content to prevent security risks.


gwendolyngoetz:
This is solely being used to clear the content of the element and it's hard-coded to use an empty string. It's fine as it is

PR #7035, opened 2024-03-25
Enable perf speed on all browser test compose files

gwendolyngoetz:
### Description

Making the postgres perf settings available on all browser test compose files.

This worked locally fine, but turns out CI loaded a different browser test compose file. While the time savings from around 250ms to 25ms is accurate the run on GitHub that appeared faster was probably due to lower weekend usage of resources.

#### General

### DIFF
--- a/browser-test/browser-test-compose.yml
+++ b/browser-test/browser-test-compose.yml
@@ -2,6 +2,8 @@
 version: '3.4'
 
 services:
+  # Load custom postgres config that disables durability for performance
+  # Not for use in a production environment
   db:
     volumes:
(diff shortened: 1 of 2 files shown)


browser-test/browser-test-compose.yml
```diff
@@ -2,6 +2,11 @@
 version: '3.4'
 
 services:
+  db:
+    volumes:
```
dkatzz, line 6: **suggestion:** May be worth adding a comment that this should only be included for non-prod environments.

dkatzz (approved):

browser-test/browser-test-compose.yml
gwendolyngoetz: done

PR #7050, opened 2024-03-25
Allow decimals to be entered for age younger than / older than predicates

### LINKED ISSUE #7049: Allow decimals to be entered for age eligibility conditions
**Describe the bug**
Currently we only allow whole numbers to be entered for the age eligibility / visibility predicates. This prevents setting up an eligibility condition for "Older than" / "Younger than" 6 months for example. This isn't very common for an age question, but could be applicable for something like "When did you move into your home?", with people being eligible if they've been there for more than 6 months. 

**To Reproduce**

1. Go to to admin view
2. Create a program with a date question
3. Add an eligibility condition
4. Click on age is greater / less than as the operator and type in .5 or another decimal
5. See error

**Expected behavior**
Decimals should be allowed

dkatzz:
### Description

Currently only a whole number is accepted for age younger than and age older than, but we should allow for decimals.

This is to allow Charlotte to create a predicate for lived in a house less than 6 months. The admin can create a date question "When did you move into your house?" and can set eligibility on that question for age less than .5.

#### General

### Issue(s) this completes

Fixes #7049

### DIFF
--- a/server/app/services/program/predicate/PredicateValue.java
+++ b/server/app/services/program/predicate/PredicateValue.java
@@ -29,7 +29,7 @@ public abstract class PredicateValue {
   }
 
   public static PredicateValue of(double value) {
-    return create(String.valueOf(value), OperatorRightHandType.LONG);
+    return create(String.valueOf(value), OperatorRightHandType.DOUBLE);
   }
 
   public static PredicateValue of(String value) {
--- a/server/app/services/program/predicate/PredicateValue.java
+++ b/server/app/services/program/predicate/PredicateValue.java
@@ -28,7 +28,7 @@ public abstract class PredicateValue {
     return create(String.valueOf(value), OperatorRightHandType.LONG);
   }
 
-  public static PredicateValue of(float value) {
+  public static PredicateValue of(double value) {
     return create(String.valueOf(value), OperatorRightHandType.LONG);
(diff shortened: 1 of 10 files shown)


server/app/services/program/predicate/PredicateValue.java
```diff
@@ -28,6 +28,10 @@ public static PredicateValue of(long value) {
     return create(String.valueOf(value), OperatorRightHandType.LONG);
   }
 
+  public static PredicateValue of(double value) {
+    return create(String.valueOf(value), OperatorRightHandType.LONG);
```
melanie-exygy, line 32: question (non-blocking):  Do we need `OperatorRightHandType.DOUBLE` here?

melanie-exygy (commented):
Looks good.  I wonder if it might be useful to add a decimal date as an eligibility condition in a browser test as well.  Maybe in `application_navigation.test.ts 'navigation with eligibility conditions'`?

server/app/services/program/predicate/PredicateValue.java
dkatzz: Yeah, good catch - updated!

melanie-exygy (approved):

melanie-exygy (approved):

PR #7056, opened 2024-03-26
Split pagination into its own test file and improved efficiency

melanie-exygy:
### Description

The pagination tests were flaky and slow.  This moves the pagination tests into their own file so that they can run in parallel and removes some unnecessary repeated actions to improve efficiency.


browser-test/src/support/ti_dashboard.ts
```diff
@@ -41,25 +41,25 @@ export class TIDashboard {
 
     await this.page.getByRole('button', {name: 'Save'}).click()
     await waitForPageJsLoad(this.page)
-    await this.expectSuccessAlertOnAddNewClient()
     await this.gotoTIDashboardPage(this.page)
     await waitForPageJsLoad(this.page)
   }
 
   async createMultipleClients(nameBase: string, copies: number) {
+    await this.page.getByRole('link', {name: 'Add new client'}).click()
```
melanie-exygy, line 49: We don't need to return to the TI dashboard between adding each client.  We can stay on the "Add client" page.

gwendolyngoetz (approved):

gwendolyngoetz:
Just an FYI, nothing run in parallel at this time. We have it hard-coded to a single worker thread.

PR #7060, opened 2024-03-26
Fix bug where TI "Applying For" banner appeared on add and edit client

### LINKED ISSUE #6899: TI client edit view shows the "You are applying for..." banner
**Describe the bug**
The TI client edit view is showing the "You are applying for..." banner.  This banner should not show on the TI dashboard or any related pages.

**To Reproduce**

1. Log in as a TI.
2. Add a client if you don't have one already.
3. On the client card, next to the name, click the "Edit" link.
4. See the following banner appear:

image

**Expected behavior**
This banner should not show.

melanie-exygy:
### Description

When a TI is applying on behalf of a client, a banner shows at the top of the page that says "You are applying for...".  This banner should only show when TI is going through the process of applying for the client.  This PR fixes a bug where the banner was appearing on the TI new client page and the TI edit client page.  

Here is an image of the banner:
image

1. Log in as a Trusted Intermediary.
3. On the TI dashboard, click the "Add new client" button.
4. Note that the "You are applying for..." banner does not appear.
5. If you don't have any clients created, go ahead and create one.
6. Click the "Back to client list" link.
7. Click "Edit" next to the client name in the client card.  
8. Note that the "You are applying for..." banner does not appear on the edit page either.
9. Click the "Back to client list" link.
10. On the client card, click "View applications".
11. Note that the "You are applying for..." banner **does** appear on the program index page.

### Issue(s) this completes

Fixes #6899


server/app/views/applicant/ApplicantLayout.java
```diff
@@ -573,7 +574,10 @@ private static boolean onTiDashboardPage(Http.Request request) {
     } catch (URISyntaxException e) {
       logger.error("Could not get the path for uri {}", request.uri());
     }
+    if (currentPath == null) {
+      return false;
+    }
 
-    return currentPath != null && currentPath.equals(tiDashboardPath);
+    return currentPath.contains(tiDashboardPath) || currentPath.contains("/admin/tiGroups");
```
caitlinshk, line 581: suggestion (non-blocking): It feels a bit weird that one part of this statement checks a constructed URI value but the other part of this statement checks a hardcoded string. Could we have them either both check a constructed URI or both check a string?

caitlinshk (approved):

server/app/views/applicant/ApplicantLayout.java
melanie-exygy: I agree.  I've refactored to get both paths from the routes.  I didn't like the hard-coded string since that would break if, for some reason, we change the path to the edit client or add client pages.  I also didn't see the need to convert the TI dashboard route to a URI since we can be sure of the format given that that path is coming from our Play routes.

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -550,30 +550,46 @@ protected Optional<DivTag> maybeRenderBackToAdminViewButton(
...
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty())
+            .url();
+
+    String editClientPath =
+        controllers.ti.routes.TrustedIntermediaryController.editClient(1L).url();
+
+    // Remove the last two path segments to get the common path prefix "/admin/tiGroups"
```
caitlinshk, line 572: Question (non-blocking): It looks like `controllers.admin.TrustedIntermediaryManagementController.index(request: Request)` corresponds to just `"/admin/tiGroups", can we use that route here?

caitlinshk (approved):

server/app/views/applicant/ApplicantLayout.java
caitlinshk: True! Good call removing the URI portion.

server/app/views/applicant/ApplicantLayout.java
melanie-exygy: Hm, we could, although that controller is for a completely different part of the application (it's for when CF Admins are managing the TI groups).  This is making me think that a better course of action would be to rename the TI client edit and add routes to "/admin/tiDash/..." given that they are subpages of the TI dashboard.  I think I'll do that.  Let me know if you see any issue with it.

Issue #6197, opened 2023-12-14
Periodically rotate server secrets


yotommy:
The Play framework uses an [application secret](https://www.playframework.com/documentation/2.8.x/ApplicationSecret) that is used for encryption (including that of Play session cookies).

Currently, this secret is generated at deployment time and never updated. It is good practice to periodically rotate the secret in order to mitigate the risk of a leaked secret.

Since user profiles are stored in Play session cookies, we must understand the impact of changing the server secret on users with applications in progress.

yotommy:
[Using AWS Parameter Store for Play Secret Rotation](https://github.com/guardian/play-secret-rotation/blob/main/aws-parameterstore/README.md)

This package above rotates Play server secrets using the AWS Parameter Store. It supports an `overlapDuration` during which the server will honor both the old and new secrets. (This seems like it would require support from Play itself.)

Here is an [explanation](https://github.com/guardian/play-secret-rotation/commit/01e7fa86688) of why this package uses the Parameter Store rather than the Secrets Manager.

yotommy:
Some deployments don't use a cloud provider.

Look into how Play handles this first.

yotommy:
The overlap period needs to be greater than or equal to session expiration time.

yotommy:
The [`play-secret-rotation` package](https://github.com/guardian/play-secret-rotation) can be used outside the context of the AWS Parameter Store by customizing the `secretStateSupplier` parameter provided to `RotatingSecretComponents`.

Posted a question on how to do this in a Java server: [Using play-secret-rotation in a Java server](https://github.com/guardian/play-secret-rotation/issues/427)

yotommy:
Posted another question to the Play discussion board: [Managing server secrets in a Java server](https://github.com/orgs/playframework/discussions/12341)

mkurz:
> I presume users with an account would just need to log in again.

Yes that should be all.

> Marking this P0, because we're going to have to do this before we move to Play 2.9, and we have to do that in the next 2 months.

Even if you don't implement secret rotation until May 31st, you can still just update `play.http.secret.key` to longer random value and you can upgrade (with the need of users to log in again). You can still implement secret rotation afterwards.

yotommy:
Thanks, @mkurz, that is encouraging.

I did some experiments:

1. A guest user has an application in progress when the secret is changed. The work is lost, but the user is able to continue to access the site (with a fresh session) with no error.
[...]

PR #7095, opened 2024-04-01
Redirect to home page if unauthorized (except for API requests)

yotommy:
### Description

Redirect to home page if the request is not authorized.

This can be useful if (1) the server secret is rotated and (2) an admin user with an active session tries to access an admin page. Before this change, they would just be presented with a page with `403 Forbidden` and no hint about how to recover.

Note there is existing handling for API requests that is untouched.

#### General

1. Run a local instance of the server.
2. Authenticate as admin and load an admin page.
3. Modify `play.http.secret.key` in `server/conf/application.conf`.
4. Reload the admin page in the browser.
5. Verify that you are redirected to the home page.
6. Verify that the page says, "You're a guest user" in the top-right corner.

### Issue(s) this completes

Relates to #6197

### DIFF
--- a/browser-test/src/security.test.ts
+++ b/browser-test/src/security.test.ts
@@ -5,9 +5,11 @@ test.describe(
   'applicant security',
   {tag: ['@uses-fixtures', '@parallel-candidate']},
   () => {
-    test('applicant cannot access admin pages', async ({page}) => {
-      const response = await page.goto('/admin/programs')
-      expect(response!.status()).toBe(403)
+    test('applicant cannot access admin pages', async ({request}) => {
+      const response = await request.get('/admin/programs')
+      await expect(response).toBeOK()
+      // Redirected to a non-admin page
+      expect(response.url()).not.toContain('/admin')
     })
 
     test('redirects to program index page when not logged in (guest)', async ({
--- a/server/test/auth/CiviFormHttpActionAdapterTest.java
+++ b/server/test/auth/CiviFormHttpActionAdapterTest.java
@@ -1,0 +1,61 @@
+package auth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static play.test.Helpers.fakeRequest;
+
+import org.junit.Test;
+import org.pac4j.core.context.HttpConstants;
+import org.pac4j.core.exception.http.StatusAction;
+import org.pac4j.play.PlayWebContext;
+import play.mvc.Result;
+
+public class CiviFormHttpActionAdapterTest {
+
+  // An UNAUTHORIZED response to an API route should remain UNAUTHORIZED.
+  @Test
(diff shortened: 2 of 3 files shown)


yotommy:
* **#7095** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @yotommy and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

server/test/auth/CiviFormHttpActionAdapterTest.java
```diff
@@ -0,0 +1,61 @@
...
+    assertThat(result.status()).isEqualTo(HttpConstants.SEE_OTHER);
+    assertThat(result.redirectLocation()).isNotEmpty();
+    assertThat(result.redirectLocation().get()).isEqualTo("/");
+  }
+
+  // An OK response should just ... be OK.
+  @Test
+  public void testOkAction() {
+    CiviFormHttpActionAdapter adapter = new CiviFormHttpActionAdapter();
+    PlayWebContext context = new PlayWebContext(fakeRequest("GET", "/non-api").build());
```
gwendolyngoetz, line 55: (question) This is testing `ok` no matter if it's api or non-api? If yes, maybe give this a different path than `/non-api` to give context that it isn't grouped with just the other non-api endpoints.

browser-test/src/security.test.ts
```diff
@@ -7,7 +7,9 @@ test.describe(
   () => {
     test('applicant cannot access admin pages', async ({page}) => {
       const response = await page.goto('/admin/programs')
-      expect(response!.status()).toBe(403)
+      expect(response!.status()).toBe(200)
+      // Redirected to a non-admin page
+      expect(response!.url()).not.toContain('/admin')
```
gwendolyngoetz, line 12: (non-blocking) If you want to speed this test up, replace `async ({page})` with `async ({request})` and swap over to using [get](https://playwright.dev/docs/api/class-apirequestcontext#api-request-context-get) with [specific response code specific assertions](https://playwright.dev/docs/api/class-apiresponseassertions#api-response-assertions-to-be-ok). 

It will be faster since it will skips clearing the database.

server/test/auth/CiviFormHttpActionAdapterTest.java
yotommy: Yes and done.

browser-test/src/security.test.ts
yotommy: Thanks, done.

gwendolyngoetz (approved):

PR #7097, opened 2024-04-01
Client data added by TIs is used to pre-population applications

### LINKED ISSUE #5430: Ensure that all client data is used to populate program applications
Right now, when TIs add clients and then apply to a program on their behalf, only client name is auto-populating in the application. In order to reduce duplicative efforts, we need to ensure at all information (such as email) that a TI adds in creating a client is brought over to the applications they fill out. 

This was reported via user testing, so the first step here is to confirm the underlying behavior. Then to ensure that we are able to reuse all client data in applications. 

This issue is complete when we confirm that a TI user can add a new client and then, when applying to a program on their behalf, see First Name, Middle Name, Last Name, DOB, and email auto-populate the application.

rockycodes:
### Description

This change makes it so that data entered by a TI when they create a client (eg. name, email, phone, date of birth) is used to pre-populate applications that contain questions tagged as "Personal Applicant Info" questions.

This is reliant on functionality hidden behind the `PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED` feature flag, and will not work correctly unless that flag is enabled.

#### General

1. Login as a CiviForm admin and enable the `PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED` feature flag
2. Create 4 questions (Date, Email, Name, Phone) and toggle on the "universal" and "primary applicant info" toggles for each question
3. Create and publish a program using those 4 questions
4. Login as a TI and create a client with name, dob, email, and phone
5. Apply to the program you created on behalf of the newly created client
6. See the client's data pre-populated

### Issue(s) this completes

Fixes #5430


browser-test/src/trusted_intermediary.test.ts
```diff
@@ -930,4 +931,142 @@ test.describe('Trusted intermediaries', () => {
...
+        page,
+        tiDashboard,
+        adminQuestions,
+        adminPrograms,
+        applicantQuestions,
+      } = ctx
+
+      await enableFeatureFlag(page, 'primary_applicant_info_questions_enabled')
+
+      // Login as TI and add a client
```
rockycodes, line 947: wrapping this in `test.step` was causing it to fail for some reason

server/app/services/applicant/question/EmailQuestion.java
```diff
@@ -46,7 +48,15 @@ public Optional<String> getEmailValue() {
...
 
-    emailValue = applicantQuestion.getApplicantData().readString(getEmailPath());
+    ApplicantData applicantData = applicantQuestion.getApplicantData();
+    Optional<String> emailValue = applicantData.readString(getEmailPath());
+
+    if (emailValue.isEmpty()
+        && applicantQuestion
+            .getQuestionDefinition()
+            .containsPrimaryApplicantInfoTag(PrimaryApplicantInfoTag.APPLICANT_EMAIL)) {
+      emailValue = applicantData.getApplicant().getEmailAddress();
```
rockycodes, line 58: Email is the only field that is not accessible via `applicantData`. All of the other fields go through `applicantData` to pull the value off of the applicant model.

server/app/services/applicant/question/Question.java
```diff
@@ -100,11 +102,41 @@ && getAllPaths().stream().anyMatch(failedUpdates::containsKey)) {
...
+    boolean isAnsweredWithApplicantData =
+        getAllPaths().stream().anyMatch(applicantQuestion.getApplicantData()::hasPath);
+
+    // Filter to see if the applicant has data saved for any of the PAI tags on the question
+    // definition.
+    // If the returned set is NOT empty, then the applicant has data saved and the question is
+    // "answered"
+    ImmutableSet<PrimaryApplicantInfoTag> tags =
+        applicantQuestion.getQuestionDefinition().getPrimaryApplicantInfoTags();
+    boolean isAnsweredWithPrimaryApplicantInfo =
```
rockycodes, line 118: This allows the answer to be shown on the preview screen.

server/test/services/applicant/question/DateQuestionTest.java
```diff
@@ -88,4 +89,29 @@ public void withMisformattedDate() {
...
+                .build());
+
+    // Save applicant's dob to the PAI column
+    applicant.setDateOfBirth("2001-01-01");
+
+    DateQuestion dateQuestion =
+        new ApplicantQuestion(dateQuestionDefinitionWithPAI, applicantData, Optional.empty())
+            .createDateQuestion();
+
+    assertThat(dateQuestion.getDateValue().get()).isEqualTo(applicant.getDateOfBirth().get());
```
rockycodes, line 115: The dob stored in the database is already present on the question, without the applicant filling anything out.

server/app/services/applicant/question/NameQuestion.java
```diff
@@ -72,8 +82,15 @@ public Optional<String> getMiddleNameValue() {
       return middleNameValue;
     }
 
-    middleNameValue = applicantQuestion.getApplicantData().readString(getMiddleNamePath());
+    ApplicantData applicantData = applicantQuestion.getApplicantData();
+    middleNameValue = applicantData.readString(getMiddleNamePath());
```
swatkat1, line 86: Please check if the ReadString()  also handles cases where no path exists. Like few applicants may not have a middle name.

server/app/services/applicant/question/EmailQuestion.java
```diff
@@ -46,7 +48,15 @@ public Optional<String> getEmailValue() {
       return emailValue;
     }
 
-    emailValue = applicantQuestion.getApplicantData().readString(getEmailPath());
+    ApplicantData applicantData = applicantQuestion.getApplicantData();
```
swatkat1, line 51: I really like that we are going from Client data --> application. I am wondering if we should also consider Application --> ClientData. Example- If in the application, they specify the email address as "xx@yy.com" and the Client is missing that info, should we populate it back?

server/app/services/applicant/question/EmailQuestion.java
rockycodes: It is kinda weird to have only the one exception, so I would be in favor of adding a `getEmailAddress` to `applicantData`. Whether we should get it from `applicantData` or directly off of `applicant` going forward... I think that depends on how easy it would be to refactor to use `applicant` instead of `applicantData`. I remember when we looked at this before we decided we would need to keep `applicantData.getApplicant`, which would still be funky, but maybe there's not too many places we would need to do that.

server/app/services/applicant/question/NameQuestion.java
swatkat1: It is not an issue. Optional<String> saved the day :)

server/app/services/applicant/question/Question.java
rockycodes: @nb1701 do you remember if we actually made a decision about this at some point or should i confirm this with product?

server/app/services/applicant/question/EmailQuestion.java
rockycodes: That's a great idea! I'll check with product on that.

rockycodes: @swatkat1 I tested this out and this was actually mostly how it behaved by default with these changes! I just had to add an update so email is fetched from PAI. I'm still waiting on the final 👍 from product, but it's looking like this is how it will work.

server/app/services/applicant/question/Question.java
```diff
@@ -100,11 +102,41 @@ && getAllPaths().stream().anyMatch(failedUpdates::containsKey)) {
...
-   * it will be considered unanswered.
+   * associated with the question or if the question is tagged with a Primary Applicant Info Tag and
+   * the applicant has data saved in the corresponding column.
    */
   public boolean isAnswered() {
-    return getAllPaths().stream().anyMatch(p -> applicantQuestion.getApplicantData().hasPath(p));
+    boolean isAnsweredWithApplicantData =
+        getAllPaths().stream().anyMatch(applicantQuestion.getApplicantData()::hasPath);
+
+    // Filter to see if the applicant has data saved for any of the PAI tags on the question
```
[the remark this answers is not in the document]
rockycodes: fixed

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -304,23 +304,25 @@ private DivTag renderCardContactInfo(AccountModel account, Messages messages) {
       return div();
     }
     ApplicantData applicantData = newestApplicant.get().getApplicantData();
-    Optional<String> maybePhoneNumber = applicantData.getPhoneNumber();
-    String email = account.getEmailAddress();
+    Optional<String> optionalPhoneNumber = applicantData.getPhoneNumber();
```
rockycodes, line 307: I updated this var name while I was here.

swatkat1 (approved):

server/app/services/applicant/question/Question.java
rockycodes: That makes sense, I'll play around with this.

server/app/services/applicant/question/EmailQuestion.java
rockycodes: done.

server/app/services/applicant/question/Question.java
rockycodes: done.

server/app/services/applicant/question/Question.java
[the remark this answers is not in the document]
rockycodes: Oh I think this might be outdated. I removed the comment when I refactored the logic.

Issue #6912, opened 2024-03-14
Update TI client edit and add pages with latest UX design


melanie-exygy:
#### Describe the solution presented in this issue.
The TI client edit and new client views differ from the latest TI dashboard mocks.  In this issue, we'll make them match as much as possible.  All input fields should use the USWDS input component.  The back link should have an arrow icon.  There shouldn't be a border around the form.

#### Link to Figma mocks: 
[mocks](https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=1112%3A1348&mode=design&t=y9ukT6E6JXHgHNZ8-1)

#### Add screenshot of relevant mocks for this issue here: 

<img width="275" alt="image" src="https://github.com/civiform/civiform/assets/116591382/a00e503b-cc11-4351-ac9c-da99528abf09">

#### Additional Context
Add any other context, link related issues, etc. 

#### Done when:
- [ ] All input fields use USWDS input component
- [ ] Labels are correct and internationalized
- [ ] No border around the form
- [ ] Back link has arrow icon
- [ ] 'Cancel' button has white background
- [ ] Buttons use USWDS styles

melanie-exygy:
@sijiayam, the current form includes a "Middle name" field, but the mocks don't.  Is that intentional?  Also, if the email field should keep the existing tooltip, could you please add that to the mocks?

sijiayam:
@melanie-exygy Thanks for catching that.

Updated mocks to:
1. add middle name
2. Note that we should likely have phone number for creating a client rather than just on edit client
3. Added a tooltip which uses `info` https://designsystem.digital.gov/components/icon/` 
4. Updated buttons to reflect our move to USWDS buttons, from left to right the styles are
    - `default`
    - `outline`
    - `accent-warm`

PR #7103, opened 2024-04-02
Update TI client edit and add pages with latest UX design

melanie-exygy:
### Description

Using USWDS button styles for TI pages, making sure headers and label text match the mocks and are translated.  Removing outline around the form and adding spacing around the form header to match mocks.  This PR does not update the input fields to use USWDS styles.  That will be in a follow-up PR.

1.  Log in as a TI.
2. Click "Add new client".
3. Notice the style changes.

This doesn't fully complete #6912 but contributes to it.


swatkat1 (approved):
Looks good!

Issue #6800, opened 2024-03-02
Address Eligibility to include option for "outside service area"


Miguel-314:
The Office of Housing Weatherization only available for residents who are in the Seattle City Light service area but outside the City of Seattle AND heat their home with electric heat. Customers in this category that heat their home with either Oil or Gas must use the King County Weatherization program. 

We have determined that we want to create a visibility condition that shows static text informing Oil and Gas heated homes outside the City of Seattle information about the King County program with a link. 

Here is a screen shot of the current visibility options. 

![Image](https://github.com/civiform/civiform/assets/126286905/596cb10c-2935-4ff9-968c-f7171a40ac8c)

In the second column, the images shows "in service area". This is the only option you can select in this drop down. 

Is it possible to create a second option that filters applicants who are "out of service area"?

gwendolyngoetz:
@Miguel-314 try setting the third drop down  to this
![image](https://github.com/civiform/civiform/assets/195162/ff2f9c77-9cbc-4a19-91f3-77fd09148e8a)

msprenke:
@gwendolyngoetz let me know if you need a design document for this.  Seems pretty straight forward.

Wondering if in the meantime it could be configured in the opposite way:  instead of shown if gas/oil Not in service area, we could do NOT show if Electric and IN the service area?

Miguel-314:
I might be reading this wrong but I don't think that will work. Here is the full eligibility I am trying to program. 



![Image](https://github.com/civiform/civiform/assets/126286905/6fb2c3b4-9a6c-467d-a979-df871e5d9e41)

Miguel-314:
I think this has been properly communicated but just in case, the feature being request is a visibility/eligibility condition that allows for both "in service area" and "outside service area".

PR #7105, opened 2024-04-02
Allow overriding the wkid in calls to Esri services

gwendolyngoetz:
### Description

The Esri API has the ability to return spatial reference data using different coordinate systems via a wellKnownId (a.k.a wkid). Details of that along with links to the long list of possible coordinate systems can be found on the [arcgis developer site](https://developers.arcgis.com/rest/services-reference/enterprise/using-spatial-references.htm).

The default wellKnownId is `4326 - GPS` which are traditional latitude and longitude value most often thought around of around coordinate systems. Custom hosted installations may set a different value as their default. For example, Seattle's installation defaults to using wellKnownId `2926 - NAD_1983_HARN_StatePlane_Washington_North_FIPS_4601_Feet` [link](https://developers.arcgis.com/rest/services-reference/enterprise/pdf/pcs_PDF_11.2.pdf).

This works fine as long as calls to correct an address and calls to mapping layers are defaulting to the same wellKnownId. Such as calling both endpoints on Seattle's instance.

Where it becomes a problem is if you want to start mixing calls to different installations.

#### Example

> 
> - Call arcgis.com to correct address, returns
> 
>   ```json5
>   // A very much truncated example
>   {
>    "spatialReference": {
>     "wkid": 4326
>    },
>    "candidates": [{
>      "location": {
>       "x": -122.329922012207,
>       "y": 47.605043979178
>      },
>     }]
>   }
>   ```
> 
> - Call Seattle map query endpoint passing in `x`, `y`, and `wkid` values from previous arcgis.com response. By no match is found because the coordinates are in 4326 format and the service expects 2926 format.
> 
> - Call King County map query endpoint passing in `x`, `y`, and `wkid` values from previous arcgis.com response. By no match is found because the coordinates are in 4326 format and the service expects 2926 format.

We can normalize this by specifying the output spatial reference when we call to correct the address using the `outSR` parameter.

#### Existing data

After we get the corrected address we save the x, y, and wkid as part of the address. Calls to map query endpoints already use the stored wkid to specify what the input spatial reference should be. Existing records will still work as is. 

> [!WARNING]
> If the override is added at a later date and is different from the previously used default you will end up having new records in a different format. If that is a concern the solution is to make sure the override value is the same as used before.
>
> This is only an issue if you change the find address endpoint at a later point in time to one with a different default. Such as using Seattle's version to using arcgis.com
> 
> If there is only a single host this will never be a problem since they'll both have the same defaults.

## Release notes

Allow overriding the default Esri wellKnownId value to allow consistent spatial reference settings if using service area validation from different sources.

#### General

### Issue(s) this relates to

Relates to #6800


server/app/services/settings/SettingsManifest.java
```diff
@@ -578,6 +578,18 @@ public Optional<Integer> getEsriExternalCallTries() {
     return getInt("ESRI_EXTERNAL_CALL_TRIES");
   }
 
+  /**
+   * Forces calls to Esri services to use the specified spatial reference wellKnownId value for the
+   * [coordinate
+   * system](https://developers.arcgis.com/rest/services-reference/enterprise/using-spatial-references.htm).
+   * If not set the default configuration from the Esri server is used. Setting this may be needed
```
dkatzz, line 585: **question:** Is it worth noting the default as 4326 - GPS?

dkatzz (approved):
Overall the changes LGTM! Could we update the docs as well https://docs.civiform.us/it-manual/sre-playbook/initial-deployment/configure-gis-service?

gwendolyngoetz:
> Overall the changes LGTM! Could we update the docs as well https://docs.civiform.us/it-manual/sre-playbook/initial-deployment/configure-gis-service?

@dkatzz  Already in progress

server/app/services/settings/SettingsManifest.java
dkatzz: Oh got it makes sense!

melanie-exygy (approved):
question (non-blocking):  Would we ever have a situation where two different mapping services (Seattle vs. King County, for example) expect a different wellKnownID?

gwendolyngoetz:
> question (non-blocking): Would we ever have a situation where two different mapping services (Seattle vs. King County, for example) expect a different wellKnownID?

@melanie-exygy It's possible multiple ones could have different defaults, but we already are set up to tell the mapping layers (service area validation) what the source spatial reference is based on the value returned from correcting the address.

Issue #6999, opened 2024-03-21
Setup Github codespaces for development


dkatzz:
#### Is your feature request related to a problem? Please describe.

Currently, developers working on CiviForm need to set up a local development environment on their computer, which involves installing docker and making other updates to their development environment.

By using codespaces, we can reduce the local docker dependency and make it easier for developers to get started working on CiviForm.

#### Done when:

Developers can run and test the CiviForm application without having docker installed

#### Other steps:

Once setup, we want to ensure the following work in the main repo:

- [x] bin/run-dev
  - [x] login (with vscode remote)
  - [x] file upload
- [x] bin/fmt
- [x] bin/sbt
- [x] bin/run-test
- [x] bin/sbt-test -> then run specific test
- [x] bin/run-ts-tests
- [x] [running browser tests](https://github.com/civiform/civiform/wiki/Getting-started#running-browser-tests)
  - [x] Running all at once (takes a long time but works)
  - [x] With `-u`
  - [x] Running individual test
- [ ] [Attaching debugger](https://github.com/civiform/civiform/wiki/Testing#attaching-a-debugger-to-unit-tests)
- [ ] bin/run-test-coverage
- [x] bin/dev-psql - will work once https://github.com/civiform/civiform/pull/7083 is merged
- [x] Java imports
- [x] Java resolving
- [x] Graphite installed by default - https://github.com/civiform/civiform/pull/7110
- [x] Using playwright extension - https://github.com/civiform/civiform/pull/7111
- [x] Persist terminal history

The following in the infra repo:

- [x] bin/fmt
- [x] bin/build-cloud-deployment

The deploy repo won't be included in the changes for now, since it doesn't require docker to be installed.

dkatzz:
Updated docs: [https://github.com/civiform/civiform/wiki/Getting-started/_compare/96b0cadb1c09097[…]12f57889c4188d5dba04...ca2cb12fefbe180d78ccde3c74efb6dfaf50578f](https://github.com/civiform/civiform/wiki/Getting-started/_compare/96b0cadb1c090976bf2112f57889c4188d5dba04...ca2cb12fefbe180d78ccde3c74efb6dfaf50578f)

dkatzz:
In order to allow persist history you can add the following to ~/.bashrc in the codespace:

```
# Append history to a file in the persistent storage volume
export HISTFILE="$HOME/.bash_history"
# Ensure unlimited history size
export HISTSIZE=-1 
export HISTFILESIZE=-1
[...]

PR #7109, opened 2024-04-02
Install graphite in devcontainer

dkatzz:
### Description

This allows developers to use graphite commands as soon as their codespace gets set up.

#### General

### Issue(s) this completes

#6999

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
+
+# Install Homebrew For Linux
+RUN /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" && \
+    eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)" && \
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
+
+# Install Homebrew For Linux
+RUN /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" && \
+    eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)" && \
+    echo "eval \"\$($(brew --prefix)/bin/brew shellenv)\"" >> /home/vscode/.zshrc && \
+    echo "eval \"\$($(brew --prefix)/bin/brew shellenv)\"" >> /home/vscode/.bashrc && \
+    # Install Graphite
+    brew install withgraphite/tap/graphite
```
gwendolyngoetz, line 9: (question) Are there any installation methods that use apt, pip, or npm? Would be nice to not have another package manage system.

dkatzz: Makes sense! Yeah they recommended brew (https://graphite.dev/docs/install-the-cli), but I agree with that and npm seems to work the same! Created https://github.com/civiform/civiform/pull/7110 which installs with npm

PR #7112, opened 2024-04-02
Using USWDS styles for input fields and labels on TI forms

### LINKED ISSUE #6912: Update TI client edit and add pages with latest UX design

melanie-exygy:
### Description

Using USWDS styles for input fields and labels on the TI add/edit client form.  Since these fields are of type `FieldWithLabel`, some refactoring was needed to move CSS class assignment further down the call stack and to separate it from attribute assignment and validation error creation.  Much of the `FieldWithLabel` logic was duplicated with the only change being USWDS classes.  This duplication is temporary given that we plan to migrate all CF input fields to USWDS with the North Star work.

1.  Log in as a Trusted Intermediary.
2. Click on the "Add new client" button.
3. Notice that the input fields now have USWDS styles.
4. Click "Save".
5. Notice that validation error messages still show, with USWDS error styles.

image

### Issue(s) this completes

Fixes #6912


browser-test/image_snapshots/end_to_end_enumerators_test/enumerator-field.png
melanie-exygy, line 1: The difference here and in some of the other snapshots is that the label with text "Administrative identifier..." now has the same styles as other labels (it isn't more bold and the letters aren't closer together).  Doing this means that the text is a bit more spaced out and that makes the tooltip icon wrap onto the next line.  I don't think it looks great, but it only looks this way with this particular screen size.  Also, the wrapping behavior is unchanged from what it was before.  Even with `display:block` the label still wraps when the screen size is narrower (you can verify this on staging).

server/app/views/applicant/EditTiClientView.java
```diff
@@ -310,13 +310,13 @@ private DivTag renderAddOrEditClientForm(
...
                 .with(
-                    firstNameField.getInputTag(),
-                    middleNameField.getInputTag(),
-                    lastNameField.getInputTag(),
-                    phoneNumberField.getInputTag(),
-                    emailField.getEmailTag(),
-                    dateOfBirthField.getDateTag(),
-                    tiNoteField.getTextareaTag(),
+                    firstNameField.getUSWDSInputTag(),
+                    middleNameField.getUSWDSInputTag(),
```
swatkat1, line 314: This is cleaver and a lot more understandable.

server/app/views/components/FieldWithLabel.java
```diff
@@ -428,7 +428,19 @@ private DivTag getNonNumberInputTag() {
     } else {
       throw new RuntimeException("non-number tag expected");
     }
-    return applyAttrsAndGenLabel(inputFieldTag);
+    return applyAttrsClassesAndLabel(inputFieldTag);
+  }
+
+  private DivTag getUSWDSNonNumberInputTag() {
```
swatkat1, line 434: This is doing almost the same things as getNonNumberInputTag(), can you combine the two? May be pass in a variable or set a class param to identify it as a USWDS component.

server/app/views/components/FieldWithLabel.java
```diff
@@ -443,7 +455,24 @@ public DivTag getTextareaTag() {
...
         textareaFieldTag.withCols(String.valueOf(this.cols.getAsLong()));
       }
-      return applyAttrsAndGenLabel(textareaFieldTag);
+      return applyAttrsClassesAndLabel(textareaFieldTag);
+    }
+
+    throw new RuntimeException("needs to be textarea tag");
+  }
+
+  public DivTag getUSWDSTextareaTag() {
```
swatkat1, line 464: This is also doing almost the same things as getTextareaTag(). Can you look to combining them?

melanie-exygy: I see what you're saying.  I considered the option of passing in a `isUSWDS` variable here, but there are 21 usages of this method throughout the app and I'd have to add the parameter in all of those places.  Then, when USWDS becomes the norm for our input fields, we'd have to remove the parameters in all those places.  By duplicating the function, we temporarily have two almost identical functions, but we'll have less to change when we are only using USWDS for input fields.  We'll be able to blend the two functions into one at that time.  Does that sound reasonable?

server/app/views/components/FieldWithLabel.java
melanie-exygy: Done

swatkat1 (approved):
Nicely done!!!

server/app/views/components/FieldWithLabel.java
swatkat1: ok sounds good! - Can you add a TODO item to make sure we followup. Thank you.

melanie-exygy: Good idea, will do!

server/app/views/components/FieldWithLabel.java
```diff
@@ -428,7 +428,19 @@ private DivTag getNonNumberInputTag() {
...
+    applyAttributesFromMap(inputFieldTag);
+    if (!this.fieldType.equals("number")) {
+      inputFieldTag.withValue(this.fieldValue);
+    } else {
+      throw new RuntimeException("non-number tag expected");
+    }
+    return applyUSWDSAttrsClassesAndLabel(inputFieldTag);
   }
 
   /** Public final tag getters * */
```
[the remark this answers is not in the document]
melanie-exygy: I didn't add this comment.  I think it's just marking the start of the public tag getter methods (vs the internal methods).

server/app/views/components/FieldWithLabel.java
melanie-exygy: Done

Issue #4763, opened 2023-05-03
Tag repo civiform/cloud-deploy-infra with release version


gwendolyngoetz:
When we create a release version within repo `civiform/civiform` also tag repo `civiform/cloud-deploy-infra` with the same version number.

dkatzz:
Another issue with the current situation is that release notes don't include updates from cloud-deploy-infra.

Notes from eng meeting: 
Nick has idea of possibly moving this repo into the civiform repo.  We might be able to bring the changes to cloud-deploy into the civiform repo release notes, but not sure exactly how.

PR #7126, opened 2024-04-03
Trigger cloud-deploy-infra tag action on release

nb1701:
This triggers the tagging action on cloud-deploy-infra when doing a release, so both repos are tagged with the same tag. This will, by default, tag the HEAD of the `main` branch of cloud-deploy-infra, but a different SHA can be specified if desired.

Addresses https://github.com/civiform/civiform/issues/4763

### DIFF
--- a/.github/ISSUE_TEMPLATE/oncall-rotation.md
+++ b/.github/ISSUE_TEMPLATE/oncall-rotation.md
@@ -29,6 +29,7 @@ Resources:
   - Changes to other stateful parts of the application (i.e. changes to the format of things stored in the database not necessarily requiring an evolution)
   - Config setting addition or removal
 - [ ] Create a Draft Release by 12 PT on Tuesday
+  - Ensure both the `civiform` and `cloud-deploy-infra` repos are tagged appropriately after running the `Create Release` action.
 - [ ] Contact Matthew Sprenke for QA on Slack
 - [ ] If Matthew is unavailable have the general team attempt a best-effort QA.
 - [ ] After QA, publish the release


.github/workflows/release.yaml
```diff
@@ -32,6 +37,14 @@ jobs:
       - name: Run bin/create-release
         run: 'GH_TOKEN=${{ secrets.GITHUB_TOKEN }} bin/create-release ${{ github.event.inputs.commit_sha }} ${{ github.event.inputs.release_number }}'
 
+      - name: Tag cloud-deploy-infra
```
gwendolyngoetz, line 40: (blocking) Add the `if success` check so this doesn't run if creating the release fails. Should be like doing the failure [here](https://github.com/civiform/civiform/blob/45d55b4e80bb4fbbd2059587856693994d65d064/.github/workflows/tests.yaml#L202).

.github/workflows/release.yaml
```diff
@@ -32,6 +37,14 @@ jobs:
...
+        run: |
+          curl --fail-with-body -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.CIVIFORM_GITHUB_AUTOMATION_PERSONAL_ACCESS_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/civiform/cloud-deploy-infra/actions/workflows/tag.yaml/dispatches \
```
gwendolyngoetz, line 46: (question) What is the process for the releaser if this step fails?

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

gwendolyngoetz (approved):

Issue #6995, opened 2024-03-20
[Disabled Programs] add `disabled` to the program visibility


azhangggggg:
#### Is your feature request related to a problem? Please describe.

Per Arkansas's ask, they'd like a way to delete programs. 

#### Describe the solution you'd like.

We provided them with the solution `disable` programs. 

#### What priority is this?

It's high as in it's blocking Arkansa launch.

#### Associated civic entity

Arkansas

#### Done when:

There's a `disabled` status for program visibility options. 
Applicants won't see the disabled programs from the dashboard if they haven't started with the application yet.

gwendolyngoetz:
@azhangggggg Can this be closed?

PR #7144, opened 2024-04-04
[Disabled Programs] add disabled as a visibility option

azhangggggg:
### Description

Added `DISABLED` enum value to DisplayMode enum and updated admin UI to have an additional option for it. The programs with `Disabled` visibility will be hidden from applicants, TI who have not started an application to the program and program admins. This PR ensures that they're hidden from applicants and TIs. Next PR will ensure that they're hidden from program admins.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

#6995


browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -16,6 +16,7 @@ test.describe('program creation', () => {
 
   test('create program page', async () => {
     const {page, adminPrograms, adminProgramImage} = ctx
+    await enableFeatureFlag(page, 'disabled_visibility_condition_enabled')
```
rockycodes, line 19: suggestion (blocking): my preference would be to add a new test that tests functionality/UI when the flag is on instead of enabling the feature flags in the existing tests.

browser-test/src/admin/admin_program_visibility.test.ts
```diff
@@ -272,4 +273,47 @@ test.describe('Validate program visibility is correct for applicants and TIs', (
...
+    const programDescription = 'Description'
+    await adminPrograms.addProgram(
+      programName,
+      programDescription,
+      'https://usa.gov',
+      ProgramVisibility.DISABLED,
+    )
+    await adminPrograms.publishAllDrafts()
+
+    // Login as applicant, verify program is hidden
```
rockycodes, line 292: suggestion (non-blocking): use [`test.step`](https://playwright.dev/docs/api/class-test#test-step) to break up these steps into testable chunks of code. If any of them fail, it will be more clear which step they failed in. Here's an [example I recently added](https://github.com/civiform/civiform/blob/a03f3c9ccef88d279f76abaf856bc63198ace892/browser-test/src/trusted_intermediary.test.ts#L982).

browser-test/src/admin/admin_program_visibility.test.ts
```diff
@@ -272,4 +273,47 @@ test.describe('Validate program visibility is correct for applicants and TIs', (
...
+    await logout(page)
+    const applicantQuestions = new ApplicantQuestions(page)
```
rockycodes, line 294: suggestion (blocking): pull this off of context at the top of the test like `const {page, tiDashboard, adminPrograms, applicantQuestions} = ctx`

rockycodes (changes requested):
this looks good! i have a few comments/suggestions about test structure.

browser-test/src/admin/admin_program_visibility.test.ts
azhangggggg: that totally makes sense since it makes it easier to debug! good suggestion, updated the test.

browser-test/src/admin/admin_program_visibility.test.ts
azhangggggg: done!

browser-test/image_snapshots/admin_program_creation_test/program-creation-page.png
rockycodes, line 1: suggestion (non-blocking): I think these files are identical. Try deleting it and rebasing off of main and see if it resolves this diff.

browser-test/image_snapshots/admin_program_creation_test/program-description-page-with-intake-form-false.png
rockycodes, line 1: same here re: rebasing off of main to resolve the diff

browser-test/image_snapshots/admin_program_creation_test/program-description-page-with-intake-form-true.png
rockycodes, line 1: same here re: rebasing off of main to resolve the diff

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -40,6 +40,64 @@ test.describe('program creation', () => {
...
+      ProgramVisibility.DISABLED,
+      'admin description',
+      /* isCommonIntake= */ false,
+      'selectedTI',
+      'confirmationMessage',
+      Eligibility.IS_GATING,
+      /* submitNewProgram= */ false,
+    )
+    await adminPrograms.expectProgramDetailsSaveAndContinueButton()
+    await validateScreenshot(
```
rockycodes, line 61: suggestion (non-blocking): you could also test that `Disabled` text shows up on the page.

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -40,6 +40,64 @@ test.describe('program creation', () => {
...
+    const {page, adminPrograms, adminProgramImage} = ctx
+    await enableFeatureFlag(page, 'disabled_visibility_condition_enabled')
+    await loginAsAdmin(page)
+
+    await adminPrograms.addProgram(
+      'program name',
+      'description',
+      'https://usa.gov',
+      ProgramVisibility.DISABLED,
+      'admin description',
```
rockycodes, line 53: nitpick (non-blocking): The params from here to the bottom all have default values so you could leave them off.

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -40,6 +40,64 @@ test.describe('program creation', () => {
...
+    const {page, adminPrograms, adminProgramImage} = ctx
+    await disableFeatureFlag(page, 'disabled_visibility_condition_enabled')
+    await loginAsAdmin(page)
+
+    await adminPrograms.addProgram(
+      'program name',
+      'description',
+      'https://usa.gov',
+      ProgramVisibility.PUBLIC,
+      'admin description',
```
rockycodes, line 82: same here re: params

browser-test/src/admin/admin_program_visibility.test.ts
```diff
@@ -272,4 +273,54 @@ test.describe('Validate program visibility is correct for applicants and TIs', (
...
+        programName,
+        programDescription,
+        'https://usa.gov',
+        ProgramVisibility.DISABLED,
+      )
+      await adminPrograms.publishAllDrafts()
+    })
+
+    await test.step('log in as an applicant and verify the program is hidden from me', async () => {
+      // Login as applicant, verify program is hidden
```
rockycodes, line 295: nitpick: you can delete this comment now since the context is represented in the `test.step` explanation

rockycodes (approved):
looks good! i left a few more comments to clean up the tests some more, but nothing blocking. please wait to merge until austin has reviewed it also though!

browser-test/src/admin/admin_program_creation.test.ts
azhangggggg: With default values, it leads this program being submitted. To test the visibility options containing `Disabled`, we need to stay on the program editing page.

browser-test/src/admin/admin_program_creation.test.ts
azhangggggg: same here re: with default values, it leads this program being submitted.

browser-test/image_snapshots/admin_program_creation_test/program-creation-page.png
azhangggggg: good idea! done.

browser-test/src/admin/admin_program_creation.test.ts
azhangggggg: done!

browser-test/src/admin/admin_program_creation.test.ts
rockycodes: Thank you for explaining!

avaleske (approved):
LGTM! Thank you @rockycodes for the quick review, too!

rough thought: We can probably remove the new tests in admin_program_creation.tests.ts after the feature flag is removed, right, since the visibility tests will fail if the Disabled option isn't present?

azhangggggg:
> LGTM! Thank you @rockycodes for the quick review, too!
> 
> rough thought: We can probably remove the new tests in admin_program_creation.tests.ts after the feature flag is removed, right, since the visibility tests will fail if the Disabled option isn't present?

yeah, I will make a note of that to remove the new tests in admin_program_creation.tests.ts after the feature flag is removed.

PR #7196, opened 2024-04-10
Skip address correction if an optional question is not answered

### LINKED ISSUE #7138: Address question cannot be optional if address correction is also selected.
**Describe the bug**
A new form has two address questions, one required question for primary address and one option question for mailing address. I have selected address correction for each address question. If you try to skip the mailing address question civiform shows and error. 

**To Reproduce**
1. Here is an image of the question that is marked with both "optional" and "address correction"

Image

2. Here is what it looks like when you try to click "save and next" without entering any data. 

Image

3. If you try to continue again by clicking "save and next" you are taken back to the original question with the error information. 

Image

4. You can click "save and next again" but it will take you through the same series of windows shown above without changes. 

**Expected behavior**
I would expect to be able to click "save and next" without entering any data in order to advance to the next question in the form. 

- OS: Windows 10 Enterprise
- Browser Firefox
- Version 22H2

Not tested on mobile.

jeffpw-goog:
### Description

An address question can be optional and also have address correction enabled. Previously address correction would run even if the user didn't enter any data, which led to an error loop that prevented the user from proceeding. Now address correction will only run if the question is answered (i.e. any address data is filled in).

#### General

Steps to reproduce are in issue #7138.

### Issue(s) this completes

Fixes #7138

### DIFF
--- a/browser-test/src/applicant/application_address_correction.test.ts
+++ b/browser-test/src/applicant/application_address_correction.test.ts
@@ -122,19 +122,19 @@ test.describe('address correction', () => {
     })
 
     await test.step('Create optional address program', async () => {
-    await adminPrograms.addProgram(optionalAddressProgram)
+      await adminPrograms.addProgram(optionalAddressProgram)
 
-    await adminPrograms.editProgramBlockWithOptional(
-      optionalAddressProgram,
-      'first block',
-      [],
-      addressWithCorrectionQuestionId,
-    )
+      await adminPrograms.editProgramBlockWithOptional(
+        optionalAddressProgram,
+        'first block',
+        [],
+        addressWithCorrectionQuestionId,
+      )
 
-    await adminPrograms.goToBlockInProgram(optionalAddressProgram, 'Screen 1')
-    await adminPrograms.clickAddressCorrectionToggleByName(
-      addressWithCorrectionQuestionId,
-    )
+      await adminPrograms.goToBlockInProgram(optionalAddressProgram, 'Screen 1')
+      await adminPrograms.clickAddressCorrectionToggleByName(
+        addressWithCorrectionQuestionId,
+      )
 
       await adminPrograms.gotoAdminProgramsPage()
       await adminPrograms.publishProgram(optionalAddressProgram)
@@ -256,8 +256,8 @@ test.describe('address correction', () => {
       await applicantQuestions.applyProgram(optionalAddressProgram)
 
       await applicantQuestions.clickNext()
-
       await applicantQuestions.expectReviewPage()
+
       await applicantQuestions.clickSubmit()
       await logout(page)
     })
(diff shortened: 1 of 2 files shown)


browser-test/src/applicant/application_address_correction.test.ts
```diff
@@ -116,6 +117,24 @@ test.describe('address correction', () => {
...
+    )
+
+    await adminPrograms.goToBlockInProgram(optionalAddressProgram, 'Screen 1')
+    await adminPrograms.clickAddressCorrectionToggleByName(
+      addressWithCorrectionQuestionId,
+    )
+
+    await adminPrograms.gotoAdminProgramsPage()
+    await adminPrograms.publishProgram(optionalAddressProgram)
+
```
gwendolyngoetz, line 137: (blocking) Wrap this with `test.step` and remove the comment.  See example [here](https://github.com/civiform/civiform/wiki/Playwright-Feature-Migration-Guide#steps) and Playwright docs [here](https://playwright.dev/docs/api/class-test#test-step). If you can to through just this `test.beforeAll` that this is in and add `test.step` to each group even better.

By using `test.step` the Playwright trace viewer and console output will better nest things. It's particularly useful in the trace viewer when you need to drill down into all the things being run.

server/app/services/applicant/question/AddressQuestion.java
```diff
@@ -281,7 +281,9 @@ public Address getAddress() {
    * the correction process.
    */
```
gwendolyngoetz, line 282: (blocking) Please update the javadoc to reflect the new behavior.

browser-test/src/applicant/application_address_correction.test.ts
```diff
@@ -226,6 +245,18 @@ test.describe('address correction', () => {
...
+    test('skips address correction if optional address question is not answered', async () => {
+      const {page, applicantQuestions} = ctx
+      await enableFeatureFlag(page, 'esri_address_correction_enabled')
+      await applicantQuestions.applyProgram(optionalAddressProgram)
+
+      await applicantQuestions.clickNext()
+
+      await applicantQuestions.expectReviewPage()
+      await applicantQuestions.clickSubmit()
+      await logout(page)
```
gwendolyngoetz, line 257: (fyi) Many of our older browser tests (still) use shared database state within the a single test file, that is it clears the database once when the file starts tests, but not between each test. You can tell if it's got
`const ctx = createTestContext(/* clearDb= */ false)` set with `false`. It's possible you may find if you run a single test in the file there may be a test that expects the database to be in a state from a change in another test.

gwendolyngoetz (changes requested):

server/app/services/applicant/question/AddressQuestion.java
jeffpw-goog: Done thanks for catching!

browser-test/src/applicant/application_address_correction.test.ts
jeffpw-goog: Done thanks!

Did the existing sections in https://github.com/civiform/civiform/pull/7221

browser-test/src/applicant/application_address_correction.test.ts
jeffpw-goog: Ack good to know!

gwendolyngoetz (approved):

gwendolyngoetz (approved):

Issue #7195, opened 2024-04-10
 [Disabled Programs] Show disabled programs from the program list for program admins


azhangggggg:
Update the controller that loads the admin program list to exclude disabled programs.
Add browser tests (from scratch, there is not currently a browser test of the program admin list view).

PR #7202, opened 2024-04-10
[Disabled Programs] refactor ActiveAndDraftPrograms

azhangggggg:
### Description

There are some duplicated code for activeNameToProgram and draftNameToProgram. For this PR, I took out a helper function to avoid them.

I also added a `Predicate<ProgramDefinition> filter` parameter for future use when we need a filter for non disabled active programs. It will be used as below.

```
ImmutableMap<String, ProgramDefinition> nonDisabledActiveNameToProgram = nameToProgram(repository, service, active, program -> program.displayMode() != DisplayMode.DISABLED);
```

#### General

### Issue(s) this completes

#7195

### DIFF
--- a/server/app/services/program/ActiveAndDraftPrograms.java
+++ b/server/app/services/program/ActiveAndDraftPrograms.java
@@ -45,6 +45,20 @@ public final class ActiveAndDraftPrograms {
     return new ActiveAndDraftPrograms(repository, Optional.empty());
   }
 
+  public ImmutableMap<String, ProgramDefinition> nameToProgram(
+      VersionRepository repository, Optional<ProgramService> service, VersionModel versionModel) {
+    ImmutableMap<String, ProgramDefinition> result =
+        repository.getProgramsForVersion(checkNotNull(versionModel)).stream()
+            .map(
+                program ->
+                    service.isPresent()
+                        ? getFullProgramDefinition(service.get(), program.id)
+                        : program.getProgramDefinition())
+            .collect(
+                ImmutableMap.toImmutableMap(ProgramDefinition::adminName, Function.identity()));
+    return result;
+  }
+
   private ActiveAndDraftPrograms(VersionRepository repository, Optional<ProgramService> service) {
     VersionModel active = repository.getActiveVersion();
     VersionModel draft = repository.getDraftVersionOrCreate();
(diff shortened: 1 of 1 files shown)


server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -45,31 +46,33 @@ public static ActiveAndDraftPrograms buildFromCurrentVersionsUnsynced(
     return new ActiveAndDraftPrograms(repository, Optional.empty());
   }
 
+  private ImmutableMap<String, ProgramDefinition> nameToProgram(
+      VersionRepository repository,
+      Optional<ProgramService> service,
+      VersionModel versionModel,
+      Predicate<ProgramDefinition> filter) {
```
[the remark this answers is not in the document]
dkatzz: +1

I also think the name of the function doesn't make it clear that any filtering is happening.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -45,31 +46,36 @@ public static ActiveAndDraftPrograms buildFromCurrentVersionsUnsynced(
...
 
+  private ImmutableMap<String, ProgramDefinition> mapNameToProgramWithFilter(
+      VersionRepository repository,
+      Optional<ProgramService> service,
+      VersionModel versionModel,
+      Optional<DisplayMode> excludeDisplayMode) {
+    return repository.getProgramsForVersion(checkNotNull(versionModel)).stream()
+        .map(
+            program ->
+                service.isPresent()
```
dkatzz, line 57: **suggestion:** Could we use ifPresentOrElse here and below? https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/util/Optional.html#ifPresentOrElse-java.util.function.Consumer-java.lang.Runnable-

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -45,31 +46,36 @@ public static ActiveAndDraftPrograms buildFromCurrentVersionsUnsynced(
...
     ImmutableMap<String, ProgramDefinition> activeNameToProgram =
-        repository.getProgramsForVersion(checkNotNull(active)).stream()
-            .map(
-                program ->
-                    service.isPresent()
-                        ? getFullProgramDefinition(service.get(), program.id)
-                        : program.getProgramDefinition())
-            .collect(
-                ImmutableMap.toImmutableMap(ProgramDefinition::adminName, Function.identity()));
+        mapNameToProgramWithFilter(repository, service, active, Optional.empty());
```
dkatzz, line 75: **suggestion:** You can create another function `mapNameToProgram` that calls `mapNameToProgramWithFilter(repository, service, active, Optional.empty());`. May make this look a little cleaner since you have an option in the cases where you don't want any filter applied. For use here and below.

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: I attempted to use `ifPresentOrElse()` in the filter one for about about an hour and couldn't really do so. Maybe I'm missing something. If you think it's doable I can give it more tries in a future PR.

server/app/services/program/ActiveAndDraftPrograms.java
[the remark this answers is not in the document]
azhangggggg: Added `excludeDisplayMode` as an optional param, as well as renaming the function to reflect the filter part. Thanks for the suggestion.

server/app/services/program/ActiveAndDraftPrograms.java
dkatzz: I'll take a look but sgtm on moving forward for now!

dkatzz (approved):

server/app/services/program/ActiveAndDraftPrograms.java
dkatzz: I actually think `ifPresent` wouldn't work in this case because it returns a void.  We could use map to do something like:

`.map(
            program ->
                service.map(svc -> getFullProgramDefinition(svc, program.id)).orElse(program.getProgramDefinition())`

But I think this is fine as is. Sorry for the confusion!

Issue #7209, opened 2024-04-11
Migrate existing program admin browser test to program admin folder 


azhangggggg:
Per the discussion [here]( https://civiform.slack.com/archives/C01QKN2UNMA/p1712785065686899), under browser test folder, currently there is folder for admin as in civiform admin, a folder for applicant, but not one for program admin. 
A folder for program admin will come out as [this PR](https://github.com/civiform/civiform/pull/7201) comes out. 
We need to migrate existing program admin browser tests under the folder for program admin.

PR #7215, opened 2024-04-11
[Disabled Programs] add tests to verify all programs are shown to program admins

azhangggggg:
### Description

Added tests for program admins, to verify that the program admins can view all the programs, including the ones with disabled visibility.

#### General

### Issue(s) this completes

#7195


browser-test/src/program_admin/program_admin_program_list.test.ts
gwendolyngoetz, line 1: (blocking) For new tests please be sure to use the fixture format. See [migration guide here](https://github.com/civiform/civiform/wiki/Playwright-Feature-Migration-Guide`).

gwendolyngoetz (changes requested):

browser-test/src/program_admin/program_admin_program_list.test.ts
azhangggggg: thanks for reminding me! adapted to the fixture format.

gwendolyngoetz (approved):

browser-test/src/program_admin/program_admin_program_list.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      await loginAsAdmin(page)
+      await adminPrograms.addProgram('Program Name One')
+      await adminPrograms.addProgram('Program Name Two')
+      await adminPrograms.addProgram('Program Name Three')
+      await adminPrograms.publishAllDrafts()
+      await logout(page)
+    })
+
+    await test.step('log in as a program admin and view the program list, verify that all active programs are shown', async () => {
+      await loginAsProgramAdmin(page)
```
avaleske, line 26: **suggestion, blocking**: It would be good to also check the page directly for the expected programs, using something like
- https://github.com/civiform/civiform/blob/60bea30f2ffa29e05716b35418a66c09d07f8ca0/browser-test/src/support/admin_programs.ts#L82
- or https://github.com/civiform/civiform/blob/60bea30f2ffa29e05716b35418a66c09d07f8ca0/browser-test/src/admin/admin_program_list.test.ts#L147

the former doesn't seem to be used, and the latter might need to be moved into support/admin_programs so you can use it, so I'm not sure which will be easier.

(here and the other test)

avaleske (changes requested):

browser-test/src/program_admin/program_admin_program_list.test.ts
azhangggggg: I think both of the functions are meant to check the program dashboard for civiform admin but not program admin. you can see it from https://github.com/civiform/civiform/blob/60bea30f2ffa29e05716b35418a66c09d07f8ca0/browser-test/src/support/admin_programs.ts#L83
and 
https://github.com/civiform/civiform/blob/60bea30f2ffa29e05716b35418a66c09d07f8ca0/browser-test/src/admin/admin_program_list.test.ts#L154

I can in the future create a folder under support for program admins. Adding it to issue #7209 to track this.

avaleske: ahh good point.

What about just a `await expect(page.getByText('Program Name One')).toHaveCount(1)`? That way there's a programatic check for the program, not just the screenshot.

(Relying solely on the screenshot could result in the test being updated incorrectly, if someone just accepts all the screenshot diffs without noticing)

azhangggggg: good point! added `await expect(programName).toHaveCount(1)` to check the programs shown on the dashboard

avaleske (approved):
Thank you!

Issue #7204, opened 2024-04-11
CiviForm admins with access to programs are not able to download files from applications


dkatzz:
**Describe the bug**
CiviForm admins that are allowed to review programs through the ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS flag, are not able to download files from programs

**To Reproduce**

1. Create a program that has a file upload question and publish
2. Fill out the program as an applicant and submit
3. Go to settings and turn on ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS
4. Login as a CiviForm admin only (not program and civiform admin)
5. Go to applications and try to download the image
6. Unauthorized error 

**Expected behavior**
Since ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS is turned on, the admin should be able to download the programs.

**Additional context**
See https://github.com/civiform/civiform/issues/7149 for more context

PR #7219, opened 2024-04-12
Allow global admin with program privileges to download applicant files

dkatzz:
### Description

CiviForm admins that are allowed to review programs through the ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS flag should be able to access files

This also adds a method in WithMockedProfiles to allow for an additional mock to be bound.

## Release notes

Allow global admin with program privileges (through ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS) to download applicant files.

#### General

1. Create a program with a file upload question and fill it out as an applicant
2. Navigate to civiform admin view with ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS on
3. Able to download file from application view

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/7204

### DIFF
--- a/server/app/controllers/FileController.java
+++ b/server/app/controllers/FileController.java
@@ -121,6 +121,8 @@ public class FileController extends CiviFormController {
     AccountModel adminAccount =
         profileUtils.currentUserProfile(request).orElseThrow().getAccount().join();
 
+    // An admin is eligible if they are a global admin with the program access flag turned on
+    // or if they have been explicitly given read permission to the program.
     return ((adminAccount.getGlobalAdmin()
                 && settingsManifest.getAllowCiviformAdminAccessPrograms(request))
             || maybeFile.get().getAcls().hasProgramReadPermission(adminAccount))
--- a/server/app/controllers/FileController.java
+++ b/server/app/controllers/FileController.java
@@ -22,6 +22,7 @@ import repository.StoredFileRepository;
 import repository.VersionRepository;
 import services.cloud.ApplicantFileNameFormatter;
 import services.cloud.ApplicantStorageClient;
+import services.settings.SettingsManifest;
 
 /** Controller for handling methods for admins and applicants accessing uploaded files. */
 public class FileController extends CiviFormController {
@@ -30,6 +31,7 @@ public class FileController extends CiviFormController {
   private final HttpExecutionContext classLoaderExecutionContext;
   private final ApplicantStorageClient applicantStorageClient;
   private final StoredFileRepository storedFileRepository;
+  private final SettingsManifest settingsManifest;
 
   @Inject
   public FileController(
@@ -37,11 +39,13 @@ public class FileController extends CiviFormController {
       StoredFileRepository storedFileRepository,
       ApplicantStorageClient applicantStorageClient,
       ProfileUtils profileUtils,
-      VersionRepository versionRepository) {
+      VersionRepository versionRepository,
+      SettingsManifest settingsManifest) {
     super(profileUtils, versionRepository);
     this.classLoaderExecutionContext = checkNotNull(classLoaderExecutionContext);
     this.applicantStorageClient = checkNotNull(applicantStorageClient);
     this.storedFileRepository = checkNotNull(storedFileRepository);
+    this.settingsManifest = checkNotNull(settingsManifest);
   }
 
   @Secure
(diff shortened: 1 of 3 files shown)


server/app/controllers/FileController.java
```diff
@@ -117,7 +121,9 @@ private Result adminShowInternal(Request request, String fileKey) {
     AccountModel adminAccount =
         profileUtils.currentUserProfile(request).orElseThrow().getAccount().join();
 
-    return maybeFile.get().getAcls().hasProgramReadPermission(adminAccount)
+    return ((adminAccount.getGlobalAdmin()
```
gwendolyngoetz, line 124: (non-blocking) Maybe add a comment on the business rules explaining what this is checking.

server/test/controllers/FileControllerTest.java
```diff
@@ -13,18 +14,22 @@
...
 public class FileControllerTest extends WithMockedProfiles {
 
   private FileController controller;
+  private SettingsManifest mockSettingsManifest;
   private final Request request = fakeRequest().build();
 
   @Before
   public void setUp() {
     resetDatabase();
+    mockSettingsManifest = Mockito.mock(SettingsManifest.class);
```
gwendolyngoetz, line 32: (praise) Thank you. We have too many tests relying on settings directly from the conf file.

server/test/controllers/FileControllerTest.java
```diff
@@ -164,6 +171,21 @@ public void acledAdminShow_globalAdmin_returnsUnauthorizedResult() {
...
+    createStoredFileWithProgramAccess(fileKey, program);
+    String encodedFileKey = encodefakeFileKey(fileKey);
+
+    Result result = controller.acledAdminShow(request, encodedFileKey);
+
+    assertThat(result.status()).isEqualTo(SEE_OTHER);
+  }
+
   @Test
   public void acledAdminShow_redirects() {
```
gwendolyngoetz, line 190: (blocking) It's unclear from the test name if this is now civiform admin or program admin. Based on the code looks like program admin, but can you update the test method name to explicitly state it.

server/test/controllers/FileControllerTest.java
```diff
@@ -154,7 +159,9 @@ public void acledAdminShow_differentFileKey_returnsNotFound() {
   }
 
   @Test
-  public void acledAdminShow_globalAdmin_returnsUnauthorizedResult() {
+  public void acledAdminShow_globalAdminWithoutPrivledges_returnsUnauthorizedResult() {
+    when(mockSettingsManifest.getAllowCiviformAdminAccessPrograms(request)).thenReturn(true);
```
gwendolyngoetz, line 163: (question) If the `AllowCiviformAdminAccessPrograms` setting is returning `true` and the user is a global admin (line 166) why is the status `unauthorized`?

server/app/controllers/FileController.java
dkatzz: Good call - added!

server/test/controllers/FileControllerTest.java
```diff
@@ -164,6 +171,21 @@ public void acledAdminShow_globalAdmin_returnsUnauthorizedResult() {
...
+    createProgramAdminWithMockedProfile(program);
+    String fileKey = fakeFileKey(1L, program.id);
+    createStoredFileWithProgramAccess(fileKey, program);
+    String encodedFileKey = encodefakeFileKey(fileKey);
+
+    Result result = controller.acledAdminShow(request, encodedFileKey);
+
+    assertThat(result.status()).isEqualTo(SEE_OTHER);
+  }
+
```
gwendolyngoetz, line 188: (blocking) We should probably have a test that mocks the return value to false and not rely on the config defaults. I'm not sure if that should be a whole separate test or just explicitly setting `acledAdminShow_redirects` with the mock. What do you think?

gwendolyngoetz (changes requested):

server/test/controllers/FileControllerTest.java
dkatzz: Good call! Updated

server/test/controllers/FileControllerTest.java
dkatzz: Wow good catch! I had written this wrong but fixing it made me realize it wasn't actually using settings manifest. Added a method to WithMockedProfiles to get this working, but let me know if you think it would be better to do this another way.

server/test/controllers/FileControllerTest.java
dkatzz: Updated the return value to false by default (unless otherwise specified). Let me know what you think.

dkatzz: Added a few other tests with the global admin role as well.

server/test/controllers/FileControllerTest.java
gwendolyngoetz: This is great! 

We absolutely should be mocking the settings to be what we expect them to be and not the config file. Thanks!!

gwendolyngoetz (approved):

PR #7226, opened 2024-04-15
Automatically enable git completion in codespaces

jeffpw-goog:
### Description

Add a postCreateCommand to devcontainer.json to automatically enable git completion in codespaces.

#### General

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -67,6 +67,7 @@
 
   "onCreateCommand": "npm install -g @withgraphite/graphite-cli@stable && npm install --no-save @playwright/test",
   "updateContentCommand": "bin/vscode-setup",
+  "postCreateCommand": "echo 'source /usr/share/bash-completion/completions/git' >> ~/.bashrc",
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   "forwardPorts": [9000, 5432, 3390]


dkatzz (approved):
We can also consider adding this into the bin/vscode-setup also, but good with this too. If we end up adding more commands like this, we can also create another script bin/post-setup or something

gwendolyngoetz:
> We can also consider adding this into the bin/vscode-setup also, but good with this too. If we end up adding more commands like this, we can also create another script bin/post-setup or something

Please don't put it in `bin/vscode-setup` or any other scripts in `bin` that may be run outside the devcontainer. We shouldn't be modifying anyone's `.bashrc` or other bash/zsh config files.

If you want to put it in a script let's create a `.devcontainer/bin` folder and put devcontainer specific scripts in there.

dkatzz:
> > We can also consider adding this into the bin/vscode-setup also, but good with this too. If we end up adding more commands like this, we can also create another script bin/post-setup or something
> 
> Please don't put it in `bin/vscode-setup` or any other scripts in `bin` that may be run outside the devcontainer. We shouldn't be modifying anyone's `.bashrc` or other bash/zsh config files.
> 
> If you want to put it in a script let's create a `.devcontainer/bin` folder and put devcontainer specific scripts in there.

Good call! SGTM! Can create another script if we end up adding more of these

PR #7237, opened 2024-04-16
[Disabled Programs] add the number of disabled programs added in question list for civiform admin

### LINKED ISSUE #7238: [Disabled Programs] Adding phrase "Added to # disabled programs" in question list for civiform admins
Adding phrase "Added to # disabled programs" in question list for civiform admins

"#" is the number of [disabled programs that are in draft and disabled] that this question has been added to.

It should look similar to the screenshot below.
Screenshot 2024-04-18 at 3 32 44 PM

azhangggggg:
### Description

Currently if a question is added to 2 programs in draft and one of the programs is disabled, it would show `Added to 2 programs` in the question referencing the program.

With this PR, same situation (the questions is added to 2 programs in draft, and one of the programs is disabled), it would show `Added to 1 disabled program` and `Added to 2 programs.`

With the next PR (not this one), same situation (the questions is added to 2 programs in draft, and one of the programs is disabled), it will eventually show `Added to 1 disabled program` and `Added to 1 program in use`

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #7238


server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -500,6 +520,8 @@ private GroupedReferencingPrograms createReferencingPrograms(
     Set<String> usedSet = Sets.intersection(activeProgramsMap.keySet(), draftProgramsMap.keySet());
     Set<String> addedSet = Sets.difference(draftProgramsMap.keySet(), activeProgramsMap.keySet());
     Set<String> removedSet = Sets.difference(activeProgramsMap.keySet(), draftProgramsMap.keySet());
+    Set<String> disabledSet =
+        Sets.difference(draftDisabledProgramsMap.keySet(), activeProgramsMap.keySet());
```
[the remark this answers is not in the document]
azhangggggg: Haha feels the same about coffee. Thanks for the comment :)

`newlyDisabledPrograms` might cause some confusion imo - say program A is disabled already (which still makes program A an active program just with disabled visibility), then we add a question to program A, it still belongs to `disabledSet` and the phrase should show `Added to 1 disabled program`. 

The criteria here is for the program to be disabled and in draft, regardless if the program was disabled before. 

Added some comments on where `disabledPrograms` are set to clarify this.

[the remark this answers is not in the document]
gwendolyngoetz: Gotcha. Makes more sense to me now. Thanks

browser-test/src/admin/admin_question_list.test.ts
```diff
@@ -159,6 +160,61 @@ test.describe('Admin question list', () => {
...
+    )
+
+    await adminQuestions.gotoAdminQuestionsPage()
+    await validateScreenshot(
+      page,
+      'questions-list-sort-dropdown-disabled-programs',
+    )
+    await adminQuestions.expectQuestionProgramReferencesText({
+      questionName: 'a',
+      expectedProgramReferencesText: 'Used in 2 programs.',
```
rockycodes, line 203: question (blocking): Can you confirm with product (Isabel and Sijia) that we don't want to separate out the disabled program count for "Used in 1 program"? It seems confusing to me to separate out the count for "Added to" but not "Used in". The screenshots in your TDD suggested it would be separated out for "Used in", but I'm seeing now those screenshots might be confusing both cases. I went back and read through the PRD and it wasn't clear there either, so I think we need clarification from product.

rockycodes (changes requested):

browser-test/src/admin/admin_question_list.test.ts
azhangggggg: Definitely I will confirm with product about the phrase and include it if a PR for   `Used in # programs` is needed.
But I don't think it should be blocking this PR. As the description suggested that this PR is only about the phrase `Added to # disabled programs`.

browser-test/src/admin/admin_question_references.test.ts
```diff
@@ -121,5 +120,52 @@ test.describe(
         await validateScreenshot(page, 'question-program-modal')
       })
     })
+
+    test('shows results for referencing disabled program', async ({
```
rockycodes, line 124: suggestion (non-blocking): Edit the test name to explain what the results shown should be. Something like "shows disabled programs listed separately"

browser-test/src/admin/admin_question_references.test.ts
```diff
@@ -121,5 +120,52 @@ test.describe(
...
+    }) => {
+      await enableFeatureFlag(page, 'disabled_visibility_condition_enabled')
+      const programName = 'Program name'
+      const disabledProgramName = 'Disabled program name'
+      const questionName = 'question-references-q'
+
+      await loginAsAdmin(page)
+      await adminQuestions.addAddressQuestion({questionName})
+
+      await test.step(`Create two programs and add ${questionName}`, async () => {
```
rockycodes, line 137: praise: super clear test!

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -461,12 +475,21 @@ abstract static class GroupedReferencingPrograms {
...
     static Builder builder() {
       return new AutoValue_QuestionsListView_GroupedReferencingPrograms.Builder();
     }
 
     boolean isEmpty() {
-      return usedPrograms().isEmpty() && addedPrograms().isEmpty() && removedPrograms().isEmpty();
+      return usedPrograms().isEmpty()
+          && addedPrograms().isEmpty()
+          && removedPrograms().isEmpty()
+          && disabledPrograms().isEmpty();
```
rockycodes, line 492: suggestion (maybe blocking): I think you want to wrap this `isEmpty` check in a feature flag check as well so that the overall `isEmpty` check isn't affected.

rockycodes (changes requested):
looks good! just a few suggestions.

server/app/views/admin/questions/QuestionsListView.java
azhangggggg: I don't think adding a feature flag check here is necessary. With the feature flag default being disabled, `disabledPrograms` is always empty. 
It won't affect isEmpty() since the logic under this function is to check `usedPrograms`, `addedPrograms` and `removedPrograms`.

But happy to discuss further if you think otherwise!

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -500,6 +532,8 @@ private GroupedReferencingPrograms createReferencingPrograms(
     Set<String> usedSet = Sets.intersection(activeProgramsMap.keySet(), draftProgramsMap.keySet());
```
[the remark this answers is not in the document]
azhangggggg: not a nit haha. updated.

browser-test/src/admin/admin_question_references.test.ts
azhangggggg: updated the test name. Let me know what you think.

server/app/views/admin/questions/QuestionsListView.java
rockycodes: I'm thinking of the situation where we're testing out the feature. Say a CE turns it on in production sets some programs to "Disabled", but then we realize there's a bug and they switch it off again. If we don't guard this check, then any programs that were set to Disabled while the flag was enabled will be counted and this check would return `false` when it should return `true`.

This check is used on line 429:
```
if (groupedReferencingPrograms.isEmpty()) {
      tag.with(p("Used in 0 programs."));
    }
```

It's also used on line 570 (although I'm not sure what the "referencing modal" is actually for):
```
    if (referencingPrograms.isEmpty()) {
      return Optional.empty();
    }
```

My brain is feeling noodley trying to parse these sets though, so let me know if I'm confused about how this is working.

browser-test/src/admin/admin_question_list.test.ts
rockycodes: thanks! please remove the snapshot it generated also.

server/app/views/admin/questions/QuestionsListView.java
azhangggggg: Say a CE turns it on in production sets some programs to "Disabled", but then we realize there's a bug and they switch it off again.  - I think that's a valid edge case. 

Wrapped isEmpty() with the feature flag. Thanks!

To answer the question, makeReferencingProgramsModal is the modal that gets generated once the user clicks on the phrase `added to # programs`.

rockycodes: Thank you for explaining!

rockycodes (approved):

gwendolyngoetz (approved):

PR #7248, opened 2024-04-17
Markdown fixes - fix lists in preview and when adding required indicator

### LINKED ISSUE #7240: Lists are formatted incorrectly in question preview
**Describe the bug**
Lists are not being formatted correctly by markdown in question preview:

image

rockycodes:
### Description

This change fixes the current markdown preview logic so that it will correctly render lists. It also fixes a bug where there markdown parser would break if the question text only contained a list.

#### General

1. Create multiple questions with different combinations of list markdown (Eg. paragraph before a list, paragraph after a list, list alone, no list)
2. See that the preview correctly renders the list markdown
3. Add the questions to a program and publish it.
4. Apply to the program as an applicant and see that the required indicators are rendered correctly:
- If a question contains a paragraph and then a list, the required indicator should come at the end of a paragraph before the list
- If a question contains only a list, the required indicator should appear next to the last item on the list
- Otherwise, the required indicator should come at the end of the question text

image

### Issue(s) this completes

Fixes #7240


server/app/views/components/TextFormatter.java
```diff
@@ -98,16 +98,51 @@ private static String addTextSize(String markdownText) {
...
+      return handleRequiredQuestionsThatEndInAList(markdownText, "<ul");
+    } else if (endsWithListTag(markdownText, "</ol>\n")) {
+      return handleRequiredQuestionsThatEndInAList(markdownText, "<ol");
+    }
+
+    // If the question doesn't end with a list, add the required indicator on to the end
     int indexOfClosingTag = markdownText.lastIndexOf("</");
+    return buildStringWithRequiredIndicator(markdownText, indexOfClosingTag);
+  }
+
```
rockycodes, line 113: The logic for where to render the required indicator felt confusing, so I broke it out into a bunch of mini methods. Lmk if you think this is overkill though and it would be simpler to contain the logic in fewer methods.

server/app/assets/javascripts/util.ts
```diff
@@ -73,7 +73,14 @@ export function formatTextHtml(text: string): Element {
...
 export function formatText(text: string): string {
   // Preserve line breaks before parsing the text
-  text = text.split('\n').join('<br>')
+  const textArray = text.split('\n')
+  for (let i = 0; i < textArray.length; i++) {
+    if (!textArray[i]) {
+      textArray[i] = '\n'
+    }
+  }
+  // the non-breaking space character (&nbsp;) is required for the markdown parser to respect blank lines
```
melanie-exygy, line 82: praise:  Thanks for the helpful comment!  I think it could help prevent others from accidentally changing this code in the future.  ;-)

server/app/views/components/TextFormatter.java
```diff
@@ -98,16 +98,51 @@ private static String addTextSize(String markdownText) {
   }
 
   private static String addRequiredIndicator(String markdownText) {
+    // If the question ends with a list (UL or OL tag), we need to handle the required indicator
+    // differently
+    if (endsWithListTag(markdownText, "</ul>\n")) {
```
melanie-exygy, line 103: question (non-blocking):  Would it be possible for the markdown text to end with a closing list tag but no new line?  If it's not possible, can you help me understand why there's always a new line at the end?

server/app/views/components/TextFormatter.java
melanie-exygy: Having shorter methods seems nice to me.

server/app/views/components/TextFormatter.java
```diff
@@ -98,16 +98,51 @@ private static String addTextSize(String markdownText) {
...
+    return buildStringWithRequiredIndicator(markdownText, indexOfClosingTag);
+  }
+
+  private static boolean endsWithListTag(String markdownText, String closingListTag) {
+    int indexOfClosingListTag = markdownText.lastIndexOf(closingListTag);
+    return indexOfClosingListTag > -1
+        && markdownText.substring(indexOfClosingListTag).equals(closingListTag);
+  }
+
+  private static String handleRequiredQuestionsThatEndInAList(
```
melanie-exygy, line 120: praise:  I really like the clarity of your method names.  Very helpful!

server/test/views/components/TextFormatterTest.java
```diff
@@ -82,21 +82,67 @@ public void rendersRequiredIndicator() {
...
+  public void insertsRequiredIndicatorBeforeLists() {
+    ImmutableList<DomContent> contentWithUnorderedList =
         TextFormatter.formatText(
             "Here is some text.\n" + "* list item one\n" + "* list item two",
             /* preserveEmptyLines= */ false,
             /* addRequiredIndicator= */ true);
-    String htmlContent = content.get(0).render();
+    String htmlContentWithUnorderedList = contentWithUnorderedList.get(0).render();
+
+    assertThat(htmlContentWithUnorderedList)
```
melanie-exygy, line 93: question (non-blocking):  Why makes separate assertions rather than a single `equals` assertion with the full html content of the static question?  Seems like that would have the added benefit of verifying that everything is in the right place.

browser-test/image_snapshots/admin_program_creation_test/program-creation-static-question-with-formatting.png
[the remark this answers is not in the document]
rockycodes: Yeah they should. In the old one they weren't actually a list (the bullet points are still rendered as *)

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -199,6 +199,29 @@ test.describe('program creation', () => {
...
+        '\n' +
+        'Here is some more text after a blank line\n' +
+        '\n' +
+        '\n' +
+        'Here is more text after more blank lines',
+    })
+
+    await page.waitForTimeout(100) // ms
+    await validateScreenshot(
+      page,
```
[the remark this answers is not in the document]
rockycodes: I love this idea!

server/app/views/components/TextFormatter.java
rockycodes: Both markdown parsers (the java one and the javascript one) add new line characters at the end of block elements (paragraphs, lists, etc.). Maybe all markdown parsers do this? I'm guessing this is so the next block element starts below the last one. New line characters aren't inserted after inline elements (em, strong, span), but those are always within block elements.

server/test/views/components/TextFormatterTest.java
rockycodes: That's a great question. I remember there was something tricky about getting the syntax right for checking it all at once, but I'll try again. If it doesn't work, I'll add a comment explaining why it's tested like this.

rockycodes: Welp, I'm not sure what I was doing before but changing these to `isEqualTo` was super easy! Updated.

browser-test/src/admin/admin_program_creation.test.ts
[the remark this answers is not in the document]
rockycodes: Done

server/app/views/components/TextFormatter.java
melanie-exygy: Got it.  Thanks for explaining.

melanie-exygy (approved):

PR #7280, opened 2024-04-19
Putting TI Account Settings on their own page and adding navigation

### LINKED ISSUE #6210: Add in page navigation tabs for Client List and Account Settings for TI dashboard
**Describe the solution**
To create a less busy TI dashboard, we'd like the Client List and Account Settings views to be on two separate sub-pages which can be accessed via a tab menu.

Mocks for the whole page:
https://www.figma.com/file/yfhnUmE2FsonLtAqfeEfoo/TI-Edit?type=design&node-id=816-494&mode=design&t=RJY2E2vZvOISPKr4-0

Mocks for the tab menu:
https://www.figma.com/file/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?type=design&node-id=1740-20643&mode=design&t=MV2sVlWhV3JxGRvq-0

image

Note that the "Contact Us" tab does not yet need to be implemented.

**Additional context**
Here is the USWDS header component, which has a tab menu as a nav bar:  [USWDS header](https://designsystem.digital.gov/components/header/).

Another option is to use HTMX.  We could implement this as two separate pages, but it would probably be more performant to use HTMX to dynamically load each view as a tab pane:  [htmx tabs](https://htmx.org/examples/tabs-hyperscript/).

**Done when**
- There is a tab menu on the TI dashboard.
- The selected tab is underlined, as in the image above.
- Clicking the "Account Settings" tab takes you to the account settings and the Organization Members table.
- Clicking the "Client List" tab takes you back to the All Clients section.

melanie-exygy:
### Description

This PR separates the TI Account Settings onto their own page which can be accessed via navigation tabs.  The navigation tabs also show on the Add Client and Edit Client views.  It also makes a few adjustments to the header band heights so that these match on all the TI pages.

The original implementation of these "tabs" used HTMX with the idea that hot-swapping the tab content would be faster than loading a new page when going to the Account Settings.  However, I realized that, rather than being a small portion of the page, the tab content was everything on the page other than the header.  This use case didn't warrant the extra complexity of HTMX. 

image

#### General

1. Log in as a TI
2. Notice the "Client list" and "Account settings" navigation tabs.
3. The "Client list" tab should be underlined and bold since it's the current tab.
4. Hover over the "Account settings" tab and notice that it is blue and underlined.
5. Click on "Account settings".
6. Note that you are directed to the Organization members table.
7. Check that the current tab is now "Account settings".
8. Click on "Client list" tab.
9. Click on the "Add new client " button.  Note that the tabs are there and "Client list" is the current tab.
10. Create a new client if you don't already have any.
11. Click on "Client list" tab.
12. In the client list, click on "Edit" next to the client name.
13. Once on the edit page, notice the tabs and that 'Client list' is the current tab.

### Issue(s) this completes

Fixes #6210


browser-test/image_snapshots/pagination_test/ti-pagination-ellipses-left.png
melanie-exygy, line 1: For all of these pagination screenshots, the diff is apparently in the "<" and ">" icons.  I can't actually see the difference, even when I zoom in, so it must be an imperceptible difference.

browser-test/image_snapshots/trusted_intermediary_test/org-members-table-many.png
melanie-exygy, line 1: Not sure what caused the height to change by 1 pixel here, but I don't think it makes a difference.

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+    if (flash.get("error").isPresent()) {
+      LoggerFactory.getLogger(TrustedIntermediaryGroupListView.class)
+          .info(request.flash().get("error").get());
+      bundle.addToastMessages(
+          ToastMessage.errorNonLocalized(flash.get("error").get()).setDuration(-1));
+    }
+    return layout.renderWithNav(request, personalInfo, messages, bundle, currentTisApplicantId);
+  }
+
+  private DivTag renderTIMembersTable(TrustedIntermediaryGroupModel tiGroup, Messages messages) {
```
melanie-exygy, line 84: These methods were just moved from `TrustedIntermediaryDashboardView.java` to here without any changes.

browser-test/image_snapshots/pagination_test/ti-pagination-ellipses-left.png
gwendolyngoetz: It's about 1-2px higher on the new version

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+import services.MessageKey;
+import services.applicant.ApplicantPersonalInfo;
+import views.BaseHtmlView;
+import views.HtmlBundle;
+import views.admin.ti.TrustedIntermediaryGroupListView;
+import views.components.ToastMessage;
+import views.style.BaseStyles;
+import views.style.ReferenceClasses;
+
+public class TIAccountSettingsView extends BaseHtmlView {
```
dkatzz, line 39: **suggestion:** Seems like the other pages spell out "TrustedIntermediary" or do pascal casing with TI to be "Ti". We should probably make this consistent with one of those standards

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+import play.twirl.api.Content;
+import services.MessageKey;
+import services.applicant.ApplicantPersonalInfo;
+import views.BaseHtmlView;
+import views.HtmlBundle;
+import views.admin.ti.TrustedIntermediaryGroupListView;
+import views.components.ToastMessage;
+import views.style.BaseStyles;
+import views.style.ReferenceClasses;
+
```
dkatzz, line 38: **suggestion:** Can you add a javadoc for this class?

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+  public Content render(
+      TrustedIntermediaryGroupModel tiGroup,
+      ApplicantPersonalInfo personalInfo,
+      Http.Request request,
+      Messages messages,
+      Long currentTisApplicantId) {
+    HtmlBundle bundle =
+        layout
+            .getBundle(request)
+            .setTitle("CiviForm")
```
dkatzz, line 59: **thought:** I wonder if we should make this title more specific?

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+                    .withData("testid", "org-members-email")
+                    .withClass("w-2/5"))
+            .with(
+                th(messages.at(MessageKey.ACCT_STATUS_LABEL.getKeyName()))
+                    .withScope("col")
+                    .withData("testid", "org-members-status")
+                    .withClass("w-1/5")));
+  }
+
+  private TrTag renderTIRow(AccountModel ti) {
```
dkatzz, line 118: **suggestion:** I know this is copied over, but similar comment to above on making the casing consistent. May be worth changing it to renderTiRow?

server/app/views/applicant/TrustedIntermediaryDashboardView.java
```diff
@@ -447,42 +412,55 @@ private DivTag renderIndexPageLink(AccountModel applicant, Messages messages) {
...
+    boolean isClientList = currentTab == TabType.CLIENT_LIST;
+    boolean isAccountSettings = currentTab == TabType.ACCOUNT_SETTINGS;
+    return nav(
+            renderClientListLink(messages, baseUrl)
+                .withClasses(
+                    "mr-10",
+                    "py-4",
+                    iffElse(isClientList, "border-b-4", "hover:border-b-4"),
+                    iffElse(isClientList, "border-blue-600", "hover:border-blue-600"),
+                    iffElse(isClientList, "text-bold", "hover:text-blue-600"))
```
dkatzz, line 453: **suggestion:** We could consider factoring this shared logic into a separate function, since its similar / the same below?

dkatzz (commented):
A little bit separate from the review, but was this where we were thinking of using htmx? I think it may be helpful to share any learnings from that and maybe include in the description why we decided to implement it this way instead.

server/app/views/applicant/TIAccountSettingsView.java
```diff
@@ -0,0 +1,144 @@
...
+      Http.Request request,
+      Messages messages,
+      Long currentTisApplicantId) {
+    HtmlBundle bundle =
+        layout
+            .getBundle(request)
+            .setTitle("CiviForm")
+            .addMainContent(
+                h1(tiGroup.getName()).withClasses(BaseStyles.TI_HEADER_BAND_H1),
+                TrustedIntermediaryDashboardView.renderTabButtons(
```
rockycodes, line 62: suggestion (maybe blocking): It's a bit of a refactor and I'm open to being convinced otherwise, but what would you think about making a parent class that is `TrustedIntermediaryDashboardView` that has `TabType`, `renderTabButtons` and any other shared logic on it? Then you could have two child classes (`TiAccountSettingsView` and `TiClientListView`... or something) that extend the parent class and only implement the parts that should be different.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -961,13 +964,56 @@ test.describe('Trusted intermediaries', {tag: ['@uses-fixtures']}, () => {
...
     })
   })
 
+  test.describe('ti navigation', () => {
+    test('marks the correct tab as current and matches screenshot', async ({
+      page,
+      tiDashboard,
+    }) => {
+      await loginAsTrustedIntermediary(page)
+      await tiDashboard.gotoTIDashboardPage(page)
```
rockycodes, line 983: suggestion (non-blocking): use `test.step` to break these up into logical chunks.

rockycodes (changes requested):
I left some non-blocking questions/comments. The biggest one is a suggestion for a refactor of the views. I second Danielle's suggestion to make the Ti casing consistent.

melanie-exygy:
> A little bit separate from the review, but was this where we were thinking of using htmx? I think it may be helpful to share any learnings from that and maybe include in the description why we decided to implement it this way instead.

Yes, I originally implemented this with htmx.  Good idea to share why I decided to implement it this way.  I'll add a paragraph to the description and share what I learned somewhere else (not sure exactly where yet).

browser-test/image_snapshots/pagination_test/ti-pagination-ellipses-left.png
melanie-exygy: Ah, I see.  Is it okay with you if I keep the change?

server/app/views/applicant/TIAccountSettingsView.java
melanie-exygy: I don't like using sentence-case for a 2-letter acronym, so I've renamed without using "TI" at all.  This also ends up being more consistent with other methods in the class.  For the method above, I renamed it `renderOrgMembersTableRow`.

browser-test/image_snapshots/pagination_test/ti-pagination-ellipses-left.png
gwendolyngoetz: Without having the two options side-by-side no one will notice a difference. I think I like it more this way having it centered between the cap height and baseline.

If anyone complains we'll just gaslight them and say it's always been this way.

server/app/views/applicant/TIAccountSettingsView.java
melanie-exygy: Good callout.  I made the title for this page and for the client list page more specific and also added it to our i18n messages file.

server/app/views/applicant/TrustedIntermediaryDashboardView.java
melanie-exygy: I used your idea, Rocky, and created a couple of `BaseStyles`.

server/app/views/applicant/TIAccountSettingsView.java
melanie-exygy: Done.

rockycodes (approved):

PR #7282, opened 2024-04-19
Refactor ProgramPredicateConfigureView.renderValueRow

jeffpw-goog:
### Description

The ProgramPredicateConfigureView.renderValueRow method had forked logic based on maybeAndNode.isPresent. Instead of duplicating the logic, we can just create an empty map if maybeAndNode is not present, and handle both cases inline, so the logic is only happening once and it's easier to see how the two cases differ.

#### General


swatkat1 (approved):
Nice!

PR #7287, opened 2024-04-22
Remove jobs from the db that are not found in the job registry

### LINKED ISSUE #6897: Delete scheduler for FixApplicantDobDataPathJob
Before we can remove the code that runs `FixApplicantDobDataPathJob`, we have to let the existing jobs run and remove the scheduler so no new jobs are scheduled. Once this is released and all deployments are on the release version with this change, we can proceed to the next step which is deleting the code that runs the job + tests. (I learned the hard way in https://github.com/civiform/civiform/issues/6284 that doing this all in one go results in an error).

rockycodes:
### Description

We need a way to easily remove Durable Jobs from our codebase, but currently an error is thrown when you remove a job from the registry because the db expects the code to be there to run the job. Both the job runner and the job scheduler reference the same registry, so it doesn't work to remove the job from just the scheduler to prevent future runs.

After talking it over with @nb1701 , we think it is safe to remove a job from the db if it is no longer in the registry. The only way a job can get into the db is through the registry, so it stands to reason if it is no longer in the registry it should no longer be in the db.

This PR deletes the job from the db when it is not also found in the registry and lowers the log level from error to info when it does so.

#### General

1. Set the time resolver for the job to 5 minutes from now
2. Run the server to add the job(s) to db (it will schedule a bunch of them)
3. Remove the logic registering the job job from `provideDurableJobRegistry` in `DurableJobModule`
4. Restart the server
5. Wait 5 minutes then check the db to see that the scheduled jobs were deleted

### Issue(s) this completes

Fixes #6897

### DIFF
--- a/server/test/durablejobs/DurableJobRunnerTest.java
+++ b/server/test/durablejobs/DurableJobRunnerTest.java
@@ -1,7 +1,6 @@
 package durablejobs;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertThrows;
 
 import annotations.BindingAnnotations;
--- a/server/test/durablejobs/DurableJobRunnerTest.java
+++ b/server/test/durablejobs/DurableJobRunnerTest.java
@@ -147,6 +147,8 @@ public class DurableJobRunnerTest extends ResetPostgres {
     PersistedDurableJobModel job = createPersistedJobToExecute();
     durableJobRunner.runJobs();
 
+    // The job should have been deleted since it does not exist in the registry.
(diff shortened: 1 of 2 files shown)


server/test/durablejobs/DurableJobRunnerTest.java
```diff
@@ -140,21 +142,30 @@ public void runJobs_runsJobsThatAreReady() {
   }
 
   @Test
-  public void runJobs_jobNotFound() {
+  public void runJobs_jobNotFound_deletesJobFromDb() {
```
rockycodes, line 145: I got turned around trying to mock out the call to `delete` to test the other code path (when delete fails). If you think we need a test for that maybe we can pair on it.

PR #7289, opened 2024-04-22
Redirect home on unauthorized

### LINKED ISSUE #7279: Add error screen for unauthorized
There is an edge case when:
- the applicant has logged in
- we change the application secret & restart the server.
- the applicant clicks on something such as _save and next_ 
- the applicant get taken to a generic Play Framework unauthorized error.

Based on Seattle's last four weeks of production logs this error doesn't actually get reached. Typically, within the application, we already have code to handle errors before they reach that point.

When this error page is reached, this generic builtin page has no Civiform UI at all to navigate back to the home page. Need plug into the framework so we can show our own page, similar to other errors. 

The actual error is 403 unauthorized. We don't typically show those kind of details in error messages.

This is needed for the Play 2.9 upgrade or we'll have a lot of calls/emails from users when their session is logged out on servers changing their app secrets.

[Start here](https://github.com/civiform/civiform/blob/main/server/app/controllers/ErrorHandler.java) 

[Similar example](https://github.com/civiform/civiform/pull/4860).

[Play 2.9 Upgrade](https://github.com/civiform/civiform/issues/5647)

gwendolyngoetz:
### Description

Normally the application handles 403s before this and redirects home. This is a failsafe for edge cases that would leave the user on a grey 403 page that is just builtin to Play. To be consistent with the rest of the app we redirect home.

This only known way to hit this is to have an active session, change the application secret, and try using your existing session.

#### General

#### In Dev Mode

- `bin/run-dev`
- start app as guest, fill out a page or two of data
- change secret
- reload
- wait for app to be fully loaded
- click save and next on open page
- get the 403 error, along with the warn/error in the console log

#### In Prod Mode

- runProd
- start app as guest, fill out a page or two of data
- stop app
- change secret
- runProd
- wait for app to be fully loaded
- click save and next on open page
- get the 403 error, along with the warn/error in the console log

 ### Issue(s) this completes

Fixes #7279

### DIFF
--- a/server/app/controllers/ErrorHandler.java
+++ b/server/app/controllers/ErrorHandler.java
@@ -126,6 +126,20 @@ public class ErrorHandler extends DefaultHttpErrorHandler {
     return Optional.empty();
   }
 
+  /**
+   * Normally the application handles 403s before this and redirects home. This is a failsafe for
+   * edge cases that would leave the user on a grey 403 page that is just builtin to Play. To be
+   * consistent with the rest of the app we redirect home.
+   *
+   * <p>This only known way to hit this is to have an active session, change the application secret,
+   * and try using your existing session.
+   */
+  @Override
+  protected CompletionStage<Result> onForbidden(RequestHeader request, String message) {
+    return CompletableFuture.completedFuture(
+        Results.redirect(controllers.routes.HomeController.index().url()));
+  }
+
   @Override
   public CompletionStage<Result> onNotFound(RequestHeader request, String message) {
     return CompletableFuture.completedFuture(


jeffpw-goog (commented):
For my edification, can you elaborate on "to be consistent with the rest of the app"? Given that this 403 case is generally unexpected, it seems like we're losing something by redirecting to home with no indication of what went wrong. In real life I get really annoyed when a website does that. :) Would it make sense / be possible to add a toast or something on the homepage even just to say "something unexpected went wrong, sorry!"?

jeffpw-goog:
> > For my edification, can you elaborate on "to be consistent with the rest of the app"? Given that this 403 case is generally unexpected, it seems like we're losing something by redirecting to home with no indication of what went wrong. In real life I get really annoyed when a website does that. :) Would it make sense / be possible to add a toast or something on the homepage even just to say "something unexpected went wrong, sorry!"?
> 
> Agreed it's annoying.
> 
> There are other parts of the application that already just redirect home. I don't like it, but I also want to be consistent with the way it's done throughout the app. There's a [long thread on slack](https://civiform.slack.com/archives/C01QKN2UNMA/p1713556144727979) you can peruse.
> 
> Retrofitting the behavior to an error message is outside the scope of what I need this for and have time for right now. The change in behavior probably make more sense to get connected with the Northstar upgrades anyway.
> 
> Also, based on the data I have from Seattle prod unhandled 403s never get logged because we typically have them handled. It's an edge case with changing the application secret value which is how I'm triggering it.

Thanks for the context! Of course I saw the beginning of that thread in Slack but not the rest of it.

jeffpw-goog (approved):

PR #7298, opened 2024-04-23
[Disabled Programs] add disabled visibility case

azhangggggg:
### Description

## Release notes

#### General

#### Database evolutions

- [ ] Assigned two reviewers

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -387,6 +387,9 @@ private LiTag renderPublishModalProgramItem(
       ProgramDefinition program, ImmutableList<Long> universalQuestionIds) {
     String visibilityText = " ";
     switch (program.displayMode()) {
+      case DISABLED:
+        visibilityText = " (Hidden from applicants and TIs) ";
```
rockycodes, line 391: suggestion (non-blocking): I think we have a preference for writing out "Trusted Intermediaries" since people might be unfamiliar with the "TI" acronym.

browser-test/src/admin/admin_publish.test.ts
```diff
@@ -70,6 +70,57 @@ test.describe(
...
+    const draftQuestionText = `${questionText} new version`
+
+    test.beforeEach(async ({page, adminPrograms, adminQuestions}) => {
+      await enableFeatureFlag(page, 'disabled_visibility_condition_enabled')
+      await loginAsAdmin(page)
+
+      // Create a disabled program
+      await adminPrograms.addDisabledProgram(disabledProgram)
+      // Create a new question referenced by the pubic program
+      await adminQuestions.addAddressQuestion({questionName, questionText})
```
rockycodes, line 91: suggestion (non-blocking): use `test.step` so you don't have to add the comments explaining what each step is doing.

browser-test/src/admin/admin_publish.test.ts
```diff
@@ -70,6 +70,57 @@ test.describe(
...
+  {tag: ['@uses-fixtures']},
+  () => {
+    const disabledProgram = 'Disabled test program'
+    const publicProgram = 'Public test program'
+    const questionName = 'admin-publish-test-address-q'
+    const questionText = 'admin-publish-test-address-q'
+    // CreateNewVersion implicitly updates the question text to be suffixed with " new version".
+    const draftQuestionText = `${questionText} new version`
+
+    test.beforeEach(async ({page, adminPrograms, adminQuestions}) => {
```
rockycodes, line 84: question (non-blocking): Are you planning to add more tests here in the future? If not, you can move the logic in the `beforeEach` block into the main test and delete the `beforeEach` block.

server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -387,6 +387,9 @@ private LiTag renderPublishModalProgramItem(
       ProgramDefinition program, ImmutableList<Long> universalQuestionIds) {
     String visibilityText = " ";
     switch (program.displayMode()) {
+      case DISABLED:
```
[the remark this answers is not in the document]
azhangggggg: If the flag is off, `Disabled` display mode won't be provided an option to the civiform admins.
https://github.com/civiform/civiform/blob/347728fb9ea161ab158bcb8b3ba4d54682116cb5/server/app/views/admin/programs/ProgramFormBuilder.java#L191

server/app/views/admin/programs/ProgramIndexView.java
azhangggggg: sounds good, changed it to "Trusted Intermediaries"

browser-test/src/admin/admin_publish.test.ts
azhangggggg: done! thanks for reminding me I will utilize `test.step` more going forward.

browser-test/src/admin/admin_publish.test.ts
azhangggggg: No, okay deleted `beforeEach`.

server/app/views/admin/programs/ProgramIndexView.java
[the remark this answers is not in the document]
rockycodes: I'm thinking of the situation like I outlined in your last PR where an admin turns the flag on, marks some programs as disabled, then we find a bug and they have to turn it off. In that case I think the pop-up would still show "Hidden from applicants and trusted intermediaries" which would be incorrect. But there might be a way to handle this deeper in the logic. Writing this now, it's making me wonder what will actually happen with programs marked "Disabled" if we have to toggle the flag off after turning it on. My guess is they should be tagged with the next most restrictive option (probably "Hide from applicants. Only individuals with the unique program link can access this program") but I think this is another product question. Will you ask Isabel and Sijia how they would want to handle this?

I can see a solution where if the flag is off we change the display mode on any programs marked "Disabled", so I think it's actually fine to merge this as is for now without wrapping it in a feature flag. It's also possible I'm overthinking this 😅 . @dkatzz what do you think?

browser-test/src/admin/admin_publish.test.ts
```diff
@@ -70,6 +70,63 @@ test.describe(
...
+    test('shows programs and questions that will be publised in the modal, including disabled programs', async ({
+      page,
+      adminPrograms,
+      adminQuestions,
+    }) => {
+      const disabledProgram = 'Disabled test program'
+      const publicProgram = 'Public test program'
+      const questionName = 'admin-publish-test-address-q'
+      const questionText = 'admin-publish-test-address-q'
+      // CreateNewVersion implicitly updates the question text to be suffixed with " new version".
```
dkatzz, line 86: **nitpick:** Could we make this the actual function name? `adminPrograms.createNewVersion` - you also may be able to use the TS @link here, but not sure if it works in a comment like this.

browser-test/src/admin/admin_publish.test.ts
```diff
@@ -70,6 +70,63 @@ test.describe(
   },
 )
 
+test.describe(
+  'publishing all programs with disabled programs feature flag on',
+  {tag: ['@uses-fixtures']},
```
dkatzz, line 75: **thought:** @gwendolyngoetz curious if you think its worth adding a tag for this feature or if that's more for larger flags like the northstar work?

server/app/views/admin/programs/ProgramIndexView.java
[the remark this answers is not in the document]
dkatzz: Yeah this is a good question - maybe we could add some sort of durable job to mark anything disabled with the next restrictive option if the flag is off? But that may be overkill and probably worth a larger discussion before doing. In the meantime, I think it can't hurt to add an if statement within this condition to say:

if (flagEnabled) {
visibilityText = " (Hidden from applicants and Trusted Intermediaries) ";
}
break;

That way, it just doesn't show that text if the flag is off.

browser-test/src/admin/admin_publish.test.ts
gwendolyngoetz: There's no harm in adding tags from playwrights perspective. Manage them may become a chore. This is what I'm thinking.

Good candidates for tags (conceptually, not necessarily the names):
- Roles (admin, applicant, etc)
- Environment (local only, real staging environment, etc)
- In development (like northstar)
- Status indicator (slow, flaky, etc)

Poor candidates for tags:
- Super fine grained (admin add question page, settings page)
- Only going to be used by a few things
- Short lived (feature flag 1, feature flag 2)

In this case perhaps we go with `in-development` and use that for longer running feature flag work.

dkatzz: That sounds great, thank you!!

server/app/views/admin/programs/ProgramIndexView.java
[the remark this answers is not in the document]
azhangggggg: Okay, I added a feature flag check to this case :)

dkatzz (approved):

PR #7308, opened 2024-04-24
Add indicator for admin of where you can use markdown in text fields

### LINKED ISSUE #6110: Add indication that markdown can be used in CiviForm Admin workflow
Now that we've added the ability to format some text with markdown in the CiviForm admin workflow, we need to add a visual cue that notifies / remind CiviForm admins that this feature is available to them. 

Right now, admins can use markdown in the following places: 
1. When authoring program descriptions 
2. When authoring custom confirmation messages 
3. When authoring questions 

One suggestion as to how to do this that came from our governance call was to use the markdown component (like Github does in issue creation) below: 
Screenshot 2023-12-04 at 9 19 48 AM

@sijiayam does the suggested approach here work for you? If so, it seems like cases 1 and 2 could be pretty simple. Thoughts on 3? Do we want to add it there as well when I believe it's only bold and italics?

rockycodes:
### Description

This change adds an indicator for where admins can use markdown to apply formatting to text that will show to applicants. It adds new fields onto `FieldWithLabel` to support showing the markdown indicator on fields that support it. Currently, the fields this applies to are program descriptions, custom confirmation screen messages, question text, question help text, and question options (only some markdown applies).

Docs update: https://github.com/civiform/docs/pull/428

### Checklist

1. Navigate to the program details page by adding a new program or editing an existing program. See the markdown indicator on the program description and custom confirmation message fields.
2. Navigate to the question edit page by adding a new question or editing an existing question. See the markdown indicator (with link to documentation) on the question text and question help text fields.
3. Add or edit a question that allows for question options (eg. radio button or checklist).
4. Under "Question settings" click the "Add answer option" button to insert a new option.
5. See the markdown indicator (with link to documentation) on the "Option Text" field.

### Issue(s) this completes

Fixes #6110


server/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -126,6 +126,8 @@ private FormTag buildProgramForm(
             .setFieldName("localizedDisplayDescription")
             .setLabelText("Describe this program for the public")
             .setRequired(true)
+            .setMarkdownSupported(true)
```
gwendolyngoetz, line 129: (question) Should these have the setMarkdownLinkText as well?

server/app/views/components/FieldWithLabel.java
```diff
@@ -87,6 +88,11 @@ public class FieldWithLabel {
   private boolean required = false;
   private boolean ariaRequired = false;
   private boolean focusOnInput = false;
+  private boolean markdownSupported = false;
+  private String markdownText = "";
+  private String markdownLinkText = "";
+  private String markdownLinkHref =
```
gwendolyngoetz, line 94: (suggestion, non-blocking) Since you won't show these if `markdownSupported` is false, what do you think about setting them to a default value. That way you only have to call `setMarkdownSupported(true)` in most cases and only override the markdown text in the one spot that you have "some markdown is supported.

server/app/views/admin/programs/ProgramFormBuilder.java
rockycodes: No, Sijia's designs only wanted to link out the docs from the question creation/edit page because we have some caveats about how to use markdown on those pages.

server/app/views/components/FieldWithLabel.java
rockycodes: Good idea!

rockycodes: I can't set a default for the link text since we rely on the `isBlank` check to render the link, but I set `markdownText` to a default which at least cleans up the program details use case.

gwendolyngoetz (approved):

PR #7316, opened 2024-04-25
Translating all content on the TI dashboard pages

### LINKED ISSUE #7046: Ensure all text on TI dashboard is localized
We need to check that all text on the TI dashboard is translated.  This should be done as the last issue in the TI dashboard epic.

melanie-exygy:
### Description

Up until recently, the TI dashboard wasn't translated.  We have translated a lot of the strings as part of the TI dashboard UX revamp, but some strings still needed to be localized.  This PR ensures that those remaining strings are translated.  In a handful of cases, this PR only submits the string to Transifex for translation (or makes a correction) so strings won't appear as translated yet.

1. Log in as a Trusted Intermediary.
2. Select a language other than English.
3. Check that all strings are translated on the Client List page, the New Client page, the Edit Client page and the Account Settings page.  Note that validation errors on the edit/add client forms will need to be localized in a future PR because the messages needed to be corrected in this PR so that "Error:" doesn't appear twice.

### Issue(s) this completes

Fixes #7046


server/app/views/BaseHtmlView.java
```diff
@@ -241,17 +250,26 @@ private LiTag renderPreviousPageButton(int page, Function<Integer, Call> linkFor
                         .withClasses("usa-icon", "h-4", "w-4")
                         .attr("role", "img")
                         .attr("aria-hidden", "true"),
-                    span("Previous").withClass("usa-pagination__link-text")));
+                    span(optionalMessages.isPresent()
+                            ? optionalMessages
+                                .get()
+                                .at(MessageKey.BUTTON_PREVIOUS_SCREEN.getKeyName())
+                            : "Previous")
```
rockycodes, line 257: suggestion (non-blocking): Is there a way to always pass in messages and have it fall back to the english translation?

rockycodes (commented):
This is great! Hooray for translations for TIs. Only blocking question is if `Displaying all clients` should be translated?

server/app/views/BaseHtmlView.java
melanie-exygy: The issue here is that `ProgramApplicationListView.java` also calls `renderPagination` but, since it's admin-facing, doesn't have access to messages, so I need to provide a way to call that method without messages:  [ProgramApplicationListView.java](https://github.com/civiform/civiform/blob/385110b91a1ed1e59000927fbb7dcfcf7fc8dd06/server/app/views/admin/programs/ProgramApplicationListView.java#L126).  Does that make sense, or is there another way you were thinking of?

rockycodes: Yep that makes sense. I was hoping there was a way to access messages from the other invocation to avoid all the ternaries, but sounds like not at this time! Thanks for explaining, I'll resolve this.

PR #7331, opened 2024-04-29
Add min/max limits for enumerator question answers

### LINKED ISSUE #7136: Allow admin to specify a minimum/maximum answers for enumerator questions
Example: Applicant can choose to have up to 3 trees. Each tree option has multiple fields that would be part of the enumerator.

Currently enumerators are unbounded.

We should add a setting to allow specifying the maximum number of times an enumerator can be filled out. Similarly a minimum setting may be useful.

jeffpw-goog:
### Description

Add a setting to enumerator questions that allows the admin to specify a minimum and maximum limit for the number of answers.

#### General

- [x] Extended the README / documentation, if necessary - https://github.com/civiform/docs/pull/471

#### User visible changes

### Issue(s) this completes

Fixes #7136


gwendolyngoetz (commented):
Maybe I missed it. What happens if I set the min to be larger than than max on the admin question page?

server/test/services/applicant/question/EnumeratorQuestionTest.java
```diff
@@ -122,6 +136,50 @@ public void withDuplicateNames_hasValidationErrors() {
                         MessageKey.ENUMERATOR_VALIDATION_DUPLICATE_ENTITY_NAME))));
   }
```
dkatzz, line 138: **suggestion:** Could we add one more test that it passes validation when they set a min / max and the number of items a person adds is within that?

jeffpw-goog:
> Maybe I missed it. What happens if I set the min to be larger than than max on the admin question page?

Good point! There's currently no handling for this.  Looking at all the other question types that have min/max, only the checkbox question has validation, here: https://github.com/civiform/civiform/blob/91bdeac8aa7e4ed226b3615ac0fbc7e7a0a9455e/server/app/services/question/types/QuestionDefinition.java#L279-L378

I didn't see it because it's in the base class rather than the concrete question class (and I would still have had to be imitating MultiOptionQuestionDefinition rather than one of the others).

Seems like the proper way to do this would be to add a validate() method for concrete question definitions to override if they want to, and then to add validation for all questions that have min/max. Really we should be using composition instead of inheritance, but I don't think I want to go down that rabbit hole.

Do you want to block this PR on adding validation? In that case I'll do a separate PR first to create the overridable validate method, move the question-specific validation logic from QuestionDefinition to the subclasses (EnumeratorQuestion has a bit too), and add validation for the other existing question types that should have it. Otherwise I can file a new issue to handle it for all questions as tech debt.

jeffpw-goog:
> If someone has already answered an enum question with 5 answers and then the question gets updated to say max is 4, what happens? Do they get a validation error when they next open the question?

Yes, they get a validation error when they open the question and then try to save.

server/test/services/applicant/question/EnumeratorQuestionTest.java
jeffpw-goog: Good idea - added!

gwendolyngoetz:
If nothing else on the admin side is currently checking min/max we can file a bug and do them in a separate PR

dkatzz:
> > If nothing else on the admin side is currently checking min/max we can file a bug and do them in a separate PR
> 
> Filed #7919 to follow up on this.

Thank you!! But yeah agreed - I think this would be good to do, but can be done later as a follow up

PR #7345, opened 2024-04-30
[Disabled Programs] create tabs for programs in use and disabled

### LINKED ISSUE #7315: [Disabled Program] Create tabs of disabled programs and programs in use for civiform admin
Adding a new tab for disabled programs in addition to visible programs
Update ProgramService to get disabled programs
Update the view of the program list with tabs, one for Active/In Draft programs, the other for Disabled programs
Add browser tests and integration tests to ensure that the programs are showing under the tab they belong

azhangggggg:
### Description

This PR Implements UI side for adding tabs for programs in use and disabled.

**Before**
Screenshot 2024-05-15 at 2 12 06 PM

**After**
Screenshot 2024-05-15 at 2 11 48 PM

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #7315


server/app/controllers/admin/AdminProgramController.java
```diff
@@ -78,9 +78,26 @@ public Result index(Request request) {
     Optional<CiviFormProfile> profileMaybe = profileUtils.currentUserProfile(request);
     return ok(
         listView.render(
-            programService.getActiveAndDraftProgramsWithoutQuestionLoad(),
+            programService.getInUseActiveAndDraftProgramsWithoutQuestionLoad(),
             questionService.getReadOnlyQuestionServiceSync(),
             request,
+            "In use",
```
rockycodes, line 84: suggestion (blocking): Make "In use" and "Disabled" into an enum that you can reference in both places. That way you don't have to worry about typing in the string incorrectly in one of the places, and it will be easier to add new program statuses in the future if we need to.

rockycodes (changes requested):
This looks good! Just one blocking comment about making the status strings into an enum.

browser-test/src/support/admin_programs.ts
```diff
@@ -230,6 +235,14 @@ export class AdminPrograms {
     return titles.allTextContents()
   }
 
+  async disabledProgramNames() {
+    await this.gotoAdminProgramsPage()
+    await this.page.click('a:has-text("Disabled")')
```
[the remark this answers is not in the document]
azhangggggg: done :)

server/app/controllers/admin/AdminProgramController.java
azhangggggg: Gotcha, added a program tab enum class. Lmk what you thin.

browser-test/src/support/admin_programs.ts
```diff
@@ -736,8 +771,13 @@ export class AdminPrograms {
     programName: string,
     blockDescription = 'screen description',
     questions: QuestionSpec[] = [],
+    isProgramdisabled: boolean = false,
```
rockycodes, line 774: nitpick: I found another one that should be `isProgramDisabled`. This might be it, but please check if there are others.

server/app/controllers/admin/AdminProgramController.java
```diff
@@ -78,9 +79,26 @@ public Result index(Request request) {
     Optional<CiviFormProfile> profileMaybe = profileUtils.currentUserProfile(request);
     return ok(
         listView.render(
-            programService.getActiveAndDraftProgramsWithoutQuestionLoad(),
+            programService.getInUseActiveAndDraftProgramsWithoutQuestionLoad(),
             questionService.getReadOnlyQuestionServiceSync(),
             request,
+            ProgramTab.IN_USE.getProgramTab(),
```
rockycodes, line 85: suggestion (maybe blocking): Instead of calling `getProgramTab` and passing along a string, keep it as the enum until the point where you need to it to be a string (I think that's when you call `.setText(status)`. At that point, you can do `.setText(status.getProgramTab()` instead)`. That way we can have greater control over what's allowed to be passed as a value.

server/app/models/ProgramTab.java
```diff
@@ -0,0 +1,26 @@
...
+  IN_USE("In use"),
+  DISABLED("Disabled");
+
+  private final String tab;
+
+  ProgramTab(String tab) {
+    this.tab = tab;
+  }
+
+  public String getProgramTab() {
```
rockycodes, line 23: suggestion (non-blocking): I see some of the other enum classes calling this `getValue` so you might want to follow that convention, but I don't feel strongly.

rockycodes (changes requested):
One more comment about using the enum and a couple of smaller comments.

server/app/models/ProgramTab.java
avaleske: Could also call it what it is, which is `getTabName()` or `getUILabel()`.

In this case I'd also recommend renaming the `tab` property to be `tabName` or `uiLabel` for consistency.

server/app/controllers/admin/AdminProgramController.java
avaleske: +1 This is sometimes (pejoratively) called being "stringly typed" as opposed to "strongly typed".

server/app/views/admin/programs/ProgramIndexView.java
```diff
@@ -58,24 +62,28 @@ public final class ProgramIndexView extends BaseHtmlView {
...
     this.programCardFactory = checkNotNull(programCardFactory);
+    this.programService = checkNotNull(programService);
     this.settingsManifest = checkNotNull(settingsManifest);
   }
 
   public Content render(
       ActiveAndDraftPrograms programs,
       ReadOnlyQuestionService readOnlyQuestionService,
       Http.Request request,
+      String selectedStatus,
```
avaleske: **suggestion** If we change this to type `ProgramTab` then the name can also just be something like `activeTab`

browser-test/src/support/admin_programs.ts
```diff
@@ -78,6 +78,11 @@ export class AdminPrograms {
     await waitForPageJsLoad(this.page)
   }
 
+  async gotoDisabledProgramIndexPage() {
```
avaleske, line 81: **suggestion, non-blocking**: this is assuming a starting place (the program index page?) is that the behavior the other `goto...` methods on this object have, or should we navigate to the index page before clicking the tab?

browser-test/src/support/admin_programs.ts
```diff
@@ -381,8 +393,18 @@ export class AdminPrograms {
...
   }
 
-  async gotoEditDraftProgramPage(programName: string) {
+  async gotoEditDisabledDraftProgramPage(programName: string) {
+    await this.gotoEditDraftProgramPage(programName, true)
+  }
+
+  async gotoEditDraftProgramPage(
+    programName: string,
+    isDisabled: boolean = false,
```
avaleske, line 402: **suggestion, non-blocking, optional**: it took me a second to figure out what `isDisabled` means here. maybe naming it `programIsDisabled` or whatever is consistent with the file?

browser-test/src/admin/admin_program_list.test.ts
```diff
@@ -44,6 +44,24 @@ test.describe('Program list page.', () => {
...
+    await loginAsAdmin(page)
+
+    const publicProgram = 'List test public program'
+    const disabledProgram = 'List test disabled program'
+    await adminPrograms.addProgram(publicProgram)
+    await adminPrograms.addDisabledProgram(disabledProgram)
+
+    await expectProgramListElements(adminPrograms, [publicProgram])
+    await validateScreenshot(page, 'program-list-in-use-tab')
+    await expectDisabledProgramListElements(adminPrograms, [disabledProgram])
```
avaleske, line 61: **thought, not-blocking*: This test's behavior would be a little clearer if there was a "navigate to disabled tab" step, but I don't know that refactoring `expectProgramListElements()` to enable that structure is worth the effort.

browser-test/src/support/admin_programs.ts
azhangggggg: change applied. also scanned through the files to avoid similar typo.

server/app/controllers/admin/AdminProgramController.java
azhangggggg: that makes sense, it also makes it less error-prone. change applied.

server/app/views/admin/programs/ProgramIndexView.java
azhangggggg: hmm how's `selectedTab`? I think `activeTab` might be confusing cuz it could be interpreted as the tab that contains active programs or smth.

browser-test/src/support/admin_programs.ts
azhangggggg: yup made the change to `isProgramDisabled` to stay consistent with the file

browser-test/src/admin/admin_program_list.test.ts
rockycodes: I had the same series of thoughts. It took me a minute to figure out when we were actually navigating to the disabled programs tab, but it wasn't too hard to track down.

rockycodes (approved):
lgtm!

server/app/views/admin/programs/ProgramIndexView.java
avaleske: oh good point, sgtm

avaleske (approved):
thank you!

PR #7354, opened 2024-04-30
[Disabled Programs] add disabled programs service 

azhangggggg:
### Description

Add service for disabled programs.

`InUse`: versionedByName in ActiveAndDraftPrograms() are map to non-disabled active and draft programs. It will be used in the first tab for civiform admins on program dashboard.
 
`Disabled`: versionedByName in ActiveAndDraftPrograms() are map to disabled active and draft programs. It will be used in the second tab for civiform admins on program dashboard.

`All`: versionedByName in ActiveAndDraftPrograms() are map to all active and draft programs regardless of visibility status. It is need because program admin needs the view to all active and draft programs.

#### General

### Issue(s) this completes

Part of #7315


server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +79,106 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
+   * Returns an ImmutableMap containing all key-value pairs from `allNameToProgram` whose keys are
+   * not present in `nameToProgram`. In other words, this filters out any entries that are shared
+   * between the two maps.
+   *
+   * @param allNameToProgram The complete map of program definitions.
+   * @param nameToProgram The map containing entries to exclude.
+   * @return A new ImmutableMap with the filtered entries.
+   */
+  private ImmutableMap<String, ProgramDefinition> filterMapNameToProgram(
+      ImmutableMap<String, ProgramDefinition> allNameToProgram,
```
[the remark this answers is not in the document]
azhangggggg: Makes sense. Changes made.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -26,14 +28,21 @@ public final class ActiveAndDraftPrograms {
   private final ImmutableMap<String, Pair<Optional<ProgramDefinition>, Optional<ProgramDefinition>>>
       versionedByName;
 
+  enum ActiveAndDraftProgramsType {
+    INUSE,
```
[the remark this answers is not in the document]
azhangggggg: Yeah sure!

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -26,14 +28,21 @@ public final class ActiveAndDraftPrograms {
...
+    DISABLED,
+    ALL
```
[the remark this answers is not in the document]
azhangggggg: That makes sense, I made the change to have ActiveAndDraftPrograms accept a list of types and now only define IN_USE and DISABLED.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +83,109 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
-                    }));
+    if (types.size() == 2
+        && types.contains(ActiveAndDraftProgramsType.DISABLED)
+        && types.contains(ActiveAndDraftProgramsType.IN_USE)) {
+      this.activePrograms = activeNameToProgramAll.values().asList();
+      this.draftPrograms = draftNameToProgramAll.values().asList();
+      this.versionedByName =
+          createVersionedByNameMap(activeNameToProgramAll, draftNameToProgramAll);
+    } else if (types.size() == 1) {
+      ActiveAndDraftProgramsType type = types.get(0);
```
dkatzz, line 120: **suggestion:** We can set the variables that are the same for both cases before the if like:

        this.activePrograms = activeNameToProgram.values().asList();
        this.draftPrograms = draftNameToProgram.values().asList();

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +83,109 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
+        this.versionedByName =
+            createVersionedByNameMap(disabledActiveNameToProgram, disabledDraftNameToProgram);
+      } else if (type.equals(ActiveAndDraftProgramsType.IN_USE)) {
+        this.activePrograms = activeNameToProgram.values().asList();
+        this.draftPrograms = draftNameToProgram.values().asList();
+        this.versionedByName = createVersionedByNameMap(activeNameToProgram, draftNameToProgram);
+      } else {
+        throw new IllegalArgumentException("Unsupported ActiveAndDraftProgramsType: " + type);
+      }
+    } else {
```
dkatzz, line 139: **suggestion:** Before the other if condition could instead consider saying if types.size > 3, so we don't have to check the size at each other step.

azhangggggg: Would you mind clarifying which `if` condition you're referring to?

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: Good call on this one! applied change.

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: Not sure which `if` is referred here but made some changes given my best guess. Let me know if that's not what you meant.

dkatzz: Sorry for the confusion. I meant that instead of having a nested if for types.size < 3 we could remove that if condition all together. If we wanted to make sure only ones with < 3 got to this point, we could create a separate if condition above it for > 3. But with `EnumSet` (mentioned above) we shouldn't need to check the size.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +80,110 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
     return mapNameToProgramWithFilter(repository, service, versionModel, Optional.empty());
   }
 
-  private ActiveAndDraftPrograms(VersionRepository repository, Optional<ProgramService> service) {
+  private ActiveAndDraftPrograms(
+      VersionRepository repository,
+      Optional<ProgramService> service,
+      ImmutableList<ActiveAndDraftProgramsType> types) {
```
dkatzz, line 86: **suggestion:** Sorry for the back and forth, but I'm realizing [EnumSet](https://docs.oracle.com/javase/7/docs/api/java/util/EnumSet.html) would actually be better than a list here. It allows you to do things like EnumSet.allOf(ActiveAndDraftProgramsType) to indicate you want a set of all of them.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +80,110 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
+          ImmutableMap<String, ProgramDefinition> disabledActiveNameToProgram =
+              filterMapNameToProgram(activeNameToProgramAll, activeNameToProgram);
+          // Disabled draft programs.
+          ImmutableMap<String, ProgramDefinition> disabledDraftNameToProgram =
+              filterMapNameToProgram(draftNameToProgramAll, draftNameToProgram);
+          this.versionedByName =
+              createVersionedByNameMap(disabledActiveNameToProgram, disabledDraftNameToProgram);
+        } else if (type.equals(ActiveAndDraftProgramsType.IN_USE)) {
+          this.versionedByName = createVersionedByNameMap(activeNameToProgram, draftNameToProgram);
+        } else {
```
dkatzz, line 131: **suggestion:** We should be able to remove these illegal argument exceptions because it won't be possible to add an invalid type into the EnumSet

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +80,110 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
-                          Optional.ofNullable(draftNameToProgram.get(programName)));
-                    }));
+    if (types.size() < 3) {
+      if (types.contains(ActiveAndDraftProgramsType.DISABLED)
+          && types.contains(ActiveAndDraftProgramsType.IN_USE)) {
+        this.activePrograms = activeNameToProgramAll.values().asList();
+        this.draftPrograms = draftNameToProgramAll.values().asList();
+        this.versionedByName =
+            createVersionedByNameMap(activeNameToProgramAll, draftNameToProgramAll);
+      } else if (types.size() == 1) {
```
dkatzz, line 116: **suggestion:** This could just be an else instead of an else if

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -26,14 +28,22 @@ public final class ActiveAndDraftPrograms {
   private final ImmutableMap<String, Pair<Optional<ProgramDefinition>, Optional<ProgramDefinition>>>
       versionedByName;
 
+  enum ActiveAndDraftProgramsType {
+    IN_USE,
+    DISABLED
+  }
+
+  private static ImmutableList<ActiveAndDraftProgramsType> allProgramTypes =
```
dkatzz, line 36: **suggestion:** This can be noted as final as well.

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: No worries. I love the idea of using `EnumSet`, thank you so much for the suggestion! Changes applied.

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +81,104 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
-        Sets.union(activeNameToProgram.keySet(), draftNameToProgram.keySet()).stream()
-            .collect(
-                ImmutableMap.toImmutableMap(
-                    Function.identity(),
-                    programName -> {
-                      return Pair.create(
-                          Optional.ofNullable(activeNameToProgram.get(programName)),
-                          Optional.ofNullable(draftNameToProgram.get(programName)));
-                    }));
+    this.activePrograms =
```
dkatzz, line 110: **suggestion:** I think this structure may be cleaner since we already have to go into an if condition, but let me know what you think:

```
if (types.containsAll(allProgramTypes)) {
this.activePrograms = ..
this.draftPrograms = ..
} else {
this.activePrograms = ..
this.draftPrograms = ..

  if (types.contains(ActiveAndDraftProgramsType.DISABLED)) {
  ...
  }

  if (types.contains(...IN_USE
  ...
}
```

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +81,104 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
+            ? activeNameToProgramAll.values().asList()
+            : activeNameToProgram.values().asList();
+    this.draftPrograms =
+        types.equals(allProgramTypes)
+            ? draftNameToProgramAll.values().asList()
+            : draftNameToProgram.values().asList();
+    if (types.equals(allProgramTypes)) {
+      this.versionedByName =
+          createVersionedByNameMap(activeNameToProgramAll, draftNameToProgramAll);
+    } else if (types.equals(EnumSet.of(ActiveAndDraftProgramsType.DISABLED))) {
```
dkatzz, line 121: **suggestion:** Can we change this to `types.contains(ActiveAndDraftProgramsType.DISABLED)` here (and similar below)

server/app/services/program/ActiveAndDraftPrograms.java
```diff
@@ -70,30 +81,104 @@ private ImmutableMap<String, ProgramDefinition> mapNameToProgram(
...
+          filterMapNameToProgram(activeNameToProgramAll, activeNameToProgram);
+      // Disabled draft programs.
+      ImmutableMap<String, ProgramDefinition> disabledDraftNameToProgram =
+          filterMapNameToProgram(draftNameToProgramAll, draftNameToProgram);
+      this.versionedByName =
+          createVersionedByNameMap(disabledActiveNameToProgram, disabledDraftNameToProgram);
+    } else if (types.equals(EnumSet.of(ActiveAndDraftProgramsType.IN_USE))) {
+      this.versionedByName = createVersionedByNameMap(activeNameToProgram, draftNameToProgram);
+    } else {
+      throw new IllegalArgumentException("Unsupported ActiveAndDraftProgramsType: " + types);
```
dkatzz, line 133: **suggestion:** I don't think we should need this with the new implementation

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: One problem with this structure is that with `if (types.contains(...IN_USE`, it will cause error `variable versionedByName might already have been assigned`
So I changed it to `else if (types.contains(...IN_USE`. But the rest change you suggested has been applied.

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: Ok, changes applied.

server/app/services/program/ActiveAndDraftPrograms.java
azhangggggg: We do need this even with the new implementation, otherwise it errors `variable versionedByName might not have been initialized`. Feel free to run the branch locally.

dkatzz (approved):

server/app/services/program/ActiveAndDraftPrograms.java
dkatzz: Sounds good, thanks!!

PR #7389, opened 2024-05-02
Make Scalar more readable

jeffpw-goog:
### Description

Make Scalar more readable:
* Inline _SCALARS variables because they add an unnecessary layer of indirection
* Keep METADATA_SCALARS in order to define a new constant for METADATA_SCALAR_KEYS to avoid the unsafe way it was being lazily initialized (https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html)

#### General

### DIFF
--- a/server/app/services/applicant/question/Scalar.java
+++ b/server/app/services/applicant/question/Scalar.java
@@ -61,6 +61,14 @@ public enum Scalar {
   UPDATED_AT("updated at", ScalarType.LONG),
   PROGRAM_UPDATED_IN("program updated in", ScalarType.LONG);
 
+  private static final ImmutableSet<Scalar> METADATA_SCALARS =
+      ImmutableSet.of(UPDATED_AT, PROGRAM_UPDATED_IN);
+  private static final ImmutableSet<String> METADATA_SCALAR_KEYS =
+      METADATA_SCALARS.stream()
+          .map(Scalar::name)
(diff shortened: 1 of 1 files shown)


server/app/services/applicant/question/Scalar.java
```diff
@@ -132,59 +88,60 @@ public static ImmutableSet<Scalar> getScalars(QuestionType questionType)
...
   public static ImmutableSet<String> getMetadataScalarKeys() {
-    if (metadataScalarKeys == null) {
-      metadataScalarKeys =
-          METADATA_SCALARS.stream()
-              .map(Scalar::name)
-              .map(String::toLowerCase)
-              .collect(ImmutableSet.toImmutableSet());
-    }
-    return metadataScalarKeys;
+    return getMetadataScalars().stream()
```
swatkat1, line 142: I understand the reasoning for others, but for metadataScalarKeys, isn't it better to store it at object level instead of calculating it over and over?

jeffpw-goog: I think it's an unnecessary optimization because CPU time is ~free and this is not called at a high frequency. But I can fix the original problem by defining the variable statically, so I'll do that instead.

swatkat1 (approved):
Thanks!

dkatzz (approved):

PR #7406, opened 2024-05-03
Simplify help text typescript

jeffpw-goog:
### Description

Simplify help text logic:
* querySelectorAll never returns null so don't need to check for that
* Use NodeList.forEach instead of converting to Array
* Add a comment that each value input has its own help text, to explain why we need querySelectorAll in the first place
* Rename configurePredicateValueInput to configurePredicateValueInputs to also remind that there can be more than one value input
* Remove the comment about the help text maybe not being present, because it no longer matters for the logic
* Inline the hidden condition; it was named backwards anyway

#### General

### DIFF
--- a/server/app/assets/javascripts/admin_predicate_configuration.ts
+++ b/server/app/assets/javascripts/admin_predicate_configuration.ts
@@ -153,7 +153,7 @@ class AdminPredicateConfiguration {
       selectedScalarType,
       selectedScalarValue,
     )
-    this.configurePredicateValueInput(
+    this.configurePredicateValueInputs(
       selectedScalarType,
       selectedScalarValue,
       questionId,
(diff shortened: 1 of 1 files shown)


server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -206,44 +206,28 @@ class AdminPredicateConfiguration {
...
     const csvHelpTexts = document.querySelectorAll(
       `#predicate-config-value-row-container [data-question-id="${questionId}"] .cf-predicate-value-comma-help-text`,
     )
-
-    // The help text div is present for inputs that allow specifying
-    // multiple values in a single text input. It won't be present
-    // for other input types e.g. multiselect inputs.
-    if (csvHelpTexts != null) {
-      const shouldShowCommaSeperatedHelpText =
+    csvHelpTexts.forEach((div: Element) =>
```
melanie-exygy, line 213: praise:  This is so much clearer and more succinct.  Thanks for cleaning up!

melanie-exygy (approved):

server/app/assets/javascripts/admin_predicate_configuration.ts
jeffpw-goog: Thank you! :)

PR #7427, opened 2024-05-06
Reduce string case shenanigans

jeffpw-goog:
### Description

Reduce string case shenanigans:
* The operator is always upper-case so it doesn't need to be upper-cased
* The scalar type is lower-cased to stuff into the element data, but otherwise it's clearer to use the upper case version, so return that from the backend and then lower-case it only where necessary in the frontend

#### General

### DIFF
--- a/server/app/assets/javascripts/admin_predicate_configuration.ts
+++ b/server/app/assets/javascripts/admin_predicate_configuration.ts
@@ -214,8 +214,7 @@ class AdminPredicateConfiguration {
     csvHelpTexts.forEach((div: Element) =>
       div.classList.toggle(
         'hidden',
-        selectedOperatorValue.toUpperCase() !== 'IN' &&
-          selectedOperatorValue.toUpperCase() !== 'NOT_IN',
+        selectedOperatorValue !== 'IN' && selectedOperatorValue !== 'NOT_IN',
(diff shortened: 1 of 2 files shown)


server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -499,8 +492,8 @@ class AdminPredicateConfiguration {
       // of EQUAL_TO with ANY_OF and NOT_EQUAL_TO with NONE_OF, we made a technical choice to
       // exclude these operators from single-select predicates to simplify the code on both
       // the form processing side and on the admin user side.
-      !(selectedScalarType in operatorOption.dataset) ||
-      (selectedScalarValue.toUpperCase() === 'SELECTION' &&
+      !(selectedScalarType.toLowerCase() in operatorOption.dataset) ||
```
gwendolyngoetz, line 495: (question) Why is this one lowercase when  everything else is all caps?

gwendolyngoetz: Nevermind just say your comment.

gwendolyngoetz (approved):

gwendolyngoetz (approved):

PR #7443, opened 2024-05-07
Refactor ProgramPredicateConfigureView.formatPredicateValue

jeffpw-goog:
### Description

Formatting predicate values based on the supported right hand types for the given operator makes two assumptions:
1. That each operator supports at most one right hand type for a given scalar type. This isn't true for AGE_BETWEEN, which supports both LONG and DOUBLE for the scalar type DATE. So the existing logic `operator.getRightHandTypes().contains(OperatorRightHandType.LONG)` for scalar type DATE obfuscated the fact that the value might be LONG or DOUBLE and only happened to work correctly by chance.
2. That an operator that supports a given right hand type supports it for the scalar where it's being checked. This assumption is probably okay today, but it opens a vector for bugs. For example, Operator LESS_THAN could support LONG for LONG scalar types and DATE for DATE scalars (today they are separate operators, but it could be implemented this way). The AGE_YOUNGER_THAN operator also supports the LONG right hand type for DATE scalars. If for DATE scalar this logic checks first that the LESS_THAN operator supports LONG before DATE, it will incorrectly format a DATE right hand value as a LONG.

If formatPredicateValue instead formats based on the predicate type, then it's explicit what kind of value it's formatting, it removes the assumptions above, and the code is easier to read.

#### General

### DIFF
--- a/server/app/views/admin/programs/ProgramPredicateConfigureView.java
+++ b/server/app/views/admin/programs/ProgramPredicateConfigureView.java
@@ -735,8 +735,7 @@ public final class ProgramPredicateConfigureView extends ProgramBaseView {
                 .setValue(
                     maybeLeafOperationNode.map(
                         leafNode ->
-                            formatPredicateValue(
-                                leafNode.scalar(), leafNode.operator(), leafNode.comparedValue())))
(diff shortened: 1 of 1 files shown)


dkatzz (approved):
Thank you for cleaning this up!!

Issue #6109, opened 2023-12-01
Ability to set eligibility based on specific BETWEEN range (all date and number questions)


elisekalstad:
**Is your feature request related to a problem? Please describe.**
As a Program Admin, I want to be able to only allow people to apply to a program if their birthday falls within a specific date range. For example, the Seattle Preschool Program only accepts applications for kids with birthdays between 9/1/2018 and 8/31/2020. I currently cannot set an eligibility condition between two date ranges.

**Describe the solution you'd like**
Ability to set, "date" "is between" "mm/dd/yyyy - mm/dd/yyyy" (similar to the existing function where by you can set "age" "is between" "3-4"

**Additional context**
![image](https://github.com/civiform/civiform/assets/86082659/785b66b2-7bc0-4ed0-8791-75259f2aaa6c)

elisekalstad:
Hi @isabelcshaw and @nb1701 -- There's a request for a small improvement to the predicate logic to be able to set eligible date ranges. Is there someone on the greater CiviForm team who has experience in the predicate code who could take this one on?

dkatzz:
https://github.com/civiform/civiform/pull/4428/files should be helpful for this change (the original PR to create the between logic).

msprenke:
Needing a "between" calculation isnt specific to just dates.  This applies to numbers as well.  Example, Number X is Between 60000 and 69999

I'm updating the title to reflect this need for a between eligibility calculation for all numbers AND date fields.

PR #7450, opened 2024-05-08
Refactor validateSubmit in admin_predicate_configuration.ts

jeffpw-goog:
### Description

* Remove unnecessary `Array.from`s
* Change ? to ! because #predicate-config-value-row-container should always exist and is an error if it doesn't, rather than failing silently
* Condition the missing value logic on whether the input is a checkbox type, rather than whether there is more than one input
* Remove unnecessary early returns that are barely optimizing anything and add complexity
* Re-work the logic to show the error message to flow more naturally

In addition to being cleaner this enables #6109 where we will have more than one non-checkbox input box.

#### General

### DIFF
--- a/server/app/assets/javascripts/admin_predicate_configuration.ts
+++ b/server/app/assets/javascripts/admin_predicate_configuration.ts
@@ -72,59 +72,46 @@ class AdminPredicateConfiguration {
     let hasValueMissing = false
 
     // Check if any scalars are missing a value.
-    Array.from(
-      document.querySelectorAll<HTMLSelectElement>('.cf-scalar-select select'),
-    ).forEach((el: HTMLSelectElement) => {
-      if (el.options[el.options.selectedIndex].value == '') {
-        hasSelectionMissing = true
-      }
-    })
-
-    // Check if any operators are missing a value.
-    Array.from(
-      document.querySelectorAll<HTMLSelectElement>(
-        '.cf-operator-select select',
-      ),
-    ).forEach((el: HTMLSelectElement) => {
-      if (el.options[el.options.selectedIndex].value == '') {
-        hasSelectionMissing = true
-      }
-    })
-
-    // Check if any inputs are missing a value.
     document
-      .querySelector('#predicate-config-value-row-container')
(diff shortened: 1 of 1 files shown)


server/app/assets/javascripts/admin_predicate_configuration.ts
```diff
@@ -72,72 +72,62 @@ class AdminPredicateConfiguration {
...
+      .querySelectorAll<HTMLSelectElement>('.cf-operator-select select')
+      .forEach((el: HTMLSelectElement) => {
+        if (el.options[el.options.selectedIndex].value == '') {
+          hasSelectionMissing = true
+        }
+      })
 
     // Check if any inputs are missing a value.
     document
-      .querySelector('#predicate-config-value-row-container')
```
jcma-google, line 96: nit: #predicate-config-value-row-container name isn't very clear, I'm not sure if this is a convention or not but if it isn't is it possible to rename to something clearer? Feel free to ignore if my request is naive and the cost outweighs the benefits.

jcma-google (commented):
Are tests already present for this use case? Should we add any?

server/app/assets/javascripts/admin_predicate_configuration.ts
jeffpw-goog: Thanks for the suggestion! Based on the naming throughout the page this name actually does make decent sense to me - this page is for predicate config, and value rows are the rows of value input boxes (vs the scalar, operator, etc. inputs at the top), and this contains all of those. The term value row is used widely throughout the page and associated utils (like for testing). I'm happy to consider if you have a specific suggestion, but I don't have a better one myself.

jeffpw-goog:
> Are tests already present for this use case? Should we add any?

Yes there are tests in admin_predicates.test.ts that cover each of the error cases: https://github.com/civiform/civiform/blob/92045e2e92c047b3126f91c8ce1dcb4cda859e53/browser-test/src/admin/admin_predicates.test.ts#L242-L284

jcma-google (approved):

PR #7453, opened 2024-05-08
Fix links to uploaded files to use correct routes

### LINKED ISSUE #7205: Applicant PDF download after applying to a program with file upload question includes a link to download file that doesn't work
**Describe the bug**
When APPLICATION_EXPORTABLE flag is turned on (on the settings page), applicants are able to download a PDF of their application after submitting. If the program includes a file upload question, there will be a link to download the file in the PDF, but this doesn't work. We should change the behavior.

Note that the full link (https://staging-aws.civiform.dev/applicants/726268/files/applicant-726268%252Fprogram-108036%252Fblock-1%252Fcharlotte+%25281%2529+%25282%2529.jpg) will work if you're already logged in, but it won't otherwise. 

We'll plan to go with the option below, which involves downloading the uploaded files when exporting the application and changing the name in the export to be: `(PDF downloaded separately)` from `UPLOADED (click to download)` and removing the link property of the text

**To Reproduce**

1. Create an application with a file upload question
2. As an admin turn on APPLICATION_EXPORTABLE
3. Login as an applicant and fill out the application and export the PDF at the end
4. Click the download link in the PDF
5. Get navigated to home page

**Expected behavior**
Download the uploaded files when exporting the application and changing the name in the export to be: `(PDF downloaded separately)` from `UPLOADED (click to download)`. The text should no longer be clickable as well.

**Screenshots**
[name-unavailable (79548)-2024-04-11T08_37_10.198442.pdf](https://github.com/civiform/civiform/files/14948474/name-unavailable.79548.-2024-04-11T08_37_10.198442.pdf)

**Additional context**
See https://github.com/civiform/civiform/issues/7149 for further context

devin-civi:
### Description

This makes two main changes:

1) Change any links we expose to download files from the deprecated adminShow to acledAdminShow. This is no-op change since adminShow just redirects to acledAdminShow.
2) For PDF exports, if an applicant is exporting, use the applicant "show" route, rather than the admin show route. I think this was the primary cause of issue #7205.

Note if the applicant isn't logged in, the link will still fail. So there are probably more improvements we could do, but they're larger in scope. (Either prompting users for login for acl'd routes, or downloading all uploaded files along with the application, which is difficult because they'd need to be in a zip or some other package, which may be confusing for applicants.)

## Release notes

Fixed issue where applicant's could not download uploaded files from the PDF export of their application.

#### General

### Issue(s) this completes

Fixes #7205

### DIFF
--- a/server/app/controllers/applicant/UpsellController.java
+++ b/server/app/controllers/applicant/UpsellController.java
@@ -213,7 +213,8 @@ public final class UpsellController extends CiviFormController {
                   pdfExporterService.generateApplicationPdf(
                       applicationMaybe.join().get(),
                       /* showEligibilityText= */ false,
-                      /* includeHiddenBlocks= */ false);
+                      /* includeHiddenBlocks= */ false,
+                      /* isAdmin= */ false);
 
               return ok(pdf.getByteArray())
                   .as("application/pdf")
--- a/server/app/controllers/applicant/UpsellController.java
+++ b/server/app/controllers/applicant/UpsellController.java
@@ -211,10 +211,7 @@ public final class UpsellController extends CiviFormController {
             check -> {
               PdfExporter.InMemoryPdf pdf =
                   pdfExporterService.generateApplicationPdf(
-                      applicationMaybe.join().get(),
-                      /* showEligibilityText= */ false,
-                      /* includeHiddenBlocks= */ false,
-                      /* isAdmin= */ false);
+                      applicationMaybe.join().get(), /* isAdmin= */ false);
 
               return ok(pdf.getByteArray())
                   .as("application/pdf")
--- a/server/app/services/applications/PdfExporterService.java
+++ b/server/app/services/applications/PdfExporterService.java
@@ -28,15 +28,10 @@ public final class PdfExporterService {
    * to review applications.
    */
   public PdfExporter.InMemoryPdf generateApplicationPdf(
-      ApplicationModel application,
-      boolean showEligibilityText,
-      boolean includeHiddenBlocks,
-      boolean isAdmin) {
+      ApplicationModel application, boolean isAdmin) {
     PdfExporter.InMemoryPdf pdf;
     try {
-      pdf =
-          pdfExporter.exportApplication(
-              application, showEligibilityText, includeHiddenBlocks, isAdmin);
+      pdf = pdfExporter.exportApplication(application, isAdmin);
     } catch (DocumentException | IOException e) {
       throw new RuntimeException(e);
     }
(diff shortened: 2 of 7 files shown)


devin-civi:
* **#7453** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/services/applications/PdfExporterService.java
```diff
@@ -28,10 +28,15 @@ public final class PdfExporterService {
    * to review applications.
    */
   public PdfExporter.InMemoryPdf generateApplicationPdf(
-      ApplicationModel application, boolean showEligibilityText, boolean includeHiddenBlocks) {
+      ApplicationModel application,
+      boolean showEligibilityText,
```
swatkat1, line 32: Devin, showEligibilityText --> is also set for only admins. I think we can rename this variable to isAdmin and make it as one variable.

server/app/controllers/applicant/UpsellController.java
```diff
@@ -213,7 +213,8 @@ public CompletionStage<Result> download(
                   pdfExporterService.generateApplicationPdf(
                       applicationMaybe.join().get(),
                       /* showEligibilityText= */ false,
-                      /* includeHiddenBlocks= */ false);
+                      /* includeHiddenBlocks= */ false,
```
swatkat1, line 216: Does include includeHiddenBlocks true only for admins? If yes, let us fold them too into into isAdmin

server/test/services/applications/PdfExporterServiceTest.java
```diff
@@ -53,10 +53,7 @@ public void generateApplicationPdf() throws IOException {
       assertThat(AnnotationAction.get(PdfName.S)).isEqualTo(PdfName.URI);
       PdfString link = AnnotationAction.getAsString(PdfName.URI);
       assertThat(link.toString())
-          .isEqualTo(
-              "http://localhost:9000/admin/programs/"
-                  + applicationOne.getProgram().id
-                  + "/files/my-file-key");
+          .isEqualTo("http://localhost:9000/admin/applicant-files/my-file-key");
```
swatkat1, line 56: So for acled file paths - no program id in the url. Am I going correctly?

swatkat1 (approved):
This is great!

server/test/services/applications/PdfExporterServiceTest.java
devin-civi: Yep. (You can see FileController for the methods that end up being called. The old adminShow takes a program ID, but doesn't use it and just calls into the new one. So all this change does is call directly into the new one, and thus doesn't need the program ID).

PR #7456, opened 2024-05-09
Refactor addValueRows in admin_predicates.ts

jeffpw-goog:
### Description

It makes more sense for addValueRows to take a number to add, rather than a PredicateSpec. This way other users can add a specific number of rows if they don't have a PredicateSpec.

#### General

### DIFF
--- a/browser-test/src/support/admin_predicates.ts
+++ b/browser-test/src/support/admin_predicates.ts
@@ -18,13 +18,9 @@ export class AdminPredicates {
     this.page = page
   }
 
-  async addValueRows(predicateSpec: PredicateSpec) {
-    const values = predicateSpec.values
-
-    if (values && values.length > 1) {
-      for (let i = 1; i < values.length; i++) {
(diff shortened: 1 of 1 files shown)


browser-test/src/support/admin_predicates.ts
```diff
@@ -48,7 +44,8 @@ export class AdminPredicates {
     }
 
     await this.clickAddConditionButton()
-    await this.addValueRows(predicateSpecs[0])
+    const totalRowsNeeded = predicateSpecs[0]?.values?.length ?? 0
+    await this.addValueRows(Math.max(totalRowsNeeded - 1, 0))
```
gwendolyngoetz, line 48: (question) Is this adding one less rows because there's already one on the screen by default?

jeffpw-goog: Yup - the existing code was doing that already, it was just less obvious because the for loop started at 1 instead of 0

gwendolyngoetz (approved):

Issue #6427, opened 2024-01-22
Upgrade to Java 17


gwendolyngoetz:
## Notes 

> [!WARNING]
> Version 2.9.x is [likely to be the final release supporting Java 11](https://www.playframework.com/documentation/2.9.x/Migration29#Minimum-required-Java-and-sbt-version).

Migration guide, but most of this will be completed in the upgrade from Play Framework v2.8 to v2.9: https://www.playframework.com/documentation/2.9.x/Migration29

**Java 17 Branch**
`java_17` (https://github.com/civiform/civiform/pull/7460)

### For Java 17
**If you work on Java 17 make sure to run `bin/build-dev` or your container will continue to run Java 11**
- [x] Basic compilation on Java 17 with Play 2.9
- [x] Play Enhancer (for ebeans) is dead :skull:. [We need to add getters/setters manually](https://github.com/playframework/playframework/issues/11991#issuecomment-1740403613). Otherwise you get `BeanNotEnhancedException`s. This appears to have gone away when trying on Java 17. Before when I tried with Java 11 I got the errors.
- [x] Fix junit tests failing due to too many postgres connection attempts. This actually looks to be related to the BeanNotEnhancedException.
- [ ] Fix errorprone (see https://issuetracker.google.com/issues/242307248)
  -  Errorprone is the cause of `Fix browser test server running in `--ci` mode` failing
- [ ] Upgrade to pac4j v6 #6348

Upgrading Play Ebean to v7.0.0 will be done as a separate step for Java 11. https://github.com/civiform/civiform/issues/6426.

gwendolyngoetz:
Related https://github.com/civiform/civiform/issues/4166

gwendolyngoetz:
Related #7461

PR #7460, opened 2024-05-10
Java 17

gwendolyngoetz:
Branch configured for Java 17 instead of Java 11

Details at #6427

> [!WARNING]
> Make sure to run `bin/build-dev` or your container will continue to run Java 11

- [x] Errorprone working

### Running

Use `bin/run-dev` to get the error. I have not filed a bug as the error requests, I am not certain it's a bug or misconfiguration.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,7 +18,7 @@
     "ghcr.io/devcontainers/features/sshd:1": {},
     "ghcr.io/devcontainers/features/github-cli:1": {},
     "ghcr.io/devcontainers/features/java:1.5.0": {
-      "version": "11",
+      "version": "17",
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
```diff
@@ -18,7 +18,7 @@
     "ghcr.io/devcontainers/features/sshd:1": {},
     "ghcr.io/devcontainers/features/github-cli:1": {},
     "ghcr.io/devcontainers/features/java:1.5.0": {
-      "version": "11",
+      "version": "17",
```
gwendolyngoetz, line 21: @dkatzz Is there anything else involved in making the devcontainers update to Java 17? Aside from the config, I assume people will need to destroy and create new instances.

gwendolyngoetz: That's good to know. Losing bash history isn't the end of the world.

dkatzz (approved):
Yay!! Thank you for doing this!

PR #7462, opened 2024-05-10
Upgrade to Play Ebean v7

gwendolyngoetz:
Initial updates for [Play Ebean v7](https://github.com/playframework/play-ebean). Previously I had been getting bean not enhanced errors, but those went away when I switched to using Java 17. Maybe from Lombok?

TODO:

- [x] Fix MetricsControllerTest - The tests fail intermittently because the counters sometimes already exist. Moved the counters to a `@Singleton` class.


gwendolyngoetz:
Closing this PR. It's was much easier to just do a new branch and not mess with this cherry-picked one.

Issue #7466, opened 2024-05-11
Roadmap for Major Dependency Updates


gwendolyngoetz:
### Status

Compiles | Tests Pass | Ready to Merge | Merged | Item | Blockers
--|----|----|----|-------------------------------|--------------------
✅ | ✅ | ✅ | ✅ |  Play 2.9.x                                  | 
✅ | ✅ | ✅ | ✅ | [Java 17 with Errorprone](https://github.com/civiform/civiform/pull/7460) #6427  |
✅ | ✅ | ✅ | ✅ |  [Play Pac4j 12.x-Play2.9/pac4j-core 6.x](https://github.com/civiform/civiform/pull/8239)             | Java 17
✅ | ✅ | ✅ | ✅ |  [Play Ebean 7.x](https://github.com/civiform/civiform/pull/7740) #6426                           | Java 17
✅ | ✅ | ✅ | ✅ |  [Play 3.x](https://github.com/civiform/civiform/pull/7467)                                       | 
✅ | ✅ | ✅ | ✅ |  Play Prometheus Filters 1.x        | Play 3.x
✅ | ✅ | ✅ | ✅ |  Play Ebean 8.x                            | Play 3.x
✅ | ✅ | ✅ | ✅ |  Play Pac4j 12.x-Play3.0              | Java 17, Play 3.x
📆 | 📆 |📆 | 📆  | Java 21                                         | Play Pac4j(?)

### Links 
- [Play Ebean](https://github.com/playframework/play-ebean)
- [Play Pac4j](https://github.com/pac4j/play-pac4j)
- [Play Prometheus Filters](https://github.com/Jyllands-Posten/play-prometheus-filters)
- [Play Framework 3](https://www.playframework.com/documentation/3.0.x/Releases)

### Dependency Order

```mermaid
gantt
    title Not a Gantt Chart
    dateFormat YYYY-MM-DD
    tickInterval 1day
    section Play 2.9
        Play 2.9                      :   play29,        2024-10-29, 1d
        Java 17                      :   java17,        after play29, 1d
        Play Pac4j 12.x_2.9    :   pac4j12_29,   after java17, 1d
        Play Ebean 7.x           :   ebean7,      after java17, 1d
    section Play 3.0
        Play 3.0                     :  play30, after ebean7, 1d
        Play Pac4j 12.x_3.0    :   pac4j12_30,   after play30, 1d
        Play Ebean 8.x           :   ebean8,      after play30, 1d
[...]

PR #7467, opened 2024-05-12
Play 3 Upgrade

gwendolyngoetz:
## Docs

[Play 2.9 -> 3.x Migration Guide](https://www.playframework.com/documentation/3.0.x/Migration30)
[Play Scala 3 Migration](https://www.playframework.com/documentation/3.0.x/Scala3Migration)

## Akka -> Pekko

The primary change in Play 3 is that switch from Akka to the Apache fork named Pekko. 

There are some environment variables that have `Akka` in their name. It would be nice to rename them, but then we're having to verify all deployment config files are changed. There is no real downside in keeping the environment variables named as they are.

## Prometheus change 

Prior version of the Prometheus library only had `applicant_id_in_profile_total`. With the new version there is an additionally generated option named `applicant_id_in_profile_created`. The tests were written assuming only the `applicant_id_in_profile_total` was there.

mfs.name == "applicant_id_in_profile"
and includes "applicant_id_in_profile_total" and "applicant_id_in_profile_created"

image

Related to: #7466

Fixes: #7032, #7394

### DIFF
--- a/server/build.sbt
+++ b/server/build.sbt
@@ -13,7 +13,8 @@ lazy val root = (project in file("."))
   .settings(
     name := """civiform-server""",
     version := "0.0.1",
-    scalaVersion := "2.13.15",
+    crossScalaVersions := Seq("2.13.15", "3.3.3"),
+    scalaVersion := crossScalaVersions.value.head,
     maintainer := "uat-public-contact@google.com",
     libraryDependencies ++= Seq(
       // Provides in-memory caching via the Play cache interface.
@@ -54,7 +55,7 @@ lazy val root = (project in file("."))
       "com.h2database" % "h2" % "2.3.232" % Test,
 
       // Metrics collection and export for Prometheus
-      "io.github.jyllands-posten" %% "play-prometheus-filters" % "0.6.1",
+      "io.github.jyllands-posten" %% "play-prometheus-filters" % "1.0.2",
 
       // Parameterized testing
       "pl.pragmatists" % "JUnitParams" % "1.1.1" % Test,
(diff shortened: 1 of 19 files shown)


server/test/controllers/applicant/ApplicantRoutesTest.java
```diff
@@ -40,6 +40,10 @@ private Counts getApplicantIdInProfileCounts() {
     for (MetricFamilySamples mfs : Collections.list(registry.metricFamilySamples())) {
       if (mfs.name.equals("applicant_id_in_profile")) {
         for (MetricFamilySamples.Sample sample : mfs.samples) {
+          if (!sample.name.equals("applicant_id_in_profile_total")) {
```
gwendolyngoetz, line 43: Prior version of the Prometheus library only had `applicant_id_in_profile_total`. With the new version there is an additionally generated option named `applicant_id_in_profile_created`. The tests were written assuming only the `applicant_id_in_profile_total` was there.

server/build.sbt
```diff
@@ -13,7 +13,8 @@ lazy val root = (project in file("."))
   .settings(
     name := """civiform-server""",
     version := "0.0.1",
-    scalaVersion := "2.13.15",
+    crossScalaVersions := Seq("2.13.15", "3.3.3"),
+    scalaVersion := crossScalaVersions.value.head,
```
gwendolyngoetz, line 17: This is how they have the [Play 3 samples](https://github.com/playframework/play-samples/blob/aca981218ee0e8b9299b8fcf3436087a00ba9226/play-java-hello-world-tutorial/build.sbt#L9-L11) configured.

gwendolyngoetz: Nah, it's normal. Scala can cross compile so there is compatibility between libraries as versions shift. For example Play is compatible with Scala 3, but the Prometheus library does not yet have a Scala 3 binary.

gwendolyngoetz: *rewrote for better clarity*

Nah, it's normal. Scala can cross compile so there are compatible libraries build as versions of Scala release. The default scala version is set to `crossScalaVersions.value.head` which is 2.13.15. This makes it so that our app will stick with the 2.x version for now, but make it easy within sbt to manually trigger a 3.3.3 compilation for later testing instead of having to come back and make additional changes. Play itself is compatible with Scala 3, but the Prometheus library does not yet have a Scala 3 binary.

PR #7478, opened 2024-05-13
Create Civiform header

greg-hyde:
### Description

Implement a basic header and add it to all existing Thymeleaf pages. It includes the login/logout button and language selector, and uses the built in mobile header

#### General

#### User visible changes

#### New Features


greg-hyde:
* **#7478** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rockycodes (commented):
I didn't do a super thorough review as the fyi reviewer, but overall lgtm!

server/app/assets/stylesheets/uswds/_uswds-theme-custom-styles.scss
```diff
@@ -62,3 +62,38 @@ See the radio-option below as an example.
...
+.usa-header--basic .usa-button {
+  @include u-line-height('sans', 2); 
+}
+
+/**
+ * Default form styles add a bottom margin to forms, but we don't want that for the language
+ * selector as it would mess with header styles.
+ */
+
+.usa-header--basic form {
```
avaleske, line 95: **Question**:  Should we add a specific class on the language selector rather then removing it from all forms in the header? Or will we never want it on any form in the header?

avaleske (commented):
Do we maybe want to only take a screenshot of the question itself for most of these, rather then the whole page?

We can have one test that makes sure the question page in general is has the header but otherwise I don't think screenshotting the header on every question permutation is necessary.

server/app/assets/stylesheets/uswds/_uswds-theme-custom-styles.scss
greg-hyde: If there was another form in the header (which seems unlikely), we would also want to remove that style so it doesn't mess with the alignment of the header buttons

greg-hyde:
> Do we maybe want to only take a screenshot of the question itself for most of these, rather then the whole page?
> 
> We can have one test that makes sure the question page in general is has the header but otherwise I don't think screenshotting the header on every question permutation is necessary.

Yeah agreed. I just filed https://github.com/civiform/civiform/issues/7533 to cover that work

server/app/views/applicant/NorthStarApplicantBaseView.java
```diff
@@ -1,39 +1,125 @@
...
+    context.setVariable(
+        "smallLogoUrl",
+        settingsManifest
+            .getCivicEntitySmallLogoUrl()
+            .orElse(assetsFinder.path("Images/civiform-staging.png")));
+    context.setVariable(
+        "civicEntityShortName", settingsManifest.getWhitelabelCivicEntityShortName(request).get());
+    context.setVariable("closeIcon", Icons.CLOSE);
+
+    // Language selector params
```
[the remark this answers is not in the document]
avaleske: ahhh yeah I guess the dot notation expects it to be a field on a child object, and maybe the underscore notation is more clunky then it's worth. This is where having `record` classes would be nice.

and yeah totally fine in a followup.

avaleske (dismissed):

avaleske (dismissed):

avaleske (approved):

PR #7509, opened 2024-05-20
Render question text with markdown formatting

### LINKED ISSUE #6420: Parse Markdown in Question Text in all places for admin
**Describe the bug**
Markdown for question text is not being correctly parsed in all places for admin.

**To Reproduce**

1. Create a question and use markdown in the question text
2. Navigate to one of the places described below
3. See the unparsed markdown

**Expected behavior**
Markdown should be parsed in all places text is shown

**Screenshots**

Question list view:
image

Question sidebar:
image

Program edit view:
image

**Additional context**
Other places may need parsing as well. Look at all the places the question text is referenced.

jcma-google:
### Description

Renders question text with markdown formatting in five new places

Some screenshot samples
- question bank
image

- program edit view
image

- question list view
image

#### General

- [x ] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >

### Issue(s) this completes

Fixes #6420


server/app/views/admin/programs/ProgramBlocksView.java
```diff
@@ -434,7 +435,8 @@ private DivTag renderBlockPanel(
       ButtonTag blockDeleteModalButton,
       boolean isIntakeFormFeatureEnabled,
       Request request) {
-    // A block can only be deleted when it has no repeated blocks. Same is true for removing the
+    // A block can only be deleted when it has no repeated blocks. Same is true for
+    // removing the
```
rockycodes, line 439: nitpick: I guess `bin/fmt` decided to put the line breaks in different places 😂 will you move the next comment line (`// enumerator question from the block.`) up to this one so we don't have this weird two-word comment line? It looks like there are a bunch of places below where comments got split up in a strange way as well. Please fix those too.

server/app/views/components/TextFormatter.java
```diff
@@ -38,6 +38,11 @@ public static ImmutableList<DomContent> formatText(
     return builder.build();
   }
 
+  /** Passes provided text through Markdown formatter. */
```
rockycodes, line 41: nitpick: 
```suggestion
  /** Passes provided text through Markdown formatter without preserving empty lines or adding the required indicator. */
```

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -510,6 +510,39 @@ test.describe('program creation', () => {
     await expect(addressCorrectionInput).toHaveValue('false')
   })
 
+  test('create program with markdown questions', async ({
+    page,
+    adminQuestions,
+    adminPrograms,
+  }) => {
+    await loginAsAdmin(page)
+    await adminQuestions.addTextQuestion({
```
rockycodes, line 519: suggestion (blocking): Add text for the other fields where we expect markdown to show up (I think this might just be `helpText`, but I'm not positive)

browser-test/src/admin/admin_question_list.test.ts
```diff
@@ -31,25 +31,25 @@ test.describe('Admin question list', () => {
 
     // Most recently added question is on top.
     await expectQuestionListElements(adminQuestions, [
-      questionTwoPublishedText,
-      questionOnePublishedText,
+      'question list test question two\n',
```
rockycodes, line 34: suggestion (blocking): Reference the const here and then add `+ \n`. That way if someone changes the text for the const, they don't have to update it in multiple places.

browser-test/src/admin/admin_question_list.test.ts
```diff
@@ -73,17 +73,39 @@ test.describe('Admin question list', () => {
...
+      questionOneDraftText + '\n',
+      questionTwoPublishedText + '\n',
     ])
   })
 
+  test('rendering markdown works', async ({page, adminQuestions}) => {
+    await loginAsAdmin(page)
+    await adminQuestions.addTextQuestion({
+      questionName: 'a',
+      questionText: '*italics*',
```
rockycodes, line 91: suggestion (blocking): add `helpText` with markdown to these questions as well.

rockycodes (changes requested):
Looks good so far! I have a few blocking suggestions. In this PR, you're actually updating the question markdown so it shows up in 5 places (not 3):
- question bank
- program edit view
- question list view
- predicate configure view
- predicate edit view

Please add browser tests for all of those places to test that markdown renders.

browser-test/src/admin/admin_question_list.test.ts
jcma-google: I've gotten into the habit of not using variables in the expect clause and I'm interested in discussing whether that makes sense for these unit tests! 

My reasoning is that by using the literal and not obfuscating them with a variable name is it directly conveys the expected value instead of having to scroll up and find what the variable matches. I also get "variable blindness" in that the names "questionOnePublishedText" and "questionTwoPublishedText" are so similar that it takes me a second or two to understand which text is expected. I think the literal strings themselves are easier to read because of the spaces :) 

But I think variables make sense if we want to make multiple assertions about a single interaction / object / class or property or complex interactions. I also believe consistency is important! So don't want to change it only here and have it be different from the current testing pattern. 

Let me know what you think!

rockycodes: That makes sense, and I actually hadn't thought about that before! I don't have a strong preference as long as we're consistent. I'd say go with either for now and let's chat about it at the next eng meet. It looks like there are some places where you're still referencing the var name in the expect statement though (eg `questionOneDraftText + '\n',`), so please update those to strings if you go that route.

browser-test/src/admin/admin_program_creation.test.ts
rockycodes: I think this comment is not resolved yet. My suggestion was to add `helpText` with markdown in it so we can cover that use case as well.

rockycodes (changes requested):
There's just one outstanding comment (plus the comment about var names in tests) and I have some questions about these screenshots:
- `eligiblity-predicates-multi-values-multi-questions-predicate-edit` (what a name!), was regenerated, but I don't think there are any edits to that test that would require a new screenshot. That said, I think you do want to add some markdown to that test so this screenshot renders with markdown.
- same for `visibility-predicates-multi-questions-predicate-edit` (add some markdown)
- `operator-help-text.png` was also regenerated, but I believe there aren't any changes. If you delete the png and rebase off of main it should go away, but if it gets tricky with git it's fine to leave it.

jcma-google:
I added markdown and regenerated screenshot tests for the eligibility predicates. For `operator-help-text.png` the screenshot was regenerated without any changes and I think I'll have to rebase to remove the file as changed. I hope that leaving it as-is is okay!

browser-test/src/admin/admin_program_creation.test.ts
rockycodes: This is still not resolved. My suggestion is to add `helpText` with markdown in it.

browser-test/src/support/admin_questions.ts
```diff
@@ -1216,9 +1225,10 @@ export class AdminQuestions {
 
     await this.expectAdminQuestionsPageWithCreateSuccessToast()
 
-    expectedQuestionText = expectedQuestionText ?? questionText
-
-    await this.expectDraftQuestionExist(questionName, expectedQuestionText)
+    if (!markdown) {
```
rockycodes, line 1228: question (blocking): Why did you wrap this in `!markdown`? The way this is intended to work is you pass in `expectedQuestionText` if it's a markdown question and it checks for that first. https://github.com/civiform/civiform/pull/6688/files#diff-23ad99c6f330f49fe04c7f59414c9dd6933551c37ea33586d1094e92e1dcf016R1222

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -609,6 +618,8 @@ test.describe('create and edit predicates', () => {
     await adminQuestions.addDateQuestion({questionName: 'date-question'})
     await adminQuestions.addCurrencyQuestion({
       questionName: 'currency-question',
+      questionText: '*currency-question*',
```
rockycodes, line 621: nitpick (non-blocking): take the `-` out of the `questionText` since generally we use those in admin names and not question titles (aka `questionText`).

rockycodes (changes requested):
There is still an outstanding comment about including `helpText` with markdown in a test case (lmk if you think you don't need that). I also have a new question about an edit you made to `addTextQuestion` to skip the `expect` if it's a markdown question. 

Unfortunately, another PR went in that made changes to the `operator-help-text` screenshot, so you need to rebase on top of main in order for your tests to pass. Let me know if you want to pair on that!

browser-test/src/admin/admin_program_creation.test.ts
jcma-google: woops, missed this comment!

browser-test/src/support/admin_questions.ts
jcma-google: Because it's difficult to assert the markdown text without some additional manipulation which I am afraid would add brittleness to the test. For example, for the currency question the expected text is "*currency question*", with the asterisks because that's the pre-markdown rendering, and the received text is "currency question" because the markdown has now become stylistic rendering.

We have tests asserting the appearance of the markdown text in screenshots and I don't think we're losing much by skipping the innerText assertions (especially considering it would be onerous to assert that markdown is rendered correctly via presence of HTML elements).

rockycodes: That makes sense! In that case, can you remove the `expectedQuestionText` param and logic (it's only used [here](https://github.com/civiform/civiform/blob/15c8a10338b236b07db5767b27f88bbdea318b25/browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts#L359) I think) since we don't need it anymore?

rockycodes: Alternatively you could pass in "currency question" for `expectedTest` if we wanted to keep this check, but I don't feel tied to that since we have other tests and it seems like it's not super clear how to use this workaround I added.

jcma-google: done! removed the expectedQuestionText usage since that felt cleaner

browser-test/src/support/admin_questions.ts
```diff
@@ -1215,10 +1222,6 @@ export class AdminQuestions {
     await this.clickSubmitButtonAndNavigate('Create')
 
     await this.expectAdminQuestionsPageWithCreateSuccessToast()
-
-    expectedQuestionText = expectedQuestionText ?? questionText
-
-    await this.expectDraftQuestionExist(questionName, expectedQuestionText)
```
rockycodes, line 1221: suggestion (blocking): We still want to keep `this.expectDraftQuestionExist` for questions that don't include markdown. I just meant to remove the `expectedQuestionText` logic. So this would become:

```
if (!markdown) {
await this.expectDraftQuestionExist(questionName, questionText)
}
```

rockycodes (approved):
lgtm! 🥳 will you make an issue to refactor the other calls to `formatText` to use the new method you added (passing in false for preserveEmptyLines and addRequiredIndicator) if you haven't already?

PR #7521, opened 2024-05-21
Add descriptive text to the program translation page.

### LINKED ISSUE #3088: CiviForm admin UX:  add descriptive text to the translation page
The user's mental model leads them to believe that "Google Translate" is imbedded within CiviForm to manage translations.

Short-term: add descriptive text on the translation page that reads: “enter text in the language you want to translate to”.

Long-term: We have enough interest that we should look into which translation option entities want to use

**Additional context**
From admin UX feedback sheet

Screen Shot 2022-08-05 at 3 31 23 PM

ssandbekkhaug:
Also, delete redundant screenshots in the browser tests. Best practice is to assert that elements exist, which the tests already do. Otherwise, we get a lot of unnecessary diffs each time the page changes.

Fixes #3088

### Description

Add descriptive text to the program translation page.

#### Before
translations_before

#### After

Screenshot 2024-05-21 at 10 31 41 AM

#### General

[Follow these instructions](https://docs.civiform.us/user-manual/civiform-admin-guide/manage-translations-for-programs-and-questions#add-or-edit-program-translations) to navigate to the program translation page.

### Issue(s) this completes

Fixes #3088


browser-test/src/admin/admin_program_translation.test.ts
```diff
@@ -297,7 +309,6 @@ test.describe(
       // Go to the question translation page and add a translation for Spanish
       await adminQuestions.goToQuestionTranslationPage(questionName)
       await adminTranslations.selectLanguage('Spanish')
```
rockycodes, line 311: suggestion (maybe blocking): I don't think we're testing that specific elements are updated when we switch to Spanish translations, so we should either keep this screenshot or add a test that those specific elements show up (eg. button text switches to "Save Spanish updates". You could use a locator to generate a screenshot of just the section we care about.

browser-test/src/admin/admin_program_translation.test.ts
```diff
@@ -358,7 +369,6 @@ test.describe(
...
-      await validateScreenshot(page, 'multi-option-question-translation')
```
rockycodes, line 361: suggestion (blocking): I think we actually want to keep this screenshot since the page for translating a multi option question looks quite different than the page for translating other question types.

rockycodes (commented):
This looks good to me, but I have some questions about some of the screenshots you're deleting. I agree with you that some of the screenshots are redundant (we have a _lot_ of screenshots, so it would be great to align on this as a group in general), but I think others are testing valid things. I'm happy to be convinced otherwise though!

avaleske (dismissed):
Approving pending the changes Rocky requested. thanks!

browser-test/src/admin/admin_program_translation.test.ts
ssandbekkhaug: This and other validateScreenshot calls were moved to in #7532

rockycodes (approved):
lgtm!

PR #7564, opened 2024-05-24
Update addProgramBlockUsingSpec in admin_programs.ts to use BlockSpec

jeffpw-goog:
### Description

Update addProgramBlockUsingSpec BlockSpec instead of just QuestionSpec

#### General

### DIFF
--- a/browser-test/src/support/admin_programs.ts
+++ b/browser-test/src/support/admin_programs.ts
@@ -720,27 +720,20 @@ export class AdminPrograms {
     blockDescription = 'screen description',
     questionNames: string[] = [],
   ) {
-    const questionSpecs: QuestionSpec[] = questionNames.map((qName) => {
-      const questionSpec: QuestionSpec = {name: qName, isOptional: false}
-      return questionSpec
+    return await this.addProgramBlockUsingSpec(programName, {
+      description: blockDescription,
+      questions: questionNames.map((questionName) => ({name: questionName})),
     })
-    return await this.addProgramBlockUsingSpec(
-      programName,
-      blockDescription,
-      questionSpecs,
-    )
   }
 
   /**
-   * Creates a new program block with the given questions as defined by {@link QuestionSpec}.
+   * Creates a new program block as defined by {@link BlockSpec}.
    *
-   * Prefer this method over {@link #addProgramBlock}: This method provides the same functionality
-   * but also makes it easy to use optional questions.
+   * Prefer this method over {@link #addProgramBlock}.
    */
   async addProgramBlockUsingSpec(
     programName: string,
-    blockDescription = 'screen description',
-    questions: QuestionSpec[] = [],
+    block: BlockSpec,
     isProgramDisabled: boolean = false,
   ) {
     await this.gotoEditDraftProgramPage(programName, isProgramDisabled)
@@ -749,13 +742,13 @@ export class AdminPrograms {
     await waitForPageJsLoad(this.page)
 
     await clickAndWaitForModal(this.page, 'block-description-modal')
-    await this.page.fill('textarea', blockDescription)
+    await this.page.fill('textarea', block.description || 'screen description')
     await this.page.click('#update-block-button:not([disabled])')
     // Wait for submit and redirect back to this page.
     await this.page.waitForURL(this.page.url())
     await waitForPageJsLoad(this.page)
 
-    for (const question of questions) {
+    for (const question of block.questions) {
       await this.addQuestionFromQuestionBank(question.name)
       if (question.isOptional) {
         const optionalToggle = this.page
--- a/browser-test/src/admin/admin_program_migration.test.ts
+++ b/browser-test/src/admin/admin_program_migration.test.ts
@@ -1,4 +1,4 @@
-import {test, expect} from '../support/civiform_fixtures'
+import {expect, test} from '../support/civiform_fixtures'
 import {
   enableFeatureFlag,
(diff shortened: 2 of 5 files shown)


browser-test/src/support/admin_programs.ts
```diff
@@ -720,42 +730,42 @@ export class AdminPrograms {
...
+   * Prefer this method over {@link #addProgramBlock}.
    */
-  async addProgramBlockUsingSpec(
-    programName: string,
-    blockDescription = 'screen description',
-    questions: QuestionSpec[] = [],
-    isProgramDisabled: boolean = false,
-  ) {
-    await this.gotoEditDraftProgramPage(programName, isProgramDisabled)
+  async addProgramBlockUsingSpec(spec: ProgramBlockSpec) {
```
gwendolyngoetz, line 749:
```suggestion
  async addProgramBlockUsingSpec(programBlockSpec: ProgramBlockSpec) {
```
(blocking) Since we now have multiple interfaces with spec in the name let's write this out in full to avoid confusion

browser-test/src/support/admin_programs.ts
```diff
@@ -58,6 +58,16 @@ export interface BlockSpec {
   questions: QuestionSpec[]
 }
 
+export interface ProgramSelectorSpec {
```
gwendolyngoetz, line 61:
```suggestion
export interface ProgramDefinitionSpec {
```
(blocking) Selector makes me think this is going to be related to a css selector or a playwright locator. Maybe `ProgramDefinitionSpec`? 

Eventually I want to get to having a full model that matches the backend datamodel, but that is a much much larger refactor.

browser-test/src/support/admin_programs.ts
```diff
@@ -58,6 +58,16 @@ export interface BlockSpec {
...
+  name: string
+  isDisabled?: boolean
+}
+
+export interface ProgramBlockSpec {
+  programSelector: ProgramSelectorSpec
```
gwendolyngoetz, line 67: (thought) Maybe we don't make this an object and just put the two properties of `ProgramSelectorSpec` onto `ProgramBlockSpec`

gwendolyngoetz (changes requested):

browser-test/src/support/admin_programs.ts
jeffpw-goog: Happy to make this change - can you help me understand where it will add clarity? Where would one see the parameter name without the type that would be unclear now?

- When writing code that calls addProgramBlockUsingSpec, the IDE will show me both the name and type of the param
- In this method there's only one object called spec - I guess it would be clearer there to call it programBlockSpec?

Are there other cases I'm missing?

browser-test/src/support/admin_programs.ts
jeffpw-goog: Thanks for the feedback! The issue is that it is a selector, not a definition... addProgramBlockUsingSpec is finding (selecting) a program and then adding a block to it, not creating (defining) the program. So for example if programs had a category, and you had to know the category to navigate to the program, we would add a category field to the selector so the code could find the program. We wouldn't want the selector to include fields like externalLink or confirmationMessage since you don't find a program based on those fields.

I'm fine to inline the interface too - see the comment below about that.

browser-test/src/support/admin_programs.ts
jeffpw-goog: Yeah I considered this... I think the selector is more extensible and the naming makes it more clear that you are selecting rather than creating/defining (though apparently not that clear since you didn't find it so :|). It also avoids the need for /* isProgramDisabled = */ which authors might skip.

On the other side, it doesn't seem that likely that we'll need another selector criteria and if it's only clearer to me and prompts confusion with other selector concepts, then I'm fine to inline it.

What do you think?

jeffpw-goog: As a side thought, this selector spec pattern would be reusable/extendible for methods like goToEditBlockVisibilityPredicatePage that currently take a program name but could automatically support disabled programs if they used ProgramSelectorSpec (or a ProgramBlockSelectorSpec that extends it with also a block name).

gwendolyngoetz: I see where you're coming from and it's not a bad idea, but I don't know if we'll ever really need to be that extensible. Let's inline it for now and refactor it down the road if we find a more pressing need.

browser-test/src/support/admin_programs.ts
gwendolyngoetz: In a code review on github it defaults to showing only a few changed lines. With a more descriptive name reviewing it will be easier otherwise you'll have to click the show more button until you get to the method definition.

The same if looking at code on github itself.

Not all editors work the same showing the same information.

jeffpw-goog: Thanks for explaining! This is moot with the latest changes.

browser-test/src/support/admin_programs.ts
jeffpw-goog: It looks like I mis-interpreted your suggestion, or maybe I forgot what I was thinking yesterday, but anyway I don't like putting programName and isProgramDisabled into a "Spec" class because they are not part of a spec (i.e. entity definition which all the specs currently are) they're for finding the program. So I just changed the PR to use BlockSpec instead of QuestionSpec.

gwendolyngoetz (approved):

PR #7590, opened 2024-05-28
Update admin_predicates.test.ts to use *UsingSpec helper methods

jeffpw-goog:
### Description

Update admin_predicates.test.ts to use *UsingSpec helper methods

* Replace the deprecated editProgramBlock and addProgramBlock methods with their preferred equivalents editProgramBlockUsingSpec and addProgramBlockUsingSpec, respectively.
* Make BlockSpec.questions optional, and update the helper methods accordingly
* Refactor editProgramBlockWithOptional to be a little clearer

#### General

### DIFF
--- a/browser-test/src/admin/admin_predicates.test.ts
+++ b/browser-test/src/admin/admin_predicates.test.ts
@@ -1,4 +1,4 @@
-import {test, expect} from '../support/civiform_fixtures'
+import {expect, test} from '../support/civiform_fixtures'
 import {
   enableFeatureFlag,
   isHermeticTestEnvironment,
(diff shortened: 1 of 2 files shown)


browser-test/src/support/admin_programs.ts
```diff
@@ -597,34 +597,27 @@ export class AdminPrograms {
 
   /**
    * Edit basic block details and required and optional questions. Cannot handle more than one optional question.
-   * @deprecated prefer using {@link #editProgramBlockUsingSpec} instead. Be aware that the new method will place
-   * the optional question in the order as defined in the question array. The older method could only handle one
-   * optional question and forced it to be the first question on the list. Tests may need to be updated to handle
-   * a different question order.
+   * @deprecated prefer using {@link #editProgramBlockUsingSpec} instead. Be aware when migrating to
+   * editProgramBlockUsingSpec that this method always puts the optional question first in the block.
```
melanie-exygy, line 601: suggestion (non-blocking):  When you say "this method" it isn't clear to me if that's `editProgramBlockUsingSpec` or `editProgramBlockWithOptional`.  How about "Be aware... that the following method always puts the optional question first in the block whereas `editProgramBlockUsingSpec` places it in the order of the question array."

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -32,12 +32,14 @@ test.describe('create and edit predicates', () => {
 
     const programName = 'Create hide predicate'
     await adminPrograms.addProgram(programName)
-    await adminPrograms.editProgramBlock(programName, 'first screen', [
-      'hide-predicate-q',
-    ])
-    await adminPrograms.addProgramBlock(programName, 'screen with predicate', [
-      'hide-other-q',
-    ])
+    await adminPrograms.editProgramBlockUsingSpec(programName, {
```
melanie-exygy, line 35: praise:  Thanks for reducing tech debt by using the new function!

browser-test/src/support/admin_programs.ts
jeffpw-goog: Thanks! Updated to just use both method names explicitly to avoid any confusion.

melanie-exygy (approved):

PR #7598, opened 2024-05-29
Refactor questionBank sorting functionality to a more general utility function

jcma-google:
### Description

Extracts the sorting function to a generalized function. I'm adding a sorting mechanism to the TI list page and want to re-use the sorting functionality already created for sorting question banks. This change is expected to be a no-op. Separating out the refactor from the feature implementation is intended to make reviewing PRs easier. 

## Release notes

#### General

### Issue(s) this completes

### DIFF
--- a/server/app/assets/javascripts/sort_selector.ts
+++ b/server/app/assets/javascripts/sort_selector.ts
@@ -6,7 +6,7 @@
  * E.g. Sort by Name A-Z will look up the value for data attribute tiname-asc and sort the list by that value.
  *
  * @param listId id for finding the portion of the document that contains the sublistSelector and elementSelector
- * @param sublistSelector selector for the parent element of all of the elements to be sorted
+ * @param sublistSelector finds all of the elements containing to-be-sorted items. Can be multiple elements. 
  * @param elementSelector individual elements to be sorted
  */
 export function sortElementsByDataAttributes(
--- a/server/app/assets/javascripts/sort_selector.ts
+++ b/server/app/assets/javascripts/sort_selector.ts
@@ -10,13 +10,11 @@
  * @param elementSelector individual elements to be sorted
  */
 export function sortElementsByDataAttributes(
-    listId: string,
+  listId: string,
   sublistSelector: string,
   elementSelector: string,
 ): void {
-  const listToBeSorted = document.getElementById(
-    listId,
-  ) as HTMLSelectElement
+  const listToBeSorted = document.getElementById(listId) as HTMLSelectElement
   const sublists = document.querySelectorAll(sublistSelector)
   if (!listToBeSorted || !sublists) {
     return
--- a/server/app/assets/javascripts/sort_selector.ts
+++ b/server/app/assets/javascripts/sort_selector.ts
@@ -1,10 +1,10 @@
-/** 
- * Responsible for controlling the sorting mechanism used by QuestionBankController. Elements (tagged with elementSelector class) 
- * have data attributes with the value to be compared and sorted according to the selector input. 
- * 
+/**
+ * Responsible for controlling the sorting mechanism used by QuestionBankController. Elements (tagged with elementSelector class)
+ * have data attributes with the value to be compared and sorted according to the selector input.
+ *
  * E.g. Sort by Name A-Z will look up the value for data attribute tiname-asc and sort the list by that value.
- * 
- * @param elementListSelector selector for finding the portion of the document that contains the sublistSelector and elementSelector 
+ *
+ * @param elementListSelector selector for finding the portion of the document that contains the sublistSelector and elementSelector
  * @param sublistSelector selector for the parent element of all of the elements to be sorted
  * @param elementSelector individual elements to be sorted
  */
--- a/server/app/assets/javascripts/sort_selector.ts
+++ b/server/app/assets/javascripts/sort_selector.ts
@@ -6,7 +6,7 @@
  * E.g. Sort by Name A-Z will look up the value for data attribute tiname-asc and sort the list by that value.
  *
  * @param listId id for finding the portion of the document that contains the sublistSelector and elementSelector
- * @param sublistSelector finds all of the elements containing to-be-sorted items. Can be multiple elements. 
+ * @param sublistSelector finds all of the elements containing to-be-sorted items. Can be multiple elements.
  * @param elementSelector individual elements to be sorted
  */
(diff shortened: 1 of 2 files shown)


jcma-google:
removing nb1701 for review as he is traveling this week.

server/app/assets/javascripts/sort_selector.ts
```diff
@@ -0,0 +1,50 @@
+/** Responsible for controlling the sorting mechanism */
+export function sortSelectorElements(
+  sortSelectId: string,
+  sortElementId: string,
```
jeffpw-goog, line 4: I'm having some trouble understanding these param names and what this method does:
1. do the params all need to start with "sort"? Seems kind of redundant
2. sortElementId and sortSublistId are not IDs - querySelectorAll supports any selector and the use case in this PR is using a class selector. Would "elementSelector" and "sublistSelector" make more sense?
3. It seems like the hierarchy goes sublistSelector -> elementSelector, so would it make more sense to put them in that order?
4. Maybe add a jsdoc explanation for each param and what this method does in general? Based on the method name I first thought it would sort a <select> element but it's much too complicated for that.

server/app/assets/javascripts/sort_selector.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (!listToBeSorted || !sublists) {
+    return
+  }
+
+  sublists.forEach((sublist) => {
+    const el: HTMLElement[] = Array.from(
+      sublist.querySelectorAll(sortElementId),
+    )
+
+    const sortedElements = el.sort((elementA, elementB) => {
```
jeffpw-goog, line 20: Consider defining the comparator as a separate entity? (Actually there seem to be two nested comparators, so two separate entities?) It's quite long to have inlined here. I realize this is technically out of scope but with it living in its own file it seems more logical to do now.

server/app/assets/javascripts/sort_selector.ts
jcma-google: good suggestions! updated the list, param ordering and descriptions to be clearer based on your comments.

server/app/assets/javascripts/sort_selector.ts
jcma-google: I don't mind it the inlining because the code is straightfoward and readable and extracting the compartor functionality to a separate entity feels like it would  be more worth it if we had need of the comparator elsewhere and not just in this function. If you feel strongly about separating the comparators I can do that.

server/app/assets/javascripts/sort_selector.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * Responsible for controlling the sorting mechanism used by QuestionBankController. Elements (tagged with elementSelector class)
+ * have data attributes with the value to be compared and will sort list using the value. The data attribute is defined
+ * in QuestionSortOption.
+ *
+ * E.g. Sort by Name A-Z will look up the value for data attribute tiname-asc and sort the list by that value.
+ *
+ * @param listSelector selector for finding the portion of the document that contains the sublistSelector and elementSelector
```
jeffpw-goog, line 8: For this one it uses getElementById which takes an ID rather than a selector, so it should be named and described differently?

server/app/assets/javascripts/sort_selector.ts
jeffpw-goog: 👍

server/app/assets/javascripts/sort_selector.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param sublistSelector selector for the parent element of all of the elements to be sorted
+ * @param elementSelector individual elements to be sorted
+ */
+export function sortSelectorElements(
```
jeffpw-goog, line 12: What does "selector" mean here in the method name? If it's for sorting QuestionBanks (per the description) then would sortQuestionBank be clearer? Otherwise it sounds like it's a generic method for sorting... something :)

jcma-google: I originally named it sortElements (which is still generic sounding!) and then renamed it to be sortSelectorElements (still too generic!) and I think maybe a better name is sortElementsByDataAttributes?

jcma-google: I plan on using this for the TrustedIntermediaryController as well so didn't want to use a name that was too narrow in scope. 

https://github.com/civiform/civiform/pull/7600/files (branch  with my changes to add TrustedIntemediaryController)

server/app/assets/javascripts/sort_selector.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Responsible for controlling the sorting mechanism used by QuestionBankController. Elements (tagged with elementSelector class)
+ * have data attributes with the value to be compared and will sort list using the value. The data attribute is defined
+ * in QuestionSortOption.
+ *
+ * E.g. Sort by Name A-Z will look up the value for data attribute tiname-asc and sort the list by that value.
+ *
+ * @param listId id for finding the portion of the document that contains the sublistSelector and elementSelector
+ * @param sublistSelector selector for the parent element of all of the elements to be sorted
```
jeffpw-goog, line 9: (non-blocking) the sublistSelector can return multiple entities, and this method happily sorts each of them separately, which is not clear from this line

server/app/assets/javascripts/sort_selector.ts
jeffpw-goog: Makes sense - I do like sortElementsByDataAttributes better, since it captures the magic within the method.

Only remaining question is then shouldn't the jsdoc reflect the generic nature of the method and not refer to QuestionBank specifically?

jeffpw-goog (dismissed):

jcma-google:
Was trying too hard to write a method for the future without referencing non-existent methods and planned methods which worked about as okay as planned haha I should have been more explicit and communicative in the PR description. I'm going to update this file with the TrustedIntermediaryController reference in the next PR that adds that file.

server/app/assets/javascripts/sort_selector.ts
jcma-google: Was trying too hard to write a method for the future without referencing non-existent methods and planned methods which worked about as okay as planned haha I should have been more explicit and communicative in the PR description. I'm going to update this file with the TrustedIntermediaryController reference in the next PR that adds that file.

jeffpw-goog (approved):

PR #7618, opened 2024-05-31
Creating the PROGRAM_FILTERING feature flag

melanie-exygy:
### Description

Creating a new feature flag to guard user-facing aspects of the Program Category Filtering feature while it is under development.  This is the first step for #7610.  The name of the flag is "PROGRAM_FILTERING_ENABLED".  It is for use while the filtering feature is under development and will be removed once the feature is released in production.

## Release notes

Creating a new feature flag to guard user-facing aspects of the Program Category Filtering feature while it is under development.  The name of the flag is "PROGRAM_FILTERING_ENABLED".

### DIFF
--- a/server/app/services/settings/SettingsManifest.java
+++ b/server/app/services/settings/SettingsManifest.java
@@ -941,8 +941,8 @@ public final class SettingsManifest extends AbstractSettingsManifest {
   }
 
   /** (NOT FOR PRODUCTION USE) Enables filtering programs by category on the homepage */
-  public boolean getProgramFiltering(RequestHeader request) {
-    return getBool("PROGRAM_FILTERING", request);
+  public boolean getProgramFilteringEnabled(RequestHeader request) {
+    return getBool("PROGRAM_FILTERING_ENABLED", request);
   }
 
   private static final ImmutableMap<String, SettingsSection> GENERATED_SECTIONS =
@@ -1953,7 +1953,7 @@ public final class SettingsManifest extends AbstractSettingsManifest {
                       SettingType.BOOLEAN,
                       SettingMode.ADMIN_WRITEABLE),
                   SettingDescription.create(
-                      "PROGRAM_FILTERING",
+                      "PROGRAM_FILTERING_ENABLED",
                       "(NOT FOR PRODUCTION USE) Enables filtering programs by category on the"
                           + " homepage",
                       /* isRequired= */ false,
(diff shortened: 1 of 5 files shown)


server/app/services/settings/SettingsManifest.java
```diff
@@ -940,6 +940,11 @@ public boolean getProgramMigrationEnabled(RequestHeader request) {
     return getBool("PROGRAM_MIGRATION_ENABLED", request);
   }
 
+  /** (NOT FOR PRODUCTION USE) Enables filtering programs by category on the homepage */
+  public boolean getProgramFiltering(RequestHeader request) {
+    return getBool("PROGRAM_FILTERING", request);
```
rockycodes, line 945: nitpick (maybe blocking): I think the pattern we've established with feature flags is to include "enabled" in the name (see `PROGRAM_MIGRATION_ENABLED` above). I'm not sure exactly when we decided on this or why, but my guess is because it makes the function name super clear (eg. `getProgramFilteringEnabled`).

rockycodes (changes requested):

server/app/services/settings/SettingsManifest.java
melanie-exygy: I saw that about half of the flags end with "enabled", but I don't like that pattern because it makes it seem like the flag itself is always enabled.  Do we then ask ourselves "Is Program Migration Enabled enabled?"  It's redundant.  Maybe we should bring this up at the eng sync on Monday.  If we want to make the function name more clear we could include the word "Flag" at the end of each FF name.  Then, the function would be `getProgramFilteringFlag`.  And we could say "Is the Program Filtering Flag enabled?"

rockycodes: I agree! Good idea to discuss on Monday.

rockycodes (approved):
approved once you fix it in the conf file!

server/conf/helper/feature-flags.conf
```diff
@@ -65,6 +65,10 @@ save_on_all_actions = ${?SAVE_ON_ALL_ACTIONS}
 north_star_applicant_ui = false
 north_star_applicant_ui = ${?NORTH_STAR_APPLICANT_UI}
 
+# Program filtering
+program_filtering_enabled = false
+program_filtering_enabled = ${?PROGRAM_FILTERING}
```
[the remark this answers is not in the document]
melanie-exygy: Done.  Thanks for the code suggestion!

PR #7627, opened 2024-05-31
Update question names for admin_predicates tests

jeffpw-goog:
### Description

Update question names for 'test predicates' tests in admin_predicates.test.ts
* Update question names to say "question" in them
* Update question names to describe the question type rather than the predicate they're later going to be associated with. Append the predicate type to the name if there is more than one of the same type.
* Move 'text-question-last-page' question to the visibility test because it's only used in that test

#### General


browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -693,40 +693,40 @@ test.describe('create and edit predicates', () => {
...
-        questionName: 'predicate-date-age-younger-than',
+        questionName: 'date-question-age-younger-than',
       })
       await adminQuestions.addDateQuestion({
-        questionName: 'predicate-date-age-between',
+        questionName: 'date-question-age-between',
       })
       await adminQuestions.addCheckboxQuestion({
-        questionName: 'both sides are lists',
+        questionName: 'checkbox-question',
```
melanie-exygy, line 723: question (non-blocking):  Are we losing some information in the name about both sides being lists (whatever that means)?

melanie-exygy (commented):
I really like these name changes.  It makes things much clearer.  Here's an idea for making these tests even easier to read:  You could add Playwright test steps.  Here's an example of where we do that:  [test steps](https://github.com/civiform/civiform/blob/6114376e95fec8e8693984fa108fb3eed275c395/browser-test/src/trusted_intermediary.test.ts#L1014).  This could be in a separate PR or, if you'd rather move on to working on something other than refactoring these tests, I completely understand.  :-)

jeffpw-goog:
> I really like these name changes. It makes things much clearer. Here's an idea for making these tests even easier to read: You could add Playwright test steps. Here's an example of where we do that: [test steps](https://github.com/civiform/civiform/blob/6114376e95fec8e8693984fa108fb3eed275c395/browser-test/src/trusted_intermediary.test.ts#L1014). This could be in a separate PR or, if you'd rather move on to working on something other than refactoring these tests, I completely understand. :-)

That is two PRs down the stack :) https://github.com/civiform/civiform/pull/7649

browser-test/src/admin/admin_predicates.test.ts
jeffpw-goog: This string still appears where the test actually configures the predicate, which is what it's referring to, so I think it's actually clearer to remove it here because it is confusing. :)

Thanks!

melanie-exygy (approved):

PR #7629, opened 2024-05-31
Integrating the last few translations for the TI client creation and edit forms

melanie-exygy:
### Description

We recently added translations for most of the TI dashboard content.  However, a few validation error messages on the client creation and edit forms could not be integrated until they were translated by Transifex.  This PR integrates those translated error messages and submits two new strings for translation.  

Note that I am already using those two untranslated strings as validation messages, but we were already using the hard-coded English strings anyway.  By incorporating the messages into the code now, the email validation errors will show as translated automatically once the Transifex translations are completed.

1. From the Dev Tools, log in as a Trusted Intermediary.
2. From the TI dashboard, select a language other than English.
3. Click on the "Add client" button.
4. Click on the blue "Save" button at the bottom of the form.
5. Note that the validation error messages are translated.


server/app/services/ti/TrustedIntermediaryService.java
```diff
@@ -73,8 +73,7 @@ public AddNewApplicantReturnObject addNewClient(
       return new AddNewApplicantReturnObject(
           form.withError(
               FORM_FIELD_NAME_EMAIL_ADDRESS,
-              "Email address already in use. Cannot create applicant if an account already"
-                  + " exists."));
+              preferredLanguage.at(MessageKey.ERROR_EMAIL_IN_USE_CLIENT_CREATE.getKeyName())));
```
azhangggggg, line 76: (question, non-blocking) do we have browser tests for this page but with a non-english language?

melanie-exygy: We don't have tests for the translated validation errors.  I thought about adding one, but remembered a conversation we had in the Eng Sync meeting (see notes from 9/18/23) concerning whether to add screenshots of every applicant-facing page in another language so that we can test that all strings are translated.  We decided not to do that because of the bulk it would add to our already-long test suite.  Our decision was, rather, to add a checkbox on the PR template making sure that all new strings were translated.

azhangggggg (approved):

PR #7649, opened 2024-06-03
Rearrange predicate tests

jeffpw-goog:
### Description

Rearrange predicate tests:
* Group the three program configuration steps (add question, add screen, configure eligibility/visibility predicate) into one section per screen
* Add test.steps for the configuration and application sections with screen numbers so it's easier to match up the two sections
* Remove the test.describe('test predicates') group now that there is no shared state between the tests

#### General

### DIFF
--- a/browser-test/src/admin/admin_predicates.test.ts
+++ b/browser-test/src/admin/admin_predicates.test.ts
@@ -689,115 +689,36 @@ test.describe('create and edit predicates', () => {
     )
   })
 
-  test.describe('test predicates', () => {
-    test.beforeEach(async ({page, adminQuestions}) => {
-      await loginAsAdmin(page)
+  test('every visibility right hand type evaluates correctly', async ({
+    page,
+    adminPrograms,
+    applicantQuestions,
+    adminPredicates,
+    adminQuestions,
+  }) => {
+    test.slow()
 
+    await loginAsAdmin(page)
+
+    const programName = 'Test all visibility predicate types'
+    await adminPrograms.addProgram(programName)
+
+    // Configure each screen so that it is not shown until the question on
+    // the screen before it is answered a certain way
+    await test.step('Configure screen 1 - question only', async () => {
       await adminQuestions.addNameQuestion({questionName: 'name-question'})
-      await adminQuestions.addTextQuestion({questionName: 'text-question'})
-      await adminQuestions.addNumberQuestion({
-        questionName: 'number-question-equal-to',
-      })
-      await adminQuestions.addNumberQuestion({
-        questionName: 'number-question-one-of',
-      })
-      await adminQuestions.addCurrencyQuestion({
-        questionName: 'currency-question',
-      })
-      await adminQuestions.addDateQuestion({
-        questionName: 'date-question-is-earlier-than',
-      })
-      await adminQuestions.addDateQuestion({
-        questionName: 'date-question-on-or-after',
-      })
-      await adminQuestions.addDateQuestion({
-        questionName: 'date-question-age-older-than',
-      })
-      await adminQuestions.addDateQuestion({
-        questionName: 'date-question-age-younger-than',
-      })
-      await adminQuestions.addDateQuestion({
-        questionName: 'date-question-age-between',
-      })
-      await adminQuestions.addCheckboxQuestion({
-        questionName: 'checkbox-question',
-        options: [
-          {adminName: 'dog_admin', text: 'dog'},
-          {adminName: 'rabbit_admin', text: 'rabbit'},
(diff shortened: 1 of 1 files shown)


browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -1054,211 +1086,266 @@ test.describe('create and edit predicates', () => {
...
-      await adminPrograms.goToEditBlockEligibilityPredicatePage(
-        programName,
-        'Screen 2',
-      )
-      await adminPredicates.addPredicates({
-        questionName: 'text-question',
-        scalar: 'text',
-        operator: 'is one of',
-        value: 'blue, green',
+      await test.step('Configure screen 2', async () => {
```
melanie-exygy, line 1116: suggestion (non-blocking):  In the visibility predicates test, you added the operator to the description like this "Configure screen 2 - string equal to".  I think we should be consistent and do the same thing in the eligibility test.

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -1054,211 +1086,266 @@ test.describe('create and edit predicates', () => {
...
-        programName,
-        'Screen 5',
-      )
-      await adminPredicates.addPredicates({
-        questionName: 'currency-question',
-        scalar: 'currency',
-        operator: 'is greater than',
-        value: '100.01',
+      await test.step('Configure screen 5', async () => {
+        await adminQuestions.addCurrencyQuestion({
```
melanie-exygy, line 1178: comment (non-blocking):  I'm torn because I like the clarity of creating the question and adding it to the block in the same step as where we add the predicate, but I think it would improve the efficiency of the tests to create the questions and add them to the blocks in a `beforeAll` clause.  What are your thoughts about that option?

jeffpw-goog: Also https://github.com/civiform/civiform/wiki/Playwright-Feature-Migration-Guide says "The beforeAll and afterAll functions do not have access to the page object and many other fixtures" and generally discourages beforeAll... and it looks like we currently have no usages of it. I guess how much of a performance improvement do you think it might be, for it to be worth all that?

melanie-exygy: I didn't realize that `beforeAll` was discouraged.  Good to know!  I guess it's better to stick with duplicating the question and block creation in that case.

browser-test/src/admin/admin_predicates.test.ts
melanie-exygy: I actually think the step name in the visibility test was confusing me, so I agree with you that removing it for either test is the best way to go.

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -742,202 +701,251 @@ test.describe('create and edit predicates', () => {
...
-      )
-      await adminPredicates.addPredicates({
-        questionName: 'number-question-equal-to',
-        action: 'shown if',
-        scalar: 'number',
-        operator: 'is equal to',
-        value: '42',
+      await test.step('Configure screen 4 - long equal to', async () => {
+        await adminQuestions.addNumberQuestion({
+          questionName: 'number-question-one-of',
```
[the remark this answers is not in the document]
melanie-exygy: Thanks for explaining.  I realize now that I was confused about each step using the question from the previous screen as the visibility condition.  I think seeing the operator in the step description followed by a different operator in the question creation was what got me confused.  If you remove the operator from the test description, I think it'll be clearer.  No need to refactor everything again.

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -742,202 +701,251 @@ test.describe('create and edit predicates', () => {
...
-        'Screen 6',
-      )
-      await adminPredicates.addPredicates({
-        questionName: 'currency-question',
-        action: 'shown if',
-        scalar: 'currency',
-        operator: 'is greater than',
-        value: '100.01',
+      await test.step('Configure screen 6 - currency greater than', async () => {
+        await adminQuestions.addDateQuestion({
```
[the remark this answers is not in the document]
melanie-exygy: Yes, I was confused about each step referring to the previous screen.  I understand now.

browser-test/src/admin/admin_predicates.test.ts
jeffpw-goog: Done!

browser-test/src/admin/admin_predicates.test.ts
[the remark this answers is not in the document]
jeffpw-goog: Done!

melanie-exygy (approved):

PR #7655, opened 2024-06-03
Only show login dialogs once per session

### LINKED ISSUE #7608: [North Star Applicant UI] Only show login modal once per session
Use local storage to record/check whether the user has already seen a login dialog, and only show the user one login prompt dialog per session

This will likely require some typescript

greg-hyde:
### Description

This adds some typescript logic to only show the login modal dialog once per session. It lives separately from the pre-northstar modal typescript logic because it operates differently.

USWDS has javascript that adds event handlers that handle the opening and closing of the dialog already, so this logic:
1) on load, check if dialog has already been seen (via localstorage), and if so, update the href of the anchor tag that launched the dialog to go straight to the bypass URL
2) when the modal dialog is launched, update localstorage and then update buttons that link to that dialog

There is a bit of weirdness because USWDS makes changes/adds event listeners on load, so merely changing the href of the anchor element isn't enough. Instead, we need to make a new <a> tag and replace the old one so it won't open the modal.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #7608

### DIFF
--- a/server/app/assets/javascripts/north_star_modal.ts
+++ b/server/app/assets/javascripts/north_star_modal.ts
@@ -8,7 +8,11 @@ export class NorthStarModalController {
     // When the button is clicked to trigger the modal, save to local storage that that modal type
     // has been shown.
     const onlyShowOnceGroup = modal.getAttribute('only-show-once-group')
-    const modalButtons = document.querySelectorAll(`a[href='#${modal.id}']`)
+    // We use dialog-id instead of the id attribute because USWDS does some DOM manipulation and
+    // moves the id to a wrapper element, so it no longer points to the dialog itself.
+    const modalButtons = document.querySelectorAll(
+      `a[href='#` + modal.getAttribute('dialog-id') + `']`,
+    )
     if (onlyShowOnceGroup) {
       modalButtons.forEach((modalButton) => {
         modalButton.addEventListener('click', () => {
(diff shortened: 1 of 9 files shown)


greg-hyde:
* **#7655** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/assets/javascripts/north_star_modal.ts
```diff
@@ -0,0 +1,71 @@
...
+          NorthStarModalController.maybeUpdateButtonHrefToBypassUrl(modal)
+        })
+      })
+    }
+  }
+
+  static maybeUpdateButtonHrefToBypassUrl(modal: HTMLElement) {
+    const onlyShowOnceGroup = modal.getAttribute('only-show-once-group')
+    if (onlyShowOnceGroup) {
+      const modalHasBeenShown = localStorage.getItem(onlyShowOnceGroup)
```
rockycodes, line 29: question (non-blocking): the way this works now, this will only ever be true for `login-prompt` right? do all modal types use the `ModalFragment`?

greg-hyde: for right now yes. In J2HTML there are two only-show-once-groups: 1 for login dialogs in the program index page and another in the review page when the user is redirected from a program slug, but I haven't implemented the latter yet

greg-hyde: and all modals in the northstar work will use modalfragment. there will be a little bit of refactoring required for that template to ensure the right ones will be present on the right pages as I build out all of the available modals

avaleske (approved):

PR #7679, opened 2024-06-05
Clean up test helpers

gwendolyngoetz:
### Description

Clean up some spots to use locators. Add some options for performance. Prep for a new test suite I have coming.

#### General

### DIFF
--- a/browser-test/src/support/admin_programs.ts
+++ b/browser-test/src/support/admin_programs.ts
@@ -759,7 +759,7 @@ export class AdminPrograms {
     editBlockScreenDetails: boolean = true,
   ) {
     await this.gotoEditDraftProgramPage(programName, isProgramDisabled)
-    return await this.addProgramBlockUsingSpecAlreadyOnEditDraftPage(
+    return await this.addProgramBlockUsingSpecWhenAlreadyOnEditDraftPage(
       block,
       editBlockScreenDetails,
     )
@@ -772,7 +772,7 @@ export class AdminPrograms {
    * @param {BlockSpec} block Desired block settings
    * @param {boolean} editBlockScreenDetails Defaults to true. If true the block name and description will be updated; if false they will not.
    */
-  async addProgramBlockUsingSpecAlreadyOnEditDraftPage(
+  async addProgramBlockUsingSpecWhenAlreadyOnEditDraftPage(
     block: BlockSpec,
     editBlockScreenDetails: boolean = true,
   ) {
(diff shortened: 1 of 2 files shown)


browser-test/src/support/admin_programs.ts
```diff
@@ -651,6 +651,28 @@ export class AdminPrograms {
     }
   }
 
+  /**
+   * Add questions to specified block. You must already be on the admin program edit block page,
+   * but if you are this is a significantly faster way to add multiple questions since it does
+   * not return to the program list page each time and navigate back to the block.
+   * @param {BlockSpec} block Block information
+   */
+  async addQuestionsToProgramBlock(block: BlockSpec) {
```
jeffpw-goog, line 660: Should this method name be suffixed with "alreadyOnEditBlockPage" like addProgramBlockUsingSpecAlreadyOnEditDraftPage below? The jsdoc does say you must already be on the admin program edit block page, but it reads more as a "by the way" than a hard prerequisite.

browser-test/src/support/admin_programs.ts
```diff
@@ -724,23 +746,50 @@ export class AdminPrograms {
...
+   * @param {string} programName Name of the program
+   * @param {BlockSpec} block Desired block settings
+   * @param {boolean} isProgramDisabled Defaults to false. Flag to determine if the program status is disabled or not
+   * @param {boolean} editBlockScreenDetails Defaults to true. If true the block name and description will be updated; if false they will not.
    */
   async addProgramBlockUsingSpec(
     programName: string,
     block: BlockSpec,
     isProgramDisabled: boolean = false,
+    editBlockScreenDetails: boolean = true,
```
jeffpw-goog, line 759: (non-blocking) This is kinda what I was worried about with the ProgramBlockSpec refactor... once you have more than one optional param, you have to specify all of them if you want to specify the last one, so you start hard-coding isProgramDisabled = false in places where it's not really helpful to the test. Which I guess isn't the worst thing, it's just more noise and doesn't scale too far.

I don't really have a suggestion, just thinking out loud. :)

jeffpw-goog (approved):

browser-test/src/support/admin_programs.ts
gwendolyngoetz: I'm only just starting to push jsdoc so most things lack them. I don't know if people will be completely aware to check at this time. But I'll make the method be 'WhenAlreadyOnDraftPage'.

browser-test/src/support/admin_programs.ts
gwendolyngoetz: I'm starting thinking through ideas for a much broader change. I'm growing increasingly irritated at the run time of the tests. This is just a means to an end for now.

PR #7727, opened 2024-06-11
Refresh enumerator entity add button status on page load

### LINKED ISSUE #7733: Enumerator response add button disable logic is not applied on page load
**Describe the bug**
When filling out an enumerator question in an application, the add button is re-enabled erroneously when you click "Save and next" with an empty entity.

**To Reproduce**

1. Set up a program with an enumerator question 
2. Preview the program as applicant
3. Add an enumerator entity and leave it empty
5. Observe that the Add button is disabled
6. Click "Save and next"
7. Observe that an error shows asking you to enter a value for each line
8. Observe that the Add button is re-enabled

**Expected behavior**
The add button should remain disabled as long as there is an empty entity.

**Screenshots**
Expected: 
image

Actual:
image

jeffpw-goog:
### Description

Refresh enumerator entity add button status on page load

Also refactor refreshAddButtonStatus so that it works when run on a page without an enumerator, and combines the two selector calls into one.

#### General

### Issue(s) this completes

Fixes #7733

### DIFF
--- a/browser-test/src/end_to_end_enumerators.test.ts
+++ b/browser-test/src/end_to_end_enumerators.test.ts
@@ -354,6 +354,19 @@ test.describe('End to end enumerator test', () => {
           page.locator('#enumerator-field-add-button'),
         ).toHaveAttribute('disabled')
       })
+
+      await test.step('Add button is still disabled after trying to save', async () => {
+        await applicantQuestions.clickNext()
+
+        // Error shows because of the empty entity
+        await expect(
+          page.locator('.cf-applicant-question-errors'),
+        ).toBeVisible()
+
+        await expect(
+          page.locator('#enumerator-field-add-button'),
+        ).toHaveAttribute('disabled')
+      })
     })
 
     test('Applicant can navigate to previous blocks', async ({
--- a/server/app/assets/javascripts/enumerator.ts
+++ b/server/app/assets/javascripts/enumerator.ts
@@ -4,6 +4,8 @@
 import {addEventListenerToElements, assertNotNull} from './util'
 
 export function init() {
+  refreshAddButtonStatus()
+
   addEventListenerToElements(
     '.cf-question-enumerator input[data-entity-input]',
     'input',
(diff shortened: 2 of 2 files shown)


server/app/assets/javascripts/enumerator.ts
```diff
@@ -169,21 +171,18 @@ function setFocusAfterEnumeratorRemoval() {
...
-  const enumeratorQuestion = assertNotNull(
-    document.querySelector('.cf-question-enumerator'),
+  const enumeratorInputValues = Array.from(
+    document.querySelectorAll(
+      '.cf-question-enumerator input[data-entity-input]',
+    ),
+  ).map((item) => (item as HTMLInputElement).value)
+
+  // validate that there are no empty inputs.
+  const addButton = <HTMLInputElement>(
```
melanie-exygy, line 181: suggestion (non-blocking):  Is there a need to cast this button as an input element?  Also, It would be best to stay consistent in the way you cast.  So, either using `as` or `<>`, but not mixing both.

server/app/assets/javascripts/enumerator.ts
```diff
@@ -4,6 +4,8 @@
 import {addEventListenerToElements, assertNotNull} from './util'
 
 export function init() {
+  refreshAddButtonStatus()
```
melanie-exygy, line 7: praise:  Clever solution to just check for empty inputs on page load.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -354,6 +354,18 @@ test.describe('End to end enumerator test', () => {
           page.locator('#enumerator-field-add-button'),
         ).toHaveAttribute('disabled')
       })
+
+      await test.step('Add button is still disabled after trying to save', async () => {
+        await applicantQuestions.clickNext()
+        // Verify that the error message is visible.
+        await expect(
+          page.locator('.cf-applicant-question-errors:visible'),
+        ).toContainText('Error: Please enter a value for each line.')
```
melanie-exygy, line 363: suggestion (non-blocking):  As much as possible, it's best to avoid relying on text either for locators or assertions because if that text changes, we'll have to update the test.  You could use `.toBeVisible()` instead.

browser-test/src/end_to_end_enumerators.test.ts
```diff
@@ -354,6 +354,18 @@ test.describe('End to end enumerator test', () => {
...
+
+        await expect(
+          page.locator('#enumerator-field-add-button'),
+        ).toHaveAttribute('disabled')
```
melanie-exygy, line 367: suggestion(non-blocking):  We started trying to use Playwright [auto-retrying assertions](https://playwright.dev/docs/test-assertions), so rather than `.toHaveAttribute('disabled')`, you could use `toBeDisabled()`

browser-test/src/end_to_end_enumerators.test.ts
jeffpw-goog: Interesting point - the unit tests already validate the error message text and that the errors occur at the right times, so the browser tests are somewhat duplicative in that sense (I copied this from a test around line 200). So I guess the question is how much should the browser tests test - a screenshot with a single error case (or two at once, maybe) could validate that (and how) error messages are presented in general.

But actually in this case I added this more so that the test reads clearly and to confirm that the page is in the expected state when validating that the add button is disabled (i.e. we're not actually on the next page where it is disabled for other reasons). I guess toBeVisible with a comment is sufficient to cover that. Updated to do that.

Thanks!

melanie-exygy (approved):

Issue #7228, opened 2024-04-15
[North Star Applicant UI] Render Toasts in the applicant view


greg-hyde:
#### Describe the solution presented in this issue.
Render toasts in the applicant view using Thymeleaf. Toasts are currently all rendered in the header, but then removed and reshown via J2HTML in toast.ts, even when we render toasts in Thymeleaf, they will be removed and re-rendered in J2HTML, so a further refactor will be needed

Block edit page
 - Eligibility screener success
 - Locale not supported warning

Review Page
 - TBD

Card list page 
 - TBD

#### Done when:
What criteria need to be met for the issue to be finished? 
- [ ] All toasts currently present in applicant side of app are replicated in Thymeleaf
- [ ] Typescript code has been refactored so that the shown toasts are actually Thymeleaf and J2HTML is not involved

greg-hyde:
Remaining toasts:

Review page: 
1) [error saving application, banner](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/ApplicantProgramReviewController.java#L109)
1) [success-banner](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/ApplicantProgramReviewController.java#L111) 
1) [Not eligible banner](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/ApplicantProgramReviewController.java#L125)

Program index view: 
1) [error, banner](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/ApplicantProgramsController.java#L92)
1) [session ended](https://github.com/civiform/civiform/blob/main/server/app/views/applicant/ProgramIndexView.java#L79)

Upsell view (maybe including common intake):
1) [error](https://github.com/civiform/civiform/blob/main/server/app/controllers/applicant/UpsellController.java#L149)

TI account settings (if migrated): error
TI Client List View (if migrated): error
General: 
1) [dev/staging warning](https://github.com/civiform/civiform/blob/main/server/app/views/BaseHtmlLayout.java#L116)

PR #7731, opened 2024-06-11
Add toasts to summary/review page

greg-hyde:
### Description

Add toasts in the review/summary page to thymeleaf

- Not eligible [
5vNVfti6p6XGivZ
](url)
- May be eligible
- Errors

#### General

#### User visible changes

### Issue(s) this completes

Progress towards #7228

### DIFF
--- a/browser-test/src/applicant/navigation/application_navigation_flow.test.ts
+++ b/browser-test/src/applicant/navigation/application_navigation_flow.test.ts
@@ -110,7 +110,7 @@ test.describe('Applicant navigation flow', () => {
           )
         })
 
-        test('shows error with incomplete submission', async ({
+        test('shows error toast with incomplete submission', async ({
           page,
           applicantQuestions,
         }) => {
--- a/browser-test/src/applicant/navigation/application_navigation_flow.test.ts
+++ b/browser-test/src/applicant/navigation/application_navigation_flow.test.ts
@@ -95,8 +95,11 @@ test.describe('Applicant navigation flow', () => {
       'review page with North Star enabled',
       {tag: ['@northstar']},
       () => {
-        test('validate screenshot', async ({page, applicantQuestions}) => {
+        test.beforeEach(async ({page}) => {
           await enableFeatureFlag(page, 'north_star_applicant_ui')
+        })
+
+        test('validate screenshot', async ({page, applicantQuestions}) => {
           await applicantQuestions.clickApplyProgramButton(programName)
 
           await validateScreenshot(
@@ -104,6 +107,34 @@ test.describe('Applicant navigation flow', () => {
             'north-star-program-preview',
             /* fullPage= */ true,
             /* mobileScreenshot= */ true,
+          )
+        })
+
+        test('shows error with incomplete submission', async ({
+          page,
+          applicantQuestions,
+        }) => {
+          test.slow()
+
+          await applicantQuestions.clickApplyProgramButton(programName)
+
+          // The UI correctly won't let us submit because the application isn't complete.
+          // To fake submitting an incomplete application add a submit button and click it.
+          // Note the form already triggers for the submit action.
+          // A clearer way to set this up would be to have two browser contexts but that isn't doable in our setup.
+          await page.evaluate(() => {
+            const buttonEl = document.createElement('button')
+            buttonEl.id = 'test-form-submit'
+            buttonEl.type = 'submit'
+            const formEl = document.querySelector('.cf-debounced-form')!
+            formEl.appendChild(buttonEl)
+          })
+          const submitButton = page.locator('#test-form-submit')
+          await submitButton.click()
+
+          await validateToastMessage(
+            page,
+            "Error: There's been an update to the application",
           )
         })
       },
--- a/browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts
+++ b/browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts
@@ -100,67 +100,6 @@ test.describe('Applicant navigation flow', () => {
       await validateAccessibility(page)
     })
 
-    test(
-      'North Star shows ineligible on home page',
-      {tag: ['@northstar']},
-      async ({page, applicantQuestions}) => {
-        await enableFeatureFlag(page, 'north_star_applicant_ui')
-        await applicantQuestions.applyProgram(fullProgramName)
-
-        await test.step('fill out application and submit', async () => {
-          await applicantQuestions.answerNumberQuestion('1')
-          await applicantQuestions.clickContinue()
-          await applicantQuestions.expectIneligiblePage()
-        })
-
-        await test.step('verify question is marked ineligible', async () => {
-          await applicantQuestions.gotoApplicantHomePage()
-          await applicantQuestions.seeEligibilityTag(
-            fullProgramName,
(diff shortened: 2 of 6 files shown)


greg-hyde:
* **#7731** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/views/components/ToastFragment.html
```diff
@@ -43,3 +51,34 @@
     ></div>
   </th:block>
 </div>
+
+<!--/* Render the toasts used in the Review/Summary page */-->
+<div th:fragment="summaryToasts">
```
melanie-exygy, line 64: question (non-blocking):  Does this mean that the "summaryToasts" fragment is only one of the fragments defined in `ToastFragment.html`?  If so, shouldn't the file be called `ToastFragments.html`?

server/app/views/components/ToastFragment.html
```diff
@@ -43,3 +51,34 @@
...
+  <!--/* Success toast */-->
```
melanie-exygy, line 65: question (blocking):  Don't we use these same toasts on the homepage and in the application blocks?  If so, then maybe it would be better to call this fragment "eligibilityToasts"?  I like the idea of making such small components reusable, rather than limiting them to use on a specific page.

server/app/views/components/ToastFragment.html
```diff
@@ -14,6 +14,22 @@
   ></div>
 </div>
 
+<div
+  th:fragment="alert(id, message, canDismiss, canIgnore, toastDuration, condOnStorageKey)"
+>
+  <div
```
melanie-exygy, line 20: question (non-blocking):  Can we include `role="alert"` or `aria-live` so that the toasts will be a11y compliant?

browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts
```diff
@@ -452,5 +369,106 @@ test.describe('Applicant navigation flow', () => {
       await applicantQuestions.gotoApplicantHomePage()
       await applicantQuestions.seeNoEligibilityTags(fullProgramName)
     })
+
+    test.describe('With north star flag enabled', {tag: ['@northstar']}, () => {
+      test.beforeEach(async ({page}) => {
+        await enableFeatureFlag(page, 'north_star_applicant_ui')
+      })
+
+      test('Shows ineligible on home page', async ({
```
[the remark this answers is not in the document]
greg-hyde: done

browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts
```diff
@@ -452,5 +369,106 @@ test.describe('Applicant navigation flow', () => {
...
+        applicantQuestions,
+      }) => {
+        await applicantQuestions.applyProgram(fullProgramName)
+
+        await test.step('fill out application and submit', async () => {
+          await applicantQuestions.answerNumberQuestion('5')
+          await applicantQuestions.clickContinue()
+        })
+
+        await test.step('verify question is marked eligible', async () => {
```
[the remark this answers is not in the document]
greg-hyde: done

browser-test/src/applicant/navigation/application_navigation_with_eligibility_conditions.test.ts
```diff
@@ -452,5 +369,106 @@ test.describe('Applicant navigation flow', () => {
...
+          await validateScreenshot(
+            page,
+            'eligible-home-page-program-tagnorthstar',
+            /* fullPage= */ true,
+            /* mobileScreenshot= */ true,
+          )
+        })
+      })
+
+      test('Shows may be eligible toast with an eligible answer', async ({
```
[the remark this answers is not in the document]
greg-hyde: done

server/app/views/components/ToastFragment.html
greg-hyde: the convention I've been using is just that Fragment.html files can contain any number of fragments. I don't really want to establish a standard where we have to rename a file when adding a second fragment. I'd be open to naming all fragment files Fragments.html but don't feel strongly about it

melanie-exygy: I think changing the name to the plural would be helpful as we switch over to Thymeleaf.  Maybe it's just me, but I wasn't expecting there to be multiple fragments in the file, so the naming would be helpful.

server/app/views/components/ToastFragment.html
greg-hyde: yes and no. there's overlap, but for instance, the block edit page has the "eligible" toast but not the "not eligible" one, and the program index page shows the "not eligible" toast but not the "eligible" toast. Thus far I've made them reusable by having the basic success/warning/alert/error fragments. I'm thinking about simplifying them some since some of the arguments don't seem to vary among all calls, but I was thinking of waiting until I implemented all toasts across the app before I do that.

server/app/views/components/ToastFragment.html
greg-hyde: that's actually handled by toast.ts. It grabs the content from these toasts and renders it to the user with role and aria-live specified.

server/app/views/components/ToastFragment.html
greg-hyde: Ok I filed https://github.com/civiform/civiform/issues/7739

server/app/views/components/ToastFragment.html
greg-hyde: Unfortunately, we will need some amount of scripting for toasts to handle features like condOnStorageKey and ignore lists and dismissing after a certain period of time. I don't particularly like relying on toast.ts mainly because it applies old non-USWDS styles instead of making things look like a USWDS alert, but I have an issue filed for that work. https://github.com/civiform/civiform/issues/7747

browser-test/src/applicant/navigation/application_navigation_flow.test.ts
```diff
@@ -106,6 +109,34 @@ test.describe('Applicant navigation flow', () => {
             /* mobileScreenshot= */ true,
           )
         })
+
+        test('shows error with incomplete submission', async ({
```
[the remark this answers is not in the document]
greg-hyde: done

server/app/views/components/ToastFragment.html
greg-hyde: The way you describe it basically just moves summaryToasts and blockEditToasts and other future ones out into the template files. I'd prefer to keep them out of the template just to keep the template files from blowing up, but I could be convinced.

My basic plan is to do the refactoring I think you have in mind once I have added toasts for all pages. Basically, I want to verify something like "ah all alert toasts we have use randomly generated ids and the same settings, so I can simplify those templates so it just looks like <div th:replace="~{this :: alert(${bannerMessage})}"></div> " and those templates get a lot simpler, but I want to triple check that there isn't one weird edge case where one alert is slightly different first. Does that make sense?

melanie-exygy: Makes sense.  The refactoring can wait.

melanie-exygy (approved):

PR #7746, opened 2024-06-12
[Status Migration] - Adding UpdateStatusDefinitions function in ApplicationsStatusRepository 

### LINKED ISSUE #7040: Ensure application status can always be changed to any available status
**Describe the bug**
1. If a submission does not have a status (eg, no default status was set), the submission cannot be changed to a different status. 
2. If a submission has Status A, and Status B was created after the submission was submitted, the submission's status cannot be changed to Status B. 

**Expected behavior**
1. Admins are able to add status to a submission that currently does not have a status. The same status dropdown shows for submissions without a status as for submissions with a status.
2. Admins are able to change a submission to any current status, including statuses created after the submission.

**Additional context**
This is needed as part of our [bulk status updates](https://docs.google.com/document/d/19jETzFmpIHEGfHfEbzqnPwSGmEB-WD-mauhXl5b9CDE/edit) work.

swatkat1:
### Description

Adding UpdateStatusDefinitions function in ApplicationsStatusRepository.

#### General

### Issue(s) this completes

Fixes #7040

### DIFF
--- a/server/app/repository/ApplicationStatusesRepository.java
+++ b/server/app/repository/ApplicationStatusesRepository.java
@@ -52,16 +52,22 @@ public final class ApplicationStatusesRepository {
         .map(ApplicationStatusesModel::getStatusDefinitions)
         .collect(Collectors.toList());
   }
+
   public void updateStatusDefinitions(String programName, StatusDefinitions statusDefinitions) {
-    //archive the previous active one
-    database.sqlQuery("UPDATE application_statuses SET status_definitions_lifecycle_stage=:obsolete_stage WHERE program_name =:programName AND status_definitions_lifecycle_stage=:active_stage")
-      .setLabel("ApplicationStatusRepository.updateStatus")
-      .setParameter("obsolete_stage",StatusDefinitionsLifecycleStage.OBSOLETE)
-      .setParameter("programName", programName)
-      .setParameter("active_stage",StatusDefinitionsLifecycleStage.ACTIVE);
-    //create new status
-    ApplicationStatusesModel newStatusDefinition = new ApplicationStatusesModel(programName,statusDefinitions,StatusDefinitionsLifecycleStage.ACTIVE);
+    // archive the previous active one
+    database
+        .sqlQuery(
+            "UPDATE application_statuses SET status_definitions_lifecycle_stage=:obsolete_stage"
+                + " WHERE program_name =:programName AND"
+                + " status_definitions_lifecycle_stage=:active_stage")
+        .setLabel("ApplicationStatusRepository.updateStatus")
+        .setParameter("obsolete_stage", StatusDefinitionsLifecycleStage.OBSOLETE)
+        .setParameter("programName", programName)
+        .setParameter("active_stage", StatusDefinitionsLifecycleStage.ACTIVE);
+    // create new status
+    ApplicationStatusesModel newStatusDefinition =
+        new ApplicationStatusesModel(
+            programName, statusDefinitions, StatusDefinitionsLifecycleStage.ACTIVE);
     newStatusDefinition.save();
   }
-
 }
--- a/server/app/repository/ApplicationStatusesRepository.java
+++ b/server/app/repository/ApplicationStatusesRepository.java
@@ -52,4 +52,16 @@ public final class ApplicationStatusesRepository {
         .map(ApplicationStatusesModel::getStatusDefinitions)
         .collect(Collectors.toList());
   }
+  public void updateStatusDefinitions(String programName, StatusDefinitions statusDefinitions) {
+    //archive the previous active one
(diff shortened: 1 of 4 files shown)


server/app/repository/ApplicationStatusesRepository.java
```diff
@@ -52,4 +52,22 @@ public List<StatusDefinitions> lookupListOfObsoleteStatusDefinitions(String prog
...
+        .eq("program_name", programName)
+        .and()
+        .eq("status_definitions_lifecycle_stage", StatusDefinitionsLifecycleStage.ACTIVE)
+        .update();
+
+    // create new status
+    ApplicationStatusesModel newStatusDefinition =
+        new ApplicationStatusesModel(
+            programName, statusDefinitions, StatusDefinitionsLifecycleStage.ACTIVE);
+    newStatusDefinition.save();
```
avaleske, line 71: **issue**: Can we wrap this in a transaction so it happens atomically. otherwise a database read between these two writes could end up with unexpected data, like there not being an ACTIVE `StatusDefinitions` for a program.

avaleske (changes requested):

server/app/repository/ApplicationStatusesRepository.java
avaleske: https://ebean.io/docs/transactions/

swatkat1: Great idea. Done!

server/app/repository/ApplicationStatusesRepository.java
```diff
@@ -52,4 +54,26 @@ public List<StatusDefinitions> lookupListOfObsoleteStatusDefinitions(String prog
         .map(ApplicationStatusesModel::getStatusDefinitions)
         .collect(Collectors.toList());
   }
+
+  public void updateStatusDefinitions(String programName, StatusDefinitions statusDefinitions) {
```
avaleske, line 58: **suggestion, optional, not blocking**: Depending on how we're calling this, a single `createOrUpdateStatusDefinitions()` method could make the callsites cleaner.

server/test/repository/ApplicationStatusesRepositoryTest.java
```diff
@@ -85,4 +89,31 @@ public void canQueryForOboseleteApplicationStatuses() {
...
+
+    StatusDefinitions statusDefinitionsResult2 = repo.lookupActiveStatusDefinitions(programName);
+    assertStatusDefinitions(statusDefinitionsResult2, "Reapply");
+  }
+
+  public void assertStatusDefinitions(
+      StatusDefinitions statusDefinitionsResult, String statusText) {
+    assertThat(statusDefinitionsResult.getStatuses().size()).isEqualTo(1);
+    assertThat(statusDefinitionsResult.getStatuses().get(0).statusText()).isEqualTo(statusText);
+  }
```
avaleske, line 118: **suggestion, optional, not-blocking**: Tests don't need to follow DRY principles as much as regular code, since the primary goal is readability. imo leaving these asserts in each test, even if the code is repetitive, is better cause it makes it easier to see what the test is doing.

avaleske (approved):
a few thoughts but overall lgtm!

server/test/repository/ApplicationStatusesRepositoryTest.java
```diff
@@ -85,4 +89,31 @@ public void canQueryForOboseleteApplicationStatuses() {
...
+    ProgramModel program =
+        ProgramBuilder.newActiveProgram("Updateprogram" + uniqueProgramId, "description").build();
+    String programName = program.getProgramDefinition().adminName();
+    StatusDefinitions statusDefinitions = new StatusDefinitions(ImmutableList.of(APPROVED_STATUS));
+    ApplicationStatusesModel applicationStatusesModel =
+        new ApplicationStatusesModel(
+            programName, statusDefinitions, StatusDefinitionsLifecycleStage.ACTIVE);
+    applicationStatusesModel.save();
+    StatusDefinitions statusDefinitionsResult = repo.lookupActiveStatusDefinitions(programName);
+    assertStatusDefinitions(statusDefinitionsResult, "Approved");
```
avaleske, line 105: **nit, non-blocking**: I might add a line between lines 103 and 104 to make it clear this is a "pre-assert" step that's a little distinct from the "arrange" step before it.

server/test/repository/ApplicationStatusesRepositoryTest.java
swatkat1: I agree with you! Let me move it back into each tests

server/app/repository/ApplicationStatusesRepository.java
swatkat1: Create has to be different as it would have the update() query.

avaleske: Yeah, inside the method it would have check if the statuses existed for the program and, if they did, call update and if they didn't call create.

_very_ optional thought though, I don't know what your callsites for this look like so I don't know if it's easier to hide that complexity in the Repository or let it be at the callsite.

Issue #7752, opened 2024-06-12
P0 - API/JSON export Broken After Setting Question As PAI


gwendolyngoetz:
Yesterday we enabled the applicant name question as a Universal name question and as PAI. That seems to have broken exporting JSON and the API.

Below stacktrace is me reproing it on my local machine. 

- Create a program
- Add a name question that is not universal and not pai
- publish
- submit an application
- Export as JSON was working
- Now set the question as universal and pai
- re-publish
- submit another application
- Export as JSON, get the exception.

Disabling the universal and pai don't appear to put it back into a working state.

```
play.api.PlayException: Execution exception[[RuntimeException: java.lang.NullPointerException: Cannot invoke "models.ApplicantModel.getFirstName()" because "this.applicant" is null]]
        at play.api.http.HttpErrorHandlerExceptions$.$anonfun$convertToPlayException$3(HttpErrorHandler.scala:407)
        at scala.Option.getOrElse(Option.scala:201)
        at play.api.http.HttpErrorHandlerExceptions$.convertToPlayException(HttpErrorHandler.scala:407)
        at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:390)
        at play.api.http.HttpErrorHandlerExceptions.throwableToUsefulException(HttpErrorHandler.scala)
        at play.http.DefaultHttpErrorHandler.throwableToUsefulException(DefaultHttpErrorHandler.java:245)
        at play.http.DefaultHttpErrorHandler.onServerError(DefaultHttpErrorHandler.java:181)
        at controllers.ErrorHandler.onServerError(ErrorHandler.java:106)
        at play.core.j.JavaHttpErrorHandlerAdapter.onServerError(JavaHttpErrorHandlerAdapter.scala:33)
        at play.core.server.AkkaHttpServer$$anonfun$invokeAction$1$1.applyOrElse(AkkaHttpServer.scala:482)
        at play.core.server.AkkaHttpServer$$anonfun$invokeAction$1$1.applyOrElse(AkkaHttpServer.scala:474)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:490)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
[...]

PR #7753, opened 2024-06-12
Fix JSON export with PAI

nb1701:
When getting the ReadOnlyApplicantProgramService inside the JsonExporterService, we are passing it an ApplicantData object without the ApplicantModel inside it. We're doing this intentionally so that any questions not in the application's program version are exported as "unanswered". This caused the lookup of data for a PAI name question (and probably any other PAI question) to fail since it relies on that ApplicantModel object.

This modifies the getters in ApplicantData to handle when the ApplicantModel is null.

In the future, we should probably figure out how to untangle this a bit to standardize how the ApplicantData object is created and used, but this fixes the immediate issue.

#### General

### Issue(s) this completes

https://github.com/civiform/civiform/issues/7752

### DIFF
--- a/server/app/services/applicant/ApplicantData.java
+++ b/server/app/services/applicant/ApplicantData.java
@@ -83,8 +83,9 @@ public class ApplicantData extends CfJsonDocumentContext {
    * @return Formatted name of the applicant
    */
   public Optional<String> getApplicantName() {
-    Optional<String> firstName = applicant == null ? Optional.empty() : applicant.getFirstName();
-    Optional<String> lastName = applicant == null ? Optional.empty() : applicant.getLastName();
+    Optional<String> firstName =
+        Optional.ofNullable(applicant).flatMap(ApplicantModel::getFirstName);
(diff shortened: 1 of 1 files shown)


server/app/services/applicant/ApplicantData.java
```diff
@@ -133,25 +133,37 @@ public Optional<String> getApplicantNameAtWellKnownPath() {
   // that we can modify both the Primary Applicant Info columns and
   // Well Known Paths at the same time.
   public Optional<String> getApplicantFirstName() {
-    return applicant.getFirstName().or(() -> readString(WellKnownPaths.APPLICANT_FIRST_NAME));
+    return applicant == null || applicant.getFirstName().isEmpty()
+        ? readString(WellKnownPaths.APPLICANT_FIRST_NAME)
```
rockycodes, line 137: With this logic, will we change `readString(WellKnownPaths.APPLICANT_FIRST_NAME)` to `Optional.empty()` when well known paths go away?

gwendolyngoetz (approved):

PR #7817, opened 2024-06-17
Use true/false for flag UI values

### LINKED ISSUE #7656: Change options for feature flags on the Settings page to "True" or "False"

The CF engineers have noticed that it is redundant and possibly confusing when feature flag names end in "ENABLED" and then also show as enabled/disabled on the Settings page:

image

We'd like to change the Settings UI to show "True" or "False" rather than "Enabled" or "Disabled".  

#### Done when:

The feature flag settings UI shows "True" or "False" as options.

jeffpw-goog:
### Description

Change setting values for boolean settings from "Enabled" and "Disabled" to "True" and "False"

## Release notes

On the Admin Settings page, the options for feature flags are now displayed as "True" and "False" instead of "Enabled" and "Disabled", to more clearly reflect the underlying true/false values and remove redundancy when "enabled" is in the feature flag name.

#### General

### Issue(s) this completes

Fixes #7656


server/app/views/admin/settings/AdminSettingsIndexView.java
```diff
@@ -297,13 +297,13 @@ private static DivTag renderEnumInput(
 
   private static DivTag renderBoolInput(
       SettingDescription settingDescription, Optional<String> value) {
-    boolean isEnabled = value.map("TRUE"::equals).orElse(false);
+    boolean boolValue = value.map("TRUE"::equals).orElse(false);
```
melanie-exygy, line 300: nit:  For the variable name, I actually prefer `isEnabled` or maybe `isOn` because otherwise I find it unintuitive to have `!boolValue`.  Just my preference, though.

melanie-exygy (approved):

server/app/views/admin/settings/AdminSettingsIndexView.java
jeffpw-goog: Yeah I didn't like any of the names I could think of. How about isTrue... it's a little redundant but logically impeccable. 😂

PR #7828, opened 2024-06-18
Seed the database with default program categories in pre-prod environments

### LINKED ISSUE #7693: [Program Filtering] Create dev task to seed database with program categories
We need to pre-populate the new Categories database table with the following list of pre-defined categories: [Translated categories]( https://drive.google.com/file/d/1SJTvHedAae4pTLOqNY7SbhxDO_va4Cru/view?usp=drive_link).  We'll start by adding a task in the pre-production environments only.  

#### Additional Context
To do this, we'll need to add category seeding to the `DevDatabaseSeedTask.java`.  See the [TDD](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit#heading=h.8ymg6avzxm8o) for more implementation details.

#### Done when:
- [ ] Pre-defined categories and their translations are loaded into the database through a button click in the Dev Tools (for dev environment).

melanie-exygy:
### Description

This gives pre-prod admin users the ability to seed the default program categories from the Dev Tools.  These categories will be the same across all instances.  Categories are seeded along with questions and programs when the user clicks "Seed sample programs".  Categories are seeded with their translations, which are stored within the project here:  "conf/i18n/categories/".

1. Go to the Dev Tools
2. Click "Clear entire database" so that you don't have any conflicts with existing programs
3. Click "Seed sample programs and categories"
4. Verify that the Categories table in the db is populated with 10 categories and their translations.

### Issue(s) this completes

Fixes #7693


server/app/controllers/dev/seeding/CategoryTranslationFileParser.java
```diff
@@ -0,0 +1,149 @@
...
+            VI_CATEGORY_TRANSLATIONS.get(key),
+            Lang.forCode("en-US").toLocale(),
+            EN_CATEGORY_TRANSLATIONS.get(key),
+            Lang.forCode("es-US").toLocale(),
+            ES_CATEGORY_TRANSLATIONS.get(key),
+            Lang.forCode("zh-TW").toLocale(),
+            ZH_CATEGORY_TRANSLATIONS.get(key)));
+  }
+
+  private static ImmutableMap<String, String> parseFileForCategoryTranslations(String fileName) {
```
avaleske, line 119: **suggestion**: Instead of parsing the file by hand, I think we should use the `Properties` java util, like we do it `MessagesTest`:
https://github.com/civiform/civiform/blob/39cdc4699c6d0346c1466f04e55979c57486c40b/server/test/conf/MessagesTest.java#L157-L172

This gives us back a map of keys-value pairs that will be easier to work with in the rest of this file, and automatically ignores comment lines.

avaleske: I also wonder if it's easier to sort these values into category-specific maps at this point, as we read in each file, instead of pivoting these lists from language to category later when we're retrieving the category.

avaleske (changes requested):

server/conf/i18n/categories/am_program_categories.txt
```diff
@@ -0,0 +1,29 @@
+# የፕሮግራሞች ዝርዝርን ከልጅ እንክብካቤ ጋር ወደሚዛመዱት አውርዶ ለማጣራት ጥቅም ላይ የሚውል መለያ።
```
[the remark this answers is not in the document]
melanie-exygy: Good ideas.  Done!

server/app/controllers/dev/seeding/CategoryTranslationFileParser.java
melanie-exygy: I've replaced the parsing with the Java `Properties` util.  Thanks for that suggestion!  

I haven't had a chance to refactor so that we're sorting into category-specific maps here, but will do that when I get back.

melanie-exygy: I refactored so that we sort into category-specific maps rather than first sorting into language-specific maps.  This meant hard-coding the categories because I needed to create category map variables before parsing all the files at once, but I think it's okay to do that since our pre-defined categories are fixed for now and once we make them more flexible, we likely won't be using the database seed tasks to create them anymore.

avaleske (approved):

PR #7829, opened 2024-06-19
Add a baseline path to gitleaks execution to suppress old findings

yotommy:
### Description

Add a baseline path to gitleaks execution to suppress old findings.

This will prevent failures from each scheduled run ([example](https://github.com/civiform/civiform/actions/runs/9579229432)).

The baseline file was created via:
```
$ gitleaks detect -r .github/workflows/gitleaks-baseline.json
```

#### General


.github/workflows/scan_for_secrets.yaml
```diff
@@ -18,3 +18,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITLEAKS_LICENSE: ${{ secrets.GITLEAKS_LICENSE}}
+        with:
+          baseline-path: .github/workflows/gitleaks-baseline.json
```
gwendolyngoetz, line 22: Can you move this up one directory into the `.github`? It's become somewhat the catch all for workflow related files and it makes it easier to find not being in with all the workflow files.

gwendolyngoetz: Is there a way to whitelist specific parts of files? The config.ts file with the test jwt for example would be nice to ignore. The dev postgres password "example". Etc.

gwendolyngoetz (changes requested):

.github/workflows/scan_for_secrets.yaml
yotommy: Moved the file as suggested.

You can exclude parts of a file via [`gitleaks:allow`](https://github.com/gitleaks/gitleaks?tab=readme-ov-file#gitleaksallow) comments.

gwendolyngoetz (approved):

gwendolyngoetz:
Haven't seen any issues so far. Will add this later if needed. Closing as it is outdated.

Issue #7846, opened 2024-06-21
Multi-select validation error is inconsistent with others


jeffpw-goog:
**Describe the bug**

Several question types allow admins to set bounds for the answer:
* ID - min/max length
* Checkbox - min/max selections
* Number - min/max value
* Text - min/max length
* Enumerator - min/max values (pending in #7136)

For all of these, the error messages if you enter an invalid value are something like:
* Must contain at least {0} characters
* Must contain at most {0} characters

However, for the checkbox question, the upper bound error message is "Please select fewer than {0}", where {0} is the upper bound + 1. This is weird technically, and could be confusing to applicants who have seen the other error message and might mis-interpret this one.

**To Reproduce**

1. Create a program with a checkbox question with 2 choices and a "maximum choices allowed" of 1
2. Apply to the program, select both checkbox options, and click next

**Expected behavior**
The error message should say "Please select at most 1" rather than "Please select fewer than 2".

**Screenshots**
<img width="279" alt="image" src="https://github.com/civiform/civiform/assets/165937506/b0eabe3d-1cd7-4cae-9765-c09908e6c3ae">

jeffpw-goog:
I am also going to use this bug as a test case for changing localized strings, to validate the documentation at https://github.com/civiform/civiform/wiki/Internationalization-(i18n)#changing-existing-strings

jeffpw-goog:
This is blocked on translations for "validation.tooManySelections.v2".

jeffpw-goog:
Translations are ready: #7918

PR #7847, opened 2024-06-21
Add new string for multi-select validation error

jeffpw-goog:
### Description

Add a new string for the error when you select too many choices in a multi-select question. The current error message says "select fewer than X" which is inconsistent with all the other similar error messages that say "select at most X". This is the first step for fixing #7846.

#### General

### DIFF
--- a/server/conf/i18n/messages
+++ b/server/conf/i18n/messages
@@ -503,6 +503,8 @@ validation.numberRequired=Must contain only numbers.
 
 validation.tooFewSelections=Please select at least {0}.
 validation.tooManySelections=Please select fewer than {0}.
+# Error message that indicates the applicant has selected too many choices on the multi-select question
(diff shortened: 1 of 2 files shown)


server/conf/i18n/messages
```diff
@@ -503,6 +503,7 @@ validation.numberRequired=Must contain only numbers.
 
 validation.tooFewSelections=Please select at least {0}.
 validation.tooManySelections=Please select fewer than {0}.
+validation.tooManySelections.v2=Please select at most {0}.
```
dkatzz, line 506: **suggestion:** Could you include a context string?

jeffpw-goog: Thanks done!

dkatzz (approved):

Issue #7490, opened 2024-05-15
[North Star Applicant UI] Update question previews in the admin page


greg-hyde:
In the admin page, we show a preview of how questions will look to applicants. We need to switch those over to render the thymeleaf implementation so it matches what the applicant will actually see

ssandbekkhaug:
<img width="1326" alt="question_preview" src="https://github.com/civiform/civiform/assets/23390490/c432ce96-7ab8-4220-9fe5-e5ee7878465d">

ssandbekkhaug:
Example from Greg to inspire my approach: https://github.com/civiform/civiform/pull/6531

ssandbekkhaug:
To enable: http://localhost:9000/dev/feature/NORTH_STAR_APPLICANT_UI/enable
To disable: http://localhost:9000/dev/feature/NORTH_STAR_APPLICANT_UI/disable

PR #7861, opened 2024-06-24
Fix converting String to QuestionType for irregular enum names

ssandbekkhaug:
### Description

Fix converting String to QuestionType for irregular enum names. This change supports question previews (#7490).

This PR is a part of #7835 being split off for ease of review. The code will be tested indirectly by browser tests, but feel free to request unit tests specific to QuestionType.

## Release notes

n/a

#### General

n/a
### Issue(s) this completes

None

### DIFF
--- a/server/app/services/question/types/QuestionType.java
+++ b/server/app/services/question/types/QuestionType.java
@@ -1,7 +1,5 @@
 package services.question.types;
 
-import java.util.Locale;
-import java.util.Map;
 import services.applicant.question.AddressQuestion;
 import services.applicant.question.CurrencyQuestion;
 import services.applicant.question.DateQuestion;
@@ -41,12 +39,6 @@ public enum QuestionType {
 
   private final String label;
   private final Class<? extends Question> supportedQuestion;
-  private static final Map<String, QuestionType> irregularMappings =
-      Map.of(
-          "FILE UPLOAD", FILEUPLOAD,
-          "RADIO BUTTON", RADIO_BUTTON,
-          "STATIC TEXT", STATIC,
-          "PHONE NUMBER", PHONE);
 
   QuestionType(String label, Class<? extends Question> supportedQuestion) {
     this.label = label;
@@ -63,16 +55,12 @@ public enum QuestionType {
   }
 
   public static QuestionType of(String name) throws InvalidQuestionTypeException {
-    String upperName = name.toUpperCase(Locale.ROOT);
-    try {
-      if (irregularMappings.containsKey(upperName)) {
-        return irregularMappings.get(upperName);
+    for (QuestionType type : QuestionType.values()) {
+      if (type.getLabel().equalsIgnoreCase(name)) {
+        return type;
       }
-
-      return valueOf(upperName);
-    } catch (IllegalArgumentException e) {
-      throw new InvalidQuestionTypeException(upperName);
     }
+    throw new InvalidQuestionTypeException(name);
   }
 
   public String getLabel() {
--- a/server/app/services/question/types/QuestionType.java
+++ b/server/app/services/question/types/QuestionType.java
@@ -56,13 +56,6 @@ public enum QuestionType {
   }
 
   public static QuestionType of(String name) throws InvalidQuestionTypeException {
-    // Match label, e.g. "Phone Number" -> PHONE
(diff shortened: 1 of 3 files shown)


server/app/services/question/types/QuestionType.java
```diff
@@ -40,6 +41,12 @@ public enum QuestionType {
 
   private final String label;
   private final Class<? extends Question> supportedQuestion;
+  private static final Map<String, QuestionType> irregularMappings =
```
greg-hyde, line 44: Can you explain why this is necessary?

gwendolyngoetz, line 44: (blocking) Please use an ImmutableMap

gwendolyngoetz: I have the same question. This is a pretty core class and I'm trying to understand what doesn't work with the new user case.

server/app/services/question/types/QuestionType.java
```diff
@@ -58,6 +65,10 @@ public boolean isMultiOptionType() {
   public static QuestionType of(String name) throws InvalidQuestionTypeException {
     String upperName = name.toUpperCase(Locale.ROOT);
     try {
+      if (irregularMappings.containsKey(upperName)) {
```
gwendolyngoetz, line 68: (blocking) Please add some tests to cover these changes.

gwendolyngoetz (changes requested):

server/app/services/question/types/QuestionType.java
ssandbekkhaug: Good question. The motivation is to render a Thymeleaf fragment inside J2HTML.

Short answer: PHONE.getLabel() -> "Phone Number" -> of("Phone Number") -> not found

Long answer can be traced in the larger PR: https://github.com/civiform/civiform/pull/7835/files

1. Thymeleaf is loaded by HTMX via QuestionEditView.
2. `htmx-get` calls `controllers.admin.routes.NorthStarQuestionController.sampleQuestion(...)`
3. The method signature is `public Result sampleQuestion(Request request, String questionType)`, so we need a String to represent the question type
  * An enum cannot be used because htmx-get doesn't support enums
  * getLabel() is called to convert the enum to a String
4. NorthStarQuestionController recieves a String representing the question type.
5. The String needs to be converted back to an enum.
6. the `of(...)` method simply converts the String to uppercase. This misses conversions for some enums.

greg-hyde: Ok that makes sense. I think I'd prefer just rewriting that method to something like

```
public static QuestionType of(String name) throws InvalidQuestionTypeException {
    for (QuestionType type : QuestionType.values()) {
      if (type.getLabel().equals(name)) {
        return type;
      }
    }
    throw new InvalidQuestionTypeException(name);
  }
```

server/app/services/question/types/QuestionType.java
```diff
@@ -56,6 +56,14 @@ public boolean isMultiOptionType() {
   }
 
   public static QuestionType of(String name) throws InvalidQuestionTypeException {
+    // Match label, e.g. "Phone Number" -> PHONE
+    for (QuestionType type : QuestionType.values()) {
```
greg-hyde, line 60: this function still feels weird to me because it's trying two different things, and why things are different isn't really documented. Maybe you could just factor this portion out to a fromLabel function and leave the existing function as is.

server/test/support/TestQuestionBank.java
```diff
@@ -560,7 +560,8 @@ public QuestionModel maybeSave(
         question.loadQuestionDefinition();
       } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
         throw new IllegalArgumentException(
-            "Questions in the TestQuestionBank better be supported QuestionTypes.");
+            "Questions in the TestQuestionBank better be supported QuestionTypes. "
```
gwendolyngoetz, line 563: (not-serious) This message isn't threatening enough. Can you add a knife emoji?

gwendolyngoetz (approved):

server/test/support/TestQuestionBank.java
ssandbekkhaug: 😂

PR #7866, opened 2024-06-24
Fix validation error for checkbox question to be consistent with other question types

### LINKED ISSUE #7846: Multi-select validation error is inconsistent with others

jeffpw-goog:
### Description

Update the validation error for too many selections on a multi-select question to be consistent with the messaging for all other questions with this type of constraint. When the limit is N, it now says "select at most N" instead of "select fewer than N+1".

* Use the new message validation.tooManySelections.v2 which is now translated
* Remove the old message validation.tooManySelections
* Update the logic to plug N into the string template instead of N+1
* Remove a test case from CheckboxQuestionRendererTest that was validating the error message which is neither generated nor rendered by CheckboxQuestionRenderer.

#### General

#### User visible changes

### Issue(s) this completes

Fixes #7846


swatkat1 (approved):
Much better messaging! Thank you!

PR #7878, opened 2024-06-25
Remove screenshot from admin_programs.ts helper

jeffpw-goog:
### Description

Remove screenshot validation for trusted intermediary program visibility from addProgram in admin_programs.ts helper because helpers should not be making assertions.

In this case, the helper already checks the TI visibility checkbox and the specific TI checkbox, so it will fail if the screen is not as expected and the screenshot is unnecessary.

#### General


swatkat1 (approved):
Nice!!

Issue #7895, opened 2024-06-26
[North Star Applicant UI] Header is too short on mobile


greg-hyde:
Default USWDS styles make the header 48 pixels high, but that's smaller than the civiform test logo and the city of seattle logo, making the logo overflow, so we should update that default height

![6UHvVRCXZUdsxM7](https://github.com/civiform/civiform/assets/156004543/cf5721f4-9af8-41ef-a2f5-292bc81227d6)

sijiayam:
We should resize the logo to fit into the header.

ssandbekkhaug:
Latest mocks from Sijia: https://www.figma.com/design/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?node-id=1832-691&t=wy3cHG9jLQyOHV2y-0

ssandbekkhaug:
Fixed by #7961

sijiayam:
Testing is blocked by #9035

sijiayam:
I am still seeing this issue on staging on screen sizes <1024px. See image attached below. I am going to reopen issue until it is resolved. 
![Image](https://github.com/user-attachments/assets/d4355798-7b5a-4ee9-a32b-48d0d748368c)

ssandbekkhaug:
Oh, that's because PR#7961 did not update the PNG used by staging: https://github.com/civiform/civiform-staging-assets/blob/main/logos/civiform-staging.png

ssandbekkhaug:
https://github.com/civiform/civiform-staging-assets/pull/4

ssandbekkhaug:
Open questions:

1. What should the "correct" dimensions of the logo be?
2. If the government provides a differently-sized logo, what should happen? (Too large or too small, very wide, very tall, etc)

sijiayam:
- Logo should have a height of 64px and max-height of 64px
- If logo height exceeds 64px, logo should be resized to fit
- If logo height is under 64px, logo should be icreased to fit

shreyachatterjee00:
@sijiayam is this still an issue? I'm seeing this on the same px size you tried - the logo isn't cutoff at the top

<img width="1012" alt="Image" src="https://github.com/user-attachments/assets/91d82d93-c9fb-4afa-87aa-bd53d836d384" />

sijiayam:
I'm still seeing it at specific breakpoints. Maybe its a tablet screen issue

[...]

Issue #7809, opened 2024-06-17
[North Star Applicant UI] Investigate why changes to the page make a 1 pixel change to questions for screenshot tests


greg-hyde:
A change to somewhere else in the page often triggers a 1 px height difference in a lot of screenshots that haven't actually changed. It looks like it's a fractional number of pixels so we need to see what we can do to prevent this churn

PR #7896, opened 2024-06-26
Add goverment website banner

### LINKED ISSUE #7772: [North Star Applicant UI] Add official government website banner
Mocks show a small banner with "An official website of the United states government. Here's how you know" that we need to add

https://www.figma.com/design/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?m=dev&node-id=2326-4257&t=4ZdRnL3ui9KmGZbW-1

greg-hyde:
### Description

Adds the banner to the top of the page explaining that it is a government website. This did make me notice that the header is too short on mobile so I filed https://github.com/civiform/civiform/issues/7895

6UHvVRCXZUdsxM7

6LEyNkv3yskeAya

#### General

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #7772


greg-hyde:
* **#7896** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ssandbekkhaug (commented):
Just a question: is it possible to take screenshots that intentionally exclude the banner (and header/footer?)

I think screenshot tests should focus on the content being tested. Including the banner in every test means that any minor change to the banner will result in us having to regenerate all the screenshots.

For the whole page ones, I think it's important to include the banner so it's not missed on a page. In my mind, that's exactly the type of the thing the whole page tests are testing. I know it causes churn right now, but I think it's worth it. As for the question types, I've narrowed the scope of those, but they annoyingly still change by 1px in height with pretty much every change (see https://github.com/civiform/civiform/issues/7809)

greg-hyde:
> Just a question: is it possible to take screenshots that intentionally exclude the banner (and header/footer?)
> 
> I think screenshot tests should focus on the content being tested. Including the banner in every test means that any minor change to the banner will result in us having to regenerate all the screenshots.

For the whole page tests, I think it's important to include them. In my mind, stuff like that is exactly what full page screenshots are for. I've narrowed the question type screenshots, but they annoyingly vary by 1 px in height (see https://github.com/civiform/civiform/issues/7809)

gwendolyngoetz:
@greg-hyde We could look into adjusting the [maxDiffPixelRatio](https://github.com/civiform/civiform/blob/b29f45ccef248d658bb7462f45ba7c60099243e7/browser-test/playwright.config.ts#L19-L25) if it ever becomes a problem. Right now it's 0 so it has to be a perfect match.

greg-hyde:
> @greg-hyde We could look into adjusting the [maxDiffPixelRatio](https://github.com/civiform/civiform/blob/b29f45ccef248d658bb7462f45ba7c60099243e7/browser-test/playwright.config.ts#L19-L25) if it ever becomes a problem. Right now it's 0 so it has to be a perfect match.

I looked into that a bit. At least at Google, it's generally not recommended because then you'll miss a small real change. But if we find the churn more annoying than potentially missing something, we can do it

server/app/views/applicant/NavigationFragment.html
```diff
@@ -36,6 +37,96 @@
...
+              ></th:block>
+            </p>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<span th:fragment="lockIcon" class="icon-lock"
```
dkatzz, line 108: **suggestion:** I'm not sure exactly where its defined, but can we use the icon fragment that we use in some other places? https://github.com/civiform/civiform/blob/87629521179c78a9f29f6a42b43479327e569e75/server/app/views/applicant/ApplicantBaseFragment.html#L68

greg-hyde: so annoyingly we can't use the Icon fragment because we're passing html into the httpsSectionContent message, meaning we have to use th:utext, but then it means it won't process anything within the inner template..

I added some content to the wiki about this, but one of my biggest annoyances with thymeleaf is adding HTML within a message. It's possible but you can't have any variables within the HTML... We're also adding a couple custom classes so it appears inline with the banner.

I added a comment to the fragment here though

dkatzz: Thanks for clarifying!

dkatzz (approved):
Confirming that this works as expected when toggled as well (like below)? May be worth adding a test for that too if it doesn't exist.

Screenshot 2024-06-27 at 12 09 53 PM

greg-hyde:
> Confirming that this works as expected when toggled as well (like below)? May be worth adding a test for that too if it doesn't exist.
> 
> Screenshot 2024-06-27 at 12 09 53 PM

Added a test to header.test.ts :)

dkatzz:
> > Confirming that this works as expected when toggled as well (like below)? May be worth adding a test for that too if it doesn't exist.
> > Screenshot 2024-06-27 at 12 09 53 PM
> 
> Added a test to header.test.ts :)

Thank you!!

PR #7939, opened 2024-07-01
Adding new overload for creating alerts

### LINKED ISSUE #5541: Fast forward to the latest program version
**Describe the bug**
As an applicant, I can start a program application but not submit it. 

A CiviForm Admin can make changes to the program application (add eligibility conditions, translations, or additional questions) 

Those changes will not show when the applicant returns to complete their application. 

This causes problems for applicants and Program Admins because the correct version of the program application was not submitted. For an applicant, they may have been able to bypass newly added eligibility criteria or required questions. For Program Admins, they must reach out to the applicant to ask for additional information to complete the application.

**To Reproduce**

As a Program Admin:
1. Create a new program
2. Publish it

As a Resident:
1. Log in
2. Complete the program application but DO NOT submit it

As a Program Admin:
1. Publish changes to the existing program by adding eligibility conditions, new questions, new translations

As a Resident:
1. Log back in
2. See the draft application
3. See that the draft application does not include the recent changes to the eligibility conditions, new questions, new translations
4. Submit the draft application

As a Program Admin:
1. See that the submitted application does not include answers to the new questions and bypassed the eligibility conditions

**Expected behavior**
Expect that a returning applicant would have their draft application upgraded to the most recent version of the program. Already completed questions would display, and any new questions/eligiblity conditions/translations would also show.

**Screenshots**
_Need to add screenshots for the workflow_

gwendolyngoetz:
### Description

Adding an overload to the `makeAlert` methods that uses the new `AlertType` enum.

#### General

### Issue(s) this completes

Fixes #5541

### DIFF
--- a/server/app/views/ViewUtils.java
+++ b/server/app/views/ViewUtils.java
@@ -36,7 +36,6 @@ import j2html.tags.specialized.PTag;
 import j2html.tags.specialized.ScriptTag;
 import j2html.tags.specialized.SpanTag;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.Locale;
 import java.util.Optional;
 import javax.inject.Inject;
@@ -421,11 +420,8 @@ public final class ViewUtils {
           default -> "";
         };
 
-    // Add the alert type class to the classes list
-    String[] updatedClasses = Arrays.copyOf(classes, classes.length + 1);
-    updatedClasses[classes.length] = alertTypeStyle;
-
-    return makeAlert(text, hidden, title, updatedClasses);
+    return makeAlert(
+        text, hidden, title, Lists.asList(alertTypeStyle, classes).toArray(new String[0]));
   }
 
   /**
--- a/server/app/views/ViewUtils.java
+++ b/server/app/views/ViewUtils.java
@@ -36,10 +36,12 @@ import j2html.tags.specialized.PTag;
 import j2html.tags.specialized.ScriptTag;
 import j2html.tags.specialized.SpanTag;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.Optional;
 import javax.inject.Inject;
 import play.i18n.Messages;
+import services.AlertType;
 import services.DateConverter;
 import services.MessageKey;
 import services.question.types.QuestionDefinition;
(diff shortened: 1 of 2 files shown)


server/app/views/ViewUtils.java
```diff
@@ -396,6 +398,36 @@ public static DivTag makeAlert(
...
+   * https://designsystem.digital.gov/components/alert/
+   *
+   * @param text The text to include in the alert.
+   * @param hidden Whether or not to set the hidden property on the component.
+   * @param title An optional title to be included in the alert.
+   * @param alertType The type of {@link AlertType} alert to show
+   * @param classes One or more additional classes to apply to the USWDS Alert component.
+   * @return DivTag containing the alert.
+   */
+  public static DivTag makeAlert(
```
jeffpw-goog, line 413: Question/suggestion: is an alert type required for all alerts? If so I think it would be preferable to make the makeAlert method above private and refactor the existing usages to use this new one. It's confusing to have two ways to accomplish the same thing (i.e. pass the alert type in the class list or call this method).

Relatedly, is the makeAlertSlim method kind of redundant then too? Actually I see calls to makeAlertSlim that also pass BaseStyles.ALERT_INFO so maybe an alert can have more than one style?! Or is that a bug? 🤔 

If all the above is possible, then would it make sense to move the ALERT_ strings out of BaseStyles so that one is not tempted to use them extra-curricularly (outside of this method)?

server/app/views/ViewUtils.java
```diff
@@ -396,6 +398,36 @@ public static DivTag makeAlert(
...
+        switch (alertType) {
+          case INFO -> BaseStyles.ALERT_INFO;
+          case ERROR -> BaseStyles.ALERT_ERROR;
+          case SUCCESS -> BaseStyles.ALERT_SUCCESS;
+          case WARNING -> BaseStyles.ALERT_WARNING;
+          default -> "";
+        };
+
+    // Add the alert type class to the classes list
+    String[] updatedClasses = Arrays.copyOf(classes, classes.length + 1);
```
jeffpw-goog, line 425: Nitpick: The array manipulation via index feels kludgy in this day and age, but it's tough with primitive arrays. The method below does this instead which I think reads cleaner (ish) and is logically simpler:
Lists.asList(BaseStyles.ALERT_SLIM, classes).toArray(new String[0]));

server/app/views/ViewUtils.java
gwendolyngoetz: I thought about doing that, but I wasn't sure if there were future cases Melanie had in mind when she created it. I was going to revisit it down the road after she gets back from vacation.

server/app/views/ViewUtils.java
gwendolyngoetz: I somewhat disagree and `new String[0]` is pretty awful syntax. But I'll go with it for consistency.

jeffpw-goog: Yeah I was hoping there was a nicer way... I had actually thrown out my comment until I saw how it was done below.

jeffpw-goog (approved):

PR #7949, opened 2024-07-02
Center CiviForm staging logo

ssandbekkhaug:
### Description

Center the staging logo in the PNG. See server/app/assets/Images/civiform-staging.png.

Also updated conf so that dev and tests read the logo file path locally ([via NorthStarBaseView](https://github.com/civiform/civiform/blob/main/server/app/views/NorthStarBaseView.java#L69)) instead of calling out to the published version on GitHub.

## Release notes

n/a

#### General

# Issues this completes

Supports #7895. That issue needs a robust fix with CSS.


ssandbekkhaug:
* **#7949** <a href="https://app.graphite.dev/github/pr/civiform/civiform/7949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/conf/application.dev-browser-tests.conf
gwendolyngoetz, line 1: This and the test file both inherit settings from the dev file, so you can just set it in the one.

Ideally see if you can set the original definition to null so everything defaults to the local image and only override if the env var is set.

server/conf/application.conf
```diff
@@ -588,3 +588,6 @@ civiform_applicant_enabled_languages = ${?CIVIFORM_APPLICANT_ENABLED_LANGUAGES}
 # Typically, this value is 1.
 num_trusted_proxies = 1
 num_trusted_proxies = ${?NUM_TRUSTED_PROXIES}
+
+# Default to the local PNG logo
+civic_entity_small_logo_url = null
```
gwendolyngoetz, line 593: (blocking) Sorry, probably didn't give enough detail in my original message. Doing this here will override the environment variables and will always be null now for any deployment. You need to edit the `whitelabel.conf` file which is imported into this base config.

What I think you want to do is set `civic_entity_small_local_url = null` [here](https://github.com/civiform/civiform/blob/afba63d9344776465884f7992a47bf2988e0e6f1/server/conf/helper/whitelabel.conf#L6C32-L6C130).

You'll need to make sure of the following though:

1) Local/deployed (staging) sites are indeed pulling from the file now instead of the url to githubusercontent as showing in the above link.

2) Verify that the image is loaded if the `CIVIC_ENTITY_SMALL_LOCAL_URL` is set. You can simulate that locally by setting the value [here](https://github.com/civiform/civiform/blob/afba63d9344776465884f7992a47bf2988e0e6f1/docker-compose.yml#L125) and running `bin/run-dev` and make sure it's now pointing back to the github link and not localhost.

gwendolyngoetz (changes requested):

server/conf/application.conf
gwendolyngoetz: Alternatively just set `civic_entity_small_local_url = null` in `application.dev.conf` and dev, browser-tests, and unit tests only should pull from the localhost image. But you'll want to test it to be certain. You don't need to set it in the `application.dev-browser-tests.conf` or `application.test.conf` files since those inherit from `application.dev.conf`

ssandbekkhaug: Thanks! Editing whitelabel.conf works and the other way doesn't. I tested locally.

I haven't deployed before, so I'll go through the instructions tomorrow to deploy to a staging environment. https://github.com/civiform/civiform/wiki/Deploy-System

Is doing a full deployment to test necessary?

gwendolyngoetz: If it works testing locally we can just test it on staging after tomorrow's release. Just don't merge it until then that way we don't have to rush if we need to rollback.

gwendolyngoetz (approved):

PR #7958, opened 2024-07-02
Add constraints to min/max question settings

### LINKED ISSUE #7919: Add input validation for custom question settings
These question types all have similar min/max settings that the admin can set to limit applicant inputs:

- Number - min/max value
- Checkbox - min/max number of choices
- ID - min/max value length
- Text - min/max value length
- File upload - max files
- Enumerator - min/max entities

Validation of these setting inputs is limited:

- Checkbox (MultiSelectQuestionDefinition) has extensive server-side validation in https://github.com/civiform/civiform/blob/91bdeac8aa7e4ed226b3615ac0fbc7e7a0a9455e/server/app/services/question/types/QuestionDefinition.java#L279-L378
- Checkbox has frontend validation that the values are at least 0: https://github.com/civiform/civiform/blame/372132b7b2d212edec0117dccefea76aa75bbe86/server/app/views/admin/questions/QuestionConfig.java#L354-L362
- File upload has frontend validation that the max value is at least 1: https://github.com/civiform/civiform/blame/372132b7b2d212edec0117dccefea76aa75bbe86/server/app/views/admin/questions/QuestionConfig.java#L128

Each question should have both server-side and client-side validation as appropriate for the setting's meaning, to help admins provide valid constraints.

jeffpw-goog:
### Description

For each question type that has min/max settings (Enumerator, Id, MultiOption, Number, Text), add constraints and validation that the input values are reasonable:

* Validate that min <= max if both are set (server-side)
* Require that min >= 0 (client-side and server-side)
* Require that max >= 1 since a max of 0 would prevent answering the question, except for Number where it validates max >= 0 since 0 is a valid Number answer (client-side and server-side)

Also included:

* Create unit test classes for several QuestionDefinition subclasses that didn't have them
* Add unit tests for the cases above

## Release notes

When an admin sets min/max settings for a question type that supports them (Number, Checkbox, ID, Text, Enumerator), CiviForm will now validate that the min <= max (if both are set), min >= 0, and max >= 1, except for Number questions where it will validate that max >= 0.

#### General

### Issue(s) this completes

Fixes #7919

### DIFF
--- a/server/app/services/question/types/MultiOptionQuestionDefinition.java
+++ b/server/app/services/question/types/MultiOptionQuestionDefinition.java
@@ -157,8 +157,8 @@ public final class MultiOptionQuestionDefinition extends QuestionDefinition {
     }
 
     if (maxChoicesAllowed.isPresent()) {
-      if (maxChoicesAllowed.getAsInt() < 0) {
-        errors.add(CiviFormError.of("Maximum number of choices allowed cannot be negative"));
+      if (maxChoicesAllowed.getAsInt() < 1) {
+        errors.add(CiviFormError.of("Maximum number of choices allowed cannot be less than 1"));
       }
 
       if (maxChoicesAllowed.getAsInt() > numOptions) {
@@ -168,17 +168,13 @@ public final class MultiOptionQuestionDefinition extends QuestionDefinition {
       }
     }
 
-    if (minChoicesRequired.isPresent() && maxChoicesAllowed.isPresent()) {
-      if (minChoicesRequired.getAsInt() == 0 && maxChoicesAllowed.getAsInt() == 0) {
-        errors.add(CiviFormError.of("Cannot require exactly 0 choices"));
-      }
-
-      if (minChoicesRequired.getAsInt() > maxChoicesAllowed.getAsInt()) {
-        errors.add(
-            CiviFormError.of(
-                "Minimum number of choices required must be less than or equal to the maximum"
-                    + " choices allowed"));
-      }
+    if (minChoicesRequired.isPresent()
+        && maxChoicesAllowed.isPresent()
+        && minChoicesRequired.getAsInt() > maxChoicesAllowed.getAsInt()) {
+      errors.add(
+          CiviFormError.of(
+              "Minimum number of choices required must be less than or equal to the maximum choices"
+                  + " allowed"));
     }
 
     return errors.build();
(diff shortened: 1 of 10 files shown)


server/app/services/question/types/MultiOptionQuestionDefinition.java
```diff
@@ -144,41 +144,25 @@ ImmutableSet<CiviFormError> internalValidate(Optional<QuestionDefinition> previo
 
     OptionalInt minChoicesRequired = getMultiOptionValidationPredicates().minChoicesRequired();
     OptionalInt maxChoicesAllowed = getMultiOptionValidationPredicates().maxChoicesAllowed();
-    if (minChoicesRequired.isPresent()) {
-      if (minChoicesRequired.getAsInt() < 0) {
```
melanie-exygy, line 148: **suggestion (blocking):** From my understanding, best practice is to validate twice, once client-side and once server-side, to be certain data is in the right format.  Client-side validation can be easy for malicious actors to bypass.  Also, server-side validation means we can guarantee the API will return valid data. Without it, we can't make that promise.  My suggestion would be to keep this validation and to add server-side validation for the min >= 0 and max is >= 1 on the other types of question configs.

[An article of using both types of validation](https://surveyjs.io/stay-updated/blog/client-server-data-validation#:~:text=In%20general%2C%20it%20is%20recommended,all%20data%20is%20validated%20correctly.)

server/test/services/question/types/IdQuestionDefinitionTest.java
```diff
@@ -0,0 +1,65 @@
...
+        new Object[] {OptionalInt.of(1), OptionalInt.of(1), Optional.<String>empty()},
+        new Object[] {
+          OptionalInt.of(2),
+          OptionalInt.of(1),
+          Optional.of("Minimum length must be less than or equal to the maximum length")
+        });
+  }
+
+  @Test
+  @Parameters(method = "getValidationTestData")
```
melanie-exygy, line 36: **praise:** I love how readable this test class is.  Thanks for using the test parameters and avoiding duplication.

jeffpw-goog: As usual this was just copied from somewhere else, but thank you :)

server/app/services/question/types/MultiOptionQuestionDefinition.java
jeffpw-goog: Thanks for the feedback and explanation! I updated all the questions to include server-side validation to the min/max values.

melanie-exygy (approved):

PR #7965, opened 2024-07-03
Moving toasts to alerts on the eligibility ui

gwendolyngoetz:
### Description

Moves from showing eligibility status from a toast message to a USWDS alert. 

Translations were pushed to main yesterday.

There a bunch of images in the PR, mostly Northstar related, that shouldn't be here. They are oddly shifted by a pixel or two. Greg had run into this issue. You'll just have to ignore them for now until we can figure out if there's a fix in Playwright to prevent this.

## Release notes

UX improvements on application review and edit pages for showing eligibility status.

#### General

#### User visible changes

### Issue(s) this completes

Related to: #5541


server/app/views/components/AlertFragment.html
```diff
@@ -0,0 +1,36 @@
+<div
+  th:fragment="alert(alertSettings, headingLevel)"
```
gwendolyngoetz, line 2: Note: I just could not get this to let me pass in an enum. Not adding the headingLevel to the AlertSettings class since setting the heading is definitely a UI thing and doesn't need to be thought about when making the AlertSettings in a service or controller.

avaleske: greg ran into issues passing enums too

gwendolyngoetz: If it's part of an object it works

server/app/views/components/AlertFragment.html
```diff
@@ -0,0 +1,36 @@
...
+  th:class="${'usa-alert usa-alert--' + alertSettings.alertType.name().toLowerCase()}"
+  aria-live="polite"
+  role="alert"
+>
+  <div class="usa-alert__body">
+    <h2
+      class="usa-alert__heading"
+      th:if="${alertSettings.title.isPresent() && headingLevel == 'H2'}"
```
melanie-exygy, line 10: **question:** It doesn't seem ideal to me that we would have so many different heading levels for our alerts.  Are we using all of these now?  If we limit the header levels here, it would encourage people to be consistent.  Would you mind checking with Sijia on what the ideal header levels should be?

browser-test/image_snapshots/admin_program_preview_test/program-preview-application-block.png
melanie-exygy, line 1: **question:** (non-blocking):  I noticed that in several of these snapshots, we're showing an alert where there wasn't previously a toast.  Should we only be showing the alert under the same circumstances that we currently show the toast?

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -877,9 +877,8 @@ test.describe('Trusted intermediaries', () => {
       await applicantQuestions.clickEdit()
       await applicantQuestions.answerNumberQuestion('5')
       await applicantQuestions.clickNext()
-      await validateToastMessage(page, 'may qualify')
+      await applicantQuestions.expectMayBeEligibileHeadingToBeVisible()
```
melanie-exygy, line 880: **suggestion:** (non-blocking):  What would you think of checking here that the TI-specific alert message is shown?  ("Your client may be eligible...")

browser-test/src/support/applicant_questions.ts
```diff
@@ -694,4 +694,22 @@ export class ApplicantQuestions {
     await this.submitFromReviewPage()
     await this.page.click('text=End session')
   }
+
+  async expectMayBeEligibileHeadingToBeVisible() {
```
melanie-exygy, line 698: **question:** Should this be `expectMayBeEligibleAlertToBeVisible`?

server/app/views/components/AlertFragment.html
```diff
@@ -0,0 +1,36 @@
+<div
+  th:fragment="alert(alertSettings, headingLevel)"
+  th:class="${'usa-alert usa-alert--' + alertSettings.alertType.name().toLowerCase()}"
+  aria-live="polite"
```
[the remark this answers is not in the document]
gwendolyngoetz: Oh, interesting. I'll make this change in a follow up PR so I can do it both for the j2html/northstar bits in a contained commit.

server/app/views/components/AlertFragment.html
```diff
@@ -0,0 +1,36 @@
...
+  role="alert"
```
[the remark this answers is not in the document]
gwendolyngoetz: I have not, but I can give it a try when I update the alerts with the assertive change

server/app/views/components/AlertFragment.html
gwendolyngoetz: I talked with Sijia before about having to move it at one point because of one place was failing accessibility tests due to the ordering of H1-6 tags or because some pages had a gap. Which were also different between current and northstar >_< 

Unless we only ever use the non-slim ones in one place with no other headings deviating we'll always need the ability to override.

I don't disagree with you, we should make it easy to be consistent. I don't want to block this PR though, we should get with Don and Sijia to make sure NS is doing H1-6 correctly as a whole.

server/app/views/components/AlertFragment.html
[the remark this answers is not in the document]
melanie-exygy: Sounds good.  Thanks!

browser-test/image_snapshots/admin_program_preview_test/program-preview-application-block.png
gwendolyngoetz: Yep. Apparently it's was found to be more friendly to people based on results from community.

browser-test/src/trusted_intermediary.test.ts
gwendolyngoetz: I don't think it's necessary. There's a self-contained class that calculates the message which I have unit tests already covering.

browser-test/src/support/applicant_questions.ts
gwendolyngoetz: Good idea!

server/app/views/components/AlertFragment.html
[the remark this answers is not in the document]
gwendolyngoetz: FYI https://github.com/civiform/civiform/issues/8026

server/app/views/components/AlertFragment.html
melanie-exygy: That sounds good.  I think Sijia already has a work item to review our headings and make sure they're consistent, so, hopefully, they'll review the alert headings as part of that.

browser-test/image_snapshots/admin_program_preview_test/program-preview-application-block.png
melanie-exygy: Ah, interesting.  Well, I guess it's an improvement!  Thanks for explaining.

melanie-exygy (approved):

PR #7974, opened 2024-07-05
[Status_Migration] Removing statuses dependency from programs 

swatkat1:
### Description

Changes required to start writing to only ApplicationStatuses table and not to Programs table. 

## Release notes

Changes required to start writing to only ApplicationStatuses table and not to Programs table. 

#### General

### Issue(s) this completes

Part of #7040


server/app/repository/ApplicationStatusesRepository.java
```diff
@@ -77,4 +78,17 @@ public void createOrUpdateStatusDefinitions(
       transaction.commit();
     }
   }
+
+  public ImmutableList<ApplicationStatusesModel> getAllApplicationStatusModels(String programName) {
+
+    return database
+        .find(ApplicationStatusesModel.class)
+        .setLabel("ApplicationStatusesModel.findList")
```
gwendolyngoetz, line 86: Don't forget to update this

server/app/services/applicationstatuses/StatusDefinitions.java
```diff
@@ -59,6 +59,10 @@ private static void assertStatusNamesNonEmptyAndUnique(ImmutableList<Status> sta
         "The provided set of statuses may not contain empty statusTexts.");
   }
 
+  public Optional<StatusDefinitions.Status> getDefaultStatus() {
+    return statuses.stream().filter(StatusDefinitions.Status::computedDefaultStatus).findFirst();
+  }
+
```
gwendolyngoetz, line 65: This needs `@JsonIgnore` from the `com.fasterxml.jackson.annotation` namespace.

avaleske (commented):
I haven't looked at everything yet but here's my comments so far.

Also:
- I might rename `StatusService` to `StatusesService` or `ApplicationStatusesService` to match the other entities
- I haven't looked closely yet, but I wonder if the `StatusService` should provide a `lookupActiveStatusDefinitions()` method that just pipes through to its internal `ApplicationStatusesRepository` so that callsites that currently use both don't need both.

server/app/services/applicationstatuses/DuplicateStatusException.java
```diff
@@ -1,4 +1,4 @@
-package services.program;
+package services.applicationstatuses;
```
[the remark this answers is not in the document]
swatkat1: I don't think we follow camel casing in packages. Also, we have had only one word for packages so far. Should I rename this package as statuses?

server/app/controllers/admin/AdminProgramStatusesController.java
```diff
@@ -75,13 +84,15 @@ public Result index(Http.Request request, long programId) throws ProgramNotFound
...
    */
   @Secure(authorizers = Authorizers.Labels.CIVIFORM_ADMIN)
-  public Result createOrUpdate(Http.Request request, long programId)
+  public Result createOrUpdate(Http.Request request, Long programId)
       throws ProgramNotFoundException {
     requestChecker.throwIfProgramNotDraft(programId);
-    ProgramDefinition program = service.getFullProgramDefinition(programId);
-    int previousStatusCount = program.statusDefinitions().getStatuses().size();
+    ProgramDefinition program = programService.getFullProgramDefinition(programId);
+    StatusDefinitions currentStatusDefinitions =
```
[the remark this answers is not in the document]
swatkat1: done!

server/app/services/applicationstatuses/DuplicateStatusException.java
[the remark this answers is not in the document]
avaleske: oh you're right:
https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names

then leaving it as `applicationstatuses` is fine. In that case I think the service itself should also be `ApplicationStatusesService`. If you leave the package as `statuses` then I think renaming the service to `StatusesService` works.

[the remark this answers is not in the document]
swatkat1: I changed the package to just "status" and left the service class as such!

server/app/services/statuses/StatusService.java
```diff
@@ -0,0 +1,258 @@
...
+   * @throws DuplicateStatusException If the updated status already exists in the list of available
+   *     statuses for application review.
+   */
+  public ErrorAnd<StatusDefinitions, CiviFormError> editStatus(
+      long programId,
+      String toReplaceStatusName,
+      Function<StatusDefinitions.Status, StatusDefinitions.Status> statusReplacer)
+      throws ProgramNotFoundException, DuplicateStatusException {
+    ProgramDefinition program = programService.getFullProgramDefinition(programId);
+    StatusDefinitions currentStatus =
```
avaleske, line 102: Similarly here, I think all the callsites have a `ProgramDefinition` available so we don't need to fetch the program again inside the method, we can just pass in the admin name.

server/app/services/statuses/StatusService.java
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Removes an existing status from the list of available statuses for application reviews.
+   *
+   * @param programId The program to update.
+   * @param toRemoveStatusName The name of the status that should be removed.
+   * @throws ProgramNotFoundException If the specified Program could not be found.
+   */
+  public ErrorAnd<StatusDefinitions, CiviFormError> deleteStatus(
+      long programId, String toRemoveStatusName) throws ProgramNotFoundException {
+    ProgramDefinition program = programService.getFullProgramDefinition(programId);
```
avaleske, line 142: ditto here re: admin name

server/app/services/statuses/StatusService.java
```diff
@@ -0,0 +1,258 @@
...
+            .collect(ImmutableList.toImmutableList());
+    if (!localizationStatusNames.equals(configuredStatusNames)) {
+      throw new OutOfDateStatusesException();
+    }
+  }
+
+  /** Get all the statusDefinitions for a given program */
+  public ImmutableList<StatusDefinitions> getAllPossibleStatusDefinitions(Long programId)
+      throws ProgramNotFoundException {
+    ProgramDefinition programDefinition = programService.getFullProgramDefinition(programId);
```
avaleske, line 253: ditto here. I think we can just pass in the admin name.

server/app/services/statuses/StatusService.java
```diff
@@ -0,0 +1,258 @@
...
+   * @param programId The program to update.
+   * @param status The status that should be appended.
+   * @throws ProgramNotFoundException If the specified Program could not be found.
+   * @throws DuplicateStatusException If the provided status to already exists in the list of
+   *     available statuses for application review.
+   */
+  public ErrorAnd<StatusDefinitions, CiviFormError> appendStatus(
+      long programId, StatusDefinitions.Status status)
+      throws ProgramNotFoundException, DuplicateStatusException {
+    ProgramDefinition program = programService.getFullProgramDefinition(programId);
```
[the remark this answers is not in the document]
avaleske: Alternatively, the programs definition should be cached, so if you think it's nicer for the caller to pass in a programId that's fine. In this case, I noticed we're literally calling `program.id()` at the callsite so `program.adminName()` is the same amount of work.

server/app/views/admin/programs/ProgramTranslationView.java
```diff
@@ -83,7 +88,9 @@ protected String languageLinkDestination(String programName, Locale locale) {
   }
 
   private ImmutableList<DomContent> formFields(
-      ProgramDefinition program, ProgramTranslationForm translationForm) {
+      ProgramDefinition program,
+      ProgramTranslationForm translationForm,
+      StatusDefinitions currenStatusDefnitions) {
```
avaleske, line 93: typo here too

server/test/controllers/admin/AdminProgramStatusesControllerTest.java
```diff
@@ -262,7 +265,9 @@ public void update_editExistingStatus() throws ProgramNotFoundException {
 
     // Load the updated program and ensure the status is present.
     assertThat(
-            programService.getFullProgramDefinition(program.id).statusDefinitions().getStatuses())
+            repo.lookupActiveStatusDefinitions(
+                    programService.getFullProgramDefinition(program.id).adminName())
```
avaleske, line 269: **thought, non-blocking**: It seems like it'd be useful to have a method on the `ProgramService` that can map from program id to admin name.

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -2702,7 +2702,7 @@ public void relevantProgramsForApplicant_withApplicationStatusAndOlderProgramVer
 
     // Publish a new program that has an updated set of status configurations that doesn't include
     // the application's status. The displayed status definition configuration should be pulled
-    // from the program version associated with the application.
+    // from the latest program version's status configuration
```
avaleske, line 2705: This comment shouldn't reference the program version at all, right? Just `the program's ACTIVE status definitions` or similar?

server/test/services/program/ProgramDefinitionTest.java
```diff
@@ -1426,7 +1432,10 @@ public void serializeThenDeserialize_allFieldsPresent() throws JsonProcessingExc
                 Locale.ITALIAN,
                 "Italian summary image description"));
     assertThat(result.externalLink()).isEqualTo("external.link");
-    assertThat(result.statusDefinitions().getStatuses())
+    assertThat(
+            applicationStatusesRepository
+                .lookupActiveStatusDefinitions(result.adminName())
+                .getStatuses())
```
avaleske, line 1438: We can just remove this assert, and the concept of statuses from this test entirely, right?

No need to test the `ApplicationStatusesRepository` here

server/test/services/statuses/StatusServiceTest.java
```diff
@@ -0,0 +1,675 @@
...
+            .withLocalizedConfirmationMessage(Locale.FRENCH, "")
+            .setLocalizedSummaryImageDescription(
+                LocalizedStrings.of(
+                    Locale.US,
+                    "English image description",
+                    Locale.FRENCH,
+                    "existing French image description"))
+            .withStatusDefinitions(
+                new StatusDefinitions(ImmutableList.of(STATUS_WITH_EMAIL, STATUS_WITH_NO_EMAIL)))
+            .build();
```
avaleske, line 405: Throughout this file, instead of using the `ProgramBuilder` to save statuses, can we save them directly using the `ApplicationStatusesRepository`? Then we don't need to deal with creating programs.

Since they were copied from the `ProgramService`, these tests are testing a lot of the program behavior unnecessarily.

server/test/services/statuses/StatusServiceTest.java
```diff
@@ -0,0 +1,675 @@
...
+        ps.updateLocalization(program.id, Locale.FRENCH, updateData);
+
+    assertThat(result.isError()).isFalse();
+    ProgramDefinition definition = result.getResult();
+    assertThat(definition.localizedName().get(Locale.FRENCH)).isEqualTo("new French name");
+    assertThat(definition.localizedDescription().get(Locale.FRENCH))
+        .isEqualTo("new French description");
+    assertThat(definition.localizedSummaryImageDescription().isPresent()).isTrue();
+    assertThat(definition.localizedSummaryImageDescription().get().get(Locale.FRENCH))
+        .isEqualTo("new French image description");
```
avaleske, line 444: for example, we don't care if this part of the localization worked correctly, that's already tested in the ProgramService tests.

server/test/services/program/ProgramServiceTest.java
```diff
@@ -2187,192 +2185,6 @@ public void newProgramFromExisting() throws Exception {
...
-    assertThat(definition.statusDefinitions().getStatuses()).isEqualTo(expectedStatuses);
-    assertThat(
-            applicationStatusesRepo
-                .lookupActiveStatusDefinitions(definition.adminName())
-                .getStatuses())
-        .isEqualTo(expectedStatuses);
-  }
-
-  @Test
-  public void updateLocalizations_updatesExistingLocale() throws Exception {
```
avaleske, line 2289: We shouldn't move this whole test into the `StatusService`, just the parts that are testing statuses. We can leave the rest here.

ditto for the other tests that we moved in their entirety.

server/test/controllers/admin/AdminProgramTranslationsControllerTest.java
```diff
@@ -302,8 +308,22 @@ public void update_outOfSyncFormData_redirectsToFreshData() throws Exception {
...
+            .get()
+            .getProgramDefinition();
+    assertThat(
+            applicationStatusesRepository
+                .lookupActiveStatusDefinitions(freshProgram.adminName())
+                .getStatuses())
+        .isEqualTo(
+            applicationStatusesRepository
+                .lookupActiveStatusDefinitions(program.getProgramDefinition().adminName())
+                .getStatuses());
```
avaleske, line 326: Was this change intentional? I don't know what this is testing when it's part of this test.

server/test/models/ProgramModelTest.java
```diff
@@ -93,6 +96,8 @@ public void canSaveProgram() throws UnsupportedQuestionTypeException {
     ProgramModel program = new ProgramModel(definition);
 
     program.save();
+    appStatusRepo.createOrUpdateStatusDefinitions(
+        program.getProgramDefinition().adminName(), new StatusDefinitions());
```
avaleske, line 100: throughout this file, I think we can just remove all mention of statuses instead of also creating status definitions to assert

avaleske (commented):
Overall this is looking good! To summarize my comments:
- double check for typos
- only move the parts of the tests that are testing statuses to the StatusServiceTest
- we create a program in a lot of places in these tests that we don't need to. We use `ProgramBuilder` because it creates a status, but we could just create the status directly using the repository
- we should remove statuses from the program definition. if not in this PR, then in a followup.

server/app/controllers/admin/AdminProgramTranslationsController.java
```diff
@@ -138,27 +153,53 @@ public Result update(Http.Request request, String programName, String locale)
...
+      ToastMessage errorMessage = ToastMessage.errorNonLocalized(joinErrors(result.getErrors()));
+      return ok(
+          translationView.render(
+              request,
+              localeToUpdate,
+              program,
+              currentStatuDefnitions,
+              translationForm,
+              Optional.of(errorMessage)));
+    }
```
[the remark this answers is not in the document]
swatkat1: Hi Austin, there are two updateLocalization() now, one in Program Service (which doesn't throw OutOfDateStatusException) and the other one in StatusService which throws it. Hence one is in try-catch block and the other isn't.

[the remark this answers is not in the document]
avaleske: Ah! Can we add a comment explaining that?

server/app/services/applicant/ApplicantService.java
```diff
@@ -1153,15 +1158,17 @@ private ApplicationPrograms relevantProgramsForApplicantInternal(
             }
             programNamesWithApplications.add(programName);
           } else if (maybeSubmittedApp.isPresent() && activeProgramNames.containsKey(programName)) {
-            // When extracting the application status, the definitions associated with the program
-            // version at the time of submission are used. However, when clicking "reapply", we use
-            // the latest program version below.
+            // When extracting the application status, the Statuses present in the current program
+            // versions are used
```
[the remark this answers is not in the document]
swatkat1: Since we renamed it, I will update the comment. Or in the other hand, we are calling it activeStatusDefinitions now, I don't see the need for this comment.

server/app/services/applicant/ApplicantService.java
```diff
@@ -1153,15 +1158,17 @@ private ApplicationPrograms relevantProgramsForApplicantInternal(
...
             ProgramDefinition applicationProgramVersion =
                 programRepository.getShallowProgramDefinition(maybeSubmittedApp.get().getProgram());
+            StatusDefinitions activeStatusDefinitions =
+                applicationStatusesRepository.lookupActiveStatusDefinitions(
```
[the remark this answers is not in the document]
swatkat1: Good catch! I will try to move it outside the loop.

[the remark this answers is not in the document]
swatkat1: I checked, the foreach is happening for each program, so for each program present, we are checking if the program is in 1. Not-started state, 2. In incomplete state, 3. Completed state. Since each program is different, we cannot move this call outside the loop.

server/app/services/statuses/StatusService.java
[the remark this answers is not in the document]
swatkat1: I changed the appendStatus to directly accept programName instead of programId. Good suggestion!

server/app/services/statuses/StatusService.java
swatkat1: done!

server/app/services/statuses/StatusService.java
swatkat1: done!

server/test/controllers/admin/AdminProgramStatusesControllerTest.java
swatkat1: since getProgramDefinition() is calling the cached program, I am not too worried!

server/test/controllers/admin/AdminProgramTranslationsControllerTest.java
swatkat1: We testing if the data is updated or not. When there is an error, it shouldn't be updated.

server/test/models/ProgramModelTest.java
swatkat1: Right now, without status, the programbuilder class is complaining. Let me tackle it in the next PR.

server/test/services/statuses/StatusServiceTest.java
swatkat1: Moved it out to ProgramServiceTest

server/test/services/program/ProgramDefinitionTest.java
swatkat1: In the next PR..

server/test/services/program/ProgramServiceTest.java
swatkat1: split these tests into programServiceTest and StatusServiceTest

server/app/services/applicationstatuses/StatusService.java
```diff
@@ -0,0 +1,260 @@
...
+        statusDefinitions.getStatuses().stream()
+            .map(StatusDefinitions.Status::statusText)
+            .collect(ImmutableList.toImmutableList());
+    if (!localizationStatusNames.equals(configuredStatusNames)) {
+      throw new OutOfDateStatusesException();
+    }
+  }
+
+  /** Get all the statusDefinitions for a given program */
+  public ImmutableList<StatusDefinitions> getAllStatusDefinitions(Long programId)
```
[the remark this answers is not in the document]
swatkat1: done ! changed to getAllPossibleStatusDefinitions()

server/test/services/statuses/StatusServiceTest.java
swatkat1: I think the problem arises because of using the same LocalizationUpdate param for both program update and status update. But I hear your point. Program data shouldn't be needed here. I plan to do a followup PR to remove status from programBuilder and get rid of everywhere. This PR is doing many things already, complexity of removing it will be more,

server/test/services/applicant/ApplicantServiceTest.java
swatkat1: done!

server/app/repository/ApplicationStatusesRepository.java
```diff
@@ -77,4 +64,23 @@ public void createOrUpdateStatusDefinitions(
...
+    ImmutableList<ApplicationStatusesModel> allApplicationStatusModels =
+        database
+            .find(ApplicationStatusesModel.class)
+            .setLabel("GetAllApplicationStatusesModel.findList")
+            .where()
+            .in("program_name", programName)
+            .query()
+            .findList()
+            .stream()
+            .collect(ImmutableList.toImmutableList());
```
gwendolyngoetz, line 80: (question) Is this list being sorted elsewhere? If not it should have a default case-insensitive sort put on it on the most logical text field the user will see. Just file an issue for your self and do it in a separate PR.

swatkat1: Done- https://github.com/civiform/civiform/issues/8131

gwendolyngoetz (approved):
Since Danielle said she already started looking, you should wait for her to complete reviewing before merging.

server/app/views/admin/programs/ProgramTranslationView.java
dkatzz, line 93: **suggestion:** Typo here with currenStatusDefnitions -> currentStatusDefinitions

server/test/controllers/admin/AdminProgramStatusesControllerTest.java
dkatzz: It would be nice just to make it simpler for the various callers and readers of the code (less about the performance), but agree its nonblocking.

dkatzz (approved):

server/app/views/admin/programs/ProgramTranslationView.java
swatkat1: changed

Issue #7528, opened 2024-05-21
[Program Migration] Handle admin name collisions in program/question import


rockycodes:
When a program or question is imported that shared an admin name with an existing program or question, programmatically rename the incoming program/question to avoid the collision.

Question admin names are auto-generated (I think) so this might be handled for questions when we add them to the db. 

Part of #7087 

Notes: 
- move helper methods to service class instead of controller
- make strings in tests into Java text blocks like in `JsonExporterServiceTest`

melanie-exygy:
Couldn't we end up with a lot of duplicate programs?  What about replacing the existing program if the name already exists?

rockycodes:
@melanie-exygy this is what was settled on in the PRD (see comments [here](https://docs.google.com/document/d/1mu4Nr-gsEj62NAXWoBjooVwkVUOP79I8ijAK4-jpsE0/edit?resourcekey=0-jwFNNYAH-SPW59-GolzzIg&disco=AAABKiFI0fo)). It makes sense to me, since this is a manual process, I think it's less likely that an admin would be importing an identical program and more likely they would be importing a different program that happens to have the same admin name.

gwendolyngoetz:
I left a more detailed comment on the PRD. But I do see it being common that a program would be imported with the same name. Example 
- Make project on Test environment
- Export -> Import to prod v1
- Make changes to program on Test
- Export -> import to prod making a v2

swatkat1:
@rockycodes  - Just letting you know we need to move the statuses data for that program from the ApplicationStatuses table too.

rockycodes:
@swatkat1 thank you for the reminder! I talked with Isabel/product a couple of weeks ago and the recommendation was to not migrate statuses for v1 of this work, but add that in for v2. Do you see any issues with ignoring statuses from a technical perspective?

swatkat1:
Yes, if an active status is not found for a give program, the code will throw an exception. If not importing the current status data, we should provide a way to create an empty status in the code. 

`  // Create a new row in  ApplicationStatuses table for the new program
      applicationStatusesRepository.createOrUpdateStatusDefinitions(
          adminName, new StatusDefinitions());
`

PR #8014, opened 2024-07-10
[Program Migration] Save questions to programs

### LINKED ISSUE #7822: [Program Migration] Bulk save questions and add to imported programs
With program import, we will need the ability to bulk add questions to the db which we cannot currently do. We will need to save the questions before saving the program so we can reference the ids of the newly saved/updated questions in the program blocks. More details to be fleshed out in the [TDD](https://docs.google.com/document/d/1a_Uo5J5zQQ3jo__vTwbPHYIzxUemdZjFrcgQBMqUBr0/edit#heading=h.mau6tuqgv74d).

rockycodes:
### Description

This PR adds the ability to save full programs with questions when you click the `Save Program` button in program import. This only handles adding questions and programs when there are no admin name collisions (no existing questions or program with admin names that match the incoming questions or program). Handling admin name collisions will be done in the [following issue](https://github.com/civiform/civiform/issues/7528).

Some important things to note about saving a program with questions:
- enumerator type questions must be saved _before_ questions that reference that enumerator so that we can insert the updated id into the `enumeratorId` field for the child questions
- eligibility and visibility predicates must be updated with the id of the imported question after it is saved
- PAI tags are stripped from incoming questions, but the "universal" setting is maintained

#### General

1. Enable the `PROGRAM_MIGRATION_ENABLED` flag
2. Add a program and questions (can seed programs and questions via the developer view)
3. Try adding different scenarios to the program like visibility and eligibility predicates, PAI questions, universal questions, enumerator questions, etc.
4. Export the program + questions via the export tab and copy the exported json
5. Clear out your db
6. Paste the json into the import tab
7. Click "Display program information" and then "Save Program"
8. Navigate to the program and confirm all information was ported over correctly

You can use staging to create and export a program, but you will need to manually remove the `enumeratorId` field from any questions that should not reference an enumerator question before importing.

### Issue(s) this completes

Fixes #7822


server/app/services/question/types/QuestionDefinitionConfig.java
```diff
@@ -13,6 +14,7 @@
 
 @AutoValue
 @JsonDeserialize(builder = AutoValue_QuestionDefinitionConfig.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
```
rockycodes, line 17: This Jackson annotation tells the serializer to only include optional fields if there are not empty. This fixes a bug where null `enumeratorId` fields were being deserialized to 0 and then our code was reading that as if it was an actual enumuratorId.

server/test/controllers/ti/TrustedIntermediaryControllerTest.java
```diff
@@ -248,7 +248,6 @@ public void addClient_WithMissingDob() {
                 profileUtils.currentUserProfile(requestBuilder.build()).get())
             .get();
     Result result = tiController.addClient(trustedIntermediaryGroup.id, requestBuilder.build());
-    System.out.println(contentAsString(result));
```
rockycodes, line 251: This was an erroneous print statement I found that I'm pretty sure was checked in by accident so I'm removing it.

server/test/services/migration/ProgramMigrationServiceTest.java
```diff
@@ -107,98 +112,24 @@ public void deserialize_notEnoughInfoToCreateProgramDef_returnsError() {
...
+    assertThat(question.getQuestionText().getDefault())
+        .isEqualTo("Please enter your first and last name");
   }
-
-  /**
-   * This contains the bare minimum needed to parse JSON into a program definition. The
-   * admin_program_migration.test.ts browser test has tests for a program with many blocks and
-   * questions.
-   */
-  public static final String EXAMPLE_PROGRAM_JSON =
```
rockycodes, line 128: I moved this to `AdminImportControllerTest` so it lives with all the other sample json strings I created for testing.

server/app/controllers/admin/AdminImportController.java
```diff
@@ -114,28 +129,172 @@ public Result saveProgram(Http.Request request) {
...
-        programMigrationService.deserialize(jsonString);
+    return programMigrationService.deserialize(jsonString);
+  }
 
-    if (deserializeResult.isError()) {
-      return ok(
-          adminImportViewPartial
-              .renderError(deserializeResult.getErrors().stream().findFirst().orElseThrow())
-              .render());
+  private ImmutableMap<String, QuestionDefinition> updateEnumeratorQuestionsAndSave(
```
rockycodes, line 176: Question for reviewer(s): Do you think all these helper methods are fine here or should they be in the `ProgramMigrationService` class?

avaleske: I think it's fine either way but I'd slightly lean toward putting them in the service

server/test/controllers/admin/AdminImportControllerTest.java
```diff
@@ -155,41 +160,859 @@ public void hxImportProgram_notEnoughInfoToCreateProgramDef_error() {
...
+            .where()
+            .eq("name", "dob")
+            .findOne()
+            .getQuestionDefinition();
+
+    assertThat(questionDefinition.isUniversal()).isTrue();
+  }
+
+  public static final String PROGRAM_JSON_WITHOUT_QUESTIONS =
+      "{\n"
```
avaleske, line 379: maybe as a followup PR, but it'd be nice to convert this to java text blocks like the `JsonExporterServiceTest` uses now.

avaleske (approved):

server/app/controllers/admin/AdminImportController.java
rockycodes: Leaving them here for now since it's just a slight preference and it's a bit of work to move them. But I'll move them in the next issue.

server/test/controllers/admin/AdminImportControllerTest.java
rockycodes: I'll add that as a todo in a follow-up issue.

server/app/services/question/types/QuestionDefinitionConfig.java
rockycodes: done.

PR #8021, opened 2024-07-11
Update template date

avaleske:
Update translations ticket template year


azhangggggg (approved):
Thanks for updating the template!

gwendolyngoetz:
Maybe you should just set it to 2025

PR #8041, opened 2024-07-14
Updating Existing Draft Application to Latest Program Version

gwendolyngoetz:
### Description

Final bit of work to enable application program fast forwarding feature. This will update draft applications to the latest published version if there is a new one.

The applicant program selection home page is now based on the latest version of the program as well; but the application is linked to the old program version which will kick off the fast forwarding process.

### Release notes

Applicants will always be presented with the most latest version of a program when applying. Draft applications will automatically be upgraded the next time the applicant loads the application. Eligibility tags when selecting programs are based on the latest version of the program.

#### General

#### New Features

- [x] Add new FeatureFlag env vars to `server/conf/helper/feature-flags.conf`

### Issue(s) this completes

Fixes: #5541

### DIFF
--- a/browser-test/src/applicant/application_version_fast_forward.test.ts
+++ b/browser-test/src/applicant/application_version_fast_forward.test.ts
@@ -9,6 +9,7 @@ import {
   logout,
   closeWarningMessage,
   AdminPredicates,
+  enableFeatureFlag,
   disableFeatureFlag,
 } from '../support'
 import {QuestionSpec} from '../support/admin_programs'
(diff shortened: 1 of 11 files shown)


browser-test/src/applicant/application_version_fast_forward.test.ts
```diff
@@ -486,6 +487,503 @@ test.describe('Application Version Fast-Forward Flow', () => {
       })
     })
   })
+
+  test('all major steps - fast forward flag enabled', async ({browser}) => {
```
gwendolyngoetz, line 491: This is a slightly modified version of the disabled version. Guess there's not a good way to show the diff since I'm keeping the feature flag disabled version for now.

browser-test/src/applicant/application_version_fast_forward.test.ts
```diff
@@ -486,6 +487,503 @@ test.describe('Application Version Fast-Forward Flow', () => {
...
+
+      Program: v3       Remove (E), add (H), no-change (G, B, A, E, F)
+        Block: First
+          Question: G
+        Block: Second
+          Question: B  ---> Eligibility
+        Block: Third
+          Question: A
+        Block: Fourth
+                    H ---> Eligibility
```
ssandbekkhaug, line 547: Super nit: add "Question:" ?

browser-test/src/applicant/application_version_fast_forward.test.ts
```diff
@@ -486,6 +487,503 @@ test.describe('Application Version Fast-Forward Flow', () => {
...
+        // a civiform admin, the program admin won't have permissions to the program.
+        await programAdminActor.login()
+        await programAdminActor.viewApplications()
+      })
+
+      await test.step('sees submitted application with questions from program version v3', async () => {
+        const cardLocator = programAdminActor.getCardLocator()
+        await expect(cardLocator).toHaveCount(1)
+
+        const cardButton = cardLocator.getByRole('link', {name: 'View'})
```
ssandbekkhaug, line 962: Question: Is it important to check the ID, or is the purpose of this step to verify that the correct version is showing?

An alternative might be to change the program name for each version and assert the appropriate text is visible/hidden.

browser-test/src/applicant/application_version_fast_forward.test.ts
```diff
@@ -739,10 +1268,21 @@ class FastForwardApplicantActor {
     programName: string,
     browser: Browser,
   ): Promise<FastForwardApplicantActor> {
-    const context = await browser.newContext()
+    const context = await browser.newContext({
+      recordVideo: {
+        dir: 'tmp/videos/',
```
ssandbekkhaug, line 1273: Non-blocking: should this be a constant?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -1367,4 +1387,35 @@ private Result handleUpdateExceptions(Throwable throwable) {
     }
     throw new RuntimeException(throwable);
   }
+
+  /**
+   * Check if the application needs to be updated to a newer program version. If it does updated and
```
ssandbekkhaug, line 1392: Super nit: "If it does, update and ..."

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -1367,4 +1387,35 @@ private Result handleUpdateExceptions(Throwable throwable) {
...
+   * return a redirect result back to the review page
+   *
+   * @return {@link Result} if application was updated; empty if not
+   */
+  private Optional<Result> updateApplicationToLatestProgramVersionIfNeeded(
```
ssandbekkhaug, line 1397: Question (Blocking): This method looks very similar to the one below. Can it live in a common file and be referenced by each caller?

server/app/repository/ApplicationRepository.java
```diff
@@ -353,7 +353,13 @@ public void updateDraftApplicationProgram(long applicantId, long programId) {
                   queryProfileLocationBuilder.create("updateDraftApplicationProgram"))
               .findOne();
 
-      // Update the existing draft
+      if (existingDraft == null) {
+        throw new RuntimeException(
+            String.format(
+                "Did not find expected draft application for applicantId=%d, programId=%d.",
```
ssandbekkhaug, line 359: Praise: This is a good error message because it includes the applicant ID and program ID, which should save time debugging :)

(No action required)

server/test/services/applicant/ApplicantServiceFastForwardEnabledTest.java
```diff
@@ -0,0 +1,3897 @@
+package services.applicant;
```
ssandbekkhaug, line 1: FYI: For this file, instead of manually reviewing 3898 lines, I did a diff with ApplicantServiceTest.java. Let me know if there's anywhere you want a closer look

browser-test/src/applicant/application_version_fast_forward.test.ts
gwendolyngoetz: I want to explicitly verify that the version id is the expected value since it changes under the covers for the applicant. I don't want to rely on just changing the text in case other things affect that and checking by ID maintains a consistent value as I mutate things.

browser-test/src/applicant/application_version_fast_forward.test.ts
gwendolyngoetz: Yep. Const in Typescript isn't the same as const in Java. It just means I can't reassign the variable. 

It's not confusing at all :|

server/app/controllers/applicant/ApplicantProgramBlocksController.java
gwendolyngoetz: They are indeed exactly the same and I expected this to come up as I went back and forth trying to decide on where to put it. :) 

Options include:
- Pushing to the base controller, but that adds more dependencies and will require touching anything that needs to now pass in those new dependencies to the base class. The touch is too broad, especially for just being used in two places.
- Move it to the service class, but now the service classes need to know about routing and URLs. Controller stuff that it shouldn't know.
- Move it to a new class and either call it from a static method or add another thing to the already long list of things being injected. I think it's pretty specialized and creating a new class seems overboard. 
- Leave it as is, duplicated in two controllers. Downside a tiny bit of duplication. Upside maintains a locality to where it's actually used.

I'm inclined to leave it as it. I'm open to suggestions if you like one of the other options or have another idea.

ssandbekkhaug: I don't have a strong opinion, but I would choose creating a new class with a static method. I'm afraid that someone will eventually update one but not the other.

I'm OK with submitting as-is, so feel free to close this thread

ssandbekkhaug (approved):

Issue #8046, opened 2024-07-15
Refactor Request building in unit tests


jeffpw-goog:
Many unit tests create a Request object using one of these methods:

- play.test.Helpers.fakeRequest()
- CfTestHelpers.requestBuilderWithSettings (3 overloads)

Many of those tests also won't pass without a CSRF token set on the request. In production, Play sets a CSRF token on the Request automatically. To imitate this behavior, there are ~400 calls to CSRFTokenHelper.addCSRFToken in tests even though they are not specifically testing CSRF.

Instead of making a call in each test we should set a CSRF token when constructing the fakeRequest, so that only tests explicitly testing CSRF have to think about it.

jeffpw-goog:
This is done! 🎉

PR #8049, opened 2024-07-15
Enhance FakeRequestBuilder to support unit test cases

jeffpw-goog:
### Description

Prepare FakeRequestBuilder to handle all request building in unit tests:

- Add factory methods and make the constructor private
- Update existing usages from the constructor to one of the factory methods
- Add CSRF to the request by default so that each test won't have to add CSRF
- Add a call() builder method that imitates the RequestBuilder(Call) constructor so it behaves like a builder

This supports #8046.

#### General


azhangggggg (approved):
Nice enhancement on fakeRequestBuilder! should be useful for other tests in the future.

Issue #8051, opened 2024-07-15
Homepage and application upsell page aren't using APPLICANT_PORTAL_NAME


avaleske:
**Describe the bug**
The banner on the homepage and the message on the applicant upsell page aren't using APPLICANT_PORTAL_NAME.

**To Reproduce**
Look at the homepage or the applicant upsell page (the confirmation page after submitting an application) when a guest user.

**Expected behavior**
Both of these messages should use the value set in APPLICANT_PORTAL_NAME in the settings UI

**Screenshots**
![image](https://github.com/user-attachments/assets/73309edd-178c-4bb6-a32c-3474587ebae9)

**Additional context**
It's working correctly in the upsell modal when starting an application while a guest user.

PR #8054, opened 2024-07-15
Use setting for APPLICANT_PORTAL_NAME instead of binding

nb1701:
Before we had the Settings system, we were binding APPLICANT_PORTAL_NAME directly from the config file into ApplicantAuthProviderName. Now that we have the settings system and need to pass in a Request object, this changes all the call sites to use the Settings getter instead (some had already made the change) and removes the now-unused binding.

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/8051

### DIFF
--- a/server/app/modules/MainModule.java
+++ b/server/app/modules/MainModule.java
@@ -2,7 +2,6 @@ package modules;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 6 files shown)


avaleske (approved):
Thanks for doing the northstar templates too!

non-blocking: Is there a test we should write to catch this so it doesn't happen again?

swatkat1 (approved):
Thank you!

PR #8096, opened 2024-07-18
[Multiple File Upload] Add remove files action

### LINKED ISSUE #7779: [Multiple File Upload] Create addFile and removeFile routes

devin-civi:
### Description

Adds the ability to remove files, both on the frontend and backend.

Includes a kind of messy hack, adding "forceUpdate" to stageAndUpdate. Without it we can never remove the final upload of a required question, since that fails validation. This skips validation when removing files to get around that, we still show an error, and validation still runs before actually submitting. (Not happy with this, but it's the best we came up with on Slack.)

#### General

#### User visible changes

  - [x] Didn't use a message in applicant facing code that isn't translated yet (it's behind a flag)

#### New Features

### Issue(s) this completes

Fixes #7779


devin-civi:
* **#8096** <a href="https://app.graphite.dev/github/pr/civiform/civiform/8096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/services/applicant/ApplicantService.java
```diff
@@ -284,7 +286,13 @@ public CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
 
     return stageAndUpdateIfValid(
-        applicantId, programId, blockId, updateMap, updates, addressServiceAreaValidationEnabled);
+        applicantId,
+        programId,
+        blockId,
+        updateMap,
+        updates,
+        addressServiceAreaValidationEnabled,
+        forceUpdate);
```
jeffpw-goog, line 295: Thought (non-blocking): this PR is quite large and it would have been nice to split it up, for example a separate PR to add support for forcing updates. And maybe frontend vs backend parts or whatever other splits are reasonable.

Actually the more I think about it the more I prefer that you split it up. But I'll leave it up to you.

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -531,6 +532,128 @@ private CompletionStage<Result> editOrReview(
             });
   }
 
+  /**
+   * Used by the file upload question. Removes the specified file and re-renders the question block.
+   */
+  @Secure
+  public CompletionStage<Result> removeFile(
+      Request request, long programId, String blockId, String fileKeyToRemove, boolean inReview) {
+    Optional<Long> applicantId = getApplicantId(request);
```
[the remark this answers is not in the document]
devin-civi: None of the other tests in this route, but the best way to do that seems to be to not use the withId method for testing except when we need to explicitly pass in an ID and rely on this path for the majority of testing, so I did that for the removeFile tests at least.

I tend to shy away from inconsistency like this, but it doesn't really seem worthwhile to change all the other tests either.

browser-test/image_snapshots/file_test/file-error-too-large-multiple-files.png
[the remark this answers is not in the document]
devin-civi: Yeah, per UX for the multiple-file approach we wanted to shrink the size of the file upload question (so remove isn't all the way to the right). The addition of a max width in the renderer code is what's causing this.

server/app/services/applicant/ApplicantService.java
devin-civi: I don't really like splitting the force update change since then you end up submitting code that isn't used by anything else, without context. (Or alternatively, if you submit removeFile first, it's submitting code that doesn't actually work.) 

I don't think splitting the frontend and backend is worth the effort at this point, but agree it should have been split.

jeffpw-goog: I disagree pretty strongly on the first point - mentioning in the PR description that it is part of a multi-PR change, and linking all related PRs in a bug, solves that problem from my perspective.

jeffpw-goog (approved):

PR #8100, opened 2024-07-18
Give admins the ability to add categories to a program when creating a new program

### LINKED ISSUE #7785: [Program Filtering] Allow admins to add categories on programs when creating a new program
Admins will be able to tag a program with categories through the program creation form (the program edit form will come later).  This is similar to changing a program description, for example.  From the admin programs list page, the user will click “Create new program”, and they will then be able to use checkboxes to add one or more categories to the program.

This issue should be done behind the "PROGRAM_FILTERING_ENABLED" feature flag so that we can polish the styles (and use USWDS checkboxes) in another issue.

#### Add screenshot of relevant mocks for this issue here:  (This is for the program edit view, but it should look the same in the program creation view.)

image

#### Additional Context
[TDD ](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit?usp=sharing)

Note:  Since the program edit form uses the same code in `ProgramFormBuilder` to render the checkboxes, we may need to hide the checkboxes when editing.  That would be just until we implement them in the program edit flow.

#### Done when:
- [ ] The list of available categories is shown as checkboxes on the program creation form
- [ ] When categories are selected, they appear on the program in the database

melanie-exygy:
### Description

Allowing admins to tag a program with categories through the program creation form (the program edit form will come later). From the admin programs list page, the user will click “Create new program”, and they will then be able to use checkboxes to add one or more categories to the program. 

This is behind the "PROGRAM_FILTERING_ENABLED" feature flag so that we can polish the styles (and use USWDS checkboxes) in another issue.

Many of the changes in this PR are because of the new categories field on the `ProgramDefinition`.  This meant adding that field everywhere that we create a ProgramDefinition.

Note:  These are not the final styles.

image

1. Go to the Dev Tools.
2. Click "Seed Sample Programs and Categories".
3. Log in as an admin.
4. Go to the Settings page and make sure "PROGRAM_FILTERING_ENABLED" and "PROGRAM_MIGRATION_ENABLED" are both set to true.
5. Go to the Programs List page.
6. Click "Create new program".
7. Make a new program and select a few categories on the program.
8. Save it.
9. Either check the db to verify that associations were created in the `programs_categories` table OR click on "Export" tab, select the new program and click "Generate JSON".  You'll find the categories array at the end of the JSON.

### Issue(s) this completes

Fixes #7785


server/test/controllers/admin/AdminImportControllerTest.java
```diff
@@ -420,6 +420,7 @@ public void saveProgram_preservesUniversalSettingOnImportedQuestions() {
           + "    \"acls\" : {\n"
           + "      \"tiProgramViewAcls\" : [ ]\n"
           + "    },\n"
+          + "    \"categories\" : [ ], \n"
```
gwendolyngoetz, line 423: (note) You don't need to change anything here. Just an FYI in case you didn't know. Now that we are on Java 17 you can use the `"""` to do [multiline strings](https://www.baeldung.com/java-multiline-string).

browser-test/src/support/admin_programs.ts
```diff
@@ -1337,4 +1337,16 @@ export class AdminPrograms {
     )
     expect(sentEmail.Body.text_part).toContain(emailBody)
   }
+
+  async seedProgramsAndCategories() {
```
gwendolyngoetz, line 1341: (suggestion) If you have a specific reason to do this here this is fine, but if not can you swap to the [seedPrograms in index.ts](https://github.com/civiform/civiform/blob/a398fb2a0c4704570f522a399b75197853cbc197/browser-test/src/support/index.ts#L258).

browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -1083,4 +1083,48 @@ test.describe('program creation', () => {
       true,
     )
   })
+
+  test('create program with categories', async ({page, adminPrograms}) => {
+    await enableFeatureFlag(page, 'program_filtering_enabled')
+
+    await test.step('seed categories', async () => {
+      await adminPrograms.seedProgramsAndCategories(page)
```
gwendolyngoetz, line 1091: (note) See commend on the admin_programs.ts file.

server/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -51,14 +56,17 @@ abstract class ProgramFormBuilder extends BaseHtmlView {
   private final SettingsManifest settingsManifest;
   private final String baseUrl;
   private final AccountRepository accountRepository;
+  private final CategoryRepository categoryRepository;
```
gwendolyngoetz, line 59: (question, blocking) The image above shows the category checkboxes sorted, but I didn't see any code that is sorting the list. Is this being sorted case-insensitively outside this PR? If not, can you put a sort on it.

browser-test/image_snapshots/admin_program_creation_test/program-creation-with-categories.png
gwendolyngoetz, line 1: (question, non-blocking) Since this says "Tag with 1 or more" can you add the required asterisk to label. If you plan on doing that with the styling step that works too.

gwendolyngoetz (changes requested):

browser-test/image_snapshots/admin_program_creation_test/program-creation-with-categories.png
melanie-exygy: The field isn't required, even though I see how the wording could make it seem that way.  My understanding is "Tag with 1 or more _if you feel like it_" rather than "You must tag with 1 or more".  There isn't a required asterisk in the mocks:  
image

I'll bring this up with Shreya and Sijia, to double-check and to suggest that we make the wording clearer.

gwendolyngoetz: Sounds good. I definitely took it as a literal command. :)

server/app/views/admin/programs/ProgramFormBuilder.java
melanie-exygy: Added sorting in the `CategoryRepository.listCategories`.  They were only sorted previously because of the order of insertion into the db.

browser-test/src/support/admin_programs.ts
melanie-exygy: Done.  Also, changed the name of `seedPrograms` to `seedProgramsAndCategories`.

gwendolyngoetz (approved):

Issue #7574, opened 2024-05-27
Set a Content-Security-Policy


gwendolyngoetz:
For increased security we should see if we can enable a [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) without breaking the application.

gwendolyngoetz:
@jeffpw-goog [here are Play's docs](https://www.playframework.com/documentation/2.9.x/CspFilter) on this if you haven't seem them yet.

jeffpw-goog:
I [added a section](https://github.com/civiform/civiform/wiki/Frontend-Development/_compare/361836bf0f3b8550106d688f7dcc38467881769d...3bc33aade8937d0bc6835c03dd33165f8d0202b8) on CSP to the [frontend development wiki](https://github.com/civiform/civiform/wiki/Frontend-Development#content-security-policy).

jeffpw-goog:
Per [discussion in Slack](https://civiform.slack.com/archives/C01R8SJA2HE/p1724072405469559), we will proceed with rolling out CSP despite #8143, and pursue the fix for that in parallel.

jeffpw-goog:
CSP is now enabled by default: https://github.com/civiform/civiform/pull/8218
Also cleaned up the staging override: https://github.com/civiform/civiform-staging-deploy/pull/163

This will go out in the release on September 24.

After that the only remaining work is to submit https://github.com/civiform/civiform/pull/8661 to clean up the config variable, which we can do after a few more weeks.

PR #8119, opened 2024-07-22
Replace inline javascript for language selector

jeffpw-goog:
### Description

Replace the inline onchange listener for the language selector dropdown with a static javascript listener in a new language_selector.ts file. This is necessary to enforce a content security policy (https://github.com/civiform/civiform/issues/7574), which disallows inline javascript.

#### General

### DIFF
--- a/server/app/assets/javascripts/applicant_entry_point.ts
+++ b/server/app/assets/javascripts/applicant_entry_point.ts
@@ -4,6 +4,7 @@
  */
 
 import * as main from './main'
+import * as languageSelector from './language_selector'
 import * as enumerator from './enumerator'
 import * as radio from './radio'
 import * as toast from './toast'
@@ -18,6 +19,7 @@ import * as trustedIntermediary from './trusted_intermediary'
 
 window.addEventListener('load', () => {
   main.init()
+  languageSelector.init()
   enumerator.init()
   radio.init()
   toast.init()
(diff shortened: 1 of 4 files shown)


melanie-exygy (commented):
This seems fine to me.  Would you mind adding a browser test?  Maybe a good place would be in `applicant_application_index.test.ts`?

I'm also wondering if there's a way to do this with HTMX.  Is that something you considered?

melanie-exygy:
> This seems fine to me. Would you mind adding a browser test? Maybe a good place would be in `applicant_application_index.test.ts`?
> 
> I'm also wondering if there's a way to do this with HTMX. Is that something you considered?

On second thought, I think we already have a few tests that cover the language selector, so there may not be a need to add another.

jeffpw-goog:
> This seems fine to me. Would you mind adding a browser test? Maybe a good place would be in `applicant_application_index.test.ts`?
> 
> I'm also wondering if there's a way to do this with HTMX. Is that something you considered?

Is there something specific you're looking to cover? Many browser tests break without this fix with CSP turned on. The browser test util method is [here](https://github.com/civiform/civiform/blob/44334a8d7c84c33fa1939c412966094008aa2f2a/browser-test/src/support/index.ts#L239) and it has 15 references.

I did not consider HTMX because I haven't seen it in use and I'm not very familiar... do you know how it would work? Maybe hx-trigger?

melanie-exygy (approved):

PR #8150, opened 2024-07-24
Validate program categories when adding them on the admin program edit form

melanie-exygy:
### Description

This PR adds server-side validation for program categories when they are added to a program through the program create/edit forms.  We check that the categories are part of the existing category options saved in the database.  This follows the validation pattern that is already part of the `ProgramService`.

This is part of the program filtering feature:  [TDD](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit?usp=sharing)


browser-test/src/admin/admin_program_creation.test.ts
```diff
@@ -1125,8 +1128,37 @@ test.describe('program creation', () => {
...
+      await adminPrograms.submitProgramDetailsEdits()
+      await adminPrograms.publishProgram(programName)
+    })
+
+    await test.step('go to program edit form and check that categories are still pre-selected', async () => {
+      await adminPrograms.gotoViewActiveProgramPageAndStartEditing(programName)
+      await page.getByRole('button', {name: 'Edit program details'}).click()
+      await waitForPageJsLoad(page)
+      await expect(
+        page.getByRole('checkbox', {name: 'Education'}),
```
azhangggggg, line 1141: question (non-blocking): can a program have multiple categories?

azhangggggg: if so, is there an upper limit to it?

melanie-exygy: Yes, a program can have multiple categories.  Right now, there isn't an upper limit, but there will only be 10 category options available (as checkboxes).

azhangggggg (approved):

PR #8190, opened 2024-07-29
[Multiple File Upload]  Populates and displays answer data for multi file uploads

devin-civi:
### Description

Adds and uses answer data for multiple file uploads. This only covers renderers/views that display answers, namely the admin application view and the applicant application review view, future PRs will cover other uses (API, PDF export, etc...)

#### General

### Issue(s) this completes


devin-civi:
* **#8190** <a href="https://app.graphite.dev/github/pr/civiform/civiform/8190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

browser-test/src/applicant/application_review.test.ts
```diff
@@ -9,9 +9,392 @@ import {
   testUserDisplayName,
   waitForPageJsLoad,
   validateScreenshot,
+  enableFeatureFlag,
 } from '../support'
 
 test.describe('Program admin review of submitted applications', () => {
+  test('all major steps with multiple file upload flag', async ({
```
melanie-exygy, line 16: **question (non-blocking):** Is your idea, in copying all the major steps, that this test will replace the following test once the feature flag is removed?  For the moment, it seems like a lot of duplication, much of it unnecessary, but I would understand if it's going to make removing the flag easier.

browser-test/src/applicant/questions/file.test.ts
```diff
@@ -336,6 +336,83 @@ test.describe('file upload applicant flow', () => {
...
+      await applicantQuestions.expectQuestionAnsweredOnReviewPage(
+        fileUploadQuestionText,
+        'file1.txt',
+      )
+
+      await applicantQuestions.expectQuestionAnsweredOnReviewPage(
+        fileUploadQuestionText,
+        'file2.txt',
+      )
+      await validateScreenshot(page, 'file-uploaded-review')
```
melanie-exygy, line 364: **suggestion (non-blocking):** Rather than taking a screenshot of the whole page including the header and footer, could you take a screenshot of the relevant section?  That'll prevent less churn when we update other parts of the page.

browser-test/src/support/applicant_questions.ts
```diff
@@ -430,6 +430,23 @@ export class ApplicantQuestions {
     return readFileSync(path, 'utf8')
   }
 
+  async downloadFileFromReviewPage(fileName: string) {
+    // Assert that we're on the review page.
+    expect(await this.page.innerText('h2')).toContain(
```
melanie-exygy, line 435: **suggestion (non-blocking):** Playwright best practices suggest using the `page.getByRole()` locator because it's looking for it from a web first way that is the same as how accessibility tools look at it.  [getByRole](https://playwright.dev/docs/locators#locate-by-role).  In this case, I think you could do: `await expect(page.getByRole('heading', { name: 'Program application summary' })).toBeVisible();`

server/app/services/applicant/AnswerData.java
```diff
@@ -75,6 +76,15 @@ public static Builder builder() {
   /** The identifier of the applicant's uploaded file after its UTF-8 encoded if applicable */
   public abstract Optional<String> encodedFileKey();
 
+  /** Identifier for the applicant's uploaded file, UTF-8 encoded. */
+  public abstract ImmutableList<String> encodedFileKeys();
```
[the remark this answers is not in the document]
devin-civi: Yep, it'll be removed later. I have to keep it for now since the single-file approach (which uses encodedFileKey) renders things differently.

browser-test/src/applicant/questions/file.test.ts
devin-civi: Done

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -204,23 +204,37 @@ private DivTag renderQuestionSummary(
 
     if (data.isAnswered() || haveAnswerText) {
       final ContainerTag answerContent;
-      if (data.encodedFileKey().isPresent()) {
+
+      if (!data.encodedFileKeys().isEmpty()) {
+        answerContent = div();
+        for (int i = 0; i < data.encodedFileKeys().size(); i++) {
+          if (i > 0) {
+            answerContent.with(br());
```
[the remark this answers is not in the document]
melanie-exygy: Rather than replacing `br` with `p`, the idea is to remove the if-statement and then wrap the anchor tag in a paragraph, so that there's naturally a break after each line.  On line 218, it would look like this:  
```
answerContent.with(
  p(
     a(fileName).withHref(fileLink).withClass(BaseStyles.LINK_TEXT)
   )
);
```
As for the other use below, where we replace new lines with the `br`, I think this would be more complicated to change since we have to respect the new lines in the answer.  I'm sure it can be done, but it might not be worth the effort.

browser-test/src/applicant/application_review.test.ts
melanie-exygy: Understood.  I'll put this on our agenda for the next eng sync so that we can discuss a best approach.  I was unsure which approach to take on the feature I'm currently working on too.

server/app/views/applicant/ApplicantProgramSummaryView.java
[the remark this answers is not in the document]
devin-civi: Right, thank you! Thinking more along those lines, this is a list, so I'll use an <ol> + <li>s.

PR #8104, opened 2024-07-19
Add North Star styles to ineligible page

ssandbekkhaug:
### Description

[UX Mocks](https://www.figma.com/design/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?node-id=2543-7139&t=k72DRV08hG8RflVQ-0)

Changes:
* Put content in the center column
* Adjust margins
* Add alert with questions
* Update copy

ineligible_v3

## Release notes
n/a

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [x] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [x] Extended the README / documentation, if necessary

#### User visible changes

- [x] Followed steps to [internationalize](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#application-strings) any new strings
  - [x] Added context strings to new [messages](https://github.com/civiform/civiform/blob/main/server/conf/i18n/messages)
  - [x] Didn't use a message in applicant facing code that isn't translated yet (unless behind a flag)
- [x] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method
- [x] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [x] Manually tested at 200% size
- [x] Manually evaluated tab order

### Issue(s) this completes

Part of #7996 (I want to get UX review before marking "Done")

Issue #7996, opened 2024-07-09
Add styles for "Ineligible" page


ssandbekkhaug:
Blocked on mocks in #7737

ssandbekkhaug:
Blocked. See comment: https://github.com/civiform/civiform/issues/7737#issuecomment-2246240099

ssandbekkhaug:
Sijia said we can continue with the MVP version for parity. Putting this issue back to "in progress". I'll continue it next week.

sijiayam:
Design QA complete. 1 minor issue opened: https://github.com/civiform/civiform/issues/9036

PR #8214, opened 2024-07-31
Prepare new ineligible alert strings for North Star

ssandbekkhaug:
### Description

Adding new ineligible alert strings for North Star. This change was originally a part of #8104 but I wanted to split it up into smaller chunks for ease of review.

## Release notes

n/a

#### General

#### User visible changes

n/a

### Issue(s) this completes

Part of #7996

### DIFF
--- a/server/app/controllers/applicant/EligibilityAlertSettingsCalculator.java
+++ b/server/app/controllers/applicant/EligibilityAlertSettingsCalculator.java
@@ -37,7 +37,12 @@ public final class EligibilityAlertSettingsCalculator {
   }
 
   public AlertSettings calculate(
-      Http.Request request, boolean isTI, boolean isApplicationEligible, long programId) {
+      Http.Request request,
+      boolean isTI,
+      boolean isApplicationEligible,
+      boolean isNorthStarEnabled,
+      boolean pageHasSupplementalInformation,
+      long programId) {
     Messages messages = messagesApi.preferred(request);
 
     boolean isEligibilityGating = isEligibilityGating(programId);
@@ -51,8 +56,16 @@ public final class EligibilityAlertSettingsCalculator {
 
     Triple triple =
         isTI
-            ? getTi(isApplicationFastForwarded, isApplicationEligible)
-            : getApplicant(isApplicationFastForwarded, isApplicationEligible);
+            ? getTi(
+                isApplicationFastForwarded,
+                isApplicationEligible,
+                isNorthStarEnabled,
+                pageHasSupplementalInformation)
+            : getApplicant(
+                isApplicationFastForwarded,
+                isApplicationEligible,
+                isNorthStarEnabled,
+                pageHasSupplementalInformation);
 
     return new AlertSettings(
         isEligibilityGating,
@@ -61,7 +74,11 @@ public final class EligibilityAlertSettingsCalculator {
         triple.alertType);
   }
 
-  private Triple getTi(Boolean isApplicationFastForwarded, Boolean isApplicationEligible) {
+  private Triple getTi(
+      boolean isApplicationFastForwarded,
+      boolean isApplicationEligible,
+      boolean isNorthStarEnabled,
+      boolean pageHasSupplementalInformation) {
     if (isApplicationFastForwarded == true && isApplicationEligible == true) {
       return new Triple(
           AlertType.SUCCESS,
@@ -83,6 +100,13 @@ public final class EligibilityAlertSettingsCalculator {
           MessageKey.ALERT_ELIGIBILITY_TI_ELIGIBLE_TEXT);
     }
 
+    if (isNorthStarEnabled == true && pageHasSupplementalInformation == true) {
+      return new Triple(
+          AlertType.WARNING,
+          MessageKey.ALERT_ELIGIBILITY_TI_NOT_ELIGIBLE_TITLE,
+          MessageKey.ALERT_ELIGIBILITY_TI_NOT_ELIGIBLE_TEXT_SHORT);
+    }
+
     // The default case: isApplicationFastForwarded == false && isApplicationEligible == false
     return new Triple(
         AlertType.WARNING,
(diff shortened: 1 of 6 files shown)


ssandbekkhaug:
* **#8214** <a href="https://app.graphite.dev/github/pr/civiform/civiform/8214?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/controllers/applicant/EligibilityAlertSettingsCalculator.java
```diff
@@ -13,16 +13,19 @@
...
 
 public final class EligibilityAlertSettingsCalculator {
   private final ProgramService programService;
   private final MessagesApi messagesApi;
+  private final SettingsManifest settingsManifest;
 
   @Inject
   public EligibilityAlertSettingsCalculator(
-      ProgramService programService, MessagesApi messagesApi) {
+      ProgramService programService, MessagesApi messagesApi, SettingsManifest settingsManifest) {
```
gwendolyngoetz, line 25: (blocking) Don't inject SettingsManifest here. Put the isNorthstar boolean on the calculate method instead. It will simplify the unit tests to a single matrix of parameters.

server/conf/i18n/messages
```diff
@@ -324,18 +324,21 @@ title.programSummary=Program application summary
 title.applicantNotEligible=Based on your responses to the following questions, you may not qualify for the {0}.
 title.applicantNotEligibleTi=Based on your responses to the following questions, your client may not qualify for the {0}.
 content.eligibilityCriteria=For eligibility criteria please refer to {0}
+# Describes how to learn more about eligibility criteria for a program. The variable text is "program details", which will become a hyperlink to another webpage.
+content.eligibilityCriteria.v2=For eligibility criteria, please refer to {0}.
```
gwendolyngoetz, line 328: (question) Are these strings new within the last couple weeks? The strings that are there were all recently updated already based on Sijia's wordage.

If they have changed again, rather than having a v2 let's just update the existing strings.

gwendolyngoetz (changes requested):

devin-civi (approved):
Changes look good to me, but obviously investigate Gwen's comments first!

server/conf/i18n/messages
gwendolyngoetz: Let's wait for @sijiayam to get back to verify. We may not need to change the text at all.

sijiayam: @gwendolyngoetz @ssandbekkhaug Can either of you help me confirm the current string? The North Star ineligible  strings we are looking at in this scenario is the string we have here:
image

I think the only suggestions I am making are:
- Adding "If you think there is a mistake"
- Adding "You can also" and removing "or"

These are to make the language on the ineligible screen more human. 

@gwendolyngoetz Are you referring to the ineligible/eligible alert message(s)that we display in the alert? Is that a different or same use case as the one above in the screenshot?

server/conf/i18n/messages
```diff
@@ -347,6 +350,8 @@ alert.eligibility_applicant_eligible_text=Based on your responses, you are likel
 alert.eligibility_applicant_not_eligible_title=You may not be eligible for this program
 # Alert text when applicant may not be eligible for a program
 alert.eligibility_applicant_not_eligible_text=Based on your responses, you may not be eligible for this program. If your information has changed, you can edit your answer and proceed with the application.
+# Alert text when applicant may not be eligible for a program (new version for North Star)
+alert.eligibilityApplicantNotEligibleText.v2=Based on your responses to the following questions, you may not be eligible for this program.
```
gwendolyngoetz, line 354: @sijiayam here

server/conf/i18n/messages
```diff
@@ -367,6 +372,8 @@ alert.eligibility_ti_eligible_text=Based on your responses, your client is likel
 alert.eligibility_ti_not_eligible_title=Your client may not be eligible for this program
 # Alert text when TI client may not be eligible for a program
 alert.eligibility_ti_not_eligible_text=Based on your responses, your client may not be eligible for this program. If your client''s information has changed, you can edit your answer and proceed with the application.
+# Alert text when TI client may not be eligible for a program (new version for North Star)
+alert.eligibilityTiNotEligibleText.v2=Based on your responses to the following questions, your client may not be eligible for this program.
```
gwendolyngoetz, line 376: @sijiayam and here

server/conf/i18n/messages
gwendolyngoetz: @sijiayam I tagged you on the specific lines I have questions on. I started this thread just at the top since there were multiple changes.

server/conf/i18n/messages
sijiayam: @ssandbekkhaug Are we removing `If your information has changed, you can edit your answer and proceed with the application.` for the North Star version because this alert will be displayed on the ineligible page with content that says a similar message? 

If so I would suggest we use the new string.

If that is not the case and this is an alert that displays without any other messaging about the ability to change answers and proceed, then we should keep the original string.

ssandbekkhaug: > @ssandbekkhaug Are we removing `If your information has changed, you can edit your answer and proceed with the application.` for the North Star version because this alert will be displayed on the ineligible page with content that says a similar message?

Yes. The page is going to look something like this when it's done:

ineligible

Here's a [link to the mocks too](https://www.figma.com/design/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?node-id=2543-7139&m=dev).

ssandbekkhaug: I should also note, the alert is used in other places, which is what I think Gwen is getting at. Here's an example:

main-summary

I can imagine having `If your information has changed, you can edit your answer and proceed with the application` on the *ineligible* page would be confusing. But having that sentence on the *application review* page seems OK.

Should we have two strings?
1. A string for editable pages, such as the application review page
2. A string for non-editable pages, such as the ineligible page

sijiayam: I do think we should have two strings for situations where we have supplemental information in the page content and situations where we don't.

ssandbekkhaug: Ok, I updated the strings. Can we resolve this thread?

server/conf/i18n/messages
ssandbekkhaug: See comments above

server/conf/i18n/messages
gwendolyngoetz: Sounds good. Just wanted to check since we just recently changed them in the non-ns part. Thanks all!

gwendolyngoetz (approved):

PR #8249, opened 2024-08-05
[Suffix]create namesuffix enum

azhangggggg:
### Description

Added suffix as a enum in ApplicantModel, building some foundation for the rest of changes on name suffix feature. Updated unit tests on ApplicantModel as well.

#### General

### Issue(s) this completes

part of #8267


server/test/models/ApplicantModelTest.java
dkatzz, line 1: **suggestion:** Can we include a test where the suffix is empty and / or null?

server/app/models/ApplicantModel.java
```diff
@@ -38,6 +40,26 @@
...
+  public enum Suffix {
+    JR("Jr."),
+    SR("Sr."),
+    I("I"),
+    II("II"),
+    III("III"),
+    IV("IV"),
+    V("V");
+
+    private final String displayName;
```
dkatzz, line 52: **suggestion (non-blocking):** You could also consider calling this `suffix` or `suffixName`, but good with either.

dkatzz:
Could you remove the release notes section since it doesn't apply in this case?

server/test/models/ApplicantModelTest.java
azhangggggg: yeah good call, test case added!

server/app/models/ApplicantModel.java
azhangggggg: gotcha, changed it to `suffixName`

server/test/models/ApplicantModelTest.java
dkatzz, line 1: **suggestion:** It's probably worth adding a case for suffix being null also.

dkatzz (approved):
One more comment about testing, but otherwise LGTM

server/test/models/ApplicantModelTest.java
azhangggggg: added a test case for when suffix is null

PR #8258, opened 2024-08-06
Update dev dump and restore to be able to dump and restore all tables

nb1701:
This adds a new script for dumping the entire dev database, replacing the old one, and modifies the restore script to simply do the pg_restore, since we are always restoring the entire database and do not need to truncate tables anymore.


bin/dev-restore-db
```diff
@@ -13,15 +13,23 @@ fi
 # set TRUNCATE_SQL to heredoc contents
 TRUNCATE_SQL=$(
   cat <<EOF
-TRUNCATE TABLE files;
-TRUNCATE TABLE applications CASCADE;
-TRUNCATE TABLE applicants CASCADE;
 TRUNCATE TABLE accounts CASCADE;
```
gwendolyngoetz, line 16: (blocking) Since the dump all script exports everything  you'll need to either not export the evolutions tables are truncate them on import.

If would say don't export them and then you don't have to truncate. This way you can restore the database without stopping the application. If someone wants to have no evolutions they can delete the db volume and run fresh.

bin/dev-dump-db
```diff
@@ -0,0 +1,27 @@
...
+readonly DUMPFILE="${1:-dev_db.dump}"
+
+readonly CMD="/usr/bin/pg_dump \
+    --no-password \
+    --format=custom \
+    --host=db \
+    --username=postgres \
+    --clean \
+    --if-exists \
+    postgres --file=/project/${DUMPFILE}"
```
gwendolyngoetz, line 16: (praise) Thank you for using the long argument name options

gwendolyngoetz (approved):

Issue #8267, opened 2024-08-06
[Suffix] Modify the service to take name suffix


azhangggggg:
Add nameSuffix enum in ApplicantModel; 
Make changes to well know path; Update applicantService and trustedIntermediaryService; 
Add tests for all of the files mentioned

PR #8275, opened 2024-08-07
[Suffix] middle layer name question service change

azhangggggg:
### Description

Added suffix to various services regarding name question, which allows name question service to take suffix.  ApplicantData#getApplicantNameSuffix will first try to read it from the ApplicantModel (the PAI column), and then fall back to reading it from the WellKnownPath.

1. Added suffix as one of the WellKnownPaths;
2. Made changes to ApplicantData, NameQuestion, QuestionAnswer, Scalar service to take suffix;
3. Added suffix `Jr.` to QuestionJsonSampler;
4. Unit tests changes.

#### General

### Issue(s) this completes

Part of #8267


server/app/services/applicant/question/Scalar.java
```diff
@@ -37,6 +37,7 @@ public enum Scalar {
   SELECTIONS("selections", ScalarType.LIST_OF_STRINGS),
   STATE("state", ScalarType.STRING),
   STREET("street", ScalarType.STRING),
+  SUFFIX("suffix", ScalarType.STRING),
```
melanie-exygy, line 40: **nitpick:** Same here, my preference would be to add the word "name".  So `NAME_SUFFIX`.

browser-test/src/applicant/application_download.test.ts
```diff
@@ -242,14 +242,14 @@ test.describe('normal application flow', () => {
     await adminPrograms.viewApplications(programName)
     const postEditCsvContent = await adminPrograms.getCsv(noApplyFilters)
     expect(postEditCsvContent).toContain(
-      'sarah,,smith,op2_admin,05/10/2021,1000.00',
+      'sarah,,smith,,op2_admin,05/10/2021,1000.00',
```
melanie-exygy, line 245: **question (non-blocking):** Is it possible to include a suffix in one of these test cases rather than having it be blank in both?

server/app/services/applicant/question/Scalar.java
avaleske: +1. I think when it's clear the `suffix` is part of a name, like in a name question in the API, we don't need to say `name_suffix`, but in lists like this, where it's a lot of concepts mixed together, it's helpful to include `name`.

server/app/services/applicant/question/NameQuestion.java
```diff
@@ -102,6 +103,20 @@ public Optional<String> getLastNameValue() {
...
+  public Optional<String> getSuffixValue() {
+    if (suffixValue != null) {
+      return suffixValue;
+    }
+
+    ApplicantData applicantData = applicantQuestion.getApplicantData();
+    suffixValue = applicantData.readString(getSuffixPath());
+
+    if (middleNameValue.isEmpty() && isPaiQuestion()) {
+      middleNameValue = applicantData.getApplicantNameSuffix();
```
avaleske, line 115: **issue**: Should this be checking and setting the `middleNameValue`, or is it a bug?

server/app/services/question/QuestionAnswerer.java
```diff
@@ -88,11 +88,13 @@ public static void answerNameQuestion(
       Path contextualizedPath,
       String first,
       String middle,
-      String last) {
+      String last,
+      String suffix) {
```
avaleske, line 92: **nit** I think setting the type of the `suffix` parameter to the `SuffixEnum` so it can only accept correct values will reduce the chance for bugs. It could be an `optional` if we need to not always set a suffix. Although since this is test code I could be convinced otherwise, but I think in prod code we should use the enum type everywhere possible.

Similarly, anywhere we take a string that is only ever going to be a Suffix, we should try to use the enum. Here, for example:
https://github.com/civiform/civiform/blob/bad18e641f95f7cbd90a699ba7a259af2ab4be93/server/app/models/ApplicantModel.java#L54

server/test/services/applicant/question/NameQuestionTest.java
```diff
@@ -57,17 +57,23 @@ public void withEmptyApplicantData_optionalQuestion() {
...
     ApplicantQuestion applicantQuestion =
         new ApplicantQuestion(nameQuestionDefinition, applicantData, Optional.empty());
     QuestionAnswerer.answerNameQuestion(
-        applicantData, applicantQuestion.getContextualizedPath(), firstName, middleName, lastName);
+        applicantData,
+        applicantQuestion.getContextualizedPath(),
+        firstName,
+        middleName,
+        lastName,
+        suffix);
```
avaleske, line 76: **issue** I don't see any updates to the asserts to make sure the new suffix field is updated correctly

server/app/services/applicant/question/Scalar.java
avaleske: Changing the scalar to `NAME_SUFFIX` will also end up changing the CSV header, so you'll need to manually set it to `suffix` the way we change the currency header, in https://github.com/civiform/civiform/blob/bad18e641f95f7cbd90a699ba7a259af2ab4be93/server/app/services/export/CsvExporterService.java#L303

avaleske (changes requested):

browser-test/src/applicant/application_download.test.ts
azhangggggg: We will def add some test cases to include suffix when I actually make some frontend changes. It's difficult now, for example methods like below we can't even apply a `.cf-name-suffix` since it's not even created yet. https://github.com/civiform/civiform/blob/fb32344e118fc977609de00fd1deb34e44e7082f/browser-test/src/support/applicant_questions.ts#L60

server/test/services/applicant/question/NameQuestionTest.java
azhangggggg: Ah good catch, added validation to make sure suffix is updated correctly.

server/app/services/question/QuestionAnswerer.java
azhangggggg: I think since it's test code and this method hasn't being used outside of testing, so maybe using string is fine. But will def apply the enum type everywhere possible in prod code.

server/test/services/export/CsvExporterServiceTest.java
```diff
@@ -153,6 +153,7 @@ public void programCsv_noRepeatedEntities() throws Exception {
             "applicant name (first_name)",
             "applicant name (middle_name)",
             "applicant name (last_name)",
+            "applicant name (name_suffix)",
```
avaleske, line 156: Can we keep this as `applicant name (suffix)`? You'll have to re-write it in the same place we re-write the currency from currency_cents to currency_dollars.

https://github.com/civiform/civiform/blob/d59163464e5151cd2a575a9f0cff65f17760ee29/server/app/services/export/CsvExporterService.java#L310-L312

avaleske (approved):
Approving with the one change to the csv export (that we'll also want in the json export)

azhangggggg:
> Approving with the one change to the csv export (that we'll also want in the json export)

change to the csv report is done! will also do that to json export when working on JsonExport service changes.

PR #8280, opened 2024-08-07
Replace inline javascript for retire api key form

jeffpw-goog:
### Description

Replace the "onsubmit" inline javascript for the "retire key" button on the admin API keys page with static javascript. The message requires the key name, which is now plumbed in a data attribute on the form.

This is necessary to enforce a content security policy (https://github.com/civiform/civiform/issues/7574), which disallows inline javascript.

Also:

- Update the retireApiKey helper so that admin_api_keys.test.ts validates that the dialog actually appears
- Fix unused variable warnings that randomly popped up in admin_api_keys.ts

#### General

### DIFF
--- a/server/app/views/admin/apikeys/ApiKeyIndexView.java
+++ b/server/app/views/admin/apikeys/ApiKeyIndexView.java
@@ -148,19 +148,13 @@ public final class ApiKeyIndexView extends BaseHtmlView {
     if (apiKey.isRetired()) {
       statsDiv.with(p("Retired " + dateConverter.formatRfc1123(apiKey.getRetiredTime().get())));
     } else {
-
-      String onSubmitJs =
-          "return confirm('Retiring the API key is permanent and will prevent"
-              + " anyone from being able to call the API with the key. Are you"
-              + " sure you want to retire "
-              + apiKey.getName()
-              + "?')";
       linksDiv.with(
           form(makeCsrfTokenInputTag(request))
               .withAction(controllers.admin.routes.AdminApiKeysController.retire(apiKey.id).url())
               .withMethod("POST")
-              .withOnsubmit(onSubmitJs)
               .withId(String.format("retire-%s", keyNameSlugified))
+              .withData("api-key-name", apiKey.getName())
+              .withClasses("retire-key-form")
               .with(
                   makeSvgTextButton("Retire key", Icons.DELETE)
                       .withClasses(ButtonStyles.OUTLINED_WHITE_WITH_ICON)));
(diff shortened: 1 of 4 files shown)


server/app/views/admin/apikeys/ApiKeyIndexView.java
```diff
@@ -148,19 +148,13 @@ private DivTag renderApiKey(
...
-              + apiKey.getName()
-              + "?')";
       linksDiv.with(
           form(makeCsrfTokenInputTag(request))
               .withAction(controllers.admin.routes.AdminApiKeysController.retire(apiKey.id).url())
               .withMethod("POST")
-              .withOnsubmit(onSubmitJs)
               .withId(String.format("retire-%s", keyNameSlugified))
+              .withData("api-key-name", apiKey.getName())
+              .withClasses("retire-key-form")
```
rockycodes, line 157: question/suggestion (non-blocking): I would prefer for this to be an id since we use it to reference one element, but if there's a reason you prefer a class I'm open to it.

rockycodes (approved):
lgtm! are there any existing tests for this? if not, do you think it needs one? it would probably just be a screenshot of the dialog popup.

server/app/views/admin/apikeys/ApiKeyIndexView.java
jeffpw-goog: Hmm there can be multiple API keys on the page and thus multiple retire key buttons. That's why the ID on line 155 has the key name in it (which actually doesn't have to be unique, but that's off topic).

rockycodes: Oh that makes sense! In the past I've seen people use [reference classes](https://github.com/civiform/civiform/blob/main/server/app/views/style/ReferenceClasses.java) for this so it's clear it's not a class to control styling.

jeffpw-goog:
> lgtm! are there any existing tests for this? if not, do you think it needs one? it would probably just be a screenshot of the dialog popup.

Thanks - admin_api_keys.test.ts is the test for this, but it doesn't actually test that the dialog appears. I was being lazy, but given how easy it is to break this functionality by not executing the javascript, I agree the test should cover it. I've updated it to do so.

server/app/views/admin/apikeys/ApiKeyIndexView.java
jeffpw-goog: Yeah I'm aware of ReferenceClasses but have never been sure when to/not to use it, and I've seen plenty of examples not using it... so I guessed that maybe it was if the string was re-used a lot within Java, since you can't reference it in JS. I think I'll leave it, but maybe we should clarify the best practices for ReferenceClasses.

rockycodes: That works for me! And agreed about clarifying best practices for Reference Classes.

Issue #7813, opened 2024-06-17
[North Star Applicant UI] Allow residents to filter programs by category on the homepage


melanie-exygy:
#### Describe the solution presented in this issue.
Migrate program filtering functionality into Thymeleaf.  This issue is for functionality only, not styling.

#### Add screenshot of relevant mocks for this issue here: 
![Image](https://github.com/user-attachments/assets/06ca970b-5e62-4dd6-8ed6-4979284a27a9)

#### Additional Context
[TDD](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit?usp=sharing)
As of 10/28/24, as agreed upon by UX/Product, we'll be adding two buttons to the filter form, on button to apply filters and one to clear them.  This will mean that we won't have to re-render the page after each filter click.
After applying filters, we should scroll to the top of the "Programs based on your selections" section.

#### Done when:
What criteria need to be met for the issue to be finished? 
- [ ] With the North Star enabled, residents can filter programs by category on the homepage
- [ ] Filtered programs appear in a "Programs based on your selections" section and all others appear in a section entitled "Other programs and services"
- [ ] Both sections show a count next to the header
- [ ] There is no flickering

PR #8282, opened 2024-08-07
Adding read-only category filter chips to the homepage

### LINKED ISSUE #8135: Add program filter chips to the homepage
Add buttons on the homepage that the resident can select to filter programs.  We should only show categories that appear in at least one of the programs displayed to the user.

This will be behind the PROGRAM_FILTERING_ENABLED feature flag.  At this point, the filter chips won't do any filtering.  We will simply display them.

#### Link to Figma mocks: 
[mocks](https://www.figma.com/file/BOES7eVIpEFIbBBShCN8AB/Program-card-discoverability?type=design&node-id=239-11037&mode=design&t=fzLPhaaMXNqOQ6yg-0)

image

#### Additional Context
[TD ](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit?usp=sharing)

#### Done when:
- [ ] There are filter chips on the homepage which display the categories that are in at least one program

melanie-exygy:
### Description

Adds filter chips on the homepage that the resident will eventually be able to select to filter programs. We should only show categories that appear in at least one of the programs displayed to the user.

This will be behind the PROGRAM_FILTERING_ENABLED feature flag. At this point, the filter chips don't do any filtering.   We simply display them and can select them.

The North Star implementation of this is part of a separate issue:  #7813. 

image

1. Go to the Dev Tools.
2. Seed programs and categories.
3. Log in as a CF Admin.
4. Edit a program or two and add some categories to them.
5. Publish the programs.
6. Go to the homepage and see the categories you selected as filter chips.
7. Click on a chip and notice that the styling changes and checkmark appears.
8. Tab to a chip and press the space-bar to select it through keyboard navigation.

### Issue(s) this completes

Fixes #8135


server/tailwind.config.js
```diff
@@ -65,5 +65,22 @@ module.exports = {
       },
     },
   },
-  plugins: [require('@tailwindcss/line-clamp')],
+  plugins: [
+    require('@tailwindcss/line-clamp'),
+
+    // Add a custom plugin to support the `has` pseudo-classes.  For example: "has-checked:bg-blue-300".
```
gwendolyngoetz, line 71: (question) Did you run into issues with the [builtin](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-descendants) `has-*` support? It was added in v3.4.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -192,6 +199,20 @@ private DivTag mainContent(
                 Math.max(relevantPrograms.unapplied().size(), relevantPrograms.submitted().size()),
                 relevantPrograms.inProgress().size()));
 
+    // Find all the categories that are on any of the relevant programs
+    List<String> relevantCategories =
+        relevantPrograms.allPrograms().stream()
```
avaleske, line 204: **question** Is there a specific sort order we want to apply, or do we not care?

avaleske (approved):
Approving pending Gwen's question about tailwind builtins

server/app/views/applicant/ProgramIndexView.java
melanie-exygy: Good question.  I'll check with Product/UX and respond back here when I get an answer.

melanie-exygy: They prefer that it be sorted alphabetically.  I added the sorting.  Thanks for noticing that!

server/tailwind.config.js
melanie-exygy: For some reason, the built-in support isn't working for me when I include it in Java, but I just found out that it works when I put it in `styles.css`, so I did that and removed the plugin.  I know it's not ideal that we're putting it in the stylesheet, but this is all going to get re-done with the North Star, so it's only temporary.

gwendolyngoetz: That's so weird. I wonder if something is escaping the `[` and `]` characters on the java side.

melanie-exygy: Hm, yeah, I wonder.  It would be nice to figure out, but I don't think it's worth the time since we're moving away from J2Html anyway.  Are you ok with me leaving it in the stylesheet?

avaleske (approved):

server/tailwind.config.js
gwendolyngoetz: Yeah yeah. That's plenty fine

gwendolyngoetz: Here's my problem. We might be moving away from j2html, but now I have a mystery that is going to consume me. lolol

PR #8301, opened 2024-08-09
[Multiple File Upload] CSV export for file upload

devin-civi:
### Description

Adds `file_urls` to JSON exports. Note that unlock most other JSON exports, this is not using the scalar (file_key_list) to build the export path, since we don't export keys, we export urls (but we store keys).

Note some of this bahvior is not hidden behind a flag, we are exposing file_urls immediately (see release notes), and then removing file_key when we turn on the feature flag.

## Release notes

In preparation of supporting multiple file uploads for a file upload question, we have added a file_urls property to the JSON API and JSON export. file_urls will contain a list of all uploaded files, and will eventually replace file_key. Since this is a backwards incompatible API change, file upload questions currently will contain both file_key and file_urls. Once the ability to upload multiple files is launched, file_key will be removed. Please contact us with any questions or if this will cause any disruptions.

#### General

### Issue(s) this completes


devin-civi:
* **#8301** <a href="https://app.graphite.dev/github/pr/civiform/civiform/8301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/test/services/export/AbstractExporterTest.java
```diff
@@ -967,6 +967,26 @@ FakeApplicationFiller answerFileUploadQuestion(
       return this;
     }
 
+    FakeApplicationFiller answerFileQuestionWithMultipleUpload(
+        QuestionModel question, String fileKey, int index) {
+      return answerFileQuestionWithMultipleUpload(question, null, fileKey, index);
+    }
+
+    FakeApplicationFiller answerFileQuestionWithMultipleUpload(
+        QuestionModel question, String repeatedEntityName, String fileKey, int index) {
```
avaleske, line 976: *nit, optional*: For other questions that take a list, the answerer method takes a list instead of an item and an index. Does this better matches the way the applicant data is filled in prod?

server/test/services/export/JsonExporterServiceTest.java
```diff
@@ -890,17 +917,46 @@ public void export_whenFileUploadQuestionIsAnswered_valueIsInResponse() {
...
+            true);
+    ResultAsserter resultAsserter = new ResultAsserter(resultJsonString);
+
+    resultAsserter.assertJsonAtApplicationPath(
+        ".applicant_file",
+        """
+{
+  "file_urls" : [ "%s/admin/applicant-files/test-file-key-1", "%s/admin/applicant-files/test-file-key-2" ],
+  "question_type" : "FILE_UPLOAD"
+}"""
```
avaleske, line 930: Did the formatter push this over because the line was so long? Otherwise, can we fix the indentation?

server/test/services/export/JsonExporterServiceTest.java
```diff
@@ -958,18 +1024,96 @@ public void export_whenFileUploadQuestionIsNotAnswered_valueInResponseIsNull() {
...
         {
           "file_key" : null,
+          "file_urls" : [ ],
           "question_type" : "FILE_UPLOAD"
         }""");
   }
 
+  @Test
+  public void
+      export_whenFileUploadQuestionIsAnswered_whenMultipleFileUploadDisabled_fileKeyListEqualsFileKey() {
```
avaleske, line 1043: should we include the `whenMultipleFileUploadDisabled/Enabled` in the other test names? It makes it clearer which scenario is being tested but also makes the names petty long...

browser-test/src/api_docs.test.ts
```diff
@@ -112,6 +112,112 @@ test.describe('Viewing API docs', () => {
     })
   })
 
+  test('Views active API docs with multiple upload', async ({
```
avaleske, line 115: Can we add a TODO and an issue to remove these duplicate tests, and reconcile them with the non-multiple-upload versions, once the feature is launched and the flag is removed?

avaleske (changes requested):
Small change but otherwise LGTM

server/test/services/export/JsonExporterServiceTest.java
devin-civi: Yeah looks like the formatter is doing this

server/test/services/export/JsonExporterServiceTest.java
devin-civi: My thought here (or actually it was a reviewer's thought in a PR I've been trying to continue :) was to rename old tests that are going to be deleted with something like "whenMultipleFileUploadDisabled", but keeping the newer copies (where the feature enabled) of those tests un-changed with the simpler name. Then when you launch the feature, you go and delete any "whenMultipleFileDisabled" named tests, and leave the others unchanged. That way you don't end up with legacy "WhenMultipleFileUpload" names long after that feature isn't relevant since all file uploads are multiple.

I don't have strong feelings either way, but I've mostly been trying to do this for my unit tests since that PR.

devin-civi (commented):
PTAL

server/test/services/export/JsonExporterServiceTest.java
avaleske: Ah makes sense to me!

avaleske (approved):

Issue #8306, opened 2024-08-09
Add "Relevant Programs" section to "Application confirmation" page


ssandbekkhaug:

ssandbekkhaug:
The section has been added. Styling the cards is tracked in #8156 (mocks) and #8157 (implementation)

![Image](https://github.com/user-attachments/assets/c3e3d0f3-a6dd-4e23-91cd-ea3595bab986)

sijiayam:
Closing since the functionality is complete. Styling will be QA'd in #8157

PR #8320, opened 2024-08-12
Set size of program cards in grid

ssandbekkhaug:
### Description

Before: cards expand to 100% of parent's width.
After: cards expand to 33% or 50% of parent's width, depending on page size.

This is a prerequisite to add a grid of cards to the application confirmation page.

I'm intentionally minimizing my time styling the program index (home) page since the mocks aren't ready yet. Expect the program index page to be imperfect for the next few weeks.

#### General

#### User visible changes

### Issue(s) this completes

Supports #8306


ssandbekkhaug:
* **#8320** <a href="https://app.graphite.dev/github/pr/civiform/civiform/8320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
```diff
@@ -206,3 +206,19 @@ body {
 main {
   flex: 1 0 auto;
 }
+
+// Specify size of program cards in a grid
+@media (min-width: 55em) {
+  .tablet-lg\:grid-col-6 {
+    flex: 0 1 auto !important;
+    width: 50% !important;
```
rockycodes, line 214: question (blocking): I'm confused about the need to set the width explicitly. Shouldn't using `grid` handle how many elements fit in a row?

rockycodes (changes requested):
Need more clarification about why the custom uswds styles are needed.

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
ssandbekkhaug: Good point. I rewrote this part, what do you think?

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
```diff
@@ -206,3 +206,23 @@ body {
...
+
+// Desktop: Program card grid has 3 columns
+@media (min-width: 64em) {
+  .cf-program-card-group {
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr;
+  }
+}
+
+.cf-application-card {
```
rockycodes, line 226: question (maybe blocking): I see the reference class being used in existing client facing code. Will these style updates affect those elements at all?

ssandbekkhaug: I don't think so, since this file is only applied to North Star, and none of the old screenshots changed when I ran the browser tests. Updated the name anyway.

ssandbekkhaug:
> Added one more question for you.
> 
> Also, just for my context.... am I correct in understanding that we are trying to phase tailwind out and just rely on uswds classes? And in that case, this is how we would control how many items are in a grid row (since we'll no longer use classes like "grid-cols-3")

Yes, we'd like to remove tailwind eventually.

The number of items in each grid row is defined by setting `class="usa-card-group"` and `grid-template-columns: <something>`.

More info here:
https://designsystem.digital.gov/utilities/layout-grid/

https://designsystem.digital.gov/components/card/

(Feel free to suggest other ways to achieve this layout. I must admit I have very little experience with CSS before this project)

rockycodes (approved):
Lgtm! Thanks for explaining the context with the css work. I think you might need to keep the `.cf-ns-application-card` so tests pass, but since that file is only referenced in north star work and snapshot tests are passing, I think it's fine to use the same class name.

Tbh I am also not a css guru, so if you want another set of expert eyes I would ping Melanie! But this looks fine to me so I'm approving 😄

PR #8336, opened 2024-08-14
Add new job type column and update model

gwendolyngoetz:
### Description

Adding the new `job_type` column to the `persisted_durable_jobs` table, along with related model changes. All existing job code is defaulted the `JobType.RECURRING`.

#### General

#### Database evolutions

- [x] Assigned two reviewers

### DIFF
--- a/server/conf/evolutions/default/74.sql
+++ b/server/conf/evolutions/default/74.sql
@@ -1,0 +1,13 @@
+--- Add job_type column
+
+# --- !Ups
+alter table if exists persisted_durable_jobs
+add column if not exists job_type varchar(32) not null
+default 'RECURRING';
+
+alter table if exists persisted_durable_jobs
+alter column job_type DROP DEFAULT;
+
+# --- !Downs
+alter table if exists persisted_durable_jobs
+drop column if exists job_type;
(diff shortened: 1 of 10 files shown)


server/conf/evolutions/default/74.sql
```diff
@@ -0,0 +1,13 @@
+--- Add job_type column
+
+# --- !Ups
+alter table if exists persisted_durable_jobs
+add column if not exists job_type varchar(32) not null
+default 'RECURRING';
+
+alter table if exists persisted_durable_jobs
+alter column job_type DROP DEFAULT;
```
rockycodes, line 9: question (non-blocking): What does this line do?

ssandbekkhaug (approved):

rockycodes (approved):
this lgtm! just a nitpick and a question.

server/conf/evolutions/default/74.sql
gwendolyngoetz: The first line of the UP adds the new column as not null so you must provide a default value which will be set for all existing rows. The second line removes the default value. It doesn't need to hang around after setting existing rows.

It's cleaner then create the column as nullable, doing an update, and making the column not null.

rockycodes: makes sense. thanks for explaining!

PR #8358, opened 2024-08-17
Adding Startup Job Runner

gwendolyngoetz:
### Description

Adding Startup Job Runner

#### General

### DIFF
--- a/server/app/durablejobs/DurableJobRegistry.java
+++ b/server/app/durablejobs/DurableJobRegistry.java
@@ -3,6 +3,10 @@ package durablejobs;
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
+import java.time.Clock;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.util.HashMap;
 import java.util.Optional;
 import models.JobType;
@@ -12,7 +16,7 @@ import models.PersistedDurableJobModel;
  * Provides the means of looking up a {@link DurableJob} by its {@link DurableJobName}. This is
  * necessary because all {@link DurableJob}s are persisted by {@link PersistedDurableJobModel}
  * records. Also provides the means of retrieving a list of all recurring jobs with their associated
- * {@link RecurringJobExecutionTimeResolver}.
+ * {@link JobExecutionTimeResolver}.
  */
 public final class DurableJobRegistry {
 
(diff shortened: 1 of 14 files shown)


server/app/durablejobs/DurableJobRegistry.java
```diff
@@ -72,30 +75,57 @@ public void registerWithNoTimeResolver(
...
+        jobName.getJobNameString(),
+        RegisteredJob.create(
+            durableJobFactory, jobName, jobType, /* jobExecutionTimeResolver= */ Optional.empty()));
+  }
+
+  /** Registers a factory for a given job name. Cannot be used with {$JobType.RECURRING}. */
+  public void registerStartupJob(
+      DurableJobName jobName, JobType jobType, DurableJobFactory durableJobFactory) {
+    validateJobName(jobName);
+    if (jobType == JobType.RECURRING) {
```
dkatzz, line 88: **suggestion:** Can we use the allowedJobTypes instead here and below? That way we don't need to remember to update this if there becomes more than one recurring job type?

server/app/modules/DurableJobModule.java
```diff
@@ -147,4 +158,23 @@ public DurableJobRegistry provideRecurringDurableJobRegistry(
 
     return durableJobRegistry;
   }
+
+  @Provides
+  @StartupJobsProviderName
+  public DurableJobRegistry provideStartupDurableJobRegistry() {
+    //  PersistedDurableJobRepository persistedDurableJobRepository) {
```
dkatzz, line 165: **suggestion:** Can we put a todo here instead with a bug tracking this work?

server/test/durablejobs/StartupDurableJobRunnerTest.java
```diff
@@ -0,0 +1,213 @@
...
+    PersistedDurableJobModel jobB = createPersistedJobToExecute();
+    PersistedDurableJobModel jobC = createPersistedJobScheduledInFuture();
+
+    recurringDurableJobRunner.runJobs();
+
+    jobA.refresh();
+    jobB.refresh();
+    jobC.refresh();
+
+    // This assertion fails occasionally. I've been unable to figure out why
```
dkatzz, line 133: **question:** Does this comment still apply?

dkatzz (approved):

server/app/modules/DurableJobModule.java
gwendolyngoetz: Typically I would, but I'm going to be adding the first job in the next day for the service area work I'm doing.

server/test/durablejobs/StartupDurableJobRunnerTest.java
gwendolyngoetz: I don't think so. I haven't had it fail at all.

server/app/durablejobs/DurableJobRegistry.java
gwendolyngoetz: Not cleanly without a bunch of refactoring. You try to pass in the scheduler you'll get a circular reference. Passing in just the list of allowed types would let the caller pass anything, at which point why ask for allowed types if you can get around the safety check.

dkatzz: Ah yeah makes sense - I wonder if its worth a comment above the allowedJobTypes for the recurring class that we'll need to update this if that changes?

Issue #8360, opened 2024-08-19
API/Json Export Failing Due to Bad Configuration


gwendolyngoetz:
There is a version of a program that is pointing to a bad question. This is causing the API and Json Exporting to fail.

This is causing the FreshBuck API integration to fail.

```
[1;31mERROR[0;39m o.p.p.f.SecurityFilter - Exception during authentication procedure 
java.util.concurrent.CompletionException: java.lang.RuntimeException: java.lang.IllegalStateException: Question type NULL_QUESTION should not be rendered. Question ID: 2597. Active program question definition is possibly pointing to an old question ID 
at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture.postFire(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(Unknown Source) 
at java.base/java.util.concurrent.CompletableFuture$Completion.run(Unknown Source) 
at play.core.j.ClassLoaderExecutionContext.$anonfun$execute$1(ClassLoaderExecutionContext.scala:64) 
at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:49) 
at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48) 
at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source) 
at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source) 
at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source) 
at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source) 
at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source) 
Caused by: java.lang.RuntimeException: java.lang.IllegalStateException: Question type NULL_QUESTION should not be rendered. Question ID: 2597. Active program question definition is possibly pointing to an old question ID 
[...]

PR #8384, opened 2024-08-21
Add logging and don't set cache when question in the program definition is null

dkatzz:
### Description

The error with the JSON / API export presented (https://github.com/civiform/civiform/issues/8360) itself when a question was null and we think this happened when a question in the program definition cache was null specifically. This change will check if there are any null questions before setting the cache and only set it if no questions are null.

#### General

### Issue(s) this completes

https://github.com/civiform/civiform/issues/8360

### DIFF
--- a/server/app/repository/ProgramRepository.java
+++ b/server/app/repository/ProgramRepository.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Provider;
 import models.AccountModel;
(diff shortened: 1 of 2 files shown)


server/app/repository/ProgramRepository.java
```diff
@@ -161,7 +162,21 @@ public void setFullProgramDefinitionCache(long programId, ProgramDefinition prog
         && getFullProgramDefinitionFromCache(programId).isEmpty()) {
       // We should never set the cache for draft programs.
       if (!versionRepository.get().isDraftProgram(programId)) {
-        programDefCache.set(String.valueOf(programId), programDefinition);
+        ImmutableList<BlockDefinition> blocksWithNullQuestion =
+            programDefinition.blockDefinitions().stream()
+                .filter(BlockDefinition::hasNullQuestion)
+                .collect(ImmutableList.toImmutableList());
+        if (blocksWithNullQuestion.size() > 0) {
+          logger.warn(
```
gwendolyngoetz, line 170: Can you flatten the `blocksWithNullQuestion` and include a csv of the questionIds with the log statement?

gwendolyngoetz (changes requested):

server/app/repository/ProgramRepository.java
dkatzz: Done! Did it within the function for now, but let me know if you think it's worth pulling out into something separate.

gwendolyngoetz (approved):

PR #8398, opened 2024-08-22
[Multiple File Upload] Redirect instead of re-serving page

devin-civi:
### Description

Redirect instead of re-serving after adding/removing files. This is needed because otherwise we end up showing the URL of the re-directed Amazon request in the URL bar, which when refreshed tries to add/remove the file again (which is a no-op, but it's ugly).

Note this does change the behavior slightly, since before we retained error information and now we don't, I think this is OK since the errors will show up when they click "next" (like all the other questions do).

#### General

### Issue(s) this completes

### DIFF
--- a/browser-test/src/applicant/questions/file.test.ts
+++ b/browser-test/src/applicant/questions/file.test.ts
@@ -366,6 +366,8 @@ test.describe('file upload applicant flow', () => {
         0,
       )
 
+      await applicantQuestions.clickNext()
+
       await applicantQuestions.expectRequiredQuestionError(
         '.cf-question-fileupload',
       )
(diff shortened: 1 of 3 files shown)


jeffpw-goog (approved):
LGTM but I know Rocky is thinking about redirect vs not so I'd defer to her opinion.

rockycodes (commented):
I think this is fine, but were you able to manually test it on dev? It's not working for me but it could be something off about my local routes.

devin-civi:
Yep I was able to test it. Note that it requires the multiple files feature flag to be enabled. I'll give it one more test before submitting just to make sure my memory isn't faulty.

Issue #8389, opened 2024-08-21
Help with ESRI setup for Bloomington


dkatzz:
#### Is your feature request related to a problem? Please describe.

From Jeremy in slack:

I've been setting up ESRI service areas. The Municipal Boundary service area works flawlessly.
However, I am stuck getting the Bloomington Economic Zone (BEZ) service area to correctly validate. Addresses I try to validate are ineligible despite being within the zone.
I might be missing something obvious. Any help would be appreciated!
Here's my configuration
    ESRI_FIND_ADDRESS_CANDIDATES_URLS.0: 'https://bloomington.in.gov/arcgis-server/rest/services/Locators/CityCountyAddressLocator/GeocodeServer/findAddressCandidates'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_URLS.0: 'https://bloomington.in.gov/arcgis-server/rest/services/DataPortal/DataPortal_PoliticalBoundaries/MapServer/3/query'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_LABELS.0: 'Municipal_Boundary'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ATTRIBUTES.0: 'muni_name'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_IDS.0: 'Bloomington'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_URLS.1: 'https://bloomington.in.gov/arcgis-server/rest/services/DataPortal/DataPortal_EconomicDevelopment/MapServer/3/query'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_LABELS.1: 'BEZ_West'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_ATTRIBUTES.1: 'objectid'
    ESRI_ADDRESS_SERVICE_AREA_VALIDATION_IDS.1: '5'

Using objectid or objectids did not produce any errors --but did not correctly validate addresses - always noted as ineligible.

Using zone_section does produce this error with HTTP 400: `Invalid service area ID: Downtown Expansion`

ESRI_ADDRESS_SERVICE_AREA_VALIDATION_URLS.1: 'https://bloomington.in.gov/arcgis-server/rest/services/DataPortal/DataPortal_EconomicDevelopment/MapServer/3/query'
  ESRI_ADDRESS_SERVICE_AREA_VALIDATION_LABELS.1: 'BEZ_West'
[...]

Issue #7134, opened 2024-04-04
Improve address correction API fields


avaleske:

avaleske:
@gwendolyngoetz here's the improvements it'd be nice to make to the address correction API fields.

gwendolyngoetz:
Related https://github.com/civiform/civiform/issues/7137

PR #8401, opened 2024-08-22
Address validation outside service area

### LINKED ISSUE #7944: Applicants are submitting ineligible Trees for Neighborhoods applications when their address is outside of Seattle
**Describe the bug**
The trees for neighborhoods program is only eligible to residents of Seattle. This program includes the following eligibility criteria. 

Image

However, two applicants were able to submit applications with addresses outside the service area on 7/1/24 at 2024/07/01 12:33:16 PM PDT and 2024/07/01 03:09:56 PM PDT. 

Image

**Expected behavior**
Applicants with addresses outside the City of Seattle service area should not be able to submit and application to Trees for Neighborhoods.

gwendolyngoetz:
### Description

Expands the ability to check address eligibility outside the service area. The former `service_area` property stored on the applicant/application address answer has been converted from a double delimited string (ex. `a_b_1,d_e_2`) to being represented as a typed array.

For backwards compatibility, the api will flatten the value back to a string. This was already documented in #7134. We need to see if any CEs are currently relying on that.

A job converting existing `service_area` string will come as a separate PR; both will be merged together.

## Release notes

- Eligibility checks can now check if addresses are outside a configured service area.
- Eligibility checks for address service area handle more than just alphanumeric characters

#### General

### Issue(s) this completes

Fixes #7944; #8389


server/app/services/CfJsonDocumentContext.java
```diff
@@ -218,6 +219,31 @@ public void putRepeatedEntities(Path path, ImmutableList<String> entityNames) {
     }
   }
 
+  /**
+   * Puts an array at a given path, building parent objects as needed.
+   *
+   * @param path the {@link Path} where the array should be added.
+   * @param entityNames a {@link List} containing service area results.
+   */
+  public void putServiceAreaInclusionEntities(
```
devin-civi, line 228: Seems like this would live better on ApplicantData since it's referencing scalars specific to that. (And same for read)

server/app/services/CfJsonDocumentContext.java
```diff
@@ -218,6 +219,31 @@ public void putRepeatedEntities(Path path, ImmutableList<String> entityNames) {
...
+      Path path, ImmutableList<ServiceAreaInclusion> entityNames) {
```
devin-civi, line 229: Optional nit: It isn't clear to me why this is called entityNames since it seems to include a lot more.

server/app/services/applicant/ServiceAreaUpdateResolver.java
```diff
@@ -158,15 +178,87 @@ private Boolean doesUpdateContainCorrectedAddress(
...
+
+  /**
+   * Attempts to find the index for a service_area array.
+   *
+   * @param updateMapKey Key from the updateMap in path form. Example:
+   *     `applicant.question_name.service_area[0].service_area_id`.
+   *     --------------------------------------^
+   * @return the int value of the index or -1 if errors/no matches found
+   */
+  private int extractIndexFromPath(String updateMapKey) {
```
devin-civi, line 200: Optional: Did you consider parsing this as a path and using the built in arrayIndex() method on that (and does that work? :) If that works I prefer it because there is less custom code, but it's just a preference.

devin-civi (approved):

server/app/services/applicant/ServiceAreaUpdateResolver.java
gwendolyngoetz: Great catch! Completely spaced on that method

server/app/services/CfJsonDocumentContext.java
```diff
@@ -241,7 +241,7 @@ public void putNull(Path path) {
    *     "$.applicant.children[3].favorite_color.text".
    * @param value the value to place; values of type Map will create the equivalent JSON structure
    */
-  private void put(Path path, Object value) {
+  protected void put(Path path, Object value) {
```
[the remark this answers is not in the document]
gwendolyngoetz: I realized there's already a putArray that is public so I swapped over to that an set put back to private.

I think it makes sense to leave readList as being protected so subclasses can get a typed array without having to add something specific to `CfJsonDocumentContext`.

server/app/services/applicant/ApplicantData.java
```diff
@@ -388,4 +392,42 @@ private Optional<String> getAccountEmail() {
...
+  /**
+   * Puts an array at a given path, building parent objects as needed.
+   *
+   * @param path the {@link Path} where the array should be added.
+   * @param entityNames a {@link List} containing service area results.
+   */
+  public void putServiceAreaInclusionEntities(
+      Path path, ImmutableList<ServiceAreaInclusion> entityNames) {
+    if (entityNames.isEmpty()) {
+      put(path, ImmutableList.of());
```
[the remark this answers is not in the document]
gwendolyngoetz: There's a private putArrayIfMissing on `CfJsonDocumentContext`. I don't know if we want to expose more stuff. I'm fine leaving it as `putArray` with an empty collection. Thoughts?

gwendolyngoetz:
Realized I was incorrectly getting the scalar names since they matched. Fixed that.

PR #8406, opened 2024-08-23
[Suffix] add suffix field to name question

### LINKED ISSUE #8405: [Suffix] suffix frontend change
Make changes to all places referenced nameQuestion for applicants and TI, and do browser/integration tests.
The name question is widely used across CiviForm, there could be a lot of test images updated.

azhangggggg:
### Description

Added name suffix as a field in name question on the frontend. 
Updated unit tests and added browser tests for when `name_suffix` feature flag is enabled.
More browser tests will be written in another PR for the change in other places, like exported pdf etc.

#### General

#### User visible changes

Log in as civiform admin and turn `NAME_SUFFIX_DROPDOWN_ENABLED` feature flag on. Log out and apply to a simply program as a guest user, on answering name question page, you should be seeing name suffix dropdown field.

### Issue(s) this completes

Part of #8405


dkatzz (commented):
More of a larger question, but instead of passing around settingsManifest and request throughout, did you consider injection / was that an option? I also wonder if you considered adding "isSuffixEnabled" on the params object, since its somewhat temporary if that would remove some of the items you have to pass around?

browser-test/src/applicant/questions/name.test.ts
```diff
@@ -217,6 +217,46 @@ test.describe('name applicant flow', () => {
     })
   })
 
+  test.describe(
+    'name question with name suffix flag enabled',
+    {tag: ['@namesuffix']},
```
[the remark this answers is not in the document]
gwendolyngoetz: +1 No need for a tag.

server/app/views/admin/questions/QuestionPreview.java
```diff
@@ -18,22 +20,38 @@
...
+    ApplicantQuestionRendererParams params;
+    if (type == QuestionType.NAME) {
+      params =
+          ApplicantQuestionRendererParams.builder()
+              .setMessages(messages)
+              .setRequest(request)
+              .setErrorDisplayMode(ErrorDisplayMode.HIDE_ERRORS)
+              .build();
+    } else {
+      params =
```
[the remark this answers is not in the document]
azhangggggg: good idea! implemented.

browser-test/src/applicant/questions/name.test.ts
```diff
@@ -217,6 +217,42 @@ test.describe('name applicant flow', () => {
...
+      page,
+      applicantQuestions,
+    }) => {
+      await applicantQuestions.applyProgram(programName)
+
+      await test.step('opens name question during application process', async () => {
+        await validateScreenshot(
+          page.locator('.cf-question-name'),
+          'name-question-with-suffix-field',
+        )
```
gwendolyngoetz, line 243:
```suggestion
        await expect(page.getByLabel('Suffix')).toBeVisible()
        await expect(page.getByLabel('Suffix')).toHaveValue('')
        
        // TODO: I don't recall if you can chain assertions, but if you can I'd do this
        // await expect(page.getByLabel('Suffix')).toBeVisible().toHaveValue('')
```

(recommended) Screenshots are a lot heavier process than asserting other things. Unless you really need screenshots here it's be better using other [locator assertions](https://playwright.dev/docs/api/class-locatorassertions). 

The above suggestion also looks up the control by it's accessible name which is what playwright recommends as that is the same way screen readers would find it.

browser-test/src/applicant/questions/name.test.ts
```diff
@@ -217,6 +217,42 @@ test.describe('name applicant flow', () => {
...
+      })
+
+      await test.step('selects an option in name suffix dropdown', async () => {
+        await applicantQuestions.answerDropdownQuestion('II')
+        await validateScreenshot(
+          page.locator('.cf-question-name'),
+          'name-suffix-with-value-selected',
+        )
```
gwendolyngoetz, line 251: same as other comment

server/app/views/questiontypes/NameQuestionRenderer.java
```diff
@@ -77,6 +81,22 @@ protected DivTag renderInputTags(
                 validationErrors.getOrDefault(nameQuestion.getLastNamePath(), ImmutableSet.of()))
             .addReferenceClass(ReferenceClasses.NAME_LAST);
 
+    SelectWithLabel nameSuffixField =
+        new SelectWithLabel()
+            .addReferenceClass("cf-dropdown-question")
```
gwendolyngoetz, line 86: (blocking) This should be added to the ReferenceClasses.java file

gwendolyngoetz (changes requested):

server/app/views/style/ReferenceClasses.java
```diff
@@ -138,6 +138,7 @@ public final class ReferenceClasses {
   public static final String NAME_FIRST = "cf-name-first";
   public static final String NAME_MIDDLE = "cf-name-middle";
   public static final String NAME_LAST = "cf-name-last";
+  public static final String NAME_SUFFIX = "cf-name-suffix";
```
rockycodes, line 141: question (non-blocking): I don't see this reference class currently being used anywhere. Are you planning to use it in the future? I checked the other reference classes and they are set in `NameQuestionRenderer` eg `.addReferenceClass(ReferenceClasses.NAME_LAST)` and referenced in tests. I think we might actually want to move away from this pattern and use `data-testid` instead so we can use Playwrights `getByTestId`, but that's out of scope for this issue. In any case, I would only include this reference class if you're planning to use it somewhere.

rockycodes (commented):
overall lgtm! just one comment about the reference class.

server/app/views/style/ReferenceClasses.java
azhangggggg: `NAME_SUFFIX` was supposed to get used in NameQuestionRenderer.java. Good catch!

server/app/views/questiontypes/NameQuestionRenderer.java
azhangggggg: replace it with `ReferenceClasses.NAME_SUFFIX` instead. thanks!

server/app/views/style/ReferenceClasses.java
rockycodes: I see it added there, but I'm still not clear why we need the reference class? I don't see it being used in in tests or js files.

gwendolyngoetz: Even if it's not currently used it should be there to be consistent with the other fields in the name renderer.

azhangggggg: I made a new change to create `DROPDOWN_QUESTION = "cf-question-dropdown"` in ReferenceClasses.java since cf-question-dropdown is actually used in tests `applicantQuestions.answerDropdownQuestion()` Sorry about the back and forth but this to me makes most sense!

browser-test/src/applicant/questions/name.test.ts
```diff
@@ -217,6 +217,42 @@ test.describe('name applicant flow', () => {
...
+          await expect(page.getByLabel('Suffix')).toBeVisible()
+          await expect(page.getByLabel('Suffix')).toHaveValue('')
+        })
+
+        await test.step('selects an option in name suffix dropdown', async () => {
+          await applicantQuestions.answerDropdownQuestion('II')
+          await expect(page.getByLabel('Suffix')).toBeVisible()
+          await expect(page.getByLabel('Suffix')).toHaveValue('II')
+        })
+      })
```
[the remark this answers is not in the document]
azhangggggg: yeah no problem! added test to show that suffix is on the review page and also submitted the application without issue :)

dkatzz (approved):

PR #8421, opened 2024-08-27
[Program Migration] Ignore program categories

### LINKED ISSUE #8422: Handle program categories
When an admin exports a program that has categories assigned it, we want to remove the categories to prevent an error on import. Long term, we want categories to be included in the export/import, but this is a quick fix to unblock launch of this feature.

rockycodes:
### Description

Currently, program import will break if the program being imported has categories on it. We want to remove the categories on export to get around this error for now. We will add handling of program categories at a later time.

#### General

1. Make sure the program migration and program filtering feature flags are one
2. Create a program with categories
3. Export it (categories should be an empty array in the json)
4. Attempt to import it
5. You should see the program preview with no errors

### Issue(s) this completes

Fixes #8422

### DIFF
--- a/server/app/controllers/admin/AdminExportController.java
+++ b/server/app/controllers/admin/AdminExportController.java
@@ -75,6 +75,9 @@ public class AdminExportController extends CiviFormController {
       return badRequest(String.format("Program with ID %s could not be found", programId));
     }
 
+    // TODO(#7087) migrate program categories
+    program = program.toBuilder().setCategories(ImmutableList.of()).build();
+
     ImmutableList<QuestionDefinition> questionsUsedByProgram =
         program.getQuestionIdsInProgram().stream()
             .map(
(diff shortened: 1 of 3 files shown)


gwendolyngoetz (approved):

server/app/controllers/admin/AdminExportController.java
```diff
@@ -75,6 +75,9 @@ public Result index(Http.Request request, Long programId) {
       return badRequest(String.format("Program with ID %s could not be found", programId));
     }
 
+    // TODO(#7087) migrate program categories
+    program = program.toBuilder().setCategories(ImmutableList.of()).build();
```
rockycodes, line 79: We want this to run whether or not the filtering feature flag is on in case admins toggled the flag on, added categories, and then toggled it off again.

server/test/support/ProgramBuilder.java
```diff
@@ -295,6 +296,11 @@ public ProgramBuilder withProgramType(ProgramType programType) {
     return this;
   }
 
+  public ProgramBuilder withCategories(ImmutableList<CategoryModel> categories) {
```
melanie-exygy, line 299: **praise:** Thanks for adding this!  It will likely come in handy.

melanie-exygy (approved):

PR #8433, opened 2024-08-28
Remove backwards compatibility for session ID in Play session

### LINKED ISSUE #6113: Store session id in user profile

jeffpw-goog:
### Description

Since March 2024, we have stored the session ID in the pac4j profile instead of the Play session. https://github.com/civiform/civiform/pull/6736 maintained backwards compatibility with existing sessions by continuing to read the session ID from the play session if it was there. Now that it's been 5 months, all sessions from March and earlier should have ended, and we no longer need to do that.

* Stop reading the session ID from the Play session (playWebContext.getNativeSession()) during login profile merge in CiviformOidcProfileCreator
* Stop reading the session ID from the Play session in order to get the id tokens for logout in CiviformOidcLogoutActionBuilder
* Update CiviformOidcProfileCreatorTest to actually test that the ID tokens are saved to the account during login when the enhanced login flag is on

#### General

### Issue(s) this completes

Fixes #6113

### DIFF
--- a/server/test/auth/oidc/CiviformOidcProfileCreatorTest.java
+++ b/server/test/auth/oidc/CiviformOidcProfileCreatorTest.java
@@ -4,7 +4,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static support.FakeRequestBuilder.fakeRequest;
-import static support.FakeRequestBuilder.fakeRequestBuilder;
 
 import auth.CiviFormProfile;
 import auth.CiviFormProfileData;
@@ -37,7 +36,7 @@ public class CiviformOidcProfileCreatorTest extends ResetPostgres {
   private static final String ISSUER = "issuer";
   private static final String SUBJECT = "subject";
   private static final String AUTHORITY_ID = "iss: issuer sub: subject";
-  private static final String SESSION_ID = "session_id";
+  private static final String ID_TOKEN_STRING = "id token string";
 
   private static OidcProfile profile;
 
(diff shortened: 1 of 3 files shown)


server/test/auth/oidc/CiviformOidcProfileCreatorTest.java
```diff
@@ -37,7 +36,7 @@ public class CiviformOidcProfileCreatorTest extends ResetPostgres {
   private static final String ISSUER = "issuer";
   private static final String SUBJECT = "subject";
   private static final String AUTHORITY_ID = "iss: issuer sub: subject";
-  private static final String SESSION_ID = "session_id";
+  private static final String ID_TOKEN_STRING = "id token string";
```
gwendolyngoetz, line 39: (question) Is this just any old value or does it mean something with pac4j?

gwendolyngoetz (approved):

Issue #5971, opened 2023-11-13
Ensure that applicants have profiles via a filter


yotommy:
**Is your feature request related to a problem? Please describe.**
Many, or perhaps all, of our applicant controller methods check for the presence of a user profile, as that is required to execute these methods. See this [example](https://github.com/civiform/civiform/blob/5decb99d8e5975d371e205fba40de3fb6a792bfe/server/app/controllers/applicant/ApplicantProgramReviewController.java#L89-L92). This is error-prone, since anyone writing such a method must remember to include this boilerplate code.

**Describe the solution you'd like**
Create a filter that does this for all applicant controller methods.

**Additional context**
We only want the filter to execute for routes with applicable prefixes.

Here is [another filter](https://github.com/civiform/civiform/blob/main/server/app/filters/SessionIdFilter.java) that can be used as a model. The new filter will also need to redirect in order to have the profile added to the session.

**Done when**
Done when the filter is implemented and configured to run and the boilerplate code is removed from the controller methods.

jeffpw-goog:
#6809 added the filter, so the remaining work is to remove special handling from controller methods ([example](https://github.com/civiform/civiform/blob/5decb99d8e5975d371e205fba40de3fb6a792bfe/server/app/controllers/applicant/ApplicantProgramReviewController.java#L89-L92)).

jeffpw-goog:
This is done! Some pages still handle the case where the profile is missing, but the remaining ones were difficult to untangle and some of them might actually be necessary with https://github.com/civiform/civiform/issues/8414.

PR #8445, opened 2024-08-30
Remove missing-profile redirects

jeffpw-goog:
### Description

#6809 created CiviFormProfileFilter which ensures that all user-facing requests have a CiviFormProfile. So all user-facing controllers should now be able to assume that the profile exists, rather than handling the case where it doesn't.

This PR has two parts:
* Remove the missing-profile handling in HomeController and ProgramSlugHandler, and instead assume a profile is always present since it's guaranteed by the filter. Remove DefaultToGuestRedirector which is then unused.
* Change CiviFormProfileFilter to apply to HEAD requests in addition to the existing GET case. The browser test script [uses curl --head](https://github.com/civiform/civiform/blob/b8091fc6026c937641f95f9c93d59a071cd0b7e8/browser-test/bin/wait_for_server_start_and_run_tests.sh#L40C9-L40C82) on the HomeController to determine if the server is up, so the HEAD request (which Play [secretly converts to a GET request](https://github.com/playframework/playframework/blob/ab50af8cc9339b50da3123942abb772c1d9b8412/core/play/src/main/scala/play/api/http/HttpRequestHandler.scala#L175-L177) so it does in fact call the HomeController.index method we're changing here) needs to be redirected to get a profile, rather than failing without one.

This is for #5971.

#### General


server/app/filters/CiviFormProfileFilter.java
```diff
@@ -45,7 +45,6 @@ private boolean profileIsMissing(Http.RequestHeader requestHeader) {
   private boolean shouldApplyThisFilter(Http.RequestHeader requestHeader) {
     return NonUserRoutePrefixes.noneMatch(requestHeader)
         && !requestHeader.path().startsWith("/callback")
-        && requestHeader.method().equals("GET")
```
dkatzz, line 48: **thought:** Would it be worth adding logging to see when the request header is not GET and the profile is missing? It may make us feel more confident making this change

server/app/controllers/HomeController.java
```diff
@@ -49,15 +48,7 @@ public HomeController(
   }
 
   public CompletionStage<Result> index(Http.Request request) {
-    Optional<CiviFormProfile> maybeProfile = profileUtils.currentUserProfile(request);
-
-    // If the user isn't already logged in within their browser session, consider them a guest.
-    if (maybeProfile.isEmpty()) {
-      return CompletableFuture.completedFuture(createGuestSessionAndRedirect(request));
```
dkatzz, line 56: **thought:** Same here - I wonder if its worth putting in a change to add some logging here to make sure this case doesn't happen then follow up with this change.

jeffpw-goog: I think that might be overkill here - the filter has simple logic to determine whether it runs (is a GET request, is not an excluded path) so a simple test shows that it always runs for this controller. Is there some other risk I'm not thinking about, and/or do you feel strongly that this a metric here is necessary to mitigate possible unknown risks?

server/app/filters/CiviFormProfileFilter.java
jeffpw-goog: Fair, thanks for the feedback! I've changed it to just add HEAD instead of removing the condition to reduce the risk.

server/app/controllers/HomeController.java
jeffpw-goog: For extra context, I'm working on another PR https://github.com/civiform/civiform/pull/8443 that makes this same change in the rest of the user-facing controllers - I separated this one out because of the required changes in the filter.

jeffpw-goog: Actually I changed this PR to include the only two controllers that explicitly have the redirect; so that linked PR (which I might abandon) will actually just address code that made the profile optional when it probably doesn't need to be anymore.

dkatzz: Sounds good, thanks! I can't think of any exact risks with this.

server/app/filters/CiviFormProfileFilter.java
dkatzz: SGTM, thanks!

dkatzz (approved):

PR #8452, opened 2024-08-31
Custom Icon Tag

gwendolyngoetz:
Custom icon tag. See javadoc for usage.

### DIFF
--- a/server/app/views/tags/IconElementTagProcessor.java
+++ b/server/app/views/tags/IconElementTagProcessor.java
@@ -1,0 +1,102 @@
+package views.tags;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static j2html.TagCreator.img;
+
+import controllers.AssetsFinder;
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.thymeleaf.context.ITemplateContext;
+import org.thymeleaf.model.IProcessableElementTag;
+import org.thymeleaf.processor.element.AbstractElementTagProcessor;
+import org.thymeleaf.processor.element.IElementTagStructureHandler;
+import org.thymeleaf.templatemode.TemplateMode;
+import play.Environment;
+
+/**
+ * Thymeleaf custom tag used to render icons.
+ *
+ * <p>Usage examples:
+ * <li>{@code <cf:icon type="some-name" />}
+ * <li>{@code <cf:icon type="some-name" class="..." />}
+ *
+ *     <p>Optionally add {@code xmlns:cf="https://civiform.us/2024/html/icon"} to an html tag to
+ *     prevent editor warnings.
+ */
+public final class IconElementTagProcessor extends AbstractElementTagProcessor {
+  private static final String ASSETS_RELATIVE_PATH = "app/assets";
+  private static final String USWDS_IMAGES_RELATIVE_PATH = "Images/uswds";
+
+  private static final Logger logger = LoggerFactory.getLogger(IconElementTagProcessor.class);
+
+  private final AssetsFinder assetsFinder;
+  private final Environment environment;
+
+  public IconElementTagProcessor(
+      String dialectPrefix, AssetsFinder assetsFinder, Environment environment) {
+    super(TemplateMode.HTML, dialectPrefix, "icon", true, null, false, 10);
+    this.assetsFinder = checkNotNull(assetsFinder);
+    this.environment = checkNotNull(environment);
+  }
+
+  @Override
+  protected void doProcess(
+      ITemplateContext context,
+      IProcessableElementTag tag,
+      IElementTagStructureHandler structureHandler) {
+    try {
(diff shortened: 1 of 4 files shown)


server/app/views/applicant/ProgramIndexTemplate.html
```diff
@@ -44,6 +47,13 @@
         <h1
           th:text="${isGuest}? #{content.saveTimeServices} : #{content.findPrograms}"
         ></h1>
+        <style type="text/css">
+          .custom-red-color {
+            filter: brightness(0) saturate(100%) invert(14%) sepia(97%)
+              saturate(7463%) hue-rotate(2deg) brightness(95%) contrast(115%);
+          }
+        </style>
```
gwendolyngoetz, line 55: This would go in with the css files.

server/app/views/applicant/ProgramIndexTemplate.html
```diff
@@ -44,6 +44,18 @@
...
+        <!-- This would go into the css file -->
+        <style type="text/css">
+          .custom-red-color {
+            filter: brightness(0) saturate(100%) invert(14%) sepia(97%)
+              saturate(7463%) hue-rotate(2deg) brightness(95%) contrast(115%);
+          }
+        </style>
+
+        <!-- Just provide the icon type -->
+        <cf:icon type="https" class="mb-6 h-8 w-8 custom-red-color" />
```
gwendolyngoetz, line 57: This has the tag processor configured to build the image name in the format of `icon-resetofname.svg`. It will handle adding the relative path and dealing with the asset finder. 

It will throw if the image isn't found to prevent trying to using an image that is spelled wrong or doesn't exist.

We could pass the type value as `icon-https`, that is the file name without the extension, but I don't want to leave it open to free form full paths. It's redundant and error prone.

server/app/views/applicant/ProgramIndexTemplate.html
```diff
@@ -29,7 +29,7 @@
 
 */-->
 <!doctype html>
-<html th:attr="lang=${preferredLanguage.code()}">
+<html th:attr="lang=${preferredLanguage.code()}" xmlns:cf="http://tempuri.org/">
```
gwendolyngoetz, line 32: The xmlns is not necessarily required, but I think it will allow editors to not show the tags are warning/errors.

We'd want to replace tempuri.org with something relevant probably on civiform.us.

server/app/views/tags/IconElementTagProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Override
+  protected void doProcess(
+      ITemplateContext context,
+      IProcessableElementTag tag,
+      IElementTagStructureHandler structureHandler) {
+    String html =
+        """
+            <img class="%s" src="%s" role="img" alt="" aria-hidden="true" />
+        """;
```
gwendolyngoetz, line 30: This template should be converted to j2html.

While we're moving away from the UI being written with it I don't see a need to remove it completely in target situations. Long term if we get to a point where this is the only thing that uses it we can rewrite it then.

server/app/views/tags/IconElementTagProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+    String type = tag.getAttributeValue("type");
+    String cssClass = tag.getAttributeValue("class");
+    String path = String.format("Images/uswds/icon-%s.svg", type);
+
+    if (Files.notExists(Paths.get("app/assets", path))) {
+      throw new RuntimeException(String.format("Can't find file: %s", path));
```
gwendolyngoetz, line 37: Probably make this throw FileNotFoundException, despite it being a checked exception. 

However, I think I like the idea of creating an empty svg that can be a placeholder if the requested image isn't found and a very large error being logged that the file is missing.

On one hand I want to error hard early so devs can fix the mistakes, but on the other I don't want a missing image to break a production system.

server/test/views/tags/IconElementTagProcessorTest.java
```diff
@@ -0,0 +1,88 @@
...
+            "cf", instanceOf(AssetsFinder.class), instanceOf(Environment.class));
+    when(mockTag.getAttributeValue("type")).thenReturn("icon-https");
+    when(mockTag.hasAttribute("class")).thenReturn(true);
+    when(mockTag.getAttributeValue("class")).thenReturn("aaa bbb ccc");
+
+    tagProcessor.doProcess(mockContext, mockTag, mockStructureHandler);
+
+    String expected =
+        """
+<img src="/assets/Images/uswds/[a-zA-Z0-9]+-icon-https.svg" aria-hidden="true" role="img" class="aaa bbb ccc" alt="">
```
ssandbekkhaug, line 44: Non-blocking Question: Why is [a-zA-Z0-9]+ part of the filepath?

ssandbekkhaug (approved):

server/test/views/tags/IconElementTagProcessorTest.java
gwendolyngoetz: A fingerprint value is added to the path by the asset system. That regex accounts for it.

PR #8469, opened 2024-09-03
Rename ProfileUtils methods

jeffpw-goog:
### Description

Rename ProfileUtils methods:

* currentUserProfile to optionalCurrentUserProfile, because it returns an Optional
* currentUserProfileOrThrow to currentUserProfile, because since #5971 the profile should always be present for user-facing controllers, and thus this method should be sufficient for most use cases, and the throwing is an edge case rather than expected

Also move currentUserProfile above the two versions of optionalCurrentUserProfile so they are next to each other.

#### General

### DIFF
--- a/server/app/actions/ApplicantAuthorizationAction.java
+++ b/server/app/actions/ApplicantAuthorizationAction.java
@@ -75,7 +75,10 @@ public class ApplicantAuthorizationAction extends Action.Simple {
 
   /** Check if the profile is authorized to access the applicant's data. */
   private CompletionStage<Void> checkApplicantAuthorization(Request request, long applicantId) {
-    return profileUtils.currentUserProfile(request).orElseThrow().checkAuthorization(applicantId);
+    return profileUtils
+        .optionalCurrentUserProfile(request)
+        .orElseThrow()
+        .checkAuthorization(applicantId);
   }
 
   /** Checks that the profile is authorized to access the specified program. */
@@ -85,7 +88,10 @@ public class ApplicantAuthorizationAction extends Action.Simple {
         .thenAccept(
             (isDraftProgram) -> {
               if (isDraftProgram
-                  && !profileUtils.currentUserProfile(request).orElseThrow().isCiviFormAdmin()) {
+                  && !profileUtils
+                      .optionalCurrentUserProfile(request)
+                      .orElseThrow()
+                      .isCiviFormAdmin()) {
                 throw new SecurityException();
               }
             });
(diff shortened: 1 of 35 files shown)


server/app/actions/ApplicantAuthorizationAction.java
```diff
@@ -75,7 +75,10 @@ private long getApplicantId(Request request, RouteExtractor routeExtractor) {
 
   /** Check if the profile is authorized to access the applicant's data. */
   private CompletionStage<Void> checkApplicantAuthorization(Request request, long applicantId) {
-    return profileUtils.currentUserProfile(request).orElseThrow().checkAuthorization(applicantId);
+    return profileUtils
+        .maybeCurrentUserProfile(request)
```
gwendolyngoetz, line 79: (blocking) Please use `optional` instead of `maybe`

gwendolyngoetz (changes requested):

server/app/actions/ApplicantAuthorizationAction.java
jeffpw-goog: Happy to make the change but not sure why - is there published guidance for this? There is tons of "maybe" in the codebase, and of the existing methods that return Optional there appears to be [one that starts with "maybe"](https://github.com/civiform/civiform/blob/50d86c656c8c2c92304b337388c59c7ef65e562d/server/app/services/LocalizedStrings.java#L179) and none that start with "optional" (based on code search for regex 'public Optional<[^>]+> maybe' and 'public Optional<[^>]+> optional').

gwendolyngoetz: It's somewhere on the wiki. We decided a while ago to use `optional` and `optionally`. Maybe is legacy naming.

jeffpw-goog: Updated.

server/app/actions/ApplicantAuthorizationAction.java
```diff
@@ -75,7 +75,10 @@ private long getApplicantId(Request request, RouteExtractor routeExtractor) {
 
   /** Check if the profile is authorized to access the applicant's data. */
   private CompletionStage<Void> checkApplicantAuthorization(Request request, long applicantId) {
-    return profileUtils.currentUserProfile(request).orElseThrow().checkAuthorization(applicantId);
+    return profileUtils
+        .optionalCurrentUserProfile(request)
```
dkatzz, line 79: **suggestion:** Should we replace this with `currentUserProfile(request)` and that will do the `orElseThrow` within the function? Here and in other cases below?

jeffpw-goog: Yes, working on that in the next PR :)

dkatzz (approved):

gwendolyngoetz (approved):

PR #8495, opened 2024-09-04
[Suffix] remove suffix as an eligibility condition

### LINKED ISSUE #8494: [Suffix] remove name suffix as an eligibility condition
Name suffix needs to be removed as an eligibility condition.

azhangggggg:
### Description

Removed suffix as an eligibility condition and added browser tests.

#### General

#### User visible changes

### Issue(s) this completes

Fixes #8494


server/app/views/admin/programs/ProgramPredicateConfigureView.java
```diff
@@ -523,9 +523,9 @@ private DivTag createScalarDropdown(
             .withText("Sorry, you cannot create a show/hide predicate with this question type.");
       }
     }
-
     ImmutableList<OptionTag> options =
         scalars.stream()
+            .filter(scalar -> !scalar.equals(Scalar.NAME_SUFFIX))
```
dkatzz, line 528: **suggestion:** Instead of doing this here, could we add a condition above for the name question where we do the `if (questionDefinition.isAddress())` - that way all the special cases are in one place? So it would be something like `if (questionDefinition.isAddress())` `else if (questionDefinition.isName()`..

azhangggggg: sure! added a condition above for the name question to filter out name_suffix before hand.

server/app/views/admin/programs/ProgramPredicateConfigureView.java
```diff
@@ -524,8 +525,23 @@ private DivTag createScalarDropdown(
       }
     }
 
+    if (questionDefinition.isName()) {
```
dkatzz, line 528: **suggestion:** I meant we could put this above where the address scalar change is happening. So it would be below this code:

```
  if (questionDefinition.isAddress()) {
      scalars = ImmutableSet.of(Scalar.SERVICE_AREA);
      maybeSelectedScalar = Optional.of(Scalar.SERVICE_AREA);
    }
```

I don't think we need to add the isName() function, unless we were using it elsewhere. Instead we could do 

```
if (questionDefinition.getQuestionType().equals(QuestionType.NAME)) {
  scalars = ImmutableSet.of(FIRST, MIDDLE, LAST);
}
```

Does that work or was there a reason it was done this way?

server/app/views/admin/programs/ProgramPredicateConfigureView.java
```diff
@@ -524,8 +525,23 @@ private DivTag createScalarDropdown(
...
+                .collect(toImmutableSet());
+      } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
+        return div()
+            .withText("Sorry, you cannot create a show/hide predicate with this question type.");
+      }
+    }
+
     ImmutableList<OptionTag> options =
         scalars.stream()
+            .filter(scalar -> !questionDefinition.isName() || !scalar.equals(Scalar.NAME_SUFFIX))
```
dkatzz, line 544: **suggestion:** When we make the proposed change, we wouldn't need this anymore.

server/app/views/admin/programs/ProgramPredicateConfigureView.java
melanie-exygy, line 528: **suggestion (blocking):** It seems like this if-condition would fit better on line 519.  So, could you do this?  
```
if (questionDefinition.isAddress()) {
      scalars = ImmutableSet.of(Scalar.SERVICE_AREA);
      maybeSelectedScalar = Optional.of(Scalar.SERVICE_AREA);
    } else {
      try {
        scalars = Scalar.getScalars(questionDefinition.getQuestionType());
        if (questionDefinition.isName()) {
            // Filter out NAME_SUFFIX
            scalars =
                scalars.stream()
                    .filter(scalar -> !scalar.equals(Scalar.NAME_SUFFIX))
                    .collect(toImmutableSet());
        }
        maybeSelectedScalar =
            assertLeafOperationNode(maybeLeafNode).map(LeafOperationExpressionNode::scalar);
      } catch (InvalidQuestionTypeException | UnsupportedQuestionTypeException e) {
        // This should never happen since we filter out Enumerator questions before this point.
        return div()
            .withText("Sorry, you cannot create a show/hide predicate with this question type.");
      }
    }

File: server/app/views/admin/programs/ProgramPredicateConfigureView.java [hunk 4]
```diff
@@ -515,6 +515,9 @@ private DivTag createScalarDropdown(
     } else {
       try {
         scalars = Scalar.getScalars(questionDefinition.getQuestionType());
+        if (questionDefinition.getQuestionType().equals(QuestionType.NAME)) {
```
dkatzz, line 518: **suggestion:** This can be after the `if (questionDefinition.isAddress()) {` before the `else`. With the current implementation we're making a call to get the scalars and not using them. If we need `maybeSelectedScalar` that could be defined within the if as well, but I'm not sure we need that?

azhangggggg: I think this should stay inside of `else` to ensure that `maybeSelectedScalar` is initialized. Unless you see any benefit taking it out of `else`?

dkatzz: Ah got it makes sense. It was just a little weird to me to have scalars predefined then rewritten so we could consider putting scalars (the line above the if) into an else within this else, but fine either way. Makes sense that we need to define maybeSelectedScalar

server/app/views/admin/programs/ProgramPredicateConfigureView.java
```diff
@@ -515,6 +515,9 @@ private DivTag createScalarDropdown(
...
+          scalars = ImmutableSet.of(Scalar.FIRST_NAME, Scalar.MIDDLE_NAME, Scalar.LAST_NAME);
```
dkatzz, line 519: **nitpick:** Can we add a comment that says something like "Suffix is not included in predicates"

dkatzz (approved):

melanie-exygy (approved):

PR #8505, opened 2024-09-05
Remove the networkidle wait on test user logins

gwendolyngoetz:
### Description

The `loginAsTestUserFakeOidc` was using `networkidle` on `waitForURL` calls. Playwright recommends avoiding the use of network idle. This PR removes it for the default of `load`. It cuts around `1000ms` from the call.

> 'networkidle' - DISCOURAGED consider operation to be finished when there are no network connections for at least 500 ms. Don't use this method for testing, rely on web assertions to assess readiness instead.

This will only affect the dev and ci. Going to let this run for a bit to make sure there no issues and look at doing the same to the other test user login methods.

#### General

### DIFF
--- a/browser-test/src/support/index.ts
+++ b/browser-test/src/support/index.ts
@@ -185,10 +185,8 @@ async function loginAsTestUserFakeOidc(
   loginButton: string,
   isTi: boolean,
 ) {
-  await Promise.all([
-    page.waitForURL('**/interaction/*', {waitUntil: 'networkidle'}),
-    page.click(loginButton),
-  ])
+  await page.click(loginButton)
+  await page.waitForURL('**/interaction/*')
 
   // If the user has previously signed in to the provider, a prompt is shown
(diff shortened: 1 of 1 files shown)


browser-test/src/support/index.ts
```diff
@@ -185,10 +185,8 @@ async function loginAsTestUserFakeOidc(
   loginButton: string,
   isTi: boolean,
 ) {
-  await Promise.all([
-    page.waitForURL('**/interaction/*', {waitUntil: 'networkidle'}),
-    page.click(loginButton),
-  ])
+  await page.click(loginButton)
+  await page.waitForURL('**/interaction/*')
```
gwendolyngoetz, line 189: >[!NOTE]
> Promise.all calls removed to follow the pattern from [the Playwright documentation](https://playwright.dev/docs/api/class-page#page-wait-for-url).

jeffpw-goog (approved):
LGTM!

Issue #8504, opened 2024-09-04
Extend CiviFormProfileFilter to all http methods


jeffpw-goog:
Since #5971, [CiviFormProfileFilter](https://github.com/civiform/civiform/blob/b8396a83255cad2ef341b863fdb5e6349ac3f406/server/app/filters/CiviFormProfileFilter.java) ensures that all user-facing requests have a pac4j profile by redirecting to create a guest profile if not. This means that all controllers can assume a profile exists and don't have to include their own handling/redirects for a missing profile.

Currently the filter [only applies to GET and HEAD requests](https://github.com/civiform/civiform/blob/a0c7ed3052b1bd8fd6078f612aa8294f650e8c92/server/app/filters/CiviFormProfileFilter.java#L48). This means that POST requests in controllers technically can't assume that the profile exists, though most will throw an exception if it doesn't.

CiviFormProfileFilter works by redirecting to CallbackController with REDIRECT_TO_SESSION_KEY set to the original request URI, so it's a transparent redirect for GET requests. For POST requests, we can't preserve the POST body (nor would we necessarily want to even try to apply that POST action to a brand new session).

Instead of leaving each controller to handle missing profiles for non-GET/HEAD requests, we can extend CiviFormProfileFilter to handle them. Instead of redirecting back to the original URL in this case, we could just redirect to home like we do in some existing cases (e.g. [ApplicantProgramReviewController](https://github.com/civiform/civiform/blob/b8396a83255cad2ef341b863fdb5e6349ac3f406/server/app/controllers/applicant/ApplicantProgramReviewController.java#L266-L271)), or return a 400 bad request like we do in other cases (e.g. [TrustedIntermediaryController](https://github.com/civiform/civiform/blob/36c688295376be637e622859151cb4a0c8f81d39/server/app/controllers/ti/TrustedIntermediaryController.java#L86-L89)).

PR #8510, opened 2024-09-05
Refactor CiviFormProfileFilter and update javadoc

jeffpw-goog:
### Description

Update CiviFormProfileFilter:
* Move the javadoc that explains when this filter applies to the shouldRedirect method, since that's where the logic is, so hopefully they stay in better sync
* Update that javadoc to reflect changes I made in #8445
* Rename shouldApplyThisFilter to shouldRedirect since redirecting is what the filter does
* Inline the profileIsMissing method because it's pretty readable and not reused
* Refactor the apply method to swap the order so it's more intuitive
* Add a TODO for #8504

#### General

### DIFF
--- a/server/app/filters/CiviFormProfileFilter.java
+++ b/server/app/filters/CiviFormProfileFilter.java
@@ -17,17 +17,8 @@ import play.mvc.Http;
 import play.mvc.Result;
 
 /**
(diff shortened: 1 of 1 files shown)


rockycodes (approved):
lgtm!

Issue #8516, opened 2024-09-05
Support French language in CiviForm


dkatzz:
#### Is your feature request related to a problem? Please describe.

Charlotte has gotten requirements to add French as a language in CiviForm. This means we'll need to get French translations for the general CiviForm content.

#### Describe the solution you'd like.

French language is supported in CiviForm

#### What priority is this?

P1 

#### Associated civic entity

Charlotte

dkatzz:
Sent out a translation request for all content in French.

dkatzz:
https://github.com/civiform/civiform/pull/8777/files

PR #8517, opened 2024-09-05
Add back context strings to messages file

dkatzz:
### Description

A while ago, we removed the context strings from the messages file for a different process that had those context strings in Transifex. However, our process now involves keeping those context strings in. We're soon going to be requesting translations for French, so this will allow us to upload this entire file into our translation request, now and in the future.

In order to do this, I used the last full translation we got for Laothian 
([messages.properties.txt](https://github.com/user-attachments/files/16900124/messages.properties.txt)) as a starting point and then included anything else we were missing from a diff between the current state and that state.

There shouldn't be a diff in the content - just additional context strings present.

#### General

### Issue(s) this completes

Related to https://github.com/civiform/civiform/issues/8516


server/conf/i18n/messages
```diff
@@ -647,13 +765,16 @@ validation.numberNonInteger=Number must be a positive whole number and can only
 # TEXT QUESTION - text specific to filling out a question with words. #
 #---------------------------------------------------------------------#
 
+# Validation errors that appear if a user enters an answer that is too long or too short.
 validation.textTooLong=Must contain at most {0} characters.
 validation.textTooShort=Must contain at least {0} characters.
 
+
```
avaleske, line 772: _very_ nit but extra line?

server/conf/i18n/messages
```diff
@@ -564,11 +672,14 @@ validation.impossibleDob=Date of Birth should be less than 150 years ago
 # ENUMERATION QUESTION - text when viewing a question where a user must name entities, such as children. #
 #--------------------------------------------------------------------------------------------------------#
 
+# Text on a button an applicant would click to add another entity
```
avaleske, line 675: Here and elsewhere, we don't say what the `{0}` value is. Was that the way the original context string was?

avaleske (approved):

server/conf/i18n/messages
dkatzz: Updated, good catch :)

PR #8520, opened 2024-09-05
Adding text area and markup support for question translation view

### LINKED ISSUE #8515: Translation field to include all text formatting features.

There are probably two issues here. Issue 1) When entering translations, the text box is too small and does not scale. 

Here is an image of the text field for entering translations followed by the scalable text field that is available when building new questions. 

Image

Image

Increasing the size of the text box will make it easier to review translations as they are pasted into the system to verify they have been entered correctly. 

The second issues is that translated text does not use all of the same text formatting features. I have verified that translated text can utilize hyperlinks and **bold**. I have verified that it does not maintain carriage return or bulleted lists. 

Here is an image of formatted text in English followed by that same text in Amharic. 

Image

Image

Text box for translation should be larger or scalable or both. Translated text should include text formatting features like return carriage and bulleted lists. 

#### What priority is this?

P1

#### Associated civic entity

Seattle

#### Done when:

swatkat1:
### Description

Adding text area and markup support for question translation view

## Release notes

Adding text area and markup support for question translation view

#### General

#### User visible changes

### Issue(s) this completes

Fixes #8515


browser-test/image_snapshots/admin_question_translation_test/multi-option-question-translation.png
melanie-exygy, line 1: **question (non-blocking):** It seems redundant to put "Markdown is supported see how it works" under every field.  Would you mind checking with product or UX to make sure that's what we want?

melanie-exygy (commented):
This seems like really useful functionality for admins.  It would be nice to add a test that the markdown feature works.  Maybe in `applicant_question_translation.test.ts`?

melanie-exygy (approved):

gwendolyngoetz:
This doesn't need to be done in this PR, but I'm almost certain the grey _English Text_ elements don't render formatted. It would probably be good to make it render correctly if it isn't.

rockycodes (changes requested):
I took a look at this since it involves markdown which I worked on. I'm not sure this fix will fully address the second issue Miguel described since the markdown indicator doesn't control markdown being applied. I could be missing something though. Would you add a test that shows a screenshot of translated text with the markdown fully applied? Specifically translated text that maintains lists and line breaks.

browser-test/image_snapshots/admin_question_translation_test/multi-option-question-translation.png
swatkat1: I will leave it for now. I have reached out to Elise and Miguel on it, if they say otherwise, I will remove it.

swatkat1:
@rockycodes  - I tried it in statging, line breaks and lists do not work.
image

The same content I copied over in Amharic
image

May be your local github did not fully switch over after my changes.

rockycodes:
Ah I see what happened. Changing it to TextArea allows for line breaks while input does not. I tested it on your branch with line breaks and then switched to main, but it was already saved in the db with line breaks so it rendered correctly. It makes sense to me now why this fixes it! I'm keeping my block for now so that it doesn't auto-merge before you've added the screenshot test, but this lgtm.

server/app/views/admin/questions/QuestionTranslationView.java
```diff
@@ -170,11 +176,14 @@ private Optional<DomContent> multiOptionQuestionFields(
       QuestionOption option = options.get(optionIdx);
       optionFieldsBuilder.add(
           fieldWithDefaultLocaleTextHint(
-              FieldWithLabel.input()
+              FieldWithLabel.textArea()
```
rockycodes, line 179: suggestion (maybe blocking): I think we should actually keep the options field as an input and say "Some markdown is supported" since it's an input field in the question editor: https://github.com/civiform/civiform/blob/2af469ff7d1408f1cb08b0da36b5c08c9d16ca3d/server/app/views/admin/questions/QuestionConfig.java#L258

image

swatkat1: I have updated the code and removed markdown on from it.

rockycodes (approved):
Lgtm! It's worth noting that some markdown works for question options (basically anything but new lines and lists) so you could add the indicator there, but it's non-blocking.

PR #8526, opened 2024-09-06
[Program Migration] Handle nested enumerators

### LINKED ISSUE #8508: Error is thrown when trying to use the program migration import for programs with nested enumerators
**Describe the bug**
When trying to import a program that has a nested enumerator, the import fails with this error:
 
Screenshot 2024-09-05 at 10 52 20 AM

**To Reproduce**

1. Turn on the program migration flag
2. Click on import program as an admin
3. Import the JSON in the attached file: [orignal_business_grant_app.json](https://github.com/user-attachments/files/16896619/orignal_business_grant_app.json)

Note that some of the questions get imported.

**Expected behavior**
Program should get imported without issues

rockycodes:
### Description

The original implementation of program migration did not account for the case where there might be nested enumerator questions. This PR fixes that by saving all of the questions up front and then going back to update the questions that have enumerator ids so they reference the correct id of the newly saved parent question.

#### General

1. Turn on the program migration flag
2. Create a program in staging that includes nested enumerator questions
3. Export it, copy the json, and import it into your local environment
4. The program should successfully save

### Issue(s) this completes

Fixes #8508


dkatzz (approved):
Amazing!! Thanks for the quick fix on this!

PR #8586, opened 2024-09-12
[Program Migration] Add program and question validation and remove ti acls

### LINKED ISSUE #8546: [Program Migration] Validate program admin names before saving
When admins migrate programs they might need to edit the program admin name to prevent naming collisions. We want to make sure that the admin name they provide meets our naming criteria (all lower case, no special characters, words separated by hyphens).

rockycodes:
### Description

This change uses our existing program and question validation functions to validate that admins are not attempting to submit program or question values that are not allowed when creating/editing a program or question.

This change also removes any trusted intermediary acls from the program since we are not migrating trusted intermediary groups with the program. Since an assigned TI group is required for the `DisplayMode.SELECT_TI` visibility, we will also show an error if an admin tries to import a program with visibility set to `DisplayMode.SELECT_TI`.

Small docs update: https://github.com/civiform/docs/pull/508

#### General

1. Enable the PROGRAM_MIGRATION_ENABLED feature flag
2. Attempt to submit program json that contains an invalid program admin name, a blank question admin name, or any of the other cases checked by the validate methods.
3. See error
4. Attempt to submit a program with visibility set to "Visible to selected trusted intermediaries only" aka `DisplayMode.SELECT_TI`
5. See error

### Issue(s) this completes

Fixes #8546


browser-test/src/admin/admin_program_migration.test.ts
```diff
@@ -175,15 +177,16 @@ test.describe('program migration', () => {
       )
     })
 
+    await seedProgramsAndCategories(page)
+    await page.goto('/')
+    await adminPrograms.goToExportProgramPage(
+      'Comprehensive Sample Program',
+      'DRAFT',
+    )
+    let downloadedComprehensiveProgram =
```
rockycodes, line 186: I moved this out of the `test.step` so we could access `downloadedComprehensiveProgram` in multiple test steps.

server/test/services/migration/ProgramMigrationServiceTest.java
```diff
@@ -202,6 +207,35 @@ public void prepForExport_clearsNotificationPreferences() {
...
+    ProgramDefinition program =
+        ProgramBuilder.newActiveProgram().withAcls(programAcls).build().getProgramDefinition();
+
+    ProgramDefinition output = service.prepForExport(program);
+
+    assertThat(output.acls().getTiProgramViewAcls()).isEmpty();
+  }
+
+  @Test
+  public void prepForExport_clearsCategories() {
```
rockycodes, line 223: We weren't testing this case previously.

server/app/controllers/admin/AdminImportController.java
```diff
@@ -145,6 +152,41 @@ public Result hxImportProgram(Http.Request request) {
...
+            program.localizedName().getDefault(),
+            program.localizedDescription().getDefault(),
+            program.externalLink(),
+            program.displayMode().getValue(),
+            notificationPreferences,
+            ImmutableList.of(), // categories are not migrated
+            ImmutableList.of() // associated TI groups are not migrated
+            );
+    if (!programErrors.isEmpty()) {
+      String errorString = joinErrors(programErrors).replace("URL", "admin name");
```
rockycodes, line 183: When submitting a new or edited program through the program UI the errors reference "URL" which makes sense. But here saying "URL" would be confusing since admin will need to update the program admin name, not the url, to fix errors.

server/test/services/migration/ProgramMigrationServiceTest.java
```diff
@@ -202,6 +207,37 @@ public void prepForExport_clearsNotificationPreferences() {
     assertThat(output.notificationPreferences()).isEmpty();
   }
 
+  @Test
+  public void prepForExport_clearsTiGroupAcls() {
```
dkatzz, line 211: **thought:** Do we have bugs to allow for this or categories to be supported? If so, could be worth adding a TODO to these (and to the error code above).

server/app/controllers/admin/AdminImportController.java
dkatzz: It may be worth adding this (or something like it) as a comment above the code as well :)

dkatzz (approved):
LGTM, just a couple of nonblocking comments!

server/test/services/migration/ProgramMigrationServiceTest.java
rockycodes: Great idea. I created [this issue](https://github.com/civiform/civiform/issues/8613) and added a TODO in as a comment.

server/app/controllers/admin/AdminImportController.java
rockycodes: I was thinking the same thing! Done.

PR #8591, opened 2024-09-12
Run the prod image locally with bin/run-dev --prod

nb1701:
Because the prod image is built prettty differently from the dev image, we sometimes run into bugs that are only present on the prod image and we don't see them when running the civiform-dev container with the full source volume mounted into the container.

This provides a mechanism for using the civiform prod image with the dev environment. It won't recompile the code when things change like the dev image does, but it will allow us to poke at the prod image. It will run in staging mode so that we still have access to dev tools.

#### General

### DIFF
--- a/bin/run-dev
+++ b/bin/run-dev
@@ -1,9 +1,16 @@
 #! /usr/bin/env bash
 
 # DOC: Run the app locally using Docker. Useful for manual testing.
+# DOC: When providing the --prod flag, it uses the production image instead
+# DOC: of the dev image. This is useful for testing bugs that only show
+# DOC: up in staging/prod. Note that in this mode, it will *not* recompile
+# DOC: the app when you make code changes. It will use the version of the code
+# DOC: compiled when the prod image was built. Also, because we only build
+# DOC: prod for linux/amd64, this will run in emulation on M1 Macs.
 
 source bin/lib.sh
 docker::set_project_name_dev
+truth::declare_var_false runprod
 
 # Default to using Localstack emulator.
 emulators::set_localstack_emulator_vars
@@ -31,6 +38,9 @@ function set_args() {
         emulators::ensure_only_one_cloud_provider_flag aws
         # Already defaulted to AWS.
         ;;
+      "--prod")
+        truth::enable runprod
+        ;;
     esac
 
     shift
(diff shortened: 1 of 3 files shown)


docker-compose.prod-dev.yml
```diff
@@ -0,0 +1,21 @@
+# Builds on docker-compose.yml
+# Runs the prod image in like dev mode, but as if it was deployed to staging.
+
+services:
+  dev-oidc:
+    ports:
+      - 3390:3390
```
gwendolyngoetz: I don't think this is necessary since it should be coming from `docker-compose.dev.yml`.

bin/run-dev
```diff
@@ -1,9 +1,16 @@
 #! /usr/bin/env bash
 
 # DOC: Run the app locally using Docker. Useful for manual testing.
+# When providing the --prod flag, it uses the production image instead
+# of the dev image. This is useful for testing bugs that only show
+# up in staging/prod. Note that in this mode, it will *not* recompile
+# the app when you make code changes. It will use the version of the code
+# compiled when the prod image was built. Also, because we only build
+# prod for linux/amd64, this will run in emulation on M1 Macs.
```
gwendolyngoetz, line 9: (blocking) Each line should have the `DOC:` part so `bin/help` works correctly.

Probably needs double checking, but I don't believe debugging is available either.

bin/run-dev
```diff
@@ -65,6 +79,9 @@ done &
...
 echo "Civiform starting locally at http://localhost:9000"
 
-docker::dev_and_test_server_sbt_command "8457" runDevServer
-
-docker::compose_dev stop civiform
+if truth::is_enabled runprod; then
+  docker::tail_prod_civiform_log
+else
+  docker::dev_and_test_server_sbt_command "8457" runDevServer
+  docker::compose_dev stop civiform
```
gwendolyngoetz, line 86: (question) Was there a reason not to include the stop option?

docker-compose.prod-dev.yml
```diff
@@ -0,0 +1,12 @@
+# Builds on docker-compose.yml and docker-compose.dev.yml
+# Runs the prod image in like dev mode, but as if it was deployed to staging.
+
+services:
+  civiform:
+    build: prod
+    image: civiform/civiform
```
gwendolyngoetz, line 7:
```suggestion
    image: civiform
```

You should use the local `civiform` name. It's consistent with the other compose files and won't risk accidentally pulling a new latest when you don't want it.

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

Issue #8194, opened 2024-07-29
[TDD] Create a TDD to add Application Note to CSV export


swatkat1:
#### Is your feature request related to a problem? Please describe.

For the issue #7366, please create a TDD. 

#### Describe the solution you'd like.

 Adding the application note is not a trivial solution since the data is stored in the application_events table and for every application to be exported, we would do a lookup of the latest application note event to extract the note raising the time complexity to O(n^2) queries which would eventually fail the export and cause SQL timeout error. We need a better design on how to implement this.

#### What priority is this?

P2

#### Done when:

TDD is ready for development.

swatkat1:
Not proceeding with a TDD at this time. I have provided a detailed explanation of approach in the PR.

swatkat1:
Work is done.

PR #8600, opened 2024-09-13
Application note addition in CSV exports

### LINKED ISSUE #7366: Add Application_Note is in .csv download

More program staff are considering CiviForm as a better option for reviewing applications and managing the enrollment process. This includes more detailed use of status updates and the use of admin notes. Program staff have requested that admin notes be include in the .csv download. Currently the application status shows up in the .csv download file. including Admin note would help program staff work through the specific business process necessary to complete the enrollment process without having to go back and forth between the .csv file and the program admin view in CiviForm. 

text in the admin note shows up as a column with data in the .csv download file. 

#### What priority is this?

P2

#### Associated civic entity

Seattle

#### User Group

Program Feedback

#### Done when:

When program admins can see their admin notes in the .csv download file.

swatkat1:
### Description

This PR does a few things

1. It adds a new column "latest_note" to the applications table so that all application notes will be strored in a standardized way (similar to the way we store latest_status) and will performant while querying for application notes in CSV exports.
2. It looks at historical data in the application_events table and updates the latest_note for every application which has a note associated with it. 
3. On every note update, it saves the latest note to the application.

This PR doesn't add more triggers to the way latest_status are populated because

- Triggers functions work best on tables if there is a one trigger for updates, 1 trigger for delete and 1 trigger for creation. Since a trigger already exists for updates on Application table and Application_events table, adding one more will be a more maintenance in the future. 
- The main reason why a trigger was selected to update the latest_status  via trigger was this-
"This approach would ensure that anytime we write / update an ApplicationEvent record, the corresponding Application.latest_status column would be updated as well. Additionally, we would want to prevent Application.latest_status from being updated by application code. This is achievable, but will be pretty easy for bugs to emerge over time if this constraint isn't properly enforced. As far as I can tell, Play framework doesn't provide any primitives for this type of thing." 
https://github.com/civiform/civiform/issues/3269#issuecomment-1232193163
- With notes update happening at only one place in the code, I personally don't see why the applications code cannot handle writing latest_note to the db. Hence this solution is proposed.  

## Release notes

Application note addition in CSV exports. 

#### General

#### Database evolutions

- [x] Assigned two reviewers

#### User visible changes

### Issue(s) this completes

Fixes #7366


server/conf/evolutions/default/76.sql
```diff
@@ -0,0 +1,30 @@
...
+    (
+        SELECT
+            application_id,
+            details->'note_event'->>'note' AS latest_note,
+            ROW_NUMBER() OVER (PARTITION BY application_id ORDER BY create_time DESC) AS row_number
+        FROM
+            application_events
+        WHERE
+            details->>'event_type' = 'NOTE_CHANGE'
+    ) b
```
avaleske, line 22: For readability can we name this something more descriptive than 'b'?

server/conf/evolutions/default/76.sql
```diff
@@ -0,0 +1,30 @@
...
+WHERE
+    b.row_number = 1
+) AS c WHERE applications.id = c.application_id;
+
+# --- !Downs
+
+alter table applications drop column if exists latest_note;
```
avaleske, line 29: **styleguide** sql keywords should be uppercased

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -227,6 +227,9 @@ public void setNote(ApplicationModel application, NoteEvent note, AccountModel a
     ApplicationEventModel event =
         new ApplicationEventModel(application, Optional.of(admin), details);
     eventRepository.insertSync(event);
+    // save the latest note on the applications table too
+    application.setLatestNote(note.note());
+    application.save();
   }
```
avaleske, line 233: Two thoughts:
- This needs to be done in the same transaction as the application event or they could get out of sync if there's a DB error partway through, or a race condition between two users.
- I'd actually move the whole `setNote` method into the `ApplicationEventsRepository` and have that be the only way to set notes on an application.

server/app/models/ApplicationModel.java
```diff
@@ -177,6 +187,15 @@ void setLatestStatusForTest(String latestStatus) {
...
   }
 
+  /**
+   * This is visible only for tests to manipulate the latest note directly in order to ensure that
+   * updates to it are overridden by the configured database trigger.
+   */
+  // @VisibleForTesting
+  public void setLatestNote(String applicationNote) {
+    this.latestNote = applicationNote;
+  }
```
avaleske, line 197: I'd remove this, or make it only `@VisibleForTesting`, and enforce that the only way to set the latest note is via a method on the `ApplicationEventsRepository` within the same transaction as the `ApplicantEvent` is created.

avaleske (commented):
Left some early comments:
- I agree that using DB triggers isn't ideal, but then we need to be very careful about enforcing consistency from application code:
  - All the changes to an application need to be in the same transaction to avoid race conditions and DB errors between the two writes
  - I'd move all writing of notes to the `ApplicantionEventsRepository` and then remove the `setLatestNote` method on the Application so no one uses it. (If possible, since you might need it accessible to do the Update if an update query proves difficult to construct.)

Also, this change in general introduces some interesting behavior to the CSV:
- Exporting the CSV for the same records could be different, if a note was added or not. Do we need a "last modified" column? Should filtering to a specific day now return applications that have notes that have been _updated_ in that day, even if they weren't submitted?

server/conf/evolutions/default/76.sql
swatkat1: Sure!

server/app/services/applications/ProgramAdminApplicationService.java
swatkat1: done!

server/app/models/ApplicationModel.java
```diff
@@ -168,6 +169,15 @@ public Optional<String> getLatestStatus() {
     return Optional.ofNullable(latestStatus);
   }
 
+  /**
+   * Returns the latest application note value associated with the application.
+   *
+   * <p>This value is updated by Porgram Admins who want to add more details to the application.
```
dkatzz, line 175: **nitpick:** Spelling Porgram -> Program

server/conf/evolutions/default/76.sql
dkatzz, line 1: **question:** In a recent tech meeting, we discussed moving towards background jobs instead of evolutions for data migrations. Was that considered here?

server/test/repository/ApplicationEventRepositoryTest.java
```diff
@@ -149,4 +149,33 @@ public void getEvents() {
...
+    AccountModel actor = resourceCreator.insertAccount();
+    ApplicantModel applicant = resourceCreator.insertApplicantWithAccount();
+    ApplicationModel application = resourceCreator.insertActiveApplication(applicant, program);
+    application.setLatestNote("initial note");
+    application.save();
+
+    repo.setNote(application, ApplicationEventDetails.NoteEvent.create("some note"), actor);
+
+    // Execute
+    ImmutableList<ApplicationEventModel> gotEvents =
```
dkatzz, line 167: **nitpick:** Thoughts on naming this applicationEvents?

browser-test/src/admin/admin_application_statuses.test.ts
```diff
@@ -344,6 +344,14 @@ test.describe('view program statuses', () => {
...
       expect(await adminPrograms.getNoteContent()).toBe(noteText)
     })
+    test('allow notes to be exported', async ({adminPrograms}) => {
+      await adminPrograms.editNote('Note is exported')
+      await adminPrograms.expectNoteUpdatedToast()
+      const noApplyFilters = false
+
+      const csvContent = await adminPrograms.getCsv(noApplyFilters)
+      expect(csvContent).toContain('Note is exported')
+    })
```
dkatzz, line 355: **suggestion:** I think it's worth adding a test here too "note gets overwritten" or something like that, which sets the note again and we see the new note instead of the original one.

dkatzz (commented):
Is product okay with just the latest note showing on the CSV download (I didn't see https://github.com/civiform/civiform/issues/7366 specify)? Also, was a TDD written for this? I see a TDD tracking issue [here](https://github.com/civiform/civiform/issues/8194), so wanted to check.

server/conf/evolutions/default/76.sql
swatkat1: Thank you for bringing that up. While background jobs are a valuable tool for managing long running or recurring tasks, for a one-time SQL update query, they might be considered overkill. We'll carefully evaluate the specific use case and potential benefits before making a decision. Are there any particular concerns or advantages you see in using a background job for this specific migration?

server/test/repository/ApplicationEventRepositoryTest.java
swatkat1: sure will change it

server/app/models/ApplicationModel.java
swatkat1: good spotting!!

browser-test/src/admin/admin_application_statuses.test.ts
swatkat1: sure. moved it to a new test

swatkat1:
Hi @dkatzz - 

Following up on the issue comment-https://github.com/civiform/civiform/issues/7366) initially requested including the program's admin note. I implemented this by adding the admin_name column.

However, after merging the code, it was clarified that the desired note is the Application note, not the admin note. I've already discussed this with Elise, Matt, and Miguel, and we're on the same page. To avoid breaking existing CSV exports, the Application note will be added as the last column. This ensures a smooth transition while incorporating the requested change.

Please let me know if you have any questions or require further clarification."

browser-test/src/admin/admin_application_statuses.test.ts
dkatzz: I guess I meant to confirm that it overwrites the old note. So maybe adding and expectation that the csv does not contain 'Some note content', which was the original note

server/conf/evolutions/default/76.sql
dkatzz: We now have the option to run a one-time background job on start up, which would be a good candidate here. The benefit to migrating data that way is that it can be fully tested, etc. Have you been able to test the cases where a few notes already exists for an application and then the most recent note gets migrated and exported correctly after this evolution runs? I'd be okay to make an exception, but if we're not sure we've fully tested all the use cases, I'd prefer us create the column in the evolution and do the data migration in a one time job with unit tests.

server/test/repository/ApplicationEventRepositoryTest.java
```diff
@@ -149,4 +149,63 @@ public void getEvents() {
...
+
+    repo.setNote(application, ApplicationEventDetails.NoteEvent.create("some note"), actor);
+
+    // Execute
+    ImmutableList<ApplicationEventModel> applicationEvents =
+        repo.getEventsOrderByCreateTimeDesc(application.id);
+
+    // Verify
+    assertThat(applicationEvents).hasSize(1);
+    ApplicationEventModel gotEvent = applicationEvents.get(0);
```
dkatzz, line 170: **nitpick:** Maybe `gotEvent` here and below can be called `firstAppEvent` or something a bit clearer?

dkatzz:
> Hi @dkatzz -
> 
> Following up on the issue comment-#7366 initially requested including the program's admin note. I implemented this by adding the admin_name column.
> 
> However, after merging the code, it was clarified that the desired note is the Application note, not the admin note. I've already discussed this with Elise, Matt, and Miguel, and we're on the same page. To avoid breaking existing CSV exports, the Application note will be added as the last column. This ensures a smooth transition while incorporating the requested change.
> 
> Please let me know if you have any questions or require further clarification."

Makes sense! Can you confirm with them that we only care about the most recent application note? For example, if an application has 3 notes, we'd only show the most recent one. I don't see that clarified anywhere in the bug, so just want to confirm we're building this out the right way for what product wants and don't have to change anything later to add all application notes.

server/conf/evolutions/default/76.sql
swatkat1: yes, I have tested this case.

swatkat1:
Can you confirm with them that we only care about the most recent application note? - > yes, as only the most recent one will be shown to the program admin to be updated. Say an application got a note- "needs i20 doc" on 09/11. Got updated to "needs salary doc" on 09/15. Then a note "Approved without conditions" on 09/20. From 09/20, only "Approved without conditions" will be associated with the application that the program admin sees.

browser-test/src/admin/admin_application_statuses.test.ts
swatkat1: oh, in the browser tests? I have added it in the unit tests. no problem adding here too.

server/conf/evolutions/default/76.sql
swatkat1: for the application_id=1, we have the following note change event- 
image
Now I ran the same update query with a where clause- id=1
image
The actual update query had this
image

server/app/models/ApplicationModel.java
```diff
@@ -177,6 +187,15 @@ void setLatestStatusForTest(String latestStatus) {
     this.latestStatus = latestStatus;
   }
 
+  /**
+   * The latest note is never directly set. It is only set using update query from the
+   * ApplicationEventRepository.
+   */
+  @VisibleForTesting
+  public void setLatestNote(String applicationNote) {
```
avaleske, line 195: **suggestion, blocking**: Two questions:
- Can we change this to something like `setLatestNoteForTest()` to make it clear it shouldn't be used?
- Does ErrorProne yell at your if you try to use this if you try to use it outside a test?

server/app/repository/ApplicationEventRepository.java
```diff
@@ -62,4 +67,27 @@ public ImmutableList<ApplicationEventModel> getEventsOrderByCreateTimeDesc(Long
                 queryProfileLocationBuilder.create("getEventsOrderByCreateTimeDesc"))
             .findList());
   }
+
+  public void setNote(
```
avaleske, line 71: **suggestion, non-blocking**: I'd suggest naming this something like `insertNoteEvent()` or similar, so it's clear it's actually inserting a note into the db and not just setting a note on some object that will be inserted later.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -62,4 +67,27 @@ public ImmutableList<ApplicationEventModel> getEventsOrderByCreateTimeDesc(Long
...
+      ApplicationModel application, ApplicationEventDetails.NoteEvent note, AccountModel admin) {
+    ApplicationEventDetails details =
+        services.application.ApplicationEventDetails.builder()
+            .setEventType(ApplicationEventDetails.Type.NOTE_CHANGE)
+            .setNoteEvent(note)
+            .build();
+    ApplicationEventModel event =
+        new ApplicationEventModel(application, Optional.of(admin), details);
+    try (Transaction transaction = database.beginTransaction(TxIsolation.SERIALIZABLE)) {
```
avaleske, line 80: **question**: Overall this is much less likely to have issues keeping the tables in sync, thank you!

There is one additional race condition that we may actually not care about, but it's this:
- What if someone submits a new application between when the `applicantModel` passed into this class was retrieved, and when this note is set? This transaction will now prevent the other issue of a note not getting copied to an application, but it doesn't prevent a note getting set on what is no longer the newest application.
- Is that actually a problem though? I think it was a possibility even with the old code, so it's probably not a problem, right?

avaleske (commented):
I haven't finished reviewing the tests but since I'm mostly out friday I wanted to share comments so far.

server/conf/evolutions/default/76.sql
dkatzz: Got it, thanks!! Seems ok to me, but including @nb1701 and @gwendolyngoetz, since this is a bit of a departure from what we recently discussed.

server/app/models/ApplicationModel.java
dkatzz: Relatedly, should this be marked as `private` instead of `public`?

server/conf/evolutions/default/76.sql
gwendolyngoetz: It's not updating a complex json blob. We don't need a one off job for something simple like this. The query could be simplified with a `distinct on` instead of double nesting and the row_number, but as long as it's working I'm fine with it.

server/conf/evolutions/default/76.sql
gwendolyngoetz: @avaleske  We haven't adopted a style guide for sql

server/app/models/ApplicationModel.java
swatkat1: Good catch! it should be protected. Same as the setLatestStatus() above it in the application model

swatkat1: Since it is not visible outside the class being a protected member, none of the repositories or services can access it

server/app/repository/ApplicationEventRepository.java
swatkat1: it doesn't prevent a note getting set on what is no longer the newest application. --> Imo, I don't think it is a problem. In the system today, we have notes and statuses associated with old applications. If a new application gets submitted, the programs admins will use the note feature to mark it as "duplicate" until we can find a solution to merge two applications and highlight the difference.

server/app/models/ApplicationModel.java
swatkat1: Also @avaleske, leaving it as protected, ErrorProne behaves well :)

dkatzz (approved):

avaleske (approved):
A few naming things and a question about errorprone, but otherwise lgtm!

server/test/models/ApplicationModelTest.java
```diff
@@ -125,4 +129,36 @@ public void isAdmin_programAdmin_isTrue() {
     ApplicationModel application = resourceCreator.insertActiveApplication(applicant, program);
     assertThat(application.getIsAdmin()).isTrue();
   }
+
+  @Test
+  public void test_oldNoteOverwritten() {
```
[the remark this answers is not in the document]
swatkat1: I initially had it there. But due to its protected status, unless it is in ApplicationsModelTest, I don't see this setLatestNote() method.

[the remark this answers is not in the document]
avaleske: Ah hmm, then maybe we remove the setLatestNote() method entirely, and set the note manually in the test with a query like we do in the repository?

Then this test can be in the right place and we don't even need the setLatestNote() method at all.

server/test/repository/ApplicationEventRepositoryTest.java
```diff
@@ -149,4 +149,31 @@ public void getEvents() {
     // NoteEvent wasn't set and should be available as an empty Optional.
     assertThat(gotEvent.getDetails().noteEvent()).isNotPresent();
   }
+
+  @Test
+  public void test_setNoteEvent() {
```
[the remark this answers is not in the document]
swatkat1: done :)

server/app/models/ApplicationModel.java
```diff
@@ -177,6 +187,15 @@ void setLatestStatusForTest(String latestStatus) {
     this.latestStatus = latestStatus;
   }
 
+  /**
+   * The latest note is never directly set. It is only set using update query from the
+   * ApplicationEventRepository.
+   */
+  @VisibleForTesting
+  void setLatestNoteForTest(String applicationNote) {
```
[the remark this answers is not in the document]
swatkat1: i removed this method per your suggestion :)

server/test/models/ApplicationModelTest.java
```diff
@@ -22,10 +24,12 @@ public class ApplicationModelTest extends ResetPostgres {
           .setLocalizedStatusText(LocalizedStrings.withDefaultValue("Approved"))
           .build();
   ApplicationStatusesRepository applicationStatusesRepository;
+  ApplicationEventRepository repo;
```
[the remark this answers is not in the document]
swatkat1: I moved this test back to [ApplicationEventRepositoryTest.java, as it makes more sense there.

server/test/models/ApplicationModelTest.java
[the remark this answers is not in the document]
swatkat1: Good idea! Used database.update() to set and rewrite the note.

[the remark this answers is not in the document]
swatkat1: no doing this instead-  ```
repo.insertNoteEvent(application, ApplicationEventDetails.NoteEvent.create("initial note"), actor);

    assertThat(application.getLatestNote()).isNotEmpty();
    assertThat(application.getLatestNote().get()).isEqualTo("initial note");

    repo.insertNoteEvent(application, ApplicationEventDetails.NoteEvent.create("new note"), actor);
```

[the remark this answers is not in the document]
swatkat1: reverting back to sql update() statement. It was missing an applictaion.refresh();

server/app/models/ApplicationModel.java
swatkat1: I have removed this method.

PR #8612, opened 2024-09-16
Fast-forward deep link bug for guest

gwendolyngoetz:
### Description

If a new guest account tries to load a deep link with an old program id it was hitting an error since the guest account may not have yet had an draft application created. This will just update drafts if found and stop erroring if one isn't found.

#### General

### DIFF
--- a/server/app/repository/ApplicationRepository.java
+++ b/server/app/repository/ApplicationRepository.java
@@ -323,7 +323,7 @@ public final class ApplicationRepository {
   }
 
   /**
-   * Updates an application to point to a new program
+   * Updates a draft application, if one exists, to point to a new program
    *
    * @param applicantId the applicant ID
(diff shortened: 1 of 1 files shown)


server/app/repository/ApplicationRepository.java
```diff
@@ -353,17 +353,11 @@ public void updateDraftApplicationProgram(long applicantId, long programId) {
                   queryProfileLocationBuilder.create("updateDraftApplicationProgram"))
               .findOne();
 
-      if (existingDraft == null) {
```
jcma-google, line 356: should we still handle the null case?

gwendolyngoetz: No. If it's null there's no match so there's no need to do the update at all.

jcma-google (approved):

PR #8615, opened 2024-09-16
Fix format on play.http.session.maxAge setting example

jeffpw-goog:
### Description

Fix the commented-out play.http.session.maxAge setting to have a value that actually works so future readers don't get confused like I did.

Relevant docs: https://www.playframework.com/documentation/2.9.x/ConfigFile#Duration-format

#### General

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -243,10 +243,11 @@ play.http {
     sameSite = null
 
     # Sets the max-age field of the cookie to 5 minutes.
+    # Format spec: https://www.playframework.com/documentation/2.3.4/Configuration#Duration-format
     # NOTE: this only sets when the browser will discard the cookie. Play will consider any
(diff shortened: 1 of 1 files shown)


server/conf/application.conf
```diff
@@ -246,7 +246,7 @@ play.http {
     # NOTE: this only sets when the browser will discard the cookie. Play will consider any
     # cookie value with a valid signature to be a valid session forever. To implement a server side session timeout,
     # you need to put a timestamp in the session and check it at regular intervals to possibly expire it.
-    #maxAge = 300
+    #maxAge = 300s
```
gwendolyngoetz, line 249: (nonblocking) Can you add the link to the duration format so it's easier to find?

gwendolyngoetz (approved):

server/conf/application.conf
jeffpw-goog: Good idea thanks!

PR #8621, opened 2024-09-17
Set a session ID on all pac4j profiles

jeffpw-goog:
### Description

#### General


gwendolyngoetz (commented):
Do we need to do something similar with the  `SamlProfileCreator`?

jeffpw-goog:
> Do we need to do something similar with the `SamlProfileCreator`?

Not as far as I can tell. This PR moves the session ID logic out of the profile creator, and SamlProfileCreator didn't have it to begin with. The only other session ID related part of OidcProfileCreator is the ID tokens thing, which looks specific to OIDC enhanced logout. Is there something else you have in mind?

gwendolyngoetz (approved):

Issue #8414, opened 2024-08-26
Make landing page not require a cookie


nb1701:
Since we changed how logins work, when you land on the CiviForm index page, a guest account is automatically created and you are redirected with a session cookie for that guest account.  This creates two problems:
1) This creates a lot of unused, empty guest accounts, because everyone who just visits the page and doesn't do anything on it gets a new guest account created.
2) Search engine crawlers can not index the page because they can not store cookies while crawling, which results in an infinite redirect that creates new guest accounts over and over.

Instead, the home page should not create a new user, and a user should only be created when hitting the "apply" button on a program (if you aren't already logged in).

I believe we made it work the current way because it was easiest just to pretend a user had clicked "log in as guest" when we removed the landing page.

This is currently a problem for AR as they are unable to get Google to scrape their new CiviForm site and include it in search results.  Seattle is currently in search results, but likely from it being scraped long ago before we implemented the login changes.

rockycodes:
related: https://github.com/civiform/civiform/issues/5542#issuecomment-1709016609

PR #8634, opened 2024-09-18
Allow viewing the home page without a cookie or guest user

nb1701:
The index page (which is really a redirect to /programs) used to be behind a login screen. When we removed that, we created a guest user automatically. This is problematic for web crawlers because they can't store cookies, so it will end up redirecting to the callback to create a guest client over and over. Additionally, this approach ends up creating a guest user any time someone loads the site, meaning we get a lot of empty guest accounts.

This changes the approach so that when you load / or /programs, a guest account is not created. It is only created when you click a link to login or start applying to a program.

This also hides the "You're a guest user." text and "End session" link when there is no profile present. On North Star, it hides the banner.

Finally, this moves the DevTools link off to the side so the layout of the other components (the language dropdown) is the same as it is on prod.

#### General

Watch the number of accounts in the database, and ensure it isn't incrementing every time you load the home page or do things on that page that don't require an account (language drop down, program category filtering, anything else?).

Try curling the home page directly and ensure it gives you back valid HTML (after the single redirect) and doesn't redirect endlessly (use the `-L` flag).

### Issue(s) this completes

https://github.com/civiform/civiform/issues/8414

### DIFF
--- a/browser-test/src/header.test.ts
+++ b/browser-test/src/header.test.ts
@@ -13,7 +13,11 @@ test.describe('Header', () => {
   /**
    * @todo (#4360) add a "Not logged in, guest mode disabled" test once we can get to the programs page without logging in, for an entity without guest mode.
    */
-  test('Check screenshots and validate accessibility', async ({page, adminPrograms, adminQuestions, applicantQuestions}) => {
+  test('Check screenshots and validate accessibility', async ({
+    page,
+    adminPrograms,
+    applicantQuestions,
+  }) => {
     await test.step('Take a screenshot with no profile/account', async () => {
       await validateScreenshot(page.getByRole('navigation'), 'not-logged-in')
     })
--- a/browser-test/src/profile_stability.test.ts
+++ b/browser-test/src/profile_stability.test.ts
@@ -75,7 +75,7 @@ test.describe('User HTTP sessions', {tag: ['@parallel-candidate']}, () => {
     await test.step('Simulate going to a user route by going to the callback', async () => {
       await page.goto('/callback?client_name=GuestClient')
     })
-    
+
     await test.step('Verify profile data', async () => {
       await page.goto('/dev/profile')
       const content = await page.content()


server/app/filters/NonUserRoutes.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.google.common.collect.ImmutableList;
+import play.mvc.Http;
+
+/**
+ * Path prefixes and static paths for HTTP routes that may not involve a user session and can be
+ * safely skipped by filters that act on the user session or account. The prefixes all should not
+ * involve a session, and the programs path may or may not, but is included here to allow it not to.
+ */
+final class NonUserRoutes {
```
rockycodes, line 11: Since we're including `/programs` which is a user route, I think it might be confusing to call this `NonUserRoutes`. I think we should either include `/programs` in the `CiviFormProfileFilter` explicitly or change the name of this. I think other filters also use `NonUserRoutes` (maybe the settings filter?) so I would opt to include `/programs` explicity in `CiviFormProfileFilter` and remove it from this class.

rockycodes:
A thought about "end session" -- could we hide that link if there is no session?

server/app/controllers/applicant/ApplicantInformationController.java
```diff
@@ -121,6 +121,25 @@ public CompletionStage<Result> setLangFromBrowser(Http.Request request, long app
             });
   }
 
+  /**
+   * When choosing a language from the switcher on the home page without being logged in, we simply
+   * redirect back with the appropriate locale set, without saving it to any particular user.
+   */
+  public CompletionStage<Result> setLangFromSwitcherWithoutApplicant(Http.Request request) {
```
rockycodes, line 128: I was thinking that switching the language should actually create a guest user and start a session. We probably want it to switch back to english after 30 minutes of inactivity. What do you think?

server/app/views/applicant/NorthStarProgramIndexView.java
```diff
@@ -140,8 +140,15 @@ public String render(
...
                     program -> program.programId(),
-                    program ->
-                        applicantRoutes.review(profile, applicantId, program.programId()).url()));
+                    program -> {
+                      if (profile.isPresent() && applicantId.isPresent()) {
+                        return applicantRoutes
+                            .review(profile.get(), applicantId.get(), program.programId())
+                            .url();
+                      } else {
+                        return applicantRoutes.review(program.programId()).url();
```
rockycodes, line 149: In North Star, is "review" the program index page? Asking because in our current flow "review" is when you review your answers so we could definitely have an applicant at that point. But I think it's being used differently in North Star.

server/app/filters/ValidAccountFilter.java
```diff
@@ -47,7 +47,7 @@ public EssentialAction apply(EssentialAction next) {
    * infinite redirect.
    */
   private boolean allowedEndpoint(Http.RequestHeader requestHeader) {
-    return NonUserRoutePrefixes.anyMatch(requestHeader) || isLogoutRequest(requestHeader.uri());
+    return NonUserRoutes.anyMatch(requestHeader) || isLogoutRequest(requestHeader.uri());
```
rockycodes, line 50: Oh wait this is the other place it's used, not SettingsFilter. I think we wouldn't want `/programs` to be included here since this is only reached if `profile.isPresent` returns true.

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -121,6 +121,25 @@ public Content render(
...
+  public Content renderWithoutApplicant(
+      Messages messages,
+      Http.Request request,
+      ApplicantService.ApplicationPrograms applicationPrograms,
+      ImmutableList<String> selectedCategoriesFromParams) {
+    return render(
+        messages,
+        request,
+        Optional.empty()
+        /* applicantId= */ ,
```
rockycodes, line 133: This is a nit, but if it's possible to have the comments appear on the same line as the vars that's preferable (maybe use `//` instead of `/*`)

server/app/views/applicant/ProgramIndexView.java
```diff
@@ -565,16 +584,20 @@ private DivTag findServicesSection(
...
         .withId("category-filter-form")
         .withAction(
-            controllers.applicant.routes.ApplicantProgramsController.indexWithApplicantId(
-                    applicantId, ImmutableList.of())
-                .url())
+            applicantId.isPresent()
+                ? controllers.applicant.routes.ApplicantProgramsController.indexWithApplicantId(
+                        applicantId.get(), ImmutableList.of())
+                    .url()
+                : controllers.applicant.routes.ApplicantProgramsController.indexWithoutApplicantId(
```
rockycodes, line 598: Did you say this would create a new guest account? Why is that?

server/app/services/applicant/ApplicantPersonalInfo.java
```diff
@@ -56,6 +56,13 @@ public String getDisplayString(Messages messages) {
 
   public abstract Representation tiPartiallyCreated();
 
+  // An empty guest user without any data. Used on the home page when a user
+  // is not logged in and hasn't created a guest account yet by applying to
+  // a program.
+  public static ApplicantPersonalInfo ofGuestUser() {
```
rockycodes, line 62: I like it.

server/app/views/applicant/ProgramCardViewRenderer.java
```diff
@@ -225,7 +225,9 @@ public LiTag createProgramCard(
     // Use external link if it is present else use the default Program details page
     String programDetailsLink =
         program.externalLink().isEmpty()
-            ? applicantRoutes.show(profile, applicantId, program.id()).url()
+            ? profile.isPresent() && applicantId.isPresent()
+                ? applicantRoutes.show(profile.get(), applicantId.get(), program.id()).url()
+                : applicantRoutes.show(program.id()).url()
```
rockycodes, line 230: It's great that those `show` and `review` routes already existed without applicant id!

rockycodes (commented):
I think this looks great! As clean an implementation and we could have. Passing them all down as optionals isn't so so bad. I don't have any real red flags except maybe including `/programs` in the `NonUserRoutes`.

server/app/filters/NonUserRoutes.java
rockycodes: I was thinking to include `/` in non user routes since it's not a route a user would see. It just directs them based on their profile if they have one. But I don't feel strongly about it, and the same concern could apply about unintended consequences for other filters using the non user routes filter.

server/app/controllers/applicant/ApplicantInformationController.java
jeffpw-goog: The language is stored in a separate cookie than the session:

image

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -394,10 +400,12 @@ private DivTag authDisplaySection(
           div(
               span(loggedInAsMessage).withClasses("text-sm"),
```
jeffpw-goog, line 401: **suggestion:** I think it should not show the loggedInAsMessage if you don't have a session. I remember when I first used CiviForm I thought it was weird that clicking "End session" had no visible effect, and I think it would make more sense if it didn't show "You're logged in" until you actually start a session.

Same suggestion for the Northstar path; I didn't dig up the code for it.

server/app/views/NorthStarBaseView.java
```diff
@@ -97,11 +97,13 @@ protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(
     // that will be embedded in the guest alert in the header.
     context.setVariable(
         "endSessionLinkHtml",
-        "<a id=\"logout-button\" class=\"usa-link\" href=\""
-            + logoutLink
-            + "\">"
-            + messages.at(MessageKey.END_YOUR_SESSION.getKeyName())
-            + "</a>");
+        profile.isPresent()
```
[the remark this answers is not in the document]
jeffpw-goog: I am confused - it looks like for a user with no profile this would show "End session" it just wouldn't be linked? Shouldn't it show nothing at all in that case?

I tried to run your PR locally to confirm that's what would happen, but it's showing the link even when I delete the session so I'm confused about that too...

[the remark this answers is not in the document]
jeffpw-goog: What would the new message be? I was thinking it wouldn't say anything at all which wouldn't require any translations.

server/test/support/ResourceCreator.java
```diff
@@ -130,6 +130,14 @@ public ProgramModel insertActiveDisabledProgram(String name) {
     return ProgramBuilder.newDisabledActiveProgram(name).build();
   }
 
+  public ProgramModel insertActiveTiOnlyProgram(String name) {
```
avaleske, line 133: **non-blocking, nit**: it'd be nice if these names were the same format, so `[insert|new]ActiveTiOnlyProgram`

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -2097,6 +2097,47 @@ applicant.id, trustedIntermediaryProfile, createRequestWithFastForwardDisabled()
             result.unapplied().get(1));
   }
 
+  @Test
+  public void relevantProgramsForNoApplicant() {
```
avaleske, line 2101: **question** do we need to duplicate this test in `ApplicantServiceFastForwardEnabledTest`?

avaleske (approved):
Thank you for doing this!

Some fast follows that might be nice:
- Fix the positioning of the dev-tools button so it doesn't move the other links out of their prod positions, maybe by absolutely positioning it nearby?
- Do we save the language once the user logs in or starts an application?

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
```diff
@@ -94,17 +94,24 @@ public ImmutableList<ProgramCardParams> getCards(
...
+          profile.isPresent() && applicantId.isPresent()
+              ? applicantRoutes.review(profile.get(), applicantId.get(), program.id()).url()
+              : applicantRoutes.review(program.id()).url();
 
       // Note this doesn't yet manage markdown, links and appropriate aria labels for links, and
       // whatever else our current cards do.
+      String detailsUrl = program.externalLink();
+      if (detailsUrl.isEmpty() || detailsUrl.isBlank()) {
+        detailsUrl =
+            profile.isPresent() && applicantId.isPresent()
```
rockycodes, line 107: suggestion (non-blocking): You do this same ternary check for `profile.isPresent() && applicantId.isPresent()` in a couple of different places. You could factor it out into a well named method for clarity and ease of reference, but I don't feel strongly about it.

server/app/views/applicant/ProgramIndexView.java
rockycodes: Awesome!

server/conf/routes
```diff
@@ -162,8 +162,9 @@ GET     /applicants/:applicantId/files/:fileKey     controllers.FileController.s
 GET     /assets/*file               controllers.Assets.versioned(file)
 
 # Methods for applicants
-GET     /applicants/:applicantId/edit                                       controllers.applicant.ApplicantInformationController.setLangFromBrowser(request: Request, applicantId: Long)
-POST    /applicants/:applicantId                                            controllers.applicant.ApplicantInformationController.setLangFromSwitcher(request: Request, applicantId: Long)
+GET     /applicants/:applicantId/edit               controllers.applicant.ApplicantInformationController.setLangFromBrowser(request: Request, applicantId: Long)
```
rockycodes, line 165: question (non-blocking): Did anything change about this route? I don't see anything different but github is marking it as changed.

server/app/services/applicant/ApplicantService.java
```diff
@@ -1055,6 +1056,28 @@ public CompletionStage<ApplicationPrograms> relevantProgramsForApplicant(
...
   }
 
+  /**
+   * Get all active programs that are publicly visible, as if it was a brand new guest account, but
+   * without requiring the account to be created yet.
+   *
+   * @param request - The request object from loading the page
+   * @return - CompletionStage of the relevant programs
+   */
+  public CompletionStage<ApplicationPrograms> relevantProgramsForNoApplicant(Request request) {
```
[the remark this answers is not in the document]
rockycodes: Or even `relevantProgramsWithoutApplicant` which matches names we have for methods elsewhere.

rockycodes (changes requested):
Looks good! Mostly small, non-blocking comments. My biggest blocking comment/question is about the look of the page when `DevTools` isn't rendered.

browser-test/image_snapshots/auth_test/guest-just-ended-session.png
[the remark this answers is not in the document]
avaleske: ahhh I missed that change in the test

avaleske (approved):

browser-test/image_snapshots/applicant_application_index_test/common-intake-form-not-set.png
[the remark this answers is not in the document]
rockycodes: Ok cool! Looks good. I still would like for Sijia to give the 👍 about this part of the new layout since it will be updating it for all homepages 
image

browser-test/image_snapshots/auth_test/user-just-ended-session.png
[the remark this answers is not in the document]
rockycodes: Makes sense. I'll resolve this.

server/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -148,18 +149,49 @@ public CompletionStage<Result> indexWithApplicantId(
...
+                      messagesApi.preferred(request),
+                      request,
+                      programs,
+                      ImmutableList.copyOf(categories)));
+        });
+  }
+
   public CompletionStage<Result> index(Request request) {
+    if (profileUtils.optionalCurrentUserProfile(request).isEmpty()) {
+      return indexWithoutApplicantId(request, ImmutableList.of());
```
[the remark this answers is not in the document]
rockycodes: Ok cool. If I understand correctly, preventing it from creating a session when filtered means if the previous user did any filtering and then walked away, the next person will see the filtered view. But it's pretty clear in the UI when it's filtered so I don't think that matters. What do you think?

Re: removing the `categories` param from `indexWithoutApplicantId` -- I'm suggesting instead of passing down an empty list everywhere you call `indexWithoutApplicantId`, just pass in an empty list here:
```
ok(
                  programIndexView.renderWithoutApplicant(
                      messagesApi.preferred(request),
                      request,
                      programs,
                      ImmutableList.of()));
```

I don't think this would break anything categories related, but if I'm wrong or if you would rather just leave it as is feel free to resolve this!

rockycodes (approved):
Lgtm pending approval from Sijia on the nav! Thanks for doing this!

PR #8664, opened 2024-09-19
Removing Status related triggers

### LINKED ISSUE #8665: Remove Triggers related to statuses from CiviForm

Following the discussion on 9/18/2024, the team unnanimously decided to remove triggers from the Code base. 
Please refer to the PPT at - https://docs.google.com/presentation/d/1ZkfmJcwbtLYW9bpfrMQsud1_lgYY-MUcNSPGAaOekXE/edit#slide=id.g2fbc8718777_0_31 for more information.

All Trigger related code should be removed/reverted in the code + any new functionality should be added in the applications code to support what the triggers do today programatically.

swatkat1:
### Description

Removing Status related triggers from the database.
Documentation link - https://docs.google.com/presentation/d/1ZkfmJcwbtLYW9bpfrMQsud1_lgYY-MUcNSPGAaOekXE/edit#slide=id.g3023ba603ce_2_10

## Release notes

Removing Status related triggers from the database. Post this, the status update should function normally as before. 

#### General

#### Database evolutions

- [x] Assigned two reviewers

#### New Features

### Issue(s) this completes

Fixes #8665


server/conf/evolutions/default/76.sql
```diff
@@ -2,7 +2,7 @@
 
 ALTER TABLE applications ADD COLUMN IF NOT EXISTS latest_note varchar;
 
--- Update historical data from application events table
+# -- Update historical data from application events table
```
devin-civi, line 5: Isn't this where the error is coming from? Is the problem maybe that this is now doing two inline comments? (Since both # and -- are ways to specify a single line comment?) I would think it's valid, but maybe SQL is weird.

swatkat1: the 44.sql has comments like this all over the place but not sure why this one is complaning. I will reset it locally and give it another shot.

server/test/models/ApplicationModelTest.java
```diff
@@ -21,67 +19,13 @@ public class ApplicationModelTest extends ResetPostgres {
...
-    ApplicationModel application =
-        resourceCreator.insertActiveApplication(
-            resourceCreator.insertApplicantWithAccount(), program);
-    assertThat(application.getLatestStatus()).isEmpty();
-
-    application.setLatestStatusForTest("some-status-value");
-    application.save();
-    application.refresh();
-    assertThat(application.getLatestStatus()).isEmpty();
+    applicationEventRepository = instanceOf(ApplicationEventRepository.class);
```
avaleske, line 28: Is this being used in this file?

server/test/models/ApplicationModelTest.java
```diff
@@ -21,67 +19,13 @@ public class ApplicationModelTest extends ResetPostgres {
...
   public void setUp() {
     applicationStatusesRepository = instanceOf(ApplicationStatusesRepository.class);
-  }
-
-  @Test
-  public void staleLatestStatusIsNotPersisted() {
-    // Tests a case where an Application (and its associated latest_status value has been loaded
-    // in-memory, a new ApplicationEventDetails is added (causing the trigger to execute), and the
-    // Application is persisted.
-    ProgramModel program = ProgramBuilder.newActiveProgram("test program", "description").build();
```
avaleske, line 36: **suggestion**: I think we actually do need to handle this race condition. Anywhere we update a submitted application, I think we're going to need to check if the `latest_status` field in the database matches the latest_status field on the `ApplicationModel` we're about to save, and if it doesn't, carry it forward to the newly updated application in the same transaction.

avaleske (changes requested):
overall lgtm! but two changes I think we need to make:
- small one is keeping `insertStatusEvent()` async
- a larger one is handling the race condition that one of the triggers was guarding against. If we have an application in memory, its status gets updated by another admin, and then we save it, we lose the latest status. Carrying the latest status forward whenever we save modifications to a submitted application, in the same transaction, should guard against that.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -68,6 +69,42 @@ public ImmutableList<ApplicationEventModel> getEventsOrderByCreateTimeDesc(Long
...
+            .setStatusEvent(newStatusEvent)
+            .build();
+    ApplicationEventModel event = new ApplicationEventModel(application, optionalAdmin, details);
+    try (Transaction transaction = database.beginTransaction(TxIsolation.SERIALIZABLE)) {
+      insertSync(event);
+      // Saves the latest note on the applications table too
+      // If the statuses are removed from an application, then the latest_status column needs to be
+      // set to null
+      // to indicate the applications has no status and not a status with empty string.
+      if (Strings.isNullOrEmpty(newStatusEvent.statusText())) {
```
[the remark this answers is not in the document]
swatkat1: yes. I think the initial idea was- if a status was removed (I really don't know how the CiviForm Admins would set it), then the application has no status

server/app/services/applicant/ApplicantService.java
```diff
@@ -466,22 +465,20 @@ public CompletionStage<ApplicationModel> submitApplication(
...
-            .setStatusEvent(statusEvent)
-            .build();
+
     // Because we are doing this automatically, set the Account to empty.
-    return applicationEventRepository.insertAsync(
-        new ApplicationEventModel(application, /* creator= */ Optional.empty(), details));
+    applicationEventRepository.insertStatusEvent(
+        application, /* optionalAdmin= */ Optional.empty(), statusEvent);
+
+    return CompletableFuture.completedFuture(application);
```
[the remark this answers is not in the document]
swatkat1: done!

server/test/models/ApplicationModelTest.java
swatkat1: adding the test back will use it Austin

server/test/models/ApplicationModelTest.java
```diff
@@ -28,6 +30,41 @@ public void setUp() {
     applicationEventRepository = instanceOf(ApplicationEventRepository.class);
   }
 
+  @Test
```
avaleske, line 33: ok my bad. you were right and can remove this test actually!

avaleske (changes requested):
ok! this looks good but for two things I missed earlier:

1. I think we also need to set insertSync() private and make this use the new way of setting events:
https://github.com/civiform/civiform/blob/476dd77dc805fde4f6b4292f4f8bc5f258c18b24/server/app/services/applications/ProgramAdminApplicationService.java#L151

And the comment here is now out of date:
https://github.com/civiform/civiform/blob/476dd77dc805fde4f6b4292f4f8bc5f258c18b24/server/app/models/ApplicationModel.java#L157-L170

And then we can remove the test I asked you to put back now that I figured out the race condition isn't possible.

server/test/models/ApplicationModelTest.java
swatkat1: I think I will leave it Austin. This test to make sure the latest_status is carried forward if updates to the existing application row is made (today only note update is possible) This check will ensure that tomorrow say something else is updated, then the status should not be dropped.

swatkat1:
@avaleske - Since insertSync() is used in testing can I make it protected + @visibleToTesting?

server/app/repository/ApplicationEventRepository.java
```diff
@@ -68,6 +64,58 @@ public ImmutableList<ApplicationEventModel> getEventsOrderByCreateTimeDesc(Long
...
+                  .eq("id", application.id)
+                  .update();
+            } else {
+              database
+                  .update(ApplicationModel.class)
+                  .set("latest_status", newStatusEvent.statusText())
+                  .where()
+                  .eq("id", application.id)
+                  .update();
+            }
```
gwendolyngoetz, line 110: (blocking) The could be simplified by getting the value `Strings.isNullOrEmpty(newStatusEvent.statusText()) ? null : newStatusEvent.statusText()` so you only have to write the update once and not sorry about having to change it in two places down the road.

server/app/repository/ApplicationEventRepository.java
```diff
@@ -68,6 +64,58 @@ public ImmutableList<ApplicationEventModel> getEventsOrderByCreateTimeDesc(Long
...
+   * transaction.
+   *
+   * <p>Previously in 44.sql, DB triggers were used to update the latest_status between
+   * application_events table and applications table. In 78.sql script, the team decided to drop
+   * triggers and use application code to update both tables. This method implements the same.
+   *
+   * <p>Note - Application code must change both tables at once as we want to avoid inconsistency
+   * between the tables.
+   */
+  public CompletionStage<ApplicationEventModel> insertStatusEvent(
```
gwendolyngoetz, line 78: (question) This look to only ever be used by forcing it to resolve the future by calling ` toCompletableFuture().join()`. Is these a reason to have this return a `CompletionStage<ApplicationEventModel>` instead of just `ApplicationEventModel`?

avaleske: The ApplicantService uses the asynchronously, since it used to use the original `supplyAsync()` method that this PR removes.

server/test/models/ApplicationModelTest.java
avaleske: but now this is unused, right?

server/test/models/ApplicationModelTest.java
avaleske: makes sense! we need to update the comment then cause it references the trigger this PR removes.

avaleske (approved):
lgtm with gwen's comment about isNullOrEmpty() and updating the comment in the test

server/app/repository/ApplicationEventRepository.java
gwendolyngoetz: Changed to blocking

gwendolyngoetz (changes requested):

server/test/models/ApplicationModelTest.java
swatkat1: sure thing!

server/app/repository/ApplicationEventRepository.java
swatkat1: Jolly good! will do it right now!

avaleske (approved):

gwendolyngoetz (approved):

PR #8668, opened 2024-09-19
Adding a step to removing temp/servers.json in pgadmin script

### LINKED ISSUE #8669: Dev-pgadmin script fails occasionally on error
**Describe the bug**
Dev-pgadmin script fails occasionally on the following error

tee: /tmp/servers.json: Is a directory

Add a step in the script to clean up the file before starting the script.

swatkat1:
### Description

Adding a step to removing temp/servers.json in pgadmin script

### Issue(s) this completes

Fixes #8669

### DIFF
--- a/bin/dev-pgadmin
+++ b/bin/dev-pgadmin
@@ -1,11 +1,11 @@
 #! /usr/bin/env bash
 
-rm -rf /tmp/servers.json
-
 # DOC: Loads the pgadmin gui tool and connects to the dev environment database
 
 source bin/lib.sh
 docker::set_project_name_dev
+
+[[ -d "/tmp/servers.json" ]] && rmdir /tmp/servers.json
 
 # The servers.json file is used to automatically configure
 # our database to show up in the pgadmin server tree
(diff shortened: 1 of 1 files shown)


bin/dev-pgadmin
```diff
@@ -1,5 +1,7 @@
 #! /usr/bin/env bash
 
+rm -rf /tmp/servers.json
```
gwendolyngoetz, line 3: (blocking) A few issues
- This is in the wrong spot, it should be below `docker::set_project_name_dev` line.
- We should not be using `rm -rf` in scripts that making changes to a user's filesystem outside the project directory.
- We should revamp how this whole thing works. Whatever the issue is that is sometimes causing this to become a directory concerns me. For now I'm fine with a targeted check to remove the specific directory like this `[[ -d "/tmp/servers.json" ]] && rmdir /tmp/servers.json`. It should be sufficient since it's an empty directory when it's happened to me.

gwendolyngoetz (changes requested):

bin/dev-pgadmin
swatkat1: Done! Thank you for your input.

gwendolyngoetz (approved):

PR #8745, opened 2024-09-27
Cleanup manual jsonnoding

gwendolyngoetz:
### Description

## Release notes

Remove this section if the title of the pull request can be used as the default release notes description. If more detail is needed to communicate to partners the scope of the PR's changes, use this release notes section (and remove this placeholder text).

#### General

#### Database evolutions

- [ ] Assigned two reviewers

#### Durable jobs

- [ ] Assigned two reviewers

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/controllers/geo/AddressSuggestionJsonSerializer.java
```diff
@@ -89,7 +89,8 @@ private <T> T deserializeJson(String json, Class<T> clazz) {
     try {
       return mapper.readValue(json, clazz);
     } catch (JsonProcessingException e) {
-      throw new BadRequestException("Error deserializing " + clazz.toString());
+      throw new BadRequestException(
```
gwendolyngoetz, line 92: `BadRequestException` doesn't take an exception so just appending the message to the string as it was getting lost making for difficult debugging. I'd change the exception but that will trickle out more than I want this PR to deal with.

server/app/services/geo/AddressSuggestionGroup.java
```diff
@@ -20,6 +20,15 @@ public static Builder builder() {
 
   public abstract Address getOriginalAddress();
 
+  /** Returns an empty instance of an AddressSuggestionGroup */
+  public static AddressSuggestionGroup empty(Address originalAddress) {
+    return AddressSuggestionGroup.builder()
+        .setWellKnownId(0)
```
gwendolyngoetz, line 26: I don't like setting this to 0, but cleaning this up is outside the scope of this PR.

server/app/services/geo/esri/models/Attributes.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  public String stateAbbreviation() {
+    // Start with getting the value from the RegionAbbr, this is the default used by ESRI for
+    // state abbreviations. Use if two characters.
+    if (regionAbbr.length() == 2) {
+      return regionAbbr;
+    }
+
+    // If regionAbbr is not two characters it is either empty or the full state name so pull
+    // from the region value. We don't use this be default because some custom implementations
```
rockycodes, line 66:
```suggestion
    // from the region value. We don't use this by default because some custom implementations
```

server/app/services/geo/esri/EsriClient.java
```diff
@@ -97,126 +102,102 @@ public CompletionStage<AddressSuggestionGroup> getAddressSuggestions(Address add
...
-    // default used by ESRI for state abbreviations
-    String result = getNodeTextValueOrEmpty(attributes, "RegionAbbr");
-
-    // Assume if it's not two characters it is either empty or the
-    // full state name and pull from the Region value
-    if (result.length() != 2) {
-      result = getNodeTextValueOrEmpty(attributes, "Region");
-    }
+  private static String getStateAbbreviationFromAttributes(Attributes attributes, Address address) {
+    String result = attributes.stateAbbreviation();
```
rockycodes, line 192: question (non-blocking): do we still need to handle the case below where `result.length() != 2` since that is now handled when calling `attributes.stateAbbreviation()`?

I also have a recommendation to rename this to "region" like you have it in `Attributes` unless it is always state. And drop the "abbreviated" part of this method name since it's not always abbreviated.

gwendolyngoetz: Yes it's still needed, I didn't want to make the fallback logic into the attributes as it is unrelated. So it still needs to fallback to the original user supplied state value here. I switched the attributes method to return empty if the abbreviation isn't two characters and now just check if it's present here or not.

I'm keeping the naming as state instead of region, for now. I know there are some assumptions in the system around addresses that it's going to be a US address. I'd rather wait to get naming right when we have a use case to test with other addresses.

rockycodes (approved):

PR #8751, opened 2024-09-30
[Program Migration] Allow importing large programs + some small frontend fixes

### LINKED ISSUE #8643: [Program Migration] add spellcheck=false to textarea to prevent spelling "errors" from being highlighted

### LINKED ISSUE #8650: [Program Migration] Allow large programs to be imported
Attempting to import a program from Seattle doesn't work because the json is truncated. The program has 5792 lines, and only 2757 lines are pasted. This program is on the smaller side of Seattle programs.

From my initial investigation, this is likely due to a data limits in the Play framework. I tried removing whitespace and removing limits on number of json rows and the problem persisted. Resolving this might require offering an upload option for program json.

rockycodes:
### Description

The main change in this PR is bumping Play's max memory limit to 256KB (and bumping the max length on the text area accordingly) to allow for large programs to be imported. I've added a note to warn admin that submitting a program above this size will error.

I'm also fixing up a couple of small frontend things while I'm here:
- increased the size of the textarea box
- added `spellcheck=false` to text area so the browser doesn't add those annoying red squiggly lines under "mispelled" words
- fixed a typo

#### General

- [x] Extended the README / documentation, if necessary. For user-facing features, consider updating [the user docs](https://github.com/civiform/docs). For "under-the-hood" changes or things more relevant to developers, consider updating [the dev wiki](https://github.com/civiform/civiform/wiki). [Docs update](https://github.com/civiform/docs/pull/519)

Try importing a reallllly large program (larger than 100KB). The easiest way to make a really large program is to create a program with a bunch of questions and then add long translations in every language to those questions.

### Issue(s) this completes

Fixes #8650 ; Fixes #8643


server/conf/application.conf
```diff
@@ -229,6 +229,9 @@ play.http {
   # If null, will attempt to load a class called ErrorHandler in the root package,
   errorHandler = "controllers.ErrorHandler"
 
+  ## Max memory
+  parser.maxMemoryBuffer = 256K
```
rockycodes, line 233: make a custom body parser instead of setting this globally: https://www.playframework.com/documentation/2.9.x/JavaBodyParsers#Writing-a-custom-max-length-body-parser

browser-test/image_snapshots/admin_program_migration_test/import-page-with-error-missing-program-fields.png
rockycodes, line 1: I'm not sure what's different here. Can you tell the difference? It could be that I changed some vertical spacing on the element that is swapped out when the error comes in.

server/app/views/admin/migration/AdminImportView.java
```diff
@@ -24,22 +24,22 @@
...
+   * https://www.playframework.com/documentation/2.9.x/JavaBodyParsers#Content-length-limits
    *
-   * <p>Note that the HTML textarea element will automatically truncate the string to be the max
-   * character length, which will likely result in JSON parsing errors. TODO(#7087): Make that
-   * truncation obvious to admins.
+   * <p>Pasting in a json string above this limit will automatically truncate the string which
+   * results in a server error.
    */
-  private static final int MAX_TEXT_LENGTH = 100000;
+  private static final int MAX_TEXT_LENGTH = 256000;
```
rockycodes, line 42: I tried adding a test for this case (what a pain) but the json string was so large that the formatter couldn't handle it. I thought about it a little more and I'm not sure we need to test the case where we're submitting a super large program since we would just be testing that these limit settings work, which is not logic we control. But I'm curious what you both think about adding a test case and, if you think we need it, how I might be able to test it?

browser-test/image_snapshots/admin_program_migration_test/import-page-with-error-missing-program-fields.png
rockycodes: Oh duh! Thank you!

server/app/views/admin/migration/AdminImportView.java
```diff
@@ -120,13 +120,16 @@ private DomContent createUploadProgramJsonForm(Http.Request request) {
...
+            .setAttribute("rows", "10")
+            .setAttribute("spellcheck", "false")
             .getTextareaTag();
 
     return div()
         .with(
-            p("To import a program, copy the JSON file content and paste into the box below")
-                .withClass("py-2"),
+            p("To import a program, copy the JSON file content and paste into the box below."),
+            p("Note: Programs larger than 256,000 characters will cause an error. If you need to"
```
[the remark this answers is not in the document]
rockycodes: That's a great question, I'm not sure. I'll look into it a little and add it in if it's simple, otherwise I'll make a follow-up issue to do that.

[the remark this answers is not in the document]
rockycodes: Made this issue: https://github.com/civiform/civiform/issues/8772

As it works right now, the json will be truncated on the page. An error will be thrown when they try to "preview" the program, so they'll never get to the "save" stage.

gwendolyngoetz (approved):

PR #8754, opened 2024-09-30
Vary program filter chip HTML element ids

### LINKED ISSUE #8632: All the program filter chips have the same id
**Describe the bug**
All the program filter chip `div` elements have the same id of `filter-chip`, which is invalid html.

**Expected behavior**
Either add a distinguishing word to the id or remove the ids from the `div` elements if they aren't needed.

melanie-exygy:
### Description

Varying program filter chip HTML element ids so that we have valid HTML.
  
### Issue(s) this completes

Fixes #8632

### DIFF
--- a/server/app/views/applicant/ProgramIndexView.java
+++ b/server/app/views/applicant/ProgramIndexView.java
@@ -603,7 +603,7 @@ public final class ProgramIndexView extends BaseHtmlView {
                         relevantCategories,
                         category ->
                             div()
-                                .withId("filter-chip")
+                                .withId("filter-chip-" + category)
                                 .with(
                                     input()
                                         .withId("check-category-" + category)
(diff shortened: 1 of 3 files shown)


server/app/views/applicant/ProgramIndexView.java
```diff
@@ -603,7 +603,7 @@ private FormTag renderCategoryFilterChips(
                         relevantCategories,
                         category ->
                             div()
-                                .withId("filter-chip")
+                                .withId("filter-chip-" + category)
```
gwendolyngoetz, line 606: (question, blocking) Can the categories have spaces or non-alphanumeric characters? If yes, they should be replaced with `-` or `_`.

melanie-exygy: Right now, all ten categories are a single word, but I can see the value of future-proofing this.  Will add the transformation logic.

melanie-exygy: Done

gwendolyngoetz (approved):

PR #8765, opened 2024-10-02
Preventing categories from being added to pre-screeners

### LINKED ISSUE #8645: Don't allow tags on pre-screener
Currently, we allow admins to add category filtering tags on pre-screener programs. Since pre-screeners are more all-encompassing than individual programs and meant for a resident to fill out before other programs, categories do not apply to pre-screeners (confirmed with AK & Charlotte). 

I'm thinking that if pre-screener is checked on a program, we gray out all category options or remove that section completely.

melanie-exygy:
### Description

This prevents categories from being added to a program if it is a pre-screener (aka "common intake form") since pre-screeners are more general and it doesn't make sense for them to be labelled with categories.  When an admin goes to either the program creation or edit forms, if the program is a pre-screener, or if they check the pre-screener checkbox, all categories are disabled and unselected.

Note that this won't automatically remove categories from any existing pre-screeners.  However, given that program filtering is still in test, we've determined that it's sufficient for admins to be able to edit the program if they want to remove the categories in that case. (See comments on the work issue:  #8645 )

### Issue(s) this completes

Fixes #8645


server/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -328,7 +328,10 @@ private DivTag showCategoryCheckboxes(
                                             .withType("checkbox")
                                             .withName("categories" + Path.ARRAY_SUFFIX)
                                             .withValue(String.valueOf(category.getId()))
-                                            .withCondChecked(categories.contains(category.getId())),
+                                            .withCondDisabled(isCommonIntakeForm)
+                                            .withCondChecked(
+                                                categories.contains(category.getId())
```
swatkat1, line 333: Non-blocking- Is there a reason to add it twice- both in withCondDisable and withCondChecked

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -21,6 +21,32 @@ class AdminPrograms {
...
+      )
+
+      programCategoryCheckboxes.forEach((checkbox) => {
+        const category = checkbox as HTMLInputElement
+
+        if (commonIntakeCheckbox.checked) {
+          category.disabled = true
+          category.checked = false
+        } else {
+          category.disabled = false
```
swatkat1, line 44: should we add catogory.checked=true here?

melanie-exygy: If we set `category.checked=true`, it will select all the categories when it isn't an intake form.  We should let the user determine which categories should be applied to the program.

server/app/views/admin/programs/ProgramFormBuilder.java
melanie-exygy: Since `disabled` and `checked` are two different HTML attributes, I don't think they can share the same conditional statement.  If you're picturing a way to do that, would you mind giving an example?

swatkat1: This question is more for my understanding.

swatkat1 (approved):

PR #8823, opened 2024-10-08
Initial Framework for Bulk status update

### LINKED ISSUE #8811: Implement changes needed for Bulk status update.

Use, https://ebean.io/docs/transactions/batch to update backend + UI changes needed for Bulk status update.
Part of https://github.com/civiform/civiform/issues/1923

#### What priority is this?
P0 

#### Done when:

Backend changes are updated to support bulk status update.

swatkat1:
### Description

Working changes needed for bulk status update. This PR only covers the Controllers unit tests. The remaining work are tracked below
#9016 
#9015 
#9017 

## Release notes

Initial changes needed for Bulk status update. This feature is under development. 

#### General

### Issue(s) this completes

Fixes #8811 

Screenshot of the new view:
image


server/app/services/applications/ApplicationNotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package services.applications;
+
+public final class ApplicationNotFoundException extends Exception {
```
gwendolyngoetz, line 3:
```suggestion
public final class ApplicationNotFoundException extends RuntimeException {
```
Don't extend directly from Exception. That create checked exceptions requiring you to either have to have excessive try/catch or add `throws blah` on any method that doesn't handle it.

server/app/forms/admin/BulkStatusUpdateForm.java
```diff
@@ -0,0 +1,42 @@
+package forms.admin;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** A form for selecting multiple applications to be considered for a status update. */
+public final class BulkStatusUpdateForm {
+  private List<String> applicationsIds;
```
ssandbekkhaug, line 8: Question: I thought applicationIds are `long`s, not `String`s. Is this correct?

server/app/forms/admin/BulkStatusUpdateForm.java
```diff
@@ -0,0 +1,42 @@
...
+  private String statusText;
+  private Boolean maybeSendEmail;
```
ssandbekkhaug, line 10: Suggestion (non-blocking): Rename this to `shouldSendEmail`. Java typically uses prefixes like "is" or "should" for boolean variables.

server/app/forms/admin/BulkStatusUpdateForm.java
```diff
@@ -0,0 +1,42 @@
...
+    this.statusText = "";
+    this.maybeSendEmail = false;
+  }
+
+  public String getStatusText() {
+    return statusText;
+  }
+
+  public void setStatusText(String statusText) {
+
```
ssandbekkhaug, line 31: Super Nit: remove blank line

server/app/repository/ApplicationEventRepository.java
```diff
@@ -111,10 +113,57 @@ public CompletionStage<ApplicationEventModel> insertStatusEvent(
...
+            .build();
+
+    List<Long> applicationIds =
+        applications.stream().map(app -> app.id).collect(Collectors.toList());
+
+    var applicationsStatusEvent =
+        applications.stream()
+            .map(app -> new ApplicationEventModel(app, optionalAdmin, details))
+            .collect(ImmutableList.toImmutableList());
+    try (Transaction transaction = database.beginTransaction(TxIsolation.SERIALIZABLE)) {
```
ssandbekkhaug, line 140: Question: Does there need to be a corresponding "catch" block?

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -238,6 +241,53 @@ public Optional<ApplicationModel> getApplication(long applicationId, ProgramDefi
     }
   }
 
+  /*
+   * Retrieves the applications for the give ApplicationIds and validates that it is associated with the given
+   * program.
+   */
+  public ImmutableList<ApplicationModel> getApplications(
```
ssandbekkhaug, line 248: Blocking: Can we add unit tests for this method?

ssandbekkhaug (commented):
I started reviewing, and then partway through, I realized this PR is over 1000 lines long(!)

Blocking: Can you split this up into smaller PRs? That would make it much easier to review (and probably improve velocity too).

server/app/forms/admin/BulkStatusUpdateForm.java
swatkat1: This was a tricky piece to get working. I tried with list<long>, it didn't work. Now tried again it worked. java is messing with me :)

server/app/repository/ApplicationEventRepository.java
swatkat1: Great question. We are starting a transaction not necessarily throwing an exception. We are signaling play that every sql query in this block should happen in one go and in a serial manner. The control should not be dropped halfway through the block. I liked this [blog](https://www.thenile.dev/blog/transaction-isolation-postgres) explaining this.

server/app/services/applications/ProgramAdminApplicationService.java
swatkat1: Sure, I have created #9015 to track it. I will be addressing it in the next PR.

server/app/forms/admin/BulkStatusUpdateForm.java
```diff
@@ -5,38 +5,37 @@
...
 public final class BulkStatusUpdateForm {
-  private List<String> applicationsIds;
+  private List<Long> applicationsIds;
   private String statusText;
-  private Boolean maybeSendEmail;
+  private Boolean shouldSendEmail;
 
-  public Boolean isMaybeSendEmail() {
-    return maybeSendEmail;
+  public Boolean sendEmail() {
```
ssandbekkhaug, line 12: Suggestion/Nit: If the public-facing method is `sendEmail()`, should the private variable be just `sendEmail` as well? Or, should the public-facing methods be named `getShouldSendEmail()` and `setShouldSendEmail(...)` for consistency?

server/app/forms/admin/BulkStatusUpdateForm.java
```diff
@@ -5,38 +5,37 @@
 
 /** A form for selecting multiple applications to be considered for a status update. */
 public final class BulkStatusUpdateForm {
-  private List<String> applicationsIds;
+  private List<Long> applicationsIds;
   private String statusText;
-  private Boolean maybeSendEmail;
+  private Boolean shouldSendEmail;
```
ssandbekkhaug, line 10: Question: should this be primitive `boolean` with lowercase "b" instead of Class `Boolean`?

IIRC, a Java primitive is "pass by value". A class is "pass by reference", which would allow code outside this class to change `shouldSendEmail`'s value (and therefore introduce subtle bugs)

server/test/controllers/admin/AdminApplicationControllerTest.java
```diff
@@ -164,6 +165,35 @@ public void index() throws Exception {
...
+            /* page= */ Optional.of(1), // Needed to skip redirect.
+            /* fromDate= */ Optional.empty(),
+            /* untilDate= */ Optional.empty(),
+            /* applicationStatus= */ Optional.empty(),
+            /* selectedApplicationUri= */ Optional.empty(),
+            /* showDownloadModal= */ Optional.empty(),
+            /* message= */ Optional.empty());
+    assertThat(result.status()).isEqualTo(OK);
+    // check if the page size is 100
+    assertThat(contentAsString(result)).contains("100");
```
ssandbekkhaug, line 194: Question (blocking): Is it possible to assert on something more specific than "100"? (Or will this be better tested in #9017?)

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -255,4 +305,71 @@ private Optional<ApplicationModel> validateProgram(
     }
     return application;
   }
+
+  /**
+   * Sets the status on the give applications
```
ssandbekkhaug, line 310: Suggestion: "Sets the status on the given applications. Assumes all applications are for the same program."

Alternatively, validate this assumption and throw an exception if it fails

server/app/services/applications/ProgramAdminApplicationService.java
```diff
@@ -255,4 +305,71 @@ private Optional<ApplicationModel> validateProgram(
...
+            .filter(s -> s.statusText().equals(newStatusText))
+            .findFirst();
+    if (statusDefMaybe.isEmpty()) {
+      throw new StatusNotFoundException(newStatusText, program.id);
+    }
+    Status statusDef = statusDefMaybe.get();
+
+    // Send email if requested and present.
+    if (sendEmail) {
+      if (statusDef.localizedEmailBodyText().isEmpty()) {
```
ssandbekkhaug, line 343: Suggestion (non-blocking): Extract sending emails into its own method

server/test/forms/admin/BulkStatusUpdateFormTest.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Test
+  public void bindFromRequest_hasAllFields() {
+    List<Long> idList = new ArrayList<>();
+    idList.add(1L);
+    idList.add(2L);
+    idList.add(3L);
+    var paramList = idList.stream().map(e -> String.valueOf(e)).collect(Collectors.toList());
+    ;
```
ssandbekkhaug, line 32: Super nit: remove extra ;

server/app/views/admin/programs/ProgramApplicationTableView.java
```diff
@@ -0,0 +1,538 @@
+package views.admin.programs;
```
ssandbekkhaug, line 1: Non-blocking (no action required): This should be tested in a browser test, but I see there are issues to finish adding tests.

server/test/controllers/admin/AdminApplicationControllerTest.java
swatkat1: Good catch - yes asserting the status update form is present (checking by the form id)

server/app/services/applications/ProgramAdminApplicationService.java
swatkat1: In line 260 - we do check it 
`  /* Validates that the given application is part of the given program. */
  private void validateApplications(
  `

server/app/services/applications/ProgramAdminApplicationService.java
swatkat1: good idea! I now globalized it for both single application status update and bulk status update :)

server/app/services/applications/ProgramAdminApplicationService.java
ssandbekkhaug: If I'm reading this correctly, validation happens before calling `setStatus`. I'm concerned that could cause a bug.

1. If a future piece of code calls `setStatus` without validating the applications, what happens?
2. If that would cause a bug/error, should this method have a comment warning the caller?
3. To remember to test this case, should we add a note to the GH issue that will add unit tests for this method?

ssandbekkhaug (approved):

server/app/services/applications/ProgramAdminApplicationService.java
swatkat1: You got a great point. If I remember correctly, when we started adding controllers and service classes, we made a decision that controllers should not be doing validation, and they are the service class responsibility. Sticking to this original plan, I moved the validation of applications to service class, and the validation methods will be called by the service class. So, all the controller is doing is handing a list of applicationIds to the service class, it will validate and do the rest of the activities. Controller will catch any exception and let the user know if any error has occurred. I followed what the current SetStatus() is doing without thinking much of it. But you pointed a great inefficiency. I will add a point to fix the single status update to the upcoming workitem.

ssandbekkhaug (approved):

PR #8869, opened 2024-10-15
Add an Index to applications ON submit_time and id.

lagosr-google:
### Description

PR #8813 adds a query that sorts the application list by submit time (desc) and id (desc).

That query will perform poorly without an Index over the same columns and order.

This change adds an Index to Applications ON submit_time (desc) and id (desc).

NOTE TO REVIEWERS:  If we are concerned about the applications table being locked while the index is built when the evolution is applied, it is possible to do a CONCURRENT index creation, which allows CONCURRENT read/write to the application table.  (The cost is that creating the index takes about 2x as long)  See: https://www.postgresql.org/docs/9.4/sql-createindex.html

## Release notes

While the index is being created, the 'applications' table will be marked as READ_ONLY by postgres.  Attempts to write to the table will block while waiting to get a writer lock. (and most likely eventually time-out).

The time that it takes to build this index is directly proportional to the size of your applications table.   The likelihood that this would be a problem for a deployment with < 100,000 applications is low.

#### General

#### Database evolutions

- [x] Assigned two reviewers
- [x] Data migrations aren't being done (please use a [Durable Job](https://github.com/civiform/civiform/wiki/Database#durable-jobs-for-data-updates) if doing a data migration)

Verified that the index exists after the evolution is applied during job startup by using pgadmin

### Issue(s) this completes

Regards #8743


gwendolyngoetz:
Play applies evolutions synchronously and won’t move on to starting the application until they are all complete. I'm not concerned about downtime, it's not going to take all that long to index. Seattle has less then 50k records in the `applications` table and no one will have more than that.

avaleske (approved):

PR #8879, opened 2024-10-15
Don't run two checks on PRs from forks

nb1701:
Neither of these are critical when running checks of PRs from forks.

### DIFF
--- a/.github/workflows/auto_request_review.yml
+++ b/.github/workflows/auto_request_review.yml
@@ -10,9 +10,7 @@ jobs:
   auto-request-review:
     name: Auto Request Review
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/auto_request_review.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     name: Auto Request Review
     runs-on: ubuntu-latest
     # Only run this for the main civiform owner. PRs from forks don't have the permissions
```
gwendolyngoetz, line 13: Can you delete this line too

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

PR #8927, opened 2024-10-17
[North Star] Create My Applications section on the homepage

### LINKED ISSUE #7814: [North Star Applicant UI] Create the “My Applications” section on the homepage
Migrate the My Applications section of the homepage to Thymeleaf.  This section will hold programs that the resident has applied to or has started applying to.  This issue does not include styling.

#### Additional Context
[TDD](https://docs.google.com/document/d/1hNL6r1JdauL5xjGRIw1dlDD3m81h9g0vUSiqlXOGDCs/edit?usp=sharing)

#### Done when:
- [ ] The “In progress” or “Submitted” sections are combined into one "My Applications" section
- [ ] There is a count next to the header
- [ ] The cards in this section don't include images or category tags.
- [ ] The "My applications" section should be above the common intake form if there is one.

melanie-exygy:
### Description

With the North Star flag enabled, on the homepage, any programs for which an application has been submitted or started now show in a section at the top with the heading "My applications".  This section goes above the intake form (pre-screener) if there is one.  Cards in this section don't show images.

Sections now also show a count next to the section heading.  Also, sections only have one heading, an `h2`.  No more nested section headings.

This PR doesn't address styling since mocks aren't finalized yet. 

1. Enable the North Star flag.
2. As an admin, create a new program that is a pre-screener.
3. Also create a couple of other programs with images, if you don't have any yet.
4. Logout.
5. Go to the homepage.
6. Start applying to a program (not the pre-screener). Don't finish applying and go back to the homepage.
7. The program should be in the "My applications" section, above the pre-screener.  It should no longer show the image.
8. Apply to the other program and submit.
9. That program should be in the "My applications" section too.

### Issue(s) this completes

Fixes #7814


browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -618,11 +618,15 @@ test.describe('applicant program index page', () => {
       }) => {
         await loginAsTestUser(page)
 
-        await test.step('Programs start in not started', async () => {
-          await applicantQuestions.expectPrograms({
-            wantNotStartedPrograms: [primaryProgramName, otherProgramName],
-            wantInProgressPrograms: [],
-            wantSubmittedPrograms: [],
+        await test.step('Programs start in not Programs and Services section', async () => {
```
ssandbekkhaug, line 621: Nit: delete 'not' from this string?

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
```diff
@@ -60,26 +60,29 @@ public ProgramSectionParams getSection(
...
-                    request,
-                    messages,
-                    buttonText,
-                    programData,
-                    preferredLocale,
-                    profile,
-                    applicantId,
-                    personalInfo));
+      ApplicantPersonalInfo personalInfo,
+      boolean isMyApplicationsSection) {
```
ssandbekkhaug, line 64: Question (Blocking): Piping a boolean specifically for the My Applications Section feels a bit awkward. What do you think of using an enum to represent different types of sections or card styles instead?

ssandbekkhaug (changes requested):

browser-test/src/applicant/applicant_application_index.test.ts
melanie-exygy: Good catch!  Done.

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
melanie-exygy: Great idea to use an enum.  Added that.

ssandbekkhaug (approved):

PR #8937, opened 2024-10-17
Pull file upload file type specifiers from config

gwendolyngoetz:
### Description

File upload used a hard-coded accept list of file type specifiers. It was brought to my attention that a new program now live in Seattle requires applicants to submit a spreadsheet filled out from a template supplied by the program.

This change just allows setting the value from the config file.

## Release notes

There is a new, optional, configuration value to control the types of files that applicants are allowed to upload on file upload questions.

The default options remain as images and PDFs. To customize set the `FILE_UPLOAD_ALLOWED_FILE_TYPE_SPECIFIERS`. This uses any valid [file type specifiers](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#unique_file_type_specifiers). Multiple are separated by commas.

#### General

### DIFF
--- a/server/conf/application.conf
+++ b/server/conf/application.conf
@@ -603,3 +603,7 @@ civiform_applicant_enabled_languages = ${?CIVIFORM_APPLICANT_ENABLED_LANGUAGES}
 # Typically, this value is 1.
 num_trusted_proxies = 1
 num_trusted_proxies = ${?NUM_TRUSTED_PROXIES}
+
+# These are file types allowed to be uploaded by the file upload components
+file_upload_allowed_file_type_specifiers = "image/*,.pdf"
+file_upload_allowed_file_type_specifiers = ${?FILE_UPLOAD_ALLOWED_FILE_TYPE_SPECIFIERS}
--- a/server/app/services/settings/SettingsManifest.java
+++ b/server/app/services/settings/SettingsManifest.java
@@ -775,6 +775,15 @@ public final class SettingsManifest extends AbstractSettingsManifest {
   }
 
   /**
+   * Specifies the allowed file types that can be uploaded. Uses any valid [file type
+   * specifiers](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#unique_file_type_specifiers).
+   * Multiple are separated by commas. Default: "image/*,.pdf"
+   */
+  public Optional<String> getFileUploadAllowedFileTypeSpecifiers() {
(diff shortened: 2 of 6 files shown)


server/conf/application.conf
```diff
@@ -603,3 +603,7 @@ civiform_applicant_enabled_languages = ${?CIVIFORM_APPLICANT_ENABLED_LANGUAGES}
 # Typically, this value is 1.
 num_trusted_proxies = 1
 num_trusted_proxies = ${?NUM_TRUSTED_PROXIES}
+
+# These are file types allowed to be uploaded by the file upload components
+file_upload_allowed_file_type_specifiers = "image/*,.pdf"
```
rockycodes, line 608: question (non-blocking): Since the default is set here, do we need the fallback logic in the classes? I'm concerned about the different default places getting out of sync. Also, should you add the default to the dev and test confs?

rockycodes (approved):
Lgtm! Just one question about if we need the defaults in multiple places. I'm also curious if/how you tested this change?

server/conf/application.conf
gwendolyngoetz: It's there only as a last resort in the event someone sets the config value in their deploy to nothing. Better to have the minimum "safe" types so it doesn't completely error out. I'm not too worried about the java classes getting out of sync since long term the non-northstar version will go away.

We don't need to add them to dev/test conf files. They all inherit it from application.conf.

rockycodes: Gotcha. I'm not worried about the java classes getting out of sync, but the conf file getting out of sync with the list in the northstar class for example. Maybe add a comment in the java class to keep that list in sync with the one in the conf file?

gwendolyngoetz:
> Lgtm! Just one question about if we need the defaults in multiple places. I'm also curious if/how you tested this change?

I tested it manually in these system states:
- multi-file upload off and northstar off (so the current single file upload)
- with multi-file upload on and northstar off
- with multi-file upload on and northstar on

In search system state, I opened a program with a file upload and clicked upload to show the file upload dialog checking that the files I want were listed and that I could upload them. Then added an addition file type, refreshed the page, and did the same process to make sure I could see and upload the new type.

server/conf/application.conf
gwendolyngoetz: Sure. I added a note on the northstar class. Not going to bother with the other one.

rockycodes (approved):

PR #8984, opened 2024-10-22
Automatic response for community contributors

melanie-exygy:
### Description

This is a Github action that will respond automatically to comments made by community contributors, inviting them to consult our Wiki page.

### DIFF
--- a/.github/workflows/contributor_response.yml
+++ b/.github/workflows/contributor_response.yml
@@ -1,4 +1,8 @@
-name: Auto-response to Outside Contributors
+# Runs on new comments and prints a welcome message if the commenter is not a bot,
+# is not an existing contributor, and has not already gotten a reply
+
+name: Community Contributor Response
+run-name: Auto-response to Community Contributor Comments
 
 on:
   issue_comment:
@@ -10,7 +14,7 @@ jobs:
     steps:
       - name: Check if commenter is a contributor
         id: check-if-commenter-is-a-contributor
-        uses: actions/github-script@v6
+        uses: actions/github-script@v7
         with:
           script: |
             const { owner, repo } = context.repo;
@@ -30,7 +34,7 @@ jobs:
             return !isBot && !isContributor && !hasReceivedResponse;
       - name: Add a comment if not a contributor and hasn't received a response
         if: steps.check-if-commenter-is-a-contributor.outputs.result === 'true'
-        uses: actions/github-script@v6
+        uses: actions/github-script@v7
         with:
           script: |
             github.rest.issues.createComment({
--- a/.github/workflows/contributor_response.yml
+++ b/.github/workflows/contributor_response.yml
@@ -1,0 +1,42 @@
+name: Auto-response to Outside Contributors
+
+on:
+  issue_comment:
+    types: [created] # Ensure it only triggers on comment creation
+
+jobs:
+  auto-response:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if commenter is a contributor
+        id: check-if-commenter-is-a-contributor
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const issue_number = context.issue.number;
+            const commenter = context.payload.comment.user.login;
+            const bots = ['dependabot[bot]', 'renovate[bot]', 'codecov', 'graphite'];
+            const isBot = bots.includes(commenter);
+
+            // Check if the commenter is a contributor
+            const contributors = await github.repos.listContributors({ owner, repo });
+            const isContributor = contributors.data.some(contributor => contributor.login === commenter);
+
+            // Check if the commenter has already received a response on this issue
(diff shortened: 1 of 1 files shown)


.github/workflows/contributor_response.yml
```diff
@@ -0,0 +1,41 @@
+name: Auto-response to Outside Contributors
```
gwendolyngoetz, line 1: (recommend, non-blocking) One place `name` is used is in the bar [listing all the workflows](https://github.com/civiform/civiform/actions). Since this bar can end up truncating the text I recommend making `name` slightly shorter.

(recommend, non-blocking) I also recommend setting the `run-name` this value can be longer and is shown in the details of the running job. If set it replace `name` in many places.

.github/workflows/contributor_response.yml
```diff
@@ -0,0 +1,41 @@
...
+  issue_comment:
+    types: [created] # Ensure it only triggers on comment creation
+
+jobs:
+  auto-response:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if commenter is a contributor
+        id: check-if-commenter-is-a-contributor
+        uses: actions/github-script@v6
```
gwendolyngoetz, line 13: (praise) This is interesting and new to me. FYI, most/all of our actions are typically shell based commands, but this looks like a good use case for `github-script`. :+1:

gwendolyngoetz, line 13:
```suggestion
        uses: actions/github-script@v7
```
(blocking) Updating to the latest version

.github/workflows/contributor_response.yml
```diff
@@ -0,0 +1,41 @@
...
+            const isContributor = contributors.data.some(contributor => contributor.login === commenter);
+
+            // Check if the commenter has already received a response on this issue
+            const comments = await github.issues.listComments({ owner, repo, issue_number });
+            const hasReceivedResponse = comments.data.some(comment => comment.user.login === 'github-actions[bot]' && comment.body.includes(commenter));
+
+            return !isBot && !isContributor && !hasReceivedResponse;
+      - name: Add a comment if not a contributor and hasn't received a response
+        if: steps.check-if-commenter-is-a-contributor.outputs.result === 'true'
+        uses: actions/github-script@v6
```
gwendolyngoetz, line 33:
```suggestion
        uses: actions/github-script@v7
```
(blocking) Updating to the latest version

.github/workflows/contributor_response.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,   
+              body: 'Thanks for your interest in contributing to CiviForm! You can find info on how to get started here:  https://github.com/civiform/civiform/wiki/Technical-contribution-guide#outside-contributors-not-part-of-exygy-google-or-a-civic-entity.'
```
gwendolyngoetz, line 40: (suggestion) We should probably add a comment to the wiki on the heading for this section to remind people that if it changes this link will break. https://github.com/civiform/civiform/wiki/Technical-contribution-guide#outside-contributors-not-part-of-exygy-google-or-a-civic-entity

gwendolyngoetz (changes requested):
My main concern is that any time someone from a Civic Entity (that hasn't committed code) replies to an issue they'll constantly get this auto message. For now though let's see how it goes.

gwendolyngoetz:
(praise) Overall good job

.github/workflows/contributor_response.yml
melanie-exygy: I thought this was a good idea, but don't know where to put the comment because the fragment wording is defined in the Table of Contents, so it seems weird to put a comment there.  Hm...

.github/workflows/contributor_response.yml
melanie-exygy: Done.

.github/workflows/contributor_response.yml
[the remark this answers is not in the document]
melanie-exygy: Good idea!  Done.

melanie-exygy: I didn't know about `run-name`.  Thanks!  I've shortened the name and added a `run-name`.

.github/workflows/contributor_response.yml
gwendolyngoetz: The fragments are defined by the specific H-tag. I added an html comment on it for you https://github.com/civiform/civiform/wiki/Technical-contribution-guide/_compare/57701d2a75fc72511ebae181e65f2162ad92098a

.github/workflows/contributor_response.yml
gwendolyngoetz: Awesome! Most of our older ones don't use them because people didn't know about it.

gwendolyngoetz (approved):

PR #8995, opened 2024-10-22
[North Star Applicant UI] Display relevant program categories as non-functional filter chips on homepage

melanie-exygy:
### Description

Displaying the program category filter chips on the homepage.  These chips don't do any filtering yet and only have basic styles.  What's important is that the correct categories are shown.  A category should only show if:
- It is on an active program.
- The program it is on is visible to the applicant (not shown to TIs only, for example).
- The program it is on hasn't been applied to yet (the program isn't in the "My applications" section).

This is behind the North Star flag.

image

1. Go to the Dev Tools -> Additional Tools.
2. Seed programs and categories, if you haven't already.
3. Log in as a CF Admin.
4. Edit a program or two and add some categories to them.
5. Publish the programs.
6. Go to the homepage and see the categories you selected as filter chips.
7. Start applying to a program that has a category that isn't on other programs.
8. Go back to the homepage.
9. Notice that that category isn't in the filter chips anymore.


server/app/views/applicant/ProgramIndexTemplate.html
```diff
@@ -82,6 +82,30 @@ <h2 th:text="#{title.getStartedSection}"></h2>
             th:replace="~{applicant/ProgramCardsSectionFragment :: cardsSection(${commonIntakeSection.get()})}"
           ></div>
         </div>
+        <div
```
melanie-exygy, line 85: **note:** I'll probably put this in a fragment eventually, but am leaving it here until I've figured out the actual program filtering.  I don't want to complicate that by adding a fragment yet.

ssandbekkhaug (approved):
The browser test is easy to understand. Nice work!

Issue #8944, opened 2024-10-18
Clicking "Continue" on an in-progress application always navigates to page 1


ssandbekkhaug:
**Describe the bug**
Clicking "Continue" on an in-progress application always navigates to page 1

**To Reproduce**

1. Use CiviForm as an applicant
2. Start an application with 3 or more pages (e.g. Comprehensive Sample Program)
3. Complete page 1
4. Click continue
5. Complete page 2
6. Click continue
7. Navigate to home page
8. On the home page, find the same program card
9. Click "Continue"

**Expected behavior**
User navigates to block 3 (since blocks 1 and 2 are completed)

**Actual behavior**
User navigates to block 1

ssandbekkhaug:
Might be related to https://github.com/civiform/civiform/issues/8876

Issue #1555, opened 2021-07-20
Static question type not visible on published forms


elisekalstad:
**Describe the bug**
The static question type does not display on the published forms. I am able to create a static question (i.e. with our Privacy Statement on it) and publish it to a program. But when residents/TIs open the form, they don't see any text. The form "skips" over the static text and goes straight to the next section of questions.

**To Reproduce**
1. Create a static question (i.e. text from the privacy statement)
2. Drag the question onto a form (in Admin view)
3. Publish form (in Admin view)
4. Login as a user/resident and start form (i.e. the Seattle Preschool Program)
5. Static text will not appear

**Expected behavior**
I would expect the static text to appear.

**Screenshots**
Attached

**Desktop (please complete the following information):**
Chrome
[Static Question Issue.docx](https://github.com/seattle-uat/civiform/files/6851715/Static.Question.Issue.docx)

elisekalstad:
[Static Question Issue.pdf](https://github.com/seattle-uat/civiform/files/6887168/Static.Question.Issue.pdf)

ettengererin:
getInProgressBlocks() seems like a safe place to add logic for showing static questions. We can check if the block has one and only one question that is static and if so, show that block. Static questions isAnswered() always returns true, so the user should still be able to progress in the flow without getting stuck.

https://github.com/seattle-uat/civiform/blob/67d108af88d5877febb42e2b55c1d59a4e07726c/universal-application-tool-0.0.1/app/services/applicant/ReadOnlyApplicantProgramServiceImpl.java#L65

lindsayzad:
We have good working code for this with unit tests.  Need clarity on whether blocks that have previously answered questions and static questions should always be shown as well, or only block with purely static questions.  After that is answered we can check in a PR.
browser-test that covers this is still TBD as well

ettengererin:
We need help from @elisekalstad on how to handle the following cases. We think we know what to do for case 1 and 2, but 3 is still unanswered.

1. Screen with **only** static questions --> always show this screen including the static questions
2. Screen with 1 or more static questions and 1 or more **incomplete** questions --> always show this screen including the static questions
3. Screen with 1 or more static questions and 1 or more **complete** questions --> ???

elisekalstad:
@ettengererin -- Is this just on the "Review" page? If so, I agree on #1 & 2, #3 don't show static text:

Screen with only static questions --> **always show this screen including the static questions**
[...]

PR #9003, opened 2024-10-23
Fix navigation when applying to a program with admin-specified block order

### LINKED ISSUE #8876: [Bug] Application skips to screen created first
1. Create a program with screens 1 and 2. 
2. Swap screens 1 and 2, such that screen 2 appears first.

image

3. As a resident, apply to the program. 
4. Program skips immediately ahead to screen 1 instead of screen 2 first. 
5. (This doesn't happen without NS flag on).

image

ssandbekkhaug:
### Description

Fix navigation when applying to a program with admin-specified block order. (See #8876 and #8944 for details)

#### General

#### User visible changes

To manually test, follow the instructions in #8876 and #8944

### Issue(s) this completes

Fixes #8876 and #8944

### DIFF
--- a/browser-test/src/applicant/applicant_application_index.test.ts
+++ b/browser-test/src/applicant/applicant_application_index.test.ts
@@ -907,7 +907,7 @@ test.describe('applicant program index page with images', () => {
           programName,
           /* northStarEnabled= */ true,
         )
-        await applicantQuestions.clickContinue()
+        await applicantQuestions.clickSubmitApplication()
         await applicantQuestions.gotoApplicantHomePage()
       })
 
--- a/browser-test/src/applicant/applicant_prevent_duplicate_submission.test.ts
+++ b/browser-test/src/applicant/applicant_prevent_duplicate_submission.test.ts
@@ -102,7 +102,6 @@ test.describe('Prevent Duplicate Submission', () => {
 
       await test.step('Submit another application to the same program without changing anything', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
--- a/browser-test/src/applicant/northstar_common_intake_upsell.test.ts
+++ b/browser-test/src/applicant/northstar_common_intake_upsell.test.ts
@@ -53,7 +53,6 @@ test.describe(
 
       await test.step('Setup: submit application', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
@@ -96,7 +95,6 @@ test.describe(
 
       await test.step('Setup: submit application', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
@@ -128,7 +126,6 @@ test.describe(
 
       await test.step('Setup: submit application', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
@@ -159,7 +156,6 @@ test.describe(
 
       await test.step('Setup: submit application', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
@@ -208,7 +204,6 @@ test.describe(
       await test.step('Setup: submit application', async () => {
         await tiDashboard.clickOnViewApplications()
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
         await applicantQuestions.submitFromReviewPage(
           /* northStarEnabled= */ true,
         )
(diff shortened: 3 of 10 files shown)


ssandbekkhaug:
* **#9003** <a href="https://app.graphite.dev/github/pr/civiform/civiform/9003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -886,10 +886,25 @@ test.describe('applicant program index page with images', () => {
...
+      page,
+      adminPrograms,
+      adminQuestions,
+      adminProgramImage,
+      applicantQuestions,
+    }) => {
       const programName = 'Wide Image Program'
       await loginAsAdmin(page)
+
+      await adminQuestions.addPhoneQuestion({
```
melanie-exygy, line 899: **question (non-blocking):** I don't understand why we're adding this question to the test and how it's related to these bug fixes.  Could you explain?

browser-test/src/applicant/applicant_prevent_duplicate_submission.test.ts
```diff
@@ -102,7 +102,6 @@ test.describe('Prevent Duplicate Submission', () => {
 
       await test.step('Submit another application to the same program without changing anything', async () => {
         await applicantQuestions.clickApplyProgramButton(programName)
-        await applicantQuestions.clickContinue()
```
melanie-exygy, line 105: **question (non-blocking):** Are you removing these `clickContinue()` because it's an unnecessary step?

browser-test/src/applicant/northstar_common_intake_upsell.test.ts
```diff
@@ -199,7 +195,7 @@ test.describe(
           lastName: 'last',
           dobDate: '2021-06-10',
         }
-        await tiDashboard.createClient(client, true)
+        await tiDashboard.createClient(client)
```
melanie-exygy, line 198: **question (non-blocking):** Is this just some general test clean-up?  I'm not seeing how it's related to this PR.

browser-test/src/applicant/northstar_ineligible.test.ts
```diff
@@ -189,6 +189,7 @@ test.describe('North Star Ineligible Page Tests', {tag: ['@northstar']}, () => {
       // Click "Continue" on the program card
       await applicantQuestions.clickApplyProgramButton(programName)
 
+      await applicantQuestions.clickEdit()
```
melanie-exygy, line 192: **question (non-blocking):** I don't understand how this test is passing.  Haven't you already clicked on the program card "Edit" button before that in `applicantQuestions.clickApplyProgramButton`?  
image
The "Edit" button meets the requirements of that locator, no?

melanie-exygy (commented):
When manually testing your branch, if I fill out page 1 of the Comprehensive Sample Program, click "Continue" and then click the Civiform logo in the header to return to the landing page, then when I click "Edit", I'm sent back to page 1 of the form.  Shouldn't I be sent to page 2 since I already completed page 1?

ssandbekkhaug:
Good question. The comprehensive sample program has a static block on page 1. Static blocks are always shown first. See https://github.com/civiform/civiform/issues/1555.

When I was implementing this PR, I encountered the same issue. I started rewriting the logic and it caused a lot of side effects. I think we need product input before we change how we handle static text questions.

melanie-exygy:
> Good question. The comprehensive sample program has a static block on page 1. Static blocks are always shown first. See #1555.
> 
> When I was implementing this PR, I encountered the same issue. I started rewriting the logic and it caused a lot of side effects. I think we need UX/product input before we change how we handle static text questions.

Ah, ok.  That makes sense.  I agree that we need UX/Product input for the static text questions.

browser-test/src/applicant/applicant_prevent_duplicate_submission.test.ts
melanie-exygy: Ah, now I get it.  Thanks for explaining.

browser-test/src/applicant/northstar_common_intake_upsell.test.ts
melanie-exygy: No, not worth removing it.  Let's keep it in here.

browser-test/src/applicant/northstar_ineligible.test.ts
melanie-exygy: So much easier to understand.  Thanks a bunch!

browser-test/src/applicant/applicant_application_index.test.ts
melanie-exygy: Oh, I see.  That makes sense now.  Rather than adding the question and answering it, what if the applicant directly clicked on "Submit application" from the review page?  Then, the card would still be put in the "My applications" section since the application would be complete.  I think that would detract less from the purpose of the test.  If that's a pain to do, then we can keep the new question.

melanie-exygy (approved):
Thanks for fixing these bugs!  This is going to make a big difference for the application experience.

browser-test/src/applicant/applicant_application_index.test.ts
ssandbekkhaug: Oh, that works, and it makes the test simpler. Thanks for the suggestion :)

ssandbekkhaug:
> When manually testing your branch, if I fill out page 1 of the Comprehensive Sample Program, click "Continue" and then click the Civiform logo in the header to return to the landing page, then when I click "Edit", I'm sent back to page 1 of the form. Shouldn't I be sent to page 2 since I already completed page 1?

Pasting reply to resolve this conversation

Good question. The comprehensive sample program has a static block on page 1. Static blocks are always shown first. See https://github.com/civiform/civiform/issues/1555.

When I was implementing this PR, I encountered the same issue. I started rewriting the logic and it caused a lot of side effects. I think we need UX/product input before we change how we handle static text questions.

melanie-exygy (approved):

shreyachatterjee00:
Confirmed (with input at PDC) that we want to take applicants to wherever they left off (eg, the first non completed question) and NOT wherever this is static text

PR #9005, opened 2024-10-23
[Program Migration] handle server errors

### LINKED ISSUE #8653: [Program Migration] Handle server errors
Several server errors are being swallowed and no error is shown to the user. These include:
- changing the language to one that doesn't exist
- changing the id of a question to one that already exists or random number (null pointer exception)

See slack for details about exceptions: 
https://civiform.slack.com/archives/C07KXU02B70/p1726678882148489
https://civiform.slack.com/archives/C07KXU02B70/p1726679184681119

rockycodes:
### Description

This is a bug fix to surface a user facing an error when a server error occurs. Previously nothing would show and it would just look like nothing was happening.

There were two cases during bug bash where testers triggered a server error:
- trying to save a program with a language that doesn't exist
- trying to save a program where the question id on the question and in the block definition do not match

This fix handles both of those cases and should handle any other server errors that occur when attempting to render the program preview.

#### General

1. Enable program migration flag
2. Copy program json into the import box
3. Edit the question id in the block definition to be different than the question ids on the questions
4. Click "Preview program"
5. See error message

### Issue(s) this completes

Fixes #8653;

### DIFF
--- a/server/test/controllers/admin/AdminImportControllerTest.java
+++ b/server/test/controllers/admin/AdminImportControllerTest.java
@@ -999,408 +999,416 @@ public class AdminImportControllerTest extends ResetPostgres {
   }
 
   public static final String PROGRAM_JSON_WITHOUT_QUESTIONS =
-      "{\n"
-          + "  \"program\" : {\n"
-          + "    \"id\" : 9,\n"
-          + "    \"adminName\" : \"no-questions\",\n"
-          + "    \"adminDescription\" : \"\",\n"
-          + "    \"externalLink\" : \"https://www.example.com\",\n"
-          + "    \"displayMode\" : \"PUBLIC\",\n"
-          + "    \"notificationPreferences\" : [ ],\n"
-          + "    \"localizedName\" : {\n"
-          + "      \"translations\" : {\n"
-          + "        \"en_US\" : \"Program With No Questions\"\n"
-          + "      },\n"
-          + "      \"isRequired\" : true\n"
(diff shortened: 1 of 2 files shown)


server/test/controllers/admin/AdminImportControllerTest.java
```diff
@@ -1874,4 +1893,101 @@ public void hxSaveProgram_noDuplicatesEnabled_addsAnEmptyStatus() {
...
+          + "      },\n"
+          + "      \"validationPredicates\" : {\n"
+          + "        \"type\" : \"name\"\n"
+          + "      },\n"
+          + "      \"id\" : 1,\n"
+          + "      \"universal\" : false,\n"
+          + "      \"primaryApplicantInfoTags\" : [ ]\n"
+          + "    }\n"
+          + "  } ]\n"
+          + "}";
```
gwendolyngoetz, line 1992: (non-blocking) Do you mind swapping this to the """ based [multiline string](https://www.baeldung.com/java-multiline-string)? It'll be easier to read and maintain.

gwendolyngoetz (changes requested):

server/test/controllers/admin/AdminImportControllerTest.java
rockycodes: done!

gwendolyngoetz (approved):
+1 for extra effort changing all the multiline strings!!

PR #9061, opened 2024-10-29
[Eligibility Message] Add localizedEligibilityMessage to block definition

azhangggggg:
### Description

Add localizedEligibilityMessage to block definition.
Add method setBlockEligibilityMessage() to update eligibility message for a block in ProgramService.

#### General

### Issue(s) this completes

#9142


server/app/services/program/ProgramService.java
```diff
@@ -1305,6 +1305,34 @@ public ProgramDefinition removeBlockPredicate(long programId, long blockDefiniti
...
+
+    BlockDefinition blockDefinition =
+        programDefinition.getBlockDefinition(blockDefinitionId).toBuilder()
+            .setLocalizedEligibilityMessage(message)
+            .build();
+    try {
+      return updateProgramDefinitionWithBlockDefinition(programDefinition, blockDefinition);
+    } catch (IllegalPredicateOrderingException e) {
+      throw new RuntimeException(
+          "Unexpected error: setting this eligibility message invalidates another");
```
gwendolyngoetz, line 1332:
```suggestion
       String errMsg = String.format("Setting this eligibility message invalidates another. [programId: %d, blockDefinitionId: %d]", programId, blockDefinitionId);
       throw new RuntimeException(errMsg, e);
```

(blocking) We should include the program and block ids and also add the caught exception to the new one.

gwendolyngoetz (changes requested):

server/app/services/program/ProgramService.java
avaleske: What does "invalidates another" mean?

server/app/services/program/ProgramService.java
```diff
@@ -1305,6 +1305,34 @@ public ProgramDefinition removeBlockPredicate(long programId, long blockDefiniti
...
+      long programId, long blockDefinitionId, Optional<LocalizedStrings> message)
+      throws ProgramNotFoundException, ProgramBlockDefinitionNotFoundException {
+    ProgramDefinition programDefinition = getFullProgramDefinition(programId);
+
+    BlockDefinition blockDefinition =
+        programDefinition.getBlockDefinition(blockDefinitionId).toBuilder()
+            .setLocalizedEligibilityMessage(message)
+            .build();
+    try {
+      return updateProgramDefinitionWithBlockDefinition(programDefinition, blockDefinition);
```
avaleske, line 1329: @gwendolyngoetz could something like this be the source of the null questions?

here we're getting a program definition, potentially from the cache but not necessarily, and then modifying it and re-saving it, and there's a potential for a race condition since it's not all in a transaction. Should we wrap this in a transaction?

gwendolyngoetz: Hmm, yeah maybe! It should definitely get rolled back if it fails if it isn't already.

server/app/services/program/ProgramService.java
gwendolyngoetz: I read it as it invalidates another existing eligibility message, but on a second reading it would be nice to make the text more clear on what the issue is.

server/app/services/program/ProgramService.java
azhangggggg: After offline discussion, we're going to do the transaction after the other parts of eligibility message are done. To track this I filed issue #9277.

server/app/services/program/ProgramService.java
azhangggggg: Yup updated the error message, also added some comments to explain it more.

avaleske (approved):

gwendolyngoetz (approved):

server/test/services/program/ProgramServiceTest.java
```diff
@@ -2208,6 +2208,70 @@ public void setProgramQuestionDefinitionPosition_invalidPosition() throws Except
...
+    ErrorAnd<ProgramBlockAdditionResult, CiviFormError> result =
+        ps.addBlockToProgram(programDefinition.id());
+    Optional<LocalizedStrings> eligibilityMsg =
+        Optional.of(LocalizedStrings.of(Locale.US, "custom eligibility message"));
+
+    ProgramDefinition updatedProgramDefinition = result.getResult().program();
+    BlockDefinition addedBlock = result.getResult().maybeAddedBlock().get();
+
+    ps.setBlockEligibilityMessage(updatedProgramDefinition.id(), addedBlock.id(), eligibilityMsg);
+    assertThat(addedBlock.localizedEligibilityMessage().equals(eligibilityMsg));
```
shane-exygy, line 2250: **note:** This assert is not evaluating because the equals is inside assertThat but it should be chained.

Issue #7756, opened 2024-06-12
Remove ApplicantModel from ApplicantData


nb1701:
When implementing the Primary Applicant Information (PAI) feature, we moved the canonical location of the applicant's primary info from well known paths in the applicant data blob to columns in the applicant database (and therefore fields in ApplicantModel).  We could change almost all the call sites there were using the ApplicantData object to find the applicant's name and DOB to use the ApplicantModel directly, but there were a few that were too cumbersome to try to get the ApplicantModel associated with the ApplicantData.  So we put the ApplicantModel as a field in the ApplicantData object.

This is not so great.  The ApplicantData object can be initialized without the ApplicantModel, so we have to account for the model not being present when we look up PAI info.  This also blurs the lines of what the ApplicantData object is for, which is supposed to only be a handler for the object blob in the applicants table, and we aren't really using it very consistently.

We should either:

1. Remove ApplicantModel from ApplicantData, and fix up all the call sites where we are doing things like ApplicantData.getFirstName() to operate directly on the ApplicantModel object, and figure out how to get the ApplicantModel where we need it when the call site doesn't have access to it currently.
2. Write a PAI wrapper that contains both the ApplicantData and ApplicantModel as fields so that ApplicantData can remain ONLY for the applicant blob (answers to questions) and not PAI, and the wrapper can figure out the right place to grab it from.

Probably prefer #1.  We only have this coupling between PAI and the applicant blob during the transition period where we have to keep carrying around WellKnownPaths in case PAI is not yet enabled.

avaleske:
I also prefer option #1. It's closer to thinking of the PAI questions as "saving to the user's profile" and keeps their "profile" and "application" data separate.

iiuc, the only time we should have to compare the ApplicantData and ApplicantModel info for PAI questions is if
- We detect a PAI answer is different then the corresponding value in the ApplicantModel.
- We're pre-filling PAI questions if their ApplicantModel has a matching value.

Issue #5503, opened 2023-08-31
Add feature flag for Primary Applicant Info Questions


nb1701:
See [Releasing a Feature Behind a Feature Flag](https://github.com/civiform/civiform/wiki/Feature-Flags#releasing-a-feature-behind-a-feature-flag) for more details.

Be sure to keep the [feature flag tracking spreadsheet](https://docs.google.com/spreadsheets/d/149xPiY0WMLAE5be51pvznyytfGVRTzu4QqrBs427DaA/edit#gid=0) updated as the flag lifecycle changes.

### Flag name
PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED

### Checklist

- [x] Feature flag created as `ADMIN_WRITEABLE`, feature guarded with flag in code
- [x] Flag [enabled in dev](https://github.com/civiform/civiform/blob/main/server/conf/application.dev.conf)
- [x] Flag [disabled for browser tests](https://github.com/civiform/civiform/blob/main/server/conf/application.dev-browser-tests.conf)
- [x] Feature fully written and guarded with flag, including unit and browser tests that manipulate the state of the flag as needed.
- [x] Flag [enabled for browser tests](https://github.com/civiform/civiform/blob/main/server/conf/application.dev-browser-tests.conf) when it's ready to be tested. (Note: If the flag is already enabled in the dev config, you can just remove the flag override in the browser tests config because the browser tests config inherits from the dev config.)
- [x] Flag enabled in staging (See [example PR](https://github.com/civiform/civiform-staging-deploy/pull/90))
- [x] Coordinated with @shreyachatterjee00 to notify CEs of new feature and ask them to turn it on and try it out in their staging
- [x] Coordinated with @shreyachatterjee00 to communicate to CEs that we are going to turn the flag on in production by default, ensuring CEs have had a chance to test out the feature and provide feedback.
- [x] Flag type changed to `ADMIN_READABLE` and default changed to true, with PR release notes mentioning the change and what feature it enables. <Edit this issue and note the date it was enabled here>
- [x] Verified feature is enabled in one or more production builds. <Edit this issue and note the date it was enabled here>
- [ ] Remove old code that utilizes Well Known Paths, as well as the preseeded questions **NOTE: This step is unique to PAI**
- [x] Flag enabled in production for at least a month
- [ ] Flag removed and feature code unguarded

PR #9067, opened 2024-10-29
Remove PAI feature flag, remove use of Well Known Paths

nb1701:
This removes the `PRIMARY_APPLICANT_INFO_QUESTIONS_ENABLED` feature flag. This also removes code we no longer need that save values to the well known paths, since we now use the PAI columns in the applicants table for referencing information about the applicant.

We now end up passing around both the `ApplicantModel` and `ApplicantData` objects to several places.  This is a bit awkward, but for now, was the least-screwy solution.  

As part of the PAI implementation, we included the `ApplicantModel` inside the `ApplicantData` object, since we needed to update and calculate things based off the PAI columns, but also fall back to the well known paths in the applicant data blob if there was no data in PAI columns. Now that we are removing well known paths, it doesn't really make sense to keep the `ApplicantModel` inside `ApplicantData`.

In `DateQuestion`, `NameQuestion`, and `EmailQuestion`, we prepopulate answers with PAI info if the user has not answered the question before. To get access to PAI data, the question needs access to `ApplicantModel`. So it's added to `ApplicantQuestion`.

`Block` creates `ApplicantQuestion` objects in `getQuestions()`, so it needs the `ApplicantModel` as well.

In `ReadOnlyApplicantProgramServiceImpl`, the `getBlocks` function creates new `Block`s, so it too needs the `ApplicantModel`.  However, in this class, we are making a copy of the passed in `ApplicantData` and then locking the values to make it immutable.  Previously, we were still doing this, but the `ApplicantModel` inside it was not locked.  That is still the case here when we pass in the `ApplicantModel` separately.  Conceptually, it somewhat violates the notion of `ReadOnlyApplicantProgramService`.  In practice, I don't think it matters too much.

I explored alternatives to passing in only the `ApplicantModel` where possible and deriving the `ApplicantData` from the model where we needed it.  But it's a bit tangled up in the `ReadOnlyApplicantProgramServiceImpl` part.  So for now, we do the awkward thing of passing in both the model and data objects.

This makes the `ApplicantData` associated with an `ApplicantModel` object a bit uncomfortably disjointed. In the future, we should look into perhaps trying to just pass around the `ApplicantModel` and derive the `ApplicantData` from it when needed.

#### General

### Issue(s) this completes

https://github.com/civiform/civiform/issues/5503
https://github.com/civiform/civiform/issues/7756

### DIFF
--- a/server/app/services/settings/SettingsManifest.java
+++ b/server/app/services/settings/SettingsManifest.java
@@ -944,7 +944,6 @@ public final class SettingsManifest extends AbstractSettingsManifest {
 
   /**
    * Upsell: Add programs cards to the confirmation screen that an applicant sees after finishing an
-   * Add programs cards to the confirmation screen that an applicant sees after finishing an
    * application.
    */
   public boolean getSuggestProgramsOnApplicationConfirmationPage(RequestHeader request) {


server/app/services/applicant/ApplicantService.java
```diff
@@ -228,14 +230,19 @@ public CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProg
...
   }
 
   /** Get a {@link ReadOnlyApplicantProgramService} from applicant data and a program definition. */
   public ReadOnlyApplicantProgramService getReadOnlyApplicantProgramService(
       ApplicantData applicantData, ProgramDefinition programDefinition) {
+    ApplicantModel applicant = new ApplicantModel();
+    applicant.setApplicantData(applicantData);
     return new ReadOnlyApplicantProgramServiceImpl(
-        jsonPathPredicateGeneratorFactory, applicantData, programDefinition);
+        jsonPathPredicateGeneratorFactory, applicant, applicantData, programDefinition);
```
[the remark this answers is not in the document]
rockycodes: Ah I gotcha. That makes sense!

server/app/durablejobs/DurableJobName.java
```diff
@@ -13,7 +13,6 @@ public enum DurableJobName {
   REPORTING_DASHBOARD_MONTHLY_REFRESH("REPORTING_DASHBOARD_MONTHLY_REFRESH"),
   UNUSED_ACCOUNT_CLEANUP("UNUSED_ACCOUNT_CLEANUP"),
   UNUSED_PROGRAM_IMAGES_CLEANUP("UNUSED_PROGRAM_IMAGES_CLEANUP"),
-  MIGRATE_PRIMARY_APPLICANT_INFO("MIGRATE_PRIMARY_APPLICANT_INFO"),
```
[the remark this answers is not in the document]
rockycodes: Yeah I fixed it so we can delete the job, but I think we have to keep the job name around as long as there are entries in the db that haven't been cleaned up. There's a job that clears old jobs out after 6 months, but we really need a better way to handle those annoying leftover job names.

rockycodes (approved):
This lgtm, but I'm curious about your thoughts about Austin's suggestions from [this slack thread](https://civiform.slack.com/archives/C01R8T4EGGG/p1732059909399649?thread_ts=1731965054.245609&cid=C01R8T4EGGG) and also generally what you think the future state should look like. Is it better to head towards handling updates in `ApplicantService` that update the data in both places, or separating the concerns more and checking for PAI higher up when we're pre-filling answers?

server/test/repository/ProgramRepositoryTest.java
```diff
@@ -380,88 +379,8 @@ public void getApplicationsForAllProgramVersions_searchById() {
     assertThat(paginationResult.getNumPages()).isEqualTo(1);
   }
 
-  private static ImmutableList<Object[]> getSearchByNameOrEmailData() {
```
rockycodes, line 383: @nb1701 I think you might also be able to remove the `@RunWith(JUnitParamsRunner.class)` annotation on this test class as well.

rockycodes (approved):

browser-test/image_snapshots/trusted_intermediary_test/pai-program-application-preview.png
avaleske, line 1: I couldn't figure out why the submitted date went away, is it cause it was getting the date from the well known paths question?

server/app/services/applicant/ApplicantService.java
```diff
@@ -228,14 +230,19 @@ public CompletionStage<ReadOnlyApplicantProgramService> getReadOnlyApplicantProg
...
+        jsonPathPredicateGeneratorFactory,
+        application.getApplicant(),
+        application.getApplicantData(),
+        programDefinition);
   }
 
   /** Get a {@link ReadOnlyApplicantProgramService} from applicant data and a program definition. */
   public ReadOnlyApplicantProgramService getReadOnlyApplicantProgramService(
       ApplicantData applicantData, ProgramDefinition programDefinition) {
+    ApplicantModel applicant = new ApplicantModel();
```
avaleske, line 242: not blocking since we've talked at length about this but just reiterating that I don't love that these overloads of `getReadOnlyApplicantProgramService()` actually have different behavior

avaleske (approved):
I didn't look closely at all the tests but thank you!

avaleske (approved):

Issue #9103, opened 2024-11-01
Update category translation handling


dkatzz:
#### Is your feature request related to a problem? Please describe.

Right now, categories are handled in separate messages files, because they weren't supposed to change and they aren't used as static text, but instead as database defined text (similar to admin translation fields). But since they are in a separate messages file, we don't get the benefit of getting this text automatically translated through the transifex integration. We'd like to add these to the messages file and then figure out how we want to update the existing databases to have the new translations (either through a one time job, or by changing the handling of these).

We specifically need to do this so Arkansas gets the new categories they need (https://github.com/civiform/civiform/issues/8485) and to handle new languages that have been added (https://github.com/civiform/civiform/issues/8833)

#### What priority is this?

High - Arkansas needs the new categories by mid november

#### Associated civic entity

Arkansas

PR #9122, opened 2024-11-04
Create a durable job to ensure categories are up to date

dkatzz:
### Description

Category translations were initially designed to be similar to admin translation fields, where the localized strings are stored in the database. This is because the categories weren't supposed to change and we plan to make them admin defined in the future. This PR creates a background job to ensure the translations are updated, since there have been a couple of new categories added. I have it as a job that runs on each startup, so if future categories are added, we can handle them. We could also consider making this a one time job or keeping it a startup job but removing the run once all translations for categories are complete. 

We don't plan to create a rollback plan and a job to undo the data changes, since this isn't any data being used in the applicant flow, but let me know if you think this is needed.

#### General

#### Durable jobs

- [x] Assigned two reviewers
- [x] Included a rollback plan and a job to undo the data changes if appropriate

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/9103

### DIFF
--- a/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
+++ b/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
@@ -78,7 +78,7 @@ public final class AddCategoryAndTranslationsJob extends DurableJob {
               LOGGER.debug("Translation change. Updated database.");
             } catch (JsonProcessingException e) {
               errorCount++;
-              e.printStackTrace();
+              LOGGER.error(e.getMessage(), e);
             }
           }
         }
--- a/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
+++ b/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
@@ -20,7 +20,7 @@ import play.Environment;
 import repository.CategoryRepository;
 import services.LocalizedStrings;
 
-/** Iterates through all categories and ensures translations aren't missing. */
+/* Iterates through all categories and ensures translations aren't missing. */
 public final class AddCategoryAndTranslationsJob extends DurableJob {
   private static final Logger LOGGER = LoggerFactory.getLogger(AddCategoryAndTranslationsJob.class);
   private static final ObjectMapper objectMapper = new ObjectMapper();
@@ -91,7 +91,7 @@ public final class AddCategoryAndTranslationsJob extends DurableJob {
             }
           }
         }
-      } catch (Exception e) {
+      } catch (RuntimeException e) {
         errorCount++;
         LOGGER.error(e.getMessage(), e);
       }
--- a/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
+++ b/server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
@@ -1,0 +1,111 @@
+package durablejobs.jobs;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import controllers.dev.seeding.CategoryTranslationFileParser;
+import durablejobs.DurableJob;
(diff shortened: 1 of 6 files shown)


server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
```diff
@@ -0,0 +1,111 @@
...
+        // Iterate through categories from the messages files and compare with categories in the DB
+        for (CategoryModel messageFileCategory : categoriesFromMessagesFile) {
+          // Get the database category or create one if there is none matching the category from the
+          // messages file
+          CategoryModel dbCategory =
+              categoryRepository.fetchOrSaveUniqueCategory(messageFileCategory);
+          LocalizedStrings dbTranslations = dbCategory.getLocalizedName();
+          LocalizedStrings fileTranslations = messageFileCategory.getLocalizedName();
+          // Update the database if the translations in the messages file differs
+          if (!dbTranslations.equals(fileTranslations)) {
```
melanie-exygy, line 69: **question:** Since this is using the default `Object.equals`, won't `!dbTranslations.equals` always return true, even if the translations are the same, since `dbTranslations` and `fileTranslations` point to two different addresses in memory?

I've just seen that you did check that no updates are made when it isn't necessary in the `run_makesNoUpdatesIfCategoriesMatch` test.  Could you please help me understand how that's working with the default `equals`?

server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
```diff
@@ -0,0 +1,111 @@
...
+            LOGGER.info("Translations mismatch for category ID: {}", dbCategory.id);
+            String updateSql =
+                """
+                update categories
+                set localized_name = CAST(:localized_name AS jsonb)
+                where id = :id
+                """;
```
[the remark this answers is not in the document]
melanie-exygy: I like that idea.

melanie-exygy (commented):
I agree that making this a durable job rather than a one-time job will be more convenient because we won't have to remember to run the job if new languages are added.  Hopefully, though, we can remember to remove the durable job when we finally do make categories editable by admins.

server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
dkatzz: Good question! Its kind of sneaky but since LocalizedStrings uses AutoValue, it uses their override of equals, which checks for the equality of the data, not the addresses in memory. Section 3 in [these docs](https://www.baeldung.com/introduction-to-autovalue) give an explanation of it, but let me know if you have questions.

server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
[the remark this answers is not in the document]
dkatzz: Done!

melanie-exygy (approved):

server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
```diff
@@ -0,0 +1,102 @@
...
+              JsonNode localizedNameToSet =
+                  objectMapper.readTree(objectMapper.writeValueAsString(fileTranslations));
+              categoryRepository.updateCategoryLocalizedName(
+                  dbCategory.id, localizedNameToSet.toString());
+
+              stepTransaction.commit();
+              LOGGER.debug("Translation change. Updated database.");
+            } catch (JsonProcessingException e) {
+              errorCount++;
+              e.printStackTrace();
```
gwendolyngoetz, line 81: (blocking) This should be handled by the logger. If we only ever want it available to log for local development set it as log.debug or log.trace.

gwendolyngoetz (changes requested):

server/app/durablejobs/jobs/AddCategoryAndTranslationsJob.java
dkatzz: Good catch! Updated

gwendolyngoetz (approved):

Issue #8482, opened 2024-09-04
Clicking "remove household member" does not remove the household member


ssandbekkhaug:
clicking "remove household member" didn't remove the household member

[Slack Message](https://civiform.slack.com/archives/C01QKN2UNMA/p1725468719357889?thread_ts=1725468051.919329&cid=C01QKN2UNMA)

Discovered by Rocky.

![Image](https://github.com/user-attachments/assets/7759d485-d58e-46c4-9795-7259a99b199e)

PR #9124, opened 2024-11-04
Allow enum entities to be removed in northstar

dkatzz:
### Description

Enumerator entities weren't being removed in northstar because the element structure was a bit different. This PR fixes the behavior to select the appropriate element and adds tests related to removing and adding entities. 

#### General

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/8482


browser-test/src/end_to_end_enumerators.test.ts
dkatzz, line 1: **note:** These are copied from the non-enum test. I'll plan to migrate this test, but wanted to include the tests related to this (now working) functionality.

browser-test/src/support/applicant_questions.ts
```diff
@@ -214,10 +214,14 @@ export class ApplicantQuestions {
...
-  async editQuestionFromReviewPage(questionText: string) {
-    await this.page.click(
-      `.cf-applicant-summary-row:has(div:has-text("${questionText}")) a:has-text("Edit")`,
-    )
+  async editQuestionFromReviewPage(
+    questionText: string,
+    northStarEnabled = false,
+  ) {
+    const element = northStarEnabled
+      ? `.block-summary:has(div:has-text("${questionText}")) a:has-text("Edit")`
```
melanie-exygy, line 222: **suggestion (non-blocking):** Would you mind using a locator?  I think that would have the advantage of making use of Playwright's auto-waiting and retrying abilities.

server/app/assets/javascripts/enumerator.ts
```diff
@@ -131,7 +138,14 @@ function removeExistingEnumeratorField(event: Event) {
 
   // Hide the field that was removed. We cannot remove it completely, as we need to
   // submit the input to maintain entity ordering.
-  const enumeratorFieldDiv = assertNotNull(removeButton.parentElement)
+  let enumeratorFieldDiv = assertNotNull(removeButton.parentElement)
```
melanie-exygy, line 141: **suggestion (non-blocking):** It might be nice to put this in a helper function since you repeat this loop above.

browser-test/src/support/applicant_questions.ts
dkatzz: Updated!

server/app/assets/javascripts/enumerator.ts
dkatzz: Yes, good call! I didn't originally because one used parentElement and the other used parentNode, but seems both work, so pulled it out into a helper!

melanie-exygy (approved):

PR #9134, opened 2024-11-05
[North Star] Filter programs by category on the landing page

melanie-exygy:
### Description

## Release notes

Remove this section if the title of the pull request can be used as the default release notes description. If more detail is needed to communicate to partners the scope of the PR's changes, use this release notes section (and remove this placeholder text).

#### General

#### Database evolutions

- [ ] Assigned two reviewers
- [ ] Every comment in script should begin with -- and not # --- unless it denotes Ups or Downs. See [here](https://www.playframework.com/documentation/2.9.x/Evolutions) for details.
- [ ] Tested both the Downs and the Ups scripts manually (When testing, include all comments from the evolution in your test script to ensure any syntax errors in the comments are caught.)

#### Durable jobs

- [ ] Assigned two reviewers

#### User visible changes

#### New Features

### Issue(s) this completes


server/app/views/applicant/ProgramIndexTemplate.html
```diff
@@ -82,35 +82,39 @@ <h2 th:text="#{title.getStartedSection}"></h2>
...
+          hx-target="#not-started-programs"
         >
           <fieldset class="grid-row grid-gap">
             <legend th:text="#{label.programFilters}"></legend>
             <th:block th:each="categoryOption : ${categoryOptions}">
-              <div class="filter-chip">
+              <div>
                 <input
-                  th:id="${'check-category-' + #strings.replace(categoryOption, ' ', '-')}"
+                  th:id="${'ns-check-category-' + #strings.replace(categoryOption, ' ', '-')}"
```
gwendolyngoetz, line 97: (question) Why does this need the `ns-` prefix?

server/app/assets/javascripts/applicant_entry_point.ts
```diff
@@ -16,6 +16,15 @@ import * as azureUpload from './azure_upload'
...
 import * as trustedIntermediary from './trusted_intermediary'
+import * as htmx from './htmx'
+
+declare global {
+  interface Window {
+    htmx: typeof htmx
+  }
+}
+
+window.htmx = htmx
```
gwendolyngoetz, line 27:
```suggestion
import htmx from './htmx'
```

(blocking) When I upgraded htmx to v2 a month or two ago I had to I had to jump through a few hoops in order for webpack to correctly build.

You should just need to reference it this way for htmx to correctly get injected by webpack.

gwendolyngoetz (changes requested):

server/app/views/applicant/ProgramIndexTemplate.html
melanie-exygy: I just needed to differentiate it because of the click listener in `main.ts` for the pre-North Star implementation.  I can remove that 'ns' once we've fully moved over to the NS.

server/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -277,4 +281,36 @@ public CompletionStage<Result> showInfoDisabledProgram(Request request, String p
...
+    return CompletableFuture.completedFuture(
+        Results.ok(
+            northStarFilteredProgramsViewPartial.render(
+                messagesApi.preferred(request),
+                request,
+                Optional.empty(),
+                ApplicantPersonalInfo.ofGuestUser(),
+                programsFuture.join(),
+                Optional.empty(),
+                ImmutableList.copyOf(categories))));
```
gwendolyngoetz, line 314:
```suggestion
                ImmutableList.copyOf(categories))).as("text/html"))
            .exceptionally(ex -> {
              logger.error("123", ex);
              return Results.badRequest("asdf").as("text/html")
            });
```

(blocking) My suggestions does two things. It makes sure to return htmx with the `text/html` mimetype since it is possible htmx might incorrectly interpret the result type.

Additionally we should return something that can safely be swapped in in case of an error. You can change the Results.badRequest to any option, even another `ok` is fine if it makes sense.

server/app/views/applicant/FilteredProgramsTemplate.html
```diff
@@ -0,0 +1,19 @@
...
+  </head>
+  <body>
+    <div
+      th:replace="~{applicant/ProgramCardsSectionFragment :: cardsSection(${recommendedSection})}"
+    ></div>
+    <th:block th:if="${otherProgramsSection.isPresent()}">
+      <div
+        th:replace="~{applicant/ProgramCardsSectionFragment :: cardsSection(${otherProgramsSection.get()})}"
+      ></div>
+    </th:block>
```
gwendolyngoetz, line 17: (blocking) Since htmx is swapping in your changes you won't want to include the html and body elements.

server/test/controllers/applicant/ApplicantProgramsControllerTest.java
```diff
@@ -433,6 +433,14 @@ public void edit_redirectsToFirstIncompleteBlock() {
                 .url());
   }
 
+  @Test
+  public void hxFilter_isOk() {
+    Result result =
+        controller.hxFilter(fakeRequest(), ImmutableList.of()).toCompletableFuture().join();
+
+    assertThat(result.status()).isEqualTo(OK);
```
gwendolyngoetz, line 441: (blocking) After adding the exceptionally in the controller please try to test the failure case.

gwendolyngoetz (changes requested):

server/app/controllers/applicant/ApplicantProgramsController.java
melanie-exygy: That makes sense.  I don't think the error response will need to be ("text/html"), though, because, if I understand [the htmx docs](https://htmx.org/docs/#requests) correctly, when they get an error response, the swap doesn't take place.  It just emits a `htmx:sendError` event.  In our case, I think the best thing to do if there's an error in filtering is just to show the unfiltered programs.  So, not swapping would make sense.  Does that sound good to you?

gwendolyngoetz: Yeah that sounds reasonable. Probably return it as a 500 error then .

server/test/controllers/applicant/ApplicantProgramsControllerTest.java
melanie-exygy: As we discussed, I wasn't able to cause an exception in the unit test.  I did test manually, however, and verified that we now log an error message server-side and, client-side, HTMX catches the error, triggers its own error event and doesn't swap the partial.

server/app/views/applicant/FilteredProgramsTemplate.html
melanie-exygy: Oh, great catch!  Done.

server/app/controllers/applicant/ApplicantProgramsController.java
melanie-exygy: Done.

gwendolyngoetz (approved):

PR #9186, opened 2024-11-09
Adding applicant program list class

gwendolyngoetz:
### Description

Adding a new fixture class to represent the applicant program list screen, i.e. `/programs`. Since this is being extracted from my fast forward tests, also update those.

This class is northstar only. I'm not going to label it that way as nothing can/will use it that isn't northstar and it'll just end up being another cleanup task.

I'm calling it a `CardGroup`, that is the container of `Card`s. Calling it CardList seemed like it would imply it's a list type element.

#### General

### DIFF
--- a/browser-test/src/support/applicant_program_list.ts
+++ b/browser-test/src/support/applicant_program_list.ts
@@ -13,104 +13,100 @@ export class ApplicantProgramList {
   }
 
   /**
-   * Get the card group locator for the desired application status card group name
-   * @param {ApplicationStatusCardGroupName} applicationStatusCardGroupName to find
-   * @returns {Locator} Locator to the card group
+   * Get the card section locator for the desired application status card section name
+   * @param {CardSectionName} cardSectionName to find
+   * @returns {Locator} Locator to the card section
    */
-  getCardGroupLocator(
-    applicationStatusCardGroupName: ApplicationStatusCardGroupName,
-  ): Locator {
-    return this.page.getByRole('region', {name: applicationStatusCardGroupName})
+  getCardSectionLocator(cardSectionName: CardSectionName): Locator {
+    return this.page.getByRole('region', {name: cardSectionName})
   }
 
   /**
-   * Get the locator for program card heading in the desired group
-   * @param {ApplicationStatusCardGroupName} applicationStatusCardGroupName to find
+   * Get the locator for program card heading in the desired section
+   * @param {CardSectionName} cardSectionName to find
    * @param {String} programName to find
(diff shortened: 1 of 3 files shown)


browser-test/src/support/applicant_program_list.ts
```diff
@@ -0,0 +1,139 @@
+import {Page, Locator} from '@playwright/test'
+
+/**
+ * Class for interacting with the applicant program list page
```
melanie-exygy, line 4: **thought (non-blocking):** We have so many names for this page.  "Landing page", "homepage", "program index page" and "applicant program list page".  It would be nice if we could start using one name for it.  I'm not a fan of "program index" but, since that's the name of the view and the template, maybe it would be helpful to name this support file accordingly?  I don't feel strongly about this.  Let me know what you think.

browser-test/src/support/applicant_program_list.ts
```diff
@@ -0,0 +1,139 @@
...
+ * @class
+ * @param {Page} page the Playwright page object in which to operate against
+ */
+export class ApplicantProgramList {
+  public page!: Page
```
melanie-exygy, line 9: **praise:** Thanks for including the types.

browser-test/src/support/applicant_program_list.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+}
+
+/**
+ * @readonly
+ * @enum {string}
+ * List of heading names used for different groups of application cards used in this test
+ * suite. This is used in relation to what the user sees on the `/programs` page
+ */
+export enum ApplicationStatusCardGroupName {
```
melanie-exygy, line 125: **suggestion (non-blocking):** We (I) might want to add the 'Programs based on your selections' and 'Other programs and services' card groups.  Could we name the enum `CardGroupName`?  Or, if we want to follow the naming in the templates, we could call it `CardSectionName`.

gwendolyngoetz: Works for me. Much easier to read. Since there is no definition in the designs that I could see on what the collection of cards should be called, I fell back to `CardSectionName` to be consistent with the templates.

browser-test/src/support/applicant_program_list.ts
gwendolyngoetz: I'm going to defer renaming this. Let's check with Sijia and decide on one name.

It doesn't feel like a landing page to me, nor a home page as you get different views depending on who you are logged in as.

melanie-exygy (approved):

PR #9223, opened 2024-11-14
Re-enable Azure browser tests

gwendolyngoetz:
### Description

Removes the disabled Azure job. Makes the AWS job generic. Dynamically builds the strategy matrix used to generate the browser test shards. Normal operation will remain running with localstack/aws. 

If your branch has `azure` in the name if will run 10 shards of the browser tests for AWS and 10 for Azure. This is done as a temporary measure to allow for CI to run Azure tests, but also to not block the rest of development.

All tests will be run now, and certain to fail at first. Once we get all the tests running we can remove the branch naming check and run them all the time.

#### Branch protection

I will have to re-jigger the requirements in the branch protection rules as the new name of the jobs are different.

#### Playwright config

I made some adjustments to the Playwright config. It should still only upload traces for failed tests, not passed ones. But instead of recording on retry we record the first try. This is to allow nixing the retry count which doesn't seem to have been all the helpful. Also trying to lower the 90 second time again. 45s is known to be too small, lets try 80s.

If it doesn't work right I'll have to add some environment variables to override some settings at runtime. Otherwise the Azure tests, which fail, take two hours to run for each shard due to the amount of failures. The coming no-op AzurePublicStorage class will probably help a sizable number of tests to pass.

#### General

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -57,6 +57,8 @@ jobs:
       - name: Run tests
         run: bin/run-ts-tests
 
+  # Dynamically builds the matrix for the next job. Defaults to only aws. If the
+  # branch name contains azure we'll run against both aws and azure.
   setup-browser-test-matrix:
     runs-on: ubuntu-latest
 
@@ -66,6 +68,9 @@ jobs:
     steps:
       - id: build-matrix
         env:
+          # To get the branch name that we want we have to consider the event type. A PR will
+          # use the head_ref whereas a push event (i.e. pushing an approve pr to main) needs
+          # to look at the ref_name.
           BRANCH_NAME: ${{ github.event_name == 'pull_request' && github.head_ref || github.ref_name }}
         run: |
           # Temporarily ignore errors so we can check grep for matches
@@ -86,7 +91,6 @@ jobs:
     needs: setup-browser-test-matrix
     strategy:
       fail-fast: false
-      # Default to only aws. If the branch name contains azure use aws and azure
       matrix: ${{ fromJson(needs.setup-browser-test-matrix.outputs.matrix) }}
     runs-on: ubuntu-latest
     steps:
@@ -139,20 +143,26 @@ jobs:
           CI: true
         run: bin/run-browser-tests-ci --shard=${{ matrix.shard_number }}/${{strategy.job-total}}
 
+      # We'll treat aws as the source of truth for images snapshots. Attempts to run both will
+      # end up with two places trying to update the same images.
       - name: Verify no new image snapshots added
         if: ${{ matrix.cloud == 'aws' }}
         # jest-image-snapshots automatically adds snapshot if it's missing.
         # Use git diff to fail if we detect new images.
         run: git add browser-test/image_snapshots ; git diff --compact-summary --exit-code HEAD
 
+      # Upload image diff files, these are the single image containing the before, after, and diff parts.
+      # We want this for aws and azure so we have access to what happened in failed tests.
       - name: Upload image diff outputs
-        if: ${{ matrix.cloud == 'aws' && failure() }}
+        if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: image diff output directory (${{ matrix.cloud }}) - shard-${{ matrix.shard_number }}-run-${{ github.run_attempt }}
           retention-days: 3
           path: browser-test/diff_output
 
+      # We only want updated snapshots for aws. Otherwise we'll hit conflicts if both aws and azure 
+      # attempt to update the same images
       - name: Upload updated snapshots
         if: ${{ matrix.cloud == 'aws' && failure() }}
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


browser-test/playwright.config.ts
```diff
@@ -25,8 +33,10 @@ export default defineConfig({
     },
   },
   use: {
-    trace: process.env.CI === 'true' ? 'on-first-retry' : 'on',
-    video: process.env.RECORD_VIDEO === 'true' ? 'on-first-retry' : 'off',
+    // retain-on-failure should remove trace and video files when a test succeeds allowing
+    // for minimizing cloud storage needs.
+    trace: process.env.CI === 'true' ? 'retain-on-failure' : 'on',
+    video: process.env.RECORD_VIDEO === 'true' ? 'retain-on-failure' : 'off',
```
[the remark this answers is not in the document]
gwendolyngoetz: When the `RECORD_VIDEO` environment variable is true, the test records the video but discards if the test succeed. That way when the artifact upload step occurs it will only have videos of failed tests to save. This is a github cloud storage saving measure. 

The outcome is the same as before. But instead of running the test, failing running again in recording mode it records on the first run. There is some overhead in recording all the time, but given our test run time it should be negligible.

.github/workflows/tests.yaml
```diff
@@ -103,109 +129,48 @@ jobs:
...
         uses: actions/upload-artifact@v4
         with:
-          name: updated snapshots output directory (aws) - shard-${{ matrix.shard_number }}-run-${{ github.run_attempt }}
+          name: updated snapshots output directory (${{ matrix.cloud }}) - shard-${{ matrix.shard_number }}-run-${{ github.run_attempt }}
           retention-days: 3
           path: browser-test/updated_snapshots
-        if: failure()
 
       - name: Upload test videos on failure
+        if: failure()
```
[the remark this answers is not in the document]
gwendolyngoetz: I'll add a comment when I deal with the comment in your other reply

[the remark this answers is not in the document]
gwendolyngoetz: Yep we do. The one I have set for aws only are the changes to images that are generated by the browser tests for snapshot diffs. Having both will generate extra clutter and possibly end up with conflicts.

.github/workflows/tests.yaml
```diff
@@ -139,20 +143,26 @@ jobs:
...
         if: ${{ matrix.cloud == 'aws' }}
         # jest-image-snapshots automatically adds snapshot if it's missing.
         # Use git diff to fail if we detect new images.
         run: git add browser-test/image_snapshots ; git diff --compact-summary --exit-code HEAD
 
+      # Upload image diff files, these are the single image containing the before, after, and diff parts.
+      # We want this for aws and azure so we have access to what happened in failed tests.
       - name: Upload image diff outputs
-        if: ${{ matrix.cloud == 'aws' && failure() }}
+        if: failure()
```
gwendolyngoetz, line 157: @rockycodes took a second pass to make sure I'm upload the things that make sense. Decided to include the image diffs as we'll want them for quick review and they aren't stored back into the repo like the image snapshots are.

.github/workflows/tests_skip.yaml
```diff
@@ -20,16 +20,37 @@ jobs:
...
     runs-on: ubuntu-latest
+
+    outputs:
+      matrix: ${{ steps.build-matrix.outputs.matrix }}
+
     steps:
-      - run: 'echo "skipping"'
+      - id: build-matrix
+        env:
+          # Need to access the branch name differently between
```
rockycodes, line 32: I think this comment got cut off

.github/workflows/tests_skip.yaml
```diff
@@ -20,16 +20,37 @@ jobs:
...
+      matrix: ${{ steps.build-matrix.outputs.matrix }}
+
     steps:
-      - run: 'echo "skipping"'
+      - id: build-matrix
+        env:
+          # Need to access the branch name differently between 
+          BRANCH_NAME: ${{ github.event_name == 'pull_request' && github.head_ref || github.ref_name }}
+        run: |
+          # Temporarily ignore errors so we can check grep for matches
```
rockycodes, line 35: this looks like it might be getting read as a part of the script and not a comment?

rockycodes (commented):
a couple of comments on the skip.yaml file

.github/workflows/tests_skip.yaml
gwendolyngoetz: Yep it is being interpreted as part of the script. That's on purpose.. I want to maintain locality with the specific `set +e/-e` commands since that can change the behavior of the step if not set correctly.

.github/workflows/tests_skip.yaml
gwendolyngoetz: I don't know how to copy/paste. Fixed

rockycodes (approved):

lagosr-google (approved):

dkatzz (approved):

PR #18, opened 2021-01-14
Integrate Postgres with the Play app using EBean.

hmhsu:
Connect the app to Postgres database in conf/application.conf.

Add EBean models under app/models/ and repository to use them under app/repository.
Add a PostgresController to allow a simple way to interact with Postgres from HTTP.
Add database evolution script for sample data. Also include tests.

Switch sbt version back to 1.3.13 due to issues with the new version (1.4.3).

PR #9323, opened 2024-11-27
Tag multi-file-upload browser tests that fail on Azure, and skip those tests.

### LINKED ISSUE #9293: Run browser tests for azure through github test action

Allow for the ability to run azure tests in the tests workflow. We should probably figure out which file changes should trigger these tests, since it would add a lot of time to the tests, but we should make it possible, so all these tests don't have to be run locally each time.

https://github.com/civiform/civiform/pull/6705/files was the PR to disable these

Allow running azure tests in tests.yaml

lagosr-google:
### Description

Add tag to for multi-file-upload tests that fail on Azure. Skip those tests.

1 - Add tag: @multi-file-upload to failing tests
2 - Detect when tests are running against an Azure server
3 - Set a reverse grep list of tags to pass on to playwright to skip those tests.

#### General

Fixes #9293


browser-test/src/applicant/application_review.test.ts
lagosr-google, line 1: The actual change is 1 line on #18 to add the tag.

All other changes occurred when running bin/fmt over the file.

gwendolyngoetz (changes requested):
I'd like to continue this discussion on Slack before we move forward with blocking tests on specific cloud providers.

bin/run-browser-tests
```diff
@@ -42,12 +42,20 @@ if [[ -z "$(docker ps --filter "name=${CONTAINER_NAME}" --format "{{.Names}}")"
     civiform-browser-test:latest
 fi
 
+# Detect if we are running tests with an Azure server, by looking for the Azurite
+# image in the docker container list.
+IS_AZURE_SERVER=
+if docker ps --filter "ancestor=mcr.microsoft.com/azure-storage/azurite" -q; then
+  IS_AZURE_SERVER=1
+fi
+
```
gwendolyngoetz, line 51: (blocking) This won't work correctly if both the dev and browser test servers are running azure.

gwendolyngoetz (changes requested):
Updating the `wait_for...sh` script like this isn't going to behave consistently. Any ad-hoc use of `--grep-invert` is going to override what is in the script.

The shell scripts already pass through arguments via the `$@` down to Playwright.

Keep the tag on the `application_review.test.ts` file, but the changes to the shell scripts are unneeded.

With the test tagged we just need to modify the [test workflow file](
https://github.com/civiform/civiform/blob/c93a43aa53cae315b3030de774f4aae79ac042cd/.github/workflows/tests.yaml#L144) to conditionally ignore the tag. Something like this 

```yaml
run: |
  if [[ "${{ matrix.cloud }}" == "azure" ]]; then
    bin/run-browser-tests-ci --shard="${{ matrix.shard_number }}/${{ strategy.job-total }}" --grep-invert "@skip-on-azure"
  else
    bin/run-browser-tests-ci --shard="${{ matrix.shard_number }}/${{ strategy.job-total }}"
  fi
```

Running it from the cli manually won't automatically pickup that you want to ignore it, but a bash alias can solve that 

```bash
alias run-browser-tests-azure='bin/run-browser-tests --grep-invert "@skip-on-azure"'
```

There should be a way to have the alias automatically created upon creation of the devcontainer.

browser-test/src/applicant/application_review.test.ts
```diff
@@ -13,385 +13,387 @@ import {
...
-  test('all major steps with multiple file upload flag', async ({
-    page,
-    adminQuestions,
-    adminPrograms,
-    applicantQuestions,
-  }) => {
-    test.slow()
+  test(
+    'all major steps with multiple file upload flag',
+    {tag: ['@multi-file-upload']},
```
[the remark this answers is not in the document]
lagosr-google: done

bin/run-browser-tests
lagosr-google: removed this part.

lagosr-google:
> Updating the `wait_for...sh` script like this isn't going to behave consistently. Any ad-hoc use of `--grep-invert` is going to override what is in the script.
> 
> The shell scripts already pass through arguments via the `$@` down to Playwright.
> 
> Keep the tag on the `application_review.test.ts` file, but the changes to the shell scripts are unneeded.
> 
> With the test tagged we just need to modify the [test workflow file](https://github.com/civiform/civiform/blob/c93a43aa53cae315b3030de774f4aae79ac042cd/.github/workflows/tests.yaml#L144) to conditionally ignore the tag. Something like this
> 
> ```yaml
> run: |
>   if [[ "${{ matrix.cloud }}" == "azure" ]]; then
>     bin/run-browser-tests-ci --shard="${{ matrix.shard_number }}/${{ strategy.job-total }}" --grep-invert "@skip-on-azure"
>   else
>     bin/run-browser-tests-ci --shard="${{ matrix.shard_number }}/${{ strategy.job-total }}"
>   fi
> ```
> 
> Running it from the cli manually won't automatically pickup that you want to ignore it, but a bash alias can solve that
> 
> ```shell
> alias run-browser-tests-azure='bin/run-browser-tests --grep-invert "@skip-on-azure"'
> ```
> 
> There should be a way to have the alias automatically created upon creation of the devcontainer.

Made the change to move the detection of the context and setting of the grep invert to the test definition file.

I think the spirit of the change is to skip this test on the CI run only.. so not making any changes to the manually cli runs by a dev is better imo ..   no need to add an alias imo.

gwendolyngoetz (approved):

gwendolyngoetz:
If you need to skip this on the azure deploy as well just append `--grep-invert "@northstar|@skip-on-azure"` to [this line](https://github.com/civiform/civiform-staging-deploy/blob/982aeefd54d7395899e16a52cd3ca09724b05785/.github/workflows/azure_deploy.yaml#L86).

dkatzz (approved):
Right now, we're running azure tests only if the branch name contains azure. I'm assuming that after this change, we'll want to run azure always? https://github.com/civiform/civiform/blob/5c2115ace7b9c8a72ed33241842baced482961d7/.github/workflows/tests.yaml#L79-L83

gwendolyngoetz:
@dkatzz if we're ready to enable it. I have something already I can push. I just need to rebase it.

dkatzz:
> @dkatzz if we're ready to enable it. I have something already I can push. I just need to rebase it.

Thank you, @gwendolyngoetz !!

Issue #7347, opened 2024-04-30
Remove durable jobs from the db that are not found in the DurableJobName enum at server start


rockycodes:
Currently, we must keep old job names around the `DurableJobName` enum to avoid getting an `IllegalArgumentException` when the job name is not found. It would be better if job names that are not found in the enum are removed from the `persisted_durable_jobs` table and never attempt to run.

It might be possible to do this with the play framework: https://www.playframework.com/documentation/2.9.x/ScheduledTasks#Starting-tasks-when-your-app-starts

PR #9326, opened 2024-11-27
Allow removal of DurableJobName values

nb1701:
### Description

Previously, we had to keep around old durable job names in the DurableJobName enum because if the job still existed in the database, we'd throw an exception looking up the enum value if we didn't. This change allows us to remove the enum value and handle the case where the database has the old job in it, but the enum no longer has that value.

#### General

`bin/run-dev` on a previous version that will add something like `MIGRATE_PRIMARY_APPLICANT_INFO` to the `persisted_durable_jobs` table. Then run this branch, and verify no exceptions are logged and that the job has been removed from the database.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/7347

### DIFF
--- a/server/app/durablejobs/AbstractDurableJobRunner.java
+++ b/server/app/durablejobs/AbstractDurableJobRunner.java
@@ -130,9 +130,9 @@ public abstract class AbstractDurableJobRunner {
     try {
       persistedDurableJob.decrementRemainingAttempts().save();
 
-      Optional<DurableJobName> jobName =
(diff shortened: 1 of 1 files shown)


rockycodes (approved):
Lgtm! Clever solution. I had to update the execution time when I tested this out manually so it would run, but it worked like a charm!

dkatzz (approved):

rockycodes (approved):

PR #9347, opened 2024-12-02
[Eligibility message] Adding eligibility message field on CiviForm admins side

azhangggggg:
### Description

Adding eligibility message field on CiviForm admins side.

A couple of notes below:
1. The text in this PR is not internationalized as it's not applicant-facing.
2. The field is designed to be mark-down compatible, but not in toast message. Markdown will be tested in a PR applicant facing. 
3. This change won't be implemented in North Star as it's admin-facing.

#### General

#### User visible changes

Turn the feature flag on for eligibility message, go to a screen to edit predicate, eligibility message field is visible. Save a new eligibility message, and a toast message pops out with the message.

### Issue(s) this completes

Part of #9370


server/app/views/admin/programs/ProgramPredicatesEditView.java
```diff
@@ -298,6 +312,45 @@ private static LabelTag renderPredicateQuestionCheckBoxRow(
                         .withClasses("mt-1", "text-sm")));
   }
 
+  private DivTag createEligibilityMessageForm(
+      Http.Request request, BlockDefinition block, ProgramDefinition program) {
+    String existingEligibilityMessage =
+        block.localizedEligibilityMessage().map(LocalizedStrings::getDefault).orElse("");
```
melanie-exygy, line 318: **question:** (maybe blocking) Shouldn't we use [`LocalizedStrings.getOrDefault(Locale)`](https://github.com/civiform/civiform/blob/208acc61f31ad51beef3d073c7f39942c6366861/server/app/services/LocalizedStrings.java#L152) here so that we get the translated message?

server/app/views/admin/programs/ProgramPredicatesEditView.java
```diff
@@ -298,6 +312,45 @@ private static LabelTag renderPredicateQuestionCheckBoxRow(
...
+    BlockEligibilityMessageForm existingEligibilityMessageForm =
+        new BlockEligibilityMessageForm(existingEligibilityMessage);
+    Form<BlockEligibilityMessageForm> form =
+        formFactory.form(BlockEligibilityMessageForm.class).fill(existingEligibilityMessageForm);
+    DivTag buttonDiv =
+        div()
```
melanie-exygy, line 324: **nitpick:** Do we need to wrap the button in an empty `<div>`?

browser-test/src/support/admin_predicates.ts
```diff
@@ -43,6 +43,11 @@ export class AdminPredicates {
     )
   }
 
+  async updateEligibilityMessage(eligibilityMsg: string) {
+    await this.page.fill('input[name=eligibilityMessage]', eligibilityMsg)
```
melanie-exygy, line 47: **suggestion (non-blocking):** We've been trying to use the `.getByRole` locator so that we're locating elements in the same way a screen reader would. So, I think you could do something like:

`await this.page.getByRole('input', {name: `eligibilityMessage`})`
`await this. page.getByRole('button', {name: 'Save eligibility message'})`

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -399,6 +399,65 @@ test.describe('create and edit predicates', () => {
...
+        questions: [{name: 'show-other-q'}],
+      })
+    })
+
+    await test.step('Eligibility message field is visible', async () => {
+      await adminPrograms.goToEditBlockEligibilityPredicatePage(
+        programName,
+        firstScreen,
+      )
+      await validateScreenshot(page, 'edit-predicate-eligibility-msg-enabled')
```
melanie-exygy, line 444: **suggestion (blocking):** We're trying to avoid having very many screenshots.  I don't think this one is needed since we have the other screenshot below on line 457.

browser-test/src/admin/admin_predicates.test.ts
```diff
@@ -399,6 +399,65 @@ test.describe('create and edit predicates', () => {
...
+
+    await test.step('Eligibility message field gets updated', async () => {
+      await adminPredicates.updateEligibilityMessage(
+        'Customized eligibility message',
+      )
+      await validateToastMessage(
+        page,
+        'Eligibility message set to Customized eligibility message',
+      )
+      await validateScreenshot(page, 'edit-predicate-eligibility-msg-updated')
```
melanie-exygy, line 457: **suggestion (blocking):** Please narrow the screenshot down to as small of an area as possible, ideally just the eligibility message element.

server/app/views/admin/programs/ProgramPredicatesEditView.java
azhangggggg: We wanna keep it in the default language English here because this page is for Civiform Admins.

browser-test/src/admin/admin_predicates.test.ts
azhangggggg: Agree on narrowing the screenshot down to a small of an area, but here imho it's better to show the toast message, which shows up at the top of the page.

browser-test/src/admin/admin_predicates.test.ts
azhangggggg: Makes sense! Deleted this screenshot and only keeping the one on line 457.

server/app/views/admin/programs/ProgramPredicatesEditView.java
azhangggggg: Gotcha, making it a ButtonTag instead of a Divtag.

browser-test/src/admin/admin_predicates.test.ts
melanie-exygy: Ah, I see.  Yes, that makes sense.  Thanks for explaining.

browser-test/src/support/admin_predicates.ts
azhangggggg: makes sense! changes applied.

avaleske (approved):
lgtm! in the future, it might be nice to gray-out the save button and then only enable it when the message is actually updated

azhangggggg:
> lgtm! in the future, it might be nice to gray-out the save button and then only enable it when the message is actually updated

Sg! I will talk to sijia about the gray-out save button.

PR #9433, opened 2024-12-11
Adding application questions to swagger2 output

gwendolyngoetz:
### Description

Adding application questions to swagger2 output

#### General


server/app/services/openapi/AbstractOpenApiSchemaGenerator.java
```diff
@@ -28,4 +47,159 @@ protected ImmutableList<Scheme> getSchemes() {
...
+  /** Get the list of scalars for a question sorted alphabetically */
+  protected static ImmutableList<Scalar> getScalarsSortedByName(
+      QuestionDefinition questionDefinition)
+      throws InvalidQuestionTypeException, UnsupportedQuestionTypeException {
+    return Scalar.getScalars(questionDefinition.getQuestionType()).stream()
+        .filter(scalar -> !excludeScalarFromSchemaOutput(scalar))
+        .map(AbstractOpenApiSchemaGenerator::remapScalar)
+        .sorted(Comparator.comparing(Enum::name))
+        .collect(ImmutableList.toImmutableList());
+  }
```
avaleske, line 84: So this is going to result in any new scalars to a question automatically being added to the API, right?

I moved us away from that in the JSON/CSV (we now manually check each scalar we want to output) in part because of the mistake requiring the comment on line 99, but I'm ok with it here since we're also mapping them to types that aren't all strings.

what do you think?

server/app/services/openapi/AbstractOpenApiSchemaGenerator.java
```diff
@@ -28,4 +47,159 @@ protected ImmutableList<Scheme> getSchemes() {
...
+      case CURRENCY_CENTS -> ApiPathSegment.CURRENCY_DOLLARS.toString().toLowerCase(Locale.ROOT);
+      case FILE_KEY_LIST -> ApiPathSegment.FILE_URLS.toString().toLowerCase(Locale.ROOT);
+      case NAME_SUFFIX -> ApiPathSegment.SUFFIX.toString().toLowerCase(Locale.ROOT);
+      default -> scalar.name().toLowerCase(Locale.ROOT);
+    };
+  }
+
+  /** Map Scalar to DefinitionType */
+  protected static DefinitionType getDefinitionTypeFromSwaggerType(Scalar scalar) {
+    // Override type because we messed up
```
avaleske, line 99: :sigh:

server/app/services/openapi/AbstractOpenApiSchemaGenerator.java
```diff
@@ -28,4 +47,159 @@ protected ImmutableList<Scheme> getSchemes() {
...
+    }
+
+    switch (scalar.toScalarType()) {
+      case DATE:
+        return Optional.of(Format.DATE);
+      case LONG:
+        return Optional.of(Format.INT64);
+      case CURRENCY_CENTS:
+      case DOUBLE:
+        return Optional.of(Format.DOUBLE);
```
avaleske, line 135: do we need to be more careful with currency conversion then this? could we end up with a double like $90.010000000000001?

server/app/services/openapi/v2/Swagger2SchemaGenerator.java
```diff
@@ -159,9 +166,72 @@ public String createSchema(ProgramDefinition programDefinition) {
...
+          ApiPathSegment.QUESTION_TYPE.name().toLowerCase(Locale.ROOT), new StringProperty());
+
+      // Enumerators require special handling
+      if (questionDefinition.getQuestionType() != QuestionType.ENUMERATOR) {
+        for (Scalar scalar : getScalarsSortedByName(questionDefinition)) {
+          String fieldName = getFieldNameFromScalar(scalar);
+          DefinitionType definitionType = getDefinitionTypeFromSwaggerType(scalar);
+          DefinitionType arrayItemDefinitionType = getArrayItemDefinitionType(scalar);
+
+          // Boolean nullable = isNullable(definitionType);
```
avaleske, line 208: leftover comment

avaleske (approved):

server/app/services/openapi/AbstractOpenApiSchemaGenerator.java
gwendolyngoetz: Ok, so going to leave this for now and look into adding some tests to validate the output of an api call with the schema if the openapi library supports it.

avaleske (approved):

PR #9445, opened 2024-12-13
Add support for Azure multi-file uploads for applicants in non-Northstar deployments

lagosr-google:
### Description

This PR implements support for multi-file uploads for Azure in non-Northstar deployments.

- The file upload form class name varies depending on whether we are using single-file upload of multi-file upload.
- The typescript module varies the events it listens on depending on the class name of the upload form.

The above is neccessary because the Azure file upload strategy requires that a TS module upload the file directly to Azure
blob storage (rather than the browser).

After this PR the status of applicant file uploads on Azure is:

Non NS Single file: OK
NS Single File: Unimplemented
Non NS Multi File: In progress
NS Multi file: Unimplemented

This PR makes it so that files can be uploaded, and deleted by the end-user, but some features still need to be implemented (hence the in-progress status).

Features missing:
  - File names rather than UID's in views when multi-file upload is enabled.   (Issue: #9444)

#### General


server/app/views/fileupload/FileUploadViewStrategy.java
```diff
@@ -166,5 +166,11 @@ private static ValidationErrorMessage fileTooLargeMessage(int fileLimitMb) {
    */
   public abstract String getUploadFormClass();
 
+  /**
+   * Returns strategy-specific class to add to the <form> element when the form implements
+   * multi-file-upload. It helps to dinstinguish client-side different strategies (AWS or Azure)
```
dkatzz, line 171: **nitpick:** client-side different strategies -> different client-side strategies

dkatzz (approved):

server/app/views/fileupload/FileUploadViewStrategy.java
lagosr-google: done

dkatzz (approved):

PR #9472, opened 2024-12-18
Issue 9381 fix navigation on the error modal  for incomplete application step

### LINKED ISSUE #9381: Review & Exit shows the wrong modal
Non-NS: 
I partially fill in a page and click _review_:
image

I partially fill in a page and click _previous_:
image

NS: 
I partially fill in a page and click _review and exit_:
image

I partially fill in a page and click _back_:
image

In NS, clicking _review and exit_ should give me the same modal as clicking _review_ does in non-NS.

haltmar:
### Description

Conditionally set navigation and display text on the error modal dialog to either previous block or review page based on whether user click on "back" or "review and exit" button the the page.

#### General

#### Database evolutions

N/A

#### Durable jobs

N/A

#### User visible changes

#### New Features

N/A

create a program with multiple steps, one of which contains required questions. Alternatively,  you can import the attached test program json.

start application, go the step which contains required questions, 
click on Back,  you should see the dialog saying

Click on "Review and exit", you should see the dialog saying " ", click on "go to review", UI should take you to the review page (whereas without the fix, user can only go back to previous page)

### Issue(s) this completes

Fixes #9381

### DIFF
--- a/browser-test/src/support/applicant_questions.ts
+++ b/browser-test/src/support/applicant_questions.ts
@@ -971,7 +971,7 @@ export class ApplicantQuestions {
       ).toBeVisible()
       await expect(
         modal.getByText(
-          "There are some errors with the information you''ve filled in. Would you like to stay and fix your answers, or go to the review page without saving your answers?",
+          "There are some errors with the information you've filled in. Would you like to stay and fix your answers, or go to the review page without saving your answers?",
         ),
       ).toBeVisible()
       await expect(
--- a/browser-test/src/support/applicant_questions.ts
+++ b/browser-test/src/support/applicant_questions.ts
@@ -961,26 +961,32 @@ export class ApplicantQuestions {
     )
   }
   async expectErrorOnReviewModal(northStarEnabled = false) {
-    const modalTitle = 'Questions on this page are not complete. Would you still like to leave and begin reviewing?';
-  const modalContent = "There are some errors with the information you've filled in. Would you like to stay and fix your answers, or go to the review page without saving your answers?";
-  const buttonReviewText = 'Continue to review page without saving';
-  const buttonStayText = 'Stay and fix your answers';
+    const modalTitle =
+      'Questions on this page are not complete. Would you still like to leave and begin reviewing?'
+    const modalContent =
+      "There are some errors with the information you've filled in. Would you like to stay and fix your answers, or go to the review page without saving your answers?"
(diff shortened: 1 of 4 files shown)


haltmar:
[test-program-for-navigation-flows-exported.json](https://github.com/user-attachments/files/18203330/test-program-for-navigation-flows-exported.json)

test Program json file exported from existing test case for reviewer's convenience.

browser-test/src/support/applicant_questions.ts
```diff
@@ -960,16 +960,38 @@ export class ApplicantQuestions {
       'This question is required',
     )
   }
+  async expectErrorOnReviewModal(northStarEnabled = false) {
+    if (northStarEnabled) {
```
dkatzz, line 964: **suggestion:** Thank you for adding all of these assertions, but I think we can assert the same text as non-northstar (i.e. don't have to add text assertions for everything). You could also consider creating variables for the text, like `modalTitle`, `modalContent`, and `modalButton` and then using that in the branches.

server/app/views/applicant/NorthStarApplicantProgramBlockEditView.java
```diff
@@ -112,10 +113,36 @@ public String render(Request request, ApplicationBaseViewParams applicationParam
       context.setVariable(
```
dkatzz, line 113: **suggestion:** This is a bit hard to parse. I think one way to make this a little more clear would be to create two functions `setContextForReview` or `setContextForPrevious` and then you can branch which function gets called based on the applicationParams.errorDisplayMode(). And beyond that you could even make a helper function `setContextForFormModal`, which gets called from each of those sub-functions and would take in variables, like errorModalFormAction, errorModalTitle, errorModalContent, errorModalButtonText and errorModalDataRedirectTo and set the context for each. And then each function would set their particular form action then call the helper function (if you think its cleaner to create that)

browser-test/src/support/applicant_questions.ts
haltmar: Thanks for the suggestion.  refactored the method. PTAL.

server/app/views/applicant/NorthStarApplicantProgramBlockEditView.java
haltmar: thanks for the suggestion.  refactored the code. PTAL

Issue #9452, opened 2024-12-16
Prevent open redirect


dkatzz:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #9476, opened 2024-12-19
Validate that redirects point to relative paths

nb1701:
Previously, if somehow a bad actor was able to modified the hidden redirectLink input field on a page, or the value in the submitted form, they could redirect anywhere. This updates several places to validate that the redirect read from the form is a relative link.

#### General

Load the applicant home page, edit the HTML in-browser to change `redirectLink` to `https://google.com`, change the language, make sure it doesn't redirect to google.com.  It should show an error.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/9452

### DIFF
--- a/server/app/controllers/CallbackController.java
+++ b/server/app/controllers/CallbackController.java
@@ -6,6 +6,7 @@ import javax.inject.Inject;
 import play.mvc.Controller;
 import play.mvc.Http;
 import play.mvc.Result;
+import services.UrlUtils;
 
 /**
  * This is a convenience class to wrap org.pac4j.play.CallbackController. Controllers outside our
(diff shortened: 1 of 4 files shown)


dkatzz:
> I hesitated a bit on whether we should handle this more gracefully or not. You should only ever hit this if you're doing something bad, and the exception is getting caught up in some other place I need to track down, so it wasn't super straightforward to catch it. But if we think people are going to balk at the unhandled exception, I can try to make it nicer.

I think this seems fine at least for now - if we want to, we can create another ticket to handle these use cases, but hopefully nobody ever sees this exception anyway.

dkatzz (approved):

PR #9481, opened 2024-12-30
Add support for multi file original names in storage and display.

### LINKED ISSUE #9444: File names are missing for applicant file uploads in Azure with multi-file-upload enabled.
**Describe the bug**

When an applicant file is uploaded to azure with multi-file-upload enabled (and Northstar disabled), the UID of the file in storage is displayed to the user rather the name of the file they uploaded.

This works ok when multi-file-upload is not enabled.

**To Reproduce**

1 - run azure env.
2 - enable multi-file upload
3 - upload a file to an application
4 - the name of the file uploaded will be a long sequence of numbers (a UID) rather than the filename.

**Expected behavior**
The name of the file uploaded should be displayed and the UID hidden.

lagosr-google:
### Description

Store and display original file names for multi-file uploads in Azure deployments using NorthStar.

#### General

#### User visible changes

- [X] Followed steps to [internationalize](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#application-strings) any new strings
  - [X] Added context strings to new [messages](https://github.com/civiform/civiform/blob/main/server/conf/i18n/messages)
  - [X] Didn't use a message in applicant facing code that isn't translated yet (unless behind a flag)
- [X] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method
- [X] Made equivalent changes in Thymeleaf for applicant-facing features and changes (or created an issue in the [Northstar epic](https://github.com/orgs/civiform/projects/1/views/94) to track that it's missing from Thymeleaf)
- [X] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [X] Manually tested at 200% size
- [X] Manually evaluated tab order
- [X] Manually tested with a screen reader if the feature is applicant-facing. See [screen reader testing](https://github.com/civiform/civiform/wiki/Testing#screen-reader-testing)

### Issue(s) this completes

Fixes #9444


server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -748,7 +758,44 @@ public CompletionStage<Result> addFileWithApplicantId(
...
+                  for (int i = 0; i < orignalFileNames.size(); i++) {
+                    String originalFileNameValue = orignalFileNames.get(i);
+                    fileUploadQuestionFormData.put(
+                        fileUploadQuestion.getOriginalFileNameListPathForIndex(i).toString(),
+                        originalFileNameValue);
+                    // OriginalFileName already exists in question, no need to append it. But we may
+                    // want
+                    // to render some kind of error in this case in the future, since it means the
+                    // user
+                    // essentially "replaced" whatever file already existed with that same name.
```
dkatzz, line 777: **thought:** In the file upload ts file, if someone uploads a file with the same name, we increment it. By the time it gets here, could the file have the same name as an existing one? https://github.com/civiform/civiform/blob/d1db8db27ffed3dcb17738a9d75914302e5daf9e/server/app/assets/javascripts/file_upload.ts#L109.

server/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -16,6 +16,15 @@
  */
 public final class FileUploadQuestion extends Question {
 
+  // There is an unfortunate dual logic in the how filenames are stored for question answers. AWS
+  // deployements store the filename in the key, Azure stores them in a separate column.
```
dkatzz, line 20: **nitpick:** deployements -> deployments

server/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -24,6 +33,7 @@ public final class FileUploadQuestion extends Question {
 
   FileUploadQuestion(ApplicantQuestion applicantQuestion) {
     super(applicantQuestion);
+    this.useOriginalFileNames = true;
```
dkatzz, line 36: **question:** Are we doing this for everything or specifically for azure? The comment above makes it seem like its just for azure, but it seems like we're always doing it in this PR? I could be misreading though. My understanding is the way the current originalFilename works is its set in the azure_upload file https://github.com/civiform/civiform/blob/d1db8db27ffed3dcb17738a9d75914302e5daf9e/server/app/assets/javascripts/azure_upload.ts#L128

dkatzz: Ah I think I may understand - we only set originalFileNameList in ApplicantProgramBlocksController if originalFileName is set, which only happens if its passed down from the azure_upload.ts file - is that right?

dkatzz (commented):
Having trouble seeing how we're branching this for AWS vs Azure as described in the comments. I may be missing something - would you mind adding some more information in the PR description? I also will plan to look at this again in the morning with fresh eyes :)

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -746,7 +756,44 @@ public CompletionStage<Result> addFileWithApplicantId(
...
+                  boolean appendValue = true;
+
+                  // Write the existing filenames so that we don't delete any.
+                  for (int i = 0; i < orignalFileNames.size(); i++) {
+                    String originalFileNameValue = orignalFileNames.get(i);
+                    fileUploadQuestionFormData.put(
+                        fileUploadQuestion.getOriginalFileNameListPathForIndex(i).toString(),
+                        originalFileNameValue);
+                    // OriginalFileName already exists in question, no need to append it. But we may
+                    // want
```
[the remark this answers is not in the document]
lagosr-google: done

server/app/services/applicant/question/FileUploadQuestion.java
lagosr-google: done

server/app/controllers/applicant/ApplicantProgramBlocksController.java
lagosr-google: yes - and thank you for this question.

Turned out that we could have the same filename, and since we do not use filenames as keys to the stored file, we should not care to store the same file name string in the record.

Removed this code that avoided uploading a record when the filename already existed, and that fixed multi-file upload for the same file for azure.

The current implementation allows the file name to be the same, since the key to the file is a UID and is unique(*) .

[FileUploadQuestion.java](https://github.com/civiform/civiform/pull/9481/files/641ee8afeb2fa7cca708f61ba354fcffc693f661#diff-59ef23a255be1c0692f36e84de257f98cb3a5c3787f273fa975f4f6f4893e69f)

The current implementation for Azure allows the same filename to exist for multiple uploads, which IMO is a better user experience.

server/app/services/applicant/question/FileUploadQuestion.java
lagosr-google: Yes thats correct.

If i understand this code correctly, there is some caching going on to avoid extra calls down to the storage layer. (the Cache fields in the class)

In the case of original filenames the class will attempt to read them by default, and then stop doing so after the first time it fails to find the data in the storage layer.

This limits the extra cost to non-azure deployments to one extra call down to storage.

Since the current implementation was not trying to cache the multi-key values , i did not attempt to cache the multi-original file name values, and just wanted to limit the extra costs to AWS deployments.   depending on how this class is instantiated (i dont think its a singleton), the actual savings is limited.

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -748,7 +779,37 @@ public CompletionStage<Result> addFileWithApplicantId(
                     fileUploadQuestion.getFileKeyListPathForIndex(0).toString(), key.get());
               }
 
-              return ensureFileRecord(key.get(), Optional.empty())
+              if (originalFileName.isPresent()) {
```
dkatzz, line 782: **suggestion:** Could you add a comment here with a reminder that this value is only set for azure?

server/app/controllers/applicant/ApplicantProgramBlocksController.java
```diff
@@ -748,7 +779,37 @@ public CompletionStage<Result> addFileWithApplicantId(
...
+                    // The actual file key is a UID in this case, and we've already checked for
+                    // key collisions above.
+                  }
+
+                  fileUploadQuestionFormData.put(
+                      fileUploadQuestion
+                          .getOriginalFileNameListPathForIndex(orignalFileNames.size())
+                          .toString(),
+                      originalFileName.get());
+                } else {
```
dkatzz, line 805: **question:** In what cases would the originalFileName.isPresent() return true but originalFileNamesOptional.isPresent() return false?

server/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -16,6 +16,15 @@
...
 public final class FileUploadQuestion extends Question {
 
+  // There is an unfortunate dual logic in the how filenames are stored for question answers. AWS
+  // deployments store the filename in the key, Azure stores them in a separate column.
+  //
+  // In order for the application to work properly we must look for data in the original file names
+  // Scalar, and absent that data, return the file keys as the filename.
+  //
+  // Remember whether to use original filenames, to avoid multiple reads for AWS deployments.
+  private boolean useOriginalFileNames;
```
dkatzz, line 26: **suggestion:** `useOriginalFileNames` feels like a confusing name to me - originally I thought this was the AWS way because we're using the actual file names and not the map of file keys to names, but then I realized this is the azure one. Maybe something like `useFileNameMap` or something would make it more clear? I also think it would be helpful to call out explicitly above, this will end up being true for azure deployments and false for AWS deployments.

server/app/services/applicant/question/FileUploadQuestion.java
dkatzz: Got it - it may be helpful adding a comment or something here since it took me a long time to understand.

server/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -88,6 +106,42 @@ public Optional<String> getOriginalFileName() {
...
+   */
+  public Optional<ImmutableList<String>> getOriginalFileNameListValue() {
+    if (useOriginalFileNames) {
+      Optional<ImmutableList<String>> originalFileNames =
+          applicantQuestion.getApplicantData().readStringList(getOriginalFileNameListPath());
+      if (originalFileNames.isPresent()) {
+        return originalFileNames;
+      }
+    }
+
```
dkatzz, line 120: **suggestion:** Similar to below, maybe include the comment `// Execution will reach this point, if we have no data for original file names.`

server/test/services/applicant/question/FileUploadQuestionTest.java
```diff
@@ -67,6 +68,65 @@ public void withApplicantData_passesValidation() {
...
+        ImmutableList.of("filekey1", "filekey2"));
+    assertThat(fileUploadQuestion.getFileKeyListValue().get())
+        .containsExactly("filekey1", "filekey2");
+    assertThat(fileUploadQuestion.getFileKeyValueForIndex(0).get()).isEqualTo("filekey1");
+    assertThat(fileUploadQuestion.getFileKeyValueForIndex(1).get()).isEqualTo("filekey2");
+    assertThat(fileUploadQuestion.getValidationErrors()).isEmpty();
+  }
+
+  @Test
+  public void getOriginalFileName_MultiFile_notAnswered_returnsNotPresent() {
```
dkatzz, line 90: **nitpick:** MultiFile -> multiFile

server/test/services/applicant/question/FileUploadQuestionTest.java
```diff
@@ -67,6 +68,65 @@ public void withApplicantData_passesValidation() {
...
+
+    FileUploadQuestion fileUploadQuestion = new FileUploadQuestion(applicantQuestion);
+    QuestionAnswerer.answerFileQuestionWithMultipleUploadOriginalNames(
+        applicantData, applicantQuestion.getContextualizedPath(), ImmutableList.of(""));
+    assertFalse(fileUploadQuestion.getOriginalFileNameListValue().isPresent());
+    assertFalse(fileUploadQuestion.getOriginalFileNameValueForIndex(0).isPresent());
+  }
+
+  @Test
+  public void getOriginalFileName_MultiFile() {
```
dkatzz, line 114: **nitpick:** MultiFile -> multiFile

server/app/services/question/QuestionAnswerer.java
```diff
@@ -88,6 +88,22 @@ public static void answerFileQuestionWithMultipleUpload(
     }
   }
 
+  public static void answerFileQuestionWithMultipleUploadOriginalNames(
```
dkatzz, line 91: **suggestion:** Maybe rename to `answerMultiFileQuestionWith..` and then maybe `NamesFromMap`? As mentioned above, I feel like OriginalFileNames makes me think its using the file names that are stored with it vs us having to provide the map. But curious what you think. I'm open to leaving it as well if others don't find it confusing.

dkatzz (commented):
In general this looks good - thanks for fixing the piece about removing files. I still find this difficult to parse though in general. I wonder if removing the abstraction of `originalFileNames` in the variables and being more explicit and just calling it `azureFileName` / `azureFileNames` or something in certain places would be clearer? Curious your thoughts.

server/app/services/applicant/question/FileUploadQuestion.java
lagosr-google: done

server/app/controllers/applicant/ApplicantProgramBlocksController.java
lagosr-google: done

server/app/controllers/applicant/ApplicantProgramBlocksController.java
lagosr-google: This happens when the upload form contains a originalFileName field indicating that the form is trying to store the original file name.

And the applicant answer data has no entries for uploaded files .. (and no entriers for original file names) -- in that case, there is no need to copy the existing records to the answer before inserting the new entry.

i think it may be possible to collapse the fileUploadQuestionFormData.put() call into a single call ..  but that might actually make it harder to read.

as is, if there are existing file upload records in the form, then copy the existing onees and add the original file name to the end, -- otherwise just add it to the front.

added a comment to clarify this.

server/app/services/applicant/question/FileUploadQuestion.java
lagosr-google: agreed -- changing this variable to useFileNameMap might make this clearer.

server/test/services/applicant/question/FileUploadQuestionTest.java
lagosr-google: done

server/test/services/applicant/question/FileUploadQuestionTest.java
lagosr-google: done

server/app/services/question/QuestionAnswerer.java
lagosr-google: added doc strings to differentiate what all theese helper methods are trying to do.

lagosr-google:
> In general this looks good - thanks for fixing the piece about removing files. I still find this difficult to parse though in general. I wonder if removing the abstraction of `originalFileNames` in the variables and being more explicit and just calling it `azureFileName` / `azureFileNames` or something in certain places would be clearer? Curious your thoughts.

given that the name originalFileName exists down to the schema, changing that verbiage could be a problem.. though i've added more comments to try to clarify this functionality.

dkatzz (approved):
The comments are really helpful, thanks for adding those!

gwendolyngoetz (approved):

dkatzz (approved):

Issue #7686, opened 2024-06-05
[North Star Applicant UI] Add styles to program index page


greg-hyde:
This covers adding styles to program cards/eligibility tags/ sections of cards etc.

melanie-exygy:
@sijiayam When we last talked, we decided that the “You are applying as a guest…” and the TI "You are applying on behalf of..." should be aria-live regions with `role="status"`.  I've since done some research into when to use aria-live and when to announce status messages.  Announcements are only recommended when there is a change in the page content.  For example, when the user hits a search button, the page content changes dynamically, so it would be good to announce "10 search results".  Here's the WCAG principle I'm referring to:  [Status Messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages).

In our case, those two banners are already part of the content on page load.  The screen reader will read that content.  We don't need to interrupt the normal flow of the screen reader.

If we want to highlight the importance of the message, we could add some screen reader only text at the beginning of the banner, such as "Important" or "For your information".  This could be a screen reader substitute for the info icon at the beginning of the banner.  

Let me know what you think.

PR #9488, opened 2025-01-02
Stretching background colors to full width on wide screens

melanie-exygy:
### Description

As requested by UX, extending the width of the background color to full screen for the Applicant Home Page blue banner and for the "You are a guest" and "Applying on behalf of..." banners.

This only makes a difference on wide desktop screens.

This is only for the North Star.  

image

1. Go to the applicant home page.
2. Start an application and go back to the home page to see the "You are a guest" banner.
3. Log in as a TI.
4. In the client list, pick a client and click the "View applications" button.  This will take you back to the home page where you'll see the "Applying on behalf of..." banner.

### Issue(s) this is part of:

Part of #7686


melanie-exygy:
> Approving pending a double check on the alert spanning the width

From @sijiayam:  "Yes the little blue bar is a little strange but i think it is okay for that to stretch the whole width. I've seen it similarly on other gov't sites too. Not a big concern on my end for now."

Issue #9088, opened 2024-10-30
[Program Overview Page] Build out new applicant facing program overview page


rockycodes:
Implement new applicant facing page that provides an overview of the program.

Acceptance criteria:
- [x] The url for the program overview page is `programs/:programSlug`
- [x] Applicants see a page that renders information about the program and visually matches the [mocks](https://www.figma.com/design/RmhVJCLgAKWGENJLWWAtNS/CiviForm%3A-Components%2C-Styles?node-id=2741-5764&m=dev) as closely as possible (the correct mock is the the one that says "Child care assistance program" at the top)
- [x] This page includes all fields input by the admin, which can include:
- Title, long description 
- Button that takes the resident to the start of the application 
- Eligibility banner (already exists, but design is updated) 
- Steps on how to apply (whichever steps admin decides to input for that program)
- Option to create an account or login
- [x] Long description is rendered with markdown (defer to #9531 )
- [x] If the long description does not exist for the program, the short description is rendered
- [x] The page uses the [USWDS process list component](https://designsystem.digital.gov/components/process-list/) to render application steps, and other USWDS components as indicated by the design.  The steps descriptions are rendered with markdown.
- [x] When a user clicks the "Start your application" button at the bottom of the screen, they are taken to the first page of the application
- [x] When a user clicks the "Log in" button they are taken to the log in screen
- [x] When a user clicks "view and apply", they are NOT shown the "Create an account or sign in" modal. If they go to the program overview page and then click "start application", they also are NOT show that modal.
- [x] #9455

See [the TDD](https://docs.google.com/document/d/1rP4kqltv122Wfp9-maJlvm5rzFkFcWrCt5w0lKRoZEs/edit?tab=t.0) for more information and implementation details

Note: The "More information" section is not being implemented at this time.
Also, the program external link will not be included.  Rather, we will let admins add links in the long description.

rockycodes:
Note: Work on this can begin in parallel with the admin UI using dummy data. Once #9085 and #9086 are completed, we can swap in the values pulled from the db.

shreyachatterjee00:
Note: made a decision to show our "Create an account or sign in" modal when the user clicks "start an application" from the program overview page, rather than when the user clicks "View and apply" (which is the current behavior before program overview). 
![image](https://github.com/user-attachments/assets/5392a1e1-745e-4d05-abe5-4b69e5721d66)

sijiayam:
[...]

PR #9518, opened 2025-01-09
Fixing details of the show program route for the North Star

### LINKED ISSUE #9455: Add test for the case when a user navigates to `/programs/programName` with a disabled program.
Add test for the case when a user navigates to /programs/programName with a disabled program.  See [this PR discussion](https://github.com/civiform/civiform/pull/9319#discussion_r1884720446).

melanie-exygy:
### Description

For the North Star, the program `show` route (`/programs/programName`), should lead to the new program overview page.  This is regardless of whether the user has started or completed an application for the program.  Essentially, we are replacing the old program details view.  This was already set up in PR #9319.  However, the current PR fixes 3 things:

- Clears the REDIRECT_TO key from the session after returning the program overview view.  This is so that the key won't be used again for the next redirect, for example, with the dev tools.
- Creates a new `showWithApplicantId` route that maintains the applicant id, this is for TIs applying on behalf of applicants.
- Navigates to the program overview even when there is an application in progress.  This distinguishes the `show` route from the `edit` or the `review` routes.

This PR also moves any tests from `northstar_program_deep_link.test.ts` that are still relevant into `northstar_program_overview.test.ts`.  Some of the tests were removed because they tested the "Create an account or sign in" modal, which used to pop up with the deep link before a guest user arrived on the program review page.  This modal won't show up before the program overview page.  We're only going to display it once the user clicks "Start an application" from the overview page.

1. Turn on the North Star and disabled visibility condition flags.
2. Create a program if you don't have one.
3. Go to this URL `/programs/<your-program-name>` (the name needs to be lowercase and have dashes instead of spaces)
4. Note that this gets you to the program overview page (which is very bare bones at the moment).
5. Go back to the applicant home page.
6. Start applying to that program.
7. Go to `/programs/<your-program-name>` again.
8. Note that you get to the program overview page.
9. Submit your application.
10. Go to that same URL.
11. Note that you still get to the program overview page.
12. Log in as a trusted intermediary.
13. Create a client if you don't have one.
14. Click "View applications" on the client card.
15. Note that the URL now has the applicant id ("/applicants/35/programs").
16. Add the program slug to the end of that URL (`/applicants/35/programs/<your-program-name>`)
17.  Note that you get to the program overview page.

### Issue this is part of

Part of #9088

### Issue(s) this completes

Fixes #9455 ;


browser-test/src/support/civiform_fixtures.ts
```diff
@@ -30,6 +31,7 @@ type CiviformFixtures = {
   adminSettings: AdminSettings
   applicantFileQuestion: ApplicantFileQuestion
   applicantProgramList: ApplicantProgramList
+  programOverview: ApplicantProgramOverview
```
gwendolyngoetz, line 34:
```suggestion
  applicantProgramOverview: ApplicantProgramOverview
```

(blocking) Please rename so this matches up with the same naming pattern.

gwendolyngoetz (changes requested):

browser-test/src/support/civiform_fixtures.ts
melanie-exygy: Good idea.  Done.

gwendolyngoetz (approved):

Issue #9558, opened 2025-01-16
Create a builder for AlertSettings


melanie-exygy:
#### Describe the solution presented in this issue.
The number of properties on the `AlertSettings.java` record is getting a little long.  It would be handy to create an AlertSettingsBuilder or something to set various properties.

rabelmervin:
Hi @melanie-exygy  Can i contribute to this issue ? This will be a good learning opportunity for me : )

gabrielc5-un:
@melanie-exygy Hello I would like to work on this, I tried to join the community contributors Slack channel as outlined in your guys contribution guide. But I am unable to as my email address is not associated with the Slack Channel, my email is gabrielcxrdero@gmail.com. Thank you :)

melanie-exygy:
Hi @rabelmervin, welcome to CiviForm!  Please introduce yourself on this [Slack channel](https://civiform.slack.com/archives/C07S691HDU0) and we can get this issue assigned to you.

melanie-exygy:
@gabrielc5-un, I've added you to the Community Contributors channel.  Please introduce yourself when you have a chance.  Thanks!

rabelmervin:
Hi @melanie-exygy 👋,  I have joined community and I'm super excited to contribute civiform.

Lynnshelsie1:
Hello @melanie-exygy can I contribute as well to this issue! I'm part of Florida International University Tech Talent Academy and we are learning and exposing ourselves to GitHub projects it would be a great opportunity to work on this.

melanie-exygy:
@rabelmervin I've assigned this issue to you since you asked first.  You can now fork the `civiform/civiform` repo and start working on it.  Let me know if you run into any problems.

melanie-exygy:
> Hello [@melanie-exygy](https://github.com/melanie-exygy) can I contribute as well to this issue! I'm part of Florida International University Tech Talent Academy and we are learning and exposing ourselves to GitHub projects it would be a great opportunity to work on this.

Hi @Lynnshelsie1, thanks for reaching out.  We'd love to have you help with CiviForm.  Please join our [CiviForm Slack community](https://civiform.slack.com/archives/C07S691HDU0) if you're able.  I assigned this particular issue to @rabelmervin since they asked first, but would you be interested in https://github.com/civiform/civiform/issues/7959 instead?

PR #9577, opened 2025-01-22
Created a Builder for AlertSettings(#9558)

rabelmervin:
### **Description** 📝

This PR introduces the AlertSettingsBuilder to improve the creation of AlertSettings objects. As the AlertSettings class had a growing number of properties, its instantiation becomes difficult. The builder pattern was implemented to provide a more structured and readable approach for setting various properties of AlertSettings

### **Solution**  💡

To solve this issue, an AlertSettingsBuilder class was implemented within the AlertSettings class. This builder pattern provides a fluent and readable way to configure and create instances of AlertSettings.

### **Change Made** 🔧 

- Added a static builder() method to instantiate the AlertSettingsBuilder.
- Retained existing properties and methods.
- Implemented as a static inner class within AlertSettings.
- Provides setter methods for each property (setShow, setTitle, setText, etc.).
- Includes a build() method to create and return a fully configured AlertSettings instance.

### **Release Notes** 📢

This change introduces a builder pattern to simplify the instantiation of AlertSettings by providing a fluent interface for setting properties.


server/app/services/AlertSettings.java
```diff
@@ -22,20 +22,59 @@ public record AlertSettings(
...
+  public static class AlertSettingsBuilder {
+    private Boolean show = false;
+    private Optional<String> title = Optional.empty();
+    private String text = "";
+    private Boolean unescapedDescription = false;
+    private AlertType alertType = AlertType.NONE;
+    private ImmutableList<String> additionalText = ImmutableList.of();
+    private Boolean isSlim = false;
+
+    public AlertSettingsBuilder setShow(Boolean show) {
```
melanie-exygy, line 41: **suggestion:** Generally, for builders in CiviForm, the method name is just the name of the property, without the word "set".  So, `public AlertSettingsBuilder show(Boolean show)`.

server/app/services/AlertSettings.java
```diff
@@ -22,20 +22,59 @@ public record AlertSettings(
     Boolean isSlim) {
 
   public static AlertSettings empty() {
-    return new AlertSettings(false, Optional.empty(), "", AlertType.NONE);
+    return new AlertSettings(false, Optional.empty(), "", AlertType.NONE, ImmutableList.of(), false);
```
melanie-exygy, line 25: **issue (blocking):** Since you're removing the constructor with 6 parameters below, I don't think this will compile.  I think you want to use either the builder or the default constructor.

server/app/services/AlertSettings.java
```diff
@@ -22,20 +22,59 @@ public record AlertSettings(
...
   }
 
-  public AlertSettings(Boolean show, Optional<String> title, String text, AlertType alertType) {
-    this(show, title, text, alertType, ImmutableList.of(), /* isSlim= */ false);
+  public static AlertSettingsBuilder builder() {
+    return new AlertSettingsBuilder();
   }
 
-  public AlertSettings(
```
melanie-exygy, line 32: **praise:** I like that you're removing this constructor so that we use the builder instead.

melanie-exygy (commented):
Nice job creating a builder!  This will make it so much clearer which properties we're setting.  The next step in this PR is to replace anywhere that the old constructors were used.  You'll want to replace them with the builder.  

Were you able to set up your environment?  That will allow you to check for compilation errors and it'll allow you to run the unit tests and the browser tests.  You'll also be able to run our linter by using the command `bin/fmt`.  Here are instructions on getting set up locally:  [Getting Started](https://github.com/civiform/civiform/wiki/Getting-started) and on running tests locally:  [Testing](https://github.com/civiform/civiform/wiki/Testing).  Let me know in Slack if you need any help.

rabelmervin:
Thanks @melanie-exygy for reviewing I'll make the changes !

gwendolyngoetz (changes requested):

Issue #9509, opened 2025-01-07
Wrap data-modifying database calls in transactions to avoid data integrity errors and race conditions


avaleske:
## The issue
In many places, we modify data in the database using a fetch, modify, and update pattern:
1. We fetch an entity from the database and load it into memory. Sometimes this is a json blob, like an ApplicantData or ProgramDefinition, and sometimes it's something like a Question that we load and hold in memory as the QuestionDefinition DTO.
2. We then modify that object in memory in application code, usually in a service class.
3. We then pass that updated object back to a repository method to save to the database.

If this _entire_ process is not wrapped in a single [serializable transaction](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE), which protects agains two transactions modifying the same data that another transaction might also be modifying, two queries can overwrite each other's data.

If two queries are both modifying the same data in memory the second query will overwrite the first if the whole fetch-modify-update cycle is not in a transaction. If two users are modifying the same Question looks like:
1. User A requests a draft QuestionDefinition
2. User B requests the same draft QuestionDefinition
3. A modifies it in memory.
4. A saves it to the database as an update to the original row.
5. B modifies it in memory.
6. B saves it to the database as an update to the original row, overwriting A's changes.

If the entire fetch-modify-update cycle was in a transaction, then #6 above would be rejected by the database because A has changed the value of the row since B read it. Additionally, this can happen even if there's only one server instance because each user could be on their own thread.

### Example
An example of this is in `QuestionService#update()`, here:
https://github.com/civiform/civiform/blob/a4d0c011d249bd11497f19fe117c08f3320063d5/server/app/services/question/QuestionService.java#L152-L173
In this code block we load a question from the database, validate the changes we made, and then store it, but this is not done in a single transaction. If another user is doing the same thing at the same time, a race condition could occur. This code block also demonstrates a far more likely race condition, which is described [below](#related_patterns_not_fixed_with_transactions).

### Related patterns
There are several other data update patterns that can lead to issues if not wrapped in a transaction, such as:
1. Updating multiple tables, such as when we update an application's status with both a row in the `application_events` table and by changing the `status` field of an application in the `applications` table. If this is not done in a single transaction (which it is) a concurrent read could see the two tables in inconsistent states.
[...]

PR #9608, opened 2025-01-24
Create TransactionManager helper class and wrap question creation

avaleske:
### Description

Create a `TransactionManager` helper class and use it to wrap question creation flow in a transaction.

Note: I haven't tested that this works correctly with async flows. It probably doesn't because transactions are tracked on thread-local storage, and some repository methods use the DatabaseExecutionContext while other application code uses the ClassLoaderExecutionContext, and I don't know if the context is propagated between thread pools.

#### General

### Issue(s) this completes

Part of #9509

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -166,7 +166,7 @@ jobs:
           retention-days: 3
           path: browser-test/diff_output
 
-      # We only want updated snapshots for aws. Otherwise we'll hit conflicts if both aws and azure 
+      # We only want updated snapshots for aws. Otherwise we'll hit conflicts if both aws and azure
       # attempt to update the same images
       - name: Upload updated snapshots
         if: ${{ matrix.cloud == 'aws' && failure() }}
@@ -176,7 +176,7 @@ jobs:
           retention-days: 3
           path: browser-test/updated_snapshots
 
-      # Upload Playwright trace files and recorded videos. We want this for aws and azure so we 
+      # Upload Playwright trace files and recorded videos. We want this for aws and azure so we
       # have access to what happened in failed tests.
       - name: Upload test videos on failure
         if: failure()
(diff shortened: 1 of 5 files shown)


avaleske:
* **#9608** <a href="https://app.graphite.dev/github/pr/civiform/civiform/9608?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/civiform/civiform/9608?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

avaleske:
> This is awesome, thank you for this! A couple non-blocking questions:
> 
> * Are there any other exceptions we should handle besides `SerializableConflictException`?
> * I wonder if we can cause an actual concurrency issue without simulating it by throwing the exception explicitly.  Maybe we can have two async transactions and the first one sleeps a bit or something?
> * I'm not sure if I'm thinking about this correctly, but if two questions attempt to get added at the same time with the same admin name, I'm not sure that's actually going to throw an exception.  The `name` field in the questions table is not unique. Maybe ID would conflict, but I forget if that's just auto-incremented on the table itself.  I think `insertQuestionSync` just throws the question into the table without any checking.

- VersionRepository's transaction related code also catches `RollbackException` and `NonUniqueResultException`, but imo those should be handled by the callsite rather than this wrapper class (although I don't feel strongly either way) since a rollback could be triggered by things other than a concurrent transaction issue, and how to handle a non-unique result is specific to the caller.

- I thought about this but that seems like something ebeans should be testing, not us. though again I don't feel strongly.

- yeah `insertQuestionSync()` doesn't check, but `checkConflicts()` checks that both the names and the derived admin names are unique. that's why this whole thing should be in a transaction, since you could verify uniqueness but have another thread insert a colliding name between when you verified uniqueness and when you inserted. (Postgres notices that another thread wrote to the table after you read from it. if you never write back to the DB that's fine and it lets you continue, but if you write back to the DB it'll assume you're working with out of date data and fail the transaction when you attempt to `.commit()` it.

PR #9613, opened 2025-01-25
Created a Builder for AlertSettings and updated other files 

rabelmervin:
### Description

This PR introduces the AlertSettingsBuilder to simplify the creation of AlertSettings objects. As the AlertSettings class grew more complex, it became harder to instantiate. The builder pattern was added to make it easier and more readable to set the properties of AlertSettings. Other files using AlertSettings have been updated accordingly.

### DIFF
--- a/server/app/controllers/HomeController.java
+++ b/server/app/controllers/HomeController.java
@@ -18,6 +18,7 @@ import play.libs.concurrent.ClassLoaderExecutionContext;
 import play.mvc.Controller;
 import play.mvc.Http;
 import play.mvc.Result;
+import repository.HealthCheckRepository;
 import services.applicant.ApplicantData;
 
 /** Controller for handling methods for the landing pages. */
@@ -29,6 +30,7 @@ public class HomeController extends Controller {
   private final Optional<String> faviconURL;
   private final LanguageUtils languageUtils;
   private final ApplicantRoutes applicantRoutes;
+  private final HealthCheckRepository healthCheckRepository;
 
   @Inject
   public HomeController(
@@ -37,7 +39,8 @@ public class HomeController extends Controller {
       MessagesApi messagesApi,
       ClassLoaderExecutionContext classLoaderExecutionContext,
       LanguageUtils languageUtils,
-      ApplicantRoutes applicantRoutes) {
+      ApplicantRoutes applicantRoutes,
+      HealthCheckRepository healthCheckRepository) {
     checkNotNull(configuration);
     this.profileUtils = checkNotNull(profileUtils);
     this.messagesApi = checkNotNull(messagesApi);
@@ -46,6 +49,7 @@ public class HomeController extends Controller {
     this.applicantRoutes = checkNotNull(applicantRoutes);
     this.faviconURL =
         Optional.ofNullable(Strings.emptyToNull(configuration.getString("favicon_url")));
+    this.healthCheckRepository = checkNotNull(healthCheckRepository);
   }
 
   public CompletionStage<Result> index(Http.Request request) {
@@ -96,7 +100,10 @@ public class HomeController extends Controller {
   }
 
   public Result playIndex() {
-    return ok("public index");
+    if (healthCheckRepository.isDBReachable()) {
+      return ok("public index");
+    }
+    return badRequest();
   }
 
   // Redirect any browsers who, by default, request favicon from root, to the
--- a/server/app/repository/HealthCheckRepository.java
+++ b/server/app/repository/HealthCheckRepository.java
@@ -1,0 +1,31 @@
+package repository;
+
+import io.ebean.DB;
+import io.ebean.Database;
+import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Implements a simple query needed to check if DB is reachable */
+public final class HealthCheckRepository {
+  private final Database database;
+  private final Logger logger = LoggerFactory.getLogger(HealthCheckRepository.class);
+
+  @Inject
+  public HealthCheckRepository() {
+    this.database = DB.getDefault();
+  }
+
+  /**
+   * Does a simple query to check if the DB is reachable. In case of any exception, error is logged
+   * and false is returned.
+   */
+  public boolean isDBReachable() {
+    try {
+      return database.sqlQuery("SELECT 1").findOne() != null;
+    } catch (RuntimeException e) {
+      logger.error(e.getMessage());
+    }
+    return false;
+  }
+}
--- a/server/test/controllers/HomeControllerTest.java
+++ b/server/test/controllers/HomeControllerTest.java
@@ -1,14 +1,22 @@
 package controllers;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 import static play.api.test.Helpers.testServerPort;
 import static play.test.Helpers.route;
 import static support.FakeRequestBuilder.fakeRequestBuilder;
 
+import auth.ProfileUtils;
+import com.typesafe.config.Config;
+import controllers.applicant.ApplicantRoutes;
 import org.junit.Test;
 import org.pac4j.core.context.HttpConstants;
+import play.i18n.MessagesApi;
+import play.libs.concurrent.ClassLoaderExecutionContext;
 import play.mvc.Http;
 import play.mvc.Result;
+import repository.HealthCheckRepository;
 import repository.ResetPostgres;
 import support.CfTestHelpers;
 import support.CfTestHelpers.ResultWithFinalRequestUri;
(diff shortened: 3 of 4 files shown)


rabelmervin:
Hi @melanie-exygy @rockycodes  I have closed previous PR as i was facing issues and please review it.... can you explain what are these checks and how can i get passed ?
                                                   
Screenshot (27)

server/app/views/applicant/NorthStarApplicantUpsellView.java
```diff
@@ -20,92 +20,97 @@
...
+        context.setVariable("bannerMessage", params.bannerMessage());
+
+        ImmutableList<String> additionalText = ImmutableList.of("Text1", "Text2");
+
+        String alertTitle = params
+                .messages()
+                .at(MessageKey.ALERT_SUBMITTED.getKeyName(), params.programTitle().orElse(""));
+        AlertSettings successAlertSettings = AlertSettings.builder()
+                .show(true)
+                .title(Optional.of("Alert Title"))
```
melanie-exygy, line 69: **issue (blocking):** The alert title can't be hard-coded because it needs to be translated.  Please keep the existing code to fetch it:  
```
    String alertTitle =
        params
            .messages()
            .at(MessageKey.ALERT_SUBMITTED.getKeyName(), params.programTitle().orElse(""));

File: server/app/views/applicant/NorthStarApplicantUpsellView.java [hunk 2]
```diff
@@ -20,92 +20,97 @@
...
+
+        context.setVariable(
+                "pageTitle", params.messages().at(MessageKey.TITLE_APPLICATION_CONFIRMATION.getKeyName()));
+
+        context.setVariable("programTitle", params.programTitle().orElse(""));
+        context.setVariable("programDescription", params.programDescription().orElse(""));
+        context.setVariable("applicationId", params.applicationId());
+        context.setVariable("bannerMessage", params.bannerMessage());
+
+        ImmutableList<String> additionalText = ImmutableList.of("Text1", "Text2");
```
melanie-exygy, line 62: **question (blocking):** Why are we adding this additional text?  I don't think that was in the original alert.

melanie-exygy (commented):
In the view classes, I'm seeing a lot more lines changed than I think we should need to change.  Not sure if that's a formatting change or what.  Please run the `bin/fmt` command.  That will automatically run the linter and apply our standard formatting.
Then, I suggest only replacing the constructor rather than making more extensive changes.

melanie-exygy:
> Hi @melanie-exygy @rockycodes I have closed previous PR as I was facing issues and please review this ... can you explain what are these checks and does all checks need to be satisfied ?
> 
Hi @rabelmervin , thanks again for your work!  Yes, all the checks will need to pass.  The good news is that I think most of these tests fail because of an easy-to-fix compilation error, which you can see if you click on the "Details" link next to the check.

image

Looks like you need to add an import statement for the `ImmutableList` class in the `NorthStarApplicantUpsellView`.

As for the first check, run `bin/fmt` in your environment and that will fix any formatting issues.  Once you check those fixes in, that check will pass.

server/app/services/AlertSettings.java
```diff
@@ -23,26 +23,59 @@ public record AlertSettings(
...
-      AlertType alertType,
-      ImmutableList<String> additionalText,
-      Boolean isSlim) {
-    this(show, title, text, /* unescapedDescription= */ true, alertType, additionalText, isSlim);
-  }
+  public static class AlertSettingsBuilder {
+    private Boolean show = false;
+    private Optional<String> title = Optional.empty();
+    private String text = "";
+    private Boolean unescapedDescription = false;
```
rockycodes, line 37: suggestion (blocking): I think we want `unescapedDescription` to default to `true` to mimic the past behavior.

server/app/views/applicant/NorthStarAddressCorrectionBlockView.java
```diff
@@ -20,100 +20,98 @@
...
-            /* show= */ true,
-            Optional.of(params.messages().at(MessageKey.ADDRESS_CORRECTION_LINE_1.getKeyName())),
-            alertMessage,
-            AlertType.WARNING);
-    context.setVariable("addressAlertSettings", addressAlertSettings);
+        AlertSettings addressAlertSettings = AlertSettings.builder()
+                .show(true)
+                .title(Optional.of(params.messages().at(MessageKey.ADDRESS_CORRECTION_LINE_1.getKeyName())))
+                .text(alertMessage)
+                .unescapedDescription(false)
```
rockycodes, line 85: suggestion (blocking): Again, double check this, but I'm pretty sure `unescaptedDescription` defaulted to `true` previously.

server/app/views/applicant/NorthStarApplicantUpsellView.java
```diff
@@ -20,92 +20,97 @@
...
+
+        String alertTitle = params
+                .messages()
+                .at(MessageKey.ALERT_SUBMITTED.getKeyName(), params.programTitle().orElse(""));
+        AlertSettings successAlertSettings = AlertSettings.builder()
+                .show(true)
+                .title(Optional.of("Alert Title"))
+                .text("Alert Text")
+                .unescapedDescription(false)
+                .alertType(AlertType.INFO)
```
rockycodes, line 72: suggestion/question (maybe blocking): please double check, but I think this was `AlertType.SUCCESS` previously.

server/app/services/AlertSettings.java
```diff
@@ -23,26 +23,59 @@ public record AlertSettings(
...
+      return this;
+    }
+
+    public AlertSettingsBuilder isSlim(Boolean isSlim) {
+      this.isSlim = isSlim;
+      return this;
+    }
 
-  /** Sanitized HTML for the alert text that processes Markdown. */
-  public String getFormattedAlertText(String text) {
```
rockycodes, line 44: suggestion (blocking): Keep this method. It is used in the template here: https://github.com/civiform/civiform/blob/3ffef58e5c896e2912b4c6fbc0476b4af6968fa8/server/app/views/components/AlertFragment.html#L39

rockycodes (changes requested):
I've added a few comments, but the formatting changes are making it a little hard to read so please ping me again after you've run `bin/fmt`. Also, if you haven't already, please make sure the app compiles and runs by running `bin/run-dev' from the root of your repository. I just tried it and it wouldn't compile for me (it gave me the error message Melanie mentioned) 
image

rabelmervin:
I apologise @rockycodes @melanie-exygy Soon ill recorrect everything

server/app/views/applicant/NorthStarApplicantUpsellView.java
melanie-exygy: Suggested solution:  You can leave out this line since we want to set the additionalText to what it was before, which is an empty list:  `ImmutableList.of()`.  Since `ImmutableList.of()` is the default for `additionalText` in your new builder, you don't need to set it here.

server/app/views/applicant/NorthStarAddressCorrectionBlockView.java
```diff
@@ -83,11 +83,16 @@ public String render(
...
-            AlertType.WARNING);
+        AlertSettings.builder()
+            .show(true)
+            .title(
+                Optional.of(
+                    params.messages().at(MessageKey.ADDRESS_CORRECTION_LINE_1.getKeyName())))
+            .text(alertMessage)
+            .unescapedDescription(true)
+            .alertType(AlertType.WARNING)
+            .isSlim(true)
```
melanie-exygy, line 94: **issue (blocking):** `isSlim` should be `false` in this case, which means that you can remove this line altogether since `false` is default.  I think this will fix some of your browser tests, by the way.

server/app/views/applicant/NorthStarApplicantUpsellView.java
```diff
@@ -65,7 +65,15 @@ public String render(UpsellParams params) {
             .messages()
             .at(MessageKey.ALERT_SUBMITTED.getKeyName(), params.programTitle().orElse(""));
     AlertSettings successAlertSettings =
-        new AlertSettings(/* show= */ true, Optional.of(alertTitle), "", AlertType.SUCCESS);
+        AlertSettings.builder()
+            .show(true)
+            .title(Optional.of(alertTitle))
+            .text("Alert Text")
```
melanie-exygy, line 71: **issue (blocking):** The alert text should be an empty string, just like in the constructor you are replacing.

server/app/views/applicant/NorthStarProgramOverviewView.java
```diff
@@ -107,14 +107,18 @@ private String getProgramDescription(
...
-            AlertType.INFO,
-            ImmutableList.of(),
-            /* isSlim= */ true);
+        AlertSettings.builder()
+            .show(true)
+            .title(Optional.empty())
+            .text(alertText)
+            .unescapedDescription(true)
+            .alertType(AlertType.INFO)
+            .additionalText(ImmutableList.of())
```
melanie-exygy, line 118: **suggestion (non-blocking):** I think you can remove this line since `ImmutableList.of()` is the default for `additionalText`.

rockycodes (changes requested):
This is looking really good! Just one outstanding comment about defaulting `unescapedDescription` to `true` to match the previous behavior. 

Is the goal to refactor all instances of `AlertSettings` to use this builder? I'm wondering if we can get rid of the params on the record and refactor the use of `AlertSettings` in `NorthStarBaseView`. @melanie-exygy curious about your thoughts on this also.

rabelmervin:
> This is looking really good! Just one outstanding comment about defaulting `unescapedDescription` to `true` to match the previous behavior.
> 
> Is the goal to refactor all instances of `AlertSettings` to use this builder? I'm wondering if we can get rid of the params on the record and refactor the use of `AlertSettings` in `NorthStarBaseView`. @melanie-exygy curious about your thoughts on this also.

Extremely sorry @rockycodes. I have changed it !

melanie-exygy:
> This is looking really good! Just one outstanding comment about defaulting `unescapedDescription` to `true` to match the previous behavior.
> 
> Is the goal to refactor all instances of `AlertSettings` to use this builder? I'm wondering if we can get rid of the params on the record and refactor the use of `AlertSettings` in `NorthStarBaseView`. @melanie-exygy curious about your thoughts on this also.

Yes, I think it would be ideal for all instances of `AlertSettings` to use this builder.  @rabelmervin, I agree with Rocky that it would be an improvement to use the builder in `NorthStarBaseView`.  However, I don't think we can get rid of the params on the record because doing that would remove those fields on the record altogether.  So, the implicit record constructor would remain, but I don't think there's any way around that.

rabelmervin:
sure @rockycodes @melanie-exygy i can add the builder to northstarbaseview too.

PR #9649, opened 2025-01-31
Add spacing at the end of applicant upsell contents (#9391)

### LINKED ISSUE #9391: "Back to homepage" button needs more padding
image

EmiliaPaz:
### Description
This PR adds spacing in between 'Back to homepage' button and footer in the `ApplicantUpsell` page when no program cards are visible. 

#### Background
Generally, each content in the page is inside a `<section>` which has a top margin. This works nicely for the main contents. However, the footer doesn’t add a top margin which could mean there is no space in between contents and footer.

Some pages handle this by adding bottom margin to the “last element” or an extra view:
1. `ApplicantProgramBlockEdit` adds margins to the last element ('Review and exit' button) via `margin-bottom-4` 
2.  `ApplicantProgramSummary` adds spacing after the last element via `<div class="page-spacing"></div>`

#### Issue
When the `ApplicantUpsell` page has program cards at the end, the program cards include a bottom margin. However, if there are no program cards there is no spacing in between the 'Back to homepage' button and footer.

|||
|----------------------------------------------|----------------------------------------------|
| Programs Card section hidden | Programs Card section visible   |

#### Solution
Add spacing between the “back to homepage” button and footer, but taking into account the Programs Card section visibility since we don’t want to add double space.

|||
|----------------------------------------------|----------------------------------------------|
| Programs Card section hidden | Programs Card section visible   |

#### Other Solutions Explored
a. Add margin to the 'Back to homepage' button. Would be more complicated to add margins conditionally on Programs Card section visibility. Also, would be more complicated if we ever go with a more consistent way of adding spacing (explained on solution b) 
b. Add top margin to the footer or page spacing before the footer to be consistent about how spacing is added in between the main content and the footer across the various pages. This would require making sure each page contents doesn't add its own bottom margins to the last element to avoid double spacing. This could be tricky if the last element is using a class that adds margins and is used somewhere else too. Additionally, this is my first contribution so I don't want to potentially affect other pages so decided to keep the fix only at the`ApplicantUpsell` page

#### General

- [n/a] Extended the README / documentation, if necessary. For user-facing features, consider updating [the user docs](https://github.com/civiform/docs). For "under-the-hood" changes or things more relevant to developers, consider updating [the dev wiki](https://github.com/civiform/civiform/wiki).
- [n/a] Ensured PII wasn't added to any new logs, unless it was guarded by `isDevOrStaging`

#### User visible changes

- [n/a] Followed steps to [internationalize](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#application-strings) any new strings
  - [n/a] Added context strings to new [messages](https://github.com/civiform/civiform/blob/main/server/conf/i18n/messages)
  - [n/a] Didn't use a message in applicant facing code that isn't translated yet (unless behind a flag)
- [n/a] Made equivalent changes in Thymeleaf for applicant-facing features and changes (or created an issue in the [Northstar epic](https://github.com/orgs/civiform/projects/1/views/94) to track that it's missing from Thymeleaf)

1. Log in to CiviForm as an admin
2. Enable North Star flag
3. Add two programs
4. Log out of CiviForm as an admin
5. As an applicant, apply for the first program. After submitting, `ApplicantUpsell` page should be visible with program cards on the bottom. Verify there is margin in between the program cards and the footer
6. Go back to the applicant home page. Apply for the second program. After submitting `ApplicantUpsell` page should be visible with no program cards on the bottom. Verify there is margin in between the 'Back to homepage' button and the footer

### Issue(s) this completes

Fixes #9391


server/app/views/applicant/ApplicantUpsellTemplate.html
```diff
@@ -124,6 +121,10 @@ <h2 th:text="#{content.otherProgramsToApplyFor}"></h2>
           ></div>
         </div>
       </section>
+
+      <div th:unless="${showProgramsCardsSection}" class="page-spacing">
+        <!-- Add spacing only when the 'Program Cards' section is hidden, since that section adds its own bottom margins. !-->
```
gwendolyngoetz, line 126: (blocking) Please swap over to use [Thymeleaf parser-level comment blocks](https://www.thymeleaf.org/doc/tutorials/3.1/usingthymeleaf.html#thymeleaf-parser-level-comment-blocks). It prevents them from being rendered to the client.

gwendolyngoetz (changes requested):
Looks good, just one minor change.

server/app/views/applicant/ApplicantUpsellTemplate.html
EmiliaPaz: Done. Is there a max char per line? I only found bin/fmt for formatting, but that doesn't include HTML

gwendolyngoetz: Unless the formatter makes changes, do what feels best.

gwendolyngoetz (approved):

PR #9650, opened 2025-02-01
Use short description on admin program index page

### LINKED ISSUE #9284: Use new short description field on admin program index page with long description as fallback
We want to default to use the new short description here on the admin program index page:
Image

Since older programs might not have a short description, we should fall back to using the existing long description field.

codekarma:
### Description

Use short description on admin program index page when available.

#### General

#### User visible changes

1. Create a new program and visit the admin program index view page.
2. Ensure that the short description is shown instead of long program description. When short description is not available, we should display the long description.

### Issue(s) this completes

Fixes #9284


dkatzz (approved):
Looks great, thank you!!

server/app/views/components/ProgramCardFactory.java
```diff
@@ -88,7 +91,11 @@ public DivTag renderCard(ProgramCardData cardData, boolean showCategories) {
                                     /* preserveEmptyLines= */ false,
                                     /* addRequiredIndicator= */ false))
                             .withClasses(
-                                "line-clamp-2", "text-sm", StyleUtils.responsiveLarge("text-base")))
+                                ReferenceClasses.ADMIN_PROGRAM_CARD_DESCRIPTION,
+                                "line-clamp-2",
+                                "text-sm",
+                                StyleUtils.responsiveLarge("text-base"),
+                                "mb-4"))
```
melanie-exygy, line 98: **question:** (maybe blocking) Did UX ask for this bottom margin?  If they didn't, would you mind sending Sijia a screenshot diff to make sure they're okay with it?  I noticed that it does make each card taller and I don't know if that makes it more difficult to go through the list of cards since less cards fit on the screen at once.

server/test/controllers/admin/AdminProgramControllerTest.java
```diff
@@ -926,9 +929,9 @@ public void update_allowsChangingCommonIntakeAfterConfirming() throws Exception
                     "localizedDisplayName",
                     newProgramName,
                     "localizedDisplayDescription",
-                    newProgramDescription,
```
melanie-exygy, line 929: **question (non-blocking):** Why not continue to use the variable for both the long and short descriptions since we refer to them both in this map?

server/app/views/components/ProgramCardFactory.java
codekarma: Thanks for pointing this out. I fixed the margins so that no extra space is added in the program card. The problem was that the Admin Notes element was adding a top margin and therefore when there was no admin notes, there was no margin between description and Visibility status. This was an existing problem that now this PR is also fixing by removing the top margins and only using bottom margins to ensure that there is always a single margin between the blocks(Note that there is no margin between Visibility status and Admin notes and I assume it is by design because there is a an existing browser test that verify this UX treatment).

browser-test/src/admin/admin_program_list.test.ts
```diff
@@ -48,6 +48,9 @@ test.describe('Program list page.', () => {
   test('view program with categories', async ({page, adminPrograms}) => {
```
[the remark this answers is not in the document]
codekarma: I've added a new test.

server/test/controllers/admin/AdminProgramControllerTest.java
codekarma: Only reason I removed the variable because it is not needed anymore. We are only verifying the short description and never using the long description in the test.

server/app/views/components/ProgramCardFactory.java
codekarma: See L110 where we remote `pt-4`. I think this was adding the top margin. Also on L102, we changed the class from `pt-4` to `mb-4` to remove top-padding and add bottom-padding. Sorry I can't seem to find that added space between Admin note and description for ex. the screenshot diff: browser-test/image_snapshots/admin_program_list_test/program-list-active-and-draft-versions.png has no space between two. Can you please point me to the screen diff you are looking at where the increase in distance is visible?

codekarma: Also the last commit has screenshots that shows that the distance reduced and back to normal: https://github.com/civiform/civiform/pull/9650/commits/880a80cfb6ba62764a6a0d8f81fbbe48112e89c5

codekarma (commented):
Addressed comments.

server/app/views/components/ProgramCardFactory.java
melanie-exygy: Got it.  I see now that the latest screenshots put the distances back to normal.  thanks!

melanie-exygy (approved):

PR #9561, opened 2025-01-17
Enable Multifile Upload

gwendolyngoetz:
### Description

Remove the flag and related code to enable the multifile upload feature. Enables the data migration job to run once.

This will enable multifile uploads for everyone. Due to the data migration there is no disabling or reverting this feature.

## Release notes

This release now allows for multiple files to be uploaded to a single file upload question.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [x] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)
- [x] Removed the release notes section if the title is sufficient for the release notes description, or put more details in that section.
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)

### Issue(s) this completes

Fixes #7493, #8362, #8563

PR #9680, opened 2025-02-06
Add multi-file upload support for Azure

### LINKED ISSUE #9444: File names are missing for applicant file uploads in Azure with multi-file-upload enabled.

lagosr-google:
### Description

Add multi-file upload support for Azure.

This PR re-introduces the multi-file support for Azure, and fixes the previous problem found with this code.

The problem was related to code that retrieved the original file names from storage, it would return either the original filenames from storage, or the filekeys if the original filenames could not be found.

This was a problem, when the controller code assumed that the retrieval code mentioned above would return nothing if the original filenames were not in storage.

The PR contains the original work to implement multi-file (as originally done in https://github.com/civiform/civiform/pull/9561) with fixes to address issue: https://github.com/civiform/civiform/issues/9444

A test was modified to reproduce the bug that caused the previous rollback of this feature, and its verified to pass now.

Separately a doc is being prepared to review how multi file uploads behave that is different in Azure vs AWS deployments, to review the tech-debt that exists, and make a decision if a refactor is warranted.

## Release notes

Enable multi-file uploads for applicants in Azure.

#### General

### Issue(s) this completes

Fixes #9444


server/app/services/applicant/ReadOnlyApplicantProgramService.java
```diff
@@ -467,8 +467,18 @@ private void addDataToBuilder(
 
           if (fileUploadQuestion.getFileKeyListValue().isPresent()) {
             ImmutableList<String> fileKeys = fileUploadQuestion.getFileKeyListValue().get();
+            Optional<ImmutableList<String>> originalFileNamesOptional =
```
dkatzz, line 470: **question:** This is the main change in this PR that's different from the original, right?

server/app/services/applicant/question/FileUploadQuestion.java
```diff
@@ -97,18 +96,28 @@ public Optional<String> getOriginalFileName() {
   }
 
   /*
-   * Returns the stored original filenames. If this data does not exist, return the file key values.
+   * Returns the stored original filenames, if they exist in storage.
    */
   public Optional<ImmutableList<String>> getOriginalFileNameListValue() {
-    return getFileKeyListValue();
+    return applicantQuestion.getApplicantData().readStringList(getOriginalFileNameListPath());
```
dkatzz, line 102: **question:** Will these always be empty for AWS?

dkatzz (commented):
We should add a browser test for the issue that came up as well. Looking forward to seeing the writeup to understand level of effort for a refactor of this code

gwendolyngoetz (approved):

server/app/services/applicant/question/FileUploadQuestion.java
lagosr-google: yes --  this should be empty for AWS, since we dont store them in the table.

The change was to not fall back in case there is no original file names in the stored table.

because i had at one point fallen back, that lead to the bug.

server/app/services/applicant/ReadOnlyApplicantProgramService.java
lagosr-google: yes.

The big difference was to not be too smart in the lower level code that got the original file names.. -- where the code would fall back to returning the filenames from the filekeys if there was no original filenames stored in the database.

that original implementation led to problems where a conditional was looking at the list of original filenames being returned by these functions. 

so now these functions can return empty lists, and code further upstream will decide to look for filenames from the filekeys when that happens.

Issue #5461, opened 2023-08-22
Use contact info from IDCS create account to populate seeded questions


elisekalstad:
**Is your feature request related to a problem? Please describe.**
As a resident, it is frustrating to repeat the same contact information when creating an account for the Seattle IDCS login and also in the program applications within CiviForm. I enter my name, email, and phone number when creating an Seattle IDCS Login and most programs ask me to re-enter that same information when I go to apply.

**Describe the solution you'd like**
I would like to enter my contact information in IDCS and have it show up automatically within the program application.

**Done when**
A resident provides their name, email, and phone to create their Seattle login AND it auto-populates within the program form. 

**Additional context**
Business requirement to onboard Fresh Bucks Public Interest Form

elisekalstad:
![Image](https://github.com/civiform/civiform/assets/86082659/766f2208-44d7-4137-a36f-0474ddaf15b0)

elisekalstad:
Reviewed with Seattle team on 9/5: Medium to Large size body of work. 

@msprenke to review and draft PRD
@swatkat1  to review PRD and draft TDD

msprenke:
@gwendolyngoetz that will be a challenge for sure.  The new seeded phone number question has the formatting.  But most applications haven't been converted to the new question due to the change it would cause in the CSV so this may be the event that makes us switch them over.  Once i'm done with the requirements document the plan will need to be figured out for the technical design.

msprenke:
PRD https://docs.google.com/document/d/1MBNZcS69znsMh9nap6SIIYMtg7h3_9BzJ4GvK6dLOOk/edit?usp=sharing&resourcekey=0-089W9JF7fqAa8mBU8YP52w

msprenke:
Moving to ready for development.  Will discuss with Exygy, but this only affects Seattle at this time

gwendolyngoetz:
As of the time of this comment Seattle's IDCS does not return phone number as part of the JWT. This is a known issue our IDCS team said they'd investigate to see if they can make it work correctly.

PR #9691, opened 2025-02-07
Adding phone validation for non-user interactive code with E.164 support

gwendolyngoetz:
### Description

Adds phone validation to support for [E164 format numbers](https://www.twilio.com/docs/glossary/what-e164) ([E164 Spec](https://www.itu.int/rec/t-rec-e.164/en)).

#### General

### Issue(s) this completes

Related to: #5461

### DIFF
--- a/server/app/services/PhoneValidationUtils.java
+++ b/server/app/services/PhoneValidationUtils.java
@@ -109,4 +109,80 @@ public final class PhoneValidationUtils {
 
     return PhoneValidationResult.create(phoneNumber, countryCode);
   }
+
+  /**
+   * Checks the supplied phone number with supported country codes to determine which, if any,
+   * country code is valid.
+   *
+   * <p>Really deep dive into E.164 can be found on the <a
+   * href="https://www.itu.int/rec/t-rec-e.164/en">ITU site</a>, but a short overview can be seen on
+   * <a href="https://www.twilio.com/docs/glossary/what-e164">Twilio's site</a>.
+   *
+   * @param phoneNumber {@link Optional<String>}} of the Phone Number
+   * @return an {@link Optional<String>} representing the <a
+   *     href="https://www.javadoc.io/doc/com.googlecode.libphonenumber/libphonenumber/8.4.1/com/google/i18n/phonenumbers/PhoneNumberUtil.html">CLDR
+   *     format</a> Country Code if the phone number is valid, or an Optional.empty() if the phone
+   *     number is invalid
+   */
+  public static PhoneValidationResult determineCountryCodeForE164PhoneNumber(
+      Optional<String> phoneNumber) {
+    PhoneValidationResult result = PhoneValidationResult.builder().build();
+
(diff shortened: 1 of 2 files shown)


gwendolyngoetz:
Added a new method to support E164 numbers and not return formatted. At some point I'd like to merge these, but I was getting too many issues with existing tests and lack time right now.

server/app/services/PhoneValidationUtils.java
```diff
@@ -109,4 +109,80 @@ private static PhoneValidationResult validate(
...
+        countryCode = Optional.of(PHONE_NUMBER_UTIL.getRegionCodeForNumber(parsedPhoneNumber));
+      }
+
+      if (!PHONE_NUMBER_UTIL.isValidNumber(parsedPhoneNumber)
+          || !PHONE_NUMBER_UTIL.isValidNumberForRegion(parsedPhoneNumber, countryCode.orElse(""))) {
+        return PhoneValidationResult.createError(
+            phoneNumber, MessageKey.PHONE_VALIDATION_INVALID_PHONE_NUMBER);
+      }
+
+      if (!POTENTIAL_COUNTRY_CODES.contains(countryCode.orElse(""))) {
```
dkatzz, line 176: **suggestion:** If the country code used isn't supported, should we throw a different error? Because the phone number may be valid, but just not a US / CA number, right?

server/test/services/PhoneValidationUtilsTest.java
```diff
@@ -136,4 +136,41 @@ public void determineCountryCode_phoneNumberHasExpectedCountryCode(
...
+      new Object[] {Optional.of("+12503613152"), COUNTRY_CODE_CA, Optional.of("2503613152")},
+      new Object[] {Optional.of("253-867-5309"), COUNTRY_CODE_US, Optional.of("2538675309")},
+      new Object[] {Optional.of("250-361-3152"), COUNTRY_CODE_CA, Optional.of("2503613152")},
+      new Object[] {Optional.of("2538675309"), COUNTRY_CODE_US, Optional.of("2538675309")},
+      new Object[] {Optional.of("2503613152"), COUNTRY_CODE_CA, Optional.of("2503613152")}
+    };
+  }
+
+  @Test
+  @Parameters(method = "numbersWithAllowedE164FormatOptions")
```
dkatzz, line 164: **praise:** Nice!

dkatzz (commented):
One comment / thought about the validation for unapproved country numbers - otherwise LGTM.

server/app/services/PhoneValidationUtils.java
gwendolyngoetz: I needed something that I can run without user interaction (for importing in from OIDC). The existing method already in use deals with more detailed validation messages, but because of the way it returns at different steps I was having difficulty getting through it in order to handle E164.

The whole thing needs to be rewritten to let the phone number library do the hard work, but because of the assumptions in how the method was going to be consumed by the UI it was going to need UI rework.

dkatzz (approved):

dkatzz (approved):

PR #9657, opened 2025-02-03
[North Star] Update program cards to use correct description and link to overview page

rockycodes:
### Description

The PR updates the program cards in the North Star Applicant Homepage so that:
1. they show the program's short description if it's available, or the long description stripped of markdown if short description is not availalbe
2. the apply button takes the applicant to the program overview page if they do not already have a submitted or draft application for that program

Most of the changed files are tests (I had to update a lot of screenshots and browser tests). All of the updated app logic lives in `ProgramCardsSectionParamsFactory.java` and `ApplicantRoutes.java`, plus a little method I added in `TextFormatter.java` to remove markdown.

I added a test file for `ProgramCardsSectionParamsFactory` to test the helper methods I added.

#### General

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [x] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)
- [x] Removed the release notes section if the title is sufficient for the release notes description, or put more details in that section.
- [x] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [x] Ensured PII wasn't added to any new logs, unless it was guarded by `isDevOrStaging`

#### User visible changes

- [x] Manually evaluated tab order
- [x] Manually tested with a screen reader if the feature is applicant-facing. See [screen reader testing](https://github.com/civiform/civiform/wiki/Testing#screen-reader-testing)

### Instructions for manual testing

To test that the apply button links correctly:
1. as an admin, turn on the north star flag and add + publish some programs 
2. as an applicant, click to apply to one of the programs -- you should be taken to the program overview page
3. as an applicant, partially apply to the program and then go back to the homepage. click on the apply button again -- you should be taken to where you left off in the application
4. an an applicant, submit your application and go back to the home page. click on the apply button again -- you should be taken to the review page

To test that the description shows up correctly:
1. as an admin, turn on the north star flag, add a program with a short description, and publish it
2. as an applicant, view the Applicant Homepage and confirm that the short description shows up on the program card

To test that the long description shows up as a backup:
[...]

PR #9692, opened 2025-02-07
[Northstar]Exclude in-progress program from filtered programs

### LINKED ISSUE #9266: [North Star] For TIs, the filtered program view includes programs that are already in My applications section
**Describe the bug**
When a TI is applying on behalf of a client, the filtered program view includes programs that are already in the "My applications" section.  Programs in the "My applications" section shouldn't also appear below in the "Un-started programs" section.

**To Reproduce**

1. Log in as a CiviForm admin.
2. Edit a few programs and add a variety of categories to them.
3. Publish your changes.
4. Log in as a Trusted Intermediary.
5. Create a new client if you don't already have one.
6. In the client list, click "View applications".
7. From the homepage, start applying to a program.
8. Go back to the homepage.
9. Select a category and click "Filter".
10. Note that the program appears both in the "My applications" section and in the "Other programs and services" section.

**Expected behavior**
Programs that are in the My applications section should not show below.

**Screenshots**
image

codekarma:
### Description

Program filters view(by category) fragment includes the programs with in-progress applications in `Other programs and services` section on the program index view for Trusted Intermediaries.  There are three part of this change:
- When applying the filter in NorthStartFilteredProgramsViewPartial, exclude the in-progress programs along with filtered programs
- Include the applicantId in the category filter form in `ProgramFiltersFragment`. This is to ensure that we fetch the `ApplicantProgram`s for a specific applicant. Since the request is made from TI account, there is no applicant Id in the request. Here we pass the applicant Id explicitly
- Parse the applicant id(Long value) from the form request and use that if provided. If not provided, we try to get applicantId from the request.

#### General

#### User visible changes

- Enable North Star and program filtering feature flag on
- Create four Programs with different categories
- Login as TI
- Create a client
- Go to the Program Index page for this client
- Apply for one of the programs but don't submit the application(in-progress program application).
- Apply for one of the programs and submit the application(submitted program application).
- Now go back to Program Index page and ensure that both in-progress and submitted programs are in `My application` section
- Now choose a category to filter the rest of the programs
- Make sure that there is chosen program is in `Programs based on your selection` section and only the fourth program that is not in progress or submitted or filtered is in the `Other programs and services` section
- Ensure that the number of programs in each section are correct.

### Issue(s) this completes

Fixes #9266, #9711


browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -982,6 +985,90 @@ test.describe('applicant program index page', () => {
           })
         })
 
+        test('with program filters enabled, categorizes programs correctly for Trusted Intermediaries', async ({
```
dkatzz, line 988: **suggestion:** I think this test may be better suited in the https://github.com/civiform/civiform/blob/main/browser-test/src/trusted_intermediary.test.ts test to keep this one specific to normal applicants

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -1341,4 +1342,160 @@ test.describe('Trusted intermediaries', () => {
...
+          /* filtersOn= */ false,
+          /* northStarEnabled= */ true,
+        )
+      })
+
+      await test.step('Select a filter, click the filter submit button and verify the Recommended and Other programs sections', async () => {
+        await applicantQuestions.filterProgramsByCategory('General')
+        await validateScreenshot(
+          page.locator('#programs-list'),
+          'north-star-in-progress-homepage-programs-filtered',
```
melanie-exygy, line 1470: **suggestion (non-blocking):** We try to avoid screenshots as much as possible if there is already another screenshot of the page.  Also, you're able to test that programs appear in the correct sections programatically.

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -1341,4 +1342,160 @@ test.describe('Trusted intermediaries', () => {
       })
     })
   })
+
+  test(
+    'with program filters enabled, categorizes programs correctly for Trusted Intermediaries',
```
melanie-exygy, line 1347: **suggestion (non-blocking):** I suggest adding "...on the applicant home page" at the end of this test name.  This is because most of the tests in this file are for the trusted intermediary home page (which is admin-side) and someone reading this could be confused as to what page we're testing.

melanie-exygy (commented):
I'm noticing a tangential bug that exists on staging, but that also prevents us from testing this current fix.  

To replicate:
1. Log in as a TI
2. Click on "View applications" for a client.
3. Apply a filter.
4. Click on "View and apply" on a card in the "Programs based on your selections" section.
5. Note that when you get to the program overview, it says that you are applying for "Fake_TI" rather than for the client.

I'm pretty sure the bug was introduced in this PR:  #9657.  I'm pretty familiar, so can help you dig into it this afternoon.  I think we need to fix that bug in order to go ahead with your current PR.

server/app/views/applicant/NorthStarFilteredProgramsViewPartial.java
```diff
@@ -76,10 +78,14 @@ public String render(
...
-    // Find all programs that don't have any of the selected categories
+    Set<ApplicantService.ApplicantProgramData> inProgressOrFilteredPrograms =
+        Stream.concat(applicationPrograms.inProgress().stream(), filteredPrograms.stream())
+            .collect(Collectors.toSet());
+
+    // Find all programs that don't have any of the selected categories or are not in-progress
     ImmutableList<ApplicantService.ApplicantProgramData> otherPrograms =
         applicationPrograms.unapplied().stream()
-            .filter(programData -> !filteredPrograms.contains(programData))
+            .filter(programData -> !inProgressOrFilteredPrograms.contains(programData))
```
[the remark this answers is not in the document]
codekarma: It appears that unapplied includes all in-progress programs. See in the screenshot, program `0` is both in unapplied and in-progress.

image

server/app/controllers/applicant/ApplicantProgramsController.java
```diff
@@ -319,7 +320,7 @@ public CompletionStage<Result> hxFilter(Request request, List<String> categories
                         northStarFilteredProgramsViewPartial.render(
                             messagesApi.preferred(request),
                             request,
-                            Optional.empty(),
+                            maybeApplicantId,
```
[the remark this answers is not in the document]
codekarma: You are right that applicantId is only needed in `northStartFilteredProgramsViewPartial` to update the language action as you pointed out. Passing of applicantId as an argument to `northStartFilteredProgramsViewPartial` is unrelated to this change(and I pass Optional.empty to clear the confusion) but we need applicantId to get `relevantProgramsForApplicant` on L313. Call on 308 returns programs that does not include information related to the applicant like which programs are in progress etc. 
Since we are logged in as TI, the request does not have applicantId(Or `getApplicantId(request)`) returns `Optional.empty()`). If you look at [ProfileUtils](https://github.com/civiform/civiform/blob/main/server/app/auth/ProfileUtils.java#L115), we extract applicantId by looking searching for `APPLICANT_ID_ATTRIBUTE_NAME` attribute in profileData. This attribute does not have any value when logged in as TI

server/app/views/applicant/NorthStarFilteredProgramsViewPartial.java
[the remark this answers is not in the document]
melanie-exygy: I think `0` in this case is the index, no?  I'm debugging (as a guest, not a TI) and I don't see an overlap between inProgress and unapplied.  Note the program ids to the right:

image

server/app/controllers/applicant/ApplicantProgramsController.java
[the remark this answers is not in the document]
melanie-exygy: Ah, I see.  Thanks for explaining.  That's very helpful.

browser-test/src/applicant/applicant_application_index.test.ts
codekarma: I moved it to `northstart_trsited_intermediary.test.ts`

browser-test/src/trusted_intermediary.test.ts
codekarma: Removed the screenshot match with specific elements state verification.

server/app/views/applicant/NorthStarFilteredProgramsViewPartial.java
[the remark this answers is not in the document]
codekarma: You were right. It turns out I just didn't need the filtering of `in-progress` application. I removed that logic.

server/app/controllers/applicant/ApplicantProgramsController.java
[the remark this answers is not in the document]
codekarma: It turns out that the `applicantId` passed to `northStartFilteredProgramsViewPartial`. It is not just used in setting the thymeleaf context but also passed to `programCardsSectionParamsFactory`. This sets up the apply for program workflow from the cards to use applicantId which is necessary downstream to access the programs list specific to applicant. This `applicantId` argument along the `requesterProfile` argument need to be passed to address the bug [9711](https://github.com/civiform/civiform/issues/9711)

browser-test/src/trusted_intermediary.test.ts
```diff
@@ -1341,4 +1342,160 @@ test.describe('Trusted intermediaries', () => {
...
+          firstName: 'first',
+          middleName: 'middle',
+          lastName: 'last',
+          dobDate: '2021-05-10',
+        }
+        await tiDashboard.createClient(client)
+        await tiDashboard.expectDashboardContainClient(client)
+        await tiDashboard.clickOnViewApplications()
+      })
+      await test.step('Apply to a program and verify that applied program is under my applicatins section of view application page', async () => {
```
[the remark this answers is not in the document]
codekarma: done. I added a test step to verify filtering after finishing the application.

browser-test/src/northstar_trusted_intermediary.test.ts
```diff
@@ -46,5 +49,187 @@ test.describe(
...
+              'Comprehensive Sample Program',
+            ],
+            expectedProgramsInRecommendedSection: [],
+            expectedProgramsInOtherProgramsSection: [],
+          },
+          /* filtersOn= */ false,
+          /* northStarEnabled= */ true,
+        )
+      })
+      async function verifyProgramSections() {
```
melanie-exygy, line 169: **suggestion (non-blocking):** Rather than putting this helper function here inside the test, would you mind moving it to one of the test support files, probably `applicant_questions.ts`?  This is so that we follow our usual pattern for helper functions.

codekarma: The verifications are very specific to test like which programs or how many are in which sections. I could make this test more generic but there will be too many arguments to pass for my liking. I could still do this change if that is more desirable.

codekarma: Removed this function and moved it to `applicant_questions.ts`.

dkatzz (approved):

PR #9708, opened 2025-02-10
Added More Information to NorthStar Applicant UI Footer

### LINKED ISSUE #7959: [North Star Applicant UI] Add domain info to footer
Mocks show "domain.gov" and some extra content in the footer that we don't have access to 
9kzGb8bcNV8Enwq

So for now our footer shows this
962nFj3WfCF9o4d

This bug tracks adding the necessary info. It depends exactly what content we decide to add how much plumbing we'll need to add

lucashurt:
### Description

Added More Information to NorthStar Applicant UI Footer such as the CIVIV_ENTITY BASE_URL for the domain and the COMMON_INTAKE_MORE_RESOURCES_LINK_HREF for the link to information services.

### Issue(s) this completes

Fixes #7959; Fixes #7959...


server/app/views/NorthStarBaseView.java
```diff
@@ -158,6 +176,15 @@ protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(
     return context;
   }
 
+  private String getBaseUrlWithoutScheme() {
+    Optional<String> baseUrlOptional = settingsManifest.getBaseUrl();
+
+    if (baseUrlOptional.isEmpty()) return "";
+
+    String baseUrlWithoutScheme = baseUrlOptional.get();
+    return baseUrlWithoutScheme.replaceFirst("^(https?://)", "");
```
gwendolyngoetz, line 185: (blocking) It's safer to parse this as a `java.net.URL` and get the hostname from that object.

server/app/views/NorthStarBaseView.java
```diff
@@ -158,6 +176,15 @@ protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(
     return context;
   }
 
+  private String getBaseUrlWithoutScheme() {
```
gwendolyngoetz, line 179:
```suggestion
  /**
  * Gets the site hostname from the `BASE_URL`.
  * <p>This value does not include the scheme. Do not use this for links.
  */
  private String getSiteHostname() {
```

(blocking) Please add javadoc and update the method name.

server/app/views/NorthStarBaseView.java
```diff
@@ -82,6 +82,24 @@ protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(
...
+          "moreResourcesLink",
+          "<a id=\"more-resources-link\" class=\"text-white underline font-bold\" href=\""
+              + CommonIntakeMoreResourceLink
+              + "\">"
+              + "Visit "
+              + CommonIntakeMoreResourceLink
+              + "</a>");
+    } else {
+      context.setVariable("moreResourcesLink", null);
+    }
```
gwendolyngoetz, line 102: (blocking) We need to go with a solution that isn't concatenating string to build the html. Based on what I see in the issue you should be able to move all the formatting in to the thymeleaf directly. 

The `Looking for more informationservices?` should be formatted separately from the `Visit cityorcounty` part. That 'Visit' part will be able to completely be a link. This will remove the need for the string concatentation as well as the need for using `utext`.

server/conf/i18n/messages
```diff
@@ -25,6 +25,8 @@ footer.returnToTop=Return to top
 # A message in the footer directing users to technical support. The placeholder is a link to
 # send an email to CiviForm technical support.
 footer.technicalSupport=For technical support: {0}
+# A Message in the footer that links the user to the official website of the specified Civic entry
+footer.informationServices = Looking for more information services? {0}
```
gwendolyngoetz, line 29: (blocking) Based on what I see in the screenshots for the issue there should be two new lines here instead of one.

- Keep `footer.informationServices`, but remove the '{0}' token.
- Add a new one that contains 'Visit {0}'

server/app/views/applicant/NavigationFragment.html
```diff
@@ -316,11 +316,26 @@
...
+        <th:block
+          class="usa-identifier__identity"
+          th:if="${moreResourcesLink != null}"
+        >
+          <div class="usa-identifier__container">
+            <p
+              class="usa-identifier"
+              th:utext="#{footer.informationServices(${moreResourcesLink})}"
+            ></p>
+          </div>
```
gwendolyngoetz, line 337: (note) See above notes on templating

gwendolyngoetz (changes requested):
There's a few things that need some adjustment, but good work overall.

melanie-exygy (commented):
Really good start for what has turned out to be a more involved work item than expected.  One thing that's missing are some browser tests.  I suggest adding to `footer.test.ts` or making a new `northstar_footer.test.ts` file, similar to `northstar_header.test.ts`.

PR #9740, opened 2025-02-15
Question title not wrapping properly if string is too long

### LINKED ISSUE #9049: New static text question:  UI layout unusable when entering a long string
**Describe the bug**
The create new static text question UI becomes unusable when entering a text string longer than 79 characters: 
image

image

**To Reproduce**

1. Go to Questions   https://staging-aws.civiform.dev/admin/questions/new?type=static&redirectUrl=%2Fadmin%2Fquestions
2. Click on Create new questions, Static Text
3. Enter a very long string in the "Question text displayed to the applicant" field 
4. The UI layout became unusable: the left pane became very narrow and user cannot interact with the left pane

On the "questions" screen
the long question text extended to the right and overlapped with other columns:
image

**Expected behavior**
The left pane should always stay visible. 

Additionally, if the intent is to limit the question to be less than certain max length, we should add error check to restrict the input text length. 

**Screenshots**
image

- OS: MacOS 14.7 (23H124)
- Browser:  Chrome Version 130.0.6723.70 (Official Build) (arm64)

Not tested

**Additional context**

gabrielc1317:
### Description

added break-words and max-w-full to ensure question title wraps properly

#### General

Create question with a long string of characthers

### Issue(s) this completes
Fixes #9049

### DIFF
--- a/browser-test/src/admin/admin_program_creation.test.ts
+++ b/browser-test/src/admin/admin_program_creation.test.ts
@@ -1434,6 +1434,9 @@ test.describe('program creation', () => {
           await adminPrograms.expectFormFieldDisabled(
             FormField.APPLICATION_STEPS,
           )
+          await adminPrograms.expectFormFieldDisabled(
+            FormField.CONFIRMATION_MESSAGE,
+          )
 
           // Changing the program type is allowed during program creation.
           // Therefore, all the program type options should be enabled.
@@ -1461,6 +1464,9 @@ test.describe('program creation', () => {
           await adminPrograms.expectFormFieldEnabled(
             FormField.APPLICATION_STEPS,
           )
+          await adminPrograms.expectFormFieldEnabled(
+            FormField.CONFIRMATION_MESSAGE,
+          )
         })
 
         await test.step('save external program', async () => {
@@ -1482,6 +1488,9 @@ test.describe('program creation', () => {
           )
           await adminPrograms.expectFormFieldDisabled(
             FormField.APPLICATION_STEPS,
+          )
+          await adminPrograms.expectFormFieldDisabled(
+            FormField.CONFIRMATION_MESSAGE,
           )
 
           // Changing the program type of an external program is disallowed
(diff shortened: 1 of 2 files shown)


melanie-exygy (commented):
Nice work.  I tested it out and can see that the problem is fixed on the questions list page.  Now, we'll need a fix for the question creation page as well.  That's this part of the bug:  
image
To fix formatting, run `bin/fmt`.

melanie-exygy (commented):
For the question preview, using `break-all` definitely helps, but the preview still stretches far to the left and eventually covers the "Save" button.  In testing, I noticed that this only happens when the `NORTH_STAR_APPLICANT_UI` feature flag is set to true.  Not sure if you've been testing with the flag on or off.  The North Star is our new UI and the question preview will appear differently based on that.

You can turn the feature flag on by logging in as an admin (using the 'DevTools' button on the home page).  Then, click the gear icon in the upper right.  Set the North Star flag to true and hit "Save".

I suspect the preview content is limited when the North Star is off, by the `w-3/5` here:  [QuestionPreview.java](https://github.com/civiform/civiform/blob/e83eed07b4c148ac9dc72650850184435e6515ad/server/app/views/admin/questions/QuestionPreview.java#L82).  So, we need to do something similar in `server/app/views/admin/questions/QuestionPreviewFragment.html` or one of the related Thymeleaf fragments.  You could try a max-width.

server/app/views/admin/questions/QuestionPreviewFragment.html
```diff
@@ -1,4 +1,4 @@
-<div id="sample-question" class="margin-y-4">
+<div id="sample-question" class="margin-y-4 w-3/5">
```
melanie-exygy, line 1: **suggestion (blocking):**  `w-3/5` is a TailwindCss class.  For the new North Star UI, we're moving off of Tailwind in favor of the US Web Design System (USWDS).  I think you could use this USWDS class instead:  `flex-5`.  You can check out their docs if you want to see other options: [USWDS Flexbox](https://designsystem.digital.gov/utilities/flex/).

PR #9754, opened 2025-02-18
Remove remaining uses of `TxScope.requiresNew()`

### LINKED ISSUE #9643: Remove uses of `TxScope.requiresNew()`, and use `TxScope.required()` instead
While ebeans gives you what look like nested transactions, there’s actually no such thing.

If you start the inner transaction with `TxScope.required()`, ebeans just adds whatever you’re doing to the outer transaction and ignores the `transaction.commit()` you call on the inner transaction. This is probably what you want to happen, since if something should get rolled back you probably want the whole thing to get rolled back.

But, if you start the inner transaction with `TxScope.requiresNew()`, what you actually get, from postgres’s perspective, is an entirely new, parallel transaction that starts from its own distinct DB snapshot. Additionally, anything that happens in that ~inner~ parallel transaction isn’t seen by the outer transaction, even after you go back to it, because it’s operating from an earlier DB snapshot that doesn’t included the changes made by the inner transaction, which were made after the outer transaction’s starting snapshot.

tldr: this test passes:
```java
@Test
  public void executeInTransaction_innerTransactionWithRequiresNewIsIndependent() {
    try (Transaction transaction =
        DB.beginTransaction(TxScope.required().setIsolation(TxIsolation.SERIALIZABLE))) {
          // Check initial number of accounts in the outer transaction's snapshot
          assertThat(DB.find(AccountModel.class).findCount()).isEqualTo(0);

          try (Transaction innerTransaction =
              DB.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE))) {
            new AccountModel().insert();

            // Assert that from within this inner transaction, we see the new account
            assertThat(DB.find(AccountModel.class).findCount()).isEqualTo(1);
            innerTransaction.commit();
          }

          // Assert that, back in the outer transaction, we still see the original 0 accounts
          // because we're working from an earlier snapshot.
          assertThat(DB.find(AccountModel.class).findCount()).isEqualTo(0);
          transaction.commit();
      }

    // Outside of both transactions, we should see the new account
    assertThat(DB.find(AccountModel.class).findCount()).isEqualTo(1);
  }
}
```

## Next steps
We should double check everywhere that uses `requiresNew()` and make sure that's what we intend. Those places are:
- [x] DevDatabaseSeedTask#inSerializableTransaction() (from seedProgramsCategories()) (#9735)
- [x] DatabaseSeedTask#inSerializableTransaction() (#9735)
- [ ] QuestionRepository#createOrUpdateDraft()
- [ ] VersionRepository#getDraftVersionOrCreate()
- [ ] private VersionRepository#publishNewSynchronizedVersion()
- [ ] public VersionRepository#publishNewSynchronizedVersion()

avaleske:
### Description

Remove remaining uses of `TxScope.requiresNew()` and replace them with `TxScope.required()`,  which is actually the behavior we want.

More details in #9643, but `.requiresNew()` is almost never the transaction isolation level we want, as it results in a separate, concurrent transaction whose changes are not visible to the outer transaction, rather than a nested "inner" transaction as one might expect.

Additionally, I verified all codepaths that call into these changes, and nothing expects the unusual behavior of `requiresNew()`.

#### General

- [n/a] Created unit and/or browser tests which fail without the change (if possible)
- [n/a] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [n/a] Extended the README / documentation, if necessary. For user-facing features, consider updating [the user docs](https://github.com/civiform/docs). For "under-the-hood" changes or things more relevant to developers, consider updating [the dev wiki](https://github.com/civiform/civiform/wiki).
- [n/a] Ensured PII wasn't added to any new logs, unless it was guarded by `isDevOrStaging`

### Issue(s) this completes

Fixes #9643

### DIFF
--- a/server/app/repository/VersionRepository.java
+++ b/server/app/repository/VersionRepository.java
@@ -113,7 +113,7 @@ public final class VersionRepository {
     // this operation inside of a transaction in order to ensure we have
     // consistent reads.
     Transaction transaction =
-        database.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE));
+        database.beginTransaction(TxScope.required().setIsolation(TxIsolation.SERIALIZABLE));
     try {
       VersionModel draft = getDraftVersionOrCreate();
       VersionModel active = getActiveVersion();
@@ -226,7 +226,7 @@ public final class VersionRepository {
   public void publishNewSynchronizedVersion(String programToPublishAdminName)
       throws CantPublishProgramWithSharedQuestionsException, ProgramNotFoundException {
     Transaction transaction =
-        database.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE));
+        database.beginTransaction(TxScope.required().setIsolation(TxIsolation.SERIALIZABLE));
     try {
       VersionModel existingDraft = getDraftVersionOrCreate();
       VersionModel active = getActiveVersion();
@@ -354,7 +354,7 @@ public final class VersionRepository {
     // in relative terms, but new drafts are very rare.  It is unlikely this will represent a real
     // performance penalty for any applicant - or even any admin, really.
     Transaction transaction =
-        database.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE));
+        database.beginTransaction(TxScope.required().setIsolation(TxIsolation.SERIALIZABLE));
     try {
       VersionModel newDraftVersion = new VersionModel(LifecycleStage.DRAFT);
       database.insert(newDraftVersion);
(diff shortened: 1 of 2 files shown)


avaleske:
* **#9754** <a href="https://app.graphite.dev/github/pr/civiform/civiform/9754?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/civiform/civiform/9754?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

server/app/repository/VersionRepository.java
```diff
@@ -369,6 +369,9 @@ public VersionModel getDraftVersionOrCreate() {
       transaction.commit();
       return newDraftVersion;
     } catch (NonUniqueResultException | SerializableConflictException | RollbackException e) {
+      // If we are in a nested transaction any serialization exceptions will be thrown when the
+      // outer transaction is committed, so
```
shane-exygy, line 373: nit: Wrap the comment up so it fills out this line

shane-exygy (changes requested):

shane-exygy (approved):

server/app/repository/VersionRepository.java
avaleske: fixed, not sure why the formatter has it out for me like that 😆

shane-exygy (approved):

Issue #8852, opened 2024-10-11
[Export Perf] Pre-compute eligibility at application submission time


auzlee:
Calculating eligibility during export-time is expensive, so we can speed up export by pre-calculating eligibility.

There are three pieces to this work:
1. Precompute eligibility for all applications going forward, and store it in a new column in the applications table.
2. Write a migration to compute eligibility for all past applications
3. Update everywhere in code that computes eligibility to use the already-computed value from the column.

More at the TDD
https://docs.google.com/document/d/1jsGEv1m9Ky2mbEbvTyZUbplFaPEiwNxopWPjm_XaRXs/edit?tab=t.0#heading=h.a2pzjye02740

dkatzz:
@azhangggggg can this be closed?

azhangggggg:
@dkatzz There are some things about this issue that needs to be confirmed before closing it. I'm on it now.

PR #9765, opened 2025-02-19
[Pre-compute eligibility]compute eligibility at submission

azhangggggg:
### Description

Add `eligibilityDetermination` as part of ApplicationModel. This PR computes eligibility determination when the application gets submitted.  The default value for `eligibilityDetermination` is `NOT_COMPUTED`. 

#### General

Create a program with eligibility criteria, apply to the program and either meet or not meet the eligibility criteria. Once submitting the application, go to `applications` table and look up the `eligibility_determination` column with the application id provided. You should see the expected result. 

### Issue(s) this completes

Part of #8852

### DIFF
--- a/server/app/controllers/applicant/ApplicantProgramReviewController.java
+++ b/server/app/controllers/applicant/ApplicantProgramReviewController.java
@@ -329,90 +329,85 @@ public class ApplicantProgramReviewController extends CiviFormController {
   private CompletionStage<Result> submitInternal(
       Request request, long applicantId, long programId) {
     CiviFormProfile submittingProfile = profileUtils.currentUserProfile(request);
+    CompletableFuture<ApplicationModel> submitAppFuture =
+        applicantService
+            .submitApplication(applicantId, programId, submittingProfile, request)
+            .toCompletableFuture();
 
-    try {
-      CompletableFuture<ApplicationModel> submitAppFuture =
-          applicantService
-              .submitApplication(applicantId, programId, submittingProfile, request)
-              .toCompletableFuture();
+    CompletableFuture<ReadOnlyApplicantProgramService> readOnlyApplicantProgramServiceFuture =
+        applicantService
+            .getReadOnlyApplicantProgramService(applicantId, programId)
+            .toCompletableFuture();
+    CompletableFuture<ApplicantPersonalInfo> applicantPersonalInfo =
+        applicantService.getPersonalInfo(applicantId).toCompletableFuture();
+    return CompletableFuture.allOf(readOnlyApplicantProgramServiceFuture, submitAppFuture)
+        .thenApplyAsync(
+            (v) -> {
+              ApplicationModel application = submitAppFuture.join();
+              Long applicationId = application.id;
 
-      CompletableFuture<ReadOnlyApplicantProgramService> readOnlyApplicantProgramServiceFuture =
-          applicantService
-              .getReadOnlyApplicantProgramService(applicantId, programId)
-              .toCompletableFuture();
-      CompletableFuture<ApplicantPersonalInfo> applicantPersonalInfo =
-          applicantService.getPersonalInfo(applicantId).toCompletableFuture();
-      return CompletableFuture.allOf(readOnlyApplicantProgramServiceFuture, submitAppFuture)
-          .thenApplyAsync(
-              (v) -> {
-                ApplicationModel application = submitAppFuture.join();
-                Long applicationId = application.id;
+              Call endOfProgramSubmission =
+                  routes.UpsellController.considerRegister(
+                      applicantId,
+                      programId,
+                      applicationId,
+                      applicantRoutes.index(submittingProfile, applicantId).url(),
+                      application.getSubmitTime().toString());
+              return found(endOfProgramSubmission);
+            },
+            classLoaderExecutionContext.current())
+        .exceptionally(
+            ex -> {
+              if (ex instanceof CompletionException) {
+                Throwable cause = ex.getCause();
+                if (cause instanceof ApplicationSubmissionException) {
+                  Call reviewPage =
+                      applicantRoutes.review(submittingProfile, applicantId, programId);
+                  String errorMsg =
+                      messagesApi
+                          .preferred(request)
+                          .at(MessageKey.BANNER_ERROR_SAVING_APPLICATION.getKeyName());
+                  return found(reviewPage).flashing(FlashKey.BANNER, errorMsg);
+                }
+                if (cause instanceof ApplicationOutOfDateException) {
+                  String errorMsg =
+                      messagesApi
+                          .preferred(request)
+                          .at(MessageKey.TOAST_APPLICATION_OUT_OF_DATE.getKeyName());
+                  Call reviewPage =
+                      applicantRoutes.review(submittingProfile, applicantId, programId);
+                  return redirect(reviewPage).flashing(FlashKey.ERROR, errorMsg);
+                }
+                if (cause instanceof ApplicationNotEligibleException) {
+                  ReadOnlyApplicantProgramService roApplicantProgramService =
+                      readOnlyApplicantProgramServiceFuture.join();
 
-                Call endOfProgramSubmission =
-                    routes.UpsellController.considerRegister(
-                        applicantId,
-                        programId,
-                        applicationId,
-                        applicantRoutes.index(submittingProfile, applicantId).url(),
-                        application.getSubmitTime().toString());
-                return found(endOfProgramSubmission);
-              },
-              classLoaderExecutionContext.current())
-          .exceptionally(
-              ex -> {
-                if (ex instanceof CompletionException) {
-                  Throwable cause = ex.getCause();
-                  if (cause instanceof ApplicationSubmissionException) {
-                    Call reviewPage =
-                        applicantRoutes.review(submittingProfile, applicantId, programId);
-                    String errorMsg =
-                        messagesApi
-                            .preferred(request)
-                            .at(MessageKey.BANNER_ERROR_SAVING_APPLICATION.getKeyName());
-                    return found(reviewPage).flashing(FlashKey.BANNER, errorMsg);
-                  }
-                  if (cause instanceof ApplicationOutOfDateException) {
-                    String errorMsg =
-                        messagesApi
-                            .preferred(request)
-                            .at(MessageKey.TOAST_APPLICATION_OUT_OF_DATE.getKeyName());
-                    Call reviewPage =
-                        applicantRoutes.review(submittingProfile, applicantId, programId);
-                    return redirect(reviewPage).flashing(FlashKey.ERROR, errorMsg);
-                  }
-                  if (cause instanceof ApplicationNotEligibleException) {
-                    ReadOnlyApplicantProgramService roApplicantProgramService =
-                        readOnlyApplicantProgramServiceFuture.join();
-
-                    try {
-                      ProgramDefinition programDefinition =
-                          programService.getFullProgramDefinition(programId);
-                      return renderIneligiblePage(
-                          request,
-                          submittingProfile,
-                          applicantId,
-                          applicantPersonalInfo.join(),
-                          roApplicantProgramService,
-                          programDefinition);
-                    } catch (ProgramNotFoundException e) {
-                      notFound(e.toString());
-                    }
-                  }
-                  if (cause instanceof DuplicateApplicationException) {
-                    return renderPreventDuplicateSubmissionPage(
+                  try {
+                    ProgramDefinition programDefinition =
+                        programService.getFullProgramDefinition(programId);
+                    return renderIneligiblePage(
                         request,
                         submittingProfile,
                         applicantId,
-                        readOnlyApplicantProgramServiceFuture.join(),
-                        programId);
+                        applicantPersonalInfo.join(),
+                        roApplicantProgramService,
+                        programDefinition);
+                  } catch (ProgramNotFoundException e) {
+                    notFound(e.toString());
                   }
-                  throw new RuntimeException(cause);
                 }
-                throw new RuntimeException(ex);
-              });
-    } catch (ProgramNotFoundException e) {
-      throw new RuntimeException("Could not find program.", e);
-    }
+                if (cause instanceof DuplicateApplicationException) {
+                  return renderPreventDuplicateSubmissionPage(
+                      request,
+                      submittingProfile,
+                      applicantId,
+                      readOnlyApplicantProgramServiceFuture.join(),
+                      programId);
+                }
+                throw new RuntimeException(cause);
+              }
+              throw new RuntimeException(ex);
+            });
   }
 
   // TODO(#7266): Delete the old codepath and inline the North Star path
--- a/server/app/controllers/applicant/ApplicantProgramReviewController.java
+++ b/server/app/controllers/applicant/ApplicantProgramReviewController.java
@@ -329,11 +329,11 @@ public class ApplicantProgramReviewController extends CiviFormController {
   private CompletionStage<Result> submitInternal(
       Request request, long applicantId, long programId) {
     CiviFormProfile submittingProfile = profileUtils.currentUserProfile(request);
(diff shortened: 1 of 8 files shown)


server/app/models/ApplicationModel.java
```diff
@@ -75,6 +78,10 @@ public Optional<String> getSubmitterEmail() {
     return Optional.ofNullable(this.submitterEmail);
   }
 
+  public Optional<EligibilityDetermination> getEligibilityDetermination() {
+    return Optional.ofNullable(this.eligibilityDetermination);
```
rockycodes, line 82: question (maybe blocking): When would this be null? I think with the default value you're setting we're guaranteed to always have an `eligibilityDetermination` and you don't have to worry about handling null, returning an optional from this method, and then handling optional everywhere else in the app. You should just be able to return `EligibilityDetermination` but let me know if I'm missing something.

server/app/repository/ApplicationRepository.java
```diff
@@ -295,6 +299,21 @@ public CompletionStage<Optional<ApplicationModel>> getApplication(long applicati
         executionContext.current());
   }
 
+  /** Updates the Applications table */
+  public void saveEligibilityDetermination(
+      ApplicationModel application, EligibilityDetermination eligibilityDetermination) {
+    try (Transaction transaction = database.beginTransaction()) {
```
rockycodes, line 305: question (maybe blocking): I have not been following what Austin is doing with transactions too closely, but does this match the preferred pattern?

server/app/repository/ApplicationRepository.java
```diff
@@ -147,7 +148,10 @@ private ApplicationModel submitApplicationInternal(
         app.setLifecycleStage(LifecycleStage.OBSOLETE);
         app.save();
       }
-
+      application.setEligibilityDetermination(
```
rockycodes, line 151: question (maybe blocking): Something about this seems off -- I need to think about it a little more, but I think this is where you're handling the case where `eligibilityDetermination` is null, which I'm pretty sure will never be the case. So then you're setting the eligibility to the existing eligibility on the application, which might be circular logic. Feel free to let me know if I'm not thinking about this correctly, otherwise I will look into in more depth tomorrow!

server/test/repository/ApplicationRepositoryTest.java
```diff
@@ -190,6 +191,16 @@ public void submitApplication_duplicateSubmissionsThrowsException() {
         .isInstanceOf(DuplicateApplicationException.class);
   }
 
+  @Test
+  public void submitApplication_eligibilityDeterminationDefualtNotComputed() {
```
rockycodes, line 195:
```suggestion
  public void submitApplication_eligibilityDeterminationDefaultNotComputed() {
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -591,6 +592,21 @@ CompletionStage<ApplicationModel> submitApplication(
           Optional<StatusDefinitions.Status> maybeDefaultStatus =
               activeStatusDefinitions.getDefaultStatus();
 
+          Optional<Boolean> optionalEligibilityStatus =
```
rockycodes, line 595: question (maybe blocking): Is there an `ApplicantService` test you can update to test that the eligibility is saved correctly here?

server/test/repository/ApplicationRepositoryTest.java
```diff
@@ -190,6 +191,16 @@ public void submitApplication_duplicateSubmissionsThrowsException() {
...
+    ApplicantModel applicant = saveApplicant("Alice");
+    ProgramModel program = createDraftProgram("Program");
+    ApplicationModel application =
+        repo.submitApplication(applicant, program, Optional.empty()).toCompletableFuture().join();
+    assertThat(
+        application.getEligibilityDetermination().equals(EligibilityDetermination.NOT_COMPUTED));
+  }
```
rockycodes, line 202: suggestion (maybe blocking): I think you'll want another test case that test that checks that the eligibility is saved correctly when it's not the default. But this connects to my comment in `ApplicationRepository` above, and I need to do some more thinking about how this is all hooked up.

rockycodes (changes requested):
Looks good, but I have some potentially blocking questions about the need to handle the null case and how the eligibility is getting set in `ApplicationRepository`.

server/app/models/ApplicationModel.java
avaleske: Agreed yeah, it's either default when we create it or default from the database.

server/app/repository/ApplicationRepository.java
avaleske: Agreed. Is this defaulting it to NOT_COMPUTED in the case where it's unset? But I don't think it's ever not set, so I don't think we need this.

server/app/repository/ApplicationRepository.java
avaleske: In this case I don't think we need the `application.save()` on line 312, because iiuc the `.update()` call is also making the update query.

Then, given that we only have one database interaction in this method, we don't need wrap it in a transaction cause it's just one update query, not a "read, modify in memory, write" cycle that could be susceptible to a race condition

All that said, I don't think we need this method at all? imo it'd be nicer to save the eligibility determination when the application is converted from a draft to active, by passing the eligibility_determination into the call to `submitApplication()` on line 570 of `ApplicantService`

server/app/services/applicant/ApplicantService.java
```diff
@@ -591,6 +592,21 @@ CompletionStage<ApplicationModel> submitApplication(
...
+          if (optionalEligibilityStatus.isPresent()) {
+            if (optionalEligibilityStatus.get()) {
+              eligibilityDetermination = EligibilityDetermination.ELIGIBLE;
+            } else {
+              eligibilityDetermination = EligibilityDetermination.INELIGIBLE;
+            }
+          } else {
+            eligibilityDetermination = EligibilityDetermination.NO_ELIGIBILITY_CRITERIA;
+          }
+          applicationRepository.saveEligibilityDetermination(application, eligibilityDetermination);
```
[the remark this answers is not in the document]
avaleske: Also it might be simpler to write this using Optional's methods, like
```
getApplicationEligibilityStatus(application, programDefinition)
    .map(es -> es ? EligibilityDetermination.ELIGIBLE : EligibilityDetermination.INELIGIBLE)
    .orElse(EligibilityDetermination.NO_ELIGIBILITY_CRITERIA)
```
or, perhaps in a different PR, update `getApplicationEligibilityStatus()` to return the new enum instead of a kind-of-cryptic optional boolean.

avaleske (changes requested):

server/app/services/applicant/ApplicantService.java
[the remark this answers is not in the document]
rockycodes: +1 to this idea. And then the logic in `ApplicationRepository` could be simplified to `application.setEligibilityDetermination(eligibilityDetermination)` with `eligibilityDetermination` being the param that is passed in.

server/test/repository/ApplicationRepositoryTest.java
rockycodes: If you end up altering the logic in `ApplicationRepository` to set the eligibility determination that is passed in as a param, you can ignore my comment above and just add a test which tests that the param passed in is saved to the application.

server/app/controllers/applicant/ApplicantProgramReviewController.java
```diff
@@ -330,84 +330,89 @@ private CompletionStage<Result> submitInternal(
...
-                      programId);
-                }
-                throw new RuntimeException(cause);
-              }
-              throw new RuntimeException(ex);
-            });
+                throw new RuntimeException(ex);
+              });
+    } catch (ProgramNotFoundException e) {
+      throw new RuntimeException("Could not find program.", e);
```
azhangggggg, line 414: Additional test on exception will com in a separate PR, to avoid making this one a huge PR.

server/app/services/applicant/ApplicantService.java
```diff
@@ -428,33 +429,69 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
    *     ApplicationSubmissionException} is thrown and wrapped in a `CompletionException`.
    */
   public CompletionStage<ApplicationModel> submitApplication(
-      long applicantId, long programId, CiviFormProfile submitterProfile, Request request) {
+      long applicantId, long programId, CiviFormProfile submitterProfile, Request request)
+      throws ProgramNotFoundException {
+    ProgramDefinition pd = programService.getFullProgramDefinition(programId);
```
rockycodes, line 434: suggestion (non-blocking): You could try/catch the `ProgramNotFoundException` in this method to avoid having to bubble it up and make changes to the `ApplicantProgramReviewController` but I don't feel strongly about it.

server/app/services/applicant/ApplicantService.java
```diff
@@ -428,33 +429,69 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
+          .thenCompose(
+              ro ->
+                  validateApplicationForSubmission(ro, programId)
+                      .thenComposeAsync(
+                          v -> submitterProfile.getAccount(), classLoaderExecutionContext.current())
+                      .thenComposeAsync(
+                          tiAccount -> {
+                            return getReadOnlyApplicantProgramService(applicantId, programId)
+                                .thenCompose(
+                                    ro2 -> {
```
rockycodes, line 446: suggestion (maybe blocking): I'm pretty sure you don't need to get another copy of the `ReadOnlyApplicantProgramService` and you should be able to reference the `ro` variable created above like this:
```
.thenComposeAsync(
                          tiAccount -> {
                            EligibilityDetermination eligibilityDetermination =
                                          Optional.of(pd.hasEligibilityEnabled())
                                              .filter(enabled -> enabled)
                                              .map(
                                                  enabled ->
                                                      ro.isApplicationNotEligible()
                                                          ? EligibilityDetermination.INELIGIBLE
                                                          : EligibilityDetermination.ELIGIBLE)
                                              .orElse(
                                                  EligibilityDetermination.NO_ELIGIBILITY_CRITERIA);
                          return submitApplication(.....
``` 

But honestly the async code always makes my head spin so this could have unintended consequences I'm not aware of.

server/app/services/applicant/ApplicantService.java
```diff
@@ -428,33 +429,69 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
+                                      EligibilityDetermination eligibilityDetermination =
+                                          Optional.of(pd.hasEligibilityEnabled())
```
rockycodes, line 448: suggestion (non-blocking): The use of `Optional` here with the corresponding methods is clever but it made the code harder for me to understand. Since `hasEligibilityEnabled()` always returns a boolean (is never null), I would prefer a classic if/else check.

server/app/services/applicant/ApplicantService.java
```diff
@@ -428,33 +429,69 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
         .thenCompose(
-            v ->
-                submitApplication(
-                    applicantId, programId, /* tiSubmitterEmail= */ Optional.empty(), request));
+            ro ->
+                validateApplicationForSubmission(ro, programId)
+                    .thenCompose(
+                        v -> {
+                          EligibilityDetermination eligibilityDetermination =
+                              Optional.of(pd.hasEligibilityEnabled())
```
rockycodes, line 480: suggestion (non-blocking): Since you have this same logic copied in two places you could factor it out into a private helper method which would make this long async chain a little easier to read and would make your custom logic easier to test.

server/test/models/ApplicationModelTest.java
```diff
@@ -64,6 +64,54 @@ public void latestStatusIsCarriedForwardEvenAfterApplicationUpdates() {
...
+    ApplicationModel application =
+        resourceCreator.insertActiveApplication(
+            resourceCreator.insertApplicantWithAccount(), program);
+    application.setEligibilityDetermination(EligibilityDetermination.INELIGIBLE);
+    assertThat(
+        application.getEligibilityDetermination().equals(EligibilityDetermination.INELIGIBLE));
+  }
+
+  @Test
+  public void eligibility_determination_no_eligibility_criteria() {
```
rockycodes, line 102: praise: Great comprehensive testing!

server/app/controllers/applicant/ApplicantProgramReviewController.java
rockycodes: I have a suggestion further down for how to avoid needing to add the try/catch here.

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -882,7 +883,7 @@ public void getReadOnlyApplicantService_getsReadOnlyApplicantServiceForTheApplic
   }
 
   @Test
-  public void submitApplication_returnsSavedApplication() {
+  public void submitApplication_returnsSavedApplication() throws ProgramNotFoundException {
```
rockycodes, line 886: suggestion (blocking): Please add a test for the `submitApplication` method that tests the three eligibility situations you're handling in `ApplicantService`: eligible, ineligible, and not computed. You might able to handle this by adding some asserts to existing tests.

rockycodes (changes requested):
Looking good! Just a few more relatively minor comments.

server/app/services/applicant/ApplicantService.java
azhangggggg: good idea, done!

server/app/services/applicant/ApplicantService.java
```diff
@@ -429,32 +430,67 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
 
-    return getReadOnlyApplicantProgramService(applicantId, programId)
-        .thenCompose(ro -> validateApplicationForSubmission(ro, programId))
-        .thenCompose(
-            v ->
-                submitApplication(
-                    applicantId, programId, /* tiSubmitterEmail= */ Optional.empty(), request));
+  private EligibilityDetermination calculateEligibilityDetermination(
+      ProgramDefinition programDefinition,
+      ReadOnlyApplicantProgramService reahOnlyApplicantProgramService) {
```
rockycodes, line 484:
```suggestion
      ReadOnlyApplicantProgramService readOnlyApplicantProgramService) {
```

server/app/services/applicant/ApplicantService.java
```diff
@@ -429,32 +430,67 @@ private CompletionStage<ReadOnlyApplicantProgramService> stageAndUpdateIfValid(
...
+    if (programDefinition.hasEligibilityEnabled()) {
+      if (reahOnlyApplicantProgramService.isApplicationNotEligible()) {
```
rockycodes, line 486:
```suggestion
      if (readOnlyApplicantProgramService.isApplicationNotEligible()) {
```

server/test/services/applicant/ApplicantServiceTest.java
```diff
@@ -4301,9 +4388,13 @@ public void getApplicationEligibilityStatus() {
...
             subject.getApplicationEligibilityStatus(ineligibleApplication, programDefinition).get())
         .isFalse();
+    assertThat(ineligibleApplication.getEligibilityDetermination())
+        .isEqualTo(EligibilityDetermination.INELIGIBLE);
     // Re-submission evaluates to eligible.
     assertThat(
             subject.getApplicationEligibilityStatus(eligibleApplication, programDefinition).get())
         .isTrue();
+    assertThat(eligibleApplication.getEligibilityDetermination())
+        .isEqualTo(EligibilityDetermination.ELIGIBLE);
```
rockycodes, line 4398: Thanks for adding these asserts! Would you add one more to a test that submits an application to a program with no eligibility settings to make sure it correctly sets it to `EligibilityDetermination.NO_ELIGIBILITY_CRITERIA`?

avaleske (approved):

Issue #9489, opened 2025-01-03
Add screen-reader-only prefix text to alerts, as a substitute for the icon


melanie-exygy:
#### Describe the solution presented in this issue.
In the North Star, add screen-reader-only text at the beginning of all info and success messages to replace the icon.  

For info messages, add "For your information,".
For success messages, add "Success!".

#### Additional Context
See the discussion in issue #9031. 

#### Done when:
What criteria need to be met for the issue to be finished? 
- [x] All info and success alerts have aria text that describes the type of alert they are.
- [ ] The prefixes are translated.

PR #9771, opened 2025-02-20
Add help text with a prefix visible only to scren readers for INFO an…

watermandrewc:
…d SUCCESS alerts.

### Description

Add extra information at the beginning of SUCCESS and INFO alerts to the text's label to account for the checkmark or info icons for screen readers.

For SUCCESS checkmarks, add "Success" to the beginning of the label. For INFO icons, add "For your information" to the beginning.

Completes code for #9489; messages will need to be translated before the issue is considered complete. This PR affects North Star only.

#### General

#### User visible changes

1. Enable North Star UI.
2. Use a screen reader to read the contents of a success or info alert. It should include the new prefix.


server/app/views/applicant/NavigationFragment.html
```diff
@@ -174,6 +174,7 @@
         <p
           class="usa-alert__text"
           th:utext="#{banner.guestBannerText(${endSessionLinkHtml})}"
+          th:aria-label="#{heading.informationHelpText(#{banner.guestBannerText(${endSessionLinkHtml})})}"
```
melanie-exygy, line 177: **issue (blocking):** When I tested with VoiceOver, I noticed that the screen reader now reads the link html code out loud.  This is only a problem when the link is part of the alert text.  Here's what the aria label looks like for the guest alert:
image
Here's a recording of what that sounds like:

https://github.com/user-attachments/assets/b173232a-b707-4445-8598-720881b2db60

I'm not absolutely sure what to do here, but I'd suggest stripping the html from the aria-label and then testing that with the screen reader.  The link should still be announced, but it should just say "end your session - link".

server/app/views/applicant/NorthStarApplicantUpsellView.java
```diff
@@ -64,8 +65,19 @@ public String render(UpsellParams params) {
...
-        new AlertSettings(/* show= */ true, Optional.of(alertTitle), "", AlertType.SUCCESS);
+        new AlertSettings(
+            /* show= */ true,
+            Optional.of(alertTitle),
+            "",
+            AlertType.SUCCESS,
+            ImmutableList.of(),
+            Optional.empty(),
+            Optional.of(helpText),
+            true);
```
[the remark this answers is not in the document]
watermandrewc: Ah, thanks! I had thought the screenshot change was just noise.

server/app/services/AlertSettings.java
```diff
@@ -24,6 +26,7 @@ public record AlertSettings(
     AlertType alertType,
     ImmutableList<String> additionalText,
     Optional<String> customText,
+    Optional<String> helpText,
```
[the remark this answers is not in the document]
watermandrewc: Updated all the different references in the PR to use ariaLabel instead of "help text" to reduce confusion. Thanks Melanie!

server/app/views/applicant/NavigationFragment.html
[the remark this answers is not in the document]
watermandrewc: Added!

watermandrewc: I updated this to just directly use the message within the HTML (which we set in NorthStarBaseView here: https://github.com/civiform/civiform/blob/82aa79d407642253830e8ca95c8711337dee8342/server/app/views/NorthStarBaseView.java#L106-L112). The screen reader doesn't announce that there's a link within the text unless you step into the component, but that's the same behavior that we have before this PR.

melanie-exygy: So.... I really wish I'd remembered this the first time I reviewed your PR, but here it goes.  

Bad news first:
The `aria-label` attribute isn't supported with the paragraph element.  See [this Mozilla list](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label#associated_roles).  The problem is that, since it's not recognized, the text from the aria-label doesn't replace the text on the `<p>` itself.  So, the screen reader repeats the alert text twice, once with the prefix you added, and once without.  

The good news is that there's an easy way to achieve what we want.  You can make the prefix screen reader only, with this USWDS class `usa-sr-only`.  So, here's an example of what you'd want to do:
```
<section
  th:fragment="guestSessionAlert"
  th:if="${isGuest and hasProfile}"
  role="region"
  th:aria-label="#{label.guestSessionAlert}"
>
  <div
    class="usa-site-alert usa-site-alert--info usa-site-alert--slim cf-alert"
  >
    <div class="usa-alert">
      <div class="usa-alert__body">
        <p class="usa-sr-only">For your information</p>
        <p
          class="usa-alert__text"
          th:utext="#{banner.guestBannerText(${endSessionLinkHtml})}"
        ></p>
      </div>
    </div>
  </div>
</section>
```
Again, so sorry that I'm only realizing this now!

watermandrewc: No worries at all! Thanks for catching it. 

With these changes, the prefix label gets read on its own -- the screen reader still needs to advance to read the rest of the alert banner, which is broken up by the "end your session" text. Is that the behavior we want?

melanie-exygy: Yep, that behavior is okay.  I just tested it on your branch with the screen reader and it seems good.

melanie-exygy (approved):

Issue #9767, opened 2025-02-19
Fix the 10 North Star tests which are failing in probers


melanie-exygy:
#### Describe the solution presented in this issue.
There are 10 NorthStar prober tests which are failing in the civiform-dev deploy pipeline.  They seem to also fail on AWS staging.  Here's a link to the [test failures](https://github.com/civiform/civiform-staging-deploy/actions/runs/13386914918/job/37385627653).

#### Additional Context
I haven't found a common denominator for the failures.  However, at least for one of the tests, I suspect the test is timing out waiting for a locator to resolve through a long-running helper function.  Fetching the locator more succinctly fixes the test in question, which is this one:  "src/applicant/applicant_application_index.test.ts:846:13 › applicant program index page › applicant program index page with northstar UI › program filtering › with program filters enabled, categorizes programs correctly @northstar › Select a filter, click the filter submit button and see the Recommended and Other programs sections ".

#### Done when:
All prober tests pass in the civiform-dev pipeline.

PR #9772, opened 2025-02-21
Getting a North Star test to pass on probers

melanie-exygy:
### Description

An `applicant_application_index.test.ts` test was failing on probers. In different environments, Playwright was having trouble locating the Recommended and Other Programs `<section>` elements, possibly because they're rendered with HTMX.  The solution of using `first()` and `last()` was the only locator combination that worked in both the local/PR environment and in probers.

I added two helper functions to `applicant_program_list.ts` because these were working for probers, but I didn't end up using them because they didn't work locally and in the PR actions.  Leaving them in the PR because I think they could be useful in the future.

### Issue(s) this is part of:

Part of #9767

### DIFF
--- a/browser-test/src/applicant/applicant_application_index.test.ts
+++ b/browser-test/src/applicant/applicant_application_index.test.ts
@@ -16,6 +16,7 @@ import {
 import {Page} from 'playwright'
 import {ProgramVisibility} from '../support/admin_programs'
 import {BASE_URL} from '../support/config'
+import {CardSectionName} from '../support/applicant_program_list'
 
 test.describe('applicant program index page', () => {
   const primaryProgramName = 'Application index primary program'
@@ -847,6 +848,7 @@ test.describe('applicant program index page', () => {
           page,
           adminPrograms,
           applicantQuestions,
+          applicantProgramList,
         }) => {
           await test.step('publish programs with categories', async () => {
             await adminPrograms.publishAllDrafts()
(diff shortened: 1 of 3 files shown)


browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -925,14 +925,16 @@ test.describe('applicant program index page', () => {
             )
           })
 
-          await test.step('Select a filter, click the filter submit button and see the Recommended and Other programs sections', async () => {
+          await test.step('Select a filter, click the filter submit button and validate screenshot', async () => {
```
melanie-exygy, line 928: **note:** In the end, the only change to this file is splitting the one large test step into two.  I did that while I was working in there so that I could more easily pinpoint the problem.

browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -925,37 +925,37 @@ test.describe('applicant program index page', () => {
...
             await applicantQuestions.filterProgramsByCategory('General')
```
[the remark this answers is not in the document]
melanie-exygy: Hurray, that fixed it!  Thanks for the great idea.  For some reason, even with the wait, it didn't recognize the region role on the recommended and other programs sections when I tried using the new methods I'd created in `applicant_program_list.ts`, so I just returned this to how it was originally.

browser-test/src/support/applicant_program_list.ts
melanie-exygy, line 1: **question:** @gwendolyngoetz Do you think I should keep these new functions even though I'm not using them?  I thought they might come in handy at some point, but am just not sure if I should be adding unused functions.

gwendolyngoetz: I'd say let's remove the unused methods. They are pretty easy to re-add when needed.

gwendolyngoetz (commented):
I'd say let's remove the unused methods. They are pretty easy to re-add when needed.

browser-test/src/support/applicant_program_list.ts
melanie-exygy: Yeah, I'm thinking that's best too.  Removed them.

gwendolyngoetz (approved):

PR #9779, opened 2025-02-24
IneligibleBlockView - Formatting

shane-exygy:
### Description

Fix line breaks and comment syntax checking


dkatzz (approved):
Thanks for cleaning that up!!

PR #9780, opened 2025-02-24
[North Star] bug fix -- maintain "submitted" tag on applications

### LINKED ISSUE #9389: Application showing "not yet submitted" tag after program submission
1. Clicking edit on a program after you've submitted takes you to a program page, instead of the review page. 
2. Then, when you try to click out of the form using the "submit" button or "review and exit" (which takes you to the submit screen) you get a pop-up saying that nothing has been changed so I can't submit an application with no change. 
3. Even if I just click back into the homescreen without acknowledging the modals, I still see "not yet submitted." 
4. When I go back to the homescreen, the application now says "not yet submitted".

Instead, I should be taken to the review page if clicking "edit" after submission. Even if there are no edits made, I should still see that it has been submitted. 

https://github.com/user-attachments/assets/425ada81-ba8d-40b5-a071-366de62eba93

rockycodes:
### Description

This PR fixes a bug where applications were being marked as "not yet submitted" after submission when a new draft was created in our North Star UI. This updates the North Star logic to match [the logic used in our legacy UI](https://github.com/civiform/civiform/blob/7e2fb646bf7e5ab6e685dd2589758eda89a60866/server/app/views/applicant/ProgramCardViewRenderer.java#L245) which maintained the "submitted" tag if an application had previously been submitted -- even if a new draft had been created.

#### General

1. As an admin create a program with at least one question
2. As an applicant apply to the program and submit your application
3. As the same applicant, go back to the homepage and see the "submitted" tag
4. As the same applicant, reapply to the same program and edit your answer (you don't have to edit your answer, but you do need to click "Continue" on the question page because that creates a new draft)
5. Do not submit your application
6. As the same applicant, go back to the homepage and see that the "submitted" tag is still here

### Issue(s) this completes

Fixes #9389


browser-test/src/applicant/applicant_application_index.test.ts
```diff
@@ -698,7 +698,7 @@ test.describe('applicant program index page', () => {
             page,
             'program-index-page-submitted-northstar',
           )
-          await expect(page.getByText('Submitted on 1/1/30')).toBeVisible()
+          await expect(page.getByText('Submitted on')).toBeVisible()
```
rockycodes, line 701: The date is actually written as whatever date you run the test, but there seems to be a funny bug where the snapshot test updates the date to be `1/1/30`. This is confusing and I would rather not rely on it, so I took the date out.

server/app/views/applicant/ProgramCardsSectionFragment.html
```diff
@@ -88,7 +88,7 @@ <h3 th:id="${section.id().get()}" th:text="${section.title().get()}"></h3>
 
       <!-- Submitted tag -->
       <div
-        th:if="${!card.applicationStatus().isPresent() and card.lifecycleStage.isPresent() and card.lifecycleStage.get().toString() == 'ACTIVE'}"
+        th:if="${!card.applicationStatus().isPresent() and card.dateSubmitted.isPresent()}"
```
rockycodes, line 91: The lifecycle stage doesn't matter, we only care if it has previously been submitted.

watermandrewc (approved):

browser-test/src/applicant/applicant_application_index.test.ts
melanie-exygy: We can rely on the "1/1/30" date because of `normalizeElements` [here](https://github.com/civiform/civiform/blob/f385921d34947354579f3205e764e7df05ba74ab/browser-test/src/support/index.ts#L329).  It's thrown me off before, though, so I understand if you want to leave it out.

server/app/views/applicant/ProgramCardsSectionFragment.html
melanie-exygy: Great job finding this!

melanie-exygy (approved):

browser-test/src/applicant/applicant_application_index.test.ts
rockycodes: Thank you! I'll still take it out since that's not obvious in this test. Good to know though.

rockycodes: Ah I see you just merged a change with an explicit call to `normalizeElements`! I'll leave it in and go that route then.

watermandrewc (approved):

PR #9789, opened 2025-02-26
Refactoring to prompt versioning

sanjanasandeep:
### Description

Adding a new folder to maintain the prompt versions. The suffix is the version number

### DIFF
--- a/pdf-to-civiform/pdf-to-civiform-gemini.py
+++ b/pdf-to-civiform/pdf-to-civiform-gemini.py
@@ -162,7 +162,7 @@ def process_pdf_text_with_llm(model, model_name, text, base_name):
     
     Output only JSON, no explanations.
     """
-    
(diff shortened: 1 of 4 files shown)


pdf-to-civiform/pdf-to-civiform-gemini.py
```diff
@@ -5,7 +5,8 @@
 from werkzeug.utils import secure_filename
 import os
 import logging
-from convert_to_civiform_json import convert_to_civiform_json 
+from convert_to_civiform_json import convert_to_civiform_json
+from prompts import pdf_extract_v1, post_process_v1, format_json_v1
```
haltmar, line 9: I suggest using a folder structure to hold versions of the prompt, rather than by file name and variable names.
It would be easier and more flexible to test/iterate over different LLM models and version.

PR #9803, opened 2025-02-26
[SessionTimeout][part1]Introduce the session timeout configurations and feature flag

### LINKED ISSUE #9820: Implement Server Settings to configure session inactivity and max session length timeout and warnings
See #843
The new server settings are:
session_timeout_enabled: Session timeout feature enabled or not.

warnings:
session_inactivity_warning_minutes: The number of minutes of inactivity before we warn the user that their session will expire.
session_max_duration_warning_minutes: The number of minutes before we hit the maximum session duration that we warn the user that their session will expire.

timeouts:
session_inactivity_warning_minutes: The number of minutes of inactivity before a session times out.

codekarma:
### Description

Introduce configurations for browser session timeouts and warning threshold. This is part 1 of browser session timeout work. The feature is by default turned off by default. Since the feature can be configured as on or off, I have not created an explicit feature flag. All subsequent changes will be guarded on the feature being on or off effectively serving as a feature flag.

#### General

#### New Features

Login as admin and ensure that all the new server settings appears in the intended sections.

### Issue(s) this completes

Fixes #9820

### DIFF
--- a/server/app/services/settings/SettingsManifest.java
+++ b/server/app/services/settings/SettingsManifest.java
@@ -822,11 +822,6 @@ public final class SettingsManifest extends AbstractSettingsManifest {
     return getInt("MAXIMUM_SESSION_DURATION_MINUTES");
   }
 
-  /** Enable session timeout based on inactivity and maximum duration. */
-  public boolean getSessionTimeoutEnabled() {
-    return getBool("SESSION_TIMEOUT_ENABLED");
-  }
-
   /**
    * The number of minutes of inactivity before we warn the user that their session will expire.
    * Default is 5.
@@ -1072,6 +1067,11 @@ public final class SettingsManifest extends AbstractSettingsManifest {
    */
   public boolean getNorthStarApplicantUi(RequestHeader request) {
     return getBool("NORTH_STAR_APPLICANT_UI", request);
+  }
+
+  /** Enable session timeout based on inactivity and maximum duration. */
+  public boolean getSessionTimeoutEnabled() {
+    return getBool("SESSION_TIMEOUT_ENABLED");
   }
 
   private static final ImmutableMap<String, SettingsSection> GENERATED_SECTIONS =
@@ -2200,7 +2200,13 @@ public final class SettingsManifest extends AbstractSettingsManifest {
                           + " experience in Applicant flows",
                       /* isRequired= */ false,
                       SettingType.BOOLEAN,
-                      SettingMode.ADMIN_WRITEABLE))),
+                      SettingMode.ADMIN_WRITEABLE),
+                  SettingDescription.create(
+                      "SESSION_TIMEOUT_ENABLED",
+                      "Enable session timeout based on inactivity and maximum duration.",
+                      /* isRequired= */ false,
+                      SettingType.BOOLEAN,
+                      SettingMode.ADMIN_READABLE))),
           "Miscellaneous",
           SettingsSection.create(
               "Miscellaneous",
@@ -2323,12 +2329,6 @@ public final class SettingsManifest extends AbstractSettingsManifest {
                           + " frontend to notify a user when their session is expired.",
                       /* isRequired= */ false,
                       SettingType.INT,
-                      SettingMode.ADMIN_READABLE),
-                  SettingDescription.create(
-                      "SESSION_TIMEOUT_ENABLED",
-                      "Enable session timeout based on inactivity and maximum duration.",
-                      /* isRequired= */ false,
-                      SettingType.BOOLEAN,
                       SettingMode.ADMIN_READABLE),
                   SettingDescription.create(
                       "SESSION_INACTIVITY_WARNING_MINUTES",
--- a/server/app/services/settings/SettingsManifest.java
+++ b/server/app/services/settings/SettingsManifest.java
@@ -822,6 +822,32 @@ public final class SettingsManifest extends AbstractSettingsManifest {
     return getInt("MAXIMUM_SESSION_DURATION_MINUTES");
   }
 
+  /** Enable session timeout based on inactivity and maximum duration. */
+  public boolean getSessionTimeoutEnabled() {
+    return getBool("SESSION_TIMEOUT_ENABLED");
+  }
+
+  /**
+   * The number of minutes of inactivity before we warn the user that their session will expire.
+   * Default is 5.
+   */
+  public Optional<Integer> getSessionInactivityWarningMinutes() {
+    return getInt("SESSION_INACTIVITY_WARNING_MINUTES");
+  }
+
+  /**
+   * The number of minutes before we hit the maximum session duration that we warn the user that
+   * their session will expire. Default is 10.
+   */
+  public Optional<Integer> getSessionMaxDurationWarningMinutes() {
+    return getInt("SESSION_MAX_DURATION_WARNING_MINUTES");
+  }
+
+  /** The number of minutes of inactivity before a session times out. Default is 30. */
+  public Optional<Integer> getSessionInactivityTimeoutMinutes() {
+    return getInt("SESSION_INACTIVITY_TIMEOUT_MINUTES");
+  }
+
   /**
    * If enabled, allows server Prometheus metrics to be retrieved via the '/metrics' URL path.  If
    * disabled, '/metrics' returns a 404.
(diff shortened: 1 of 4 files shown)


server/app/services/settings/SettingsManifest.java
```diff
@@ -2297,5 +2323,32 @@ public boolean getNorthStarApplicantUi(RequestHeader request) {
...
+                      SettingMode.ADMIN_READABLE),
+                  SettingDescription.create(
+                      "SESSION_MAX_DURATION_WARNING_MINUTES",
+                      "The number of minutes before we hit the maximum session duration that we"
+                          + " warn the user that their session will expire. Default is 10.",
+                      /* isRequired= */ false,
+                      SettingType.INT,
+                      SettingMode.ADMIN_READABLE),
+                  SettingDescription.create(
+                      "SESSION_INACTIVITY_TIMEOUT_MINUTES",
```
dkatzz, line 2348: **suggestion:** There is already MAXIMUM_SESSION_DURATION_MINUTES that controls how long before the session expires, which I believe would be the same as this one. We should merge these into one variable?

server/conf/env-var-docs.json
```diff
@@ -703,6 +703,30 @@
     "description": "The amount of time, in minutes, that a session lasts. The default is 600 minutes, or 10 hours. Note that there isn't yet messaging on the frontend to notify a user when their session is expired.",
     "type": "int"
   },
+  "SESSION_TIMEOUT_ENABLED": {
+    "mode": "ADMIN_READABLE",
+    "description": "Enable session timeout based on inactivity and maximum duration.",
```
dkatzz, line 708: **nitpick:** Optional but maybe we should make a note of what the feature flag is to utilize these?

server/app/services/settings/SettingsManifest.java
codekarma: These are two different configuration settings:
MAXIMUM_SESSION_DURATION_MINUTES: Governs total length of session. A session cannot last longer than this and we force user to re-authenticate again. This depends on the session creation time.
SESSION_INACTIVITY_TIMEOUT_MINUTES: Governs the timeout due to inactivity. This depends on lastActivityTime that gets reset after any request on server side.

server/conf/env-var-docs.json
codekarma: This guards the session timeout feature but the reason it is not feature flag is because this setting can be turned off or on even after the feature is stabilized. I have not added any new feature flag. Can we have server settings also behind feature flag?

server/app/services/settings/SettingsManifest.java
dkatzz: Ah yes makes sense! Thank you for clarifying.

server/conf/env-var-docs.json
dkatzz: Got it! Can we move it into experimental still under feature flags and then we can relocate it here once the work is complete? I wouldn't want anyone turning it on thinking its ready in the meantime, so right now it would function as a feature flag and later it would be a setting that can be turned on / off.

server/conf/application.conf
```diff
@@ -611,3 +611,16 @@ file_upload_allowed_file_type_specifiers = ${?FILE_UPLOAD_ALLOWED_FILE_TYPE_SPEC
...
+session_timeout_enabled = false
+session_timeout_enabled = ${?SESSION_TIMEOUT_ENABLED}
+
+session_inactivity_timeout_minutes = 30
+session_inactivity_timeout_minutes = ${?SESSION_INACTIVITY_TIMEOUT_MINUTES}
+
+session_inactivity_warning_minutes = 5  # 5 minutes warning for inactivity
+session_inactivity_warning_minutes = ${?SESSION_INACTIVITY_WARNING_MINUTES}
+
+session_max_duration_warning_minutes = 10     # 10 minutes warning for total duration
```
rockycodes, line 625: question: Does this warn the user 10 mins _before_ the total duration? If so, can you make that clear here?

server/conf/env-var-docs.json
```diff
@@ -703,6 +703,24 @@
     "description": "The amount of time, in minutes, that a session lasts. The default is 600 minutes, or 10 hours. Note that there isn't yet messaging on the frontend to notify a user when their session is expired.",
     "type": "int"
   },
+  "SESSION_INACTIVITY_WARNING_MINUTES": {
+    "mode": "ADMIN_READABLE",
+    "description": "The number of minutes of inactivity before we warn the user that their session will expire. Default is 5.",
```
rockycodes, line 708: question: Does this mean we will warn the user after 5 minutes of inactivity or 5 minutes before their session expires due to inactivity? The language is a little unclear.

rockycodes (changes requested):
It's exciting to see this work getting started! I have a couple questions/thoughts about making the language more clear to differentiate between if the variable is used in reference to time elapsed or time _before_ the session times out, but overall this lgtm.

server/conf/application.conf
gwendolyngoetz: (question) How does `session_inactivity_timeout_minutes` differ from `maximum_session_duration_minutes` which is already defined?

rockycodes: Danielle asked the same thing and amit answered it here https://github.com/civiform/civiform/pull/9803#discussion_r1973747373

This is making me think we might need something more in the description to make that clearer.

server/conf/env-var-docs.json
codekarma: How about this:
"How long before the inactivity timeout, in minutes, a user receives a warning that their session is about to expire."

server/conf/application.conf
gwendolyngoetz: I think we should try to unify the naming on all of these. fyi @dkatzz 

- maximum_session_duration_minutes
- session_timeout_enabled
- session_inactivity_timeout_minutes
- session_inactivity_warning_minutes
- session_max_duration_warning_minutes


or at least make `session_max_duration_warning_minutes` -> `session_duration_warning_threshold_minutes`  or something like that. 

It's hard to follow since things are so similarly named.

codekarma: Yes. that is correct. I updated the comment.

server/conf/env-var-docs.json
codekarma: updated the wording.

server/conf/application.conf
codekarma: The pattern I following is 
For warning: session_<reason>_warning_minutes
For timeout: session_<reason>_timeout_minutes
the reason can be either inactivity or max_duration.
@gwendolyngoetz Love to get your thoughts on how to unify this.

codekarma: How about this?
session_timeout_enabled
maximum_session_duration_minutes(existing)
session_inactivity_duration_minutes
session_duration_warning_threshold_minutes
session_inactivity_warning_threshold_minutes

codekarma: Updated the naming to this:
```
session_timeout_enabled
maximum_session_duration_minutes(existing)
session_inactivity_timeout_minutes
session_duration_warning_threshold_minutes
session_inactivity_warning_threshold_minutes
```

server/conf/env-var-docs.json
rockycodes: This is much clearer. Thank you!

server/conf/application.conf
rockycodes: This is much more clear to me. I'm ready to approve but want to hear if this works for @gwendolyngoetz and @dkatzz first.

dkatzz: Agreed this is much more clear!

My only thought is whether we should change 'session_inactivity_timeout_minutes' to 'maximum_session_inactivity_timeout_minutes' for consistency with the existing on or update 'maximum_session_duration_minutes' to 'session_duration_timeout_minutes' but would need to ensure nobody is setting that so we don't cause issues with renaming the variable. If we renamed that would be a separate PR. 

Curious what others think!

codekarma: I think we `maximum_session_duration_minutes` to `session_duration_timeout_minutes` makes more sense to me. I can make that change in this PR or we can change it later.
 
The problem with `maximum_session_inactivity_timeout_minutes` is that prefix `maximum` is confusing. It is not maximum of the session inactivity timeout but rather It is just a timeout.

dkatzz: Sounds good! I'm good with this but we'll just want to check nobody is setting the other flag before we make the other change

rockycodes: that makes sense to me!

rockycodes (approved):

server/conf/application.conf
gwendolyngoetz: Yeah looks good. Thanks!

PR #9806, opened 2025-02-26
[SessionTimeout][part2]Enforce session length and inactivity timeout 

### LINKED ISSUE #9821: Enforce max session and inactivity timeout
Currently `ValidAccountFilter` validate that every request has a valid account and session. This task is to augment this filter to also check if the user session has breached either session max duration or inactivity timeout and log user out if they did.

codekarma:
### Description

This change augment the existing CiviformUserProfileData to also store last activity time. ValidAccountFilter now checks the length of user session and the inactivity time and logs user out if they exceed the configured limits.

#### General

#### New Features

### Issue(s) this completes

Fixes #9821


server/test/filters/ValidAccountFilterTest.java
```diff
@@ -0,0 +1,228 @@
...
+    when(mockProfile.getProfileData()).thenReturn(mockProfileData);
+    when(mockProfile.getAccount()).thenReturn(CompletableFuture.completedFuture(mockAccount));
+
+    play.i18n.Messages mockMessages = mock(play.i18n.Messages.class);
+    when(messagesApi.preferred(any(Http.RequestHeader.class))).thenReturn(mockMessages);
+    when(mockMessages.at(anyString())).thenReturn("Logout message");
+  }
+
+  @Test
+  public void testValidProfile_UpdatesLastActivityTime() throws Exception {
```
dkatzz, line 58: **suggestion:** I think it may be helpful to include the flag value in the test name for clarity `testValidProfile_sessionTimeoutDisabled_updatesLastActivityTime` and similar for below

server/test/filters/ValidAccountFilterTest.java
dkatzz, line 1: **praise:** Thank you for adding this test file!!

server/app/auth/CiviFormProfile.java
```diff
@@ -226,6 +228,17 @@ public CompletableFuture<String> getEmailAddress() {
...
 
+  public Optional<Long> getSessionStartTime() {
+    return getAccount()
+        .thenApply(
+            account ->
+                account
+                    .getActiveSession(getProfileData().getSessionId())
+                    .map(SessionDetails::getCreationTime)
+                    .map(Instant::toEpochMilli))
+        .join();
```
dkatzz, line 239: **thought:** We've tried to avoid calling `.join` in the applicant path where possible - maybe there isn't a way to avoid it here, but thought I'd call that out

server/app/auth/CiviFormProfileData.java
```diff
@@ -41,8 +43,13 @@ public CiviFormProfileData() {
   }
 
   public CiviFormProfileData(Long accountId) {
+    this(accountId, Clock.systemUTC());
```
gwendolyngoetz, line 46: (blocking) I don't like how we're using mixed version of the clock. Internal to this class clock is being used as method parameters, a constructor parameter, use of static Clock. This class isn't being created by Guice so we can't just `@Inject` it, but there needs to be one entry point of the clock into this class.

The Clock typically provided when injecting it into the constructor used the timezone that is configured by CiviForm admins. Having each method accept the clock may result in mismatched clocks being used when setting or comparing times.

Please adjust this so that there is only a single clock that is used in this class.

gwendolyngoetz (changes requested):

server/app/auth/CiviFormProfileData.java
```diff
@@ -41,8 +43,13 @@ public CiviFormProfileData() {
   }
 
   public CiviFormProfileData(Long accountId) {
```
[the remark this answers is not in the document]
codekarma: Just in tests. I am updating all the test and removing this method to address Gwen's comment.

server/app/filters/ValidAccountFilter.java
```diff
@@ -4,53 +4,137 @@
...
+  private Optional<LogoutReason> getLogoutReason(Optional<CiviFormProfile> maybeProfile) {
+    if (maybeProfile.isEmpty()) {
+      return Optional.empty();
+    }
+
+    CiviFormProfile profile = maybeProfile.get();
+
+    // Check basic validity first
+    if (!profileUtils.validCiviFormProfile(profile)) {
+      return Optional.of(LogoutReason.INVALID_PROFILE);
```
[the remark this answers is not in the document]
codekarma: We checked for profile and session validity before without showing user any message. I consolidated the logic into this getLogoutReason function where we now have an opportunity to present user with message. We can chose to keep the old behavior if that is desirable.

server/conf/i18n/messages
```diff
@@ -42,6 +42,12 @@ heading.informationAriaLabelPrefix=For your information: {0}
 heading.successAriaLabelPrefix=Success: {0}
 # Toast message that tells the user their session has ended, to help indicate that they are no longer logged in to an account (but still as a guest).
 toast.sessionEnded=Your session has ended.
+# Toast message that tells the user that they are logged out because their session is invalid.
+toast.logoutInvalidSession=Your session is no longer valid. Please log in again.
+# Toast message that tells the user that they are logged out because their session has timed out due to inactivity.
+toast.logoutTimeoutInactivity=Your session has timed out due to inactivity.
+# Toast message that tells the user that they are logged out because their session has expired.
+toast.logoutTimeoutDuration=Your session has expired. Please log in again.
```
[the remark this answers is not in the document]
codekarma: Fair point. I can simplify the logic quite a lot of we don't think showing specific message adds any value. I was just following TDD that calls for specific messages detailing the reason for the logout.

server/app/filters/ValidAccountFilter.java
```diff
@@ -4,53 +4,137 @@
...
+  private static final int DEFAULT_MAX_SESSION_DURATION_MINUTES = 480;
+
   private final ProfileUtils profileUtils;
-  private final Config config;
+  private final Provider<SettingsManifest> settingsManifest;
+  private final MessagesApi messagesApi;
+  private final Clock clock;
+
+  public enum LogoutReason {
+    INVALID_PROFILE(MessageKey.TOAST_LOGOUT_INVALID_SESSION.getKeyName()),
```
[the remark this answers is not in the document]
codekarma: Yes because there would be the same message displayed for both of these logout reasons. As it was pointed out in other PR comment, these messages may not be that useful to users so it make sense to simplify the code and remove the messaging.

server/app/auth/CiviFormProfile.java
codekarma: Not sure how to avoid join and still be able to read the session creation time. I'm open to any ideas/

server/app/filters/ValidAccountFilter.java
```diff
@@ -4,53 +4,137 @@
...
+import services.settings.SettingsManifest;
 
 /**
  * A filter to ensure the account referenced in the browser cookie is valid. This should only matter
  * when the account is deleted from the database which almost will never happen in prod database.
  */
 public class ValidAccountFilter extends EssentialFilter {
+
+  private static final int DEFAULT_INACTIVITY_TIMEOUT_MINUTES = 30;
+  private static final int DEFAULT_MAX_SESSION_DURATION_MINUTES = 480;
```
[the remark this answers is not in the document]
codekarma: This is the case when application.conf is missing these properties. If they are, we are providing a default which is also noted in the SettingsManifest.

server/test/filters/ValidAccountFilterTest.java
```diff
@@ -0,0 +1,228 @@
...
+    when(settingsManifest.getSessionReplayProtectionEnabled()).thenReturn(false);
+
+    EssentialAction action =
+        filter.apply(
+            EssentialAction.of(requestHeader -> Accumulator.done(play.mvc.Results.ok("Success"))));
+
+    Result result =
+        action.apply(request.build()).run(NoMaterializer$.MODULE$).toCompletableFuture().get();
+
+    verify(mockProfileData).updateLastActivityTime(any());
```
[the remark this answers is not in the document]
codekarma: Yes. Added a new test to check that updateLastActivityTime is not updated of the feature is off.

server/app/auth/CiviFormProfileData.java
codekarma: This constructor was only used for tests. I have removed this constructor and now we only use the clock passed to the profile data.

server/conf/i18n/messages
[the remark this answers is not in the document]
codekarma: I've removed the logout message as we did not show any message on logout due to invalid session or invalid profile. We also have the frontend that warns user of the session timeout.

server/app/auth/CiviFormProfile.java
gwendolyngoetz: Agreed. This should return a CompletableFuture and the filter should be using the appropriate methods to get the value in an async way.

https://www.playframework.com/documentation/3.0.x/JavaHttpFilters

https://www.playframework.com/documentation/3.0.x/JavaAsync

server/app/filters/ValidAccountFilter.java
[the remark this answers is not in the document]
codekarma: I've updated the hardcoded value in filter to match the default config value. It looks like we are using the hardcoded default value for maximumSessionDurationMinutes in the code(See AccountRepository where we have a default value 600).

server/app/auth/CiviFormProfileData.java
```diff
@@ -70,6 +73,20 @@ public String getSessionId() {
     return getAttributeAsString(SESSION_ID);
   }
 
+  public void updateLastActivityTime(Clock clock) {
+    addAttribute(LAST_ACTIVITY_TIME, clock.instant().toEpochMilli());
+  }
```
gwendolyngoetz, line 78:
```suggestion
  public void updateLastActivityTime() {
    addAttribute(LAST_ACTIVITY_TIME, clock.instant().toEpochMilli());
  }
```

gwendolyngoetz: (blocking) Please use the class level clock.

server/app/auth/CiviFormProfileData.java
```diff
@@ -70,6 +73,20 @@ public String getSessionId() {
...
+
+  /**
+   * Returns the timestamp of the last activity for this session. For backward compatibility with
+   * existing sessions created before session timeout feature, returns current time if the attribute
+   * is not present. This effectively gives legacy sessions a fresh activity timestamp when the
+   * timeout feature is first enabled.
+   */
+  public long getLastActivityTime(Clock clock) {
+    return (Long) getAttributes().getOrDefault(LAST_ACTIVITY_TIME, clock.instant().toEpochMilli());
+  }
```
gwendolyngoetz, line 88:
```suggestion
  public long getLastActivityTime() {
    return (Long) getAttributes().getOrDefault(LAST_ACTIVITY_TIME, clock.instant().toEpochMilli());
  }
```

(blocking) Please use the class level clock.

server/app/auth/CiviFormProfile.java
```diff
@@ -226,6 +228,22 @@ public CompletableFuture<String> getEmailAddress() {
         .thenApplyAsync(AccountModel::getEmailAddress, classLoaderExecutionContext.current());
   }
 
+  /**
+   * Gets the session start time asynchronously.
+   *
+   * @return A CompletableFuture that resolves to an Optional containing the session start time in
+   *     milliseconds, or empty if no active session is found.
+   */
+  public CompletableFuture<Optional<Long>> getSessionStartTime() {
```
gwendolyngoetz, line 237: (praise) Fantastic!! :D

gwendolyngoetz (changes requested):
Great job! Just a few other changes I'd like to see.

server/app/auth/CiviFormProfileData.java
```diff
@@ -40,9 +42,10 @@ public CiviFormProfileData() {
     addAttribute(SESSION_ID, UUID.randomUUID().toString());
   }
 
-  public CiviFormProfileData(Long accountId) {
+  public CiviFormProfileData(Long accountId, Clock clock) {
     this();
     this.setId(accountId.toString());
+    addAttribute(LAST_ACTIVITY_TIME, clock.instant().toEpochMilli());
```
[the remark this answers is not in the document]
codekarma: Since this class is serializable, it's must have a no argument constructor which will not have access to clock provided by the IoC framework. This is the reason I did not keep the class level Clock. Do you have a suggestion on how to get around that problem?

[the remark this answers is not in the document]
codekarma: Also if I keep the Class level Clock object, even with transient, wouldn't we have backward compatibility issues and may have to bump the version(I vaguely remember making this change and it failed test with serialization issues)? 
I think that this current design keep the class simpler considering the Clock initialization issue during deserialization that would need ugly null checks.

dkatzz (approved):

gwendolyngoetz (approved):

PR #9810, opened 2025-02-26
[SessionTimeout][part3]Extend session endpoint and a new filter to add timeout cookie

codekarma:
### Description

This change is a backend for session length and inactivity timeout warning functionality. A new endpoint is added that updates the last activity time on user profile effectively granting user more time to finish any workflow they have been working on. There is also a hard limit on total session length and we warn user some time[configurable] before their session length approaches this limit. Frontend for this change is going to be in a subsequent PR.

#### General

#### New Features

Only way to verify this change is to check the session cookie in chrome devtools.

### Issue(s) this completes

Fixes [9823](https://github.com/civiform/civiform/issues/9823)


server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.pekko.stream.Materializer;
+import play.libs.Json;
+import play.mvc.Filter;
+import play.mvc.Http;
+import play.mvc.Result;
+import services.settings.SettingsManifest;
+
+public class SessionTimeoutFilter extends Filter {
+  public static final int DEFAULT_INACTIVITY_TIMEOUT_MINUTES = 30;
+  public static final int DEFAULT_MAX_SESSION_DURATION_MINUTES = 600;
```
gwendolyngoetz, line 29:
```suggestion

```
(blocking) This came up in the previous PR, and I'm going to bring it up again since these are starting to get used elsewhere now.

Just delete these.

Instead of  `settingsManifest.get().getMaximumSessionDurationMinutes().orElse(DEFAULT_MAX_SESSION_DURATION_MINUTES)`

Do `settingsManifest.get().getMaximumSessionDurationMinutes().orElseThrow()`

The default value is already in the config and this is just adding unnecessary code.

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+    this.settingsManifest = checkNotNull(settingsManifest);
+    this.clock = checkNotNull(clock);
+  }
+
+  @Override
+  public CompletionStage<Result> apply(
+      Function<Http.RequestHeader, CompletionStage<Result>> nextFilter,
+      Http.RequestHeader requestHeader) {
+
+    Optional<CiviFormProfile> profileOpt = profileUtils.optionalCurrentUserProfile(requestHeader);
```
gwendolyngoetz, line 55:
```suggestion
    Optional<CiviFormProfile> optionalProfile = profileUtils.optionalCurrentUserProfile(requestHeader);
```
(blocking) You can find the [preferred naming of optionals](https://github.com/civiform/civiform/wiki/Development-standards#prefer-optional-instead-of-null) here in the wiki.

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+import javax.inject.Inject;
+import javax.inject.Provider;
+import org.apache.pekko.stream.Materializer;
+import play.libs.Json;
+import play.mvc.Filter;
+import play.mvc.Http;
+import play.mvc.Result;
+import services.settings.SettingsManifest;
+
+public class SessionTimeoutFilter extends Filter {
```
gwendolyngoetz, line 27: (blocking) Please add javadoc

server/app/controllers/SessionController.java
```diff
@@ -0,0 +1,38 @@
...
+
+import auth.ProfileUtils;
+import java.time.Clock;
+import javax.inject.Inject;
+import play.mvc.Controller;
+import play.mvc.Http;
+import play.mvc.Result;
+import services.settings.SettingsManifest;
+
+public class SessionController extends Controller {
```
gwendolyngoetz, line 11: (blocking) Please add javadoc

server/app/controllers/SessionController.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Inject
+  public SessionController(
+      ProfileUtils profileUtils, SettingsManifest settingsManifest, Clock clock) {
+    this.profileUtils = profileUtils;
+    this.settingsManifest = settingsManifest;
+    this.clock = clock;
+  }
+
+  public Result extendSession(Http.Request request) {
```
gwendolyngoetz, line 24: (blocking) Please add javadoc

server/app/controllers/SessionController.java
```diff
@@ -0,0 +1,38 @@
...
+    if (!settingsManifest.getSessionTimeoutEnabled()) {
+      return badRequest("Session timeout is not enabled");
```
gwendolyngoetz, line 26:
```suggestion
      return badRequest();
```
(blocking) Please remove the message. We don't need to expose the reason to the client.

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  private CompletableFuture<Http.Cookie> createSessionTimestampCookie(CiviFormProfile profile) {
+    return profile
+        .getSessionStartTime()
+        .thenApply(
+            maybeSessionStartTime -> {
+              // Get timestamps in seconds
+              long currentTime = clock.instant().toEpochMilli() / 1000;
+              long sessionStartTimeInMillis = maybeSessionStartTime.orElse(currentTime);
```
dkatzz, line 73: **nitpick:** Name is confusing because it says "InMillis", but I believe its in seconds? Or at least currentTime is.

server/app/controllers/SessionController.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    return profileUtils
+        .optionalCurrentUserProfile(request)
+        .map(
+            profile -> {
+              profile.getProfileData().updateLastActivityTime(clock);
+              return ok("Session extended");
+            })
+        .orElse(unauthorized("No active session"));
```
gwendolyngoetz, line 36:
```suggestion
              return ok();
            })
        .orElse(unauthorized());
```
(blocking) Same here

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+              long lastActivityTime = profile.getProfileData().getLastActivityTime(clock) / 1000;
+              long sessionStartTime = sessionStartTimeInMillis / 1000;
```
dkatzz, line 75: **nitpick:** Maybe put this above next to the sessionStartTimeInMillis definition or remove that one and define this as `maybeSessionStartTime.orElse(currentTime) / 1000`

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+                createSessionTimestampCookie(profileOpt.get()).thenApply(result::withCookies));
+  }
+
+  private CompletableFuture<Http.Cookie> createSessionTimestampCookie(CiviFormProfile profile) {
+    return profile
+        .getSessionStartTime()
+        .thenApply(
+            maybeSessionStartTime -> {
+              // Get timestamps in seconds
+              long currentTime = clock.instant().toEpochMilli() / 1000;
```
gwendolyngoetz, line 72: (blocking) I'm seeing with the inclusion of these additional filters the logic to do all the calculations is starting to spread out. Please create a new service class and centralize all of the time/threshold calculations that are in all the new session related filters into that new class.

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+                      .get()
+                      .getSessionInactivityTimeoutMinutes()
+                      .orElse(DEFAULT_INACTIVITY_TIMEOUT_MINUTES);
+              int totalLengthMinutes =
+                  settingsManifest
+                      .get()
+                      .getMaximumSessionDurationMinutes()
+                      .orElse(DEFAULT_MAX_SESSION_DURATION_MINUTES);
+              int inactivityWarningMinutes =
+                  settingsManifest.get().getSessionInactivityWarningThresholdMinutes().orElse(5);
```
dkatzz, line 89: **question:** Should we handle these similar to DEFAULT_INACTIVITY_TIMEOUT_MINUTES and DEFAULT_MAX_SESSION_DURATION_MINUTES and define this above?

server/test/filters/SessionTimeoutFilterTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    when(mockProfile.getProfileData()).thenReturn(mockProfileData);
+
+    filter =
+        new SessionTimeoutFilter(
+            materializer, profileUtils, () -> settingsManifest, instanceOf(Clock.class));
+  }
+
+  @Test
+  public void testNoProfile_ClearsCookie() throws Exception {
```
dkatzz, line 56: **nitpick:** Could we camel case the test name: testNoProfile_clearsCookie here and below

server/app/controllers/SessionController.java
dkatzz, line 26: **question:** I'm assuming this wouldn't be called if timeout isn't enabled, so we should never reach this, right?

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,141 @@
...
+              String cookieValue =
+                  Base64.getEncoder()
+                      .encodeToString(Json.stringify(timestamps).getBytes(StandardCharsets.UTF_8));
+
+              // Set Cookie with appropriate max age
+              return Http.Cookie.builder(TIMEOUT_COOKIE_NAME, cookieValue)
+                  .withHttpOnly(false)
+                  .withPath("/")
+                  .withMaxAge(COOKIE_MAX_AGE)
+                  .build();
```
gwendolyngoetz, line 130: (question) Will you be using javascript to read the values of the cookie on the client side?

gwendolyngoetz (changes requested):

server/app/filters/SessionTimeoutFilter.java
gwendolyngoetz: It's probably a good idea to rename all the time variables to include the units to avoid confusion.

server/app/filters/SessionTimeoutFilter.java
codekarma: yes. That is correct. See session.ts here: https://github.com/civiform/civiform/pull/9776

server/app/filters/SessionTimeoutFilter.java
```diff
@@ -0,0 +1,105 @@
...
+      return nextFilter.apply(requestHeader).thenApply(this::clearTimeoutCookie);
+    }
+    return nextFilter
+        .apply(requestHeader)
+        .thenCompose(
+            result ->
+                createSessionTimestampCookie(optionalProfile.get()).thenApply(result::withCookies));
+  }
+
+  private CompletableFuture<Http.Cookie> createSessionTimestampCookie(CiviFormProfile profile) {
```
gwendolyngoetz, line 72:
```suggestion
  /**
   * This cookie is used by client-side Javascript. It is not encrypted because the client needs to 
   * read it. We should avoid using values in the cookie that are sent on client requests. Extra
   * validation should be performed if we need to use the request cookie on the server.
   */
  private CompletableFuture<Http.Cookie> createSessionTimestampCookie(CiviFormProfile profile) {
```

(blocking) Adding a note on the usage of the cookie. Please reword if needed.

server/app/filters/SessionTimeoutFilter.java
gwendolyngoetz: (blocking) Please remove these and just use the defaults values from the application.conf

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):
Looks great!

PR #9839, opened 2025-03-02
Use correct value for program slug

gwendolyngoetz:
### Description

This fixes a Northstar bug on the applicant program cards I came across by chance.

The adminName property was being use to build the url on the cards instead of the slug. In the past the adminName was pulled from the `programs.name` column, but was later changed to use the `programs.localized_name`. When creating new programs these days the adminName and the slug are the same thing, but programs created prior to this point might not be different. This causes the older programs to build a URL that doesn't work. Changing this to the slug as that is the correct value to be used in the URL.

Note, that even if the Civiform admin were to edit the program and change the program name it doesn't update the value in the `programs.name` column so it's not fixable from within the application. We should probably deprecate the `programs.name` column and phase it out.

#### General

### DIFF
--- a/server/app/views/applicant/ProgramCardsSectionParamsFactory.java
+++ b/server/app/views/applicant/ProgramCardsSectionParamsFactory.java
@@ -140,7 +140,7 @@ public final class ProgramCardsSectionParamsFactory {
         getActionUrl(
             applicantRoutes,
             program.id(),
-            program.adminName(),
+            program.slug(),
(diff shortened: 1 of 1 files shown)


gwendolyngoetz:
@shreyachatterjee00, FYI. I didn't see an existing issue, but if there is one it can be closed when this is merged.

rockycodes (approved):
Good catch!

PR #9880, opened 2025-03-05
Introduce flag to hide the government name in the header

dkatzz:
### Description

This adds a flag that enables governments to hide the government name in the header. This was something Charlotte was interested in, since their logo already has their name within it. For northstar, we'll show the name when the logo doesn't show on mobile. By default this is turned off, so no governments will see a change unless they turn this on. 

#### General

#### User visible changes

Turn on and off the flag with and without northstar on. 

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/9901


server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
```diff
@@ -251,6 +251,11 @@ it is around the label */
     @include u-padding-top(4);
     @include u-padding-bottom(2);
   }
+
+  // Hide the government name on desktop (if set to hidden)
+  .cf-hidden-gov-name {
```
rockycodes, line 256: suggestion (non-blocking): I'd prefer to call this `cf-gov-name` since it's not always hidden.

server/app/views/applicant/NavigationFragment.html
```diff
@@ -25,7 +25,15 @@
               />
             </div>
             <em class="usa-logo__text">
-              <span th:text="${civicEntityShortName}"></span>
+              <span
+                class="cf-hidden-gov-name"
```
rockycodes, line 29: suggestion (non-blocking): Would you add a comment here that briefly explains how this works? I was confused at first about why were including `civicEntityShortName` twice.

rockycodes (approved):
Lgtm! Great idea for how to resolve this problem.

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
dkatzz: The confusing thing is that this element actually doesn't exist unless they have the flag turned on to hide the government name. But the reason governments may want to hide the name is if the logo has the name in it and with northstar, the logo disappears on mobile, so this is basically to show the hidden name if the logo is hidden (on mobile) and to hide it on desktop. Thoughts on the naming based on that? `cf-gov-name-with-hidden-flag` seems too long but something like that may be good? Or maybe `cf-gov-name` still makes sense but wanted to make sure it was clear first. 

https://github.com/user-attachments/assets/4c285938-3718-4e2c-b33a-4b696f6532e7

dkatzz: Yeah, I like that idea - updated!

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
```diff
@@ -251,6 +251,11 @@ it is around the label */
     @include u-padding-top(4);
     @include u-padding-bottom(2);
   }
+
+  // Hide the government name on desktop (if set to hidden). This class is only applied when the HIDE_CIVIC_ENTITY_NAME_IN_HEADER flag is enabled
+  .cf-gov-name {
```
gwendolyngoetz, line 256: Instead of defining these custom classes can you use the `.display-block` / `.display-none` USWDS [utility classes](https://designsystem.digital.gov/utilities/display/).

gwendolyngoetz, line 256: Instead of creating custom classes can you use the `.display-none` / `.tablet:display-block` USWDS [utility classes](https://designsystem.digital.gov/utilities/display)?

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
gwendolyngoetz: Instead of creating custom classes can you use the `.display-none` / `.tablet:display-block` USWDS [utility classes](https://designsystem.digital.gov/utilities/display)?

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
```diff
@@ -260,6 +265,11 @@ it is around the label */
     display: none;
   }
 
+  // Show the government name when the logo is hidden
+  .cf-hidden-gov-name {
```
rockycodes, line 269: Should this be `.cf-gov-name` also?

rockycodes (commented):
This lgtm, just one comment about making the class names match on mobile. I'm also curious about Gwen's question about using USWDS styles.

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
dkatzz: Me forgetting to change that made me realize that we don't need this, only the other one :) but good catch!

server/app/assets/stylesheets/northstar/_uswds-theme-custom-styles.scss
dkatzz: Good call! I had followed the pattern of cf-header-logo, but makes sense to use the utility class here. Updated - kept the custom class just for the test.

rockycodes (approved):

PR #9981, opened 2025-03-12
Fix S3 document upload policy fields

lagosr-google:
### Description

2 document upload policy fields needed changes:

-  content-length-range should have paramters of type <string, int, int> but was <string, string, string>
- servicename was changed accidentally to aws-s3 during the recent refactor of the storage classes.  set back to s3 now.

## Release notes

S3 Storage file upload policy for upload max length range was updated to specify the min, max values as integers, rather than strings.

#### General


gwendolyngoetz (changes requested):
This isn't the direction I think we should take. We already have a pattern for determining the [cloud storage provider](https://github.com/civiform/civiform/blob/main/server/app/modules/CloudStorageModule.java) that I'd prefer we follow. It's a little extra work to create some new base classes, but it's going to be better for the long term. Both in terms of being prepared for making any specific GCP changes that may come up and from a logical perspective. All these classes are specific to AWS in naming and it's going to be confusing to see it configured as AWS, but pointing at GCP.

lagosr-google:
PR was updated to now only fix the bug that was found on the S3 document upload policy

gwendolyngoetz (approved):

dkatzz (approved):

tallulahkay (approved):
LGTM

PR #9984, opened 2025-03-13
[North Star] Set program card image dimension to consistent height and clean up card section layout

### LINKED ISSUE #9961: [North Star] Program card images should be the same size
**Describe the bug**
Images on the program cards should be the same height.  Also, if the description is one long word, the word should be broken up and wrap so that the program card width doesn't stretch to fit (see screenshot 2).

**To Reproduce**

1. Make sure you have two images with different dimensions.
2. Add the images to two programs.
3. Go to the applicant home page.
4. Note that the height of the images is different.

**Expected behavior**
Just like in pre-North Star, images should be cropped so that they all have the same dimensions.

**Screenshots**

Image

Image

**Additional context**
- Cards in the same row should all have the same width and height.  This seems to be the existing behavior and we want to maintain that.
- We may want to make use of these USWDS classes:  `usa-card__media`, `usa-card__img`, `usa-card__media--exdent`, if we aren't already using them.

melanie-exygy:
### Description

North Star program card images had varying heights depending on the dimensions of the uploaded image.  This sets the aspect ratio to 16/9, which is what is currently in production for pre-North Star.

This also fixes a bug where a card with a single long word as a description would stretch the card size to very large (See screenshot in #9961.  This is fixed by using `overflow-wrap: anywhere`.

Lastly, there were some unused and/or conflicting card layout styles which I've cleaned up.

1. As an admin, create a program with a very long word as the short description.
2. Publish your changes.
3. From the applicant home page, note that the long word is broken up and wraps.  Also note that all cards in a single row have the same size.
4. Make sure you have two images with different heights.
5. Add the images to two programs.
6. Go to the applicant home page.
7. Note that the height of the images on the program cards is the same.

### Issue(s) this completes

Fixes #9961


server/app/views/applicant/ProgramCardsSectionFragment.html
```diff
@@ -33,7 +30,7 @@ <h3 th:id="${section.id().get()}" th:text="${section.title().get()}"></h3>
 <li
   th:fragment="card (card, sectionType)"
   th:with="showImage=(${sectionType != 'MY_APPLICATIONS' and card.imageSourceUrl().isPresent() and not #strings.isEmpty(card.imageSourceUrl().get())})"
-  class="usa-card tablet-lg:grid-col-6 widescreen:grid-col-4 cf-application-card"
```
melanie-exygy, line 36: **note:** The `tablet-lg` and `widescreen` styles weren't being applied because they were being overwritten by the styles for `cf-program-card-group`.  Here, I'm using the USWDS out-of-the-box layout styles to replace those from `cf-program-card-group`.

server/app/views/applicant/ProgramCardsSectionFragment.html
```diff
@@ -51,11 +48,10 @@ <h3 th:id="${section.id().get()}" th:text="${section.title().get()}"></h3>
 
     <!--*/ Optional image */-->
     <div th:if="${showImage}" class="usa-card__media usa-card__media--exdent">
-      <div class="usa-card__img">
+      <div class="usa-card__img add-aspect-16x9">
```
melanie-exygy, line 51: **note:** This aspect ratio mimics what we currently have in production.

EmiliaPaz: Sounds good.
Side note (outside this PR): is this communicated somewhere to the administrators? It would be nice for them to know the ratio, so they can properly upload images

browser-test/image_snapshots/northstar_applicant_application_index_test/north-star-program-image-wide.png
EmiliaPaz, line 1: (non-bloking) This looks the same as applicant_application_index_test/north-star-program-image-wide.png. Could they use the same screenshot?
Edit: quick look, the tests are the same ([#1](https://github.com/civiform/civiform/blob/fb36a8ef453b0f9e60e63882ca8d07a6a060cb23/browser-test/src/applicant/applicant_application_index.test.ts#L1206-L1238), [#2](https://github.com/civiform/civiform/blob/fb36a8ef453b0f9e60e63882ca8d07a6a060cb23/browser-test/src/applicant/northstar_applicant_application_index.test.ts#L837-L876)). Maybe one can be dropped?

EmiliaPaz (approved):
Thanks Melanie!

browser-test/image_snapshots/northstar_applicant_application_index_test/north-star-program-image-wide.png
melanie-exygy: Tallulah just migrated the applicant_application_index tests to a new test file.  I think we should these should have been removed from the non-North Star file.  Let me check with her and report here...

melanie-exygy: The duplicate tests and screenshots that are in the non-North Star file are going to be removed in another PR.  Good catch!

PR #9991, opened 2025-03-13
Update screen reader behavior for required fields

### LINKED ISSUE #9910: [a11y][Mobile][iOS/android][WCAG 2.4.8 ] [v2.1 AA] Required fields are failing to properly announce it
OBSERVED RESULTS:
--

There are some required fields (Marked with a red asterisk) that are announcing it incorrectly while using the screen reader in both android and iOS, verbalizing like "First Name, star" or directly not mentioning the status of required

[iOS](https://drive.google.com/file/d/1BZo2XhScSuLcsZ7NeasU2AV9lC6sSyjw/view?usp=drive_link)

[Android](https://drive.google.com/file/d/1Ej5UeTbWThbGWCrLuc7cOSuLgVgiziRv/view?usp=drive_link)

EXPECTED RESULTS:
--

Required fields are important so providing the correct context to users that rely on screen readers is a priority, something like "FIELDNAME, text field, REQUIRED" is an ideal verbalization to let the user know that he must fill those fields

STEPS TO REPRODUCE:
--
0. Turn on the screen reader (VoiceOver or Talkback )
1. Go to [URL](https://ns.civiform.dev/programs/76/blocks/1/edit)
2. Navigate in a linear manner and notice the verbalization of the required fields that are marked with a red asterisk

RELEVANT GAR SHEET:
--

[GAR sheet](https://docs.google.com/spreadsheets/d/1mgAhVZ3tuJ0O2EeB5JZ0HAvSrxVeIUjlJxVAhu3JH7M/edit?gid=1240238308#gid=1240238308)

ADDITIONAL INFORMATION:
--

[Success Criterion 1.3.5 Identify Input Purpose](https://www.w3.org/TR/WCAG21/#identify-input-purpose)

ENVIRONMENT DETAILS:
--

---
- Devices: iPhone 14 plus: iOS 18.2.1
- Screen reader name: VoiceOver
- Browser name: Chrome Mobile: v133.0.69
---
- Devices: Pixel 7: android 15
- Screen reader name: TalkBack
- Browser name: Chrome Mobile: v133.0.69
---

watermandrewc:
### Description

A few question types in North Star do not have an aria-required label set when the question is required. Adding the label so that the screen reader will recognize when the question is required.

Also hiding the required (*) from screen readers so that it does not get read as "star". The label will not have a required indication when read by a screen reader, but the input itself will. This change impacts both North Star and legacy UIs.

#### General

#### User visible changes

Go to a program with required questions, and use a screen reader to read the elements on the page aloud. It would be best to test multiple types of questions. The screen reader should not read the asterisks aloud, but inputs with a label that have one should be labeled as "required" when read by the screen reader.

Test in both NS and legacy UIs.

### Issue(s) this completes

Fixes #9910

### DIFF
--- a/server/app/views/questiontypes/AddressQuestionFragment.html
+++ b/server/app/views/questiontypes/AddressQuestionFragment.html
@@ -151,7 +151,7 @@
         <span
           th:if="${!question.isOptional()}"
           class="usa-hint--required"
-          th:aria-label="#{content.required}"
+          th:aria-hidden="true"
         >
           *</span
         >
--- a/server/app/views/questiontypes/AddressQuestionFragment.html
+++ b/server/app/views/questiontypes/AddressQuestionFragment.html
@@ -40,7 +40,11 @@
       </th:block>
       <label class="usa-label" th:for="${inputId}">
         <span th:text="#{label.street}"></span>
-        <span th:if="${!question.isOptional()}" class="usa-hint--required">
+        <span
+          th:if="${!question.isOptional()}"
+          class="usa-hint--required"
+          th:aria-hidden="true"
+        >
           *</span
         >
       </label>
@@ -106,7 +110,11 @@
       </th:block>
       <label class="usa-label" th:for="${inputId}">
         <span th:text="#{label.city}"></span>
-        <span th:if="${!question.isOptional()}" class="usa-hint--required">
+        <span
+          th:if="${!question.isOptional()}"
+          class="usa-hint--required"
+          th:aria-hidden="true"
+        >
           *</span
         >
       </label>
@@ -140,7 +148,11 @@
       </th:block>
       <label class="usa-label" th:for="${inputId}">
         <span th:text="#{label.state}"></span>
-        <span th:if="${!question.isOptional()}" class="usa-hint--required">
+        <span
+          th:if="${!question.isOptional()}"
+          class="usa-hint--required"
+          th:aria-label="#{content.required}"
+        >
           *</span
         >
       </label>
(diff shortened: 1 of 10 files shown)


server/app/views/questiontypes/AddressQuestionFragment.html
```diff
@@ -140,7 +148,11 @@
       </th:block>
       <label class="usa-label" th:for="${inputId}">
         <span th:text="#{label.state}"></span>
-        <span th:if="${!question.isOptional()}" class="usa-hint--required">
+        <span
+          th:if="${!question.isOptional()}"
+          class="usa-hint--required"
+          th:aria-label="#{content.required}"
```
melanie-exygy, line 154: **question (blocking):** Shouldn't this be `th:aria-hidden="true"` like the other address fields?

server/app/views/questiontypes/NameQuestionFragment.html
```diff
@@ -71,6 +76,7 @@
       autocomplete="additional-name"
       th:value="${nameQuestion.getMiddleNameValue().orElse('')}"
       th:name="${nameQuestion.getMiddleNamePath()}"
+      th:aria-required="${!question.isOptional()}"
```
melanie-exygy, line 79: **issue (blocking):** The middle name field is never required.

melanie-exygy (commented):
Thanks for doing this really important work to make accessibility better!

server/app/views/questiontypes/AddressQuestionFragment.html
watermandrewc: Oops yes, missed this one when backpedaling on an earlier approach. Thanks!

melanie-exygy (approved):

PR #10050, opened 2025-03-18
[SessionTimeout][part6]Display session timeout warning and enforcement in frontend JS

codekarma:
### Description

Frontend code to read the session_timeout_data cookie and display warning message in a modal dialog or logout user if the session has timed out due to inactivity or total session length above a threshold. The session cookie is only set when session timeout feature is enabled and therefore the SessionTimeoutHandler initialization is not under feature flag. Browser tests are coming in the next PR.

#### General

#### User visible changes

#### New Features

- Configure the session timeouts in server configuration file(in applicantion.conf). To test session inactivity warning dialog, set the session_inactivity_timeout_minutes to 2 minutes and session_inactivity_warning_threshold_minutes to 1 minute. Set maximum_session_duration_minutes to 2 minutes and session_duration_warning_threshold_minutes to 1 minutes for session length warning dialog
- Log in as Admin
- Enable the session timeout in server settings(if not enabled in step 1).
- Visiting any page should now display and waiting for a minute would show a warning modal dialog.
- Logout as Admin
- Login as Applicant
- Any page visit should display the warning dialog.
- Now configure the session_inactivity_timeout_minutes to 1 minute
- Wait for 1 minute and you should be logged out automatically.
- Now configure the maximum_session_duration_minutes to 1 minute
- Wait for 1 minute and you should be logged out automatically.

### Issue(s) this completes

### DIFF
--- a/server/app/assets/javascripts/admin_entry_point.ts
+++ b/server/app/assets/javascripts/admin_entry_point.ts
@@ -21,6 +21,7 @@ import * as devIcons from './dev_icons'
 import * as modal from './modal'
 import * as questionBank from './questionBank'
 import PreviewController, * as preview from './preview'
+import {SessionTimeoutHandler} from './session'
 import * as enumerator from './enumerator'
 import * as phoneNumber from './phone'
 import * as adminQuestionEdit from './admin_question_edit'
@@ -33,6 +34,7 @@ import htmx from './htmx'
 
 window.addEventListener('load', () => {
   initializeEverything()
+  SessionTimeoutHandler.init()
   htmx.on('htmx:afterSettle', () => {
     afterSettle()
   })
--- a/server/app/assets/javascripts/admin_entry_point.ts
+++ b/server/app/assets/javascripts/admin_entry_point.ts
@@ -34,7 +34,6 @@ import htmx from './htmx'
 
 window.addEventListener('load', () => {
   initializeEverything()
-  SessionTimeoutHandler.init()
   htmx.on('htmx:afterSettle', () => {
     afterSettle()
   })
@@ -69,6 +68,7 @@ function initializeEverything(): void {
   trustedIntermediaryController.init()
   fileUpload.init()
   azureUpload.init(AZURE_ADMIN_FILEUPLOAD_FORM_ID)
+  SessionTimeoutHandler.init()
 }
 
 function afterSettle(): void {
--- a/server/app/assets/javascripts/session.test.ts
+++ b/server/app/assets/javascripts/session.test.ts
@@ -403,7 +403,7 @@ describe('SessionTimeoutHandler', () => {
 
     afterEach(() => {
       jest.useRealTimers()
-      document.cookie = `${SessionTimeoutHandler['COOKIE_NAME']}=; expires=Thu, 01 Jan 1970 00:00:00 GMT`
+      document.cookie = `${SessionTimeoutHandler['TIMEOUT_COOKIE_NAME']}=; expires=Thu, 01 Jan 1970 00:00:00 GMT`
     })
 
     it('immediately logs out if timeout is reached', () => {
--- a/server/app/assets/javascripts/session.test.ts
+++ b/server/app/assets/javascripts/session.test.ts
@@ -1,0 +1,392 @@
+import {jest, expect, describe, it, beforeEach, afterEach} from '@jest/globals'
+import {SessionTimeoutHandler} from './session'
+import {ToastController} from './toast'
+import {WarningType} from './session'
+
+type SessionTimeoutHandlerType = typeof SessionTimeoutHandler & {
+  logout: () => void
+  checkAndSetTimer: () => void
+  showWarning: (type: WarningType) => void
+  inactivityWarningShown: boolean
+  totalLengthWarningShown: boolean
+}
+
+describe('SessionTimeoutHandler', () => {
+  let container: HTMLElement
+  let inactivityModal: HTMLElement
+  let lengthModal: HTMLElement
+  let extendSessionForm: HTMLFormElement
+  let consoleSpy: ReturnType<typeof jest.spyOn>
+
+  beforeEach(() => {
+    // Set up DOM elements
+    container = document.createElement('div')
+    container.id = 'session-timeout-modals'
+    document.body.appendChild(container)
+
+    // Create inactivity warning modal with new structure
+    inactivityModal = document.createElement('div')
+    inactivityModal.id = 'session-inactivity-warning-modal'
+    inactivityModal.classList.add('is-hidden', 'usa-modal')
+    inactivityModal.setAttribute(
+      'data-modal-type',
+      'session-inactivity-warning',
+    )
+    container.appendChild(inactivityModal)
+
+    // Create extend session form
+    extendSessionForm = document.createElement('form')
+    extendSessionForm.id = 'extend-session-form'
+    extendSessionForm.setAttribute('hx-post', '/extend-session')
+    extendSessionForm.setAttribute('hx-target', 'this')
+    extendSessionForm.setAttribute('hx-swap', 'none')
+
+    // Add CSRF token input
+    const csrfInput = document.createElement('input')
+    csrfInput.setAttribute('name', 'csrfToken')
+    csrfInput.value = 'test-csrf-token'
+    extendSessionForm.appendChild(csrfInput)
+
+    // Create primary button (extend session)
+    const primaryButton = document.createElement('button')
+    primaryButton.textContent = 'Extend Session'
+    primaryButton.classList.add('usa-button')
+    primaryButton.setAttribute('data-modal-primary', '')
+    primaryButton.setAttribute('data-modal-type', 'session-inactivity-warning')
+    extendSessionForm.appendChild(primaryButton)
+
+    inactivityModal.appendChild(extendSessionForm)
+
+    // Create secondary button (cancel)
+    const secondaryButton = document.createElement('button')
+    secondaryButton.textContent = 'Cancel'
+    secondaryButton.classList.add('usa-button', 'usa-button--unstyled')
+    secondaryButton.setAttribute('data-modal-secondary', '')
+    secondaryButton.setAttribute(
+      'data-modal-type',
+      'session-inactivity-warning',
+    )
+    inactivityModal.appendChild(secondaryButton)
+
+    // Create close button
+    const closeButton = document.createElement('button')
+    closeButton.textContent = 'Close'
+    closeButton.classList.add('usa-button', 'usa-modal__close')
+    closeButton.setAttribute('data-close-modal', '')
+    closeButton.setAttribute('data-modal-type', 'session-inactivity-warning')
+    inactivityModal.appendChild(closeButton)
+
+    // Create session length warning modal
+    lengthModal = document.createElement('div')
+    lengthModal.id = 'session-length-warning-modal'
+    lengthModal.classList.add('is-hidden', 'usa-modal')
+    lengthModal.setAttribute('data-modal-type', 'session-length-warning')
+    container.appendChild(lengthModal)
+
+    // Create primary button (logout)
+    const logoutButton = document.createElement('button')
+    logoutButton.textContent = 'Logout'
+    logoutButton.classList.add('usa-button')
+    logoutButton.setAttribute('data-modal-primary', '')
+    logoutButton.setAttribute('data-modal-type', 'session-length-warning')
+    lengthModal.appendChild(logoutButton)
+
+    // Create secondary button (cancel)
+    const lengthCancelButton = document.createElement('button')
+    lengthCancelButton.textContent = 'Cancel'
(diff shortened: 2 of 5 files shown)


server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+  static init() {
+    void this.checkAndSetTimer()
+    this.setupModalEventHandlers()
+  }
+
+  /**
+   * Set up event handlers for the server-rendered modals.
+   */
+  private static setupModalEventHandlers() {
+    document.addEventListener('htmx:afterRequest', (event: Event) => {
```
dkatzz, line 58: **suggestion:** Maybe add a comment noting this sets up the handlers for extending sessions (assuming I'm understanding correctly)

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+    return (
+      typeof d.inactivityWarning === 'number' &&
+      typeof d.inactivityTimeout === 'number' &&
+      typeof d.totalWarning === 'number' &&
+      typeof d.totalTimeout === 'number' &&
+      typeof d.currentTime === 'number'
+    )
+  }
+
+  private static getCookie(name: string): string | null {
```
dkatzz, line 181: **suggestion:** Can you add a comment with more explanation of what is happening in this function?

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+    const value = `; ${document.cookie}`
+    const parts = value.split(`; ${name}=`)
+    if (parts.length === 2) {
+      return parts.pop()?.split(';').shift() || null
+    }
+    return null
+  }
+
+  private static checkAndSetTimer() {
+    const data = this.getTimeoutData()
```
dkatzz, line 191: **question:** Just to check my understanding, timeoutdata will be null if the feature flag is off (session timeouts aren't enabled), right?

server/app/assets/javascripts/session.ts
dkatzz, line 1: **praise:** Thanks for the comment explanations!

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+      const modalType = target.getAttribute('data-modal-type')
+      if (!modalType) return
+
+      if (target.hasAttribute('data-modal-primary')) {
+        if (modalType === 'session-inactivity-warning') {
+          // Handle extend session
+          const form = document.getElementById(
+            'extend-session-form',
+          ) as HTMLFormElement
+          form?.requestSubmit()
```
dkatzz, line 113: **question:** what does requestSubmit() do here? Can you add a comment?

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  private static hideModal(modalType: string) {
+    const modal = document.getElementById(`${modalType}-modal`)
+    if (modal) {
+      modal.classList.add('is-hidden')
+    }
+  }
+
+  private static getTimeoutData(): TimeoutData | null {
+    const cookieValue = this.getCookie(this.COOKIE_NAME)
```
dkatzz, line 140: **thought:** No strong preference but you could consider pulling this cookie check out of this function and put it directly into the parent function or create another isSessionDataInCookie() or something

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,263 @@
...
+    try {
+      // Decode Base64 and parse JSON
+      const jsonString = atob(decodeURIComponent(cookieValue))
+      const data: unknown = JSON.parse(jsonString)
+      if (!this.isTimeoutData(data)) {
+        console.error('Invalid timeout data format')
+        return null
+      }
+      // Calculate clock skew between client and server
+      const clientNow = Math.floor(Date.now() / 1000)
```
dkatzz, line 152: **thought:** Could there be any weirdness with timezones here vs on the server side?

dkatzz (commented):
Didn't get a chance to fully review / look at tests, but added some initial thoughts and comments.

server/app/assets/javascripts/session.ts
codekarma: Added lots of comments

server/app/assets/javascripts/session.ts
codekarma: Yes that is correct. There will be no cookie to read timeoutData from and therefore this would be just null.

server/app/assets/javascripts/session.ts
codekarma: It is just an html(x) form submission. Added comment that it hits the extend-session endpoint.

server/app/assets/javascripts/session.ts
codekarma: So this function encapsulate getting the timeout data from the cookie. Reading cookie is already a separate function that returns null if there in no session timeout data cookie. Checking for cookie validity is minimal(only checking if it is null. The next part is to check the data is actually timeout data and that is a separate function already.

server/app/assets/javascripts/session.ts
codekarma: Since we are using Unix times, these are all long values representing millisecs after epoch and we adjust for the clock skew already, I don't see any problem with timezone difference.

server/app/assets/javascripts/admin_entry_point.ts
```diff
@@ -33,6 +34,7 @@ import htmx from './htmx'
 
 window.addEventListener('load', () => {
   initializeEverything()
+  SessionTimeoutHandler.init()
```
rockycodes, line 37: question (non-blocking): Is there a reason you're initializing this separately rather than including it in the `initializeEverything` method?

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,396 @@
...
+interface TimeoutData {
+  /** When to show the inactivity warning modal */
+  inactivityWarning: number
+  /** When to logout due to inactivity */
+  inactivityTimeout: number
+  /** When to show the total session length warning modal */
+  totalWarning: number
+  /** When to logout due to total session length */
+  totalTimeout: number
+  /** Server's current time when cookie was set - used for clock skew calculation */
```
rockycodes, line 16: praise: really helpful comments!

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,396 @@
...
+  private static totalLengthWarningShown = false
+  /** Tracks if inactivity warning has been shown at least once */
+  private static hasInactivityWarningBeenShown = false
+  /** Tracks if total length warning has been shown at least once */
+  private static hasTotalLengthWarningBeenShown = false
+  /** Current active timeout timer */
+  private static timer: number | null = null
+
+  static init() {
+    void this.checkAndSetTimer()
```
rockycodes, line 52: nitpick (non-blocking): It would be helpful to have the `checkAndSetTimer` method defined directly under this the `init` method for easy reference.

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,396 @@
...
+/**
+ * Handles session timeout management including:
+ * - Reading timeout data from cookies
+ * - Showing warning modals before timeout
+ * - Handling session extension requests
+ * - Managing automatic logout
+ */
+export class SessionTimeoutHandler {
+  /** Name of cookie storing timeout data */
+  private static COOKIE_NAME = 'session_timeout_data'
```
rockycodes, line 39: suggestion (non-blocking): name this var `TIMEOUT_COOKIE_NAME` to match the one in `SessionTimeoutFilter.java` and add a comment noting that they must be kept in sync. [Here's](https://github.com/civiform/civiform/blob/1f7644c8116708a88a176f3eb42d116a0a172694/server/app/assets/javascripts/admin_export_view.ts#L4) an example of where we do this elsewhere

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    // 2 & 3. Show warnings if they haven't been shown before and their time has passed
+    if (!this.hasInactivityWarningBeenShown && data.inactivityWarning <= now) {
+      this.showWarning(WarningType.INACTIVITY)
+      this.hasInactivityWarningBeenShown = true
+      return
+    }
+
+    if (!this.hasTotalLengthWarningBeenShown && data.totalWarning <= now) {
```
rockycodes, line 291: praise: great job making this complex logic easy to understand!

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    timeouts.push({
+      time: data.totalTimeout,
+      action: () => this.handleTimeout(),
+    })
+
+    // Sort by earliest time
+    timeouts.sort((a, b) => a.time - b.time)
+
+    if (timeouts.length > 0) {
```
rockycodes, line 336: question (non-blocking): Is there ever a situation where we would get here and there would be no timeouts?

rockycodes (commented):
Commenting for now since I'm not quite done reviewing but need to sign off for the day.

server/app/assets/javascripts/admin_entry_point.ts
codekarma: I moved it into intializeEverything().

server/app/assets/javascripts/session.ts
codekarma: Make sense to have these methods closer.

server/app/assets/javascripts/session.ts
codekarma: You are right. Now since we are always adding the timers for timeout, this will always have legth 2 or more(2, 3, or 4).

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,395 @@
...
+     *   - Resets warning visibility flags
+     *   - Rechecks timeout conditions
+     *
+     * Uses data attributes to identify buttons and their actions:
+     * - data-modal-type: Identifies which modal the button belongs to
+     * - data-modal-primary: Primary action button (extend/logout)
+     * - data-modal-secondary: Secondary action button (dismiss)
+     * - data-close-modal: Close button (X)
+     */
+    document.addEventListener('click', (event) => {
```
rockycodes, line 217: question (maybe blocking): Is there a way to apply this click handler just to the modals instead of the entire document?

rockycodes: Applying them separately to the different modals based on their ids could help simplify some of the if/else logic.

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,395 @@
...
+    })
+  }
+
+  /**
+   * Hides the specified modal by adding the 'is-hidden' class.
+   * Updates visibility tracking flags when hiding a modal.
+   *
+   * @param modalType Type of modal to hide ('session-inactivity-warning' or 'session-length-warning')
+   */
+  private static hideModal(modalType: string) {
```
rockycodes, line 255: suggestion (non-blocking): move this method to `modal.ts` and call it `hideUswdsModal` or something like that so it can easily be used in other places.

server/app/assets/javascripts/session.ts
```diff
@@ -0,0 +1,395 @@
...
+   * @param type Type of warning to show (inactivity or total length)
+   */
+  private static showWarning(type: WarningType) {
+    // Check if any warning is already shown to prevent showing multiple dialogs
+    if (this.inactivityWarningShown || this.totalLengthWarningShown) return
+
+    // Get the appropriate modal element
+    const modalId =
+      type === WarningType.INACTIVITY
+        ? 'session-inactivity-warning-modal'
```
rockycodes, line 365: suggestion (non-blocking): Since you use the strings `session-inactivity-warning` and `session-length-warning` multiple times you could store them in a const at the top of the file and note that they should be kept in sync with wherever they are set on the elements.

server/app/assets/javascripts/session.test.ts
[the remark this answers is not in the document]
rockycodes: +1! Really fantastic tests!

rockycodes (changes requested):
Only potentially blocking comment is about adding the click handler to the whole document vs the modals directly.

server/app/assets/javascripts/session.ts
codekarma: Adding separate click handler for dialogs. Event delegation help simplify code and improve perf in more complex SPAs but in our case, maybe it was an overkill.

server/app/assets/javascripts/session.ts
codekarma: done. Created an enum.

server/app/assets/javascripts/session.ts
rockycodes: Ah that's good to know!

rockycodes (approved):
looks great!

PR #9983, opened 2025-03-13
[Northstar] fix: Admin Program Card Preview doesn't display category tag text

tallulahkay:
### Description

Color tokens do not have an effect in admin - manually added the ink color so it appears on the card preview in admin.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [ ] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)
- [ ] Removed the release notes section if the title is sufficient for the release notes description, or put more details in that section.
- [ ] Created unit and/or browser tests which fail without the change (if possible)
- [ ] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [ ] Extended the README / documentation, if necessary. For user-facing features, consider updating [the user docs](https://github.com/civiform/docs). For "under-the-hood" changes or things more relevant to developers, consider updating [the dev wiki](https://github.com/civiform/civiform/wiki).
- [ ] Ensured PII wasn't added to any new logs, unless it was guarded by `isDevOrStaging`

### Instructions for manual testing

Enable north star, navigate to a program, add tags to the program, confirm that the tags show up with the correct text color in the card preview.

### Issue(s) this completes

Fixes #9985;

PR #10053, opened 2025-03-18
[North Star] Fix styling of the admin program image preview

### LINKED ISSUE #9962: [North Star] Program card image preview has layout issues
**Describe the bug**
In the program card preview, when an admin is adding an image to a program card, the preview card is off-center and too narrow.  The program name also extends beyond the boundaries of the card.    

**To Reproduce**

1. As an admin, from the Admin Program List view, click "Edit" on a program card.
2. Click "Edit program image"
3. Add an image.  
4. See the preview on the right side of the page.

**Expected behavior**
- The card preview should have a size similar to what it will be on the applicant home page.
- The program name should stay within the bounds of the card.
- The card should be centered in the right half of the page.  

**Screenshots**

Image

- OS: Mac OS
- Browser: Chrome

melanie-exygy:
### Description

#### User visible changes

### Issue(s) this completes

Fixes #9962


browser-test/image_snapshots/admin_program_image_test/program-image-preview.png
melanie-exygy, line 1: **note:** Even though we want to avoid screenshots whenever possible, we do want one screenshot of the page.  There wasn't one, so I'm adding it.  This is also the best way to test the styles.

server/app/views/applicant/ProgramCardsSectionFragment.html
```diff
@@ -30,7 +33,7 @@ <h3 th:id="${section.id().get()}" th:text="${section.title().get()}"></h3>
 <li
   th:fragment="card (card, sectionType)"
   th:with="showImage=(${sectionType != 'MY_APPLICATIONS' and card.imageSourceUrl().isPresent() and not #strings.isEmpty(card.imageSourceUrl().get())})"
-  class="usa-card tablet:grid-col-6 desktop:grid-col-4 cf-application-card"
```
melanie-exygy, line 33: **note:** These grid column sizes only work for the layout on the applicant home page.  I moved the styles into the `cf-program-card-group`, so that we can set this at the card group level (and not set it for the admin image preview page).

server/app/views/admin/programs/ProgramCardPreviewFragment.html
```diff
@@ -1,6 +1,6 @@
-<div id="card-preview" class="margin-y-4">
+<div id="card-preview">
   <!-- Card group containing one card. -->
-  <ul class="usa-card-group cf-program-card-group">
+  <ul class="usa-card-group margin-x-15 display-flex flex-column">
```
tallulahkay, line 3: suggestion (non-blocking): usa-card-group already makes this a flex box, so you can use a `flex-justify-center` class to center the content - that way you will no longer need the margin

browser-test/image_snapshots/admin_program_image_test/program-image-preview.png
tallulahkay: agreed, good screenshot! I'm adding one with the category tags in this pr, but maybe I should just update the one you added as to not have too many screenshots? https://github.com/civiform/civiform/pull/9983/files#diff-fbeb0ea82b0b33cfd4359b647ed3313b20e2ebebd77c5242ac531bd5a7ef6334

tallulahkay (approved):

server/app/views/admin/programs/ProgramCardPreviewFragment.html
melanie-exygy: That was my original approach, but what ends up happening without the margin is that the card stretches to take up the whole space, like this:  
image

PR #10075, opened 2025-03-20
[North Star] Adding a link to skip to main content of the page

### LINKED ISSUE #6067: Add skip links so that keyboard and screen reader users can skip to main content
Screen reader and keyboard users must first navigate through all of the header content before getting to the main content of the page.  This can be especially frustrating to users with motor impairments.

We should provide a skip link at the top of each applicant-facing page that allows the user to skip to the main content.  See these resources for more information and examples:
- [WCAG](https://www.w3.org/TR/WCAG20-TECHS/G1.html)
- [WebAIM](https://webaim.org/techniques/skipnav/)

Here's an example of how the USWDS site has done this.  It's the first item that is focused when you start tabbing.
image

### LINKED ISSUE #9916: [a11y] [WCAG 2.4.1]: The page doesn't offer a way to skip repetitive content.
OBSERVED RESULTS:
--
When navigating the TestCity CiviForm page using the keyboard, the header repeats on every product page. Users have to tab through all the icons in the header, with no option to skip them. This makes navigation slower and more cumbersome as there is no "Skip to main content" button to bypass the repetitive elements.

USER IMPACT:
--
The Screen reader users may experience difficulty navigating through repetitive content, such as menus or headers, if they cannot easily skip past these blocks to access the main content.

EXPECTED RESULTS:
--
When navigating the TestCity CiviForm page using the keyboard, users should be able to bypass the repetitive header elements and directly access the main content. This can be achieved by providing a "Skip to main content" button, allowing for a smoother and more efficient navigation experience for keyboard users.

STEPS TO REPRODUCE:
--
0. Enable ChromeVox using "CONTROL+ALT+Z".
1. Go to [TestCity CiviForm](https://ns.civiform.dev/programs)
2. Navigate through the page with the keyboard and notice that there is no method to skip the repetitive header content.

SCREENCAST:
--
- Chromebook+ChromeVox:[Evidence](https://drive.google.com/file/d/1lEm1SagPMTnjyrFeZjrJRy-onP7abCA6/view?usp=drive_link&resourcekey=0-LXShg_3qnWDpCjBXa4UlOw)
- Windows+NVDA:[Evidence](https://drive.google.com/file/d/1hlxMRPrZAyE_ByZkAtQqM2P9CUsk_vg-/view?usp=drive_link&resourcekey=0-FgKcm-SLZ7dk7RoYooWKsQ)

ADDITIONAL INFORMATION:
--
**WCAG rule:**

[WCAG 2.4.1](https://www.w3.org/TR/WCAG22/#bypass-blocks)

**Enviroment details:**

Machine: Chromebook Version 13.3

Browser: Chrome (132.0.6834.208 (Official Build) (64-bit))

Machine: Windows 11 Enterprise 23H2

Screen reader: ChromeVox, NVDA

Bug Filling Time: 20 mins

melanie-exygy:
### Description

Adding a skip link so that keyboard and screen reader users can skip to main content.  The link is the first focusable element on the page and is visible on keyboard focus only.  Once the user has clicked the link, clicking "Tab" will go to the first focusable element in the main content, rather than back up to the top of the page.

The skip link needs to be present on every applicant-facing page of the North Star.

1. Turn the North Star flag on. 
2. From any applicant-facing page, press "Tab".
3. The "Skip to main content" link should become visible and should have focus.
4. Press "Enter".
5. Press "Tab".
6. Focus should be on the first interactive element of the main content of the page.
7. Refresh the page.
8. Turn the screen reader on.
9. Focus should go straight to the "Skip to main content link" and the link should be announced.
10. Click the link.
11. It should skip to main content and start reading there.

### Issue(s) this completes

Fixes #9916; Fixes #6067

### DIFF
--- a/browser-test/src/applicant/northstar_applicant_application_index.test.ts
+++ b/browser-test/src/applicant/northstar_applicant_application_index.test.ts
@@ -14,7 +14,7 @@ import {
   selectApplicantLanguage,
   normalizeElements,
 } from '../support'
-import {Page} from 'playwright'
+import {Locator, Page} from 'playwright'
 import {ProgramVisibility} from '../support/admin_programs'
 import {BASE_URL} from '../support/config'
 
@@ -90,7 +90,25 @@ test.describe('applicant program index page', {tag: ['@northstar']}, () => {
     await applicantQuestions.expectTitle(page, 'Find programs')
   })
 
-  test('validate accessibility', async ({page}) => {
+  test('validate accessibility and validate skip link', async ({page}) => {
+    const skipLinkLocator: Locator = page.getByRole('link', {
+      name: 'Skip to main content',
+    })
+    await test.step('Tab and verify focus on skip link', async () => {
+      await page.keyboard.press('Tab')
+      await expect(skipLinkLocator).toBeFocused()
+      await expect(skipLinkLocator).toBeVisible()
+    })
+
+    await test.step('Click on skip link and skip to main content', async () => {
+      await skipLinkLocator.click()
+      await expect(page.locator('main')).toBeFocused()
+      await page.keyboard.press('Tab')
+      await expect(
+        page.getByRole('link', {name: 'View and apply'}).first(),
+      ).toBeFocused()
+    })
+
     await validateAccessibility(page)
   })
 
--- a/server/app/services/MessageKey.java
+++ b/server/app/services/MessageKey.java
@@ -288,6 +288,7 @@ public enum MessageKey {
   LINK_PROGRAM_DETAILS_SR("link.programDetailsSr"),
   LINK_REMOVE_FILE("link.removeFile"),
   LINK_SELECT_NEW_CLIENT("link.selectNewClient"),
+  LINK_SKIP_TO_MAIN_CONTENT("link.skipToMainContent"), // North Star only
   LINK_HOME("link.home"), // North Star only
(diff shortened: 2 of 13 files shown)


browser-test/src/applicant/northstar_applicant_application_index.test.ts
```diff
@@ -90,7 +90,25 @@ test.describe('applicant program index page', {tag: ['@northstar']}, () => {
...
 
-  test('validate accessibility', async ({page}) => {
+  test('validate accessibility and validate skip link', async ({page}) => {
+    const skipLinkLocator: Locator = page.getByRole('link', {
+      name: 'Skip to main content',
+    })
+    await test.step('Tab and verify focus on skip link', async () => {
+      await page.keyboard.press('Tab')
+      await expect(skipLinkLocator).toBeFocused()
+      await expect(skipLinkLocator).toBeVisible()
```
gwendolyngoetz, line 100: No change required. I think `toBeFocused` will fail if the element is not visible, making the second line redundant. No need to change anything, but it would be good for us to look into for general knowledge.

gwendolyngoetz (approved):

browser-test/src/applicant/northstar_applicant_application_index.test.ts
melanie-exygy: Got it.  Will look into it and let you know what I discover.

PR #10084, opened 2025-03-20
Slow question bank perf

### LINKED ISSUE #9094: Question Bank Causes Slow Page Loads with Many Questions
On the admin program edit page, the full contents of the Question Bank get loaded when the page loads.

Seattle has 850+ questions and this makes the page weight over 1MB and causes very slow loading.

The duplication of all the tailwind classes and the SVGs contribute to this large size. We don't have any paging on the question bank, but they way it filters on the client side currently requires all the questions.

We could probably htmx-ify this pretty easily. 

In general re-vamping the bank as a whole might be useful. It gets difficult finding questions when there are this many.

swatkat1:
### Description

Seattle question backs many seconds to load to due having about 6.6. MB of questions. 

This PR does two things
1. The svg in the question bank are repeated each time. This was making them link to just a single instance.
2. Removing selected tailwind for a custom css classes

## Release notes

Performance improvements on Question bank loading page. 

#### General

#### Database evolutions

- [ ] Assigned two reviewers
- [ ] Every comment in script should begin with -- and not # --- unless it denotes Ups or Downs. See [here](https://www.playframework.com/documentation/2.9.x/Evolutions) for details.
- [ ] Tested both the Downs and the Ups scripts manually (When testing, include all comments from the evolution in your test script to ensure any syntax errors in the comments are caught.)
- [ ] Update the model documentation in our [wiki](https://github.com/civiform/civiform/wiki/Database) if necessary

#### Durable jobs

- [ ] Assigned two reviewers

#### User visible changes

#### New Features

### Issue(s) this completes

Fixes #9094


server/app/views/components/Icons.java
```diff
@@ -467,8 +469,21 @@ public enum Icons {
   /**
    * Returns SVG element for given question. Note that callers need to size this element using
    * Tailwind classes like any other element.
+   *
+   * <p>public static SvgTag questionTypeSvg(QuestionType type) { Icons icon; switch (type) { case
```
swatkat1, line 473: I plan to remove this method fully

browser-test/image_snapshots/admin_program_creation_test/open-question-search.png
gwendolyngoetz, line 1: The css I gave you probably had the wrong color. Can you adjust the color so the add button maintains the existing color scheme. It should match the create new button.

It's a little weird that the screen 1 tab on the far left now has the blue high light. It should have that. It's just odd the deleted color isn't the blue.

server/app/views/admin/questions/QuestionsListView.java
swatkat1, line 1: from L:308 to L 313- Can also be converted css classes, but I am not sure what w-full or my-4 would be.

server/app/assets/stylesheets/styles.css
```diff
@@ -153,4 +153,37 @@
     cursor: default;
   }
 
+  .question-bank-add-button {
```
gwendolyngoetz, line 156: Can you add a comment for the purpose of these classes are to help reduce the page load size cause by duplication of all these on each question. You can remove the line by line comments.

server/app/views/components/UseTag.java
```diff
@@ -0,0 +1,9 @@
+package views.components;
+
+import j2html.tags.ContainerTag;
+
+public final class UseTag extends ContainerTag<UseTag> {
```
gwendolyngoetz, line 5: Add some javadoc with info this tag. It's custom cause I didn't find it in j2html and the SvgTag was custom. Info on `use` can be found here https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/use

server/app/views/components/Icons.java
```diff
@@ -525,13 +540,81 @@ public static SvgTag questionTypeSvg(QuestionType type) {
...
+        break;
+      case CHECKBOX:
+        icon = Icons.CHECKBOX;
+        break;
+      case CURRENCY:
+        return svg(Icons.CURRENCY)
+            .attr("fill", "none")
+            .attr("stroke-linecap", "round")
+            .attr("stroke-linejoin", "round")
+            .attr("stroke-width", "2");
```
gwendolyngoetz, line 560: This should just set the icon type and not return an svg

server/app/views/components/Icons.java
```diff
@@ -525,13 +540,81 @@ public static SvgTag questionTypeSvg(QuestionType type) {
...
+        break;
+      case ENUMERATOR:
+        icon = Icons.ENUMERATOR;
+        break;
+      case STATIC:
+        return svg(Icons.ANNOTATION)
+            .attr("fill", "none")
+            .attr("stroke-linecap", "round")
+            .attr("stroke-linejoin", "round")
+            .attr("stroke-width", "2");
```
gwendolyngoetz, line 593: This should just set the icon type and not return an svg

server/app/views/admin/questions/QuestionsListView.java
swatkat1: done

server/app/views/components/Icons.java
```diff
@@ -525,13 +540,81 @@ public static SvgTag questionTypeSvg(QuestionType type) {
...
+      case TEXT:
+        icon = Icons.TEXT;
+        break;
+      case PHONE:
+        icon = Icons.PHONE;
+        break;
+      default: // fall through
+        icon = Icons.UNKNOWN;
+    }
+    return svgLink(icon);
```
gwendolyngoetz, line 603:
```suggestion
    return new SvgTag().with(new UseTag().attr("href", "#svg-link-" + icon.name().toLowerCase(Locale.ROOT)));
```

This should be the only return for this method. Since nothing else uses the methods `svgLink` or `use` we can just inline this and delete those two new methods I added.

Although since we need to build the id in two places having a method that takes the icon and returns the id would be good to have.

```
private String makeSvgLinkId(Icon icon) { 
  return String.format("svg-link-%s", icon.name().toLowerCase(Locale.ROOT));
}
```

Keep in mind you only need the `#` on the `use` not for the `id` attribute on the other method.

server/app/views/components/ProgramQuestionBank.java
```diff
@@ -188,6 +189,27 @@ private FormTag questionBankPanel() {
...
+                Icons.questionTypeSvgWithId(QuestionType.ENUMERATOR),
+                Icons.questionTypeSvgWithId(QuestionType.FILEUPLOAD),
+                Icons.questionTypeSvgWithId(QuestionType.ID),
+                Icons.questionTypeSvgWithId(QuestionType.NAME),
+                Icons.questionTypeSvgWithId(QuestionType.NUMBER),
+                Icons.questionTypeSvgWithId(QuestionType.RADIO_BUTTON),
+                Icons.questionTypeSvgWithId(QuestionType.STATIC),
+                Icons.questionTypeSvgWithId(QuestionType.TEXT),
+                Icons.questionTypeSvgWithId(QuestionType.PHONE),
+                Icons.questionTypeSvgWithId(QuestionType.NULL_QUESTION))
```
gwendolyngoetz, line 210: Can you create a program and manually verify all the different types still load correctly?

swatkat1: Done! works pretty well :) ever faster

server/app/views/components/Icons.java
```diff
@@ -525,7 +581,67 @@ public static SvgTag questionTypeSvg(QuestionType type) {
       default: // fall through
         icon = Icons.UNKNOWN;
     }
-    return svg(icon);
+    return new SvgTag()
+        .with(new UseTag().attr("href", "#svg-link-" + icon.name().toLowerCase(Locale.ROOT)));
+  }
```
gwendolyngoetz, line 586:
```suggestion
    return svg(icon).withId(makeSvgLinkId(icon));
  }
  
  /**
  * Generate the HTML ID for the icon
  */
  private String makeSvgLinkId(Icon icon) { 
    return String.format("svg-link-%s", icon.name().toLowerCase(Locale.ROOT));
  }
```

Sorry, I think I miscommunicated on the id. The one that is hidden, that is the actual SVG, needs an ID which can't have a `#` in the name. And it needs to be built the same with as the full SVG to maintain it's styling.

Below, where we create the link version we need the `#` before the id and it just has the `use` element.

server/app/views/components/Icons.java
```diff
@@ -525,7 +581,67 @@ public static SvgTag questionTypeSvg(QuestionType type) {
...
+        icon = Icons.TEXT;
+        break;
+      case PHONE:
+        icon = Icons.PHONE;
+        break;
+      default: // fall through
+        icon = Icons.UNKNOWN;
+    }
+    return new SvgTag()
+        .with(new UseTag().attr("href", "#svg-link-" + icon.name().toLowerCase(Locale.ROOT)));
```
gwendolyngoetz, line 644:
```suggestion
    return new SvgTag()
        .with(new UseTag().attr("href", "#" + makeSvgLinkId(icon)));
```

gwendolyngoetz (changes requested):

server/app/views/admin/questions/QuestionsListView.java
```diff
@@ -383,9 +378,9 @@ private Pair<DivTag, ImmutableList<Modal>> renderActiveOrDraftRow(
   private DivTag renderInfoCell(QuestionDefinition definition) {
     DivTag questionText =
         div()
-            .withClasses("font-bold", "text-black", "flex", "flex-row", "items-center")
+            .withClasses("question-bank-info-cell")
             .with(
-                Icons.questionTypeSvg(definition.getQuestionType())
+                Icons.questionTypeSvgWithId(definition.getQuestionType())
```
gwendolyngoetz, line 383:
```suggestion
                Icons.questionTypeSvg(definition.getQuestionType())
```

Adding this id here may result in a conflict on this page. Let's hold off on doing the `/admin/questions` list for a different PR.

gwendolyngoetz (changes requested):

server/app/views/admin/questions/QuestionsListView.java
gwendolyngoetz: After this I think we're good!

swatkat1: I updated. Thanks Gwen!

swatkat1: One question- But our goal was to improve the performance in the question list page, if we don't reuse the svg, won't it be against the effort?

avaleske (approved):

gwendolyngoetz (approved):

PR #10085, opened 2025-03-20
Updates Switch Pattern Matching For Errorprone 1

gwendolyngoetz:
### Description

Updates to patterns needed to support the latest version of Errorprone.

There may be some defaults that are gone. This is normal because all the enum cases would be covered by the pattern matching. 

#### General

### DIFF
--- a/server/app/auth/ClientIpResolver.java
+++ b/server/app/auth/ClientIpResolver.java
@@ -65,15 +65,10 @@ public final class ClientIpResolver {
    */
   public String resolveClientIp(PlayWebContext context) {
 
-    switch (clientIpType) {
-      case DIRECT:
-        return context.getRemoteAddr();
-      case FORWARDED:
-        return getClientIpFromRequest(context);
-      default:
-        throw new IllegalStateException(
-            String.format("Unrecognized ClientIpType: %s", clientIpType));
-    }
+    return switch (clientIpType) {
+      case DIRECT -> context.getRemoteAddr();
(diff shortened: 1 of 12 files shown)


server/app/auth/ClientIpResolver.java
```diff
@@ -65,15 +65,10 @@ public String resolveClientIp(Http.RequestHeader request) {
...
-      case FORWARDED:
-        return getClientIpFromRequest(context);
-      default:
-        throw new IllegalStateException(
-            String.format("Unrecognized ClientIpType: %s", clientIpType));
-    }
+    return switch (clientIpType) {
+      case DIRECT -> context.getRemoteAddr();
+      case FORWARDED -> getClientIpFromRequest(context);
+    };
```
dkatzz, line 71: **question:** Is there a reason we removed the default case (here and below)?

dkatzz (commented):
One question around removing the defaults in 2 cases, but otherwise LGTM!

server/app/auth/ClientIpResolver.java
dkatzz: Can you add this comment to the PR description: `There may be some defaults that are gone. This is normal because all the enum cases would be covered by the pattern matching`

dkatzz (approved):

server/app/auth/ClientIpResolver.java
gwendolyngoetz: Done

PR #10100, opened 2025-03-23
[External Program Cards] Use ProgramType on ProgramForm

EmiliaPaz:
### Description

External program cards feature will require introducing a new entry to `ProgramType`. To avoid having a `isExternalProgram` variable on `ProgramForm`, we can introduce a `programType` variable that is used to determine which program type is currently being used.

No functionality was changed.

#### General

### Issue(s) this completes

Part of #10185


EmiliaPaz:
* **#10100** <a href="https://app.graphite.dev/github/pr/civiform/civiform/10100?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/civiform/civiform/10100?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

browser-test/src/support/admin_programs.ts
```diff
@@ -1495,7 +1495,7 @@ export class AdminPrograms {
   }
 
   getCommonIntakeFormToggle() {
-    return this.page.locator('input[name=isCommonIntakeForm]')
+    return this.page.locator('input[id=common-intake-checkbox]')
```
gwendolyngoetz, line 1498: (blocking) Please look this up by it's accessible name using `this.page.getByRole(...)`

server/app/services/program/ProgramType.java
```diff
@@ -21,4 +21,11 @@ public enum ProgramType {
   public String getValue() {
     return dbValue;
   }
+
+  public static ProgramType fromValue(String value) {
+    if (value.equals("common_intake_form")) {
+      return COMMON_INTAKE_FORM;
+    }
+    return DEFAULT;
+  }
```
gwendolyngoetz, line 30:
```suggestion
  public static ProgramType fromValue(String value) {
    for (ProgramType programType : values()) {
        if (programType.value.equals(value)) {
            return programType;
        }
    }
    throw new IllegalArgumentException("Unknown value: " + value);
  }
```

(question) Do you expect `value` to ever be an empty string or a string value not part of the enum? If no, just loop through and throw as shown in the above sample. If yes, in what case(s) would this default  to `DEFAULT`?

server/app/controllers/admin/AdminProgramController.java
```diff
@@ -133,17 +133,16 @@ public Result create(Request request) {
             ImmutableList.copyOf(programData.getCategories()),
             ImmutableList.copyOf(programData.getTiGroups()),
             applicationSteps,
-            programData.getIsCommonIntakeForm()
-                ? ProgramType.COMMON_INTAKE_FORM
-                : ProgramType.DEFAULT);
+            ProgramType.fromValue(programData.getProgramType()));
```
gwendolyngoetz, line 136: (nonblocking) There are a log of calls to ProgramType.fromValue. Consider adding that as a method on programData to simplify the code.

gwendolyngoetz (changes requested):

server/app/forms/ProgramForm.java
```diff
@@ -115,12 +116,21 @@ public void setLocalizedShortDescription(String localizedShortDescription) {
...
+    this.programTypeValue = programTypeValue;
+  }
+
+  public ProgramType getProgramType() {
+    for (ProgramType programType : ProgramType.values()) {
+      if (programType.getValue().equals(programTypeValue)) {
+        return programType;
+      }
+    }
+    throw new IllegalArgumentException("Unknown ProgramType for: " + programTypeValue);
```
gwendolyngoetz, line 133: (blocking) This should be moved to be a static method on the ProgramType enum

gwendolyngoetz (changes requested):

server/app/forms/ProgramForm.java
EmiliaPaz: Changed to `ProgramForm` per [comment](https://github.com/civiform/civiform/pull/10100#discussion_r2023216787). 
I do agree that this method better belongs in `ProgramType` enum, however it makes the calls in `AdminProgramController.java` much nicer by moving it to `ProgamForm` (as seen on the last commit). I don't have strong opinions, so happy to change it back

EmiliaPaz (commented):
Picking this up, sorry for the delay between reviews. Thanks!

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

PR #10103, opened 2025-03-23
[SessionTimeout][part7]Browser tests for session timeout

codekarma:
### Description

Added browser tests for session timeout frontend changes.

#### General

#### User visible changes

#### New Features

### Issue(s) this completes


browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,266 @@
...
+  currentTime: number
+}
+
+type TimeoutType = 'inactivityWarning' | 'totalWarning' | 'inactivityTimeout'
+
+/**
+ * Gets the adjusted timeout data that accounts for clock skew
+ * This matches what the SessionTimeoutHandler uses internally
+ */
+async function getAdjustedTimeoutData(page: Page): Promise<TimeoutData | null> {
```
rockycodes, line 28: suggestion (blocking): Move this and other helper methods into a `support/session_timeout.ts` file. We generally try to keep test files just for tests and move helper methods into support files eg. https://github.com/civiform/civiform/blob/main/browser-test/src/support/admin_programs.ts

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,267 @@
...
+  await advanceToSessionInactivityWarning(page)
+  await page.waitForSelector(
+    '#session-inactivity-warning-modal div.usa-modal',
+    {
+      state: 'visible',
+      timeout: 20000,
+    },
+  )
+
+  await validateScreenshot(page, `${uiPrefix}session-inactivity-warning-modal`)
```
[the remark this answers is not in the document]
rockycodes: Thinking about this more, since the modals are identical for applicant, admin, and north star, you don't need a screenshot for each use case and one screenshot of each of the modals should be sufficient.

server/app/assets/javascripts/session.ts
```diff
@@ -59,10 +59,47 @@ export class SessionTimeoutHandler {
...
+  private static nextTimeoutTime: number | null = null
 
   static init() {
+    if (this.isInitialized) {
+      return
+    }
+
     void this.checkAndSetTimer()
     this.setupModalEventHandlers()
+    // Add listener for timechange event to support testing with mocked time
```
rockycodes, line 76: suggestion (non-blocking): Make it extra obvious this logic is used only for testing by factoring it out into a helper method and adding a method comment that it is used only for testing.

server/app/assets/javascripts/session.ts
```diff
@@ -59,10 +59,47 @@ export class SessionTimeoutHandler {
   private static hasTotalLengthWarningBeenShown = false
   /** Current active timeout timer */
   private static timer: number | null = null
+  /** Tracks if handler has been initialized */
+  private static isInitialized = false
+
+  /** Stores the next timeout action and time for testing */
```
rockycodes, line 65: nitpick:
```suggestion
  /** Used for testing. Stores the next timeout action and time. */
```

server/app/assets/javascripts/session.ts
```diff
@@ -123,39 +160,48 @@ export class SessionTimeoutHandler {
           this.showWarning(WarningType.INACTIVITY)
           this.hasInactivityWarningBeenShown = true
         },
+        type: WarningType.INACTIVITY,
```
rockycodes, line 163: question (non-blocking): Where are we using `timeout.type`? I'm not seeing it.

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  await expect(
+    page.locator('#session-inactivity-warning-heading'),
+  ).toContainText('Warning')
+  await expect(page.locator('#session-inactivity-description')).toContainText(
+    'Your session will expire soon due to inactivity. Would you like to extend your session?',
```
dkatzz, line 187: **thought:** Should we also test the case in which they press cancel here? Would that trigger logout? I guess that gets a bit more complicated with the flow, but maybe it can be done at the end?

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,266 @@
...
+  await validateScreenshot(page, `${uiPrefix}session-extended-success`)
+
+  // Test length warning
+  await advanceToSessionLengthWarning(page)
+  await page.waitForSelector('#session-length-warning-modal div.usa-modal', {
+    state: 'visible',
+    timeout: 20000,
+  })
+  await validateScreenshot(page, `${uiPrefix}session-length-warning-modal`)
+  await page.click('#session-length-warning-modal button:text("Cancel")')
```
dkatzz, line 203: **thought:** Could we test the logout case here too?

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    await loginAsTestUser(page)
+    await page.goto('/programs')
+
+    await verifySessionTimeoutBehavior({page})
+    // Test timeout
+    await advanceToSessionInactivityTimeout(page)
+    await page.waitForURL((url) => url.href.includes('logout'))
+  })
+
+  test('shows warning modals and logout for Admins', async ({page}) => {
```
dkatzz, line 245: **suggestion:** Logout isn't checked for this flow?

server/conf/application.dev-browser-tests.conf
```diff
@@ -31,3 +31,8 @@ name_suffix_dropdown_enabled = true
 # In the test environment we don't need to have the jobs running at the
 # default 5 second interval
 durable_jobs.poll_interval_seconds = 3600
+
+session_inactivity_warning_threshold_minutes = 10 # Warn the user 10 mins before the inactivity timeout
```
rockycodes, line 35: question (non-blocking): It doesn't hurt to set these here, but since they're not directly used in your test (unless I'm missing something) I think you don't need them. I'm pretty sure the test conf file will inherit from the dev conf file which inherits from the prod conf file.

rockycodes (changes requested):
👏 👏 👏 for figuring out these complex tests! I mostly have comments about organization of the tests and removing screenshots.

server/conf/application.dev-browser-tests.conf
codekarma: Unfortunately we need this to order all warnings happen before the timeouts as it makes the tests easier to write(trigger all warnings first and then test the timeout). If the timeout event happen before warning(which is what happens in default), It is not possible to test warning by our current approach of advancing time.

server/app/assets/javascripts/session.ts
codekarma: I used it for logging and using it for unit tests but since then I have changed the tests and removed logs from session.ts so removed the timeout.type.

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  advanceToSessionLengthWarning,
+  advanceToSessionTimeout,
+} from './support/session_timeout'
+
+test.describe('Session timeout warnings for Admin', () => {
+  test.beforeEach(async ({page, adminSettings}) => {
+    await setupSessionTimeout({page, adminSettings})
+    await loginAsAdmin(page)
+    await page.goto('/admin/programs')
+    await page.waitForLoadState('networkidle')
```
gwendolyngoetz, line 25:
```suggestion
    await waitForPageJsLoad(page)
```

(blocking) You'll be redirected to the correct page already, you just need to use our wait function. Also you should almost never need to use `networkidle` with Playwright per their docs.

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  test('shows inactivity warning modal, extends session, show session length warning modal and logout', async ({
+    page,
+  }) => {
+    await test.step('Show and verify inactivity warning', async () => {
+      await advanceToSessionInactivityWarning(page)
+      await page.waitForSelector(
+        '#session-inactivity-warning-modal div.usa-modal',
+        {state: 'visible', timeout: 20000},
+      )
```
gwendolyngoetz, line 36:
```suggestion
      await expect(page.locator('#session-inactivity-warning-modal div.usa-modal')).toBeVisible()
```

(blocking) Use the web first assertions instead of waitForSelector. Like `networkidle` there should almost never be a reason to manually set timeouts. Playwright will handle waiting and will timeout based on global defaults.

Anything with explicit timeout needs to be updated.

browser-test/src/support/session_timeout.ts
gwendolyngoetz, line 1: (blocking) This should follow the pattern used in the other support files making this a class. The index.ts file is a special legacy case which I have been slowing phasing out.

browser-test/src/support/session_timeout.ts
```diff
@@ -0,0 +1,228 @@
...
+      console.error('Failed to parse timeout cookie:', e)
+      return null
+    }
+  })
+}
+
+/**
+ * Helper function to mock the Date object and advance time
+ */
+async function mockDateAndAdvanceTime(page: Page, advanceMilliseconds: number) {
```
gwendolyngoetz, line 54:
```suggestion

```
(blocking) I would prefer that the [Playwright clock api](https://playwright.dev/docs/clock#introduction) be used instead of a custom solution.

browser-test/src/support/session_timeout.ts
```diff
@@ -0,0 +1,228 @@
...
+  const data = await getAdjustedTimeoutData(page)
+  if (!data) {
+    throw new Error('No timeout data found')
+  }
+
+  // Get current browser time
+  const browserNow = await page.evaluate(() => Math.floor(Date.now() / 1000))
+
+  // Get target time based on timeout type
+  const targetTime = data[timeoutType]
```
gwendolyngoetz, line 128:
```suggestion
  const targetTime = data[timeoutType]
```
You should be able to slot in the clock api here

browser-test/src/support/session_timeout.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export async function advanceToSessionLengthWarning(page: Page) {
+  return advanceToTimeout(page, 'totalWarning')
+}
+
+export async function advanceToSessionTimeout(page: Page) {
+  return advanceToTimeout(page, 'totalTimeout')
+}
+
+export async function setupSessionTimeout({
```
gwendolyngoetz, line 149:
```suggestion

```
(blocking) This function can be removed. Please use `await enableFeatureFlag(page, 'session_timeout_enabled')` to set the feature flags.

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  setupSessionTimeout,
+  verifyTimeoutModalBehavior,
+  verifyModalDismissal,
+  verifyAutoLogout,
+  advanceToSessionInactivityWarning,
+  advanceToSessionLengthWarning,
+  advanceToSessionTimeout,
+} from './support/session_timeout'
+
+test.describe('Session timeout warnings for Admin', () => {
```
gwendolyngoetz, line 20: (blocking, question) The only difference here compared to the other two application options appears to primarily be the addition of the screenshots. Can this reuse the same calls as the others.?

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      await validateScreenshot(
+        page.locator('#session-length-warning-modal div.usa-modal'),
+        'admin-session-length-warning-modal',
+      )
+    })
+
+    await test.step('Logout and verify', async () => {
+      await page.click('#session-length-warning-modal button:text("Logout")')
+      await page.waitForURL((url) => url.pathname === '/programs')
+      await expect(page.locator('a[href="/applicantLogin"]')).toHaveText(
```
gwendolyngoetz, line 82: (blocking) This selector needs to be more targeted. This will fail if there are more than one matches.

server/app/assets/javascripts/session.ts
```diff
@@ -77,7 +91,11 @@ export class SessionTimeoutHandler {
    * be shown again even if its time passes again.
    */
   private static checkAndSetTimer() {
+    console.log('checkAndSetTimer called')
```
gwendolyngoetz, line 94: (blocking) Please remove all console.log debugging

gwendolyngoetz (changes requested):

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  })
+})
+
+function defineApplicantTests(options: {isNorthStar?: boolean} = {}) {
+  const description: string = options.isNorthStar
+    ? 'with NorthStar UI'
+    : 'Session timeout warnings for Applicants'
+  const config = options.isNorthStar ? {tag: ['@northstar']} : {}
+
+  test.describe(`${description}`, config, () => {
```
gwendolyngoetz, line 47: (blocking) There's no need to wrap this in a function. It'll make removing it later when pre-northstar goes away more difficult.

gwendolyngoetz (changes requested):

browser-test/src/session_timeout.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    page,
+  }) => {
+    const sessionTimeout = new SessionTimeout(page)
+    await sessionTimeout.verifyModalDismissal()
+    await sessionTimeout.verifyAutoLogout()
+  })
+})
+
+// NorthStar UI
+test.describe('Session timeout warnings with NorthStar UI', () => {
```
gwendolyngoetz, line 61: This still needs to hav the northstar tag

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

PR #10128, opened 2025-03-24
Link to program overview page from eligibility message

### LINKED ISSUE #10117: Program Details link on eligibility message should go to program overview
Image

The "program details" in this link should take the user to the program overview page for that program.

watermandrewc:
### Description

In the hyperlink on the ineligible page in North Star UI, link to the program overview page instead of the external program URL.

#### General

As an admin, create a program with a gating eligibility condition. As an applicant, apply to the program with an answer that does not meet the eligibility criteria. On the ineligible page, "program details" should link to the program overview page, whether or not an external program URL was set by the admin.

### Issue(s) this completes

Fixes #10117

### DIFF
--- a/server/app/views/applicant/NorthStarApplicantIneligibleView.java
+++ b/server/app/views/applicant/NorthStarApplicantIneligibleView.java
@@ -95,7 +95,8 @@ public class NorthStarApplicantIneligibleView extends NorthStarBaseView {
 
     // Manually construct a hyperlink with a runtime href and localized string. The hyperlink will
     // be inserted into another localized string in the Thymeleaf template.
-    String linkHref = applicantRoutes.show(program.slug()).url();
+    String linkHref =
+        applicantRoutes.show(params.profile(), params.applicantId(), program.slug()).url();
     String linkText =
         params.messages().at(MessageKey.LINK_PROGRAM_DETAILS.getKeyName()).toLowerCase(Locale.ROOT);
     String linkHtml =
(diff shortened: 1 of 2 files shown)


server/app/views/applicant/NorthStarApplicantIneligibleView.java
```diff
@@ -95,11 +95,7 @@ public String render(Params params) {
 
     // Manually construct a hyperlink with a runtime href and localized string. The hyperlink will
     // be inserted into another localized string in the Thymeleaf template.
-    // TODO: Update this to point to the new northstar details page.
-    String linkHref =
-        program.externalLink().isEmpty()
-            ? applicantRoutes.review(params.profile(), params.applicantId(), program.id()).url()
-            : program.externalLink();
+    String linkHref = applicantRoutes.show(program.slug()).url();
```
rockycodes, line 98: question (maybe blocking): I think we'll want to use the [`show` method](https://github.com/civiform/civiform/blob/bed74a5e10b9d469aca15c6ec5ddf2b482d823cd/server/app/controllers/applicant/ApplicantRoutes.java#L86) that expects an applicant id since we should have one at this point, but I'm not positive. If we can't use that for everyone, I think we'll need to direct to the different routes based on whether or not we have an applicant id like we do [here](https://github.com/civiform/civiform/blob/bed74a5e10b9d469aca15c6ec5ddf2b482d823cd/server/app/views/applicant/ProgramCardsSectionParamsFactory.java#L238) to handle the TI use case.

Would you try applying as a TI on behalf of a client and see what happens?

rockycodes (changes requested):

server/app/views/applicant/NorthStarApplicantIneligibleView.java
watermandrewc: You're right; under the previous commit, it redirects to the program overview page in a new tab, but starting an application from that page results in starting an application for the TI, not for the client.

Under the new commit, it redirects to the program overview page, where the TI can start/edit the application for the client.

rockycodes (approved):

PR #10129, opened 2025-03-25
Support processing large PDFs

sanjanasandeep:
### Description

Split the pdf to json and post processing steps to accept requests in chunks. The restriction is mainly large output tokens leading to malformed jsons.


haltmar (approved):

haltmar (changes requested):
as discussed, let's revise the approach to only creating chunks when LLM fails to process.

Issue #9980, opened 2025-03-12
[Security] Restrict style source on our Content Security Policy


melanie-exygy:
#### Describe the solution presented in this issue.
To additionally protect against XSS, we should add the `style-src` directive to our Content Security Policy.  We can probably use a nonce to validate the source in the same way we do for scripts. 

#### Additional Context
[style-src directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src)

watermandrewc:
The new policy has been enabled in dev for a few weeks with no known issues; next steps are:
1. Set the policy on one or more of our staging instances, and if no known issues:
2. Set the policy for prod using an environment variable

From @gwendolyngoetz:
> Ideally we should put it behind a flag, but we'd have to do it differently from a feature flag since we need it in the config file itself.
> directives.style-src = ${?VARIABLE_NAME}
> The ${?VARIABLE_NAME} would only set style-src if that environment variable has a value. So we keep the default off and we can set the value in a deploy config environment variable.

watermandrewc:
Before we can follow either of the next steps above, there's a blocking issue with tests failing on Azure/CSP errors getting thrown for Azure deployments on the file upload page (e.g. see failing tests in https://github.com/civiform/civiform/pull/10751).

Root cause is an issue with an older version of the Azure blob storage library: https://github.com/Azure/azure-sdk-for-js/issues/13268, so these steps above for setting the new policy are blocked on [#10227](https://github.com/civiform/civiform/issues/2349).

cc @dkatzz

melanie-exygy:
> Before we can follow either of the next steps above, there's a blocking issue with tests failing on Azure/CSP errors getting thrown for Azure deployments on the file upload page (e.g. see failing tests in [#10751](https://github.com/civiform/civiform/pull/10751)).
> 
> Root cause is an issue with an older version of the Azure blob storage library: [Azure/azure-sdk-for-js#13268](https://github.com/Azure/azure-sdk-for-js/issues/13268), so these steps above for setting the new policy are blocked on [#10227](https://github.com/civiform/civiform/issues/2349).
> 
> cc [@dkatzz](https://github.com/dkatzz)

Couldn't we still set the policy for our AWS staging instance?

dkatzz:
FYI https://github.com/civiform/civiform/pull/10720 should fix the azure-related issues!

watermandrewc:
[...]

PR #10142, opened 2025-03-26
Remove all inline styles

watermandrewc:
### Description

Remove all existing instances of inline styles. Adds nonce to some <links> so that they will continue to load once our csp is updated.

Part of #9980

#### General

Should be manually tested in both NS and legacy UIs to make sure no CSS has been broken by the change. E.g. navigate with dev tools open on various pages and verify that no CSP errors show up in the console.


browser-test/src/support/civiform_fixtures.ts
```diff
@@ -104,12 +104,6 @@ export const test = base.extend<CiviformFixtures>({
   },
 
   page: async ({page, request}, use) => {
-    page.on('console', (msg) => {
```
watermandrewc, line 107: Removed this since I think there's some race condition that was causing this check to fail with loading the css stylesheets. When testing manually, CSP errors weren't appearing in the console, but in the test env, they were even though by the test screenshots, it seems to have loaded correctly.

browser-test/image_snapshots/northstar_ineligible_test/northstar-eligibility-msg.png
gwendolyngoetz, line 1: Broken styling

gwendolyngoetz (changes requested):
I'll look at it in more depth later or on Monday, but a quick scan of the changed images the eligibility box is missing styling.

browser-test/image_snapshots/northstar_ineligible_test/northstar-eligibility-msg.png
watermandrewc: Thanks, looks like the bullet point was missing. Fixed!

browser-test/src/support/civiform_fixtures.ts
gwendolyngoetz: I'm not seeing any particular issue. Can you provide a more specific example to repro?

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -479,24 +481,46 @@ protected DivTag renderProgramApplicationTitleAndProgressIndicator(
...
-                "w-1",
-                "rounded-full")
-            .withStyle("width:" + percentComplete + "%");
+                "w-auto",
+                "flex-grow"));
+
+    // List of divs representing number of progress steps. Used to render correct progress
+    // percentage on progress bar.
+    ArrayList<DivTag> innerProgressBars = new ArrayList<DivTag>();
+    // Render a set of divs with blue civiform backgrou d
```
gwendolyngoetz, line 501:
```suggestion
    // Render a set of divs with blue civiform background
```

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -479,24 +481,46 @@ protected DivTag renderProgramApplicationTitleAndProgressIndicator(
...
       boolean forSummary,
       Messages messages) {
-    int percentComplete = getPercentComplete(blockIndex, totalBlockCount, forSummary);
 
-    DivTag progressInner =
-        div()
-            .withClasses(
-                BaseStyles.BG_CIVIFORM_BLUE,
+    double numerator = forSummary ? blockIndex : blockIndex + 1;
+    double denominator = forSummary ? totalBlockCount : totalBlockCount + 1;
```
gwendolyngoetz, line 486: (blocking) These names no long match their purpose. Please rename to make it easier to follow.
(blocking) These don't need to be doubles. Make them int will simplify the casting later.

server/app/views/applicant/ApplicantLayout.java
```diff
@@ -479,24 +481,46 @@ protected DivTag renderProgramApplicationTitleAndProgressIndicator(
...
+    // Render a set of divs with blue civiform backgrou d
+    for (int i = 0; i < numerator; i++) {
+      ArrayList<String> innerProgressBarClasses = new ArrayList<>(progressClassesList);
+      innerProgressBarClasses.add(BaseStyles.BG_CIVIFORM_BLUE);
+      if (i == 0) {
+        innerProgressBarClasses.add("rounded-l-full");
+      }
+      if (i == (int) numerator - 1) {
+        innerProgressBarClasses.add("rounded-r-full");
+      }
```
gwendolyngoetz, line 510: See if you can deal with this using first/last child css pseudo elements instead.

server/conf/application.conf
```diff
@@ -358,6 +358,7 @@ play.filters {
     reportOnly = false
     nonce.enabled = true
     directives.script-src = ${play.filters.csp.nonce.pattern} "'strict-dynamic' https: 'unsafe-inline'"
+    directives.style-src = ${play.filters.csp.nonce.pattern} "'strict-dynamic' 'self'"
```
gwendolyngoetz, line 361: (blocking) Let's getting all the CSS changes in first as a separate clean up commit. Anything specific to enabling CSP needs to be a separate commit. I'd like to enable it in dev first for a bit before moving this toward production.

gwendolyngoetz (changes requested):

server/app/views/docs/ApiDocsView.java
```diff
@@ -214,7 +214,6 @@ private DivTag apiResponseSampleDiv(ProgramDefinition programDefinition) {
 
     apiResponseSampleDiv.with(
         pre(code(fullJsonResponsePreviewPretty))
-            .withStyle("max-width: 100ch;")
```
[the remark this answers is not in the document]
watermandrewc: Tailwind doesn't seem to support ch for width units, so I replaced with max-w-2/3; let me know if there's a better configuration to use.

browser-test/src/support/civiform_fixtures.ts
watermandrewc: Hmm, I'm no longer able to repro -- perhaps because there were a few trailing inline styles that I missed. Before, the tests were failing due to a console error complaining about failing to load a stylesheet due to the security policy, but I added this back in and tests are no longer failing.

server/app/views/components/AlertFragment.html
```diff
@@ -60,8 +60,7 @@
     <!--/* padding-left-205 indents the list */-->
     <ul
       th:if="${not #lists.isEmpty(alertSettings.additionalText)}"
-      class="padding-left-205"
-      th:style="'list-style-type: disc;'"
+      class="padding-left-205 list-style-type-disc"
```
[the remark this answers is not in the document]
watermandrewc: Done. Still needed to add some custom styling for usa-list since some tailwind configurations were causing the list-style-type to get automatically removed from ol and ul elements

server/app/views/applicant/ApplicantLayout.java
watermandrewc: Tried doing this, but it started getting messy, since we don’t want to round the border of the last child of the #progress-indicator, but instead we want to round the last one that has the blue background. We *could* add a parent container over the filled-in/blue progress indicators, but that feels like it’s getting a bit more unnecessarily complex than it already is.

server/conf/application.conf
watermandrewc: SGTM -- removed this for now and we can add in a follow-up PR!

browser-test/playwright.config.ts
```diff
@@ -35,6 +35,8 @@ export default defineConfig({
     video: process.env.RECORD_VIDEO === 'true' ? 'on-first-retry' : 'off',
     // Fall back support config file until it is removed
     baseURL: process.env.BASE_URL || BASE_URL, // 'http://civiform:9000'
+    // // Bypass CSP to prevent flaky tests due to stylesheets
```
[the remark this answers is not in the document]
watermandrewc: Nope, was playing around with this when trying to get tests to pass and forgot to revert the changes. Thanks for catching!

server/app/views/applicant/BlockSummaryFragment.html
```diff
@@ -13,7 +13,7 @@
       ></a>
     </div>
 
-    <ul class="zero-padding usa-list usa-list--unstyled">
+    <ul class="zero-padding usa-list--unstyled">
```
melanie-exygy, line 16: **question (non-blocking):** This change doesn't seem related to removing inline styles.  Should it be in this PR?

server/app/views/applicant/ApplicantProgramSummaryView.java
```diff
@@ -316,8 +316,7 @@ private DivTag renderQuestionSummary(
             "border-b",
             "border-gray-300",
             "flex",
-            "justify-between")
-        .withStyle("word-break:break-word");
+            "justify-between");
```
melanie-exygy, line 319: **suggestion (blocking):** Needs a replacement.  Tailwind has the `break-words` class.

server/app/views/NorthStarBaseView.java
```diff
@@ -62,7 +62,6 @@ protected ThymeleafModule.PlayThymeleafContext createThymeleafContext(
     context.setVariable("civiformImageTag", settingsManifest.getCiviformImageTag().get());
     context.setVariable("addNoIndexMetaTag", settingsManifest.getStagingAddNoindexMetaTag());
     context.setVariable("tailwindStylesheet", assetsFinder.path("stylesheets/tailwind.css"));
-    context.setVariable("uswdsStylesheet", assetsFinder.path("dist/uswds.min.css"));
```
melanie-exygy, line 65: **question (non-blocking):** Why do we need to remove this variable?

browser-test/image_snapshots/text_test/text-max.png
melanie-exygy, line 1: **issue (blocking):** The text isn't being broken-up anymore.

server/app/views/applicant/BlockSummaryFragment.html
watermandrewc: This change is so that when the list-style-type is added back to the usa-list, other instances won't have a disc show up where not desired (see https://github.com/civiform/civiform/pull/10142#discussion_r2021249100)

melanie-exygy: Ah, got it.  Thanks for explaining.

server/app/views/NorthStarBaseView.java
watermandrewc: We don't have to remove it, but it didn't appear to be used anywhere. Let me know if you'd rather me keep it in for this PR to reduce risk of breakage by removing it.

server/app/views/applicant/ApplicantProgramSummaryView.java
watermandrewc: Done -- I had to also add "break-all" since "break-words" wasn't working on its own (I think because it sets `overflow-wrap: break-word` instead of directly setting `word-break`).

browser-test/image_snapshots/text_test/text-max.png
watermandrewc: This should be fixed by the above break-words/break-all change

server/app/views/NorthStarBaseView.java
melanie-exygy: Got it.  I think it makes sense to remove it and clean things up.  Thanks!

melanie-exygy (approved):

gwendolyngoetz (approved):

gwendolyngoetz (approved):

PR #10144, opened 2025-03-26
[North Star] Send focus to recommended programs upon filtering programs

### LINKED ISSUE #9915: [a11y][WCAG 3.2.1]: Screen reader focus is not moving to Filters results region after applying filters
Observed Results:
--
After applying filters in "Programs" screen, screen reader focus is not moving to Filters output region and screen reader remains silent.

Screencast: Chromebook + Chromevox[here]( https://drive.google.com/file/d/1I9OW1f_02TkRtq0yfi7a4qbo3Q_T0OLa/view?usp=drive_link ) \
Screencast: Windows + NVDA[here](https://github.com/user-attachments/assets/1b4a29ba-4400-4679-8948-2a1d03d8a2d5)

User Impact:
-------
Users who rely on assistive technologies like screen reader will face difficulty if focus is not moving to output screen after activating any control

Expected Results:
--
After selecting any checkbox under "Available programs and services" region and hit "Apply selections" button, focus should move to the filters results region and should announce as " Programs based on your selections(1)"

Steps to Reproduce:
-------
0. Enable Chromevox using Ctrl+Alt+Z
1. Launch the URL[here]( https://ns.civiform.dev/programs )
2. Navigate to Checkbox controls at "Available programs and services - filter by category" region.
3. Select any checkbox among (Economic, General and Utilities) and hit "Apply selections" button
4. Now observe the screen reader verbalization after activating filter selections button

WCAG Details:
--
[WCAG 3.2.1][On-focus]( https://www.w3.org/WAI/WCAG22/quickref/?versions=2.0#on-focus )

Additional Information
--
N/A

Environment Details:
--
Machine: chromebook 13.3/Windows \
Browser: Version 132.0.6834.208 (Official Build) (64-bit) \
Screenreader: Chromevox/NVDA 2024.4.2135031 \
Bug Filling Time: 15 mins.

melanie-exygy:
### Description

After selecting any program filter checkbox on the applicant home page and hitting the "Apply selections" button, focus now moves to the "Programs based on your selections" heading and the screen reader announces that heading.  This lets screen reader users know that the focus has moved, whereas it was previously silent and focus did not go to the search results.

1. Make sure at least one program has category tags.
2. Go to the applicant home page.
3. Turn on the screen reader.
4. Select a program filter and click "Apply selections".
5. Focus will move to the "Programs based on your selections" heading and the screen reader will read that heading text.

### Issue(s) this completes

Fixes #9915


server/app/views/applicant/ProgramCardsSectionFragment.html
```diff
@@ -2,12 +2,23 @@
...
 >
   <div th:if="${showHeading}" class="grid-row usa-prose section-bottom">
-    <h3 th:id="${section.id().get()}" th:text="${section.title().get()}"></h3>
+    <h3
+      th:unless="${sectionType == 'RECOMMENDED'}"
+      th:id="${section.id().get()}"
+      th:text="${section.title().get()}"
+    ></h3>
+    <h3
+      th:if="${sectionType == 'RECOMMENDED'}"
```
melanie-exygy, line 16: **note:** I would have preferred to conditionally apply the `autofocus` and `tabIndex` attributes, but couldn't find a way to do that in Thymeleaf.

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
```diff
@@ -41,7 +41,8 @@ public enum SectionType {
     MY_APPLICATIONS,
     COMMON_INTAKE,
     UNFILTERED_PROGRAMS,
-    STANDARD;
+    STANDARD,
```
EmiliaPaz, line 44: What is the difference between standard, recommended and unfiltered programs?
My understanding is when no filter is applied, programs are on UNFILTERED_PROGRAMS. When a filter is applied, programs are on STANDARD. Now if the programs are matching, they are on RECOMMENDED.
I think we should add a comment for each type because it's a bit confusing.

server/app/views/applicant/ProgramCardsSectionFragment.html
EmiliaPaz: Would `th:autofocus="${sectionType == 'RECOMMENDED'}"` work? Haven't tried it, so no idea if it's valid

EmiliaPaz: nit-super optional: If previous comment doesn't work, having the 'if statement' first is easier to read (following if - else style)
```
<h3
      th:if="${sectionType == 'RECOMMENDED'}"
      th:id="${section.id().get()}"
      th:text="${section.title().get()}"
      tabindex="-1"
      autofocus
></h3>
<h3
      th:unless="${sectionType == 'RECOMMENDED'}"
      th:id="${section.id().get()}"
      th:text="${section.title().get()}"
></h3>
```
and I would even replace the unless to be ` th:unless="${sectionType == 'STANDARD'}"` because it makes it easier to read and showHeading means those are the only two options

EmiliaPaz (commented):
Thanks Melany! I wasn't able to reproduce it, for some reason pulling the branch give me "switch statement" errors. Branch may not be up to date with my local. Regardless, it looks good.

Just a couple comments

server/app/views/applicant/ProgramCardsSectionFragment.html
melanie-exygy: Unfortunately, your `th:autofocus` suggestion is exactly what I tried and I got a big Thymeleaf compilation error, so I guess they don't support it.  
Sure, I can put the `th:if` first.  As for replacing the `th:unless` with `th:unless=${sectionType == 'STANDARD'}`, I can't do that because that won't evaluate correctly.  It's basically equivalent to saying "if the section type is NOT standard".

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
melanie-exygy: Originally, there were just two types:  COMMON_INTAKE and STANDARD (which just means default).  Then, I've had to add other types to the point where now the only "standard" program section is the "Other programs and services" section (the programs that don't match any filters).  You're right that it's confusing now.  
I think I'll just rename STANDARD to be OTHER_PROGRAMS.  That way it matches the heading.  Sound ok?

server/app/views/applicant/ProgramCardsSectionFragment.html
EmiliaPaz: Ah, that's unfortunate. This looks good!

server/app/views/applicant/ProgramCardsSectionParamsFactory.java
EmiliaPaz: That sounds better. Although, what about a bit more descriptive:
- UNFILTERED_PROGRAMS
- FILTERED_PROGRAMS_MATCHING
- FILTERED_PROGRAMS_NON_MATCHING
(could be done in a separate PR to land the a11y fix)

melanie-exygy: I started changing the name and realized that the "STANDARD" section is also used on the upsell page.  So, I can't give it a specific name.  I'm renaming it to "DEFAULT", since that's clearer and adding a comment that this is the section type when that section doesn't require any special styles.

EmiliaPaz: sgtm, thanks for looking into it!

melanie-exygy: Done

EmiliaPaz (approved):
Thanks!

Issue #10061, opened 2025-03-19
Tracking: Add support for External Programs


EmiliaPaz:
Tracking issue for adding External Programs support

## Resources
- [Technical Design Doc](https://docs.google.com/document/d/17DDMTE_V8fJyB8xq1h3IlNs1g2m2B6M2DSrDJQKUYzs/edit?usp=sharing)
- [Mocks](https://www.figma.com/design/wGYsNhmRoKATfcZ7br3kgI/External-Program-View-Iteration?node-id=1-20&p=f&t=TDYa9mROncNR3tkO-0)

## Tasks

To be completed once TDD is approved

### UX
- [x] Applicant side mocks
- [x] Admin panel mocks
- [ ] Finalize strings with UX/PM

### Setup
- [ ] Add launch flag

### QA and bug fixes

msprenke:
I see we have this listed in production under EXTERNAL_PROGRAM_CARDS_ENABLED
(NOT FOR PRODUCTION USE) Enable showing external program cards on North Star applicant UI.

When do you anticipate it being complete? @EmiliaPaz

dkatzz:
> I see we have this listed in production under EXTERNAL_PROGRAM_CARDS_ENABLED (NOT FOR PRODUCTION USE) Enable showing external program cards on North Star applicant UI.
> 
> When do you anticipate it being complete? [@EmiliaPaz](https://github.com/EmiliaPaz)

This should actually be ready for use - looks like we just haven't removed that messaging yet - @rockycodes FYI

rockycodes:
Looks like the last step was to coordinate the announcement of the feature, so I think you're right! https://github.com/civiform/civiform/issues/10363

@shreyachatterjee00 can you confirm?

msprenke:
thank you!

msprenke:
I noticed i'm not able to switch a program to external.  Its' greyed out as an option.  Was this by design or is there something I need to do to allow me to switch a program to external.  The program was using a wonky redirect inside a static questions to tell people to click on it and go to seattle.gov.


Cheers,



Matt



Matthew Sprenke

Business Systems Analyst

Affordable Seattle Team | Information Technology Department

City of Seattle

Learn more about Affordable Seattle<https://seattle.gov/affordable> and CiviForm<https://civiform.us/>



[...]

PR #10157, opened 2025-03-27
Use USWDS on checkbox and radio options in ProgramFormBuilder (part 2)

EmiliaPaz:
### Description

In the program create / edit pages, use USWDS styling only for:
- Common intake radio button
- TI List checkboxes

This is done to have all checkboxes and radio buttons in the create / edit page to use USDWS styling, as explained on #10147.

This is only a UI change and shouldn't have any impact on functionality.

#### General

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [x] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)

#### User visible changes

- [x] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [x] Manually tested at 200% size
- [x] Manually evaluated tab order

### Instructions for manual testing

1. Log in as an admin
2. Click on create a program
3. Verify form fields look as expected

### Issue(s) this completes

Part of #10147

PR #10161, opened 2025-03-27
Show the required indicator in the first application step on program form iff program is not common intake 

### LINKED ISSUE #10165: Show the required indicator in the first application step on program form iff program is not common intake
Pre-screeners have certain fields disabled that are not applicable (#9508).  If the field was required, the required label is removed.

There are two bugs:
1. If the pre-screener checkbox is unselected, the required label is not added back to the required application step #1
2. When editing a pre-screener, the required label appears on application step #1 even when field disabled.

EmiliaPaz:
### Description

Pre-screeners have certain form fields disabled. This PR fixes two related bugs:
1. If the pre-screener checkbox is unselected, the required label is added back to the required application step 
2. When editing a pre-screener, the required label should be hidden on application step 

This PR also adds helper methods for:
- Updating the disabled state given a fieldSelector or fieldElement, in `server/app/assets/javascripts/admin_programs.ts`
- Verifying the form field disabled state in `browser-test/src/support/admin_programs.ts`
These methods will be very helpful when adding external programs to the form (#10061), since they will disable other form fields

#### General

#### User visible changes

1. Login as an admin
3. Click on 'Create new program'
    => Verify all fields are enabled
    => Verify step 1 has '*'
4. Check 'set program as pre-screener'
    => Verify long program description (if NS enabled), program categories (if filtering enabled) and step 1 are disabled
    => Verify step 1 doesn't have '*'
5. Uncheck 'set program as pre-screener'
    => Verify all fields are enabled
    => Verify step 1 has '*'
6. Save program
7. Edit program
    => Verify long program description (if NS enabled), program categories (if filtering enabled) and step 1 are disabled
    => Verify step 1 doesn't have '*'
 
### Issue(s) this completes

Fixes #10165


EmiliaPaz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/civiform/civiform/10161?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10161** <a href="https://app.graphite.dev/github/pr/civiform/civiform/10161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/civiform/civiform/10161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10157** <a href="https://app.graphite.dev/github/pr/civiform/civiform/10157?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

server/app/views/admin/programs/ProgramFormBuilder.java
```diff
@@ -382,7 +382,8 @@ static DivTag buildApplicationStepDiv(
             .setMarkdownSupported(true)
             .setValue(descriptionValue);
 
-    if (indexPlusOne.equals("1")) {
+    Boolean isRequired = indexPlusOne.equals("1") && !isCommonIntakeForm;
```
EmiliaPaz, line 385: Functionality change (2/2): before, '*' would appear when editing a pre-screener

This is tested by including whether the span containing '*' is disabled in `expectFormFieldDisabled`

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -24,88 +24,124 @@ class AdminPrograms {
...
-          applicationStepOneDiv?.querySelectorAll('span')
-        requiredIndicators?.forEach((indicator) => {
+      const requiredIndicators = applicationStepOneDiv?.querySelectorAll('span')
+      requiredIndicators?.forEach((indicator) => {
+        if (commonIntakeCheckbox.checked) {
           indicator.classList.add('hidden')
-        })
-      }
+        } else {
+          indicator.classList.remove('hidden')
```
EmiliaPaz, line 78: Technically, we should add 'optional' to match the styling of other fields.
However, that adds more complexity since we would need to change the actual text. I think this is fine, since not having '*' implies optional

(separately, wish we didn't have the word optional since not having '*' implies that.. but that is a separate conversation that can be addressed with the admin ui redesign)

browser-test/src/support/admin_programs.ts
```diff
@@ -219,17 +244,76 @@ export class AdminPrograms {
...
+  /**
+   * Verifies whether specific form fields are properly disabled or enabled based on expected state.
+   *
+   * @param formField - The specific form field type to verify (from FormField enum)
+   * @param isDisabled - Boolean indicating whether the field should be disabled (true) or enabled (false)
+   *
+   * @throws Will throw an error if the elements' states don't match the expected disabled/enabled state
+   * @throws Will throw an error if an invalid or unsupported form field type is provided
+   */
+  async expectFormFieldDisabled(formField: FormField, isDisabled: boolean) {
```
rockycodes, line 256: suggestion (blocking): Instead of using this singular method and passing in a boolean, create two methods to separately handle the enabled and disabled cases. I'm suggesting this for a few reasons:
- The method name `expectFormFieldDisabled` is confusing when we're actually expecting the field to be enabled.
- I learned recently that we want to avoid if/else statements in testing helper functions because they're harder to reason about and can have unintended consequences.
- We generally try to avoid passing booleans as params (with varying success as you can tell 😂 ), especially in functions that will have many call sites, because they always require a param hint to know what the boolean is meant to signify and it's easy to forget to add that.

There's some logic in here that might be a pain to duplicate (eg. the for-loop for application steps) so you could make a singular helper method that takes in a boolean passed from the two public methods and does the if/else thing as a compromise. Up to you.

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -24,88 +24,124 @@ class AdminPrograms {
...
+        /* shouldDisable= */ commonIntakeCheckbox.checked && northStarUiEnabled,
       )
 
-      const applicationStepTitles = document.querySelectorAll(
-        'input[id^="apply-step"]',
-      )
-      const applicationStepDescriptions = document.querySelectorAll(
-        'textarea[id^="apply-step"]',
+      // Application steps
+      this.updateTextFieldsDisabledState(
```
rockycodes, line 62: suggestion (blocking): Is it possible to combine `updateTextFieldsDisabledState` and `updateTextFieldDisableState`? They are so similarly named that it tripped me up for a minute. Alternatively, would you update the name of one of them so it's more unique?

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -24,88 +24,124 @@ class AdminPrograms {
...
-        const requiredIndicators =
-          applicationStepOneDiv?.querySelectorAll('span')
-        requiredIndicators?.forEach((indicator) => {
+      const requiredIndicators = applicationStepOneDiv?.querySelectorAll('span')
+      requiredIndicators?.forEach((indicator) => {
+        if (commonIntakeCheckbox.checked) {
           indicator.classList.add('hidden')
-        })
-      }
+        } else {
```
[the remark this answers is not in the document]
rockycodes: Praise: Thank you for fixing this!

server/app/assets/javascripts/admin_programs.ts
rockycodes: I 100% agree with you, it's fine to not add "optional" in and I would love to get rid of that text entirely when we re-do this page.

server/app/assets/javascripts/admin_programs.ts
EmiliaPaz: I can see the confusion. I think it's valuable to have them separate, so the test is easier to read. Updated the name of them so they reflect the parameter type

browser-test/src/support/admin_programs.ts
EmiliaPaz: That makes sense, thanks for the explanation. The method ddivided on disabled and enabled are actually easier to read :)

server/app/assets/javascripts/admin_programs.ts
```diff
@@ -24,88 +24,124 @@ class AdminPrograms {
...
+    checkboxes.forEach((checkbox) => {
+      const checkboxElement = checkbox as HTMLInputElement
+      if (shouldDisable) {
+        checkboxElement.disabled = true
+        checkboxElement.checked = false
+      } else {
+        checkboxElement.disabled = false
+      }
+    })
+  }
```
EmiliaPaz, line 143: IMO is better to keep the `shouldDisable` boolean than separate on two methods for these ones because it's easier to see what is the enabled and disabled state.
However, I don't feel strongly so can be changed to `enableUSWDSCheckboxes` and `disableUSWDSCheckboxes` (similarly the other methods)

rockycodes (dismissed):

rockycodes (approved):

PR #10164, opened 2025-03-28
[SessionTimeout][part8] Increase gap between timeout events in browser test and handle context destruction.

codekarma:
### Description

These changes simply increase the time between the timeout events. The session timeout browser test currently test for 4 events to happen in this order: session inactivity warning, session length warning, inactivity timeout and then session length timeout. The test also rely on Playwright's clock API to advance time. The tests were flaking because the session length warning timestamp was too close to inactivity timeout and in traces, it can be seen that the timeout was processed right before the session length warning dialog was fully displayed. Also, when advancing the time to a timeout could sometime lead to user logout before we actually send the timechange event. This was result in an error where the page context is destroyed(due to navigation) at the time of sending timechange event. This change skip sending timechange event in this case.

#### General

### Issue(s) this completes

### DIFF
--- a/browser-test/src/support/session_timeout.ts
+++ b/browser-test/src/support/session_timeout.ts
@@ -224,19 +224,28 @@ export class SessionTimeout {
     // Run for a small amount of time to ensure all timers are processed
     await this.page.clock.runFor(100)
 
-    // Dispatch a custom event to notify the application that time has changed
-    // This helps trigger any listeners that might be waiting for time changes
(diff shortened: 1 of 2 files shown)


gwendolyngoetz (changes requested):
I'll look at this Monday. I'd like to find a better approach. A destroyed page context is indicative of deeper issues.

gwendolyngoetz:
@tallulahkay do we still need this?

Issue #10169, opened 2025-03-31
The short descript is not getting shown in the program cards but the long one is.


swatkat1:
**Describe the bug**
In a program card, we want to show the program's short descript for the public but currently only the long descript is being shown.  

**To Reproduce**

1. Create a program and add two different texts for program short description and program's long description.
2. Click on publish
3. View the program as an applicant.
4. See the long description being shown the program cards. 

**Expected behavior**
The short description to should be shown to the applicants. 

**Screenshots**

Program details page. The program short description is "Buy fruits and veggies"

![Image](https://github.com/user-attachments/assets/95699bc7-fa0b-40e5-86f0-8732341f9817)

Program long description (which is optional is)

![Image](https://github.com/user-attachments/assets/d0c232ff-00f5-4fc4-b9c8-b937682f1857)

After publish:
"buy icecreams" is shown to the applicants. 
![Image](https://github.com/user-attachments/assets/3b550348-3ecf-47df-811d-50cb2071dc00)

**Additional context**

melanie-exygy:
@swatkat1, I believe the short description is only used in the cards with the North Star flag on.  @shreyachatterjee00 , can you confirm?

swatkat1:
though the messaging is a little confusing. short description is only for NS. Closing for now.

PR #10177, opened 2025-03-31
Adding program's long description to program pdf export

### LINKED ISSUE #10126: Form PDF download to include program description and sections

In the top right section of a form, CiviForm admins can download a PDF. This is for review and translation purposes. The PDF does not include new information that will be part of the North Star update. 

Information not included in the PDF includes: 

- New short description.
- How to apply Step title and description
- Custom eligibility message
- Submission confirmation message
- Custom status and status message 
- Program tile description

PDFs should include this information as well for one stop shop review and for translation submissions. 

If some of these things are easier to add than other, I think we should add the most easy items as possible. This feature will still be successful even if it doesn't include all the new items listed above. 

#### What priority is this?

P2

#### Associated civic entity

Seattle

#### User Group

If this request or piece of feedback comes from a resident, add the label 'Resident feedback'.
If this request or piece of feedback comes from a CBO or TI, add the label 'TI'.
If this request or piece of feedback comes from an admin, add the label 'Program Feedback'.

swatkat1:
### Description

Adding program's long description to program pdf export

## Release notes

Adding program's long description to program pdf export

#### General

### Issue(s) this completes

Fixes #10126

### DIFF
--- a/server/app/services/export/PdfExporter.java
+++ b/server/app/services/export/PdfExporter.java
@@ -320,6 +320,10 @@ public final class PdfExporter {
               SMALL_GRAY_FONT));
       document.add(
           new Paragraph(
+              "Admin long description: " + programDefinition.localizedDescription().getDefault(),
+              SMALL_GRAY_FONT));
+      document.add(
+          new Paragraph(
               "Time of export: "
                   + timeCreated.format(DateTimeFormatter.ofPattern("yyyy-MM-dd hh:mm a")),
               SMALL_GRAY_FONT));
--- a/server/test/services/export/PdfExporterTest.java
+++ b/server/test/services/export/PdfExporterTest.java
@@ -397,10 +397,12 @@ public class PdfExporterTest extends AbstractExporterTest {
     assertThat(linesFromPdf.get(3))
         .contains(
             "Admin short description: " + programDef.localizedShortDescription().getDefault());
-    assertThat(linesFromPdf.get(4)).contains("Time of export:");
-    assertThat(linesFromPdf.get(5)).isEqualTo("Origin of export: http://localhost:9000");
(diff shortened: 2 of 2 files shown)


server/app/services/export/PdfExporter.java
```diff
@@ -318,6 +318,10 @@ private byte[] buildProgramPdf(
               "Admin short description: "
                   + programDefinition.localizedShortDescription().getDefault(),
               SMALL_GRAY_FONT));
+      document.add(
+          new Paragraph(
+              "Admin long description: " + programDefinition.localizedDescription().getDefault(),
```
dkatzz, line 323: **question:** I think the long description is optional. Could we have this added only if it is set?

swatkat1: I don't think localizedDescription is null by default. But showing it as empty, when not set, will let the admin know the value is missing when pulling the document for translation. 

16.sql--> 
update programs set localized_description = ('{"en_US" : "' || description || '"}')::jsonb where localized_description is null;

dkatzz: Can we check if its empty and hide it if so? I think we support it being unset, since the short description will be used in that case.

swatkat1: It doesn't use the short description. Infact, it is the text that goes on the program cards. Only today I found out! 
In the below program, the long description is empty as I didn't set it while creating it. But the reason for adding it- because it needs translation since it's a localized text. If it isn't captured here- program cards will remain untranslated. 
image
I filed this today - https://github.com/civiform/civiform/issues/10169

swatkat1: Will getting the product's input help with the review? Please feel free to involve other folks. I will also talk to Miguel in Seattle.

Issue #9914, opened 2025-03-07
[a11y][WCAG 2.4.3]: Linear navigation focus is not visible on checkbox controls


dkatzz:
Observed Results:
--
While navigating in Linear Navigation (Search + right/left) arrow keys, screen reader foucs is not visible on the checkbox controls rather it is displaying somewhere else in the page.

Screencast: Chromebook + Chromevox[here]( https://drive.google.com/file/d/1q6IQ53AYKNJU-R3cTKZ8VzSLOjwyoKVc/view?usp=drive_link ) \
Screencast: Windows + NVDA[here]( https://screencast.googleplex.com/cast/NjcyMjI2NzUyMTIyMDYwOHw3YmM0ZGYyMi05YQ )

User Impact:
-----
Users who are low vision and depend on screen readers will face difficulty or get confused if focus is not on the controls while navigating.

Expected Results:
--
While navigating using linear navigation (Search+right/left) arrow keys, checkbox controls under "Available programs and services" region,screen reader focus should visible on the checkbox control and name of the checkbox control individually

Steps to Reproduce:
-------
0. Enable Chromevox using Ctrl+Alt+Z
1. Launch the URL[here]( https://ns.civiform.dev/programs )
2. Using Linear Navigate(Search +left/right arrow) over the Checkbox controls at "Available programs and services - filter by category" region.
3. Observe the linear navigation focus on checkbox controls

WCAG Details:
--
[WCAG 2.4.3][Focus-Order]( https://www.w3.org/WAI/WCAG22/quickref/?versions=2.0#focus-order )

Additional Information
--
N/A

Environment Details:
--
Machine: chromebook 13.3/Windows \
Browser: Version 132.0.6834.208 (Official Build) (64-bit) \
Screenreader: Chromevox/NVDA 2024.4.2135031 \
Bug Filling Time: 15 mins.

dkatzz:
Second screencast:

https://github.com/user-attachments/assets/09fe1cae-e884-4755-a846-5dad139b73ad

melanie-exygy:
Tested on NVDA and had a similar experience, with a small difference depending on whether I was using the up/down arrows or form navigation (F and Shift-F).  I believe with NVDA, most users would be using form navigation.  With the arrow keys, the red focus box only appear after a second click.  With form navigation, the red focus box doesn't appear at all.

Note that I tested this on a checkbox question within the application form and had the same experience.

[...]

PR #10071, opened 2025-03-19
[A11y] update styles for radio and checkbox focus in high contrast mode

tallulahkay:
### Description

Some custom scss styling was suppressing the checkbox in high contrast mode. I updated the styling so that it adds an outline rather than a border to the component on focus.

#### General

Read the full guidelines for PRs [here](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#creating-a-pull-request)

- [x] Added the correct label: < feature | enhancement | bug | under-development | dependencies | infrastructure | ignore-for-release | database >
- [x] Assigned to a specific person, `civiform/developers`, or a [more specific round-robin list](https://github.com/civiform/civiform/wiki/Technical-contribution-guide#adding-reviewers)
- [ ] Added an additional reviewer from outside your organization as FYI (if the primary reviewer is in the same organization as you)
- [x] Removed the release notes section if the title is sufficient for the release notes description, or put more details in that section.
- [ ] Created unit and/or browser tests which fail without the change (if possible)
- [x] Performed manual testing (Chrome and Firefox if it includes front-end changes)
- [ ] Extended the README / documentation, if necessary. For user-facing features, consider updating [the user docs](https://github.com/civiform/docs). For "under-the-hood" changes or things more relevant to developers, consider updating [the dev wiki](https://github.com/civiform/civiform/wiki).
- [ ] Ensured PII wasn't added to any new logs, unless it was guarded by `isDevOrStaging`

#### User visible changes

- [ ] Followed steps to [internationalize](https://github.com/civiform/civiform/wiki/Internationalization-%28i18n%29#application-strings) any new strings
  - [ ] Added context strings to new [messages](https://github.com/civiform/civiform/blob/main/server/conf/i18n/messages)
  - [ ] Didn't use a message in applicant facing code that isn't translated yet (unless behind a flag)
- [ ] Wrote browser tests using the [validateAccessibility](https://sourcegraph.com/github.com/civiform/civiform/-/blob/browser-test/src/support/index.ts?L437:14&subtree=true) method
- [ ] Made equivalent changes in Thymeleaf for applicant-facing features and changes (or created an issue in the [Northstar epic](https://github.com/orgs/civiform/projects/1/views/94) to track that it's missing from Thymeleaf)
- [x] Tested on mobile view. See [mobile device mode](https://developer.chrome.com/docs/devtools/device-mode/)
- [ ] Manually tested at 200% size
- [ ] Manually evaluated tab order
[...]

Issue #8491, opened 2024-09-04
On radio buttons, focus outline is around the button only rather than parent element


ssandbekkhaug:
On radio buttons, focus outline is around the button only rather than parent element.

[Slack Message](https://civiform.slack.com/archives/C01QKN2UNMA/p1725469054941769?thread_ts=1725468051.919329&cid=C01QKN2UNMA)

Focus should look like this:

![image](https://github.com/user-attachments/assets/49757aee-f9f5-4405-b9a1-dde719773f86)

Reported by Melanie

ssandbekkhaug:
![radio](https://github.com/user-attachments/assets/4ff859b6-1560-4905-b120-faef252da099)

PR #10182, opened 2025-04-01
[North Star] Making screen reader linear navigation outline visible on applicant home page checkboxes

melanie-exygy:
### Description

When the screen reader is on and the user is NOT navigating the checkboxes by using "Tab", but, rather, by using other form navigation controls which vary by screen reader, a navigation outline surrounds the element that the screen reader is currently reading.  Note that this outline is not the focus outline.  It is a different color and width, depending on the screen reader.  

The USWDS checkbox component is a custom component.  It shifts the native checkbox outside the viewport and uses its own customized element as the visible checkbox.  Since the screen reader navigation outline is based on the native element, it is also outside the viewport and is missing from view.

This PR places the native checkbox input element around the custom one, but sets the opacity to 0 so that it is only visible during screen reader linear navigation.  This PR limits itself to fixing the program filter checkboxes on the applicant home page since that is what was discovered in testing.  However, we will need a follow-up PR to fix this same problem for the checkbox questions within the application form.

Since this would require access to the NVDA screen reader, I have provided a video here (using NVDA) in place of manual testing:

https://github.com/user-attachments/assets/54ac3333-35ae-4108-9023-945e517e3a3e

### Issue(s) this partially completes

Partially completes #9914


EmiliaPaz (approved):
Fix lgtm!
However, I'm wondering if we are adding too much customization to the USWDS checkbox. USWDS checkbox default behavior highlights the "square" on tab navigation. This was overriden on #8491, which caused other a11y customizations like #10071 and seems more will be needed for similar buttons.
Not a fan of adding customization to a button, because then more and more is needed.. but we already opened that box for this button so this is good

melanie-exygy:
> > Fix lgtm! However, I'm wondering if we are adding too much customization to the USWDS checkbox. USWDS checkbox default behavior highlights the "square" on tab navigation. This was overriden on #8491, which caused other a11y customizations like #10071 and seems more will be needed for similar buttons. Not a fan of adding customization to a button, because then more and more is needed.. but we already opened that box for this button so this is good
> 
> I agree with this concern! Can we bring it up in the wg-ux-north-star chat? I think changing it back to the basic button makes sense but would want to get kyle and shreya's input

Before this, the only other checkbox customization was putting the focus outline around the whole label rather than just the checkbox square.  I agree that it's turned out to be a bit of a hassle.  We did it because it was recommended by Sijia and requested by Seattle.  See the UX decision here:  #5947.  It also seems like a better accessibility choice to make the focus outline match the clickable area and making the focus outline larger just makes it easier for people with low vision to see.

That said, it's a divergence from USWDS and maybe we should just trust their decision for the focus outline.  I'll bring it up again in our Slack channel, for reconsideration.

As for the customization in this particular PR, I've brought it up for discussion here:  [Slack conversation](https://civiform.slack.com/archives/C01QKN2UNMA/p1743531390171779).  As I said there, I'm leaning toward filing a bug with USWDS, but please do chime in if you think differently.

tallulahkay:
I'm noticing that the outline falls out of sync with the focus when you go backwards:

https://github.com/user-attachments/assets/10336be2-4aee-41e5-a7d1-6e296067f121

I am using mac's voiceover though, not NVIDIA. Not sure if it's an issue that needs fixing or maybe a separate issue - seems out of scope with this issue. I'm fine to approve and would still like to have the conversation about using USWDS styling, but agree we'd have to keep this customization either way

tallulahkay (approved):

melanie-exygy:
> I'm noticing that the outline falls out of sync with the focus when you go backwards:
> 
>  Screen.Recording.2025-04-01.at.12.05.16.PM.mov 
> I am using mac's voiceover though, not NVIDIA. Not sure if it's an issue that needs fixing or maybe a separate issue - seems out of scope with this issue. I'm fine to approve and would still like to have the conversation about using USWDS styling, but agree we'd have to keep this customization either way

I tried reproducing the issue, but couldn't.  What command are you using?  "Tab" and "Shift-Tab" or something else?  Which browser are you on?  Also, do you have the Keyboard Navigation setting turned on for your Mac? (System Settings -> Keyboard -> toggle "Keyboard navigation" on).

tallulahkay:
> > I'm noticing that the outline falls out of sync with the focus when you go backwards:
> > Screen.Recording.2025-04-01.at.12.05.16.PM.mov
> > I am using mac's voiceover though, not NVIDIA. Not sure if it's an issue that needs fixing or maybe a separate issue - seems out of scope with this issue. I'm fine to approve and would still like to have the conversation about using USWDS styling, but agree we'd have to keep this customization either way
> 
> I tried reproducing the issue, but couldn't. What command are you using? "Tab" and "Shift-Tab" or something else? Which browser are you on? Also, do you have the Keyboard Navigation setting turned on for your Mac? (System Settings -> Keyboard -> toggle "Keyboard navigation" on).

I'm using ctrl+alt+arrow keys with mac voiceover on chrome

melanie-exygy:
> > > I'm noticing that the outline falls out of sync with the focus when you go backwards:
> > > Screen.Recording.2025-04-01.at.12.05.16.PM.mov
> > > I am using mac's voiceover though, not NVIDIA. Not sure if it's an issue that needs fixing or maybe a separate issue - seems out of scope with this issue. I'm fine to approve and would still like to have the conversation about using USWDS styling, but agree we'd have to keep this customization either way
> > 
> > 
> > I tried reproducing the issue, but couldn't. What command are you using? "Tab" and "Shift-Tab" or something else? Which browser are you on? Also, do you have the Keyboard Navigation setting turned on for your Mac? (System Settings -> Keyboard -> toggle "Keyboard navigation" on).
> 
> I'm using ctrl+alt+arrow keys with mac voiceover on chrome

Okay, I see it now.  This does seem like a separate problem.  It's also coming from the USWDS component.  Let's discuss whether or not to create an new bug in Slack.

PR #10199, opened 2025-04-01
Durable job for pre compute eligibility

### LINKED ISSUE #10070: The visibility editor when editing a screen's visibility as a civiform admin shows an eligibility message field
**Describe the bug**
As a civiform admin, when you edit the visibility of a screen, there is an option to include an eligibility message.

**To Reproduce**

1. Log in as an admin
2. Go to a program
3. Create a new screen or navigate to an existing screen
4. Select 'Edit visibility condition'
5. See the 'eligibility message' field

**Expected behavior**
There is no eligibility message field because this is the visibility editing.

**Screenshots**
Image

- MacOS
- Chrome

azhangggggg:
### Description

We added a column `eligibility_determination` to application table while `eligibility_determination` value for existing applications defaulted to `NOT_COMPUTED`
This durable job is to calculate `eligibility_determination` of existing applications. We will run this job every Sunday 2AM till all the existing applications are handled.

Step 4 of this [TDD](https://docs.google.com/document/d/1jsGEv1m9Ky2mbEbvTyZUbplFaPEiwNxopWPjm_XaRXs/edit?tab=t.0#heading=h.a2pzjye02740)

#### General

#### Durable jobs

- [x] Assigned two reviewers

### Issue(s) this completes

Part of #8852


server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public void run() {
+    logger.info("Starting job to calculate eligibility determination.");
+
+    try (Transaction jobTransaction = database.beginTransaction()) {
+      String filter =
+          """
+          eligibility_determination = 'NOT_COMPUTED'
+          """;
+      try (var query = database.find(ApplicationModel.class).where().raw(filter).findIterate()) {
```
gwendolyngoetz, line 54: (blocking) This should also filter for applications that have `LifecycleStage.Active` or `LifecycleStage.Obsolete`. That should cut down on a lot of empty draft applications.

It would be nice to not have to do obsoleted records as they are superseded by the active application, but they are returned in the API. Unless the PRD/TDD indicates we can skip them best to include them.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,77 @@
...
+            ReadOnlyApplicantProgramService roAppProgramService =
+                applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
+            EligibilityDetermination eligibilityDetermination =
+                applicantService.calculateEligibilityDetermination(
+                    programDefinition, roAppProgramService);
+            application.setEligibilityDetermination(eligibilityDetermination);
+            application.save();
+            jobTransaction.commit();
+          } catch (RuntimeException e) {
+            logger.error(e.getMessage(), e);
```
gwendolyngoetz, line 71: (blocking) Can you add the application id to the error message.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,77 @@
...
+            ProgramModel pm = application.getProgram();
+            ProgramDefinition programDefinition = pm.getProgramDefinition();
+            ReadOnlyApplicantProgramService roAppProgramService =
+                applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
+            EligibilityDetermination eligibilityDetermination =
+                applicantService.calculateEligibilityDetermination(
+                    programDefinition, roAppProgramService);
+            application.setEligibilityDetermination(eligibilityDetermination);
+            application.save();
+            jobTransaction.commit();
```
gwendolyngoetz, line 69: (blocking) I believe this will throw when you try to commit the second application since the transaction is started outside the loop.

gwendolyngoetz (dismissed):
Do you have a sense of how long it would take to run a large chunk of records. Filtering out the drafts will significantly help, but it would be good to have a rough idea of how long to expect this to run against a prod database as the startup jobs block the Civiform from finishing loading.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
shane-exygy: Anna you manually tested this correct?  I assume you didn't see Gwen's suggested error?

I agree it seems like this shouldn't work.

shane-exygy: Sorry correction, I think the expected behavior is that subsequent commit() are no-ops, so I would think only the first record is being changed.

shane-exygy: In reading move I think subsequent calls may be no-ops which would imply that only the first Application is updated.  But the test seems to say otherwise.

server/test/durablejobs/jobs/CalculateEligibilityDeterminationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    ApplicationModel firstApp =
+        resourceCreator.insertActiveApplication(
+            resourceCreator.insertApplicantWithAccount(), program);
+    ApplicationModel secondApp =
+        resourceCreator.insertActiveApplication(
+            resourceCreator.insertApplicantWithAccount(), program);
+    ApplicationModel thirdApp =
+        resourceCreator.insertActiveApplication(
+            resourceCreator.insertApplicantWithAccount(), program);
+
```
shane-exygy, line 103: **issue:** Verify the eligibility Determination is the expected default value to ensure it is't NO_ELIGIBILITY_CRITERIA from the start.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
azhangggggg: Yes I've manually tested this with multiple applications. I didn't see suggested error. But looking over your suggestions now.

azhangggggg: Yeah you're right it was throwing an exception. I didn't notice it in the log cuz the applications were executed correctly. Made changes to code to avoid the exception. Good catch!

Screenshot 2025-04-23 at 10 38 59 AM
Screenshot 2025-04-23 at 10 41 12 AM
Screenshot 2025-04-23 at 10 41 28 AM

server/test/durablejobs/jobs/CalculateEligibilityDeterminationTest.java
azhangggggg: Added verifying the eligibility determination step before the job ran.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
auzlee: +1

server/app/modules/DurableJobModule.java
```diff
@@ -150,6 +153,13 @@ public DurableJobRegistry provideRecurringDurableJobRegistry(
                 publicStorageClient, versionRepository, persistedDurableJob),
         new RecurringJobExecutionTimeResolvers.ThirdOfMonth2Am());
 
+    durableJobRegistry.register(
+        DurableJobName.CALCULATE_ELIGIBILITY_DETERMINATION_JOB,
+        JobType.RECURRING,
+        persistedDurableJob ->
+            new CalculateEligibilityDeterminationJob(applicantService, persistedDurableJob),
+        new RecurringJobExecutionTimeResolvers.Sunday2Am());
```
auzlee, line 161: iiuc this is going to process 5000 applications per week. We should either up the limit, if we determine that it won't slow down the server for a long time, or make a "daily at 3am" or "every 3 hours at :23" TimeResolver so this doesn't take months to work its way through all the old applications.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
auzlee: (also not blocking at all but it should be possible to build the filter in the ebeans syntax instead of as a raw strong)

server/app/modules/DurableJobModule.java
gwendolyngoetz: We don't need to limit the job to 5000. Only one instance of the job will run at a time in the background.

Do do that something like would work:

```
while true:
  begin tx
  find applications
    if hasNext is false
      break exiting outer while
  iterate batch of n applications
  save    
  commit/rollback where needed.
```

You can also choose to end the job if too many errors have counted up which might indicate a bigger problem.

shane-exygy: @azhangggggg does the PR description need updating?  It mentions 5k still

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,91 @@
...
+            ProgramDefinition programDefinition = pm.getProgramDefinition();
+            ReadOnlyApplicantProgramService roAppProgramService =
+                applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
+            EligibilityDetermination eligibilityDetermination =
+                applicantService.calculateEligibilityDetermination(
+                    programDefinition, roAppProgramService);
+            application.setEligibilityDetermination(eligibilityDetermination);
+            application.save();
+          } catch (RuntimeException e) {
+            errorCount++;
```
shane-exygy, line 73: **suggestion (non-blocking):** Consider stopping after N errors, maybe 10.  If we're getting a lot of Runtime errors that is likely an indication of a larger issue that should be addressed before continuing to process more records.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,91 @@
...
+            logger.error("Error message {}", e.getMessage());
+          }
+        }
+      }
+
+      if (errorCount == 0) {
+        logger.info("Job succeeded");
```
shane-exygy, line 80: **suggestion:** Swap these two lines so we know commit() succeeded before logging it.

server/test/durablejobs/jobs/CalculateEligibilityDeterminationTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+import repository.ResetPostgres;
+import services.applicant.ApplicantService;
+import services.applications.ApplicationService;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CalculateEligibilityDeterminationTest extends ResetPostgres {
```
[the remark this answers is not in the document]
azhangggggg: The reason that we're getting NO_ELIGIBILITY_CRITERIA is that none of the test programs has an eligibility criteria set. But it does verify that [CalculateEligibilityDetermination()](https://github.com/civiform/civiform/blob/b2d86f32d7523f30543b4438a3215e0aabd8164c/server/app/services/applicant/ApplicantService.java#L493) was called. There are other tests like [this](https://github.com/civiform/civiform/blob/6c209846a280ccc40b52abae42560583c2292387/server/test/services/applicant/ApplicantServiceTest.java#L1775) testing CalculateEligibilityDetermination() in more details. For this PR we're just trying to see if the method was called imo.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
azhangggggg: yes, done

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,97 @@
...
+          }
+        }
+      }
+
+      if (errorCount == 0) {
+        jobTransaction.commit();
+        logger.info("Job succeeded");
+      } else {
+        logger.error(
+            "Failed to compute eligibility determination for existing applications. Error: {}",
```
shane-exygy, line 91: **suggestion:** Since we're stopping after N errors, I'd suggest updating the phrasing to imply that, as well indicate that the value is an error count "Error: 9" is not that readable.

Perhaps: "Stopping early after {} errors."

server/test/durablejobs/jobs/CalculateEligibilityDeterminationTest.java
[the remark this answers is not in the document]
shane-exygy: I think we need to ensure that a valid eligibility update is done at least once.  Otherwise how does the test ensure changes to the underlying code don't have a bug that simply sets everything to NO_ELIGIBILITY_CRITERIA, or the loop accidentally uses the first application in every loop etc.

The test shouldn't care how the operation is done (Eg verify a method is called), but that the correct outcome is achieved, and we have no situation in which eligibility is actually computed here.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,97 @@
...
+            ProgramDefinition programDefinition = pm.getProgramDefinition();
+            ReadOnlyApplicantProgramService roAppProgramService =
+                applicantService.getReadOnlyApplicantProgramService(application, programDefinition);
+            EligibilityDetermination eligibilityDetermination =
+                applicantService.calculateEligibilityDetermination(
+                    programDefinition, roAppProgramService);
+            application.setEligibilityDetermination(eligibilityDetermination);
+            application.save();
+          } catch (RuntimeException e) {
+            errorCount++;
```
shane-exygy, line 75: **issue (blocking):** Most of these error handling lines aren't tested.

server/app/modules/DurableJobModule.java
shane-exygy: I think it's still not right

"We will run this job once and have it handle as many applications in batches as it can in one run."

As I understand it .commit() is only called if there are 0 errors, so from the outside we're not updating applications in batches, they'll be updated all or nothing.  and if there's an error it'll have to be run again assuming the error is resolvable with a re-run.

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
```diff
@@ -0,0 +1,114 @@
...
+  public PersistedDurableJobModel getPersistedDurableJob() {
+    return persistedDurableJobModel;
+  }
+
+  @Override
+  public void run() {
+    logger.info("Starting job to calculate eligibility determination.");
+
+    try (Transaction jobTransaction = database.beginTransaction(TxIsolation.SERIALIZABLE)) {
+      jobTransaction.setBatchMode(true);
```
shane-exygy, line 52: **note (non-blocking):** FYI the default batch size is 20, you may want to increase that via `jobTransaction.setBatchSize()`

"The batch buffer size default value is set via databaseConfig.persistBatchSize which defaults to 20."

https://ebean.io/docs/transactions/batch

shane-exygy (approved):

server/app/durablejobs/jobs/CalculateEligibilityDeterminationJob.java
azhangggggg: Setting the batch size to 50, avoiding increasing memory usage too much.

shane-exygy (approved):

PR #10217, opened 2025-04-03
Clean up TextFormatter Class

### LINKED ISSUE #10127: Refactor TextFormatter class to be more readable
The `TextFormatter` class has gotten pretty complex and it's not always clear when to use which methods. We should think about how to refactor this class to make it more clear how to use it. For example, we could rename the methods to their intended use eg `formatTextForAdmin`, `formatTextForApplicant`, and `formatTextForApplicantLegacy`.

Also remove unnecessary calls to `toLowerCase` when we pass in the ariaLabel.

rockycodes:
### Description

- Rename some methods in `TextFormatter.java` for clarity
- Remove extraneous calls to `toLowerCase` when passing in the aria label from the message key
- Remove `ariaLabel` class var and pass it as a param instead
- Remove unused `removeMarkdown` method

#### General

### Issue(s) this completes

Fixes #10127; .


server/app/views/questiontypes/StaticContentQuestionRenderer.java
```diff
@@ -51,7 +51,8 @@ public DivTag render(ApplicantQuestionRendererParams params) {
           TextFormatter.formatText(
               question.getQuestionText(),
               /* preserveEmptyLines= */ true,
-              /* addRequiredIndicator= */ false);
+              /* addRequiredIndicator= */ false,
+              /* ariaLabel= */ "opens in a new tab");
```
rockycodes, line 55: This call is for admin, but it's the only situation where we need to preserve empty lines on the admin side so I opted to use the applicant method and pass in the aria label. If there were more situations on the admin side like this, it would warrant an admin specific method, but I don't think it's worth it for this one case.

server/app/views/questiontypes/StaticContentQuestionRenderer.java
```diff
@@ -51,7 +51,8 @@ public DivTag render(ApplicantQuestionRendererParams params) {
           TextFormatter.formatText(
               question.getQuestionText(),
               /* preserveEmptyLines= */ true,
-              /* addRequiredIndicator= */ false);
+              /* addRequiredIndicator= */ false,
+              /* ariaLabelNewTab= */ "opens in a new tab");
```
rockycodes, line 55: This call is for admin, but it's the only situation where we need to preserve empty lines on the admin side so I opted to use the applicant method and pass in the aria label. If there were more situations on the admin side like this (where we need empty lines or a required indicator), it would warrant an admin specific method, but I don't think it's worth it for this one case.

rockycodes: "opens in a new tab" is the same string as the english default message, but I'm not sure how to pull it out of the messages file, since we don't have the `messages` object on the admin side. Do you know how?

server/app/views/components/TextFormatter.java
```diff
@@ -21,77 +21,52 @@ public final class TextFormatter {
...
-  }
-
-  /**
-   * Passes provided text through Markdown formatter, generating an HTML String. This function does
-   * not add translated aria labels to links and should only be used in admin facing views.
-   */
-  static String formatTextToSanitizedHTML(
-      String text, boolean preserveEmptyLines, boolean addRequiredIndicator) {
+  /** Passes provided text through Markdown formatter, generating an HTML String. */
+  static String internalFormatTextToSanitizedHTML(
```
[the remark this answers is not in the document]
rockycodes: Oh yeah I see your point! I can't remember why this was initially factored out, but I think you're right it no longer needs to be.

server/app/views/components/TextFormatter.java
```diff
@@ -21,77 +21,52 @@ public final class TextFormatter {
...
-   * Passes provided text through Markdown formatter. This function does not add translated aria
-   * labels to links and should only be used in admin facing views.
-   */
-  public static ImmutableList<DomContent> formatText(
-      String text, boolean preserveEmptyLines, boolean addRequiredIndicator) {
-    ImmutableList.Builder<DomContent> builder = new ImmutableList.Builder<DomContent>();
-    builder.add(rawHtml(formatTextToSanitizedHTML(text, preserveEmptyLines, addRequiredIndicator)));
-    return builder.build();
+  public static ImmutableList<DomContent> formatTextForAdmins(String text) {
+    return formatText(text, false, false, "opens in a new tab");
```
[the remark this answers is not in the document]
rockycodes: good call

server/app/services/applicant/question/StaticContentQuestion.java
```diff
@@ -42,7 +42,7 @@ public ImmutableList<Path> getAllPaths() {
   }
 
   public String getFormattedTextForRendering(String ariaLabel) {
```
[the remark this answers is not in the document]
rockycodes: I think this was the only call site remaining with this generic var name, but lmk if there are others you see that I didn't catch.

server/app/views/components/TextFormatter.java
```diff
@@ -21,77 +21,52 @@ public final class TextFormatter {
...
-   * This function should be used when the resulting html will be applicant facing.
+   * Passes provided text through Markdown formatter, returning a String with the sanitized HTML.
+   * This is used by Thymeleaf to render strings containing markdown.
    */
-  public static String formatTextToSanitizedHTMLWithAriaLabel(
+  public static String formatTextToSanitizedHTML(
       String text,
       boolean preserveEmptyLines,
       boolean addRequiredIndicator,
       String ariaLabelNewTab) {
```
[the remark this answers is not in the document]
rockycodes: updated

server/app/views/components/TextFormatter.java
```diff
@@ -30,43 +30,37 @@ public static ImmutableList<DomContent> formatText(
...
    * addRequiredIndicator set to false. This function does not add translated aria labels to links
    * and should only be used in admin facing views.
    */
   public static ImmutableList<DomContent> formatTextForAdmins(String text) {
-    return formatText(text, false, false, "opens in a new tab");
+    return formatText(
+        text,
+        /* preserveEmptyLines= */ false,
+        /* addRequiredIndicator= */ false,
+        "opens in a new tab");
```
shane-exygy, line 52: **suggestion:** 
```suggestion
        "/* ariaLabelForNewTabs= */ opens in a new tab");
```

shane-exygy (approved):

server/app/views/questiontypes/StaticContentQuestionRenderer.java
shane-exygy: Given this isn't an injected class it's probably not worth the effort.

ezraeinhorn624 (approved):

shane-exygy (approved):

Issue #10188, opened 2025-04-01
Audit Respository classes for transaction correctness


shane-exygy:
Tracking issue to look at all Repository classes and ensure any read-write code is in a Transaction

- [x] [AccountRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/AccountRepository.java)
- [x] [ApiKeyRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ApiKeyRepository.java)
- [x] [ApplicationEventRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ApplicationEventRepository.java)
- [x] [ApplicationRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ApplicationRepository.java)
- [x] [ApplicationStatusesRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ApplicationStatusesRepository.java)
- [x] [CategoryRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/CategoryRepository.java)
- [x] [ExportServiceRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ExportServiceRepository.java)
- [x] [HealthCheckRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/HealthCheckRepository.java)
- [x] [PersistedDurableJobRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/PersistedDurableJobRepository.java)
- [x] [ProgramRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ProgramRepository.java)
- [x] [QuestionRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/QuestionRepository.java)
- [x] [ReportingRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/ReportingRepository.java)
- [x] [SettingsGroupRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/SettingsGroupRepository.java)
- [x] [StoredFileRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/StoredFileRepository.java)
- [x] [VersionRepository](https://github.com/civiform/civiform/blob/main/server/app/repository/VersionRepository.java)

PR #10231, opened 2025-04-07
Add TransactionManager.executeWithRetry and usage

shane-exygy:
### Description
TransactionManger:

- Extract `execute(Supplier)` from the existing method.
- Add `execute(Runnable)`
- Add `executeWithRetry(Supplier)`

Update `AccountRepository` to use the new methods.

Note: I'll rename `executeInTransaction` in a subsequent PR to not cloud this one.

#### General

### Issue(s) this completes

Addresses #[10188](https://github.com/civiform/civiform/issues/10188)


server/test/repository/TransactionManagerTest.java
```diff
@@ -120,6 +176,84 @@ public void executeInTransaction_rollsBackTransactionSuccessfully() {
...
+            // Update the account in a different Transaction (requiresNew)
+            // before the current one finishes to trigger a serialization
+            // exception in the outer transaction.
+            try (Transaction innerTransaction =
+                DB.beginTransaction(TxScope.requiresNew().setIsolation(TxIsolation.SERIALIZABLE))) {
+              AccountModel innerAccount = accountRepo.lookupAccount(accountId).orElseThrow();
+              innerAccount.setEmailAddress(innerEmail);
+              innerAccount.save();
+              innerTransaction.commit();
+              LOGGER.error("set inner email");
```
rockycodes, line 238: question (non-blocking): are you leaving this logger in intentionally?

rockycodes (approved):
lgtm! great tests -- they're super cool and helped me understand what was going on.

server/test/repository/TransactionManagerTest.java
shane-exygy: nope!   thank you, removed it.

server/app/repository/AccountRepository.java
```diff
@@ -142,9 +144,14 @@ public CompletionStage<Optional<AccountModel>> lookupAccountByEmailAsync(String
    * we create one.
    */
   private ApplicantModel getOrCreateApplicant(AccountModel account) {
-    Optional<ApplicantModel> applicantOpt =
-        account.getApplicants().stream().max(Comparator.comparing(ApplicantModel::getWhenCreated));
-    return applicantOpt.orElseGet(() -> new ApplicantModel().setAccount(account).saveAndReturn());
+    return transactionManager.executeWithRetry(
```
dkatzz, line 147: **question (non-blocking):** Is the reason we know to call executeWithRetry here vs execute because it involves reading then writing, so there's a chance of a concurrency error the first time?

dkatzz (approved):

server/app/repository/AccountRepository.java
shane-exygy: What you describe is the reason to have it in a transaction at all, as this could end up making two Applicants.  (That may also just fail due to key collisions though).

The reason I chose Retry is that this is in effect a read, and it doesn't really care who made the applicant, but if it exists it should return it.  So if say the user loaded the url twice though extra clicking of a URL, etc and two of these ran in parallel, there's no reason to throw an exception due to a key collision in the DB or otherwise not return the applicant the other thread made.

dkatzz: Thank you, makes sense! I think it would be great to add any best practices like this to the wiki at some point. I guess in this case it would be when looking for a value in the database or creating it if it doesn't exists, we should use `executeWithRetry`. But it probably is a bit case by case as well.

PR #10312, opened 2025-04-15
Giving the button aria role to links that behave as buttons

### LINKED ISSUE #9911: [a11y][Mobile][iOS][WCAG 2.4.8 ] [v2.1 AA] Focus indicator remains persistent on some components
OBSERVED RESULTS:
--

While using VoiceOver on iOS, there are certain components where the focus indicator persist even when the element is not currently being focused by the user, as shown in the following video where linear navigation is conducted and some components display incorrectly the focus indicator

[iOS](https://drive.google.com/file/d/1bjGTVI9FoAudxxSew2cZPLYnK8iAC_5x/view?usp=drive_link)

EXPECTED RESULTS:
--

Information about the user's location within the page should be accurate as to avoid confusion about the current location of selection of components.

STEPS TO REPRODUCE:
--
0. Turn on the screen reader (VoiceOver)
1. Start the navigation on a linear manner 
2. Pay attention to the focus indicator, notice how there are certain elements that display a persistent focus indicator.
3. Focus indicator only shifts to the next component once a new component with the capacity of displaying the focus indicator is selected. 

RELEVANT GAR SHEET:
--

[GAR sheet](https://docs.google.com/spreadsheets/d/1mgAhVZ3tuJ0O2EeB5JZ0HAvSrxVeIUjlJxVAhu3JH7M/edit?gid=1240238308#gid=1240238308)

ADDITIONAL INFORMATION:
--

[Success Criterion 2.4.8 Location](https://www.w3.org/TR/WCAG22/#focus-visible)

ENVIRONMENT DETAILS:
--

---
- Devices: iPhone 14 plus: iOS 18.2.1
- Screen reader name: VoiceOver
- Browser name: Chrome Mobile: v133.0.69
---

melanie-exygy:
### Description

On iOS when using the VoiceOver screen reader, keyboard focus (blue outline) doesn't follow the VoiceOver cursor (black outline) to anchor tags.  However, keyboard focus does follow the cursor to buttons.  This PR assigns the button aria role to any anchor tags that act as buttons.  The button role should be used "when the user- action causes a change in either back-end or the front-end of the website." (see https://www.digitala11y.com/links-vs-buttons-a-perennial-problem/).  By doing this, we're solving the keyboard focus bug for those particular buttons.

However, some anchor tags in CiviForm need to keep the implicit link role because they function as links (they take the user to a new location within the page or to another page).  Since screen reader users often navigate by using the list of links in a page, if a link behaves like a link, it needs to keep that link role so that it will show up in the screen reader link list.  For those links, we will continue to have the problem on iOS of keyboard focus not syncing with the VoiceOver cursor.  However, this problem can be replicated on other reliable websites such as w3c.org and the USWDS site, as shown in #9911.  Therefore, this is expected VoiceOver behavior on iOS and we won't attempt to fix it.  

1. Check out the `melanie-exygy/focus-links` branch.
2. Use these steps to open your local CF on your iPhone:  https://prowe214.medium.com/tip-how-to-view-localhost-web-apps-on-your-phone-ad6b2c883a7c
3. Turn VoiceOver on.
4. Swipe right to move from one interactive element to the next.  
5. When you get to the "Log in" button in the nav bar, the blue focus outline surrounds the element.

### Issue(s) this completes

Fixes #9911

### DIFF
--- a/server/app/views/applicant/NavigationFragment.html
+++ b/server/app/views/applicant/NavigationFragment.html
@@ -224,7 +224,6 @@
 
 <div th:fragment="showDebugTools">
   <a
-    type="button"
     class="usa-button usa-button--outline margin-right-0"
     id="debug-content-modal-button"
     href="#debug-content-modal"
@@ -247,7 +246,7 @@
 <th:block th:fragment="loginOrLogout">
   <th:block th:if="${isGuest}">
     <a
-      type="button"
+      role="button"
       class="usa-button usa-button--outline margin-right-0"
       th:text="#{button.login}"
       id="login-button"
@@ -258,7 +257,7 @@
 
   <th:block th:if="${!isGuest}">
     <a
-      type="button"
+      role="button"
       class="usa-button usa-button--outline margin-right-0"
       th:text="#{button.logout}"
       id="logout-button"
@@ -325,6 +324,7 @@
               class="mobile-lg:grid-col-6 desktop:grid-col-auto usa-footer__primary-content"
             >
               <a
+                role="button"
                 class="usa-footer__primary-link"
                 th:href="${adminLoginUrl}"
                 th:text="#{link.adminLogin}"
--- a/server/app/views/NorthStarBaseView.java
+++ b/server/app/views/NorthStarBaseView.java
@@ -128,7 +128,7 @@ public abstract class NorthStarBaseView {
     // that will be embedded in the guest alert in the header.
     context.setVariable(
         "endSessionLinkHtml",
-        "<a id=\"logout-button\" class=\"usa-link\" href=\""
+        "<a id=\"logout-button\" class=\"usa-link\" role=\"button\" href=\""
             + logoutLink
             + "\">"
             + messages.at(MessageKey.END_YOUR_SESSION.getKeyName())
--- a/server/app/views/applicant/CreateAccountAlertFragment.html
+++ b/server/app/views/applicant/CreateAccountAlertFragment.html
@@ -19,6 +19,7 @@
       <div class="grid-col-auto">
         <a
           class="usa-button usa-button--inverse"
+          role="button"
           th:text="${createAccountLinkText}"
           id="create-account"
           th:href="${createAccountLink}"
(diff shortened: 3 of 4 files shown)


server/app/views/applicant/NavigationFragment.html
```diff
@@ -224,7 +224,6 @@
 
 <div th:fragment="showDebugTools">
   <a
-    type="button"
```
melanie-exygy, line 227: **note:** `type="button"` is not valid markup.   The type of an anchor element should be the media type of the linked document (“text/html”, for example).

rockycodes: question (non-blocking): should this still have `role=button`? Or it doesn't matter since this is only shown to admin?

rockycodes (approved):
This looks good to me and solves the focus issue for the two elements you mentioned, but I noticed one of the "Start an application" buttons on the program overview page is still being treated as a link, so we might need to add it there as well? I'll put some time on for us to look at this together.

melanie-exygy:
> This looks good to me and solves the focus issue for the two elements you mentioned, but I noticed one of the "Start an application" buttons on the program overview page is still being treated as a link, so we might need to add it there as well? I'll put some time on for us to look at this together.

I think "Start an application" leans more towards link behavior because it takes the user to another page, but one could argue that it does allow the user to take an action (starting an application).  Sometimes deciding between the two roles seems very subjective.  

With the other "Start an application with an account" button, I gave it the role button by modifying the `CreateAccountAlertFragment` in this PR.  That one does lead the user to sign in, which seems like more of a user action (so, button role).  The "Start an application as a guest" link below it is the only one that isn't quite right, IMO, because that one does take the user to a new location, just like the "Start an application" button does.  But, that's part of the `CreateAccountAlertFragment`, so any modifications there apply to other uses of that fragment, such as the one on the Upsell page.  That's why I decided to leave it as a button.  My reasoning is that it isn't a problem because a screen reader user will still have the option to use the "Start an application" link above, so they won't be missing anything.  But, let me know if you think that's not going to work.

Issue #9904, opened 2025-03-06
Implement setting custom primary colors in North Star UI


watermandrewc:
Allow admins to set primary colors which will be used for branding in CiviForm in the North Star applicant UI.

TDD: https://docs.google.com/document/d/10najhbuM_aomsWSnDzXdSQKUO2aOLoBRrNDgQto-yPQ/edit?tab=t.0#heading=h.bknox3ib8wpb

Feature flag tracked in #9882

dkatzz:
@watermandrewc we can close this right?

watermandrewc:
Yep! Closing

PR #10320, opened 2025-04-15
Add color contrast check when setting theme colors

watermandrewc:
### Description

Add an accessibility check for color contast when setting theme colors.

Validate theme colors in the settings page by checking if they have a contrast ratio of at least 4.5:1 with white, following [WCAG guidelines](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) and the [formula for relative luminance](https://www.w3.org/TR/WCAG21/#dfn-relative-luminance).

[9904](https://github.com/civiform/civiform/issues/9904)

#### General

- [x] Noted in the PR description which, if any, code in this PR was generated by AI.

#### User visible changes

1. Sign in as a CiviForm admin and go to settings page.
2. Test various colors and verify that if they meet a contrast ratio of 4.5:1 with #FFFFFF, that the setting saves properly, and if not, an error message is shown and the setting doesn't save. You can use https://webaim.org/resources/contrastchecker/ to check the contrast ratio for a given color.


server/app/services/settings/SettingsService.java
```diff
@@ -189,6 +190,22 @@ private ImmutableMap<String, SettingsGroupUpdateResult.UpdateError> validateSett
 
                     if (error.isPresent()) {
                       validationErrors.put(settingDescription.variableName(), error.get());
+                    } else {
+                      if (settingDescription.variableName().equals("THEME_COLOR_PRIMARY")
+                          || settingDescription.variableName().equals("THEME_COLOR_PRIMARY_DARK")) {
+                        // Only allow admins to set theme colors that have a contrast ratio of 4.5:1
```
dkatzz, line 196: **thought:** IIUC this checks if the colors contrast with white, but should we also be checking that they contrast with each other?

browser-test/src/admin/admin_settings.test.ts
```diff
@@ -64,4 +64,61 @@ test.describe('Managing system-wide settings', () => {
...
+  }) => {
+    await adminSettings.gotoAdminSettings()
+
+    await test.step('contrast ratio not met on primary color', async () => {
+      await adminSettings.setStringSetting('THEME_COLOR_PRIMARY', '#19baff')
+      await adminSettings.saveChanges(
+        /* expectUpdated= */ false,
+        /* expectError= */ true,
+      )
+      await expect(
```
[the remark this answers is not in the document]
watermandrewc: Added! This will make it easier if we update the strings (which we'll likely want to do -- working w/ UX to fine-tune them at the moment)

server/app/services/ColorUtil.java
```diff
@@ -0,0 +1,54 @@
...
+    int blueCode = Integer.valueOf(hex.substring(5, 7), 16);
+
+    double rsrgb = getSRGB(redCode);
+    double gsrgb = getSRGB(greenCode);
+    double bsrgb = getSRGB(blueCode);
+
+    return (rsrgb * 0.2126) + (gsrgb * 0.7152) + (bsrgb * 0.0722);
+  }
+
+  /** Returns the sRGB value for a given R,G,B value. */
```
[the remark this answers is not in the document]
watermandrewc: Added another link to the relative luminance section in the w3 docs, which explains how to calculate it

server/app/services/ColorUtil.java
```diff
@@ -0,0 +1,54 @@
...
+        : normalizedColorValue / 12.92;
+  }
+
+  /** Converts a 3-digit hex color code to a 6-digit one. */
+  private static String convert3DigitHexTo6Digits(String hex) {
+    char redCode = hex.charAt(1);
+    char greenCode = hex.charAt(2);
+    char blueCode = hex.charAt(3);
+
+    return String.format(
```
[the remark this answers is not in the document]
watermandrewc: Ah thanks this is much better!

server/app/services/settings/SettingsService.java
watermandrewc: AFAIU for accessibility, that shouldn't really matter; we mainly just want to check for contrast to make sure that the text is readable, so we're essentially checking font color vs. button/background color here. @melanie-exygy what do you think?

melanie-exygy:
As I'm testing, if I enter something invalid, like `#000`, I see this error message:  "Invalid input, must match ^#(?:[0-9a-fA-F]{3}){1,2}$".  I wonder if that regex will just be more confusing than anything.  Might be best to put something like "Must be # followed by 6 digits."

server/app/services/settings/SettingsService.java
melanie-exygy: Yes, I agree, the color contrast is between the button background and the font (white) and the button background and the page background (also white).  IMO, we want there to be enough of a difference between the two theme colors so that the user notices the hover state, but that contrast isn't a WCAG requirement, as far as I can tell.

melanie-exygy (approved):

server/app/services/settings/SettingsService.java
dkatzz: Makes sense, thank you!

dkatzz (approved):

PR #10327, opened 2025-04-16
Make shouldLogoutUser fully async

nb1701:
Previously, we were doing `profile.getAccount().join()`. This was causing the main thread to lock up with the pool all waiting on `join()`s that were never to finish. By making this truly async and running the necessary functions in the `DatabaseExecutionContext`, this prevents the thread starvation from occurring.

### Further details

On versions from v2.33.0 on, we were finding that instances may stop responding and require a restart.  After much intensive debug, particularly using Postman to hit a prod instance with both valid and invalid endpoints, we found a few things.  First, we are doing the callback to create guest accounts on `/error` and invalid endpoints when we shouldn't be.  We'll fix this in a different PR.  

[Here](https://github.com/civiform/civiform/blob/87e9d8080b5a1619e1dcd5c57790cf97b7a60acd/server/app/filters/ValidAccountFilter.java#L62) we call `shouldLogoutUser`.  In that function, we [synchronously call](https://github.com/civiform/civiform/blob/87e9d8080b5a1619e1dcd5c57790cf97b7a60acd/server/app/filters/ValidAccountFilter.java#L82) `profileUtils.validCiviFormProfile`. In [that function](https://github.com/civiform/civiform/blob/87e9d8080b5a1619e1dcd5c57790cf97b7a60acd/server/app/auth/ProfileUtils.java#L96), we call `profile.getAccount().join()`.  Now profile.getAccount() [does indeed run in the dbContext](https://github.com/civiform/civiform/blob/87e9d8080b5a1619e1dcd5c57790cf97b7a60acd/server/app/auth/CiviFormProfile.java#L95), however, the `.join()` waiting for that to come back is blocking the thread running in the default context, which is the context that also handles requests coming in.

On every single request now, we are having to fetch settings from the database in order to check if the session timeout feature flag is enabled.  This is not a small amount of data (although not huge either).  But this is not something we were doing before and takes some time to fetch.  As best we can tell, under sufficient load, you can get into a state where:

- All the database threads are in the middle of fetching settings from the database (or most of them are, one or two may be doing other things)
- The threads that requested that settings data are parked waiting for the database callback
- The threads actively being executed are all completely blocked on `profile.getAccount().join()`.
- We are unable to dispatch new database execution threads, because the active database threads are waiting to return their settings data.
- The threads waiting for settings data are unable to become active to accept the data back because all active threads in the thread pool are blocked.

By making the `profileUtils.validCiviFormProfile` function fully asyncrhonous and not doing a `.join()`, this allows the async handler to park the thread, let the threads waiting for the settings data to execute and receive the data, and get things moving again.

In the future, we are going to look into modifying how we handle settings so that 1) we cache the data and don't fetch it from the database each time, and 2) we don't inject settings in every request to make the code more efficient and reduce the amount of data being passed around.

#### General

The way we tested this was to point Postman at prod (or staging) instance, and run a performance test with 100 virtual users, hitting both `/programs` and `/callback?client_name=%2Fetc%2Fpasswd%00.html`, as this is one of the ones we see with people hitting CiviForm with Metasploit and results in a server error.  In reality, it probably doesn't matter too much which endpoints you use.

You can also test this locally using `bin/run-dev --prod`.  It may work to do this as-is, but I also constrained its resources by setting the following in docker-compose.yml.
```
db:
    image: postgres:16.8
    deploy:
      resources:
        limits:
          cpus: '2'
          memory: 1G
```
and
```
civiform:
    deploy:
      resources:
        limits:
          cpus: '1'
          memory: 4G
```
in order to simulate how it runs in AWS by default.

Within a minute of running the Postman perf test, all requests to the server should stop responding.  With the fix, it should continue to respond.

Addresses: #10330

### DIFF
--- a/server/test/filters/ValidAccountFilterTest.java
+++ b/server/test/filters/ValidAccountFilterTest.java
@@ -12,7 +12,6 @@ import auth.ProfileUtils;
 import java.time.Clock;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import javax.inject.Provider;
 import models.AccountModel;
 import org.junit.Before;
 import org.junit.Test;
@@ -35,7 +34,6 @@ public class ValidAccountFilterTest extends WithApplication {
   private AccountModel mockAccount;
   private SessionTimeoutService sessionTimeoutService;
   private Clock clock;
-  private DatabaseExecutionContext dbContext;
 
   @Before
   public void setUp() {
@@ -43,17 +41,16 @@ public class ValidAccountFilterTest extends WithApplication {
     settingsManifest = mock(SettingsManifest.class);
     sessionTimeoutService = mock(SessionTimeoutService.class);
     clock = mock(Clock.class);
-    Provider<SettingsManifest> settingsManifestProvider = () -> this.settingsManifest;
-    dbContext = instanceOf(DatabaseExecutionContext.class);
 
     filter =
         new ValidAccountFilter(
             profileUtils,
-            settingsManifestProvider,
+            () -> settingsManifest, // Provider<SettingsManifest>
             mat,
             clock,
-            () -> sessionTimeoutService,
-            dbContext);
+            () -> sessionTimeoutService, // Provider<SessionTimeoutService>
+            () -> instanceOf(DatabaseExecutionContext.class) // Provider<DatabaseExecutionContext>
+            );
 
     mockProfile = mock(CiviFormProfile.class);
     mockProfileData = mock(CiviFormProfileData.class);


server/test/filters/ValidAccountFilterTest.java
```diff
@@ -42,10 +44,16 @@ public void setUp() {
...
 
     filter =
         new ValidAccountFilter(
-            profileUtils, settingsManifestProvider, mat, clock, () -> sessionTimeoutService);
+            profileUtils,
+            settingsManifestProvider,
+            mat,
+            clock,
+            () -> sessionTimeoutService,
+            dbContext);
```
shane-exygy, line 56: **question:** this doesn't need "() -> " similar to the proceeding Provider?

shane-exygy (approved):

shane-exygy (approved):

PR #10328, opened 2025-04-17
Adding aria-label to differentiate "Remove file" links and giving role of button

### LINKED ISSUE #9467: Multiple file upload "Remove file" links need more precise label and should be buttons rather than links
### Checklist (remove this section after completion): 
- [X] Add the correct epic label to the 'epic' category 
- [ ] Add the correct status to the 'status' category
- [X] Complete the info below

Discovered when porting browser-test/src/applicant/questions/file.test.ts to North Star, the test "remove button has correct aria-labelled by" fails because there is no aria-labelledby class on the "Remove File" link.

For the test to pass as is, there needs to be an aria-labelledby class present in the link. 

#### Additional Context

Legacy HTML:

`<a __playwright_target__="call@400" href="link:///programs/2/blocks/1/removeFile/applicant-4%2Fprogram-2%2Fblock-1%2Ffile-upload.png/false" aria-labelledby="uploaded-file-1" class="text-blue-600 hover:text-blue-500 underline ml-2 shrink-0">Remove file</a>`

North Star HTML:

`<a __playwright_target__="call@394" href="link:///programs/3/blocks/1/removeFile/applicant-6%2Fprogram-3%2Fblock-1%2Ffile-upload.png/false" style="outline: rgb(0, 106, 177) solid 2px; background-color: rgba(111, 168, 220, 0.498);">Remove file</a>
`
#### Done when:

Either Sijia says we don't care about the aria-labelledby class in North Star, or someone adds it to North Star.

melanie-exygy:
### Description

On file upload questions, there is a "Remove file" link next to each uploaded file.  When multiple files are uploaded, this means the links have duplicated names.  A screen reader user would have no way of knowing which file they were removing.  This adds an aria label which includes the name of the file in the link label so that users can differentiate between links.  For example: "Remove puppy.jpg file".

Also, semantically these should be buttons, not links. The purpose of a link would be to take the user to a different location, whereas the purpose of a button is for the user to take action.  So, this PR gives the anchor tags the aria role of "button" so that screen readers will treat them as buttons.

1. Create a program with a file upload question.
2. Turn the NS flag on.
3. Apply to the program.
4. When you get to the file upload question, upload a couple of files.
5. Turn the screen reader on.
6. Tab to the "Remove file" button.
7. It should be announced as a button and the screen reader should say "Remove 'filename' file".

### Issue(s) this completes

Fixes #9467

### DIFF
--- a/server/conf/i18n/messages
+++ b/server/conf/i18n/messages
@@ -160,6 +160,8 @@ link.home=Home
 link.applicationForProgram=Application for {0}
 # Anchor which when clicked removes a file the user has previously uploaded.
 link.removeFile=Remove file
+# Screen reader text for anchor which when clicked removes a file the user has previously uploaded. {0} is the file name.
+link.removeFileSr=Remove {0} file
 # The title of a pop-up informing the user that they tried to go to the review page but there were errors in the information that they inputted
 modal.errorSaving.review.title=Questions on this page are not complete. Would you still like to leave and begin reviewing?
 # The title of a pop-up informing the user that they tried to go to the previous page but there were errors in the information that they inputted
--- a/server/conf/i18n/messages
+++ b/server/conf/i18n/messages
@@ -160,7 +160,7 @@ link.home=Home
 link.applicationForProgram=Application for {0}
 # Anchor which when clicked removes a file the user has previously uploaded.
 link.removeFile=Remove file
-# Screen reader text for anchor which when clicked removes a file the user has previously uploaded. {0} is the file name.
+# Screen reader text for a link which when clicked removes a file the user has previously uploaded. {0} is the file name.
 link.removeFileSr=Remove {0} file
 # The title of a pop-up informing the user that they tried to go to the review page but there were errors in the information that they inputted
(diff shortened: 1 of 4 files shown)


server/conf/i18n/messages
```diff
@@ -160,6 +160,8 @@ link.home=Home
 link.applicationForProgram=Application for {0}
 # Anchor which when clicked removes a file the user has previously uploaded.
 link.removeFile=Remove file
+# Screen reader text for anchor which when clicked removes a file the user has previously uploaded. {0} is the file name.
```
shane-exygy, line 163: **suggestion:** 
```suggestion
# Screen reader text for an anchor which when clicked removes a file the user has previously uploaded. {0} is the file name.
```

shane-exygy (approved):

orwant (approved):

server/conf/i18n/messages
melanie-exygy: Done

shane-exygy (approved):

